PR #5724, opened 2016-06-13
Correct UnitarySystem and Multispeed Heat Pump heat recovery calculation to honor maximum outlet temperature

rraustad:
## Pull request overview

Fixes 5516 defect for Multispeed coil heat recovery calculation where heat transfer still occurs when outlet water temperature is above the maximum limit. No diff's are expected in example files.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??

### DIFF
--- a/src/EnergyPlus/HVACMultiSpeedHeatPump.cc
+++ b/src/EnergyPlus/HVACMultiSpeedHeatPump.cc
@@ -3263,7 +3263,6 @@ namespace HVACMultiSpeedHeatPump {
 		Real64 HeatRecOutletTemp; // Heat reclaim outlet temp [C]
 		Real64 HeatRecMassFlowRate; // Heat reclaim mass flow rate [m3/s]
 		Real64 CpHeatRec; // Heat reclaim water inlet specific heat [J/kg-K]
-		Real64 HeatRecInletEnth; // Heat reclaim water inlet enthalpy [J/kg]
 
 		// Begin routine
 		HeatRecInNode = MSHeatPump( MSHeatPumpNum ).HeatRecInletNodeNum;
@@ -3271,7 +3270,6 @@ namespace HVACMultiSpeedHeatPump {
 
 		// Inlet node to the heat recovery heat exchanger
 		HeatRecInletTemp = Node( HeatRecInNode ).Temp;
-		HeatRecInletEnth = Node( HeatRecInNode ).Enthalpy;
 
 		// Set heat recovery mass flow rates
 		HeatRecMassFlowRate = Node( HeatRecInNode ).MassFlowRate;
@@ -3283,9 +3281,13 @@ namespace HVACMultiSpeedHeatPump {
 			CpHeatRec = GetSpecificHeatGlycol( PlantLoop( MSHeatPump( MSHeatPumpNum ).HRLoopNum ).FluidName, HeatRecInletTemp, PlantLoop( MSHeatPump( MSHeatPumpNum ).HRLoopNum ).FluidIndex, RoutineName );
 
 			HeatRecOutletTemp = QHeatRec / ( HeatRecMassFlowRate * CpHeatRec ) + HeatRecInletTemp;
-			if ( HeatRecOutletTemp > MSHeatPump( MSHeatPumpNum ).MaxHeatRecOutletTemp ) HeatRecOutletTemp = MSHeatPump( MSHeatPumpNum ).MaxHeatRecOutletTemp;
+			if ( HeatRecOutletTemp > MSHeatPump( MSHeatPumpNum ).MaxHeatRecOutletTemp ) {
+				HeatRecOutletTemp = max( HeatRecInletTemp, MSHeatPump( MSHeatPumpNum ).MaxHeatRecOutletTemp );
+				QHeatRec = HeatRecMassFlowRate * CpHeatRec * ( HeatRecOutletTemp - HeatRecInletTemp );
+			}
 		} else {
 			HeatRecOutletTemp = HeatRecInletTemp;
+			QHeatRec = 0.0;
 		}
 
 		SafeCopyPlantNode( HeatRecInNode, HeatRecOutNode );
--- a/src/EnergyPlus/HVACUnitarySystem.cc
+++ b/src/EnergyPlus/HVACUnitarySystem.cc
@@ -11155,7 +11155,6 @@ namespace HVACUnitarySystem {
 		Real64 HeatRecOutletTemp; // Heat reclaim outlet temp [C]
 		Real64 HeatRecMassFlowRate; // Heat reclaim mass flow rate [m3/s]
 		Real64 CpHeatRec; // Heat reclaim water inlet specIFic heat [J/kg-K]
-		Real64 HeatRecInletEnth; // Heat reclaim water inlet enthalpy [J/kg]
 		Real64 ReportingConstant;
 
 		ReportingConstant = TimeStepSys * SecInHour;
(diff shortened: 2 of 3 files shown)


rraustad:
This is somewhat confusing. Heat balance at each component looks correct. Suspect plant loop is not updating node conditions sufficiently (i.e., needs more iteration). Also, temperature of tank when plant flow = 0 is 25C, yet MSHP heat recovery inlet sees 80C.

@EnergyArchmage, where could I force another plant iteration to see if that helps line up heat transfer rate mismatch?

Where is water node enthalpy calculated?

What happens in the plant loop when mass flow rate = 0? do node conditions get passed around?

Heat transfer rates agree with component node mass flow \* delta H

5516-heattransferrates

Adding pump heat to MSHP heat transfer shows what heat transfer should occur at tank.

5516-massflow_heattransferrates

Delta H at each component shows why heat transfer rates differ.

5516-deltah

Temperature at outlet of water tank shows tank temperature, yet MSHP sees 80C during no flow condition.

5516-tempvsflow

rraustad:
This is interesting. The tank heat transfer does not match pump power when MSHP heat transfer = 0 and tank temp is rising. It does match when tank reaches max limit and is constant. This might have something to do with the 1-sec tank model integrating over a time step different than the plant?

5516-tanknotmatchingpumppower

rraustad:
Tried to justify difference by accounting for storage component where tank water temp changes. Unfortunately, it makes sense that change in water temp is included in heat transfer rate. Plot shows calculation of water mass \* DeltaH / time - storage losses (sign issue?, a negative number means a loss) versus tank source side heat transfer rate. Tank outlet temp and water flow are included for comparison.

Equation used: Storage - Losses = =+(((0.2 x 971)*(DeltaH))/600) - Tank Losses
 0.2 tank volume, 971 kg/m3 at 80C, DeltaH of tank this time step minus previous time step, 10 min

The "energy not conserved" comment means I couldn't find a report variable to account for loss.

5516-energybalance

Update: "Not sure what this is" = standby losses. At around x-axis 485-515 tank temp is around 25C.
Question: why is standby losses non-zero near end of plot where water temp decreases gradually and zero at end of plot where tank losses offset change in water temp?

Update: Standby losses are about 127 W from x-axis 590 to end. For some reason, water temp doesn't change (source inlet/outlet node). There's an issue with node enthalpy when tank flow is 0. Tank outlet node enthalpy is constant, tank inlet node enthalpy is decreasing. Inlet/outlet tank water temps are constant and identical. Oh, and I used tank outlet node temp as surrogate for tank temp. Might be cause of some confusion here.

Found the vent report variable: Water Heater Venting Heat Transfer Rate [W]

lgu1234:
@rraustad You caught the storage term. I think energy is balanced during a period of heat recovery. Why is not zero for the blue line between 433 and 450 (x-axis). The previous figure shows zero values in the same period.

rraustad:
I think I've resolved the tank heat balance but it brings up another question. If I look at a time when only standby losses exist when pump is running at 261 W, I calculate 40 W net for the tank water temperature increase while I see 135 W entering the tank. What am I doing wrong here?
Tank = 0.2 m3, water density = 971 kg/m3 at 80C,  S = 4180 J/kg, dT = 0.173, dt = 3600 sec or 40 W.
Pump heat = 261 W, tank losses = 126 W, tank heat transfer = 135 W which yields 0.173 C/hr ??

5516-heatbalancelooksgood

EnergyArchmage:
The ConvervenceLimits object has a field called Minimum Plant Iterations, you can use that to force more plant iteration.

rraustad:
Hmmm, all the waste heat as a function of temperature curves are unity. Not sure how this could converge. First iteration the heat recovery calculates an outlet temp (higher than the tank temp since the coil HR inlet is the same as or higher [pump heat] than the tank temp), the water heater sees this temp and calculates a higher tank temp, then back to the coil where heat recovery outlet temp is a little higher, and so on.

Still not sure why Qcoil, heat recovery is twice as high as Q tank, source side. Especially since the Qcoil, HR is constant, plus pump heat, should make Qtank, source side always higher than Qcoil, HR. RE: the Water Heater Heat Balance plot above, the 40 W does work at the tank. 170 W Source side heat transfer - ~130 W tank losses does equal the 40 W needed for 0.173C/hr slope. So everything I've seen says the models in themselves are working as expected. It's the connection between them that I believe is causing this diff.

```
Curve:Biquadratic,
  HAPCCoolWHFT Speed 1,    !- Name
  1.0,                     !- Coefficient1 Constant
  0.0,                     !- Coefficient2 x
  0.0,                     !- Coefficient3 x**2
  0.0,                     !- Coefficient4 y
  0.0,                     !- Coefficient5 y**2
  0.0,                     !- Coefficient6 x*y
  0,                       !- Minimum Value of x
  50,                      !- Maximum Value of x
  0,                       !- Minimum Value of y
  50,                      !- Maximum Value of y

A18, \field Speed 1 Waste Heat Function of Temperature Curve Name
   \type object-list
   \object-list BiquadraticCurves
   \object-list BiVariateTables
   \note curve = a + b*odb + c*odb**2 + d*db + e*db**2 + f*odb*db
   \note odb = Outdoor air dry-bulb temperature (C)
   \note db = entering coil dry-bulb temperature (C)
```

rraustad:
I guess the unity curve has nothing to do with this since for any given iteration, the waste heat capacity is a constant (i.e., db and odb are constant for a given iteration). Seems like this curve should be a function of water temp.

Also, I did try the ConvergenceLimits today with minimum plant iterations at 6, this didn't help.

rraustad:
Trying to resolve energy around the plant loop. Came across the fact that the pump outlet node conditions are the same as the inlet ?????

5516-pumpconditions

Myoldmopar:
Yes, the pump doesn't attempt to immediately change the temperature of the fluid across its inlet and outlet nodes.  Instead the pump heat is added to the loop mass at the inlet of each loop side.  So for a supply side pump, the pump heat is stored and added to the loop mass at the supply inlet node on the next timestep.  This approximates the diffusion of pump energy from kinetic energy into thermal energy over a short, but not instantaneous, time.

EnergyArchmage:
plant loop pump heat would show up in the change in temperature over time for these Plant Demand Side Lumped Capacitance Temperature and Plant Supply Side Lumped Capacitance Temperature.

rraustad:
OK, so things do resolve around the loop. Not the way I would expect, but the energy balance is near 0 (< 100 W). If I sum the Q's around the loop (actually sum Q's - Tank Heat Transfer) I should get 0. The schematic shows the equation. The first plot shows the plant energy balance (in Watts), the second plot shows the Q at each plant interface node pair. And finally, if I compare the tank heat transfer plus the plant interface node energy content to the MSHP heat recovery, I get a match. I'm calling this issue complete.

5516-schematic

5516-plantenergybalance

rraustad:
@mjwitte @EnergyArchmage the difference between the MSHP heat recovery rate and the Big Water Heater source side heat transfer is in the plant interface node energy storage. I think that since this loop far exceeds the temperature set point (25C) that this stored energy is not getting removed by the plant equipment. This seems like an issue to me, just not related to the MSHP heat recovery. I fixed several calculations in the HR routine. This should probably be closed and a more broad issue opened if this is something that needs to be addressed.

EnergyArchmage:
I am not sure what to say, it is all too easy to have a heat recovery loop that isn't "sized" correctly and runs differently than expected.  I haven't looked closely at this file's hot water use to judge if there is a real issue (beyond a input file that lacks HW demand during cooling), but if the temperatures are not running away then I sort of doubt it.

rraustad:
Plant flow rate is 0.00055 m3/s, Plant volume is 0.41 m3. 12.5 minutes to circulate the plant volume. I increased the flow rate x10 and things got worse. I decreased the plant flow rate x10 and things line up now. Would be nice to figure out a way to warn the user.

5516-decreasedplantmdot

rraustad:
Here's the result when increasing the plant flow rate as a reference. 

5516-increasedplantmdot

EnergyArchmage:
Could maybe do the  Mcp Dt/dt calculation on the capacitance tanks and compare the magnitude to the loads on the loop.  If  capacitance change is always greater than the demand side's loads, then throw a warning.

nrel-bot-3:

nrel-bot-3:

Issue #4944, opened 2015-05-29
Diffuse solar shading changes if the transmittance schedule changes during the year vs constant all year


mjwitte:
Helpdesk ticket 10499
User was comparing diffuse solar on a shaded window with a fixed shade (no transmittance schedule), no shading surface at all, and a seasonal shade (transmittance schedule = 1 in winter, = 0 in summer).

For the seasonal shade, the seasons with transmittance = 1 match the no shade case exactly.
But the summer with transmittance = 0 compared to the fixed shade case shows the beam matches exactly but the diffuse is significantly different.  The diffuse is close to the fixed shade case, but not exactly.

Added another case with the transmittance schedule = 0 year-round and now the summer results match the fixed shade case.

And another case with transmittance schedule = 0 thru Sep 30, then = 1 for the rest of the year yields a different result for June.

Defect files have Timestep shading calcs and DetailedSkyDiffuse options set, but other options show similar results.

See output for ZN001:WALL-SOUTH:WIN001:Surface Outside Face Incident Sky Diffuse Solar Radiation Rate per Area [W/m2](Hourly)
Happened to choose June 16 to look at but other summer dates show similar results.

JasonGlazer:
@mjwitte @EnergyArchmage @Myoldmopar The Surface Outside Face Incident Sky Diffuse Solar Radiation output variable corresponds with QRadSWOutIncidentSkyDiffuse(). The differences seem to trace to a line that computes this based on AnisoSkyMult() which is calculated in two different routines if shading transmittance varies or not. The equation is identical and the values are different at a specific timestep only for DifShdgRatioIsoSky() and DifShdgRatioHoriz() variables. The larger difference is with the DifShdgRatioIsoSky() array which is also calculated almost identically in two places depending on if it is done once at the beginning or if it done throughout the simulation. The calculation is the ratio of WithShdgIsoSky() and WoShdgIsoSky(). The values for these are very different (by several orders of magnitude) for the two test files that should not be showing any differences. These are calculated in for the non-varying case at:

https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SolarShading.cc#L8340-L8386

and for the varying case at:

https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SolarShading.cc#L3608-L3624

[...]

PR #5741, opened 2016-06-21
Fix diffuse Solar when using shading with transmittance schedule

JasonGlazer:
This fix addresses #4944 when the diffuse solar in cases that shading uses a transmittance schedule is inconsistent with shading without a transmittance schedule.
- [x] Unit Test(s)


src/EnergyPlus/SolarShading.cc
```diff
@@ -1334,6 +1334,12 @@ namespace SolarShading {
 		++NumAnisoSky_Calls;
 #endif
 
+		Real64 curSunlitFrac = 0.;
+		Real64 CurDifShdRadioHoriz = 0.;
+		if ( TimeStep == 4 && HourOfDay == 9 && DayOfSim == 167 ){
+			int x = 1;
+		}
```
EnergyArchmage: is this just for testing?

JasonGlazer: Just a debugging aid. 

I put a breakpoint on the "int x=1;" line and it is an easy way to get the simulation to stop at a certain time. I have tried using the hit count breakpoints but because how EnergyPlus iterates, it doesn't always come out to be the same time for different (but similar) test files.

src/EnergyPlus/SolarShading.cc
```diff
@@ -3639,17 +3636,26 @@ namespace SolarShading {
...
 		Real64 Fac1WithShdg; // Intermediate calculation factor, with shading
+		int IPhi; // Altitude step counter
+		int ITheta; // Azimuth step counter
+		Real64 DPhi; // Altitude step size
+		Real64 DTheta; // Azimuth step size
+		Real64 DThetaDPhi; // Product of DTheta and DPhi
+		Real64 PhiMin; // Minimum altitude
+		Real64 Phi; // Altitude angle
+		Real64 Theta; // Azimuth angle
+		Real64 FracIlluminated; // Fraction of surface area illuminated by a sky patch
```
Myoldmopar, line 3647: Can these variables be defined in the little block below when they are used?  I'm commenting as I read, so I may answer my own question shortly.

src/EnergyPlus/SolarShading.cc
```diff
@@ -3679,25 +3685,58 @@ namespace SolarShading {
 
 		//   Note -- if not the below, values are set in SkyDifSolarShading routine (constant for simulation)
 		if ( DetailedSkyDiffuseAlgorithm && ShadingTransmittanceVaries && SolarDistribution != MinimalShadowing ) {
-			CosPhi = 1.0 - SUNCOS( 3 );
+			WithShdgIsoSky = 0.;
```
Myoldmopar, line 3688: The decimal point after the zero here is completely fine, but also unnecessary.  The compiler knows it is a floating point, so even just `0` is fine.

PR #5746, opened 2016-06-23
Add an optional fuel type to OtherEquipment

### LINKED ISSUE #5656: FuelEquipment
## Issue overview

EnergyPlus currently has `ElectricEquipment` and `GasEquipment` objects that use electricity or natural gas and contribute to internal gains. There is also a `OtherEquipment` which uses no fuel and contributes to internal gains. Additionally, there is a `Exterior:FuelEquipment` that uses a fuel specified in a field (Electricity, NaturalGas, PropaneGas, FuelOil, etc) but does not contribute to internal gains. 

For residential use there exists equipment that uses fuels other than electricity or natural gas and contributes to internal gains (i.e. a propane oven/range or a propane clothes dryer). Therefore there is a need for a generic `FuelEquipment` object that allows the user to specify a fuel and contributes to internal gains.

This could be implemented in a couple of ways:
1. Create a `FuelEquipment` to replace (or deprecate) `GasEquipment` and `ElectricEquipment`.
2. Create a `FuelEquipment` that has all of the fuel types except electricity or natural gas.
3. Create separate `PropaneEquipment`, `FuelOilEquipment`, etc objects.
4. Equipment that replaces all interior/exterior, gas/electric/other. From [below](https://github.com/NREL/EnergyPlus/issues/5656#issuecomment-220482638)
5. Create a `FuelEquipment` to replace `GasEquipment`. From [below](https://github.com/NREL/EnergyPlus/issues/5656#issuecomment-219066932)
6. Add "Fuel Type" field to `OtherEquipment`. From [below](https://github.com/NREL/EnergyPlus/issues/5656#issuecomment-221892458)

My preference is 1. Before I go make an NFP, are there any thoughts on how this should be done?

@Myoldmopar @shorowit

Table previously in @nealkruis [post](https://github.com/NREL/EnergyPlus/issues/5656#issuecomment-221677604):

| Object | Generates Heat in Zone | Exterior (no zone heat gain) | Generates CO2 | Consumes Electricity | Consumes Natural Gas | Uses District Heat | Consumes any Fuel Type | Consumes no Energy Resource | Net Number of Objects |
| --- | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: | :-: |
| ElectricEquipment | x |  |  | x |  |  |  |  | NA |
| GasEquipment | x |  | x |  | x |  |  |  | NA |
| HotWaterEquipment | x |  |  |  |  | x |  |  | NA |
| SteamEquipment | x |  |  |  |  | x |  |  | NA |
| OtherEquipment | x |  |  |  |  |  |  | x | NA |
| Exterior:FuelEquipment |  | x |  | x | x | x | x |  | NA |
| Option 1 | x |  | x | x | x | x | x |  | -1 |
| Option 2 | x |  | x |  |  | x | x |  | +1 |
| Option 3 | x |  | x |  |  |  |  |  | +11 |
| Option 4 | x | x | x | x | x | x | x | x | -5 |
| Option 5 | x |  | x |  | x | x | x |  | 0 |
| Option 6 | x |  |  | x | x | x | x | x | 0 |

nmerket:
## Pull request overview

Adding an optional fuel type to `OtherEquipment`.  Fixes #5656.
### Work Checklist
- [x] Code and test feature
- [x] Documentation
- [x] Unit test
- [x] Transition
  - [x] Document in spreadsheet
  - [x] Write conversion program
  - [x] Transition existing test files
- [x] Add new test file
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


tst/EnergyPlus/unit/InternalHeatGains.unit.cc
```diff
@@ -0,0 +1,136 @@
...
+#include <DataGlobals.hh>
+#include <DataHeatBalance.hh>
+#include <ExteriorEnergyUse.hh>
+
+#include "Fixtures/EnergyPlusFixture.hh"
+
+using namespace EnergyPlus;
+using namespace ObjexxFCL;
+
+TEST_F( EnergyPlusFixture, InternalHeatGains_OtherEquipment )
```
nmerket: @Myoldmopar I would like to do more here, especially test failure conditions on the input processing. I'm having a hard time figuring that out. Google test has the [exception assertions](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#exception-assertions), but we tend to error out a little differently. Would it be possible to check the error messages we submit to `ShowSevereError` like the python [assertRaisesRegExp](https://docs.python.org/2.7/library/unittest.html#unittest.TestCase.assertRaisesRegexp)?

mjwitte: @nmerket [EnergyPlusFixture](https://github.com/NREL/EnergyPlus/blob/develop/tst/EnergyPlus/unit/Fixtures/EnergyPlusFixture.cc#L432-L478) has several functions that can be used to check the contents of various output streams.  [MixedAir.unit.cc](https://github.com/NREL/EnergyPlus/blob/ae55b4e023f6ba80f4fb7f70d6b72a02932c77da/tst/EnergyPlus/unit/MixedAir.unit.cc#L1223-L1235) has an example of checking an error message.

mjwitte: And one trick to get the exact error string with all the right spacing and asterisks is to use a string that will fail the match, run the test and then use the output from the failed test to copy/paste the error string.

nmerket: Thanks, @mjwitte, that's exactly what I was looking for. Thanks for the pointer on how to get the error message too!

nmerket:
@Myoldmopar, I think this is ready to go except for the transition program, which I leave to you.

nrel-bot-3:
@Myoldmopar @nmerket @lgentile it has been 15 days since this pull request was last updated.

nmerket:
@Myoldmopar, @mbadams5, Is there any reason `ErrorsFound` needs to be static as shown [here](https://github.com/NREL/EnergyPlus/blob/f05c758b24dfdc528e5f25a354a1e5f3c590f4a8/src/EnergyPlus/InternalHeatGains.cc#L286)? It's messing up my unit tests.

As [this guy](http://www.cprogramming.com/tutorial/statickeyword.html) describes, the use of the `static` keyword works like:

> The use of static inside a function is the simplest. It simply means that once the variable has been initialized, it remains in memory until the end of the program. You can think of it as saying that the variable sticks around, maintaining its value, until the program completely ends.

(I'm sure you all know that, but it was news to me, so I wanted to note it so I wouldn't forget it.)

In my unit tests I have places where I intentionally have erroneous inputs and make sure it errors out. Later, another test needs to call `GetInternalHeatGainsInput`, but `ErrorsFound` is already true because it was true from my previous test, so that test errors out even though there are no errors in the input. This only happens when the tests are run together.

The easy fix I see is to make `ErrorsFound` just a normal local variable that doesn't persist between calls. I don't see how having it stay set between function calls is useful in any way, but I wanted to check with you guys first in case that was done that way for a good reason that I'm not seeing.

Myoldmopar:
Uh, it shouldn't anymore.  It should be a namespace level variable so it is easily reset by the clear_state functions.

nmerket:
Alright, @Myoldmopar, I feel pretty comfortable with this pull request at this point. It does need a transition program, though, which I leave to you. Beyond that, I think it's ready for review and merging.

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -2146,66 +2148,102 @@ namespace InternalHeatGains {
 						if ( errFlag ) ErrorsFound = true;
 					}
 
-					ZoneOtherEq( Loop ).SchedPtr = GetScheduleIndex( AlphaName( 3 ) );
+					std::string FuelTypeString( "" );
+					if ( AlphaName( 2 ) == "NONE" ) {
+						ZoneOtherEq( Loop ).OtherEquipFuelType = 0;
```
Myoldmopar: Raw Integer?  How about a (class) enum?  This zero isn't meaningful to anyone else looking at this code.

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -2237,6 +2280,12 @@ namespace InternalHeatGains {
 					// Zone total report variables
 					if ( RepVarSet( ZoneOtherEq( Loop ).ZonePtr ) ) {
 						RepVarSet( ZoneOtherEq( Loop ).ZonePtr ) = false;
+
+						if ( ZoneOtherEq( Loop ).OtherEquipFuelType > 0 ) {
```
Myoldmopar: If you used an enum above, this would also change, to something like: 

```
if ( ZoneOtherEq( Loop ).OtherEquipFuelType != FuelTypeNone ) {
```

testfiles/5ZoneDetailedIceStorageSimpleCtrl.idf
```diff
@@ -3151,7 +3151,8 @@ Schedule:Compact,
     For: Weekdays SummerDesignDay, !- Field 2
     Until: 24:00,1.0,        !- Field 3
     For: AllOtherDays,       !- Field 4
-    Until: 24:00,0.0;        !- Field 5
+    Until: 24:00,0.0;        !- Field 5
+
```
Myoldmopar, line 3155: Uhh something happened here; line ending?

Myoldmopar, line 3155: And 5ZoneSwimmingPool looks like a line ending problem too maybe...

Myoldmopar:
OK, I'm going to take a look at the line endings issues I noted above.  I might also fix up the enum thing; at a minimum I could just set a parameter `int const noFuelType = 0;` to the code so there isn't a stray zero literal.  I'll be working on the transition changes.  I'll be running tests too.  I'll let you know when I'm done if I need something; or I'll merge.

testfiles/5ZoneDetailedIceStorageSimpleCtrl.idf
Myoldmopar, line 3155: OIC...you fixed the bad line ending that was there in develop.  Yikes, and thanks.

```
$ diff 1eplus/testfiles/5ZoneDetailedIceStorageSimpleCtrl.idf 5eplus/testfiles/5ZoneDetailedIceStorageSimpleCtrl.idf | cat -t
3154c3154,3155
<     Until: 24:00,0.0;        !- Field 5^M^M
---
>     Until: 24:00,0.0;        !- Field 5
```

Myoldmopar, line 3155: And same thing in develop for SwimmingPool.  Thanks @nmerket.

Myoldmopar:
OK, I still have some testing to do, but now transition is done, and I addressed all my own review comments so far 👍 .

Myoldmopar:
@nmerket Did you end up adding a test file to actually demonstrate this new capability?  If not, could you go ahead and do that?  You don't necessarily need to add a new test file, just incorporate it into an existing file.

Myoldmopar:
@nmerket Did you end up adding a test file to actually demonstrate this new capability?  If not, could you go ahead and do that?  You don't necessarily need to add a new test file, just incorporate it into an existing file.

Myoldmopar:
And this merge from develop will clean up the stray unit test failure.

nmerket:
@Myoldmopar added a test file and it should be ready to merge. Thanks!

Myoldmopar:
I'll let CI have its way, but yeah, I don't have anything further, so if it is happy then we'll be good to merge.

Myoldmopar:
Although CI will now see the changes from the other PR I just merged.  The changes are orthogonal enough though, I'm just going to merge this after all.

Issue #5711, opened 2016-06-03
HVAC Sizing Simulation crash with bad subscript


mjwitte:
## Issue overview

User file crashes [here](https://github.com/NREL/EnergyPlus/blob/d5db2e7f81c9d462bc85583ed347623e3384bcdf/src/EnergyPlus/SizingAnalysisObjects.cc#L254) in SizingAnalysisObjects.cc SizingLog::FillSysStep with a bad subscript.  `tmpSysStepStamp.stStepsIntoZoneStep = -1`.

Workaround is to turn off Do HVAC Sizing Simulation.
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (current develop 3b72813e45)
- helpdesk ticket number 11403
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (DefectFiles\5711)
- [x] Ticket added to Pivotal for defect (development team task)
- [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

EnergyArchmage:
Okay, I see it now, the defect file has also turned on report during warmup and the value for DataGlobals::DayOfSim is not as expected during warmup days.

mjwitte:
Closed via #5751

PR #5751, opened 2016-06-30
Corrections to sizing simulation logging to fix crashes, subtimesteps, and report during warmup

EnergyArchmage:
## Pull request overview

This PR is for issue #5711.  Multiple issues were found and corrected in the logging for HVAC Sizing Simulations. 
1. Changed how zone timestamp constructor works, now it uses DataGlobals::TimeStep instead of OutputProcessor:TimeValue.  Problem always was that TimeValue is out of sync during system timesteps and caused a lot of problems.  
2. DayOfSim increments up during successive warmup days and if ReportDuringWarmup is on the logging system had problems because there were too many days. Added trap for when logging is called during warmup.
3. Add traps for bad indices to avoid hard crashes.

Most of the code line diffs are formatting and whitespace clean up to match current practice. I will make line comments to point out where the real changes are occuring. 

CI shows errors/changes on 8 integration test files that all use plant coincident sizing.  These are expected, the logging has changed and there are small differences in the coincident size results triggering numerical diffs because the equipment size outcomes changed.  
### Review Checklist

- [x] Unit Test(s)
    expanded unit test to add one that tests sub-system timesteps.  Also had to alter existing tests for change in constructor. 
- C++ checks:
  - [x] Argument types
- IDD changes n/a


nrel-bot-3:

mjwitte:
@EnergyArchmage @Myoldmopar  Ran local annual simulations of the 8 test files that show diffs using the python test suite scripts just to see the annual diffs, but tablediff had errors like this:

image

Not sure why the CI tablediff ran successfully and the local one didn't, but this led to noticing that the last three columns in the Plant Loop Coincident Design Fluid Flow Rate Adjustments are mislabeled as {TIMESTAMP}
image

@EnergyArchmage Please fix that here in this PR, either remove {TIMESTAMP} from those column headers, or combine these into a single column that is a real timestamp (if that makes any sense here).  And add an entry to [OutputChanges8-5-0-to-8-6-0.md](https://github.com/NREL/EnergyPlus/blob/develop/src/Transition/OutputRulesFiles/OutputChanges8-5-0-to-8-6-0.md).

mjwitte:
Looked at annual diffs locally.  Some of the test files show tiny changes in annual energy use, but as exepcted, the big diffs are only in the reported peak times in the coincident sizing report.  Merging.

Issue #5757, opened 2016-07-05
Outside Movable Insulation Does Not Differentiate Between Solar and Thermal Absorptivity


RKStrand:
## Issue overview

While working on a bug in the inside movable insulation, it was noticed that the inside movable insulation always grabs the solar absorptivity internally in the subroutine that evaluates inside movable insulation.  That actually is ok because code in the one place where thermal absorptivity is needed does not use the solar absorptivity returned by the subroutine call but actually grabs the correct data.  Unfortunately, in the calls for the OUTSIDE movable insulation, the call that needs the thermal absorptivity does not make this correction in the code surrounding the subroutine call.  This needs to be fixed.
### Details

Some additional details for this issue (if relevant):
- Platform: all platforms
- Version of EnergyPlus: problem was noted in Version 8.5 and has probably been there a while.
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ x ] No defect file but a new test suite file will likely be added as part of this modification
- [ x ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #5762

PR #5762, opened 2016-07-06
124975981 Correction of Absorptivity Assignment in Movable Insulation

RKStrand:
## Pull request overview

This defect came about be a discovery that the movable insulation code was not making any distinction between the solar and thermal absorptivity for OUTSIDE movable insulation.  The code was basically always using the solar absorptivity when in one situation it should have used the thermal absorptivity.  This problem was fixed.  In addition, during the testing of this fix, it was discovered that the material properties it was using for solar absorptivity were only correct if the material layer was a glazing layer.  For other material layers, the calculation of solar absorptivity was incorrect.  This was fixed for both inside and outside movable insulation.  It should be noted that even though this was established for inside movable insulation that this still does not allow the proper modeling of transparent movable insulation on the INSIDE.  This would be a future enhancement idea though not very high priority as there is not much interest in movable insulation and generally transparent insulation only makes sense on the exterior of the building.  This addresses GitHub Issue #5757 as originally stated and the additional issues as well.
### Work Checklist

  - Defect: This work started with the correction of GitHub Issue #5757 and then corrected additional issues that were discovered along the way.
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


RKStrand:
@mjwitte @Myoldmopar Could one of you also verify this fix?  I made the mistake of changing the title.  So, this was originally GitHub Issue #5757 and I should have left it as that.  When I made the pull request, I made the mistake of changing the title and this now is being called #5762.  My apologies for that confusion.  Other than that, this will hopefully be another relatively straightforward fix.  It includes a couple of interrelated issues, two unit tests, and a documentation clarification.

doc/input-output-reference/src/overview/group-advanced-surface-concepts.tex
```diff
@@ -1,4 +1,10 @@
-\section{Group -- Advanced Surface Concepts}\label{group-advanced-surface-concepts}
+\documentclass[10pt]{article}
```
Myoldmopar: @RKStrand This standalone LaTeX preamble will have to be removed back out so that the documentation can be built as part of the full pdf.

doc/input-output-reference/src/overview/group-advanced-surface-concepts.tex
```diff
@@ -1877,3 +1883,5 @@ \subsubsection{Inputs}\label{inputs-17}
     Lshaped Zone:South Wall,Lshaped Zone:East Wall,0.101310,
   <snip>
 \end{lstlisting}
+\]
```
Myoldmopar: Similar to [above](https://github.com/NREL/EnergyPlus/pull/5762/files#r71266857)...

src/EnergyPlus/HeatBalanceMovableInsulation.cc
```diff
@@ -177,8 +179,12 @@ namespace HeatBalanceMovableInsulation {
 
 			HMovInsul = 1.0 / ( MovInsulSchedVal * Material( Surface( SurfNum ).MaterialMovInsulExt ).Resistance );
 			RoughIndexMovInsul = Material( Surface( SurfNum ).MaterialMovInsulExt ).Roughness;
-			AbsExt = max( 0.0, 1.0 - Material( Surface( SurfNum ).MaterialMovInsulExt ).Trans - Material( Surface( SurfNum ).MaterialMovInsulExt ).ReflectSolBeamFront );
-
+			MaterialGroupNum = Material( Surface( SurfNum ).MaterialMovInsulExt ).Group;
```
Myoldmopar: Minor comments:
- Consider defining `MaterialGroupNum` directly where it is used, no need to define variables at the top of each function like we had to do with FORTRAN.  Making that a habit can ensure the variable is scoped properly, which can aid in the compiler auto-optimizing the code.
- Consider cleaning up empty and unnecessary comment block sections from the old FORTRAN template (`DERIVED TYPE DEFINITIONS`, etc.) as you go, to slowly clean the code up.
- There are a number of lookups for the `Surface( SurfNum ).MaterialMovInsulExt`.  The compiler _may_ be optimizing these away, but it may be better to reduce these calls down by using a single reference.  Add in `auto & thisSurface( Surface( SurfNum ) )`, and then change each of those `Surface( SurfNum )` calls to just be `thisSurface`. Like I said, it may not affect the compiler, but it will make a lighter looking block of code.

src/EnergyPlus/HeatBalanceMovableInsulation.cc
```diff
@@ -140,7 +140,8 @@ namespace HeatBalanceMovableInsulation {
 		// (I)BLAST legacy routine OMVINS
 
 		// USE STATEMENTS:
-		// na
+		using DataHeatBalance::WindowGlass;
```
Myoldmopar: In general, I'd say just go ahead and write `DataHeatBalance::WindowGlass` in the below code instead of a using statement.  For this small of a function, it's not a big deal either way, but with larger functions, if there are many using statements, it gets hard to understand where each variable is coming from.

src/EnergyPlus/HeatBalanceMovableInsulation.cc
```diff
@@ -245,8 +253,15 @@ namespace HeatBalanceMovableInsulation {
 			}
 
 			HMovInsul = 1.0 / ( MovInsulSchedVal * Material( Surface( SurfNum ).MaterialMovInsulInt ).Resistance );
-			AbsInt = Material( Surface( SurfNum ).MaterialMovInsulInt ).AbsorpSolar;
 
+			MaterialGroupNum = Material( Surface( SurfNum ).MaterialMovInsulInt ).Group;
```
Myoldmopar: Similar comments to above.

Myoldmopar:
This looks good.  The unit tests are nice and to the point.  I made a few pretty minor comments.  If you could make those changes, or at least respond to my thoughts, and then pull in develop (resolving the merge conflict along the way), this should be able to go in.

doc/input-output-reference/src/overview/group-advanced-surface-concepts.tex
RKStrand: @Myoldmopar No idea how this got in there.  I simply edited the text on line 232/238 and nothing else.  I will eliminate these lines and the one at the end and hope they don't magically pop up again.

src/EnergyPlus/HeatBalanceMovableInsulation.cc
RKStrand: @Myoldmopar I'll take care of these other comments as soon as I can.  Not sure how a merge conflict got introduced, but glad you know what it is and how to resolve it.

RKStrand:
@Myoldmopar Ok, I think I have addressed all of the issues you brought up in the latest commit that I made.  And I made another commit which I think addresses the merge conflict?  I'm not sure why it is considered a conflict unless I missed something along the way.  Let me know if I missed anything or if I messed anything up.  Oh, and by the way, I just happened to fix another bug along the way that I missed earlier.

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -51,7 +51,11 @@ set( test_src
   General.unit.cc
   GroundHeatExchangers.unit.cc
   HeatBalanceManager.unit.cc
+<<<<<<< HEAD
```
Myoldmopar: This is the problem @RKStrand; some git merge conflict markers were left in the source after the merge.....  Looks like the resolution is to simply **keep** both unit.cc files listed here and delete the markers.  I'll do that now actually...

Myoldmopar:
OK, this should now build fine.  You mention that this addresses multiple bugs.  Are there multiple _Github tickets_ to go along with those bugs, or just the one Github issue?

RKStrand:
@Myoldmopar This was a single GitHub issue.  I happened to see another obvious one line issue and I simply corrected it along with the other fixes for the bug that caused this GitHub issue in the first place rather than creating another GitHub issue.  Hope that was ok.

Issue #5284, opened 2015-10-27
EMS function @RhovFnTdbRhLBnd0C is being parsed as @RhovFnTdbRh


EnergyArchmage:
While working on another issue (to detect unitialized variables), found that RhovFnTdbRhLBnd0C is actually being parsed as RhovFnTdbRh and a variable LBnd0C.   The substrings match and parser gets confused.  

function RhovFnTdbRhLBnd0C will not work in EMS programs.

rraustad:
Just move the latter up above so it's checked first ?

```
} else if ( SameString( String.substr( Pos, 12 ), "@RhovFnTdbRh" ) ) {
    if ( DeveloperFlag ) gio::write( OutputFileDebug, fmtA ) << "FUNCTION \"" + String.substr( Pos, 12 ) + "\"";
    Token( NumTokens ).Operator = FuncRhovFnTdbRh;
    Token( NumTokens ).String = String.substr( Pos, 12 );
    Pos += 11;
} else if ( SameString( String.substr( Pos, 18 ), "@RhovFnTdbRhLBnd0C" ) ) {
    if ( DeveloperFlag ) gio::write( OutputFileDebug, fmtA ) << "FUNCTION \"" + String.substr( Pos, 18 ) + "\"";
    Token( NumTokens ).Operator = FuncRhovFnTdbRhLBnd0C;
    Token( NumTokens ).String = String.substr( Pos, 18 );
    Pos += 17;
```

rraustad:
I assume there is a similar issue here ?

```
} else if ( SameString( String.substr( Pos, 12 ), "@RhFnTdbRhov" ) ) {
    if ( DeveloperFlag ) gio::write( OutputFileDebug, fmtA ) << "FUNCTION \"" + String.substr( Pos, 12 ) + "\"";
    Token( NumTokens ).Operator = FuncRhFnTdbRhov;
    Token( NumTokens ).String = String.substr( Pos, 12 );
    Pos += 11;
} else if ( SameString( String.substr( Pos, 18 ), "@RhFnTdbRhovLBnd0C" ) ) {
    if ( DeveloperFlag ) gio::write( OutputFileDebug, fmtA ) << "FUNCTION \"" + String.substr( Pos, 18 ) + "\"";
    Token( NumTokens ).Operator = FuncRhFnTdbRhovLBnd0C;
    Token( NumTokens ).String = String.substr( Pos, 18 );
    Pos += 17;
```

[...]

PR #5763, opened 2016-07-07
Correct EMS parsing of RhovFnTdbRh and RhovFnTdbRhLBnd0C. Similar issue w…

rraustad:
## Pull request overview

EMS psychrometric functions were not parsed correctly for the *LBnd0C functions which were being parsed as the similarly named function without the LBnd0C at the end of the function name.
This corrects issue #5284. No diff's expected in example files.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,258 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+			"set Var28 = @RhovFnTdbWPb 0.0 1.0 2.0,",
+			"set Var29 = @RhFnTdbRhov 0.0 1.0,",
+			"set Var30 = @RhFnTdbRhovLBnd0C 0.0 1.0,",
+			"set Var31 = @RhFnTdbWPb 0.0 1.0 2.0,",
+			"set Var32 = @TwbFnTdbWPb 0.0 1.0 2.0,",
+			"set Var33 = @VFnTdbWPb 0.0 1.0 2.0,",
+			"set Var34 = @WFnTdpPb 0.0 1.0,",
+			"set Var35 = @WFnTdbH 0.0 1.0,",
+			"set Var36 = @WFnTdbTwbPb 0.0 1.0 2.0,",
+			"set Var37 = @WFnTdbRhPb 0.0 1.0 2.0,", // this looks suspicious
```
rraustad: This EMS function accepts a 4th argument ?

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,258 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+		EXPECT_TRUE( anyRan );
+
+		int index( 0 );
+		int offset( 24 );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).Operator, FuncRound );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).NumOperands, 1 );
+//		EXPECT_EQ( size( DataRuntimeLanguage::ErlExpression( 1 ).Operand ), 1 ); // why doesn't this compile? ... Array1D< ErlValueType > Operand
+		index = 1 + offset;
+		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( index ).Name, "VAR1" );
+//		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( 25 ).Value, 2.0 ); // why doesn't this compile? same issue as above ... ErlValueType Value
```
rraustad: I couldn't fully test out each EMS function since I couldn't figure out correct syntax for these

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,258 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 34 ).NumOperands, 2 );
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 35 ).Operator, FuncWFnTdbH );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 35 ).NumOperands, 2 );
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 36 ).Operator, FuncWFnTdbTwbPb );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 36 ).NumOperands, 3 );
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 37 ).Operator, FuncWFnTdbRhPb );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 37 ).NumOperands, 4 ); // why is this 4?
```
rraustad, line 1275: W Fn Tdb Rh Pb, or w as a function of 3 independent variables. 4th argument is CalledFrom string for warning messages

rraustad:
@Myoldmopar I've never seen this message before, do I need to merge this branch?

rraustad:
@Myoldmopar the message disappeared when I clicked send on the previous message. There was a message with a green button that said "Confirm Merge".

Myoldmopar:
Were you trying to merge it?

tst/EnergyPlus/unit/EMSManager.unit.cc
EnergyArchmage: the .Value is a structure, you probably want to use DataRuntimeLanguage::ErlVariable( 25 ).Value.Number.  

Operand is an Array1D.  maybe size() doesn't work with that, dunno.

rraustad:
@Myoldmopar no not trying to merge. Just created this PR, changed the title and description. Then scrolled down to see if CI had any results and saw this message. Disregard as it must have been a fluke.

tst/EnergyPlus/unit/EMSManager.unit.cc
rraustad: .Value.Number works. Thanks.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,416 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 49 ).NumOperands, 2 );
+		index = 49 + offset;
+		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( index ).Name, "VAR49" );
+		EXPECT_NEAR( DataRuntimeLanguage::ErlVariable( index ).Value.Number, 0.0, 0.00000001 ); // TrendDirection 0.5 1.5 = ???
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 50 ).Operator, FuncTrendSum );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 50 ).NumOperands, 2 );
+		index = 50 + offset;
+		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( index ).Name, "VAR50" );
+		EXPECT_NEAR( DataRuntimeLanguage::ErlVariable( index ).Value.Number, 0.0, 0.00000001 ); // TrendSum 0.5 1.5 = ???
```
rraustad: @EnergyArchmage the previous 6 tests return 0 as a result for @Trend functions. Either I'm doing something wrong or there is an issue here. I assume it's me.

EnergyArchmage: Those trend functions aren't setup correctly in the unit test.  

 They take a special trend variable as argument 1, and some number of timesteps as the argument 2.   

The trend variable would need to be setup and filled with history, maybe by marching through timesteps, but I think that is way out of scope for this issue.  I would just remove those trend functions from the unit test.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,416 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+
+		bool anyRan;
+		EMSManager::ManageEMS( DataGlobals::emsCallFromHVACIterationLoop, anyRan );
+		EXPECT_TRUE( anyRan );
+
+		int index( 0 );
+		int offset( 24 ); // first 24 values in ErlExpression are key words.
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).Operator, FuncRound );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).NumOperands, 1 );
+//		EXPECT_EQ( size( DataRuntimeLanguage::ErlExpression( 1 ).Operand ), 1 ); // why doesn't this compile? ... Array1D< ErlValueType > Operand
```
rraustad: @nealkruis @Myoldmopar what do I need to do to get this to compile?

tst/EnergyPlus/unit/EMSManager.unit.cc
rraustad: Got it. I can probably set something up to test the result correctly.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,475 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 15 ).Operator, FuncRandG );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 15 ).NumOperands, 4 );
+		index = 15 + offset;
+		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( index ).Name, "VAR15" );
+
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 16 ).Operator, FuncRandSeed );
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 16 ).NumOperands, 1 );
+		index = 16 + offset;
+		EXPECT_EQ( DataRuntimeLanguage::ErlVariable( index ).Name, "VAR16" );
```
rraustad, line 1110: I thought these 3 random functions (14, 15 and 16) would return different values each time the unit test was executed so I didn't test the result. I see now that the result is always @RandomUniform(0.0, 1.0)=0.60066077, @RandomNormal (0.0, 1.0, 2.0, 3.0)=2.0, RandomSeed(0.0)=0.0. Is this correct?

So this doesn't generate a random number?
    } else if ( SELECT_CASE_var == FuncRandU ) {
        RANDOM_NUMBER( tmpRANDU1 );
        tmpRANDU1 = Operand( 1 ).Number + ( Operand( 2 ).Number - Operand( 1 ).Number ) \* tmpRANDU1;
        ReturnValue = SetErlValueNumber( tmpRANDU1 );

rraustad, line 1110: Fixing the min/max for RandomNormal changes the answer but the result is still static.
 @RandomUniform(0.0, 1.0)=0.7526 (OK, I did get a different result here)
 @RandomNormal (0.0, 1.0, 0.0, 1.0)=1.0
 @RandomSeed(1.0)=1.0

Adjusting min/max again gives a different static result:
 @RandomNormal (0.0, 1.0, -10.0, 10.0)=0.0

EnergyArchmage, line 1110: You have to first control the seed to get random results.   This way it is easier to get repeatable energy simulations.  There are built in variables called ActualDateAndTime and ActualTime which are useful for the seed.

@SeedRandom(ActualDateAndTime) would be called once early to initialize the seed used for subsequent calls to the random functions. 

BTW, watch out with these "@" symbols in github commenting.  I think you may have spammed someone named RandomSeed.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -976,310 +976,418 @@ TEST_F( EnergyPlusFixture, EMSManager_TestFuntionCall ) {
...
-		EMSManager::ManageEMS( DataGlobals::emsCallFromHVACIterationLoop, anyRan );
+		EMSManager::ManageEMS( DataGlobals::emsCallFromHVACIterationLoop, anyRan ); // process trend functions again using above data
 		EXPECT_TRUE( anyRan );
 
 		int index( 0 );
 		int offset( 25 ); // first 24 values in ErlExpression() are key words + 1 EMS global variable
 		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).Operator, FuncRound );
 		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).NumOperands, 1 );
-//		EXPECT_EQ( size( DataRuntimeLanguage::ErlExpression( 1 ).Operand ), 1 ); // why doesn't this compile? ... Array1D< ErlValueType > Operand
+		EXPECT_EQ( DataRuntimeLanguage::ErlExpression( 1 ).Operand.size(), 1 );
```
Myoldmopar: I think these lines with the `size()` check are what throws the build warning.  The `size()` call gives a const unsigned long, and the literal `1` is inferred as a constant integer.  You can see this demonstrated in the Objexx unit test code [here](https://github.com/NREL/EnergyPlus/blob/d5fe22242257821429bbb5123dd69bdc8dce5de1/third_party/ObjexxFCL/tst/unit/Array1.unit.cc#L35).  Just change `1` to `1u` and it will be created as an unsigned and the build warning should go away.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -826,6 +827,583 @@ TEST_F( EnergyPlusFixture, EMSManager_CheckIfAnyEMS_OutEMS ) {
...
+			"set Var6 = @ArcSin 0.70710678090550916,",
+			"set Var7 = @DegToRad 45.0,",
+			"set Var8 = @RadToDeg 0.785398163,",
+			"set Var9 = @Exp 1.0,",
+			"set Var10 = @Ln 2.0,",
+			"set Var11 = @Max 0.5 1.5,",
+			"set Var12 = @Min 0.5 1.5,",
+			"set Var13 = @Abs 1.3,",  // -3.1 does not work, (-3.1) does not work, "set absNum = -3.1" and use @Abs absNum does work if EMS line of code occurs before this line
+			"set Var14 = @SEEDRANDOM 2.65,",
+			"set Var15 = @RANDOMUNIFORM 0.0 1.0,",
```
EnergyArchmage, line 864: I think the behavior of these random functions could vary on different hardware.  I don't really know how the C++ is implementing this but it used to be that the fortran called the processor's random generator and behavior of that would vary with the type of processor hardware.  I do not think these are going to be stable enough for unit testing in general and they cause test failures in CI now.   So I think these three random function tests should be removed.

EnergyArchmage:
all is good here.  Windows 64 CI came back clean before so... merging.

Issue #5442, opened 2016-01-29
Inconsistent reference hot water temperature are used to calculate density in water flow autosize


Nigusse:
While working adding unit test for #4980 , I came across a difference in water reference temperature used to calculated hot water density. In the Unit Heater parent object, the design hot water vol flow rate is calculated at 60.0C, just like the hot water Cp, but ReportSizingOutput uses, 5.05 C as the reference water temperature. And this difference in temperature leads to slight change of water mass flow rate and design coil capacity. This introduced about 1.7% difference between the Design Coil Load and the Design Coil Capacity. Shouldn't we use consistent reference temperature for sizing calculation? I went ahead checked FanCoil and Unit Ventilator objects and found out the same issue.

I am assuming that 5.05C is the reference temperature for water volume flow rate calculation. Shall I change reference water temperature used to calculated water density from 60.0C to 5.05C in the UnitHeater parent object design hot water vol flow rate calculation? @mjwitte @rraustad @Myoldmopar

@rraustad Response:
When the RequestSizing function was created, the component sizing methods were moved to a central function. At that time, the HW coil was sizing using InitConvTemp and I couldn't change it then since it would have caused diff's (I do recall seeing this). Also, we used to call CPCW and CPHW, where Cp was passed back as 4180 regardless of temperature. Now we use GetDensityGlycol (I don't recall when this was changed) and temperature does matter. So yes this should be corrected using a separate issue.

@mjwitte Response:
The change from CPCW/CPHW to GetDensityGlycol should be revisited. If a variable temp is being used, seems it should be tied to the Sizing:Plant design supply temp or the coil object Rated Inlet Water Temperature for coils that have that input, or the average of the design input/output water temps?

Nigusse:
@mjwitte @rraustad @Myoldmopar There are three issues that I see here before I proceed forward:
- Correct the inconsistent use of 60.0C and 5.05C for hot water density calculation in autosizing hot water Volume Flow Rate.  What value should be used? should we take the value from the Sizing:Plant object or Coil:Heating:Water objects? Keep in mind that the values in these two objects can be different, which value should the precedence, the plant or the coil object? For density calculation I suggest to use an inlet condition, not an average.
[...]

PR #5774, opened 2016-07-14
Addresses inconsistent reference hot water temperature are used to calculate density in water flow autosize

Nigusse:
Addresses #5442
## Pull request overview

### Work Checklist

### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


Nigusse:
@mjwitte @rraustad @EnergyArchmage @Myoldmopar I added two global variables (HWCoilInitTemp  and CWCoilInitTemp) that hold Init water temperatures for density and Cp values calculation during sizing (in some cases it is also used during initialization).  In this commit I changed all numeric values of 60.0C used for hot water by HWCoilInitTemp variable for sizing. For chilled water in almost of all cases the variable InitConvTemp (which is initialized to 5.05 C) is used but in a few I cases I also noted that numeric value 5.0 was used. 

In this commit, all chilled water related properties calculations that use InitConvTemp variables were replaced with the variable CWCoilInitTemp. The CWCoilInitTemp variable is initialized to 5.05C.  Diffs were not expected in this commit. This also confirmed by CI machine run above. 

In the next step I will make the following selected changes in **Hot Water Coils** related calculations:
(1) If there is any mismatch between parent and child object use of Water Init Temp for property (density and CP) calculation, then they will be forced to use HWCoilInitTemp, which is 60.0C 

(2) In the few places that I noted 5.0C use for chilled water related sizing calculation will be replaced with CWCoilInitTemp, which is 5.05C.  

Note that there are several places/modules (e.g., cooling towers, Fluid coolers, Evapcoolers, PlantHeatExchangersFluidToFluid, pumps, etc) that cannot be categorized as chilled or hot water. In those modules the code will remain to use InitConvTemp as is. The values of InitConvTemp and CWCoilInitTemp are identical.

These two changes are expected to bring diffs in several example files for sizing and annual run results.

If you have any concerns or suggestion please let me know; otherwise, I will go ahead with the proposed changes.

Nigusse:
Removed unused variables to get rid of the warning message.

Myoldmopar:
I support this change too, but I'd like to see all the `using DataGlobals` changed to simply reference the variable itself in the code.  I'll likely make that change and push it up.  That should also address the [build warnings](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-127d6de36dc57bb529bb0c738459dbe005ef61ff-x86_64-Linux-Ubuntu-14.04-gcc-4.8-IntegrationCoverage-Debug.html).

Nigusse:
@Myoldmopar I see in several places that CWCoilInitTemp variable not referenced. Will you make a second pass?

Myoldmopar:
Well, instead of booting up the Ubuntu machine to find the rest of the error message and correct a single instance or two, I just eliminated all the `using` of InitConvTemp, CWInitConvTemp, and HWInitConvTemp.  This is nicer now anyway.  CI should come back clean and this should be mergeable.

Myoldmopar:
And bonus, net of 500 lines of source removed. (Almost entirely empty comment blocks, but still)

image

Myoldmopar:
Yay, looking good enough to merge before bed.  Thanks @Nigusse.

PR #5776, opened 2016-07-15
Improve ESO file write speed

mbadams5:
## Pull request overview

This improves the speed in which the ESO file (and MTR file) are written.

WriteIntegerData and especially WriteRealData are a large bottleneck when requesting numerous Output:Variables and/or Timestep interval data.

This uses [Milo's dtoa implementation](https://github.com/miloyip/dtoa-benchmark) which in his benchmarks is 9.1x faster than sprintf.

This uses [Milo's itoa implementation](https://github.com/miloyip/itoa-benchmark) which in his benchmarks is 25.5x faster than sprintf. His code is based off [this sse-based code](http://0x80.pl/articles/sse-itoa.html)

In my testing, UpdateDataandReport goes from 26908 ms to 12088 ms, or a 55% reduction for that function when MANY Output:Variables are requested in my test file. (Most of the remaining time is just writing to the stream)

This PR does subtlety change how numbers are output and the precision they are output. I have updated the unit tests which show these changes. I don't feel like this should affect any third party code since they are still fundamentally the same numbers, just sometimes in different notation.
### Work Checklist

### Review Checklist

- [ ] Unit Test(s)


JasonGlazer:
@mbadams5 Great improvement! Have you looked at applying this to the outputs other than ESO/MTR?

mbadams5:
@JasonGlazer It definitely can. I just didn't try very hard to look for other places yet. If you know of specific spots where numerics are being converted to strings to be output then let me know.

JasonGlazer:
Could it be used in RealToStr()? See OutputReportTabular.cc line 13998. That is used all over.

EnergyArchmage:
Also maybe General::TrimSigDigits()  and General::RoundSigDigits()

Myoldmopar:
This is great.  There are build warnings:
image

Milo's code is pretty complex, so I didn't take the time to understand everything it's doing.  Does that library not come with its own unit tests?

mbadams5:
@Myoldmopar I'll look into it.

nrel-bot-3:
@Myoldmopar @mbadams5 @lgentile it has been 14 days since this pull request was last updated.

mbadams5:
For posterity, I did look into `RealToStr()` and it was difficult to just drop in a replacement since that function changes output of the string based on passed in number of digits. There is a solution there, I just didn't have enough time to find the best one.

Issue #4188, opened 2014-04-01
Apparent Timestep-Dependent Plant Volume Autocalculation


Myoldmopar:
An email from @larryscheier described a possible issue with plant volume autocalculation.

I was doing some CPU timings for Trane using the RefBldgHospital example and noticed this: 

| Plant Loop Name | Number of Timesteps: | 1 | 2 | 3 | 4 | 6 | 10 | 15 | 20 | 30 | 60 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| COOLSYS1 | Plant Loop Volume [m3] | 420.7 | 209 | 139.2 | 104.4 | 69.8 | 41.8 | 27.9 | 20.9 | 14 | 7 |
| HEATSYS1 | Plant Loop Volume [m3] | 129.2 | 65.7 | 43.9 | 33 | 21.9 | 13.2 | 8.8 | 6.6 | 4.4 | 2.2 |
| SWHSYS1 | Plant Loop Volume [m3] | 1.299 | 0.649 | 0.433 | 0.325 | 0.216 | 0.13 | 0.087 | 0.065 | 0.043 | 0.022 |

In PlantManager, when Plant Loop Volume is autosized:

```
PlantLoop(LoopNum)%Volume = PlantLoop(LoopNum)%MaxVolFlowRate*TimeStepZone*SecInHour/0.8d0
```

which would be the Plant Loop Volume during a timestep.  Instead, shouldn't this be:

```
PlantLoop(LoopNum)%Volume = PlantLoop(LoopNum)%MaxVolFlowRate*SecInHour/0.8d0
```

As far as I can tell, PlantLoop(LoopNum)%Volume is only used to calculate the PlantLoop(LoopNum)%Mass a few lines down in PlantManger:

```
PlantLoop(LoopNum)%Mass = PlantLoop(LoopNum)%Volume * FluidDensity
```

Then later in HVACInterfaceManager.f90:

   ThisTankMass = FracTotLoopMass \* PlantLoop(LoopNum)%Mass

*MJW 06 Feb 2017 - Added defect files in EnergyPlusDevSupport to test time-step dependency and the new input field for both PlantLoop and CondenserLoop obejcts.*

mjwitte:
@EnergyArchmage The autocalculation of plant loop volume is still timestep dependent as noted in this issue.  Changing this will cause broad diffs, because nearly every file autocalculates the plantloop volume.  Current code is:

[...]

Issue #4126, opened 2014-01-14
Would be nice to diagnose missmatched plant loop volume and operating flow rates. (CR #9355)


axelstudios:
###### Added on 2013-10-24 16:14 by Brent Griffith
--
#### Description
User had a difficult to identify problem where the autocalculated plant loop volume was very large compared to the flow rate.  Supply fluid reset strategies did not work because the loop capacitance was so large that the rest neve makes it to the demand side.  Other users have found that for domestic hot water systems, the autocalculation produces such a large capacitance that the mains water temperature model does not work well enough.  

Suggestion is to find a way to diagnose missmatched sizing of plant loop volume compared to operating flow rates.  Approaches to investigate include:
1. monitor temperature differences between supply outlet and demand inlet and detect if they persist over long periods of time.
2. monitor the maximum operating flow rate and compare that to the magnitude of impact on the capacitance model.  Some criteria to compare Mdot*c*DeltaT flow term to McDT/Dt term.

--
External Ref: Ticket 5556
Last build tested: `13.10.18 V8.1.0.008`

rglazer2:
@mjwitte Any more advice on exactly what needs to be done for this fix, and what would be the best way to accomplish it?

mjwitte:
@EnergyArchmage Any suggestions here?  Or would you rather take this one?

EnergyArchmage:
Hi @rglazer2, here is my $0.02

note that issue #5516 is also related to this and has some discussion
1. change up how the autocalculate sizes plant volume, if they come out smaller this problem gets less. #4188 
2.  add some output variables so time series reporting can more easily diagnose. I suggest adding some calculations and data storage to the routine called UpdateHalfLoopInletTemp. 

I would add two new variables to the plant loop/loop side data structure, along side of LoopSideInlet_TankTemp.  Variables called something like LoopSideInlet_MdotCpDeltaT and LoopSideInlet_McpDTdt, with output names like (for both loop sides)
Plant Demand Side Lumped Capacitance Heat Transport Rate [W]
Plant Demand Side Lumped Capacitance Heat Storage Rate [W]
Plant Supply Side Lumped Capacitance Heat Transport Rate [W]
Plant Supply Side Lumped Capacitance Heat Storage Rate [W]

The calculations should be easy there with the newly calculated TankFinalTemp and all the terms like LastTankOutletTemp available, something like this at the end of the routine. 

```
[...]

PR #5777, opened 2016-07-15
Eliminates timestep dependence in plant volume autocalculation and adds warning when plant loop volume and operating flow rates are mismatched

rglazer2:
## Pull request overview

This pull request addresses defects described by issue #4188 Timestep Dependence in Plant Volume Autocalculation and issue #4126 Mismatched Plant Loop Volume and Operating Flow Rates. 
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


rglazer2:
@mjwitte and @EnergyArchmage I finished this up, let me know if there is anything else I need to add or change.

src/EnergyPlus/PlantManager.cc
```diff
@@ -3282,7 +3288,8 @@ namespace PlantManager {
 		// Small loop mass no longer introduces instability. Checks and warnings removed by SJR 20 July 2007.
 		if ( PlantLoop( LoopNum ).VolumeWasAutoSized ) {
 			// Although there is no longer a stability requirement (mass can be zero), autosizing is formulated the same way.
```
Myoldmopar, line 3290: This comment should be removed or clarified to avoid confusion later.

src/EnergyPlus/PlantManager.cc
```diff
@@ -4531,6 +4541,20 @@ namespace PlantManager {
 		}
 
 	}
+	void
+	CheckOngoingPlantWarnings() 
+	{
+		int LoopNum;
```
Myoldmopar, line 4547: This variable should be defined in the for loop init statement.  Or better yet...since this isn't an order dependent loop, you could do range based:

```
for ( auto & plantLoop : PlantLoop ) {
  if ( plantLoop.LoopSide...blahblahblha...
```

Myoldmopar:
Hmm.  So the autocalculate changes are great.  And the plant tracking stuff looks good, until you notice that even our small performance test suite shows a significant performance hit [(2.5% negative...)](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-fe42ebb58303e396f14008672cf804620ea5bec6-PR5777-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html#performance).  Could you move that section of code that is hit in HVACInterfaceManager out to a higher level in SimHVAC and maybe reduce the number of calls to it?  If not we'll have to get fancier about how we are tracking that.

src/EnergyPlus/PlantManager.cc
```diff
@@ -4531,6 +4541,20 @@ namespace PlantManager {
...
+		for (LoopNum = 1; LoopNum <= TotNumLoops; ++LoopNum) {
+			//Warning if the excess storage time is more than half of the total time
+			if ( PlantLoop( LoopNum ).LoopSide( DemandSide ).LoopSideInlet_CapExcessStorageTime > PlantLoop( LoopNum ).LoopSide( DemandSide ).LoopSideInlet_TotalTime / 2) {
+				ShowWarningError("Plant Loop: " + PlantLoop( LoopNum ).Name + " Demand Side is storing excess heat the majority of the time." );
+			}
+			if ( PlantLoop( LoopNum ).LoopSide( DemandSide ).LoopSideInlet_CapExcessStorageTime > PlantLoop( LoopNum ).LoopSide( DemandSide ).LoopSideInlet_TotalTime / 2) {
```
mbadams5, line 4553: @rglazer2 This seems like a copy and paste error. From the warning message below I think it should be `SupplySide` otherwise this is the exact same if statement as above.

rglazer2:
@Myoldmopar Where do you think it would be best to put that in SimHVAC? It seems like a very complicated function.

EnergyArchmage:
This big performance hit seems difficult to explain to me.  That is not a lot of extra calculations for four filling new output variables.  (I wish you tell which file did what in the performance results.)

We may need to separate out the volume changes from the new output variables.  I would not assume the volume change is innocent on the performance hit.  Iit might be that the lower volume plant loop iterates more in the hospital file, or something like that.

mjwitte:
@EnergyArchmage It would be nice to report the individual file changes directly, but you can get a feel by comparing the file-by-file data from another pull request.  Looking at [this branch](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-fe42ebb58303e396f14008672cf804620ea5bec6-PR5777-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html#performance) vs [5698](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-96c71013166de08d73c1522d2159d6c1c9e1a9df-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html#performance) shows that the big hit comes from PipingSystem_Underground_FHX.  All the others do show increases but this one goes from 676... to 751...

rglazer2:
What should I do then? I can separate out the two changes so we can be sure which is causing the problems, if you think that would be best?

Myoldmopar:
Seems like a good next step.  I'm still not sure why there's such a performance hit.  Go ahead and do that change, and if we still have the performance hit, we'll take a deeper look early next week.

rglazer2:
I just created two new pull requests for issue #4188 and #4126, so I am closing this pull request. Hopefully that will make it easier to determine what is causing such a difference is speed.

Issue #5783, opened 2016-07-20
AvailabilityManager:OptimumStart not working as expected - appears to be crosstalk between systems


mjwitte:
## Issue overview

_From user:_
I'm using AvailabilityManager:OptimumStart to start VAV system 'AHU.SP' up to 1 hour early in the attached IDF. The results show the AHU starting at the earliest possible time (5 AM) on all operating days of the year in 'OptStart.idf', regardless of zone temperature conditions. The normal start time is 6 AM which can be seen working normally in 'NoOptStart.idf'

I tried to limit the applicability by setting the applicability schedule in this object to zero when average temperature of connected zones is between 18 to 24 (OptStart_AppSched_18-24.idf). The applicability schedule seems to have no effect either. In another attempt, I set the applicability schedule to always zero. My expectation was that Optimum Start would not be triggered and the AHU would start as per it's fan schedule (6AM), but it still started at the earliest allowable optimum start time (5AM).

_Reply to user:_
It appears something is broken.  I added some more reporting and then decided just to see what would happen if I changed the Maximum Value for Optimum Start Time from 1 to 5 just for system SP.  This resulted _all_ of the systems starting 5 hrs early for two days and then _none_ of the systems started early after that.
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (v8.5.0)
- helpdesk ticket 11489
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Only works with single air loop, and even then there are deficiencies.

rraustad:
See PR https://github.com/NREL/EnergyPlus/pull/5787

Myoldmopar:
Closed via #5787

PR #5787, opened 2016-07-22
AvailabilityManager:OptimumStart does not work with multiple air loops.

rraustad:
## Pull request overview

Optimum start availability managers are not programmed to retain information about the start time for each HVAC system. This fixes issue #5783. Also, heating coefficients were used to calculate pre-start time for control type = MaximumOfZoneList.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??

… availability manager


rraustad:
I'm a little surprised the example file didn't change. A single air loop using OptStart out of a 4 air loop input. Didn't change much code wise but I did alter how the zone (max out of a zone list) with the peak delta T was calculated so was expecting diff's.
See lines 3128 - 3144.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2301,29 +2290,64 @@ namespace SystemAvailabilityManager {
 		int I;
 		int J;
 		Real64 TempDiff;
-		static Real64 TempDiffHi( 0.0 );
-		static Real64 TempDiffLo( 0.0 );
-		//  LOGICAL, ALLOCATABLE, SAVE, DIMENSION(:) :: ZoneCompOptStartControlType
+		Real64 TempDiffHi;
+		Real64 TempDiffLo;
```
rraustad, line 2301: Must keep track of the result by air loop num since this calc is used the next time through.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2301,29 +2290,64 @@ namespace SystemAvailabilityManager {
...
-		static bool ATGUpdateFlag1( false );
-		static bool ATGUpdateFlag2( false );
+		Real64 AdaTempGradHeat;
+		Real64 AdaTempGradCool;
+		Real64 ATGUpdateTime1( 0.0 );
+		Real64 ATGUpdateTime2( 0.0 );
+		Real64 ATGUpdateTemp1( 0.0 );
+		Real64 ATGUpdateTemp2( 0.0 );
+		bool ATGUpdateFlag1( false );
+		bool ATGUpdateFlag2( false );
```
rraustad, line 2317: Same concept of keeping track just not sure if all these need to be stored.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2301,29 +2290,64 @@ namespace SystemAvailabilityManager {
...
+		TempDiffHi = OptStartSysAvailMgrData( SysAvailNum ).TempDiffHi;
+		ATGWCZoneNumLo = OptStartSysAvailMgrData( SysAvailNum ).ATGWCZoneNumLo;
+		ATGWCZoneNumHi = OptStartSysAvailMgrData( SysAvailNum ).ATGWCZoneNumHi;
+		CycleOnFlag = OptStartSysAvailMgrData( SysAvailNum ).CycleOnFlag;
+		ATGUpdateFlag1 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateFlag1;
+		ATGUpdateFlag2 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateFlag2;
+		NumHoursBeforeOccupancy = OptStartSysAvailMgrData( SysAvailNum ).NumHoursBeforeOccupancy;
+		FirstTimeATGFlag = OptStartSysAvailMgrData( SysAvailNum ).FirstTimeATGFlag;
+		OverNightStartFlag = OptStartSysAvailMgrData( SysAvailNum ).OverNightStartFlag;
+		OSReportVarFlag = OptStartSysAvailMgrData( SysAvailNum ).OSReportVarFlag;
```
rraustad: Save and set on each op start manager.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2301,29 +2290,64 @@ namespace SystemAvailabilityManager {
...
+			}
+			AdaTempGradTrdHeat = OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradTrdHeat;
+			AdaTempGradTrdCool = OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradTrdCool;
+			AdaTempGradHeat = OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradHeat;
+			AdaTempGradCool = OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradCool;
+			ATGUpdateTime1 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTime1;
+			ATGUpdateTime2 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTime2;
+			ATGUpdateTemp1 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTemp1;
+			ATGUpdateTemp2 = OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTemp2;
+		}
```
rraustad, line 2366: These only need to be set for the AdaptiveTemperatureGradient method.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2341,32 +2365,38 @@ namespace SystemAvailabilityManager {
...
 			}
 			if ( ! allocated( OptStartData.ActualZoneNum ) ) OptStartData.ActualZoneNum.allocate( NumOfZones );
-			OptStartData.OptStartFlag = false;
-			OptStartData.OccStartTime = 99.99; //initialize the zone occupancy start time
+			if ( BeginDayFlag ) {
+				NumHoursBeforeOccupancy = 0.0; //Initialize the hours of optimum start period. This variable is for reporting purpose.
+				OptStartData.OccStartTime = 22.99; //initialize the zone occupancy start time
+				OptStartData.OptStartFlag = false;
+			}
+
```
rraustad, line 2395: The OptStartData needs to be reset each day. Can get rid of reset for NumHoursBeforeOccupancy, I think. Changed OccStartTime to 22.99 since this is a time of day variable and code used CIELNG+1 so 22.99 = 24.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2341,32 +2365,38 @@ namespace SystemAvailabilityManager {
...
-						goto Loop3_exit;
-					}
+					if ( DayValuesTmr( J, I ) <= 0.0 ) continue;
+					FanStartTimeTmr = I - 1 + 1 / NumOfTimeStepInHour * J;
+					exitLoop = true;
+					break;
 				}
+				if ( exitLoop ) break;
 			}
-			Loop3_exit: ;
```
rraustad, line 2369: Got rid of GoTo in the above 2 loops. No changes otherwise.
Tested and breaks out when schedule is >0.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2808,12 +2834,8 @@ namespace SystemAvailabilityManager {
...
 
 			} else if ( SELECT_CASE_var == AdaptiveTemperatureGradient ) {
-				NumPreDays = OptStartSysAvailMgrData( SysAvailNum ).NumPreDays;
+
 				if ( OptStartSysAvailMgrData( SysAvailNum ).CtrlType == ControlZone ) {
-					if ( ! allocated( AdaTempGradTrdHeat ) ) {
-						AdaTempGradTrdHeat.allocate( NumPreDays );
-						AdaTempGradTrdCool.allocate( NumPreDays );
-					}
 					ZoneNum = OptStartSysAvailMgrData( SysAvailNum ).ZoneNum;
```
rraustad: These got moved up to top of function.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -3110,21 +3128,19 @@ namespace SystemAvailabilityManager {
...
-									} else {
-										ATGWCZoneNumHi = OptStartSysAvailMgrData( SysAvailNum ).ZonePtrs( 1 );
 									}
 									TempDiff = TempTstatAir( OptStartSysAvailMgrData( SysAvailNum ).ZonePtrs( ZoneNum ) ) - OccRoomTSetPointHeat( OptStartSysAvailMgrData( SysAvailNum ).ZonePtrs( ZoneNum ) );
 									TempDiffLo = min( TempDiffLo, TempDiff );
 									if ( TempDiff == TempDiffLo ) {
 										ATGWCZoneNumLo = OptStartSysAvailMgrData( SysAvailNum ).ZonePtrs( ZoneNum );
-									} else {
-										ATGWCZoneNumLo = OptStartSysAvailMgrData( SysAvailNum ).ZonePtrs( 1 );
 									}
```
rraustad, line 3168: This test for zone farthest from set point was just wrong. Save when we find it, don't overwrite if we don't.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -3141,8 +3157,8 @@ namespace SystemAvailabilityManager {
 					} else if ( DayOfSim == BeginDay && BeginDayFlag ) {
 						AdaTempGradTrdHeat = OptStartSysAvailMgrData( SysAvailNum ).InitTGradHeat;
 						AdaTempGradHeat = OptStartSysAvailMgrData( SysAvailNum ).InitTGradHeat;
-						AdaTempGradTrdCool = OptStartSysAvailMgrData( SysAvailNum ).InitTGradHeat;
-						AdaTempGradCool = OptStartSysAvailMgrData( SysAvailNum ).InitTGradHeat;
+						AdaTempGradTrdCool = OptStartSysAvailMgrData( SysAvailNum ).InitTGradCool;
+						AdaTempGradCool = OptStartSysAvailMgrData( SysAvailNum ).InitTGradCool;
 					} else {
```
rraustad, line 3186: Correct typo on original commit.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -3416,6 +3432,26 @@ namespace SystemAvailabilityManager {
...
+		if ( OptStartSysAvailMgrData( SysAvailNum ).CtrlAlgType == AdaptiveTemperatureGradient ) {
+			OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradTrdHeat = AdaTempGradTrdHeat;
+			OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradTrdCool = AdaTempGradTrdCool;
+			OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradHeat = AdaTempGradHeat;
+			OptStartSysAvailMgrData( SysAvailNum ).AdaTempGradCool = AdaTempGradCool;
+			OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTime1 = ATGUpdateTime1;
+			OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTime2 = ATGUpdateTime2;
+			OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTemp1 = ATGUpdateTemp1;
+			OptStartSysAvailMgrData( SysAvailNum ).ATGUpdateTemp2 = ATGUpdateTemp2;
+		}
```
rraustad, line 3478: Save all air loop specific data for use next time through.

rraustad:
I think this is good (best I can tell), waiting on response from user if he thinks results look reasonable. Still need to clean up this function (i.e., lots of OptStartSysAvailMgrData( SysAvailNum ).\* in code could be changed to pointer) but didn't want to clutter code diff's at this point.

rraustad:
User review of results:

Hi Richard

 I can see that the system starts at 3 different times up to 2 hours early (5:00, 5:15, 5:30). That seems correct if the “Maximum Value for Optimum Start Time” is set to 2 hours. It was set to 1 hour in the file I provided.

 What I found interesting was that the heating coil does not operate before 7 AM, which is the as per the original AHU schedule, though the SAT is clearly below the SAT set-point at start-up on several occasions. This would be expected if the heating coil was on a separate availability schedule. I see in the IDF that the heating coil has not been assigned an availability schedule, so I’m not sure why it would operate in that way. 

 That’s all I could gather from the results file at this point.

My response to user:

Thanks for taking the time to review these results.

 The 2 things I see here are that the Optimal start times are <= 1 and that the pump is scheduled off before 7 AM.

 The Optimal start time is always less than or equal to 1 (see columns D-I). If you see a start time < 5 AM, it's due to an adjustment for daylight savings time.

 And the water heating coil won't turn on since the pump is scheduled off.

 Schedule:Day:Interval,
 HVAC_Weekday_Schedule, !- Name
 Fraction, !- Schedule Type Limits Name
 No, !- Interpolate to Timestep
 Until: 01:00, !- Time 1 {hh:mm}
 0, !- Value Until Time 1
 Until: 02:00, !- Time 2 {hh:mm}
 0, !- Value Until Time 2
 Until: 03:00, !- Time 3 {hh:mm}
 0, !- Value Until Time 3
 Until: 04:00, !- Time 4 {hh:mm}
 0, !- Value Until Time 4
 Until: 05:00, !- Time 5 {hh:mm}
 0, !- Value Until Time 5
 Until: 06:00, !- Time 6 {hh:mm}
 0, !- Value Until Time 6
 Until: 07:00, !- Time 7 {hh:mm}
 0, !- Value Until Time 7
 Until: 08:00, !- Time 8 {hh:mm}
 1, !- Value Until Time 8

 Also, the EMS program setup to get the OptimalStart model working probably won't work. For one, it's run after the HVAC system, and two, the OptimalStart model wasn't set up correctly to handle multiple OptimalStart availability managers. If you try running a simulation with only 1 OptimalStart availability managers it most likely will work.

 EnergyManagementSystem:ProgramCallingManager,
 AHU_Control, !- Name
 AfterPredictorAfterHVACManagers, !- EnergyPlus Model Calling Point
 AHU_TempReset; !- Program Name 1

rraustad:
@EnergyArchmage This defect file uses OptimalStart on the HVAC system. The heating coil is not turning on before 7 AM since the pump schedule is 0 up until that time. Is there a flag that needs to be set to over-ride the pump schedule? As is done with fans using the TurnFansOn and TurnFansOff flags. I can see how the fans are turned on with avail managers, but what about the pumps needed for water coils?

rraustad:
I think I see how the plant is activated with the Availability Manager. The user needs to enter the name of the avail manager in the plant loop.

```
AirLoopHVAC,
  AHU.SP,                  !- Name
  AHU.SP Availability Managers,  !- Availability Manager List Name

PlantLoop,
  HHW_Loop Hot Water Loop, !- Name
  ,                        !- Availability Manager List Name
```

I don't see how the plant can handle more than 1 list. If there are multiple air loops served by the same plant, the plant can only look at 1 of the avail managers at a time.

```
AvailabilityManagerAssignmentList,
  AHU.WP Availability Managers,  !- Name
  AvailabilityManager:OptimumStart,  !- Availability Manager 1 Object Type
  AHU.WP Availability;     !- Availability Manager 1 Name

AvailabilityManagerAssignmentList,
  AHU.WI Availability Managers,  !- Name
  AvailabilityManager:OptimumStart,  !- Availability Manager 1 Object Type
  AHU.WI Availability;     !- Availability Manager 1 Name

AvailabilityManagerAssignmentList,
  AHU.SP Availability Managers,  !- Name
  AvailabilityManager:OptimumStart,  !- Availability Manager 1 Object Type
  AHU.SP Availability;     !- Availability Manager 1 Name
```

I guess the plant could have it's own list where all these are listed in a single list for the plant.

rraustad:
New issue is POSSIBLY needed to address pump override when a pump schedule is used and the availability manager requests operation needing a pump (e.g., air loop fans are turned on when water coils are used in the air loop). I say possibly since this schedule is a flow fraction schedule, not a component availability schedule.

```
Pump:VariableSpeed,
   HVACOperationSchd;       !- Pump Flow Rate Schedule Name
```

A user file had a pump flow schedule, I added that to the OptimalStart_RefBldgLargeOfficeNew2004_Chicago example file. The fans turn on prior to the start time, yet the pumps do not. So the question is, should the pumps operate similar to fans when an availability manager might be expected to override a pump flow schedule?

In function SetupPumpMinMaxFlows, the max/min flows are set. If there is a schedule, the schedule fraction is read:

```
//Retrive the pump speed fraction from the pump schedule
if ( PumpEquip( PumpNum ).PumpScheduleIndex != 0 ) {
    PumpSchedFraction = GetCurrentScheduleValue( PumpEquip( PumpNum ).PumpScheduleIndex );
    PumpSchedFraction = BoundValueToWithinTwoValues( PumpSchedFraction, 0.0, 1.0 );
} else {
    PumpSchedFraction = 1.0;
}
```

Then that schedule fraction is used to limit flow:

```
//The pump outlet node Min/MaxAvail
PumpMassFlowRateMin = max( InletNodeMin, PumpMassFlowRateMinLimit );
PumpMassFlowRateMax = min( InletNodeMax, PumpOverridableMaxLimit * PumpSchedFraction );
```

Then if an availability manager is used, the pump can only turn off. This obviously won't override a pump flow fraction schedule = 0 when avail managers request CycleOn.

```
// Override pump operation based on System Availability Managers, should be done elsewhere?  I suppose this should be OK though
if ( allocated( PlantAvailMgr ) ) {
    if ( PlantAvailMgr( LoopNum ).AvailStatus == ForceOff ) {
        PumpMassFlowRateMax = 0.0;
        PumpMassFlowRateMin = 0.0;
    }
}
```

I changed this avail manager test to include CycleOn and the pump does turn on. However, there are other issues here. Handling VFD is one issue I see, where code may need to be modified to set PumpOverridableMaxLimit according to ... something.

```
// Override pump operation based on System Availability Managers, should be done elsewhere?  I suppose this should be OK though
if ( allocated( PlantAvailMgr ) ) {
    if ( PlantAvailMgr( LoopNum ).AvailStatus == ForceOff ) {
        PumpMassFlowRateMax = 0.0;
        PumpMassFlowRateMin = 0.0;
    } else if ( PlantAvailMgr( LoopNum ).AvailStatus == CycleOn )  {
        PumpMassFlowRateMax = min( InletNodeMax, PumpOverridableMaxLimit );
        PumpMassFlowRateMin = 0.0;
    }
}
```

Additionally, if an availability manager is used on an air loop, for example, optimum start avail manager, and the air loop has a chilled and hot water coil, the availability manager used for the air loop also needs to be entered in the plant loop object. The bad news is that the CycleOn command will turn on both the hot and cold water loop pumps. I expect that this also impacts NightCycle manager and possibly others.

If a pump flow schedule is not used, it appears that the plants operate as expected (since PumpSchedFraction = 1).

rraustad:
This is ready for review.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2316,46 +2321,55 @@ namespace SystemAvailabilityManager {
 		Real64 NumHoursBeforeOccupancy; // Variable to store the number of hours before occupancy in optimum start period
 		bool exitLoop; // exit loop on found data
 
+		auto & OptStartMgr( OptStartSysAvailMgrData( SysAvailNum ) );
```
Myoldmopar, line 2324: While under the hood, this is a pointer, we typically refer to this as a [reference](http://www.tutorialspoint.com/cplusplus/cpp_references.htm).  The word pointer raised my eyebrows, but when I saw it was a reference, I lowered them back down.  The change to a reference is a nice clean up of the code.  Possibly for performance, but even if not, it helps readability.

rraustad:
Bogus diff's, all green before and expect the same result when merged with develop.

rraustad:
So should I pull develop and run CI again?

nrel-bot-3:

Myoldmopar:
I'll merge develop in to clean this back up to green.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -1643,7 +1645,12 @@ namespace SystemAvailabilityManager {
 		if ( allocated( HiTurnOnSysAvailMgrData ) ) for ( auto & e : HiTurnOnSysAvailMgrData ) e.AvailStatus = NoAction;
 		if ( allocated( LoTurnOffSysAvailMgrData ) ) for ( auto & e : LoTurnOffSysAvailMgrData ) e.AvailStatus = NoAction;
 		if ( allocated( LoTurnOnSysAvailMgrData ) ) for ( auto & e : LoTurnOnSysAvailMgrData ) e.AvailStatus = NoAction;
-		if ( allocated( OptStartSysAvailMgrData ) ) for ( auto & e : OptStartSysAvailMgrData ) e.AvailStatus = NoAction;
+		if ( allocated( OptStartSysAvailMgrData ) ) {
+			for ( auto & e : OptStartSysAvailMgrData ) {
+				e.AvailStatus = NoAction;
+				e.isSimulated = false;
```
Myoldmopar, line 1651: So should we be resetting `isSimulated` on all of these availability manager types?

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2341,32 +2381,44 @@ namespace SystemAvailabilityManager {
...
 			for ( I = 1; I <= 24; ++I ) {
 				for ( J = 1; J <= NumOfTimeStepInHour; ++J ) {
-					if ( DayValues( J, I ) > 0.0 ) {
-						FanStartTime = I - 1 + 1 / NumOfTimeStepInHour * J;
-						goto Loop1_exit;
-					}
+					if ( DayValues( J, I ) <= 0.0 ) continue;
+					FanStartTime = I - 1 + 1 / NumOfTimeStepInHour * J;
+					exitLoop = true;
+					break;
```
Myoldmopar, line 2407: I very much appreciate the removal of these goto statements.  Although they aren't _entirely_ evil, they are often _misused_ in evil ways.  In this code, I don't suspect a big performance hit from adding the `if ( exitLoop ) break;` check in there.  In performance critical code, I think another approach would be to put this nested loop in its own function, and do a `return` when you are done, essentially jumping like a `goto`, but in a much cleaner way.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2808,13 +2857,9 @@ namespace SystemAvailabilityManager {
 				}
 
 			} else if ( SELECT_CASE_var == AdaptiveTemperatureGradient ) {
-				NumPreDays = OptStartSysAvailMgrData( SysAvailNum ).NumPreDays;
-				if ( OptStartSysAvailMgrData( SysAvailNum ).CtrlType == ControlZone ) {
-					if ( ! allocated( AdaTempGradTrdHeat ) ) {
```
Myoldmopar, line 2813: I guess you found that these `allocated` checks are just not needed?  It's always nice to remove checks like this when we know for sure they aren't needed since the compiler surely can't optimize these out.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2301,35 +2297,79 @@ namespace SystemAvailabilityManager {
...
+		ATGUpdateFlag1 = OptStartMgr.ATGUpdateFlag1;
+		ATGUpdateFlag2 = OptStartMgr.ATGUpdateFlag2;
+		NumHoursBeforeOccupancy = OptStartMgr.NumHoursBeforeOccupancy;
+		FirstTimeATGFlag = OptStartMgr.FirstTimeATGFlag;
+		OverNightStartFlag = OptStartMgr.OverNightStartFlag;
+		OSReportVarFlag = OptStartMgr.OSReportVarFlag;
+
+		if ( OptStartMgr.CtrlAlgType == AdaptiveTemperatureGradient ) {
+			NumPreDays = OptStartMgr.NumPreDays;
+			if ( ! allocated( AdaTempGradTrdHeat ) ) {
```
Myoldmopar, line 2348: Oh, I see, you just moved the checks up here.  OK cool.

src/EnergyPlus/SystemAvailabilityManager.cc
Myoldmopar, line 1651: nm, I see that isSimulated is only on that one struct.

rraustad, line 1651: Yes, only on this struct. Some AvailManagers can be called by air loops and plants. I was trying to avoid resimulating the same AvailManager twice in the same time step. Function ManageSystemAvailability has two loops, one for air loops and one for plants. This change should correct that. When the plant does simulate the same AvailManager, the AvailStatus is returned without resimulating (line 2327). There may be other AvailManagers that also need to be monitored so they are not simulated twice.

PR #5627, opened 2016-04-22
Refactor Controller:OutdoorAir functions

mjwitte:
## Pull request overview

Partially addresses #4633 and more.  See [design doc](https://github.com/NREL/EnergyPlus/blob/develop/design/FY2016/OAControllerMakeover-Issue4663.md) for details.

This pull request is the initial refactoring of `MixedAir::CalcOAController` and related functions.  
- Portions of `CalcOAController` have been split out into `CalcOAEconomizer` and `CalcMechVentController`.  
- Merged `DataAirloop::OAControllerInfo` into `MixedAir::OAController`.
- Deleted stray using statements.
- `CalcOAController`, `CalcOAEconomizer`, `SizeOAController`, `UpdateOAController` and `Checksetpoints` are now member functions of `OAControllerProps`.
- `ReportOAController` was an empty function and was deleted.
- `SetOAControllerData` was used by `HVACStandAloneERV` to populate `OAController objects` for `ZoneHVAC:EnergyRecoveryVentilator:Controller`. This function was deleted. `HVACStandAloneERV` now fills the `OAController` data directly.
- Converted some parts of `InitOAController` to use a single reference instead of full variables like `OAController( ControllerNum )`.
- In `CalcMechVentController` eliminated loops over all People objects just to count up occupants in one zone (both nominal and current total occupants are available from other data.).  Still needed to keep [one of these loops](https://github.com/NREL/EnergyPlus/blob/0fce018d6514b182caa183b16a2280d6267fb063/src/EnergyPlus/MixedAir.cc#L3881-L3886) for `SystemOAMethod == SOAM_ProportionalControlDesOcc` that adds up the CO2 output for the design number of people at the current activity level.  That could probably be collected elsewhere also to avoid extra steps here.

This PR should produce no diffs.  Then a new branch and PR will make the needed functional changes.
### Work Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
- [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
  - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
  - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
  - NewFeature: This pull request includes code to add a new feature to EnergyPlus
  - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
[...]

Issue #4633, opened 2014-12-17
Controller:OutdoorAir Field: Minimum Fraction of Outdoor Air Schedule Name


kbenne:
I have reviewed the source code pertaining to "Minimum Fraction of Outdoor Air Schedule Name" and made observations in actual simulations that are at least worth some discussion, if not considered a bug.  My investigation has been motivated by the way in which OpenStudio handles ventilation and the ControllerOutdoorAir.  First it should be noted that when OpenStudio creates an outdoor air system the Controller:OutdoorAir object will always reference a Controller:MechanicalVentilation object which in turn references a DesignSpecificationOutdoorAir object that ultimately defines the ventilation for the attached zones.  That is to say OpenStudio puts the mechanical ventilation controller in charge, and therefore overpowering the minimums in the oa controller itself.  This was by design to support DCV as well as the other calculation methods offered by the mechanical ventilation controller such as VentilationRateProcedure, IndoorAirQualityProcedure, ProportionalControl, and IndoorAirQualityProcedureGenericContaminant.  One complication with this strategy is that the E+ promoted method of turning off mechanical ventilation during night cycle or off hours, that is using the Controller:OutdoorAir "Minimum Outdoor Air Schedule Name", is powerless to turn things off against the mechanical ventilation controller.  This is consistent with the documentation about mechanical ventilation controller so no issue there.  OpenStudio has acknowledged the limited power of min oa schedule, and as a result promoted using the max oa fraction schedule to get things turned off instead.  I tested this guidance on a single zone system I believed to see it working.  https://drive.google.com/drive/#folders/0B6AYyX2ggNyBajM5a29JYTlyNnM/0B6AYyX2ggNyBRW9SUmJWN0ZIcWc  Turns out it doesn't work that well, despite the evidence I thought I had in the linked idf.  Doesn't work at all for multi zone vav systems.  Not a big deal actually.  (Except I have given bad advice) The documentation doesn't indicate that it would work, and after reviewing the source code I can't see any reason to expect it would.  

[...]

PR #5794, opened 2016-07-28
Fix Controller:OutdoorAir and Controller:MechanicalVentilation schedules and precedence of limits

mjwitte:
## Pull request overview

Part 2 addressing #4633 and more.  See [design doc](https://github.com/NREL/EnergyPlus/blob/121784677-OAControllerFixes4633/design/FY2016/OAControllerMakeover-Issue4663.md) for details.
#5627 was the initial refactoring of `MixedAir::CalcOAController` and related functions.

This pull request implements the following fixes/changes:
- Fix Controller:MechanicalVentilation input processing when a ZoneList name is used and the Design Specification Zone Air Distribution Object Name is blank.  The correct DesignSpecification:ZoneAirDistribution data is now found in the corresponding Sizing:Zone objects.  Previously the wrong index was used and could cause array bounds errors.
- For Controller:MechanicalVentilation, if the Design Specification Outdoor Air Object Name is blank and there is no corresponding name in a Sizing:Zone object, then the default of 0.0944 m3/s-person is properly applied.
- For Controller:MechanicalVentilation, if the Design Specification Zone Air Distribution Object Name is blank and there is no corresponding name in a Sizing:Zone object, then the defaults of effectiveness=1.0 and recirculation=0.0 are  properly applied.
- Streamline input processing for Controller:MechanicalVentilation (no user impact)
- ~~In Controller:OutdoorAir, autosizing of Minimum Outdoor Air Flow Rate is now set to zero when a Controller:MechanicalVentilation is used.~~ _Follow-up issue #5846._
- In Controller:OutdoorAir, change and document the order of various limit checks such that the Maximum Fraction of Outdoor Air Schedule becomes the last word (unless there is EMS or demand-limiting).  This schedule, if zero, will now shut out all outdoor air, regardless of other settings in Controller:OutdoorAir or Controller:MechanicalVentilation.
- In Controller:MechanicalVentilation, use the Outdoor Air Schedule if specified in DesignSpecification:OutdoorAir. Previously, this schedule was ignored by Controller:MechanicalVentilation. Now it is applied for all System Outdoor Air Method options.  The schedule is multiplied times all zone-level OA flow rates, including IAQ signals.
- Add new output variable "Air System Outdoor Air Mechanical Ventilation Requested Mass Flow Rate."
### Summary of Test Suite Diffs
- [x] Fixes to Controller:MechanicalVentilation getinput for ZoneList resulted in eio diffs for HeatPumpIAQP_DCV, HeatPumpIAQP_GenericContamControl, HeatPumpProportionalControl_DCV, HeatPumpSimpleDCV, HeatPumpVRP_DCV.  eio output we previously incomplete. e.g.:

```
 <Controller:MechanicalVentilation>,Name,Availability Schedule Name,Demand Controlled Ventilation {Yes/No},System Outdoor Air Method,Zone Maximum Outdoor Air Fraction,Number of Zones,Zone Name,DSOA Name,DSZAD Name
- Controller:MechanicalVentilation,DCVOBJECT,VENTSCHEDULE,Yes,IndoorAirQualityProcedure,1.00,3,WEST ZONE,CM DSOA WEST ZONE,CM DSZAD WEST ZONE,EAST ZONE,CM DSOA DCV ZONE LIST,,NORTH ZONE,CM DSOA DCV ZONE LIST,
+ Controller:MechanicalVentilation,DCVOBJECT,VENTSCHEDULE,Yes,IndoorAirQualityProcedure,1.00,3,WEST ZONE,CM DSOA WEST ZONE,CM DSZAD WEST ZONE,EAST ZONE,CM DSOA DCV ZONE LIST,CM DSZAD DCV ZONE LIST,NORTH ZONE,CM DSOA DCV ZONE LIST,CM DSZAD DCV ZONE LIST
```
- [x] RDD diffs from new output variable in files with Controller:MechanicalVentilation.
### Work Checklist

- [x] Unit tests
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes??


testfiles/HospitalBaseline.idf
```diff
@@ -57774,8 +57774,8 @@
     AHU2_OAMinOAFracSchedule,!- Minimum Fraction of Outdoor Air Schedule Name
     ,                        !- Maximum Fraction of Outdoor Air Schedule Name
     ,                        !- Mechanical Ventilation Controller Name
-    No,                      !- Time of Day Economizer Control Schedule Name
-    ,                        !- High Humidity Control
+    ,                        !- Time of Day Economizer Control Schedule Name
+    No,                      !- High Humidity Control
```
Myoldmopar, line 57778: @mjwitte All of these switches intentional?  From how far I've gotten into the PR description and the code changes I don't see the exact reason nor transition changes if we need to apply this change to user files.

mjwitte, line 57778: @Myoldmopar Yes, these are intentional.  These files were wrong from day 1 way back when, but there was never any getinput code to validate these.  And since I reverted the economizer changes back out of this, there still isn't any.  I could at least add those input checks back into this PR if you like.

Myoldmopar, line 57778: No worries, just putting all the changes together in my head.

Myoldmopar:
And I don't really have anything else here.  @EnergyArchmage or @wfbuhl any additional comments before merge?

mjwitte:
@Myoldmopar There is one minor detail here yet.  I think you call them, oh, what is it . . . unit tests.

wfbuhl:
Do we still do those?

mjwitte:
@Myoldmopar This has some minor IDD changes in it - one field name change and several \notes changes.  If I need to undo any of that, let me know.  Working on merge and unit test today.

mjwitte:
@kbenne @macumber This pull request fixes the OA controller schedule problems which you raised in #4633. There are some minor I/O changes in here.  Let me know if these are OK to let in yet this week, or if you want me to hold them for post-v8.6 and just put in the functional code and doc changes.

It contains some minor IDD changes, one field name and some \notes in [DesignSpecification:OutdoorAir](https://github.com/NREL/EnergyPlus/pull/5794/files#diff-23ccf090b80d26e885712256b9a6d888R31041) and Controller:OutdoorAir (just \notes).

It also includes some new output columns in the [Demand Controlled Ventilation table of System Summary report](https://github.com/NREL/EnergyPlus/pull/5794/files#diff-cd588b6af83972de5496d052036a1457R67).

v8.5
image

v8.6 (this branch)
image

kbenne:
@mjwitte I'm ok with the IO changes coming in now.

kbenne:
Also. Thank you!

Myoldmopar:
@kbenne Thanks.  

@mjwitte After the green button is pressed, I'm going to make a tag/pre-release for the new IO freeze point.

mjwitte:
@Myoldmopar @lgu1234 @kbenne  I'm second-guessing this part:  "In Controller:OutdoorAir, autosizing of Minimum Outdoor Air Flow Rate is now set to zero when a Controller:MechanicalVentilation is used."  It makes sense when the mech vent controller is looking at the flows set in DesignSpecification:OutdoorAir and applying DCV.

But for System Outdoor Air Method=IndoorAirQualityProcedure, IndoorAirQualityProcedureGenericContaminant, and IndoorAirQualityProcedureCombined (_which I just realized is present but currently undocumented_) if the IAQ setpoint is satisfied then the system OA could go to zero. It never uses the values in DesignSpecification:OutdoorAir.  Before these changes, the DesignSpecification:OutdoorAir flows will set the floor on min OA flow if they are also references in the Sizing:System object.

Maybe I should autosize the Min OA Flow Rate to zero only if Controller:MechanicalVentilation uses a method _other than_ IAQ\* ?

mjwitte:
@Myoldmopar Here's the unit test.  Didn't get  any comments back about the autosize to zero on min oa flow, so no change to that yet.

mjwitte:
@Myoldmopar @kbenne @MatthewSteen @rraustad  I have decided to back off on this change: 

> "In Controller:OutdoorAir, autosizing of Minimum Outdoor Air Flow Rate is now set to zero when a Controller:MechanicalVentilation is used."

mainly because I am concerned that users (and reviewers) will be surprised by output like this with big zeros in "Minimum Outdoor Air":
image

image

The problem still exists that the autosized Minimum Outdoor Air could silently prevent DCV,  but I think we need a smarter solution than simply setting Min OA flow to zero, because savvy users may already be putting in separate spec for Sizing:Zone to set a real non-zero minimum on DCV OA flow rates.  I'll post a followup issue to improve the autosizing or warnings, and/or add some user control.  If you disagree - speak now.

rraustad:
I don't understand this:
"The problem still exists that the autosized Minimum Outdoor Air could silently prevent DCV"

I assume you mean autosized min OA at 0. If DCV required some ventilation rate, why would it instead be 0?

mjwitte:
@rraustad It's the opposite problem.  Say you use the same DesignSpecification:OutdoorAir in Sizing:Zone and in Controller:MechanicalVentilation and it has X cfm/person and Y cfm/sf.  This field, Minimum Outdoor Air Flow Rate will autosize to the flow at design occupancy and that will always set the outdoor airflow.  DCV will calculate smaller values when occupancy is below design, but this min will always be greater and take precedence.

nrel-bot-3:

Myoldmopar:
@kbenne This has a field name change and some note changes.  Previously you had said it was fine to drop in, but it's pretty late now.  Do those types of changes affect you?

Issue #5770, opened 2016-07-11
User file generating many fancoil max iteration warnings


mjwitte:
## Issue overview

Three files with increasing levels of electric equipment in the data hall zones.  The -P33 file has no fancoil iteration errors, -P66 has a few, and -P100 has a bunch more.  Runtime tripled between -P33 and -P100.  These files may have other issues, but there should be some way for the fancoil to come to a solution quickly even if it can't meet the load or is thrashing around to nail a tiny load.  
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (v8.5.0-https://github.com/NREL/EnergyPlus/commit/39aabddf19b23bc045868ab6effb8431c57510b8)
- helpdesk ticket number 11373
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added to EnergyPlusDevSupport\DefectFiles, run with Singapore weather
- [x] Ticket added to Pivotal for defect (development team task)
- [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Moved to RegulaFalsi and warnings are now gone in P66 version after I moved water coil sensor node to coil outlet node. Testing original P66 and P100 versions now.

rraustad:
Original P66 runs with no fan coil or SimHVAC warnings. P100 runs with no fan coil warnings yet does have SimHVAC max iteration warnings. I wonder if the P100 SimHVAC warnings are due to air loop operation and not fan coil operation? There are so many HVAC systems in these files it's hard to tell what's causing the SimHVAC warnings.

rraustad:
See PR #5800

mjwitte:
Closed via #5800.

PR #5800, opened 2016-08-03
Fan coil max iteration warnings

rraustad:
## Pull request overview

Addresses #5770.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types


tst/EnergyPlus/unit/FanCoilUnits.unit.cc
```diff
@@ -416,17 +416,8 @@ namespace EnergyPlus {
...
-		DataGlobals::DoingSizing = false;
-
-		PlantLoop.deallocate();
-		ZoneSysEnergyDemand.deallocate();
-		FanCoil.deallocate();
-		Node.deallocate();
-		WaterCoil.deallocate();
-		ZoneEquipConfig.deallocate();
-		Zone.deallocate();
-		CoilNames.deallocate();
```
rraustad, line 428: Each unit test deallocated arrays which is done in clear_state. Deleted these as redundant.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2204,15 +2206,20 @@ namespace FanCoilUnits {
 			// speed fan selection only for multispeed cycling fan
 			if ( UnitOn && ( FanCoil( FanCoilNum ).CapCtrlMeth_Num == CCM_CycFan ) ) {
 				QZnReq = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputRequired;
+				if ( QZnReq < 0 ) {
+					Node( FanCoil( FanCoilNum ).ColdControlNode ).MassFlowRate = FanCoil( FanCoilNum ).MaxColdWaterFlow;
+				} else if ( QZnReq > 0 && FanCoil( FanCoilNum ).HCoilType_Num != HCoil_Electric ) {
+					Node( FanCoil( FanCoilNum ).HotControlNode ).MassFlowRate = FanCoil( FanCoilNum ).MaxHotWaterFlow;
+				}
```
rraustad, line 2235: Not sure how this could have worked before without setting the water flow on the nodes. This is not done in Init.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2204,15 +2206,20 @@ namespace FanCoilUnits {
...
 				Node( InletNode ).MassFlowRateMax = FanCoil( FanCoilNum ).LowSpeedRatio * FanCoil( FanCoilNum ).MaxAirMassFlow;
-				Calc4PipeFanCoil( FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOutMax );
 				FanCoil( FanCoilNum ).SpeedFanSel = 1;
 				FanCoil( FanCoilNum ).SpeedFanRatSel = FanCoil( FanCoilNum ).LowSpeedRatio;
+				Calc4PipeFanCoil( FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOutMax );
```
rraustad, line 2240: FanCoil().SpeedFanRatSel is used inside Calc4PipeFanCoil so it needs to be set prior to call.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2244,51 +2248,72 @@ namespace FanCoilUnits {
...
-						Error = ( QZnReq - QUnitOut ) / QZnReq;
-						AbsError = QZnReq - QUnitOut;
-						DelPLR = ( QZnReq - QUnitOut ) / QUnitOutMax;
-						PLR += Relax * DelPLR;
-						PLR = max( 0.0, min( 1.0, PLR ) );
-						++Iter;
-						if ( Iter == 32) Relax = 0.5;
-						if ( Iter == 65 ) Relax = 0.25;
-						if ( Iter > 70 && PLR == 0.0 && DelPLR < 0.0 ) Error = 0.0; // exit loop if PLR = 0
-					}
```
rraustad, line 2270: This iteration loop was oscillating on PLR and not converging until higher iterations were reached and the relaxation factor lowered.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2244,51 +2248,72 @@ namespace FanCoilUnits {
...
+				if ( QUnitOutMax < QZnReq ) {
+					// more cooling than required, find reduced air and water flow rate to meet the load
+					//solve for the cold water flow rate with no limit set by flow rate lockdown
+					Par( 1 ) = double( FanCoilNum );
+					Par( 2 ) = 0.0; // FLAG, IF 1.0 then FirstHVACIteration equals TRUE, if 0.0 then FirstHVACIteration equals false
+					if ( FirstHVACIteration ) Par( 2 ) = 1.0;
+					Par( 3 ) = ControlledZoneNum;
+					Par( 4 ) = QZnReq;
+					Par( 5 ) = double( FanCoil( FanCoilNum ).ColdControlNode );
+					SolveRegulaFalsi( 0.001, MaxIterCycl, SolFlag, PLR, CalcFanCoilPLRResidual, 0.0, 1.0, Par );
```
rraustad, line 2277: New iteration method solves in only a few iterations.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2301,39 +2326,69 @@ namespace FanCoilUnits {
...
-
-					// adjust the PLR to meet the heating load calling Calc4PipeFanCoil repeatedly with the PLR adjusted
-					while ( std::abs( Error ) > ControlOffset && std::abs( AbsError ) > SmallLoad && Iter < MaxIterCycl && PLR != 1.0 ) {
-						// the water flow rate is at the maximum flow rate time the PLR
-						//    Node(FanCoil(FanCoilNum)%HotControlNode)%MassFlowRate = PLR * FanCoil(FanCoilNum)%MaxHotWaterFlow
-
-						if ( FanCoil( FanCoilNum ).HCoilType_Num == HCoil_Water ) {
-							mdot = PLR * FanCoil( FanCoilNum ).MaxHotWaterFlow;
-							SetComponentFlowRate( mdot, FanCoil( FanCoilNum ).HotControlNode, FanCoil( FanCoilNum ).HotPlantOutletNode, FanCoil( FanCoilNum ).HWLoopNum, FanCoil( FanCoilNum ).HWLoopSide, FanCoil( FanCoilNum ).HWBranchNum, FanCoil( FanCoilNum ).HWCompNum );
+				if ( QUnitOutMax > QZnReq ) {
```
rraustad, line 2343: Same methodology used for heating.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -4395,6 +4538,78 @@ namespace FanCoilUnits {
 	}
 
 	Real64
+		CalcFanCoilPLRResidual(
```
rraustad, line 4554: New iteration  function for cycling fan mode.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -4724,87 +4939,6 @@ namespace FanCoilUnits {
 	}
 
 	Real64
-	CalcFanCoilAirFlowResidual(
```
rraustad, line 4727: Function not used anywhere.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2244,51 +2248,72 @@ namespace FanCoilUnits {
...
-				if ( QUnitOutMax < 0.0 ) { // protect against QUnitOutMax = 0 (no water flow, plant off, etc.)
-					// calculate the PLR, if load greater than output, PLR = 1 (output = max)
-					PLR = std::abs( QZnReq / QUnitOutMax );
-					if ( PLR > 1.0 ) PLR = 1.0;
-
-					// adjust the PLR to meet the cooling load calling Calc4PipeFanCoil repeatedly with the PLR adjusted
-					while ( std::abs( Error ) > ControlOffset && std::abs( AbsError ) > SmallLoad && Iter < MaxIterCycl && PLR != 1.0 ) {
-						// the water flow rate is at the maximum flow rate times the PLR
-						//    Node(FanCoil(FanCoilNum)%ColdControlNode)%MassFlowRate = PLR * FanCoil(FanCoilNum)%MaxColdWaterFlow
-						mdot = PLR * FanCoil( FanCoilNum ).MaxColdWaterFlow;
```
rraustad, line 2258: The old iteration method adjusted water flow rate proportional to PLR. However, if fan speed is above 1, shouldn't water flow rate remain at the max? The new routine does the same thing but it looks wrong. For example, whatever load can be met at fan speed = 1 and full water flow is a limit for first stage fan. If the load is slightly higher than that then water flow should remain constant and air flow should be adjusted to meet the load. I'll wait and see what CI returns before I do anything here.

rraustad:
Here's one comparison where zone conditions were not met in V8.5

5770-zoneconditions

Update: the above comparison with V8.5 is not valid since I didn't change anything in the Multispeed operating mode. I compared my branch with develop and do not see where zone temps do not meet set point. I'll look closer for why the CI shows diff's.

5770-zoneconditions_mybranch

UPDATE 2: this is the file with small diff's so there nothing here to worry about. I compared each column and see no more than E-9 diff's. I'll look at the other, big diff files to see what's happening.

5770-fancoilautosize_msfan

rraustad:
Now I can see why the original model appeared to work even though I see some things that don't make sense. The comparison of develop with my branch for this file with large diff's shows relatively the same results. I feel more comfortable now with these changes. I can only see a small difference in FC performance near the end of the cooling operation.

5770-5zonefancoildoascool

rraustad:
This is what I was referring to with water flow rate vs fan speed level. The model performs that same as before, however, I was thinking the water flow rate should max out at speed level 2 or higher. I guess it's just as reasonable to assume the water flow rate could also modulate at higher fan speeds.

Upper figure is water flow rate, lower figure is air flow rate.

5770-chw flow

Here's another view that shows water and air are both proportional to PLR.

5770-chw vs air flow

rraustad:
Since I have to correct a warning, if someone could comment on these results, namely the last comment on water and air flow rate vs FCU fan speed, before that final check in that would be helpful. @Myoldmopar @mjwitte @EnergyArchmage @Nigusse @lgu1234 @wfbuhl

Nigusse:
Just to clarify, for the MultiSpeedFan capacity control the water flow rate is  either at full flow or off position when the Fancoil unit is running at a fan speed higher than the minimum (speed 1).  In this mode the PLR is zero, instead SpeedRatio is used. This is what they call it the "two-position value control" operating mode. When the Fancoil unit is running at speed 1 (minimum speed), then the water flow rate is assumed to vary proportionally to the PLR. 

For the CyclingFan capacity control method, the water flow rate is always proportional to the PLR. That is to say the FanCoil unit is always cycling at any speed level to meet the load. At a given time step the FanCoil unit runs a particular speed level, i.e., doesn't switch speed level. 

CyclingFan and MultispeedFan capacity control methods of FanCoils use different water flow control methodology.

mjwitte:
@rraustad While testing DOA to ZoneHVAC for #5676 I ran the existing DOAToFanCoilSupply and DOAToFanCoilInlet test files as annual simulations.  They both generate errors like this

```
   ** Warning ** Cold Water control failed in fan coil unit SPACE4-1 FAN COIL
   **   ~~~   **   Bad cold water mass flow limits
   **   ~~~   ** ..Water flow rate set to lower limit  Environment=Chicago Ohare Intl Ap IL USA TMY3 WMO#=725300, at Simulation time=03/19 15:45 - 16:00
   ** Warning ** Hot Water control failed in fan coil unit SPACE2-1 FAN COIL
   **   ~~~   **   Bad hot water mass flow limits
   **   ~~~   ** ..Water flow rate set to lower limit  Environment=Chicago Ohare Intl Ap IL USA TMY3 WMO#=725300, at Simulation time=04/15 07:00 - 07:04
```

I'm curious if these errors still happen on this branch.  Would you please test them?

rraustad:
Will do.

rraustad:
Still fails. I was working on CyclingFan mode, these are ConstantFanVariableFlow. RegulaFalsi is already used in this part of the code, there is just a minor glitch in the logic. When RegulaFalsi is called, the residual must be on either side of 0, otherwise RegulaFalsi returns -2 and this message is posted. This is an issue of precisely protecting RegulaFalsi so that a result is always possible.

rraustad:
This code here breaks the RegulaFalsi protection rule. One would think that the FC would always return the same result when water flow rate = 0. So this code would not impact anything. However, in this case, the QUnitOutNoHC result on FirstHAVCIteration = 1112 W, and the next time through it's 1107 W. This means the calculation of the actual load the FC needs to meet is off by 5 W and down goes RegulaFalsi (i.e., the residual result at no flow and full flow do not bound 0).

Water flow rate is set to 0 just prior to this call:

```
// constant fan variable flow
if ( SELECT_CASE_var == CCM_ConsFanVarFlow ) {
      .........

// obtain unit output with no active heating/cooling
Calc4PipeFanCoil( FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOutNoHC, 0.0 );
if ( FirstHVACIteration ) {
    FanCoil( FanCoilNum ).QUnitOutNoHC = QUnitOutNoHC;
} else {
    QUnitOutNoHC = FanCoil( FanCoilNum ).QUnitOutNoHC;
}
// get the loads at the coils
QCoilHeatSP = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputReqToHeatSP - QUnitOutNoHC;
QCoilCoolSP = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputReqToCoolSP - QUnitOutNoHC;
```

I only see this variable (FanCoil( FanCoilNum ).QUnitOutNoHC ) used here so I commented out the IF/ELSE block and no more errors (leaving just these: *\* Severe  *\* FindRootSimpleController: ). Not sure if this was actually needed when Fred changed to RegulaFalsi for that previous issue, or if this was added along the way during that fix and wasn't actually needed. I guess I could run the defect file for that issue to see if it still works.

nrel-bot-3:

mjwitte:
@rraustad Started to review this, merged in develop, but there are several unit tests which fail on this branch, for example :
EnergyPlusFixture.ConstantFanVariableFlowFanCoilCoolingTest:
The difference between QZnReq and QUnitOut is 794.62794312950427, which exceeds 5.0, where QZnReq evaluates to -4000, QUnitOut evaluates to -4794.6279431295043, and 5.0 evaluates to 5.

rraustad:
I'll have a look.

mjwitte:
Tested defect files and the iteration warnings are fixed with small ~1% reductions in energy use with no change in unmet tstat hours.  Ran the six files with big diffs as annual simulations and there is <1% change in energy use and no change in unmet tstat hours, and **HUGE** speed improvement for these files:
FanCoilAutoSize_MultiSpeedFan-Annual drops from _100s to 19s_, and 
DOASDualDuctSchool-Annual went from _3:24 to 1:10_

Removed a few extraneous spaces that crept into the source, will wait for one or two CI reports then merge.

PR #5802, opened 2016-08-04
Post CI-built documentation to s3 for reviewer inspection

Myoldmopar:
@mjwitte For reference, my intent here is to get CI to build the docs and post them to the dashboard on each commit for quick inspection.

@lefticus would this really be all that is needed?  
- The existence of the `s3_upload` key should tell `decent_ci` that there is _something_ to be uploaded, which I guess would be the clue to add the little doc icon?  
  image
- `BUILD_DOCS` was already turned on for this build configuration.
- The documentation build folder (`{CMAKE_BUILD_DIR}/doc-build`) has the pdfs, and a bunch of build subdirectories, but it [looks](https://github.com/lefticus/decent_ci/blob/36e0505b59c461a1958735b36a0e654ca300edb4/send_to_s3.py#L23) like `decent_ci` only cares about the files at the root of the upload folder
- It looks like `decent_ci` will consider the pdfs as binary blobs [here](https://github.com/lefticus/decent_ci/blob/36e0505b59c461a1958735b36a0e654ca300edb4/send_to_s3.py#L34)...is `octect` right?  I always thought it was octet.

OK, if this is all it takes, then I'll feel real dumb for not having enabled it sooner.


Myoldmopar:
Works!

Dashboard:
image

Posted pdfs on Amazon:
image

Merging now so people can take advantage of this.

Issue #5772, opened 2016-07-13
Warmup Convergence Issue in File with Only 3 Surfaces


RKStrand:
## Issue overview

During the review of a fix for GitHub Issue #5261, it was discovered that there was a warmup convergence issue with a file that was testing a new error message.  It seems that there was problems converging during the warmup phase.  This needs to be further investigated to determine the cause of the convergence issue.
### Details

Some additional details for this issue (if relevant):
- Version of EnergyPlus: current version, V8.5
- Unmethours link or helpdesk ticket number
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [x] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte @Myoldmopar 

Investigation of problem:
[...]

PR #5804, opened 2016-08-05
Correction of View Factors for Three or Less Surface Zones

RKStrand:
## Pull request overview

This was originally started as GitHub Issue #5772 where a three surface zone that was pretty basic was not converging during warmup.  A variety of analysis including stepping through with the debugger and code analysis was used to try to figure out what was causing the problem.  Initially, focus was put on the internal radiation algorithm because of recent changes and the fact that the defect file was created during work for allowing zones with fewer than six surfaces without causing warning messages (unless three or less surfaces are defined for a zone since this would not be a complete enclosure).  Upon looking in detail at the view factor algorithms and testing out various combinations, it was discovered that when having a zone of three surfaces (or less), the existing algorithm though guaranteeing reciprocity was leading to view factors for individual surfaces that summed to greater than unity (in some situations the values are greater than 3-4).  This is obviously not physically possible and was corrected by this work.  The original input file that started this defect was investigated further and it was noted that the lack of convergence was actually not a problem with the EnergyPlus algorithm but rather a logical result of the input file that had only two surfaces connected to the exterior environment with relatively high R-values and some mass on the interior.  Debugging showed that there were no instabilities or other anomalies and that the solution was proceeding as it should but simply very slowly toward steady periodic conditions.  The input file was corrected by adding additional exterior surface area to the one exterior wall and this file was added to the test suite.  In addition, documentation was added to the Engineering Reference to clarify the medications to the view factor calculations for zone of three or fewer surfaces.
### Work Checklist

  - Defect: This pull request repairs a github defect issue.  The GitHub Issue resolved was #5772.
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


Myoldmopar:
@RKStrand I tested locally, then pushed some cleanups and merged in develop.  As long as CI comes back clean this will go in.

mjwitte:
@Myoldmopar This is merged, but the branch is not deleted, and #5772 is still open?

PR #5809, opened 2016-08-09
Add Heating Coil that can handle other fuel types

### LINKED ISSUE #5752: Additional fuel type options for furnace heating coils
## Issue overview

Similar to #5746 , furnaces currently only support gas or electric heating coils.  Propane and fuel oil are commonly used fuel types, especially for residential models.  This is similar to #5656 / #5746  which added a fuel type field to OtherEquipment.  In this case, I would suggest adding a fuel type field to Coil:Heating:Gas, and possibly change it's name to Coil:Heating:FuelFired, but maybe not at this point.
### Details

- Platform (all)
- Version of EnergyPlus (v8.5.0)
- Helpdesk ticket number 11461.
### Checklist

nmerket:
## Pull request overview

Fixes #5752 by adding a new heating coil object that can handle other fuel types.
### Work Checklist
- [x] NFP
- [x] idd changes
  - [x] rename object
- [x] code changes
- [x] unit tests
- [x] transition
  - [x] Expand Objects
- [x] integration/regression test files
  - [x] update old files
  - [x] create a new one
- [x] documentation updates
  - [x] IO Ref
  - [x] Engineering Ref
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


nmerket:
@nealkruis I'm fine with that. I know people get weird about deprecating frequently used object. I was kind of leaning towards doing that under the hood anyway. At least combining this new object with the Gas heating object, that is. I was thinking of leaving electric alone.

nmerket:
Okay, new plan. I'd like to make the following changes to the idd:
- `Coil:Heating:Gas` is to be replaced with `Coil:Heating:Fuel` or something like that. It's identical to the original except that it has a user specified fuel type. The file will rename and transition objects from earlier EnergyPlus releases.
- `Coil:Heating:Electric` will remain the same.

Myoldmopar:
That sounds like a good forward-looking plan.  I think the next phase of this would be to go for Coil:Heating ( or _Coil_ ), and this starts a pattern of generalizing the coil.  I'm not sure about the Fuel name, but you are right, "something like that".  Let me kick off that conversation with some _really good_ ideas:
- Coil:Heating:NotElectric
- Coil:Heating:GasOrWhatever
- Coil:Heating:Fire
- Coil:Heating:DinosaurPowered
- Coil:Heating:UserSpecifiedFuelType

nmerket:
My favorites are GasOrWhatever and DinosaurPowered.

EnergyArchmage:
What about multistage flavors of the coils?  Are you sure don't also need a Coil:Heating:Fuel:MultiStage?

If you have to transition Coil::Heating:Gas to anything else, I say might as well go for changing it all the way to Coil:Heating  (and Coil:Heating:MultiStage) .  The down side of trying to phase things is that the name changes keep happening more often.

nmerket:
The reason to not go to `Coil:Heating` is that I won't be getting rid of `Coil:Heating:Electric`. As for multi stage coils. It's certainly doable, but I'm trying to get this done before the code freeze, so I'm trying to keep it simple.

nmerket:
So, how do we make a final decision about what to call this thing? This will end up replacing Coil:Heating:Gas.

shorowit:
`Coil:Heating:Fuel` seems like a logical name to me.

EnergyArchmage:
"DirectFired" is used in an absorption chiller heater.  so maybe Coil:Heating:DirectFired.

mjwitte:
In this context DirectFired (or the earlier suggestion of Direct) could be confused with direct-fired gas burners (where the combustion products go into the supply air) which is _not_ what we model here.  

I'm ok with `Coil:Heating:Fuel`. 
Other options:
`Coil:Heating:FuelFired`
`Coil:Heating:Combustion`
`Coil:Heating:Burner`.

nmerket:
`Coil:Heating:Fuel` it is.

nmerket:
@Myoldmopar this is ready for review and to be merged.

mjwitte:
@nmerket In the IDD, the Fuel Type field should not be `\required-field`, because it has a `\default`.  (EnergyPlus allows this, but it's confusing to interfaces, so #5817 gets rid of those combinations, among other things.)

mjwitte:
And this object should have a `\min-fields 7` to ensure the default gets filled.

nmerket:
@mjwitte that should take care of your comments on the idd.

PR #5815, opened 2016-08-15
Fixing calculation of EMPD surface mass transfer coefficient based on known heat-transfer coefficient

nmerket:
In the process of setting up the new EMPD model #5093, @jdwoods21 found and fixed the following bug:

> the equation below was changed because the previous equation was incorrect. The specific heat (PsyCpAir...) is defined on a dry-air basis [J/kg_da-K], so you want to multiply only by the density of air, (PsyRhoAir...) [kg_da/m^3]. Adding to this the vapor density (RhoVaporAirIn) puts the density on a total air/vapor mixture basis [kg_total/m^3], which is inconsistent with how the specific heat is defined, and thus gives the incorrect answer.

@Myoldmopar We pulled this out of the new feature EMPD branch.


mjwitte:
@nmerket Can you make the title and description more specific about the scope of this change?  If appears this impacts only surface moisture balance calculations?

nmerket:
@Myoldmopar What do we need to get this in?

nrel-bot-3:
@nmerket @lgentile it has been 14 days since this pull request was last updated.

mjwitte (commented):
@nmerket @jdwoods21 Does this equation show up in the engineering reference?

nmerket:
Not that I'm aware of. @jdwoods21 did you see this equation in the engineering reference?

jdwoods21:
No, I did not see it in the engineering reference. I’m fairly certain it is not in there.

mjwitte:
@nmerket @jdwoods21 Merged.

nmerket:
Thank you.

Issue #5825, opened 2016-08-21
Daylighting followup


mjwitte:
## Issue overview

Followups from DElight integration #5725 
- [x] Restore delight stream and unit test or remove commented code in simulationmanager.
- [x] Question about eio diffs - DElight reporting is now for two zones instead of one.  Is this correct?
- [x] Complete or remove TEST_F(EnergyPlusFixture, DElightManagerF_InputGenerator_Test)
- [x] Add transition of daylighting reference point output:variable key names
- [x] Make Glare Calculation Azimuth Angle of View Direction Clockwise from Zone y-Axis not required
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
- Unmethours link or helpdesk ticket number
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

JasonGlazer:
The EIO changes were incorrect, added code to detect condition.

Myoldmopar:
Closed via #5826

PR #5826, opened 2016-08-22
Fix DElight illuminance reference point zero outputs; other daylighting cleanup

JasonGlazer:
Fixed DElight "Daylighting Reference Point Illuminance" output always zero.

Various cleanup from daylighting enhancements added for v8.6:
Fixed EIO differences. Cleaned up commented code in simulationmanager. Removed unsupported stream support for Delight.in file from EnergyPlusFigure.cc. Removed unused unit test. 
Reference issue #5825 
Reference original pull request #5725 
### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types
- IDD changes:
  - [x] If transition, add rules to spreadsheet
  - [x] If transition, add transition source
  - [x] If transition, update idfs

….


mjwitte:
@JasonGlazer I just remembered one more item added to #5825 
Add transition of daylighting reference point output:variable key names

JasonGlazer:
The EIO diffs are as expected.

JasonGlazer:
@Myoldmopar and @mjwitte This is ready to be merged.

JasonGlazer:
The EIO diffs are the same as previous and are OK.
The AUD and Table diffs have to do with the new field that can be defaulted and show up as changed lines in the AUD file and changed values for the number of fields defaulted in the Table files. 

@mjwitte and @Myoldmopar, this is ready to be reviewed and merged.

JasonGlazer:
@mjwitte and @Myoldmopar I hope this is on your list of pull requests to try to review and merge prior to the 8.6.0 release.

mjwitte (changes requested):
@JasonGlazer Tested 5 somewhat random daylighting files and I'm still getting a variable not found warning with PurchAirWithDoubleFacadeDaylighting.idf

```
   ** Warning ** The following Report Variables were requested but not generated
   **   ~~~   ** because IDF did not contain these elements or misspelled variable name -- check .rdd file
   ************* Key=WEST ZONE_DAYLREFPT1, VarName=DAYLIGHTING REFERENCE POINT 1 ILLUMINANCE, Frequency=Hourly
   ************* Key=WEST ZONE_DAYLREFPT1, VarName=DAYLIGHTING REFERENCE POINT 1 GLARE INDEX, Frequency=Hourly
   ************* Key=WEST ZONE, VarName=DAYLIGHTING LIGHTING POWER MULTIPLIER, Frequency=Hourly
```

All of these should have Key=WEST ZONE_DAYLCTRL

src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
```diff
@@ -718,6 +718,17 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
...
                 ENDIF
+
+                 ! For Output:Variable that reference a specific Daylighting:Controls object need update to the new reference point name
+                 IF (InArgs(1) .NE. '*') THEN
+                   IF (SameString(InArgs(2)(1:29),'Daylighting Reference Point 1')) THEN
+                     OutArgs(1) = TRIM(InArgs(1)) // '_DaylRefPt1'
+                   ENDIF
+                   IF (SameString(InArgs(2)(1:29),'Daylighting Reference Point 2')) THEN
+                     OutArgs(1) = TRIM(InArgs(1)) // '_DaylRefPt2'
+                   ENDIF
```
mjwitte: @JasonGlazer Looking at the source code, seems like all of these need a transition rule.

```
Zone,Average,Daylighting Reference Point 1 Illuminance [lux]
Zone,Average,Daylighting Reference Point 1 Glare Index []
Zone,Sum,Daylighting Reference Point 1 Glare Index Setpoint Exceeded Time [hr]
Zone,Sum,Daylighting Reference Point 1 Daylight Illuminance Setpoint Exceeded Time [hr]
Zone,Average,Daylighting Lighting Power Multiplier []
```

And it looks like the Daylighting Lighting Power Multiplier is getting setup differently for delight (zone name) vs splitflux (daylighting control name).  Maybe `DELightManagerF::SetupDElightOutput4EPlus` should be removed and drop the if splitfulx in `DaylightingManager::GetdayLightingControls` for this SetUpOutputVariable.

mjwitte (changes requested):
Still getting warnings after transitioning PurchAirWithDoubleFacadeDaylighting:

```
   ** Warning ** The following Report Variables were requested but not generated
   **   ~~~   ** because IDF did not contain these elements or misspelled variable name -- check .rdd file
   ************* Key=WEST ZONE_DAYLREFPT1, VarName=DAYLIGHTING REFERENCE POINT 1 ILLUMINANCE, Frequency=Hourly
   ************* Key=WEST ZONE_DAYLREFPT1, VarName=DAYLIGHTING REFERENCE POINT 1 GLARE INDEX, Frequency=Hourly
   ************* Key=WEST ZONE, VarName=DAYLIGHTING LIGHTING POWER MULTIPLIER, Frequency=Hourly
```

These should all have Key=West Zone_DaylCtrl which is the daylighting:controls name.

src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
mjwitte: And the transition rules file needs to be updated to include these rules.  This will conflict with #5864 so the rules file will need a manual merge either here or there.

JasonGlazer: @mjwitte which rules files are you talking about? I just made changes to OutputChanges8-5-0-to-8-6-0.md but the Report Variables 8-5-0 to 8-6-0.csv doesn't change because the variable names are still the same, just the keys changed.

mjwitte: I mean the xlsx file.  Seems this needs to be mentioned somewhere so interface developers know that the key has changed.

JasonGlazer:
@mjwitte please review the changes I made. The Rules8-4-0-to-8-5-0.xls format is not perfect for documenting this kind of transition but I put something in.

JasonGlazer:
@mjwitte should I re-transition the files also or is there a final application of transition that occurs in the CI somehow? If I do, where should I grab the 8.5.0 files from? I understand that other changes have also been made to them?

mjwitte:
I don't think you need to re-transition anything, we just want to make sure any daylighting testfiles that have specific keys are corrected if need be.  There may only be this one file but I haven't done a search to be sure.  But as you point out, at this point, they should be corrected by hand because other changes have been made.  If you pull out all of the files that have daylighting and run them, you should be able to see quickly if there are any output variable warnings.

mjwitte:
@JasonGlazer Here are two more test files that use a complete set of name output variables.  The outputs for old Daylighting:Controls objects are transitioned OK, but there are still some problems for DElight inputs.
1.  Currently, transition is keeping the old object name from Daylighting:DELight:Controls rather than naming the new Daylighting:Controls object with <zonename>__DaylCtrl so the output variable key name rule doesn't work.  Either need to make the output variable rule smarter, or change the transition to use the <zonename>_DaylCtrl object name.
2.  Transitioned Daylighting:DElight:\* objects are present in the transitioned file as comments.  Was that intentional?  I don't think we want that.
3.  The old output variable name for Daylighting:DElight:ReferencePoint is "Daylighting Reference Point Illuminance" (no number in the name).  These are not getting transitioned to Reference Point 1, 2, 3, etc. Illuminance.  And their old keys are reference point names, which are gettting transitioned to "RefPt 1_DaylCtrl".

[DElightCFSLightShelf-TestAllOutputs.idf.txt](https://github.com/NREL/EnergyPlus/files/483244/DElightCFSLightShelf-TestAllOutputs.idf.txt)
[DElightCFSLightShelf-TestAllOutputs_V850.idf.txt](https://github.com/NREL/EnergyPlus/files/483242/DElightCFSLightShelf-TestAllOutputs_V850.idf.txt)
[DaylightingDeviceShelf-TestAllOutputs.idf.txt](https://github.com/NREL/EnergyPlus/files/483243/DaylightingDeviceShelf-TestAllOutputs.idf.txt)
[DaylightingDeviceShelf-TestAllOutputs_V850.idf.txt](https://github.com/NREL/EnergyPlus/files/483245/DaylightingDeviceShelf-TestAllOutputs_V850.idf.txt)

Myoldmopar:
@kbenne This one also has some small IDD changes.  Not structural, like field order, but a default change to a field sort of thing.  @macumber had asked for this change.

We can drop this in if you are OK with this.

kbenne:
Im good.

Myoldmopar:
@JasonGlazer so what's the plan here?  I'm happy to work on the transition stuff, but I'd rather not have to catch up on everything on my own.  If you want to have a chat to get me up to speed on the issues, I will update the transition source appropriately.  We could do a GoToMeeting if it would be helpful.

Also, note that the Rules file is conflicted.  When you resolve the conflict, just start with a copy of develop's version and make your changes and use that so that the extra EMPD worksheet doesn't get lost again.

JasonGlazer:
@Myoldmopar I am working on it now. I have resolved the Rules file conflict and committed it but not pushed it yet.

Myoldmopar:
OK, great.  Let me know if you do want to chat today.  And uh...better do another fetch because the PRs that just went in may have re-conflicted you.

JasonGlazer:
@Myoldmopar hopefully this address the output variable transition issues that @mjwitte found for the DElight test file.

src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
```diff
@@ -728,6 +727,36 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
                   OutArgs(1)='*'
                   nodiff=.false.
                 ENDIF
+
+                ! For Output:Variable that reference a specific Daylighting:Controls object need update to the new reference point name
+                IF (InArgs(1) .NE. '*') THEN
+                  isDElightOutVar = .FALSE. ! first assume that it is not a DElight related variable
+                  IF (SameString(InArgs(2)(1:27),'Daylighting Reference Point')) THEN
```
Myoldmopar, line 734: So when I transition up the 3 DElight\* example files from 8.5, I don't see a change in the output variables.  This code implies it _should_, right?  Or maybe I'm missing the intent.

Myoldmopar: Oh! It's only if it uses a specific key.  OK.

JasonGlazer: Yes only for a specific key.

Myoldmopar:
Alright, transitioned up the problem example file identified above and it now transitions properly.  I'm ready to merge.  @mjwitte did you have any other items before I push the green button?

src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
mjwitte: @JasonGlazer I'm not seeing code to convert from the delight "Daylighting Reference Point Illuminance" to "Daylighting Reference Point 1 Illuminance", "Daylighting Reference Point 2 Illuminance", etc.
@Myoldmopar You sure the extra test file gets fully transitioned? DElightCFSLightShelf-TestAllOutputs.idf [attached here](https://github.com/NREL/EnergyPlus/pull/5826#issuecomment-248374806)

Myoldmopar: You mean I was supposed to follow along all the way to the third item in that comment?  

OK, yeah, that still needs some work.  @JasonGlazer do you have a path forward to make those transition things?  I can take a crack at it tomorrow if needed.

JasonGlazer: I guess we know my attention span :-)

I will loop back on that one I missed tomorrow morning.

JasonGlazer: @mjwitte and @Myoldmopar: The change I made yesterday removed the _DaylCtrl from the Daylighting Reference Point Illuminance variables which is the right name for the variables. That does not produce any warning messages and when I manually added the 1, 2, 3 that is when an warning message is generated. So the reason I thought everything working yesterday is because the files were not generating warnings, which was my goal.

On the other hand, looking at the CSV generated, those three variables produce all zeros. A different issue that I will try to resolve.

mjwitte: So that means the old DELight output variable name is still getting setup.  That needs to go, unlessssss, you want to   keep Daylighting Reference Point Illuminance and set it up for all reference points.  So, the same value could be reported as <RefpointName>:Daylighting Reference Point Illuminance and <Daylighting_ctrl name>:Daylighting Reference Point 1 Illuminance.  Actually, that's probably better.

mjwitte: In other words, maybe your transition work is complete and you just need to fix the SetUpOutput for reference point illuminance.

mjwitte: Oh, I see now, the outputs are currently dependent on the method, the numbered approach for SplitFlux and the reference point name approach for DElight.  I hadn't understood that before.   How about setting up both types for both methods?

JasonGlazer: I can do that but it seems like it would be confusing to users. Are you sure?

mjwitte: I see three options:
1.  Current, different illuminance output names for SplitFlux vs DElight - if you switch methods, then the available output variable names change.  That could be confusing.
2.  Generate both sets for both methods.  This would let the old variable names pass through and the user can choose between the ones keyed off the control object vs keyed off the reference point name.  It's kind of like node-name variable vs component inlet temperature variables.  This could be confusing because the names are so similar.
3.  Choose one or the other and use it for both methods.

JasonGlazer: I just fixed and pushed an update for the DElight illuminance values always being zero. This was wrong in the 8.5.0 version as well. My guess it has been wrong a while, the wrong intermediate file was being referenced.

JasonGlazer: I think part of what is confusing is due to a subtle design decision that I made to try to make it as seamless as possible for current SplitFlux and DElight users:
- For SplitFlux the name of the Daylighting:Controls object is used as the key for all of "reference point" related output variables. That is why they need to be numbered 1, 2, 3... This is how it was in 8.5.0
- For DElight the name of the Daylighting:ReferencePoint is used as the key for all the "reference point" related output variables. Since the names are unique they are not numbered. This is how it was in 8.5.0

If we are going to redo things, I would say we should drop the numbering and change the key names for the SplitFlux output variables to be the name of the Daylighting:ReferencePoint objects. They would be unique and this would be consistent between DElight and SplitFlux. 

Unfortunately, this is probably a little more work and a little riskier given the timeline we are on for the release. It means changing stuff in EnergyPlus and Transition.  Please let me know if you want me to proceed.

mjwitte: I understand the keeping things similar for current users, but it may look a bit odd for new users under the combined scheme.  If the variables are working now, then probably best to leave as-is and consider one of the other options for future work.

JasonGlazer: Ok, I just added a new issue, #5880

Myoldmopar:
> If the variables are working now, then probably best to leave as-is and ...

When you say as-is, do you mean you are ready to drop it in without any change?

mjwitte:
Yes, assuming the two test files attached [here](https://github.com/NREL/EnergyPlus/pull/5826#issuecomment-248374806) are now transitioning correctly and producing output without variable warnings.  Do you want me to re-test?

Myoldmopar:
If you don't mind.  I'm wrapping up the scalable sizing PR work right now.  Do you want me to be reviewing the new PIU PR?

mjwitte (approved):

mjwitte:
@Myoldmopar I've tested the transition and it works as designed.  And the DElight reference point illuminance outputs now have non-zero values.  I'll leave it to you to merge.  Do we want to change the label/title on this to defect and emphasize fixing the DElight outputs?

Myoldmopar:
Updating title, label, and merging.  Thanks all.

Issue #5665, opened 2016-05-17
Logic problem in coincident plant sizing


EnergyArchmage:
Plant coincident sizing doesn't alwasy work correctly when demand triggers to set a new size but the max flow timestamp is not valid.  The routine ResolveDesignFlowRate should work if there is only a demand-based timestamp but the way it is coded it needs to also have a max flow timestamp.

The || here should be an &&. 
https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SizingAnalysisObjects.cc#L542

And several places need to check one or the other timestamp is null. 
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@EnergyArchmage Is there a defect file for this?

EnergyArchmage:
No, sorry there isn't a defect file, just a unit test.  The large set of fixes that went in with https://github.com/NREL/EnergyPlus/pull/5751 corrected the problem that was generating null timestamps in the first place, now that they are gone this logic problem is hard to reproduce in a running file.  I considered rolling this small change into that PR but there was already a separate issue.

mjwitte:
Got it

mjwitte:
Closed via #5829

PR #5829, opened 2016-08-23
Fix plant coincident sizing logic problem

EnergyArchmage:
## Pull request overview

This PR is for issue #5665.  The logic checking for valid timestamps would not allow for sizing with a valid max demand timestamp and an invalid max flow rate timestamps.  Changed a "||" to an "&&" and added some robustness checks.  

Added unit test to test the condition with a null flow timestamp.

Note that the presence of null timestamps that need to be checked has been corrected by https://github.com/NREL/EnergyPlus/pull/5751 , making this issue somewhat moot.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types


nrel-bot-3:

mjwitte:
@EnergyArchmage Is there a defect file that shows this bug?

mjwitte (approved):
Looks good.  Will merge after CI reports back.

Issue #5833, opened 2016-08-23
Coil:Heating:Gas transition is missing from develop


mjwitte:
## Issue overview

Somehow the transition for Coil:Heating:Gas got lost from the transition source during the flurry of merges.  Need to double check rules file and transition source for anything else that may have been stomped on.
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
- Unmethours link or helpdesk ticket number
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mitchute:
@mjwitte I can roll this into my work over on #5805, so I'll handle this unless you say otherwise.

mjwitte:
Closed via #5836

Issue #5805, opened 2016-08-05
Several testfiles have old version numbers


mjwitte:
## Issue overview

Happened to notice version warnings while testing all of the new (and old) DOATo\* testfiles.  Searching current develop finds files listed below with old version numbers.  Maybe some need to be transitioned up?  Most probably just need the version number changed to 8.6, but should test them all.  

Also, would be good to run all the testfiles through transition with the MakingPretty option set to clean up old field names and fix formatting.
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Ticket added to Pivotal for defect (development team task)
- [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)
### Details

```
\testfiles\AdvancedOutput\ExerciseOutput1.idf(7):   VERSION,8.3;
\testfiles\AdvancedOutput\ExerciseOutput1A-Solution.idf(7):   VERSION,8.3;
\testfiles\5ZoneFanCoilDOAS_HumidifierOnOASystem.idf(108): Version,8.4;
\testfiles\CoolingCoilFreezingPrevention.idf(105):   Version,8.4;
\testfiles\DOAToWaterToAirHPInlet.idf(109):     8.4;                     !- Version Identifier
\testfiles\DOAToWaterToAirHPSupply.idf(109):     8.4;                     !- Version Identifier
\testfiles\ZoneCoupledGroundHTBasement.idf(65):     8.4; !- Version Identifier
\testfiles\ZoneCoupledGroundHTSlabInGrade.idf(52):     8.4; !- Version Identifier
\testfiles\ZoneCoupledGroundHTSlabOnGrade.idf(52):     8.4; !- Version Identifier
\testfiles\BasicsFiles\AdultEducationCenter.idf(87):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise1A.idf(17):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise1B-Solution.idf(80):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise1C-Solution.idf(126):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise1D-Solution.idf(149):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise2.idf(34):   VERSION,8.4;
\testfiles\BasicsFiles\Exercise2A-Solution.idf(131):   VERSION,8.4;
[...]

PR #5836, opened 2016-08-23
Re-adding and completing Coil:Heating:Gas transition.

mitchute:
The transition for this got squished during a merge with develop. It's back, by popular demand, in all of its fantastic glory.  Addresses #5833.
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?

### DIFF
--- a/src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
+++ b/src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
@@ -139,6 +139,7 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
   LOGICAL :: ErrFlag
 
   INTEGER :: I, CurField, NewField, KAindex=0, SearchNum
+  INTEGER :: AlphaNumI
 
   If (FirstTime) THEN  ! do things that might be applicable only to this new version
     FirstTime=.false.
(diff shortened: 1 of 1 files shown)


mjwitte:
@Myoldmopar @mitchute @nmerket Getting there.  The `Coil:Heating:Gas` object is getting transitioned to `Coil:Heating:Fuel`, but references to the coil in other objects are not getting changed.  Looks like that was never part of the original transition for the new coil type.

Myoldmopar:
Wow, that's a great catch @mitchute.  @mjwitte any tips?  Because my inclination would be to just look for all the possible references (obvious ones like Branch and AirLoopHVAC) and find the field numbers that could contain them.    Or....since we _are_ looping over every object already...maybe we just loop over all the fields in that object and replace it if it is Coil:Heating:Gas.  💪 **Brute force** 💪 !

Myoldmopar:
Or, yeah, you know, great catch, @mjwitte 💯 .  The comment author wasn't clear in the email, sorry.

nmerket:
Oh right, I didn't think of that. I defer to your wise judgement, @Myoldmopar. And, yes, thanks for catching that @mjwitte.

mjwitte:
@Myoldmopar There is precedent for that, sort of, when "Load Range Based Operation" was forked into "Cooling Load Range Based Operation and Heating Load Range Based Operation" back in the v1.0.3 transition. There might be another more recent example, but this was the first one I stumbled onto.

 [This block](https://github.com/NREL/EnergyPlus/blob/8280b7e5e1d2ed7a1b0beae2a916669d15039dde/src/Transition/CreateNewIDFUsingRulesV1_0_3.f90#L294-L314) searches the entire idf for plant and condenser operation scheme objects and replaces fields.  Note this comment:
`!  Actually going to "act" directly on the IDFRecord ... scary`
Also note that this is all done early - prior to the remaining transition steps if I'm following correctly.  I'm guessing that was done so that the revised field strings would simply flow through normally from old to new idf.  So, for this one, you could add a list of all possible objects that reference coil:heating:gas/fuel, or eliminate that part and simply cycle through every alpha field (>1) in the entire incoming idf.

Myoldmopar:
I mean, I could imagine a fancy way, creating a map filled with objects mapping to lambda functions that would return a list of alpha index numbers that could have that field.  And then I see just looping over all the Alphas searching for any that start with Coil:Heating:Gas.  I vote for the latter.

Myoldmopar:
OK @mjwitte @mitchute that should take care of that issue.  5ZoneAutoDXVAV transitions up fine with field name updates, branch object cleanups, heating coil fuel changes, and references to the coil cleaned up in the terminal units, branches, etc.

mjwitte:
Tested this with a group of files that I've been using to test another fix.  Now they are all transitioning as expected.  I'm going to merge this, but I'm going to add a comment to #5805 that we need to run all of the 8.5 testfiles through transition and run them to check for any lingering issues.

PR #5841, opened 2016-08-25
Putting latest version of IDF Editor into EnergyPlus repo

JasonGlazer:
Update the EnergyPlus repo with the latest source and binary for IDF Editor. Wrong version was pulled from EnergyPlusBuildSupport repo when transitioned. This version of IDF Editor is 1.48b which is the same version as was distributed in EnergyPlus 8.5.0 installer.


mjwitte:
Pulled this branch, built Win64 release package  and tested IDF Editor.  All good.

Issue #5497, opened 2016-02-19
Incorrect centroid calculated for non-convex quads


dmac000:
When creating a small non-convex building surface, then I get the error:

```
** Severe  ** SetOutBulbTempAt: Surface Outdoor Temperatures < -100 C
**   ~~~   ** ...check Surface Heights - Maximum Surface Height=[110465].
**   ~~~   ** ...according to your maximum Z height, your building is somewhere in the Stratosphere.
**  Fatal  ** Program terminates due to preceding condition(s).
```

with the vertices:

```
9.857091, -27.472747, 11.0,  !- Vertex 1 position
5.065178, -24.431707, 11.0,  !- Vertex 2 position
9.857181, -27.471673, 11.0,  !- Vertex 3 position
14.461132, -20.217025, 11.0; !- Vertex 4 position
```

It looks like the high z-value is coming from `CalcSurfaceCentroid` in `SurfaceGeometry.cc` where 4-sided polygons are split into 3-sided polygons. This split is not correct for non-convex surfaces. The value is calculated as:

```
TotalArea = 0.00485596
Tri1Area = 24.380 / TotalArea = 5020.642
Tri2Area = 24.384 / TotalArea = 5021.642

Tri1Area * 11 + Tri2Area * 11 = 110465
```

mjwitte:
If I'm following the vertices correctly, this is not a well-formed polygon, the sides cross over each other.  Seems that vertices 3 and 4 are reversed?  Also, vertex 3 is nearly on top of vertex1 and the x and y differences between these two are less than the tolerance for geometry.  From the I/O Ref section titled "Surface Vertices":

> Note that the resolution on the surface vertex input is 1 millimeter (.001 meter). Therefore, using vertices that are very close together (<1 mm) may result in invalid dot product and fatal errors during shading calculations.

Were there any other warning messages about this surface in the error file prior to the fatal?

dmac000:
The surface is a v-shape which should be well-formed. The extreme example produces a user-visible error because of the 'stratosphere' check. This is the only error that is shown. More moderate examples produce the wrong values but are not enough to produce an error.

Another example is something like (10, 10, 0), (20, 50, 0), (30, 10, 0), (20, 60, 0). As points 2 and 4 get closer together, the error in the centroid increases because it's forming a triangle using the first 3 points which include the area outside the polygon. Because the area of the triangles over the area of the polygon doesn't add up to 1, then the center point becomes a multiple of what it should be.

[...]

PR #5843, opened 2016-08-29
For non-convex quad fix triangles chosen for proper centroid calculation to avoid unrealistic z values

JasonGlazer:
Addresses #5497 
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


JasonGlazer:
@mjwitte the one file that is showing the new warning message is the 5ZoneGeometryTransform.idf file that contains the GeometryTransform object which is processed in the TransformVertsByAspect() routine. The issue is happening because the GrossArea of the surfaces is computed prior to the application of the GeometryTransform. In each place that the GrossArea is computed right after the CreateNewellAreaVector() call, the call to TransformVertsByAspect happens later. It happens in several places in SurfaceGeometry. This does not make much sense to me but maybe I am not understanding the bigger picture. Do you know why those two calls are in the order that they are?

mjwitte:
Not sure.  Does TransformVertsByAspect need the GrossArea info to do the transform?  Perhaps TransformVertsByAspect needs to re-compute GrossArea?  And possibly other surface properties?  What about the "length" and "width" that get reported in the surface details report?

JasonGlazer:
No, it just uses and changes the vertex() array.

mjwitte:
If you run 5ZoneGeometryTransform with and without the transform, do the various surface reports look correct?  Both table and eio surface details.

JasonGlazer:
The speed seems slow so I am going to merge it again and see how the performance tests go. The 1 failing file, 5zoneGeometryTransform.idf is a known issue and #5849 was posted about it.

JasonGlazer:
This pull request got messed up by merging two active branches by accident. Starting a new branch and pull request. #5852

Issue #5849, opened 2016-09-02
GeometryTransform Not Impacting IVRS Zone Summary or Envelope Summary 


JasonGlazer:
## Issue overview

If you take the 5ZoneGeometryTransform.idf file and comment out the GeometryTransform object, the ZoneInformation portion (as well as others parts) of the EIO file changes but in the HTML file, the Input Verification and Results Summary as well as the Envelope Summary do not change. These should have corresponding changes. 

In addition, code inspection of the SurfaceGeometry.cc file the use of the GeometryTransform object which occurs in the TransformVertsByAspect() function occurs after the area computation which  the computation of Surface().GrossArea by the CreateNewellAreaVector() call followed by the VecLength() call. This does not seem intuitively correct. This code combination occures multiple time in SurfaceGeometry.cc. 

```
CreateNewellAreaVector( SurfaceTmp( CurNewSurf ).Vertex, SurfaceTmp( CurNewSurf ).Sides, SurfaceTmp( CurNewSurf ).NewellAreaVector );
SurfaceTmp( CurNewSurf ).GrossArea = VecLength( SurfaceTmp( CurNewSurf ).NewellAreaVector );

```

This was noticed while working on #5843 and one of the commits rearranges the call but because it caused more diffs than expected it was reverted.
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

PR #5852, opened 2016-09-08
For non-convex quad fix centroid calculation to avoid unrealistic z values

JasonGlazer:
Addresses #5497

This is the second pull request related to this fix. The first got messed up so this is a redo.
### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types


JasonGlazer:
The diffs in the CI are all on the 5ZoneGeometryTransform.idf file. There were issues related to this file found and posted to #5849 during the development of the fix. In addition, the centroid for the areas for some of the surfaces in the file were being miscalculated originally. Since the centroid calculation was fixed, this impacted the surfaces and flowed down to the zone and HVAC impacts. 

@Myoldmopar and @mjwitte I believe this is ready to be reviewed and merged.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -10052,16 +10052,36 @@ namespace SurfaceGeometry {
...
 				}
 
 				// get area fraction of triangles.
-				Real64 const Tri1Area( AreaPolygon( 3, Triangle1 ) / TotalArea );
-				Real64 const Tri2Area( AreaPolygon( 3, Triangle2 ) / TotalArea );
+				Real64 Tri1Area( AreaPolygon( 3, Triangle1 ) / TotalArea );
+				Real64 Tri2Area( AreaPolygon( 3, Triangle2 ) / TotalArea );
+
+				// check if sum of fractions are slightly greater than 1.0 which is a symptom of the triangles for a non-convex quadralateral using the wrong two triangles
+				if ( (Tri1Area + Tri2Area) > 1.05 ) {
```
mjwitte, line 10067: @JasonGlazer What's the basis for >1.05 as the test?  Could a bad surface slip through with, say 1.04?

mjwitte (commented):
Tested with the defect file before/after.  Couldn't find a way to report surface centroids?  So added reporting of outside surface wind speed and temperatures to verify that other surfaces keep the same values and that the offending surface has a reasonable value. That much all looks good.

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: I tried 1.0 and that tripped up a lot of surfaces that were ok so I used engineering judgment to put it at 1.05. I did not do any kind of analysis to justify that number.

mjwitte (approved):

Issue #5853, opened 2016-09-09
Fancoil cold/hot water control failed warnings


mjwitte:
## Issue overview

While testing DOA to ZoneHVAC for #5676 I ran the existing DOAToFanCoilSupply and DOAToFanCoilInlet test files as annual simulations.  They both generate errors like this

```
   ** Warning ** Cold Water control failed in fan coil unit SPACE4-1 FAN COIL
   **   ~~~   **   Bad cold water mass flow limits
   **   ~~~   ** ..Water flow rate set to lower limit  Environment=Chicago Ohare Intl Ap IL USA TMY3 WMO#=725300, at Simulation time=03/19 15:45 - 16:00
   ** Warning ** Hot Water control failed in fan coil unit SPACE2-1 FAN COIL
   **   ~~~   **   Bad hot water mass flow limits
   **   ~~~   ** ..Water flow rate set to lower limit  Environment=Chicago Ohare Intl Ap IL USA TMY3 WMO#=725300, at Simulation time=04/15 07:00 - 07:04
```
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (v8.6.0 5b5301a77f )
- These errors still show after the changes in #5800.
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Why these new warnings? Seems related to chilled water temp and not zone load control.

```
 ** Severe  ** FindRootSimpleController: Controller error for controller = "DOAS COOLING COIL CONTROLLER"
 **   ~~~   **  Environment=Chicago Ohare Intl Ap IL USA TMY3 WMO#=725300, at Simulation time=01/17 12:45 - 13:00
 **   ~~~   **   Controller function is inconsistent with user specified controller action = Reverse action
 **   ~~~   **   Actuator will be set to maximum action
 **   ~~~   ** Controller control type=Temperature
 **   ~~~   ** Controller temperature setpoint = 11.60 [C]
 **   ~~~   ** Controller sensed temperature = 12.15 [C]
 **   ~~~   ** Controller actuator mass flow rate set to 0.59 [kg/s]
 **   ~~~   ** Controller actuator temperature = 12.20 [C]
[...]

Issue #5726, opened 2016-06-13
Untranslated LaTex equations in Component Sizing section of Engineering Reference


wfbuhl:
Generally, the Component Sizing section of of EnergyPlus 8.5.0 as published in PDF looks like crap. Plus, we've got untranslated LaTex equations on pages 505, 523, 544, and 548. These should definitely get fixed.

nealkruis:
@michael-okeefe can you look into this?

michael-okeefe:
Sure, looking into it now. I can do a quick pass through. It's going to require manual correction of the LaTeX source at this point.

Myoldmopar:
Closed via #5736

Issue #5276, opened 2015-10-21
Fan coil unit hot water coil waterside control problem


EnergyArchmage:
Helpdesk issue 10848.  User noticed a severe mismatch in heating energy between the water side and air side during some times of the day.  The water flow is sometimes running at full mass flow rate and large temperature difference and the McdeltaT can be a factor of 10 higher than the air side coil heating rate.  

Fan coil unit has many places where there is While loop that finds a PLR, need make sure that each one of those has code like this before it calls Calc4PipeFanCoil.

```
mdot = PLR * FanCoil( FanCoilNum ).MaxHotWaterFlow;

SetComponentFlowRate( mdot, FanCoil( FanCoilNum ).HotControlNode, FanCoil( FanCoilNum ).HotPlantOutletNode, FanCoil( FanCoilNum ).HWLoopNum, FanCoil( FanCoilNum ).HWLoopSide, FanCoil( FanCoilNum ).HWBranchNum, FanCoil( FanCoilNum ).HWCompNum );
```

EnergyArchmage:
Not having any luck with this.  Initial idea did not work out.  I have tried a number of code changes and nothing seems to work.  Cannot reproduce in an example file.  

In additon to the water side showing more heat transfer that the air side, it also can show air side heat transfer with no water mass flow at all.   

Does not appear that I can deliver a fix on this, have to cut and run at this point.

wfbuhl:
Oh this is ugly. We have a zone in a big, complicated building with only a fancoil. Some of the time in heating mode the fancoil works well - hw flow and heating output move in sync and the zone temperature is controlled. At other times the zone temperature is not well controlled. The hw flow goes from either max (zone temp below set point) to zero (zone temperature above setpoint). Meanwhile hw coil output seems to be random. The good news is that the hw flow rate seems to be trying to do the right thing, but the air side (the hw coil output) is not responding. Could there be something wrong with the hw coil model? Or how the coil output is calculated? This should be interesting.

Myoldmopar:
OK, @mjwitte, this appears to have gone through @EnergyArchmage and @wfbuhl without much promise of a fix yet.  If we want this included in the bug fix release, I guess we need to have a call and figure out a team effort way to fix it??  Or we push it off until after the bug fix release.  What do you think?

mjwitte:
The best thing would be to try to reproduce it with a smaller file.  Is this happening with a particular fan coil control type?

wfbuhl:
[...]

PR #5855, opened 2016-09-10
Fan Coil control may fail when used with DOAS as inlet to fan coil

rraustad:
## Pull request overview

Fan coil may fail to converge when DOAS is used to pretreat outdoor air and delivers that air to the inlet of the fan coil. If the inlet conditions fluctuates during the iteration procedure, the fan coil may not find the correct solution and show a warning related to Cold/Hot water control failed. This correction addresses issue #5853 with a minor change to existing FC control logic. Diff's are not expected in existing example files.

### Work Checklist

- [x] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


rraustad:
I'm not sure I can think up a unit test for this. Any ideas?

rraustad:
I don't like this fix, it doesn't seem like it's reliable. I would rather force the water flow to 0 regardless of flow lock and test the coil off air-side capacity. The SetComponentFlowRate calls just above my attempt to fix appears to return the air-side capacity with what the plant dictates the water flow to be if the plant is locked. If the model needs to know the no-flow capacity, then calculate the no-flow capacity.

This does seem to work but I bet there's a better way.

rraustad (commented):
Added a method to test the coil off capacity prior to setting water flow rate using SetComponentFlowRate. This always provides the off capacity without relying on a previous iteration result. Tested with original defect file #5726 and DOAToFanCoilSupply annual run.

rraustad:
If the CI returns results similar to before then I am comfortable with this change. @wfbuhl I changed the way the off capacity is measured for FCs (previously used FanCoil( FanCoilNum ).QUnitOutNoHC to store results from a previous iteration). This method sets node mass flow to 0 before calling SetComponentFlowRate. Now the model always tests for coil off capacity each iteration. Please review and comment as needed.

rraustad:
Correction: original defect file is #5276. I also uploaded a revised defect file for that issue since they were outdated. Hint: it would be nice to be able to edit our summary.

rraustad:
Well this if F'd up. The SetCompFlowRate function uses the node inlet mass flow rate to set the flow when the flow is locked. How does this work if the model sets the flow between 0 and max and everywhere in between when it iterates on a solution? I would think SetCompFlowRate would use MassFlowRateMin and MassFlowRateMax (or MinAvail and MaxAvail) to lock the flow to a specific value.

Of course, there are 2 unit tests failing because I don't quite understand the methodology. Still, the way I have the FC code now is the way I think it should work. There's no way I am saving the node flow in an equipment model before I set it to something else since the plant should have done that for me.

```
//Set loop flow rate
if ( loop_side.FlowLock == FlowUnlocked ) {
} else if ( loop_side.FlowLock == FlowLocked ) {
    Node( OutletNode ).MassFlowRate = Node( InletNode ).MassFlowRate;
    CompFlow = Node( OutletNode ).MassFlowRate;
```

So this is on hold.

rraustad:
Here's what the unit test does:

```
FirstHVACIteration = false;
PlantLoop( 2 ).LoopSide( 1 ).FlowLock = 1;
Node( FanCoil( 1 ).ColdControlNode ).MassFlowRate = 0.2;
// cooling simulation with flow lock on and locked flow > flow that meets load; bypass extra flow
Sim4PipeFanCoil( FanCoilNum, ZoneNum, ControlledZoneNum, FirstHVACIteration, QUnitOut, LatOutputProvided );
```

I thought this looked odd the first time I looked at it. Now I see why it's like this. The problem is, how can the FC iterate on a solution to meet the zone load if the water-side flow rate is fixed?

@Myoldmopar @wfbuhl chime in here...

rraustad:
Here's another question, what is MassFlowRateMax, MassFlowRateMin, MassFlowRateMinAvail and MaxFlowRateMaxAvail equal to when loop_side.FlowLock == FlowLocked ????

@EnergyArchmage probably need your input as well.

rraustad:
Here's a RegulaFalsi routine that sets the water node flow as it iterates:

```
CalcFanCoilCWLoadResidual(
   Node( FanCoil( FanCoilNum ).ColdControlNode ).MassFlowRate = CWFlow;
   Calc4PipeFanCoil( FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOut, 1.0 ); 
```

This works great to find the flow that meets the load. Then the next time SetCompFlowRate is called, THIS flow is used, not the flow the plant expects. What should happen is we iterate on a solution to find the flow that meets the load, then call SetCompFlowRate to find the flow the plant expects, then do something within the model to resolve the difference (e.g., bypass excess flow, limit flow, etc.).

Still waiting on suggestions.....

EnergyArchmage:
The FC should have already iterated some number of times with FlowLock set to false.  I would want to track down why the FC is getting called again with FlowLock set to true.  

The iterations have to be cut off at some point, or the runtime performance police will shut you down.  It sounds like this is happening at the end and some system is not going to be satisfied.

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:
@rraustad @lgentile it has been 20 days since this pull request was last updated.

Myoldmopar:
So the air-side does get a chance to iterate, right?  Can you add some diagnostic info out to figure out what this does during flow lock false and why it still isn't satisfied once we hit flow lock true?

rraustad:
The airside does iterate. The problem occurred before flow was locked. What was happening with this issue was that the coil off capacity was only calculated when FirstHVACIteration = true. You would think that the coil off capacity would never change, but in this case it did change on the following iteration. Not sure why, just know it did change. Then on this next iteration, RegulaFalsi failed since the load was not between the ACTUAL min/max output of the fan coil.

rraustad:
I will have to circle back to this to refresh my memory on why I chose that sequence of events leading in to the RegulaFalsi function.

rraustad:
Reverted this code and made a simple change. Both example files DOAToFanCoilInlet and DOAToFanCoilSupply now run clean except for:

    ** Warning ** Chiller:Electric:EIR = MAIN CHILLER:  Integrated Part Load Value (IPLV) calculated is not at the AHRI test condition.
    **   ~~~   **  Check limits in Cooling Capacity Function of Temperature Curve, Curve Type = BIQUADRATIC, Curve Name = MAIN CHILLER RECIPCAPFT
    **   ~~~   **  Check limits in EIR Function of Temperature Curve, Curve Type = BIQUADRATIC, Curve Name = MAIN CHILLER RECIPEIRFT

Running defect file in V8.6 and this branch now. Will update when finished.
Not sure where I got the impression that 5276 was the defect file for this issue. Ran it in V8.6 and this branch and did not see any FC control failed warnings. There are RootFinder errors in the 2 example files since the CW temp is greater than the mixed air temp SP (needs to be changed from 12.2 to 11.2 in the CW OA temp reset SPM).

Also need to address unit test.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -1938,13 +1941,14 @@ namespace FanCoilUnits {
...
 			Calc4PipeFanCoil( FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOutNoHC, 0.0 );
-			if ( FirstHVACIteration ) {
-				FanCoil( FanCoilNum ).QUnitOutNoHC = QUnitOutNoHC;
-			}
-			else {
+
+			if ( ColdFlowLocked || HotFlowLocked ) {
 				QUnitOutNoHC = FanCoil( FanCoilNum ).QUnitOutNoHC;
+			} else {  // continue to update QUnitOutNoHC while flow is unlocked
+				FanCoil( FanCoilNum ).QUnitOutNoHC = QUnitOutNoHC;
```
rraustad, line 1936: This is the simple fix. Continue updating QUnitOutNoHC until flow is locked (instead of only on FirstHVACIteration). The only problem I see is if the plant resolver locks the flow at something other than what it was the previous iteration (starved flow, etc).

rraustad:
This looks like it correct the problem. Ready for final review.

mjwitte:
@rraustad One of the linux builds failed here complaining about the license year. Please merge in develop one more time and this should pass the tests.

rraustad:
@Myoldmopar what do the errors mean in the x86_64-Linux-Ubuntu-14.04-gcc-4.8 tests (although there is a green check mark, 57 tests failed)?

First of 57 errors:

    regression.1ZoneUncontrolledFourAlgorithms 4.14175 

    Device id: x86_64-Linux-Ubuntu-14.04-gcc-4.8
    Comparing `/home/elee/ci/EnergyPlus-14e9219508-x86_64-gcc-4.8/build/testfiles/1ZoneUncontrolledFourAlgorithms` with `/home/elee/ci/EnergyPlus-9dee229531-x86_64-gcc-4.8/build/testfiles/1ZoneUncontrolledFourAlgorithms`
    Processing (Diffs) : 1ZoneUncontrolledFourAlgorithms
    Traceback (most recent call last):
     File "/home/elee/ci/EnergyPlus-9dee229531-x86_64-gcc-4.8/regressions/Testing/RegressionSuite/Scripts/RunComparison.py", line 78, in  runner.process_diffs_for_one_case(entry, base_dir, mod_dir, mod_dir)
     File "/home/elee/ci/EnergyPlus-9dee229531-x86_64-gcc-4.8/regressions/Testing/RegressionSuite/Scripts/runtests.py", line 445, in process_diffs_for_one_case path_to_tablediff_log)))
     File "/home/elee/ci/EnergyPlus-9dee229531-x86_64-gcc-4.8/regressions/Testing/RegressionSuite/Scripts/TableDiff.py", line 408, in table_diff diff_dict[h] = map(lambda x, y: thresh_abs_rel_diff(abs_thresh, rel_thresh, x, y), hdict1[h], hdict2[h])
    KeyError: /#Shading Surfaces

mjwitte:
These started popping up in January after CI was restarted (or maybe after the VS2015 changes were done?)  @Myoldmopar Hasn't diagnosed it yet, but it's not your branch.

Myoldmopar:
Concur.  Not your branch.  Still diagnosing.  Carry on.

src/EnergyPlus/FanCoilUnits.cc
Myoldmopar: This seems quite reasonable.  I'll build/test locally, but this makes sense.  My only question is really whether this same catch needs to be extended to other component types.

Myoldmopar:
Looks good.  Merging.

src/EnergyPlus/FanCoilUnits.cc
rraustad: I think this may also impact other water coil models and there is no way to tell unless a user points something out. So to test I have been thinking about a fatal error in SetComponentFlowRate to see if there are failures in the example files (just to get us started on figuring this out). The problem is that when flow is locked, the water outlet node mass flow rate is set equal to the inlet node mass flow rate instead of some other value which we could compare against.

    } else if ( loop_side.FlowLock == FlowLocked ) {
        Node( OutletNode ).MassFlowRate = Node( InletNode ).MassFlowRate;
        CompFlow = Node( OutletNode ).MassFlowRate;
    } else {
        ShowFatalError( "SetComponentFlowRate: Flow lock out of range" ); //DEBUG error...should never get here
    }

when flow is locked down, if the Node() data held the value of the locked flow then we could compare inlet node flow to the locked value.

    } else if ( loop_side.FlowLock == FlowLocked ) {
        Node( OutletNode ).MassFlowRate = Node( InletNode ).LockedMassFlowRate;
        if(  Node( InletNode ).MassFlowRate !=  Node( InletNode ).LockedMassFlowRate ) ShowFatalError....

@EnergyArchmage and @mjwitte might have other suggestions.

PR #5858, opened 2016-09-13
Fix ASIHPMixedTank crash, sizing problems, and output variable names

shenb:
Addresses #5845 regarding the crash with running ASIHPMixedTank.idf test file on Linux.  After the fix, diffs in running ASIHPMixedTank.idf are expected. Also fix size inconsistencies in the test file, and update the output variable names.

PR #5859, opened 2016-09-13
Updating version numbers. Transition as required.

mitchute:
Stems from issue #5805. Some files have old version numbers. This brings the version number up to date and runs transition as necessary.


mjwitte:
@mitchute Is this ready to merge?  Should the new py script stay?

mitchute:
@mjwitte It should be. The only question I had was regarding the ASIHPMixedTank integration failure on the previous commit. That should not have anything to do with what I've done though since the input file has not changed. 

I pulled this branch into the makingPretty branch and removed the py script over there. We can remove it here too, or wait for that branch to drop in when it will be removed.

mjwitte:
The ASIHP integration failure is known and has a pending PR #5858.  The py script can stand here.  I'll merge once the last to CI reports come in.

PR #5245, opened 2015-10-04
Improve Dataset Support

Myoldmopar:
- [x] Add CI input processing to help validate the inputs of dataset files
- [x] Update the version number comments of the dataset files to V X.Y+ so that we don't have to force change them unless the inputs change
- [x] Try out typedef'd test fixture for datasets to help organize unit test results
- [x] Address dataset unit test failure: Chillers  -- Ran the Chillers file through transition, passes now
- [x] Address dataset unit test failure: PrecipitationSchedulesUSA_fullset_Release  -- Not an actual dataset file, just plain text.  Removed it from the unit test list
- [x] Address dataset unit test failure: RefrigerationCases  -- Not packaged with installer, appears out of date, removed
- [x] Address dataset unit test failure: RefrigerationCasesSetNumberOne  -- Not packaged with installer, appears out of date, removed
- [x] Address dataset unit test failure: ResidentialACsAndHPsPerfCurves  -- The objects in this dataset are full-on stubs missing many required fields.  ~~I'm removing the unit test from this until we decide to either flush out the objects, or comment the stubbed objects out.~~ I updated the dataset file to comment out, but leave in, the generic stubbed out coil objects. The curves are still being processed by the input processor.

PR #5860, opened 2016-09-14
Prettify IDFs

mitchute:
Ran transition with makingPretty enabled on the IDFs. This is the result. This cleans up field names, formatting, etc.


mjwitte:
@mitchute Thanks.  This _should_ be no diffs.  Interesting that min-fields filled out a bunch of VAV terminal units with some additional blank fields at the end.  Also, we need to be sure to change the `MakingPretty=.true.` lines back to false before this gets merged.

mitchute:
@mjwitte That should bring `makingPretty = .false.` back.

mjwitte:
@mitchute This [comment](https://github.com/NREL/EnergyPlus/pull/5245#issuecomment-145610764) in #5245 explains the original justification for having the blanket comment that doesn't necessarily need to change with every release.

mitchute:
OK. Works for me. The way it is worded looks like we can leave it.

mjwitte (approved):
@Myoldmopar This is good to go anytime.  I'll hold on the merge to avoid clouding other pull requests that are ready.

Myoldmopar:
The Version is also a `\unique-object`, so if someone imports the whole dataset with, for example, `EP Macro`, it will get two version IDs, and I _think_ E+ will balk at that.

mjwitte:
@mitchute Didn't notice until now that merging #5859 caused conflicts in this branch.

mjwitte:
Or it might have been a different merge.

mitchute:
@mjwitte The issue was in `RadLoTempCFloHeatCool_AddedAutosizing.idf`. Only numbers had changed so I accepted the `develop` values.

mjwitte:
Merging this one - may cause conflicts with other pull requests that have test idf file changes.

Issue #5847, opened 2016-08-31
Transition for VAV terminal damper heating action


mjwitte:
## Issue overview
#5646 added a new choice for VAV terminal "Damper Heating Action" and changes the default.  Transition rules and code need to be added.
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
- Unmethours link or helpdesk ticket number
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Fixed via #5864

PR #5864, opened 2016-09-15
Update transition file for missing pieces

Myoldmopar:
Some things just got missed in the Rules 8.5 to 8.6 file.


src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
```diff
@@ -478,7 +478,16 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
...
-              CASE('BRANCH')
+            CASE('AIRTERMINAL:SINGLEDUCT:VAV:REHEAT')
+                nodiff=.false.
+                OutArgs=InArgs
+                IF (SameString(InArgs(16), 'REVERSE')) THEN
+                  IF ( (.NOT. SameString(InArgs(17),'')) .OR. (.NOT. SameString(InArgs(18),''))  ) THEN
+                    OutArgs(16)='ReverseWithLimits'
+                  END IF
+                END IF
+
```
Myoldmopar, line 489: @wfbuhl (@mjwitte) Is this correct?  I followed your Rules file changes.  Basically 
- if F16 is "REVERSE" then we do stuff:
  - if F17 is NOT BLANK; or if F18 is NOT BLANK then:
    - Change it to ReverseWIthLimits
- Otherwise just leave it.

Note, when I test-transitioned a file, it changed `Reverse` to `ReverseWithLimits` because F17 and F18 were `AutoCalculate`.  That's good, right?

Myoldmopar:
Addresses #5847

src/Transition/CreateNewIDFUsingRulesV8_6_0.f90
wfbuhl: Yes. That's the most frequent case.

Myoldmopar:
@nmerket I think I'll get a notification when you push a commit up here, but just in case, tag me.  Or if you haven't done anything yet and want me to do it, just let me know.

nmerket:
@Myoldmopar, I added the EMPD sheet back in.

Myoldmopar:
Thanks @nmerket; done.

Issue #5728, opened 2016-06-14
High temp radiant with scalable sizing causes crash for DX coil sizing later


mjwitte:
## Issue overview

User file crashes [here](https://github.com/NREL/EnergyPlus/blob/09819874170d77a0257048ab77f435d89bcbc38e/src/EnergyPlus/ReportSizingManager.cc#L2023-L2024) in ReportSizingManager::RequestSizing because DataScalableCapSizingON = true but CurZoneEqNum = 0.

The ZoneHVAC:HighTemperatureRadiant apparently is setting DataScalableCapSizingON to true but not resetting it to false.  Seems like RequestSizing should be protecting against CurZoneEqNum in all of these scalable sizing blocks or before them.

_Workaround:_ In zone equipment such as ZoneHVAC:HighTemperatureRadiant, specify `HeatingDesignCapacity` for the Heating Design Capacity Method.  (Similar for cooling capacity if other types of zone equipment are causing the same problem.)
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (v8.5.0)
- Helpdesk ticket 11421
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (DefectFiles\5728)
- [x] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #5866

PR #5866, opened 2016-09-19
Correct Scalable Sizing Crash for Various Zone Equipment

RKStrand:
## Pull request overview

This pull request corrects a crash that happened with the scalable sizing routines.  At issue was a variable was not being reset to false consistently across all of the different equipment that use scalable sizing.  This was causing a crash when scalable sizing was being used for a high temperature radiant heater and then a DX coil was being sized.  The unexpected wrong value for a critical flag caused the wrong code to be executed and as a result a hard crash.  This was corrected for the high temperature radiant heater as well as other equipment which use this same flag.  The new code corrects the problem and the user defect input file now runs as it should.  This problem was eliminated for other equipment combinations as well.  The original problem was logged as GitHub Issue #5728.
### Work Checklist

  - Defect: This pull request repairs a github defect issue #5728.
### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types
- IDD changes n/a


mjwitte:
@rraustad Mr. Sizing Guru, do these changes look ok to you?

tst/EnergyPlus/unit/HighTempRadiantSystem.unit.cc
```diff
@@ -137,5 +141,35 @@ namespace EnergyPlus {
 
 	}
 
+	TEST_F( EnergyPlusFixture, HighTempRadiantSystemTest_SizeHighTempRadiantSystemScalableFlagSetTest )
+	{
+		int RadSysNum;
+		int SizingTypesNum;
+		
+		HighTempRadiantSystem::clear_state();
```
mjwitte: This doesn't need to be here.  I'll remove it.  It gets called from the fixture.

mjwitte (commented):
@RKStrand Based on a search of the code for `DataScalableCapSizingOn`, it looks like HVACVariableRefrigerantFlow.cc, PackagedTerminalHeatPump.cc, and WindowAC.cc are missing trailing `DataScalableCapSizingON = false` statements.

tst/EnergyPlus/unit/HighTempRadiantSystem.unit.cc
RKStrand: @mjwitte I'll look at these and add as appropriate (DataScalableCapSizingON = false).

RKStrand: @mjwitte And thanks for removing the extraneous clear_state.  Still not 100% on my understanding of those, as you can tell.

RKStrand:
@mjwitte I added the proper assignment of the flag to the other models that use the DataScalableCapSizingON flag.  So, I think that addresses your comments, Mike.

mjwitte (approved):

rraustad:
Just got back from vacation. These changes look correct to me.

PR #5867, opened 2016-09-20
EP-Launch-Lite (v2.0)

Myoldmopar:
## Pull request overview

This removes the old AppleScript based EP-Launch-Lite tool, and replaces it with a Python based tool.  The tool uses PyGTK, but is packaged up using py2app so that it is a standalone app that comes along with the E+ installer.  For some very strange reason, I couldn't get CPack to put it in the root of the install folder, so right now it is sitting inside PreProcess.  I will try to figure out how to get it in there, but if I can't figure it out in less than 20 minutes I'm just going to change the Darwin post-flight script to move it.

Special thanks to @JasonGlazer for testing this out many times over the 3 days since it was started.

Also, I couldn't get it to run Macro or ExpandObjects files due to permissions issues.  They work fine if I give the installer folder user write permission.  We already do this for the ExampleFiles folder, so I didn't feel too bad about doing this.  I'm open to suggestion on a better solution, but I don't have time to investigate too deeply.
### Work Checklist
- [x] Test it out on someone else's machine
- [x] Test HVACTemplate/Macro files (note Macro files have to be set up with `##fileprefix`, etc.


Myoldmopar:
OK, @nmerket tested this locally.  His one comment was about the open input file dialog, and I just pushed a commit ( 3255e73 ) to resolve that.  I also added French translation.  I also updated the version number.  

The only thing I couldn't resolve yet was placement of the built app bundle in the installation folder.  For now it will remain in PreProcess.  For reference though, on Macs, if you use spotlight to open the app, it will be seamless, see?...

image

I'd love to go back and have someone double check the translations, and also update them to allow th unicode characters for accents and such.  So in that sense maybe it's a _little_ rough, but fine enough to go out.  This is ready to go in.  

@nmerket would you be interested in clicking the merge button?

Myoldmopar:
(Also: opened #5871 to track any follow-up issues from this initial release)

nmerket:
Ooh, the power. Sure I can merge it.

Myoldmopar:
Thanks @nmerket I'll close the relevant Pivotal tickets and Github issues.

jasondegraw:
If only there was someone from France who already has a signed CLA...

Issue #5753, opened 2016-07-01
Meter:Custom with Zone Packaged Terminal Air Conditioner Electric Energy causes a fatal error


mjwitte:
## Issue overview

Using "Zone Packaged Terminal Air Conditioner Electric Energy" in a Meter:Custom object results in some ambiguous warnings and then ultimately a fatal error:

```
   ** Warning ** Referenced variable or meter used in the wrong context "OFFICE_101_PTAC" of type "ZONEHVAC:PACKAGEDTERMINALAIRCONDITIONER"
   ** Warning ** Referenced variable or meter used in the wrong context "OFFICE_1103_PTAC" of type "ZONEHVAC:PACKAGEDTERMINALAIRCONDITIONER"

. . .

   **  Fatal  ** GetInternalVariableValue: passed index beyond range of array.
```

The warnings come from `OutputProcessor::GetMeteredVariables` which is called from the infamous `SystemReports::CreateEnergyReportStructure`.  This leaves zeroes in the system reports data structure which then trip the fatal error later.  One solution here may be to discern between custom meters and standard meters in `SystemReports::CreateEnergyReportStructure`.
### Details

Some additional details for this issue (if relevant):
- Platform (all)
- Version of EnergyPlus (v8.5.0)
- Helpdesk ticket number 11457
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #5869

PR #5869, opened 2016-09-21
Fix fatal error in GetInternalVariableValue when non-metered variable used on a custom meter

mjwitte:
## Pull request overview

Addresses issue #5753.  If an output variable which is not normally metered is used in a custom meter (`Meter:Custom`) then it would cause warnings and in some cases a fatal error.

Three types of changes:
1. In `OutputProcessor::GetNumMeteredVariables` return number of variables that are connected only to a standard meter.  Previously connections to custom meters were also counted here, but those a not applicable for this function.
2. In `SystemReports::*` anywhere that `NumVariables = GetNumMeteredVariables` is used, check if `NumVariables` is zero and skip entire blocks.  This eliminates the "wrong context" meter warnings and prevents the fatal bad index error later on.
3. In `SystemReports:*` throughout, add references to get rid of lonnnnnng variable constructs like `PrimaryAirSystem( AirLoopNum ).Branch( BranchNum ).Comp( CompNum ).MeteredVar( VarNum )`.
### Work Checklist

### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes n/a


mjwitte:
@Myoldmopar Maybe I shouldn't have clouded this with all the local reference changes.  I can back this up to just the bug fix if you prefer.

Myoldmopar:
No, don't revert those.  They are nice.  If we had more time...maybe it would be nice to break it into two pieces, but now isn't the time for that.  I don't see any red flags with these changes.

Myoldmopar (approved):

PR #5872, opened 2016-09-21
Addition of Scalable Sizing for the Simple Cooling Panel

RKStrand:
## Pull request overview

This is sort of a defect/completion of a new feature that was added earlier.  The previous version of the model did not include autosizing.  This work upgrades the cooling panel to use the latest sizing procedure that has been implemented in all of the other radiant-convective units.
### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes n/a


Myoldmopar:
So with annual runs, there are gobs of ControlCompOutput messages in the error files of all three example files that had been added to CMakeLists previously.  I added the fourth CoolingPanel example file so it would be run also, but I haven't run it yet.

@RKStrand What are your thoughts on the ControlCompOutput issue?

PR #5875, opened 2016-09-22
Battery, storage, and other minor documentation fixes

Myoldmopar:
These documentation changes were suggested from coworkers here in both the commercial group and the battery group.


Myoldmopar:
There's really only a couple sentences with meaningful changes that tend to clarify the usage of each model for different battery types.  This was informed by comments from battery researchers based on their understanding of the models being used inside E+.

PR #5876, opened 2016-09-22
Fix memo and extensible flags on IDD object

Myoldmopar:
Fixes the IDD object for ZoneHVAC:CoolingPanel:RadiantConvective:Water


Myoldmopar:
@macumber @mjwitte Bringing this in on its own since it looks like #5872 is not ready yet.

PR #5877, opened 2016-09-22
Don't fix up executables on CentOS

macumber:
ldd does not provide output on CentOS for fixup bundle, this will only affect builds on RedHat


macumber:
@Myoldmopar

Myoldmopar:
I do not mind this going in at all, assuming it works fine when i try to
build later tonight. @kbenne do you have any issues with this?

Issue #5635, opened 2016-04-29
WaterHeater:Stratified with HorizontalCylinder tank shape uses incorrect height to check if heaters are inside the tank


mjwitte:
## Issue overview

User file with this tank input

```
WaterHeater:Stratified,
    Storage Tank,            !- Name
    Water Heater,            !- End-Use Subcategory
    0.360,                       !- Tank Volume {m3}
    0.4,                !- Tank Height {m}
    HorizontalCylinder,      !- Tank Shape
    ,                        !- Tank Perimeter {m}
    100,                     !- Maximum Temperature Limit {C}
    MasterSlave,             !- Heater Priority Control
    Standard Tariff,      !- Heater 1 Setpoint Temperature Schedule Name
    8,                       !- Heater 1 Deadband Temperature Difference {deltaC}
    3000,                       !- Heater 1 Capacity {W}
    0.535,                       !- Heater 1 Height {m}
```

throws this fatal error:

```
   ** Severe  ** WaterHeater:Stratified = STORAGE TANK: Heater 1 is located higher than overall tank height.
   **   ~~~   ** Tank Height = 0.4000
   **   ~~~   ** Heater 1 Height = 0.5350
   **  Fatal  ** Errors found in getting WaterHeater:Stratified input. Preceding condition causes termination.
```

This check needs to use the true tank height (diameter) which is 2*SQRT((0.36/0.4)/Pi)=1.07m not the input value of 0.4 which is the horizontal "height".  Quick look at the code finds two places where the true height is calculated from the input height and neither place saves this into the data structure so the variable names gets confusing WaterThermalTank.Height (input value) and TankHeight (local height dependent upon shape input).  Seems this calculation could be done once and stored in a new structure variable like ActualHeight  and WaterThermalTank.Height should be renamed to something like WaterThermalTank.InputHeightOrLength.  Then this error check could simply be changed to use .ActualHeight instead of .Height.

And perhaps change the field name to "Tank Height or Length"?
### Details

Some additional details for this issue (if relevant):
- Platform Win64
- Version of EnergyPlus 8.5.0
- Helpdesk ticket 11313
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (reproduced with an example file, attached here)
  [WaterHeaterStandAlone-Horizontal.idf.txt](https://github.com/NREL/EnergyPlus/files/242889/WaterHeaterStandAlone-Horizontal.idf.txt)
[...]

PR #5881, opened 2016-09-23
Change test tank height for heaters in stratified water heater

Myoldmopar:
This just changes which height to use in checking whether the heaters are out of range.  I didn't do a deep inspection of the whole module, but the defect file now runs to completion with a clean error file, so I think this is good to go.  Fixes: #5635

### DIFF
--- a/src/EnergyPlus/WaterThermalTanks.cc
+++ b/src/EnergyPlus/WaterThermalTanks.cc
@@ -2921,7 +2921,8 @@ namespace WaterThermalTanks {
 					WaterThermalTank( WaterThermalTankNum ).HeaterHeight1 = rNumericArgs( 7 );
 
 					//Test if Heater height is within range
-					if ( ( !WaterThermalTank( WaterThermalTankNum ).HeightWasAutoSized ) && ( WaterThermalTank( WaterThermalTankNum ).HeaterHeight1 > WaterThermalTank( WaterThermalTankNum ).Height ) ) {
+					Real64 tankHeightForTesting = 2.0 * sqrt( ( WaterThermalTank( WaterThermalTankNum ).Volume / WaterThermalTank( WaterThermalTankNum ).Height ) / DataGlobals::Pi );
+					if ( ( !WaterThermalTank( WaterThermalTankNum ).HeightWasAutoSized ) && ( WaterThermalTank( WaterThermalTankNum ).HeaterHeight1 > tankHeightForTesting ) ) {
 						ShowSevereError( cCurrentModuleObject + " = " + cAlphaArgs( 1 ) + ": Heater 1 is located higher than overall tank height." );
 						ShowContinueError( cNumericFieldNames( 2 ) + " = " + RoundSigDigits( rNumericArgs( 2 ), 4 ) );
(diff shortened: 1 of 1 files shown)


Myoldmopar:
@mjwitte I put you down to review, but if you are out of time, I can handle it.  The defect file from the issue now runs successfully though.

JasonGlazer:
@Myoldmopar When running the unit tests locally, EnergyPlusFixture.HPWHWrappedDummyNodeConfig fails as well as EnergyPlusFixture.StratifiedTankUseEnergy

Myoldmopar:
Yeah, I saw the failures come through, I'll take a look in a bit.  Can you post the failure output?

JasonGlazer:
[ RUN      ] EnergyPlusFixture.HPWHEnergyBalance
........\tst\EnergyPlus\unit\WaterThermalTanks.unit.cc(734): error: Value of: WaterThermalTanks::GetWaterThermalTankInput()
  Actual: true
Expected: false
[  FAILED  ] EnergyPlusFixture.HPWHEnergyBalance (13131 ms)

[ RUN      ] EnergyPlusFixture.StratifiedTankUseEnergy
........\tst\EnergyPlus\unit\WaterThermalTanks.unit.cc(1592): error: Value of: WaterThermalTanks::GetWaterThermalTankInputData( ErrorsFound )
  Actual: true
Expected: false
[  FAILED  ] EnergyPlusFixture.StratifiedTankUseEnergy (8177 ms)

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -2921,7 +2921,8 @@ namespace WaterThermalTanks {
 					WaterThermalTank( WaterThermalTankNum ).HeaterHeight1 = rNumericArgs( 7 );
 
 					//Test if Heater height is within range
-					if ( ( !WaterThermalTank( WaterThermalTankNum ).HeightWasAutoSized ) && ( WaterThermalTank( WaterThermalTankNum ).HeaterHeight1 > WaterThermalTank( WaterThermalTankNum ).Height ) ) {
+					Real64 tankHeightForTesting = 2.0 * sqrt( ( WaterThermalTank( WaterThermalTankNum ).Volume / WaterThermalTank( WaterThermalTankNum ).Height ) / DataGlobals::Pi );
```
mjwitte: @Myoldmopar @JasonGlazer I think tankHeightForTesting needs to be sensitive to the Tank Shape type.

Myoldmopar: That makes _all_ the sense.  #dumb

I'll make that change shortly.

JasonGlazer:
Also, the code seems to be assuming a horizontal cylinder but looks like it would run for a vertical or horizontal orientation.

Myoldmopar:
I think 222ee92 should take care of this now.  Unless I'm missing something from @JasonGlazer 's last comment.

mjwitte:
@Myoldmopar Code changes look correct to me.  I don't have time to test at the moment, but can tomorrow.

JasonGlazer:
@Myoldmopar and @mjwitte the code change looks good to me too and the CI tests are clean.

PR #5879, opened 2016-09-23
Parallel fan-powered induction unit zone fan control improvements

mjwitte:
## Pull request overview

Initial phase of control improvements for `AirTerminal:SingleDuct:ParallelPIU:Reheat` zone fan control. Begins to address #4989.

**If Fan On Flow Fraction > 0 (Existing controls)**
_When the zone fan schedule is on:_ The zone fan will operate whenever the primary airflow is <= the Fan On Flow Fraction.  Even if the central fan is off, the zone fan will operate if the zone fan schedule is on.

_When the zone fan schedule is off:_ The zone fan will only operate if an availability manager cycles on the central fan or zone fans.

**If Fan On Flow Fraction = 0 (New control option)**
_When the zone fan schedule is on:_ The zone fan will only operate if there is a heating load, if reheat is required, or if an availability manager cycles on zone fans only.

_When the zone fan schedule is off:_ The zone fan will operate if an availability manager cycles on zone fans regardless of load. If an availability manager cycles on the central fan, then the zone fan will only operate if there is a heating load or reheat is required.
### Work Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
- [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
  - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
  - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
  - NewFeature: This pull request includes code to add a new feature to EnergyPlus
  - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
  - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog
- [x] Doc changes
### Review Checklist

This will not be exhaustively relevant to every PR.
- [ ] Code style (parentheses padding, variable names)
- [ ] Functional code review (it has to work!)
- [ ] If defect, results of running current develop vs this branch should exhibit the fix
- [ ] CI status: all green or justified
- [ ] Performance: CI Linux results include performance check -- verify this
- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
  - [ ] If any virtual classes, ensure virtual destructor included, other things
- IDD changes:
  - [ ] Verify naming conventions and styles, memos and notes and defaults
  - [ ] Open windows IDF Editor with modified IDD to check for errors
[...]

Issue #4989, opened 2015-06-25
Parallel fan powered terminal unit fan controls need option to cycle with thermostat


rongpengzhang:
Currently, the input is for a minimum damper position at or below which the zone fan turns on. This leaves the fan on during deadband hours and potentially always during nighttime hours when the damper is closed! In most real systems and in DOE2, the fan is controlled thermostatically so that it can be used as the first stage of heating, before the reheat coil is energized. So that would require a separate terminal fan Tstat schedule or maybe just an input for offset from the heating Tstat.

mjwitte:
Also, allow Maximum Secondary Air Flow Rate to be specified as a fraction of the primary flow rate.

mjwitte:
User reports that the changes in #5879 still have some issues:

Made the following changes to use the new Night Cycle Availability Manager.
1.      Using the “CycleOnAnyCoolingOrHeatingZone” control type option.
2.      Put all zones in to the zone list “CoolingZones”.
3.      Put VAV reheat terminal zones to the zone list “Heating Zones”
4.      Put all PIU zones into the “Heating Fan Only Zones”

Testing result shows that the zone lists in Night Cycle Availability Manager work:
  a.      Keeping central AHU fan off during unoccupied period when any of the zones on the “heating fan only” Zones call for heating.
  b.      Activate central AHU fan when any zone on the “Cooling Zones” call for cooling during unoccupied period
  c.      Activate central AHU fan when any zone on the “Heating Zones” call for heating during unoccupied period

There are still two issues remaining:
1.      Parallel PIU fan still seem to be running during some cooling operation and deadband operation periods (when the primary air flow ratio is low). Ideally we would like the fan to be operating only when zone thermostat calls for heating.
2.      During unoccupied period, it seems like if any of the PIU zones call for heating, PIU fans for all zones cycle on. What we expect is only PIU fan in the zones that call for heating cycles on, others remain off.

_Defect file added to EnergyPlusDevSupport/DefectFiles._

mjwitte:
Issue 2 above can be seen in the defect file output for time  01/06  04:40:00, Space 2-1 falls below setpoint and then in the next timestep all of the zone fans turn on.  The other zones have baseboard heat which keeps them at setpoint.  Space 2-1 has undersized baseboard.  For this to work, however, the baseboard needs to run first for heating, otherwise the PIU will think it should be serving the heating load.  Second flavor of defect file added.

_Update_
[...]

PR #5918, opened 2016-10-26
Fix parallel fan-powered induction unit zone fan controls

mjwitte:
## Pull request overview

Completes work to address #4989.  Initial work was started in #5879 which added more control options to  `AirTerminal:SingleDuct:ParallelPIU:Reheat`.

This pull requests fixes some logic problems.
### Work Checklist

### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:
@mjwitte @lgentile it has been 26 days since this pull request was last updated.

Myoldmopar:
@mjwitte Do the original 4989 test files demonstrate the logic issues when run with the develop branch?  I have this branch built, and I'm running the defect files now, and also looking at extracting out a unit test...

mjwitte:
Probably not.  I have other feedback that I need to incorporate - so this one should wait.

nrel-bot-3:

nrel-bot-3:

mjwitte:
@Myoldmopar This is ready for review.  The original issue that prompted these changes was an input error, so you'll see no results changes when you run the defect files.  However, the new logic is more robust and avoids the risk of crosstalk between zones.  The old logic touched `DataHVACGlobals::TurnFansOn` in the PIU calc routine.  The new logic uses local fan flags instead to avoid flipping the global flag set by the availability manager.

mjwitte:
@rraustad Please edit the description at the top here, and I don't see a pivotal story for this one.

mjwitte:
Whoops - that comment was meant for a different PR.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1588,8 +1573,8 @@ namespace PoweredInductionUnits {
 		Real64 SecAirMassFlow; // secondary air mass flow rate [kg/s]
 		Real64 CpAirZn; // zone air specific heat [J/kg-C]
 		Real64 FanDeltaTemp( 0.0 ); // fan temperature rise [C]
-		//unusedREAL(r64)    :: MaxSteamFlow
-		//unusedREAL(r64)    :: MinSteamFlow
+		bool PIUTurnFansOn( false ); // If True, overrides fan schedule and cycles PIU fan on
+		bool PIUTurnFansOff( false ); // If True, overrides fan schedule and PIUTurnFansOn and cycles PIU fan off
```
Myoldmopar: These could be defined below when they are initialized, but no big deal.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1608,6 +1593,9 @@ namespace PoweredInductionUnits {
 		QZnReq = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputRequired;
 		QToHeatSetPt = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputReqToHeatSP;
 		CpAirZn = PsyCpAirFnWTdb( Node( ZoneNode ).HumRat, Node( ZoneNode ).Temp );
+		// Initialize local fan flags to global system flags
+		PIUTurnFansOn = ( DataHVACGlobals::TurnFansOn || DataHVACGlobals::TurnZoneFansOnlyOn );
+		PIUTurnFansOff = DataHVACGlobals::TurnFansOff;
```
Myoldmopar: Doesn't seem like this local is currently really being used anywhere here.  But I don't think it causes an issue, and it does set a nice pattern along with the local `On` variable.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1637,9 +1625,14 @@ namespace PoweredInductionUnits {
 				// no primary air flow
 				PriAirMassFlow = 0.0;
 				// PIU fan off if FanOnFlowFrac is 0.0 and there is no heating load, also reset fan flag if fan should be off
-				if ( ( QZnReq <= SmallLoad ) && ( PIU( PIUNum ).FanOnFlowFrac <= 0.0 ) ) {
-					SecAirMassFlow = 0.0;
-					DataHVACGlobals::TurnFansOn = DataHVACGlobals::TurnZoneFansOnlyOn;
+				if ( PIU( PIUNum ).FanOnFlowFrac <= 0.0 ) {
+					if ( QZnReq <= SmallLoad ) {
```
Myoldmopar, line 1627: 👍

tst/EnergyPlus/unit/PoweredInductionUnits.unit.cc
```diff
@@ -0,0 +1,264 @@
...
+	PoweredInductionUnits::InitPIU( SysNum, FirstHVACIteration ); // Run thru init once with FirstHVACIteration set to true
+	Fans::InitFan( 1, FirstHVACIteration );
+	DataGlobals::BeginEnvrnFlag = false;
+	FirstHVACIteration = false;
+
+	// Note that the fan schedule is always off, so the PIU fan should only run if the night cycle turn on flag is true
+
+	// First test - Heating load, TurnZoneFansOn is false - expecting no flow
+	DataZoneEnergyDemands::ZoneSysEnergyDemand(1).RemainingOutputRequired = 2000.0; // Heating load - expect min flow rate
+	DataZoneEnergyDemands::CurDeadBandOrSetback(1) = false;
```
Myoldmopar: You don't adjust this value throughout all the unit tests here.  Was there a reason you copied this line into each block below?  I just want to make sure you weren't intending on also adding another set with ` = true;`.

Myoldmopar (approved):
Changes are good.  I had a couple comments and one question in the unit tests, but this is good.

src/EnergyPlus/PoweredInductionUnits.cc
mjwitte: Yeah, the night cycle manager doesn't use TurnFansOff, but other availability managers might.  So, for completeness it seemed like the right thing to do to include that and pass the pair of flags, instead of just passing PIUTurnFansOn to the fan simulation.  I could probably remove it and not see any changes.  I need to look more closely at how the fan simulation uses these.

tst/EnergyPlus/unit/PoweredInductionUnits.unit.cc
mjwitte: No, just happened to copy it along.  But I could add a case(s) that sets it to true.

Myoldmopar: I'm all for more tests if they are relevant to this fix and don't add a major level of effort to whip up, but I won't hold this up.  I was just making sure you hadn't accidentally forgotten to switch some of them on.

src/EnergyPlus/PoweredInductionUnits.cc
Myoldmopar: Do you need to look more closely at the fans before this goes in then?  Or just to look at for later?

mjwitte: Oh, that's it, these two optional parameters are assumed to get used as a pair, in [Fans::SimulateFanComponents](https://github.com/NREL/EnergyPlus/blob/8971859c03f8c5d8e350d46877364a5a6c60d1e4/src/EnergyPlus/Fans.cc#L271).
`		if ( present( ZoneCompTurnFansOn ) && present( ZoneCompTurnFansOff ) ) {`

tst/EnergyPlus/unit/PoweredInductionUnits.unit.cc
mjwitte: Adding some more now.

mjwitte:
Moved the bools and added some more test cases.  One with deadband true and several with primary airflow > 0.

Issue #5878, opened 2016-09-22
Processing of non-rectangular window surfaces


rongpengzhang:
## Issue overview

Current EnergyPlus prohibits building models with 4-sided but non-rectangular window or glassdoor surfaces. Such surfaces lead to severe errors in the surface pre-processing and the termination of the simulation.

```
   ** Severe  ** ProcessSurfaceVertices: Suspected 4-sided but non-rectangular Window, Door or GlassDoor:
   **   ~~~   ** Surface=WF-1
   **  Fatal  ** Errors found in Building Input, Program Stopped
```

This strict restriction is related with the calculation of Nusselt number  at window air gaps. We find that this restriction can be released by transforming the surface into an equivalent rectangular surface with the same area and aspect ratio. This approach is similar to the approach implemented in the current EnergyPlus to handle triangular windows. The current severe error caused by the detection of non-rectangular window will be replaced by a warning message before the shape transformation.
### Details

Some additional details for this issue (if relevant):
- Platform Win64
- Version of EnergyPlus 8.5.0
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (list location of defect file here)
- [x] Ticket added to Pivotal for defect (development team task)
- [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

[Uploading 5ZoneAirCooled_with non-rectangular window.zip…]()

mjwitte:
Closed via #5920

PR #5920, opened 2016-10-27
Processing of non-rectangular window surfaces

rongpengzhang:
## Pull request overview

This is to fix Issue #5878. Current EnergyPlus prohibits building models with 4-sided but non-rectangular window or glassdoor surfaces. Such surfaces lead to severe errors in the surface pre-processing and the termination of the simulation. This strict restriction is related with the calculation of Nusselt number at window air gaps. We find that this restriction can be released by transforming the surface into an equivalent rectangular surface with the same area and aspect ratio. This approach is similar to the approach implemented in the current EnergyPlus to handle triangular windows. The current severe error caused by the detection of non-rectangular window will be replaced by a warning message before the shape transformation.
### Work Checklist

### Review Checklist

- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types
- IDD changes n/a


nrel-bot-3:
@rongpengzhang @lgentile it has been 14 days since this pull request was last updated.

mjwitte (changes requested):
This should have a unit test.

I would expect that this needs some documentation changes - I/O Ref and Engineering Ref.

Do we really want this to always display a warning?  If the method is adequate, then perhaps it should be silent (same as the triangular window currently is, if I am following correctly).  Or this could warn only with DisplayExtraWarnings.

rongpengzhang:
@mjwitte Thanks for the comments. I have updated the codes and IO reference based on your advice. I have also prepared a unit test for the rectangular, trapezoid, and parallelogram windows.

mjwitte (approved):

PR #5921, opened 2016-10-28
Allow Multiple Air Loops to One Thermal Zone

mjwitte:
## Pull request overview

This new feature removes the limitation of one airloop (`AirLoopHVAC`) system per zone, and adds options for user-control of sizing and return air flow allocation.
It also allows an airloop with no return path.

### Work Checklist

- [x] New feature proposal
- [x] Design document
- [ ] Source code
- [ ] Example files
- [ ] Unit tests
- [ ] Documentation
- [ ] Deal with furnaces searching for any ADU/DirectAir to find out the ZoneSequenceCoolingNum etc. - happens in Furnaces and HVACMultiSpeedHeatPump
- [ ] Review whether optimum start avail managers really works with multiple airloops in the sames zone - if more than one is opt start then it may be goofy
- [ ] Optimal start could be refactored to loop over the zones on the airloop instead of all zones for every decision
- [ ] Sort out NoHeatToReturnAir node by node instead of zone by zone
- [ ] Get rid of all crlzonnum vs actualzonenum - they're the same now, right?
- [ ] Sort out zone loads in SystemReports::ReportSystemEnergyUse
- [ ] Sort out unbalanced exhaust warnings - currently by zone, not by airloop in ZoneEquipmentManager::SimZoneEquipment
- [ ] Sort out peak system load in OutputReportTabular::WriteComponentLoadSummaryTables
### Review Checklist

- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
- IDD changes:
- [ ] ExpandObjects changes?


design/FY2017/MultipleAirloopsInOneZone.md
```diff
@@ -0,0 +1,213 @@
+Allow Multiple Air Loops to One Thermal Zone
```
hongtz68, line 1: Mike,

This is long overdue :) Thanks for your effort on this new feature.

I have several questions:
1. Do we allow multiple outdoor air loops (DOAS systems) to a zone?
2. How do we control the operation of multiple air loops? For example, if we have two air loops to a zone with each sized to meet half of the zone peak loads, at 50% zone load, do we operate only one air loop at 100%, or two with each at 50%? We may need a control scheme for this to make it more flexible. 
3. How to avoid fight of different air loops (e..g, some doing cooling/dehumidifying while others doing heating/humidifying)? We saw this in data centers with multiple RTUs.
4. Any high level report to summarize how various air loops operate in a zone? E.g., what loads they meet, how many hours each loop operate?

Tianzhen

mjwitte: 1.  I don't see any reason this would not work, but the user would need to specify the OA requirements to avoid over-ventilating the zone.
2.  That's actually a second project that is also funded for the next release "Allow Multiple/Partial HVAC in One Zone" similar, but focused on the control aspect for all equipment serving a zone, including ZoneHVAC equipment.
3.  That's a good question - at this point I was assuming all of the system would work off the same thermostat, but that may not be adequate in the long run (this kind of fits into the part two project as well).
4.  That would be useful - will think about what that might contain.

larryscheier:
Excellent start. Some good comments/questions from Tianzhen.

The proposed DesignSpecification:AirTerminal:Sizing object contains many fields that currently exist on some AirTerminal objects, but not others.

1) Should users expect the new fields to be used solely for sizing the AT or will some also have an effect on the HVAC simulation? For example, for VAV AT's, will the "Design Spec OA Object Name" field on the new DesignSpecification:AirTerminal:Sizing object have an effect on the HVAC simulation of minimum airflow as is now done for the AT:SingleDuct:VAV:NoReheat and AT:SingleDuct:VAV:Reheat objects?

2) It would be nice to have a table of the new fields on the new DesignSpecification:AirTerminal:Sizing object vs each AirTerminal object, i.e., which new DsnSpec:AT fields match functionality of existing AirTerminal fields, which new DsnSpec:AT fields do not apply to existing AirTerminals fields, which new DsnSpec:AT fields provide new functionality, etc.

mjwitte:
@larryscheier 

> Excellent start. Some good comments/questions from Tianzhen.
> The proposed DesignSpecification:AirTerminal:Sizing object contains many fields that currently exist on some AirTerminal objects, but not others.
> 
> 1) Should users expect the new fields to be used solely for sizing the AT or will some also have an effect on the HVAC simulation? For example, for VAV AT's, will the "Design Spec OA Object Name" field on the new DesignSpecification:AirTerminal:Sizing object have an effect on the HVAC simulation of minimum airflow as is now done for the AT:SingleDuct:VAV:NoReheat and AT:SingleDuct:VAV:Reheat objects?

Good point - I didn't think about those overlaps.  The intent of the proposal was that the new sizing object would work the same way that `Sizing:Zone` currently works - just for sizing.  Maybe a better solution is to allow generic `Sizing:Zone` objects that aren't linked to a specific zone?  Or to shorten the proposed new object to be exactly like `Sizing:Zone`.

> 2) It would be nice to have a table of the new fields on the new DesignSpecification:AirTerminal:Sizing object vs each AirTerminal object, i.e., which new DsnSpec:AT fields match functionality of existing AirTerminal fields, which new DsnSpec:AT fields do not apply to existing AirTerminals fields, which new DsnSpec:AT fields provide new functionality, etc.

This would be good in general for the existing sizing and terminal unit objects.  I'll work on this table and that may help guide this proposal or at least provide some discussion points.

lgu1234:
@mjwitte You have an option: ### 6. _Optional_ - Allow an airloop with no return path. How do you handle mass balance without return path? Do you plan to use other objects?

mjwitte:
@lgu1234 The zone air balance would work as it currently does, except there wouldn't be a return node for that air loop.  The zone air mass balance should work the same as it currently does.  If there is more supply than exhaust (plus return to another airloop if present) then the excess air is assumed to exfiltrate and doesn't impact the zone air balance.  Of if the ZoneAirMassBalance option is active then that would take over and balance by adjusting infiltration and/or mixing.

For the airloop itself without return, it would need to take in outdoor air to match the supply and complain if the OA controller didn't allow that.

shorowit:
Allowing an airloop with no return path has been identified as a high priority for modeling airflow in residential buildings. If the funding allows for this, we would definitely like to see it included in the work.

Myoldmopar:
> Rather than develop more workaround components, it is time to remove the restriction of one airloop per zone.

This line wins @mjwitte 130 internet points today.

I don't have any specific concerns moving forward.  I just look forward to see plenty of demos and testing results showing the edge cases working well.  I think it would be really great to work up some unit tests right from the start......

mjwitte:
@Myoldmopar Feel free to write and push those unit tests to this branch.  How many points get me a free pizza?

Myoldmopar:
```
TEST_F(ZoneAirLoopTestBase, TwoAirLoopTest_Case1)
 airLoop1 = AirLoop.new(m)
 airLoop2 = AirLoop.new(m)
 zone = Zone.new(m)
 airLoop1.attachToZone(zone)
 airLoop2.attachToZone(zone)
 zone.simulateHVAC() 
 EXPECT_EQ(zone.drybulb, expected_db)
 EXPECT_EQ(airLoop1.sensibleCoolingProvidedToZone(zone), expected_sen_clg_airloop1)
```

---

_Edwin Wakes Up_

---

Oh, sorry, my worlds collided a little bit there...

Myoldmopar:
> How many points get me a free pizza?

42*42

nrel-bot-3:

design/FY2017/MultipleAirloopsInOneZone.md
```diff
@@ -0,0 +1,546 @@
...
+
+This object modifies the sizing of a given terminal unit given the base sizing results from the corresponding `Sizing:Zone` inputs. Any given `DesignSpecification:AirTerminal:Sizing` object is generic and may be used by multiple terminal units with similar characteristics. 
+
+
+*Name*
+
+Name of the design specification air terminal sizing object. This name may be reference by any `AirTerminal:*` object.
+
+*Fraction of Design Cooling Load*
+
```
rraustad: Currently a "sensible" load

rraustad: Currently a "sensible" load

design/FY2017/MultipleAirloopsInOneZone.md
```diff
@@ -0,0 +1,546 @@
...
+In struct `EquipConfiguration`
+
+ - Delete `int ReturnAirNode`
+ - Add `int NumReturnNodes`
+ - Add `Array1D_int ReturnNode`
+ - Add to IDD and input processing for Return Air Node or Nodelist
+
+ - Delete `int AirLoopNum`
+ - Add `int NumAirLoops` (this may not always equal `NumReturnNodes` if an airloop has no return path)
+ - Add `Array1D_int AirLoopPointer`
```
rraustad: Would be nice if these lined up (i.e., each has same size as NumAirLoops). Remember the confusion with zone indexes for conditioned vs non-conditioned zones (if I recall correctly). This way, AirLoopPointer(6) always uses ReturnNode(6).

mjwitte: @rraustad Valid point.  But NumReturnNodes will get set when ZoneHVAC:EquipmentConfguration input is processed, and at that point, we haven't figured out which airloops are serving the zone yet.  But I could make sure these align when the AirLoopPointers get set later on.  Will think about that.

design/FY2017/MultipleAirloopsInOneZone.md
```diff
@@ -0,0 +1,546 @@
...
+  - Step 4 - Allocate unbalanced exhaust air flows to each air loop (*modified step*)
+      - For a given zone, the unbalanced exhaust air flow is distributed to any air loop that has an outdoor air system in proportion to the loop's supply flow to that zone. Air systems without an outdoor air inlet are excluded.
+  - Step 5 - Balance each air loop (*existing step*)
+      - LoopReturn0 = Sum of Return Node flow rates on the loop after Steps 1-4.
+      - LoopReturn = LoopSupply - LoopExhaust
+      - For each return node on the loop AdjustedReturnFlow = ReturnFlow*LoopReturn/LoopReturn0
+
+### 5. Allocate return air heat gains to specific return air nodes
+This applies to return air heat gain from lights, refrigerated case under-case return,and airflow windows.
+
```
EnergyArchmage, line 132: Does this allocation of return air heat gains need more user input? Something like a return air node name in the Lights object?  or allocate based on relative mass flow rates?

mjwitte: Yes, a new (optional) field is proposed [further down](https://github.com/NREL/EnergyPlus/pull/5921/files#diff-9bc0a8d4cebbce9ddf4c43d508ca2b2cR310) for Lights, refrigerated case, and airflow window property.

design/FY2017/MultipleAirloopsInOneZone.md
```diff
@@ -0,0 +1,546 @@
...
+- `UpdateSysSizing` uses `TermUnitSizing`, `ZoneSizing` and `FinalZoneSizing` as inputs.
+- `UpdateSysSizing` is basically four functions in one, split up by a big case statement for `CallIndicator` = BeginDay, DuringDay, EndDay, or EndSysSizingCalc (tempted to split that into four separate functions).
+- Lots of sizing calcs are done here, and the most pertinent section is that it makes final adjustments to `TermUnitFinalZoneSizing`.
+
+#### Proposed Changes ####
+- Add new struct `DataSizing::AirTerminalSizingData` to hold input for the new `DesignSpecification:AirTerminal:Sizing` object. (Similar to ZoneHVACSizingData)
+- The new array will be called `AirTerminalSizingSpecifications`.
+- Add new function `SizingManager::GetAirTerminalSizing` to process input for the new ` DesignSpecification:AirTerminal:Sizing` object.
+
+- Change `TermUnitSizing` and `TermUnitFinalZoneSizing` to be indexed by air distribution unit (ADU) number instead of by zone. Direct air terminal units will be added to the end of the ADU list for this indexing.
```
rraustad: Documenting thought...
This seems messy. If you can pre-calculate fractional values for these indexed arrays (using air node fraction * zone level value), then when the air loop sums the specific flows it should also be possible to use air node fraction * zone level value. During sizing you only know CurZoneEqNum (in RequestSizing) but would have to know which terminal unit is sizing so you could still use air node fraction * zone level value in the sizing routines (or pass the correct air volume flow rate in a Data* var). It would be easier to start with this method and index the arrays later if you run into problems (fairly easy, yet tedious, to index these arrays).

mjwitte: @rraustad I just don't want to be reapplying the factor multiple times in different places.  So, will revisit this and consider which approach is cleanest (and safest).

nrel-bot-3:

nrel-bot-2:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

mjwitte:
This has been superceded by #6248.

Issue #5272, opened 2015-10-19
Constructions and materials tables differ between EIO and SQL


macumber:
I have not seen different values but there are different fields included in the Construction and Material summaries in the EIO and in the Constructions and Materials tables in the SQL output.  Also there is no WindowConstruction table in sql.

JasonGlazer:
@macumber I am starting to review the differences

## Material Details

| Field | In both | In EIO only | In SQLite only |
| --- | --- | --- | --- |
| Material Name | x |  |  |
| ThermalResistance {m2-K/w} | x |  |  |
| Roughness | x |  |  |
| Thickness {m} | x |  |  |
| Conductivity {w/m-K} | x |  |  |
| Density {kg/m3} | x |  |  |
| Specific Heat {J/kg-K} | x |  |  |
| Absorptance:Thermal |  | x |  |
| Absorptance:Solar |  | x |  |
| Absorptance:Visible |  | x |  |
| Material Type |  |  | x |
| ThermGradCoef |  |  | x |
| VaporDiff |  |  | x |

JasonGlazer:
## Construction CTF

| Field | In both | In EIO only | In SQL only |
| --- | --- | --- | --- |
| Construction Name | x |  |  |
| Index |  | x |  |
| Number Layers | x |  |  |
| Number CTFs |  | x |  |
| Time Step {hours} |  | x |  |
| ThermalConductance(UValue in SQL) {w/m2-K} | x |  |  |
| OuterThermalAbsorptance | x |  |  |
| InnerThermalAbsorptance | x |  |  |
| OuterSolarAbsorptance | x |  |  |
| InnerSolarAbsorptance | x |  |  |
| Roughness | x |  |  |
| InsideAbsorpVis |  |  | x |
| OutsideAbsorpVis |  |  | x |

## WindowConstruction

These appear in Construction Table in SQLite

| Field | In both | In EIO only | In SQL only |
| --- | --- | --- | --- |
| Construction Name | x |  |  |
| Index |  | x |  |
| Number Layers | x |  |  |
| Roughness | x |  |  |
| Conductance {W/m2-K} (Uvalue in SQL) | x |  |  |
| SHGC |  | x |  |
| Solar Transmittance at Normal Incidence |  | x |  |
| Visible Transmittance at Normal Incidence |  | x |  |

## WindowMaterial:Glazing

| Field | In both | In EIO only | In SQL only |
| --- | --- | --- | --- |
[...]

PR #5928, opened 2016-11-02
Add EIO tables to new Tabular summary report

JasonGlazer:
In process of addressing #5272 decided to create a tabular report (usually a HTML file) that contains all the content of the EIO file. This would allow the SQL file to contain all the data from the EIO file. 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
@mjwitte and @Myoldmopar As I am working on the new "Initialization Summary" report that pulls the content of the EIO file into the tabular reports, I realize that the EIO is not quite consistently formatted. How important is leaving the EIO file alone? For example: 
- Some data lines have what I will be using as a row header such as Surface Convection Parameters which has the surface name as the first field and I can use as a row header but others output in the EIO does not have anything useful for a row header such as Shading Summary that just has three numbers. How bad would it be to add the building name to each of these rows in the EIO so that the Initialization Summary report has a row header that makes sense?
- Some data lines could really be reorganized slightly to make better row headings like Component Sizing Information which the Component Type (AirTerminal:SingleDuct:VAV:Reheat) is the first data field even though the second data field is Component Name (SPACE1-1 VAV REHEAT). It would make much more sense to reverse these so that the row heading is the component name.
- The header rows are shortcuts for some rows such as the line:
  `! <HeatTransfer/Shading/Frame/Divider_Surface>,...`
  which is a shortcut for rows called HeatTransfer_Surface, Shading_Surfaces, etc.. and I would prefer if each one was explicitly shown rather than combined.

I am not suggesting changing any data that appears in the EIO file but just the layout of specific description and data rows to make them more consist. Consistency will allow the straight forward routine that I am writing to just create the Initialization Summary report in the tabular file automatically which is what I would prefer to programming lots of exceptions.

So how much of an issue would adding fields and rearranging some fields in the EIO file?

mjwitte:
I don't really know if any tools are parsing the eio.  I would ask some of the major players directly.  In theory, as long as the changes are documented in the output rules, this should be ok.  Consistency would be good.

Myoldmopar:
I ran into those types of issues when I had an eio parsing script once.  That, and I believe the CTF coefficient reports, are both goofy.  Cleaning them up can only do good.  I agree with @mjwitte that it wouldn't _hurt_ to ask some relevant players, but in the end, this will be very helpful, and documenting it in a change log should be sufficient.

JasonGlazer:
Here is a [link ](https://drive.google.com/open?id=0Bw2rC5uvcT-kQUY1M1V2Sm8tYWc)to the current state of the HTML and SQL output files. You probably need to download them to view them. I still have some clean up to do and support unit conversions. @macumber could you take a look at the SQL and see if you can get to the tables you originally were interested in?

AndyTindale:
DesignBuilder does parse the EIO output file so if the format changes it will involve extra work for us, but as long as the changes are documented we would be happy to see the format improve and for the data to be included in the HTML output file.

JasonGlazer:
I spot checked the EIO diffs and they are as expected.

The ERR diffs are showing stuff that I will try to fix: 
- 5ZoneAutoDXVAV.idf shows [Internal] and [m3/s-m2] conversion factors missing
- PurchAirTables.idf, PurchAirTables_wAnnual.idf, and regression._SmallOffice_Dulles.idf  shows [Interal] conversion factors missing

JasonGlazer:
The EIO diffs are as expected. The one ERR diff is actually removing a warning related to unit conversions. I think this is ready for review.

nrel-bot-3:

nrel-bot-3:

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10785,6 +10802,166 @@ namespace OutputReportTabular {
...
 	}
 
+    // Parses the contents of the EIO (initializations) file and creates subtables for each type of record in the tabular output files
+	// Glazer - November 2016
+	void
+	WriteEioTables() {
+
+		if (displayEioSummary)
+		{
+			// all arrays are in the format: (row, column)
```
Myoldmopar: I suppose you mean all 2D arrays?

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10785,6 +10802,166 @@ namespace OutputReportTabular {
...
+			Array1D_int columnWidth;
+			Array1D_string rowHead;
+			Array2D_string tableBody;
+			Array1D_int colUnitConv;
+
+			// setting up  report header 
+			WriteReportHeaders("Initialization Summary", "Entire Facility", isAverage);
+
+			// since the EIO initilization file is open at this point must close it to read it and then reopen afterward.
+			gio::close(OutputFileInits); 		// for testing only
```
Myoldmopar: for testing only?

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10785,6 +10802,166 @@ namespace OutputReportTabular {
...
+
+			std::ifstream eioFile;
```
Myoldmopar, line 10811: Hmm.  I didn't quite realize this was the approach.  Reading output files that E+ just wrote.  In an ideal world, it would've been great to create a very lightweight class, and when the eio is written by a component, it first populates an instance of the class, then writes the eio.  Then anyone who wants to do anything with the eio data could just grab from that same in-memory data, instead of having to re-read the eio.  Even with two simplifications (instantiate the class globally so you can access it from anywhere, and have it pretty much just store an array of strings, one per line), it would be very straightforward to implement and just much better than reading back the eio data.  What would you think about making this ticket a higher complexity, but pursuing something like this?

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10785,6 +10802,166 @@ namespace OutputReportTabular {
...
+			// reopen the EIO initilization file and position it at the end of the file so that additional writes continue to be added at the end.
+			int write_stat;
+			{ IOFlags flags; flags.ACTION("write"); flags.STATUS("UNKNOWN"); flags.POSITION("APPEND"); gio::open(OutputFileInits, DataStringGlobals::outputEioFileName, flags); write_stat = flags.ios(); }
+			// as of Oct 2016 only the <Program Control Information:Threads/Parallel Sims> section is written after this point
+		}
+	}
+
+	// changes the heading that contains and SI to IP as well as providing the unit conversion index
+	// Glazer Nov 2016
+    int
```
Myoldmopar: Indentation problem?

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -14180,6 +14357,37 @@ Label900: ;
 		return StringOut;
 	}
 
+	// adapted from http://stackoverflow.com/questions/4654636/how-to-determine-if-a-string-is-a-number-with-c
```
Myoldmopar: `stof()`?  with a try block...

```
#include <iostream>
#include <string>
#include <exception>
int main()
{
  std::string str;
  std::cout << "Give me a string:";
  getline (std::cin, str);
  try {
      std::stof(str);
      std::cout << "Yep, it's a number!\n";
  } catch (std::exception & e) {
      std::cout << "Not a number!\n";
  }
}
```

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -14180,6 +14357,37 @@ Label900: ;
...
+	bool 
+	isNumber(std::string s)
+	{
+		std::string trimmed = stripped(s);
+		return(strspn(trimmed.c_str(), "-.0123456789eE") == trimmed.size());
+	}
+
+	// return the number of digits after the decimal point
+	// Glazer - November 2016
+	int
```
Myoldmopar, line 14347: So I was starting to draft a more efficient way to find the number of digits after the decimal, and then I realized this may just be an artifact of having to read the eio *back in*, so if we cleaned that process up so we were just storing the data in memory and writing it out directly without having to read the eio, this is probably all gone.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -14828,6 +15043,8 @@ Label900: ;
 				stringOutWithIP = stringInWithSI.substr( 0, posLBrac + 1 ) + UnitConv( selectedConv ).ipName + stringInWithSI.substr( posRBrac );
 			} else if ( modeInString == misParen ) {
 				stringOutWithIP = stringInWithSI.substr( 0, posLParen + 1 ) + UnitConv( selectedConv ).ipName + stringInWithSI.substr( posRParen );
+			} else if (modeInString == misBrce) {
```
Myoldmopar: Parentheses padding

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -774,6 +775,15 @@ namespace OutputReportTabular {
 	WriteSurfaceShadowing();
 
 	void
+	WriteEioTables();
+
+	int
+	unitsFromHeading(std::string & heading);
+
+	std::vector<std::string>
```
Myoldmopar: `std::vector` +1

src/Transition/OutputRulesFiles/OutputChanges8-6-0-to-8-7-0.md
```diff
@@ -0,0 +1,51 @@
+Output Changes
```
Myoldmopar, line 1: Thanks for adding this file

Myoldmopar (changes requested):
I very much like the cleaning work that's going on in here, and having the report in the html.  As I noted in the review comments, I'd like to hear your thoughts on pursuing an approach where we don't have to re-open the eio to read it, but instead just store the eio data inside E+ and query that data.  I believe it would clean things up a lot.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: @Myoldmopar I actually spent a few days trying to figure out an approach that would work more like what you suggest.  There are 822 lines that write to the EIO file so I wanted to make a change that tried to keep the write lines as similar as possible to how they are now but instead of writing directly would call a function/method. Basically the same approach as you are suggesting. I ran into a few problems. The heavy use of Fortran style Formats for almost every line means that I would need to overload the gio::write and << operator. I tried to work on this but did not get far. I also tried another approach using stringstream figuring I could just leverage the Fortran style Format handling from the objexx library but did not get vary far. I started planning some meta code with Python that would get rid of all the Fortran Formats and replace them with normal C++ output manipulators. I also a little exploration of how C++ functions could accept any number of arguments of any types and that seemed impossible. It is a tough nut to crack.

Myoldmopar: Frack.  OK, I see now.  First question: How did you come up with the 822?  I get 724:

```
$ grep -r 'gio::write( OutputFileInits' * | wc -l
     724
```

I don't ask this to correct your number, I ask because I'm imagining setting up a script to convert the lines into function calls.  It would be nice to convert these lines:

```
gio::write( OutputFileInits, fmtA ) << "! <Site:GroundTemperature:BuildingSurface>, Months From Jan to Dec {C}";
```

into:

```
writeToEIO( formatString, args... );
```

possibly using variadics.  But ugh.  I don't think I want to go down that path right now.  I would love to see a more C++ approach to IO brought in, but that isn't a part of this PR.  Can you just respond to the other changes I mentioned and we'll move to get this merged.

JasonGlazer: Visual Studio "Find in Files" got the 822 but it includes the unit tests.

I would really like to revisit this but maybe it deserves new feature level funding with an NFP and a team effort figuring out the best approach.

Myoldmopar: Yep, let's wrap it up as best we can here and then we'll discuss bigger things later.  @jasondegraw is dying to discuss it.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: weird, shows up ok in VS. I deleted all leading characters and entered spaces.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: @Myoldmopar did you want to take a shot at refactoring this still?

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: thanks

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -774,6 +775,15 @@ namespace OutputReportTabular {
...
+	splitCommaString(std::string inputString);
```
mbadams5: Should use `std::string const & inputString`

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -940,6 +950,12 @@ namespace OutputReportTabular {
 	std::string
 	DateToString( int const codedDate ); // word containing encoded month, day, hour, minute
 
+	bool
+	isNumber(std::string s);
```
mbadams5: Should use `std::string const & s`

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -940,6 +950,12 @@ namespace OutputReportTabular {
...
+
+	int
+	digitsAferDecimal(std::string s);
```
mbadams5: Should use `std::string const & s`

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -14180,6 +14356,41 @@ Label900: ;
 		return StringOut;
 	}
 
+	bool 
+	isNumber(std::string s)
+	{
+		bool is = true;
+		try {
+			std::stof(s);
+		} catch ( std::exception & e ) {
```
mbadams5: @Myoldmopar I know you suggested this, but I disagree with using a try/catch to check if it is a number. That is an expensive paradigm that I don't want to spread in E+ code. Also, I don't think `stof()` will function properly for very large numbers since you are checking the broad `std::exception`. `stof` throws `std::invalid_argument` when out of range of floats (which could happen since we are using doubles in all of E+). You should use `std::stod`.

An alternate approach is to use `strtod`:
```cpp
bool
isNumber(std::string const & s) {
  char* p;
  strtod( s.c_str(), &p, 10 );
  return *p == 0;
}
```

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -1943,9 +1945,15 @@ namespace OutputReportTabular {
 					displayEconomicResultSummary = true;
 					WriteTabularFiles = true;
 					nameFound = true;
-				} else if ( SameString( AlphArray( iReport ), "EnergyMeters" ) ) {
+				} else if (SameString(AlphArray(iReport), "EnergyMeters")) {
 					WriteTabularFiles = true;
 					nameFound = true;
+				} else if (SameString(AlphArray(iReport), "EIO")) {
+					WriteTabularFiles = true;
```
mbadams5: Should this be `displayEioSummary` since it is checking for "EIO"? I could be wrong.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -1943,9 +1945,15 @@ namespace OutputReportTabular {
 					displayEconomicResultSummary = true;
 					WriteTabularFiles = true;
 					nameFound = true;
-				} else if ( SameString( AlphArray( iReport ), "EnergyMeters" ) ) {
+				} else if (SameString(AlphArray(iReport), "EnergyMeters")) {
 					WriteTabularFiles = true;
 					nameFound = true;
+				} else if (SameString(AlphArray(iReport), "EIO")) {
```
mbadams5: Is "EIO" supposed to be in the list of choices in the IDD?

mbadams5 (commented):
@Myoldmopar We should really put a test in the performance tests that stress the generation of ALL outputs.

My biggest comment about this PR is my concern over memory consumption and performance when the EIO file is large.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: This did not pass all the unit tests. Specifically:	

EXPECT_TRUE(isNumber("0 "));

failed.

src/EnergyPlus/OutputReportTabular.hh
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.hh
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.hh
JasonGlazer: ok

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: No, it is a undocumented shortcut. There are a few others and this one made sense to include to me, primarily for developers.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: Thanks

src/EnergyPlus/OutputReportTabular.cc
Myoldmopar: I'm not opinionated on the strtod thought.  I trust you know the performance impact better than myself.  I was expecting this wouldn't be a big performance issue since it was happening only one time during the simulation and we do much worse grotesque performancy things during "one-time" parts of the simulation.

Out of curiosity, since you'd have to check the return value of isNumber every time anyway, how much worse is performance when using a try/catch over an if/then?

mbadams5: If an exception is rarely thrown, then it would be similar to (and maybe just slightly faster) than an if statement. However, when an exception is thrown there is a large cost associated with it. So the big question is how often this function is checking actual numbers versus non-numbers. If it isn't overwhelmingly numbers then you'd have to profile to see the actual performance impact.

I understand this is reporting code and is only run once. I was on the fence about suggesting a change to begin with because of that reason. I just know how programming patterns have a tendency to move around in E+ and I don't want this pattern in the simulation code itself eventually.

Also if @JasonGlazer sticks with the exception based approach then it should change to `std::stod` like I mentioned in my original post.

Myoldmopar: Great!  Thanks for the insight, that helps, and makes sense.  If @JasonGlazer can move forward with your advised function, that's awesome.

@JasonGlazer I'm not sure about the trailing whitespace issue.  Is there a different way to parse the lines that ensures you don't end up with whitespace padded strings?  Or would you have to trim the strings before passing them in?

mbadams5: If trailing whitespace is a concern here is some updated (untested) code to deal with it.
```cpp
bool
isNumber(std::string const & s) {
  char* p;
  strtod( s.c_str(), &p, 10 );
  for ( ; isspace( *p ); ++p ); // handle trailing whitespace
  return *p == 0;
}
```

JasonGlazer: @mbadams5 MSVS doesn't like the third argument on strtod(), and the c++ web sites I usually use only show two arguments. What is the 10 supposed to do?

mbadams5: I thought it took a third argument to specific the base of the number, hence the 10. Looking at the documentation, it doesn't seem to take a third parameter like you said and instead automatically determines the base. You should just remove that last input parameter. I did say it was untested 😉

JasonGlazer: No problem. I am compiling the test suite as I am typing.

nrel-bot-3:

src/EnergyPlus/OutputReportTabular.cc
Myoldmopar: Nope.

Myoldmopar:
@JasonGlazer It still looks like there are parentheses padding issues.  I commented and saw you received the message, but they are still there.  I'm hesitant to merge in the inconsistency.  Could you address those, and resolve the merge conflict?  

@mbadams5 Do you want to invest any time in discussing a more memory-conscious approach here before it goes in?  How intense do you expect it to be for 99.5% of simulations?

JasonGlazer:
@Myoldmopar hopefully I caught all the weird formatting stuff. I reconfigured my VS settings to try to do this better. Let me know if you spot anything that is still a problem.

Myoldmopar (approved):

Issue #5118, opened 2015-08-18
Custom meters assigned to a custom meter are read as all zeros


ejhw:
It would be nice to allow this functionality, but at a minimum this should trigger a warning and be mentioned in the documentation.

8.3 defect file illustrating issue: https://gist.github.com/ejhw/06541ff6e4ecbb2f8bc9

cc @nmerket

mjwitte:
Closed via #5942

PR #5942, opened 2016-11-16
Add warning when Meter:Custom references another Meter:Custom

JasonGlazer:
Add warning and no longer produce output in the case when a Meter:Custom references another Meter:Custom.

Addresses #5118

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/OutputProcessor.cc
+++ b/src/EnergyPlus/OutputProcessor.cc
@@ -1672,6 +1672,14 @@ namespace OutputProcessor {
 		cCurrentModuleObject = "Meter:Custom";
 		NumCustomMeters = GetNumObjectsFound( cCurrentModuleObject );
 
+        //make list of names for all Meter:Custom since they cannot refer to other Meter:Custom's
+		std::vector<std::string> namesOfMeterCustom;
+		for ( Loop = 1; Loop <= NumCustomMeters; ++Loop ) {
+			GetObjectItem (cCurrentModuleObject, Loop, cAlphaArgs, NumAlpha, rNumericArgs, NumNumbers, IOStat, lNumericFieldBlanks, lAlphaFieldBlanks, cAlphaFieldNames, cNumericFieldNames);
+			namesOfMeterCustom.push_back(cAlphaArgs(1));
+		}
+
+
 		for ( Loop = 1; Loop <= NumCustomMeters; ++Loop ) {
 			GetObjectItem( cCurrentModuleObject, Loop, cAlphaArgs, NumAlpha, rNumericArgs, NumNumbers, IOStat, lNumericFieldBlanks, lAlphaFieldBlanks, cAlphaFieldNames, cNumericFieldNames );
 			lbrackPos = index( cAlphaArgs( 1 ), '[' );
@@ -1689,6 +1697,22 @@ namespace OutputProcessor {
 			VarsOnCustomMeter = 0;
 			MaxVarsOnCustomMeter = 1000;
 			NumVarsOnCustomMeter = 0;
+			// check if any fields reference another Meter:Custom
+			int found = 0;
+			for ( fldIndex = 4; fldIndex <= NumAlpha; fldIndex += 2 ) {
+				for ( auto nameOfMeterCustom : namesOfMeterCustom ) {
+					if ( SameString (cAlphaArgs (fldIndex), nameOfMeterCustom)) {
+						found = fldIndex;
+						break;
+					}
+				if ( found != 0) break;
+				}
+			}
+			if ( found != 0 ) {
+				ShowWarningError (cCurrentModuleObject + "=\"" + cAlphaArgs (1) + "\", contains a reference to another " + cCurrentModuleObject + "in field: " + cAlphaFieldNames (found) + "=\"" + cAlphaArgs (found) + "\".");
+				continue;
+			}
+
 			for ( fldIndex = 3; fldIndex <= NumAlpha; fldIndex += 2 ) {
 				if ( cAlphaArgs( fldIndex ) == "*" || lAlphaFieldBlanks( fldIndex ) ) {
 					KeyIsStar = true;
@@ -1816,7 +1840,7 @@ namespace OutputProcessor {
 			}
 			if ( NumVarsOnCustomMeter == 0 ) {
 				ShowWarningError( cCurrentModuleObject + "=\"" + cAlphaArgs( 1 ) + "\", no items assigned " );
-				ShowContinueError( "...will not be shown with the Meter results" );
+				ShowContinueError( "...will not be shown with the Meter results. This may be caused by a Meter:Custom be assigned to another Meter:Custom." );
 			}
 
 		}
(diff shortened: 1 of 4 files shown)


src/EnergyPlus/OutputProcessor.cc
```diff
@@ -1672,6 +1672,14 @@ namespace OutputProcessor {
 		cCurrentModuleObject = "Meter:Custom";
 		NumCustomMeters = GetNumObjectsFound( cCurrentModuleObject );
 
+        //make list of names for all Meter:Custom since they cannot refer to other Meter:Custom's
+		std::vector<std::string> namesOfMeterCustom;
```
mbadams5: @JasonGlazer I would use either `std::set` or `std::unordered_set`. Depending how many Meter:Custom are typically in an IDF that use them, there will be a point where `unordered_set` is faster than `set`, especially since this is only used for searching. However, both will be faster than linear searching a vector.

Also, you should use `reserve()`, especially in the case for `unordered_set`. 
`namesOfMeterCustom.reserve( NumCustomMeters );`
This reserves capacity in the vector/set but doesn't allocate it. This allows you to avoid costly capacity increases when adding more items to the container.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -1672,6 +1672,14 @@ namespace OutputProcessor {
...
+		for ( Loop = 1; Loop <= NumCustomMeters; ++Loop ) {
+			GetObjectItem (cCurrentModuleObject, Loop, cAlphaArgs, NumAlpha, rNumericArgs, NumNumbers, IOStat, lNumericFieldBlanks, lAlphaFieldBlanks, cAlphaFieldNames, cNumericFieldNames);
+			namesOfMeterCustom.push_back(cAlphaArgs(1));
```
mbadams5: You should use `emplace_back` in this case for vector, and with a change to `std::set` you would use `emplace`. This constructs the item in the vector/set in-place instead of copying, which saves memory/time.

You should also call `MakeUPPERCase( cAlphaArgs( 1 ) )` when you put them into the set. This will affect my suggested code refactor below.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -1689,6 +1697,22 @@ namespace OutputProcessor {
 			VarsOnCustomMeter = 0;
 			MaxVarsOnCustomMeter = 1000;
 			NumVarsOnCustomMeter = 0;
+			// check if any fields reference another Meter:Custom
+			int found = 0;
+			for ( fldIndex = 4; fldIndex <= NumAlpha; fldIndex += 2 ) {
+				for ( auto nameOfMeterCustom : namesOfMeterCustom ) {
```
mbadams5: With a change to using set, I would rewrite all this code as follows...

``` cpp
bool referenceMeterCustom = false;
for ( fldIndex = 4; fldIndex <= NumAlpha; fldIndex += 2 ) {
  if ( namesOfMeterCustom.find( MakeUPPERCase( cAlphaArgs( fldIndex ) ) ) != namesOfMeterCustom.end() ) {
    referenceMeterCustom = true;
    break;
  }
}
if ( referenceMeterCustom ) {
  ShowWarningError (cCurrentModuleObject + "=\"" + cAlphaArgs (1) + "\", contains a reference to another " + cCurrentModuleObject + "in field: " + cAlphaFieldNames (found) + "=\"" + cAlphaArgs (found) + "\".");
  continue;
}
```

With these changes, this goes from an O( n \* m ) algorithm to O( n \* log( m ) ) for set or O( n ) for unordered_set aka faster.

tst/EnergyPlus/unit/OutputProcessor.unit.cc
```diff
@@ -3355,6 +3355,72 @@ namespace EnergyPlus {
 			} ) );
 
 		}
+		TEST_F (SQLiteFixture, OutputProcessor_GenOutputVariablesAuditReport)
```
mbadams5: This unit test doesn't seem to test the code changes in this PR. You don't have any `Meter:Custom` in the IDF snippet. Is this unit test for a different pull request?

tst/EnergyPlus/unit/OutputProcessor.unit.cc
```diff
@@ -3355,6 +3355,72 @@ namespace EnergyPlus {
...
+
+			auto timeStep = 1.0 / 6;
+
+			SetupTimePointers ("Zone", timeStep);
+			SetupTimePointers ("HVAC", timeStep);
+
+			TimeValue (1).CurMinute = 50;
+			TimeValue (2).CurMinute = 50;
+
+			EnergyPlus::sqlite = std::move (sqlite_test);
```
mbadams5: @JasonGlazer Make sure you `std::move` back to `sqlite_test`  from `EnergyPlus::sqlite` after you are finished running any functions that require `EnergyPlus::sqlite` to exist.

`sqlite_test = std::move( EnergyPlus::sqlite );`

mbadams5: @JasonGlazer Make sure you `std::move` back to `sqlite_test`  from `EnergyPlus::sqlite` after you are finished running any functions that require `EnergyPlus::sqlite` to exist.

`sqlite_test = std::move( EnergyPlus::sqlite );`

src/EnergyPlus/OutputProcessor.cc
JasonGlazer: @mbadams5 thanks for the suggestions. I will go ahead and implement them but one thing you should know is that I have never seen more than just 3 of these objects in an input file.

tst/EnergyPlus/unit/OutputProcessor.unit.cc
JasonGlazer: It was intended to be part of this PR but you are right it does not test Meter:Custom. I decided to test another part of outputprocessor.cc that wasn't yet covered by any unit tests.

tst/EnergyPlus/unit/OutputProcessor.unit.cc
JasonGlazer: I should probably just remove that line.

src/EnergyPlus/OutputProcessor.cc
shorowit: FWIW,  I just looked at an example input file produced by BEopt and we have over 30 Meter:Custom objects in it. (That said, we eventually would like to move away from using so many. For example, now that E+ supports oil/propane heating coils and equipment, we can remove a bunch.)

JasonGlazer: @shorowit thanks for letting me know. That does change my perspective on the value of optimizing this.

tst/EnergyPlus/unit/OutputProcessor.unit.cc
JasonGlazer: @mbadams5 thanks for your review.

JasonGlazer:
@mjwitte @Myoldmopar or @mbadams5 I think this is ready to review.

JasonGlazer:
Thanks for the unit test @mbadams5

mbadams5:
CI looks good with the new unit test. Merging.

Issue #5935, opened 2016-11-09
Incorrect Zone column content in Lighting Summary - Daylighting HTML table 


asparke2:
Issue overview
--------------
In the HTML file of models with daylighting controls, the Lighting Summary - Daylighting table has a column labeled `Zone`, but this column contains the name of the `Daylighting:Controls` object instead.  Rather than changing the heading, it would be preferable to change the contents to actually be the `Zone` names.

Here's an example from the HTML file:

![image](https://cloud.githubusercontent.com/assets/2009290/20144706/ac40aebc-a65a-11e6-9f85-ccd156889fbf.png)

And the corresponding IDF:

```
  Daylighting:Controls,
    Lobby_Records_Flr_1_DaylCtrl,  !- Name
    Lobby_Records_Flr_1,     !- Zone Name
    SplitFlux,               !- Daylighting Method
    ,                        !- Availability Schedule Name
    Stepped,                 !- Lighting Control Type
    0.3,                     !- Minimum Input Power Fraction for Continuous or ContinuousOff Dimming Control
    0.2,                     !- Minimum Light Output Fraction for Continuous or ContinuousOff Dimming Control
    3,                       !- Number of Stepped Control Steps
    1.0,                     !- Probability Lighting will be Reset When Needed in Manual Stepped Control
    Lobby_Records_Flr_1_DaylRefPt1,  !- Glare Calculation Daylighting Reference Point Name
    270.0,                   !- Glare Calculation Azimuth Angle of View Direction Clockwise from Zone y-Axis {deg}
    22.0,                    !- Maximum Allowable Discomfort Glare Index
    ,                        !- DElight Gridding Resolution {m2}
    Lobby_Records_Flr_1_DaylRefPt1,  !- Daylighting Reference Point 1 Name
    0.04,                    !- Fraction of Zone Controlled by Reference Point 1
    375.0,                   !- Illuminance Setpoint at Reference Point 1 {lux}
    Lobby_Records_Flr_1_DaylRefPt2,  !- Daylighting Reference Point 2 Name
    0.04,                    !- Fraction of Zone Controlled by Reference Point 2
    375.0;                   !- Illuminance Setpoint at Reference Point 2 {lux}

  Daylighting:ReferencePoint,
    Lobby_Records_Flr_1_DaylRefPt1,  !- Name
    Lobby_Records_Flr_1,     !- Zone Name
    1.397000,                !- X-Coordinate of Reference Point {m}
[...]

Issue #5894, opened 2016-09-30
Lights object references discontinued Daylighting:DElight:Controls object in v8.6 IDD


aaron-boranian:
## Issue overview

In the IDD file for v8.6, the **Fraction Replaceable** input field of the **Lights** object has a note that references the **Daylighting:DElight:Controls** object, which is no longer available. Since this error is in a note, it's only a documentation error and doesn't affect usability.

JasonGlazer:
@aaron-boranian good catch!

mjwitte:
Closed via #5945

PR #5945, opened 2016-11-21
Fix daylighting table columns in lighting summary report

JasonGlazer:
Address #5935.

The column labeled Zone used to show the daylighting control name but has been fixed to show the zone name. In addition, the control name now appears in a column with the heading control name. Also the column heading "Daylighting Type" was changed to "Daylighting Method" to be more consistent with the input field of that it is reporting. This column also was always just showing "Detailed" previously and now shows either SplitFlux or DElight as appropriate.

Also address small change in IDD in #5894 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes:
   - [x] minor \notes change
   - [na] If transition, add rules to spreadsheet
 - [na] If new idf included, locally check the err file and other outputs
 - [na] Documentation changes in place
 - [na] ExpandObjects changes?


JasonGlazer:
I spot checked the AUD diffs in the last CI runs and they seem reasonable given that the number of columns changed in one table.

JasonGlazer:
@mjwitte and @Myoldmopar looks like the same audit file differences. This should be ready for review and merging.

nrel-bot-3:

Issue #5695, opened 2016-05-31
Hot water preheat coil not running but electric preheat works as expected


mjwitte:
## Issue overview

From user:
I have constructed two VAV air systems. The difference between the two systems is the preheat coil in the outdoor air system, one is a hot water preheat coil, the other is a electric resistance heating coil. After the heating coil I have a setpoint manager with a constant temperature setpoint of 11.67 C. When I look at the heating design day simulation results, I see the following:
1. Both have economizer turned on.
2. For the case with hot water preheat coil, the preheat coil seems to be locked out by the economizer - it's not operating.
3. For the case with electric resistance preheat coil, the preheat coil is operating heating air to its setpoint of 11.67 C
### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus v8.5.0
- Helpdesk ticket 11367
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (DefectFiles\5695)
- [x] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

EnergyArchmage:
closed via PR #5973

PR #5973, opened 2017-01-05
Hot water preheat coil now runs when economizing

Nigusse:
Now hot water preheat coils in the outdoor air system works as expected just like electric or gas preheat coil.  Previously, the hot water coil could be locked out during air-side economizing (like is desirable for chilled water cooling coil) and the coil would not run to preheat too-cold of air.  

Addresses issue #5695.

Pull request overview
---------------------
Program was modified to change how hot water heating coils located in the outdoor air system behave in the context of economizer lockout.  Previously the hot water coils would not run during economizing, now the heating coils are allowed to run. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Nigusse:
Investigating HW preheat coil in OA system does not run under certain input conditions. This happens when the economizer is active (economizer on). Economizer ON does not mean the OA flow is full it could be at minimum for some conditions (documentation). This logic works for electric and steam coils (I tested both) and probably gas coils as well.

Below is the code that forces the HW coil Controllers skip the controller calculation:

// if the controller can be locked out by the economizer operation and the economizer is active, leave the controller inactive
if ( AirLoopControlInfo( AirLoopNum ).EconoActive && PrimaryAirSystem( AirLoopNum ).CanBeLockedOutByEcono( AirLoopControlNum ) ) {
				ControllerConvergedFlag = true;
				continue;
}
The PrimaryAirSystem( AirLoopNum ).CanBeLockedOutByEcono( AirLoopControlNum )  variable is always true if the HW coil is on OA system. I just turned off this variable for hot water coils in OA system [here](https://github.com/NREL/EnergyPlus/compare/137048597_HWCoilOnAOSys?expand=1#diff-f51329ac8013440dabda3acd6c2d2a9a) and the defect file worked as expected just like electric and steam coils but not sure if this change has unintended effect. @mjwitte ?

Nigusse:
Added unit test and did cleanup. This branch is ready for review and merging.

EnergyArchmage:
@Nigusse, can you fix the PlantLoop objects in two new unit tests here, or should I?  Looks like they need to be lengthened to include the new circulation time field.

Myoldmopar:
This branch is still 9 commits behind develop.  I think those are fixed in develop.

Nigusse:
@EnergyArchmage I will update the branch, and push it.

Nigusse:
@EnergyArchmage Fixed the unit tests failed due to new input field addition in the PlantLoop object. Pushed the latest commit. These unit tests are in this branch, so updates in develop does not capture them.

Nigusse:
@EnergyArchmage This is all green, ready for merge.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1105,6 +1105,15 @@ namespace SimAirServingZones {
 								}
 							}
 						}
+						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
+						for (OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum) {
+							for (OACompNum = 1; OACompNum <= OutsideAirSys( OASysNum ).NumComponents; ++OACompNum) {
+								CompType = OutsideAirSys( AirSysNum ).ComponentType( OACompNum );
```
nealkruis, line 1111: @Nigusse @EnergyArchmage Should this be `OutsideAirSys( OASysNum )` instead of `OutsideAirSys( AirSysNum )`?

I'm seeing some array extents assertion failures related to this line.

Nigusse: @nealkruis I agree it should have been OASysNum.

PR #5980, opened 2017-01-12
EMPD R-coating bugfix

nmerket:
Pull request overview
---------------------
This fixes a bug in the EMPD calculation where the resistance of the coating layer is erroneously included in an intermediate calculation. 

@jdwoods21 

### Work Checklist
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description. *I didn't make a GitHub issue for this one. I just fixed it.*
   - ~~Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring~~
   - ~~NewFeature: This pull request includes code to add a new feature to EnergyPlus~~
   - ~~Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus~~
   - ~~DoNoPublish: This pull request includes changes that shouldn't be included in the changelog~~
 - [x] Unit Tests

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


src/EnergyPlus/MoistureBalanceEMPDManager.cc
```diff
@@ -599,7 +600,7 @@ namespace MoistureBalanceEMPDManager {
 		}
 		// Calculate resistance between surface-layer/air interface and center of surface layer. [s/m]
 		// This is the physical surface of the material.
-		RSurfaceLayer = 1.0 / hm_surf_layer - 1.0 / h_mass_conv_in_fd - Rcoating;
+		RSurfaceLayer = 1.0 / hm_surf_layer - 1.0 / h_mass_conv_in_fd;
```
nmerket, line 603: This is the bug fix.

nmerket:
@Myoldmopar @mjwitte This is a minor bug fix that's ready to go, I believe. Let me know if you need anything else from me.

nmerket:
Ping @Myoldmopar @mjwitte.

src/EnergyPlus/MoistureBalanceEMPDManager.cc
Myoldmopar: Can you clarify what this is doing?  The h terms seem to be conductance like, such that resistances are added/subtracted as reciprocals, but why the subtraction here?  Taking a previously calculated total resistance and shaving off some of it?  I'm not saying it's wrong, but it is odd.

nmerket: @jdwoods21, I think you can answer his question better than I can.

jdwoods21: The subtraction is meant to find only a portion of the overall resistance, such that we can calculate the humidity ratio at the actual surface (with the material resistance on one side, and the air boundary layer on the other side). This is required because the other part of the model (HeatBalanceSurfaceManager) uses this surface condition (rather than the condition inside the surface) to calculate moisture transfer from the surface to the zone air.

The mistake that was corrected was that I originally also subtracted the paint-layer resistance, but this is no included in the HeatBalanceSurfaceManager calculations, so it needed to be removed.

Myoldmopar (approved):

Myoldmopar:
Was there a github issue or pivotal ticket for me to accept?  I don't see them.

nmerket:
I don't think there was.

On Feb 9, 2017, at 7:32 AM, Edwin Lee <notifications@github.com<mailto:notifications@github.com>> wrote:


Was there a github issue or pivotal ticket for me to accept? I don't see them.

-

PR #5790, opened 2016-07-26
Eliminates timestep dependence in plant volume autocalculation

rglazer2:
## Pull request overview

This pull request addresses defects described by issue #4188 Timestep Dependence in Plant Volume Autocalculation.
- Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description

Superceded by #6000.

### Review Checklist

This will not be exhaustively relevant to every PR.
- [ ] Code style (parentheses padding, variable names)
- [ ] Functional code review (it has to work!)
- [ ] If defect, results of running current develop vs this branch should exhibit the fix
- [ ] CI status: all green or justified
- [ ] Performance: CI Linux results include performance check -- verify this
- [ ] Unit Test(s)
- C++ checks:
  - [ ] Argument types
  - [ ] If any virtual classes, ensure virtual destructor included, other things
- IDD changes:
  - [ ] Verify naming conventions and styles, memos and notes and defaults
  - [ ] Open windows IDF Editor with modified IDD to check for errors
  - [ ] If transition, add rules to spreadsheet
  - [ ] If transition, add transition source
  - [ ] If transition, update idfs
- [ ] If new idf included, locally check the err file and other outputs
- [ ] Documentation changes in place
- [ ] ExpandObjects changes??
- [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #6000, opened 2017-02-06
Revise plant and condenser loop volume autocalculate and add new field for loop circulation time

mjwitte:
Pull request overview
---------------------
Addresses #4188 to remove timestep dependency from the PlantLoop and CondenserLoop volume autocalculate methodology, and adds a new field "Loop Circulation Time" to allow user control over the ratio of Loop Volume to Maximum Loop Flow Rate.  The new field defaults to 2 minutes, which is typical for chilled water loops and results in significantly smaller loop volumes than the prior equation.

### Work Checklist

### Review Checklist
*Test files in EnergyPlusDevSupport test timestep-dependency and new input field for both plant and condenser loop objects*
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
Replaces #5790.  Some relevant discussion is shown there.

mjwitte:
@Myoldmopar These changes result in plant loop volumes which are 10-50% of the prior values (depending on the timestep in the test file).  This causes many diffs, and will likely need offline annual test suites to evaluate if any test files show significant diffs in energy use.  

One particular test file, PipingSystem_Underground_FHX, had previously shown a significant increase in run time with similar changes which prompted adding the new input file.  In the defect files folder, there are two test files, one with the default 2 minute loop time, and another with 20 minute loop time.  The 20 minute file gets the run time back to where it was with v8.6, but the annual energy results are varying widely across the three cases.  That seems suspect.

Myoldmopar (approved):
Everything looks good here.  Just what I'd expect.  I  haven't built it locally, but the code/doc changes seem good.

EnergyArchmage:
Note there is a third spot where the volume calc needs the change, see
https://github.com/NREL/EnergyPlus/blob/4188_3TimestepDependentPlantLoopAutocalc/src/EnergyPlus/SizingAnalysisObjects.cc#L606

mjwitte:
@EnergyArchmage Thanks.  Will add.

mjwitte:
@Myoldmopar @mitchute  I never did make a change to the default loop time for PipingSystem_Underground_FHX.

mitchute:
@mjwitte could you remind me what needs to happen with the PipingSystem_Underground_FHX file?

Myoldmopar:
I think nothing.  I believe the change was just intended to make it a no-diffs result.

mjwitte:
Not so fast @mitchute From [this comment](https://github.com/NREL/EnergyPlus/pull/6000#issuecomment-277830588):

> One particular test file, PipingSystem_Underground_FHX, had previously shown a significant increase in run time with similar changes which prompted adding the new input file. In the defect files folder, there are two test files, one with the default 2 minute loop time, and another with 20 minute loop time. The 20 minute file gets the run time back to where it was with v8.6, but the annual energy results are varying widely across the three cases. That seems suspect.

I'd have to re-run this both ways to confirm, but it seemed that the results and run time were both too sensitive to loop volume - implying a problem somewhere.

Myoldmopar:
If that's the case then it's appropriate to open a new issue to investigate the time step dependence, not lump it with the small idd/idf tweaks that I just merged in.

mjwitte:
True.  I'll leave it to @mitchute to investigate and post an issue if need be.

Issue #6001, opened 2017-02-07
Construction name missing from surface details report and other eio output problems


mjwitte:
Issue overview
--------------
1. The surface details report (`Output:Surfaces:List,Details or DetailsWithVertices;`) is not reporting the surface construction name (since v8.2).

2. The new table output InitializationSummary which echoes the eio output has some formatting/heading issues:

- The Zone Surfaces and Shading Surfaces reports contain the string "\#" which trips up the TableDiff comparison utility resulting in `KeyError: /#Shading Surfaces` failures.
- The Heat Transfer Surface table is misaligned when reporting DetailsWithVertices or Vertices only.
- The Shading Surface table does not show when reporting DetailsWithVertices or Vertices only.
- Now that the Heat Transfer Surface and Shading Surface headers are separate - trailing inapplicable columns can be removed.  Other inapplicable columns should stay so that eio output columns align (e.g. Width is always in the same column for heat transfer, shading, and frame surfaces).  *Not sure if this is important.  Maybe all inapplicable columns should be removed.??*
- There are no units in the main headers for the surface reports.
- If InitializationSummary is the only table output specified, the table file is not written.
- Surfaces with more than four vertices do not show the additional vertices in the table output - this could make the table extremely wide - is this ok as-is?

### Details
Some additional details for this issue (if relevant):
 - Platform - All
 - Version of EnergyPlus 8.6+https://github.com/NREL/EnergyPlus/commit/593b02cfce2f6eb2d0f77e67cf4fb9e820ddb277

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
1ZoneUncontrolledFourAlgorithms - because this was the first one reporting a CI failure
4ZoneWithShading_Simple_1* - because this has shading surfaces
5ZoneAirCooled-January - well, this has shading surfaces also
PurchAirWithDaylighting* - because this has window frame and divider
RefBldgOutPatientNew2004_Chicago - because this has some surfaces with more than 4 vertices (e.g. Floor 1 Reception Slab)

 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6002

PR #6002, opened 2017-02-07
Initialization output cleanup - eio output and table output

mjwitte:
Pull request overview
---------------------
Addresses #6001.

Restores the construction name in the surface details output and cleans up some issues with the new (in v8.7) InitializationSummary table output. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
And this should eliminate the `KeyError: /#Shading Surfaces` CI regression test failures.

mjwitte:
@Myoldmopar @mbadams5 Wanted to unit test eio output, so the latest commit here starts a partial implementation of eso_stream.  Is it ok to leave a mix of writes to eso_stream and gio::write for the same file?  And just convert more to stream as needed.

src/EnergyPlus/OutputReports.cc
```diff
@@ -1987,52 +1983,67 @@ DetailsForSurfaces( int const RptType ) // (1=Vertices only, 10=Details only, 11
 					ExtConvCoeffCalc = ConvCoeffCalcs( std::abs( Surface( surf ).ExtConvCoeff ) );
 				}
 				if ( Surface( surf ).ExtBoundCond == ExternalEnvironment ) {
-					{ IOFlags flags; flags.ADVANCE( "No" ); gio::write( unit, Format_705, flags ) << "ExternalEnvironment"; }
-					{ IOFlags flags; flags.ADVANCE( "No" ); gio::write( unit, Format_705, flags ) << ExtConvCoeffCalc; }
-					{ IOFlags flags; flags.ADVANCE( "No" ); gio::write( unit, Format_705, flags ) << IntConvCoeffCalc; }
+					if ( DataGlobals::eio_stream ) *DataGlobals::eio_stream << "ExternalEnvironment" << "," << ExtConvCoeffCalc << "," << IntConvCoeffCalc << ",";
```
Myoldmopar: :+1: to all these changes.  The only question I'd bring up is whether we actually want to protect every single call with an if check.  Or if we can collect the `if` check to a more centralized location and abort early if it is null.  Not necessary to change now, but if we're already in here cleaning out the commented lines, then maybe it's worth discussion.

src/EnergyPlus/OutputReports.cc
```diff
@@ -1738,16 +1738,15 @@ DetailsForSurfaces( int const RptType ) // (1=Vertices only, 10=Details only, 11
 	std::string AlgoName;
 
 	// Formats
-	static gio::Fmt Format_700( "('! <Zone Surfaces>,<Zone Name>/#Shading Surfaces,# Surfaces')" );
-	static gio::Fmt Format_700b( "('! <Shading Surfaces>,<Zone Name>/#Shading Surfaces,# Surfaces')" );
+	static gio::Fmt Format_700( "('! <Zone Surfaces>,Zone Name,# Surfaces')" );
+	static gio::Fmt Format_700b( "('! <Shading Surfaces>,Number of Shading Surfaces,# Surfaces')" );
```
Myoldmopar: Yay!!

Myoldmopar (changes requested):
This all looks good.  I'll leave it as "request changes" since, at a minimum, we need to clean out the code you left in as comments.  I cannot find anything wrong with your changes and appreciate you going through the effort to add the eio stream here.  If @mbadams5 has any comments, since the unit test structure here is more of his baby, they'd far outweigh my opinion.  But if he doesn't have any issues, then 👍

mbadams5:
The unit test and fixture side looks good.

You might need to update unit tests in the future as you convert more gio::write to the stream but that is the only thing I can see.

mjwitte:
@Myoldmopar Forgot to merge in develop, and also checking out possible issue with reporting to tubular daylighting devices.  So, not quite ready yet.

mjwitte:
@Myoldmopar This should be ready for review now.
@mbadams5 Thanks for looking at this.

PR #6007, opened 2017-02-11
Fix Linux Builds

Myoldmopar:
Looks like Linux is failing to complete the build of E+ because of a doc build issue.  Let's see if this remedies Linux builds....if not I'll keep digging.

@mitchute This so far simply brings back in image900.png from the IO ref, which was, possibly inadvertently, deleted in the ground domain branch.  Can you confirm whether the image900 I re-added here is still the correct version?  Or push a better version to this branch if not?


mjwitte:
@Myoldmopar If you want to get the performance back to where it was, change the loop time to 18.75 (or round it to 20 minutes) in PipingSystem_Underground_FHX in both testfiles and performance.

Myoldmopar:
I haven't figured out yet why those two tests are failing on Linux.  They pass on my local machine, and on my 14.04 VM with a release build.  I'll have to test it on the CI box itself tomorrow.  If the circulation time thing gets performance back up and the other results all look good besides those last two Linux tests, I'll merge this in so that branches can start pulling in develop and getting it much more green on the dashboard.

Myoldmopar:
OK, merging this.  It fixes many problems.  I'm still investigating those last two failures.

PR #6017, opened 2017-02-16
Allow regular expressions as keys for Output:Variable

mbadams5:
Pull request overview
---------------------
This pull request allows regular expressions to be used to better filter Output:Variable keys. Previously, only the exact key or all keys (*) were valid options. Now most typical regular expressions are allowed as options. Check [RE2 Syntax](https://github.com/google/re2/wiki/Syntax) for allowed regular expressions as well as what regular expression syntax RE2 supports.

One key note is that ANY regular expression with a comma (`,`) will fail regardless if it is valid regular expression syntax or not. This is due to the nature of the EnergyPlus input processor since fields are delimited by commas. Thus any comma is seen as the end of one field and the beginning of another.

How this regular expression filtering is powerful:

Given the following IDF snippet
```
NodeList, Zone1Inlets, SalesFloor InletNode 1;
NodeList, Zone2Inlets, SalesFloor InletNode 2;
NodeList, Zone3Inlets, SalesFloor InletNode 3;
NodeList, Zone4Inlets, Office InletNode 1;
NodeList, Zone5Inlets, Office InletNode 2;
NodeList, Zone6Inlets, Office InletNode 3;
```

Previously, if you wanted all `SalesFloor InletNode`'s `System Node Mass Flow Rate`, you had to individually write each as an Output:Variable as follows...

```
Output:Variable, SalesFloor InletNode 1, System Node Mass Flow Rate, timestep;
Output:Variable, SalesFloor InletNode 2, System Node Mass Flow Rate, timestep;
Output:Variable, SalesFloor InletNode 3, System Node Mass Flow Rate, timestep;
```
Or you could select ALL nodes `System Node Mass Flow Rate`, which would include the Office nodes...
```
Output:Variable, *, System Node Mass Flow Rate, timestep;
```
Now you can write it much more concisely with a regular expression...
```
Output:Variable, SalesFloor InletNode [\d]+, System Node Mass Flow Rate, timestep;
```

This is just a simple example, but begins to show the filtering capabilities.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
lol   https://github.com/NREL/EnergyPlus/pull/6017/files?w=1#diff-064872bdf78c6174c034845f5426b83f

Myoldmopar:
The E+ changes are super minimal here, and look good.  Unit tests look good.  I think this can drop right in easily.  This will be a big improvement!  I'll give CI some time to get through the platforms.

mjwitte:
@mbadams5 Will this alter the case of the variable keys/names in the output?
And I assume the licensing of the 3rd party package has been cleared with @jasondegraw

mjwitte:
@mbadams5 Are there I/O Ref changes for this?

jasondegraw:
@mjwitte re2 is BSD-style, so we're all good.

mbadams5:
@mjwitte Yes, the `\retaincase` will alter the case of the variable keys in the output, at least as far as how I understand it to work.  The issue was that without it, the input processor was uppercasing the key and was causing the regular expressions to not match. The code will still use a case-insensitive comparison fall back method if the regular expression engine doesn't find a match.

There are no needed changes to the existing I/O Ref but I guess I could add new text about the regular expressions if needed.

Myoldmopar:
Yeah, good call.  The IO ref could definitely use some info.  @mbadams5 Can you add some content to the relevant output object(s) to describe the capabilities and the limitations (no comma, repetition)?

mbadams5:
@Myoldmopar I updated the I/O Ref to include information about the new regular expression option.

Myoldmopar (approved):

Myoldmopar:
OK, I'm going to get this in.  If anything comes up due to the abrupt merge I'll deal with it.  Thanks @mbadams5, this is a very cool enhancement.

mjwitte:
Then the case change needs a mention in the output rules doc.

mbadams5:
For posterity, the `\retaincase` ends up NOT affecting the ESO file and outputs. Everything is still upper-cased, so there is no need to update the output rules doc.

jmarrec:
Major +1 @mbadams5! Thanks for this!

PR #6022, opened 2017-02-17
Run transition with "makePretty" flag set to cleanup idf formatting

mitchute:
I was working with transition today on another project and saw need to run this. The python file was cleaned up slightly and transition run with MakingPretty enabled. I have not checked the diffs in every file, but the 10-15 random ones I did check looked good.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@mitchute @Myoldmopar All of the idfs still are version 8.6.  Do we want to change them to 8.7 here?

rraustad:
What are the strange characters in 5ZoneSwimmingPool?

mitchute:
@rraustad strange. I'm not sure but I will see if I can get them cleaned up. Thanks for pointing that out.

nrel-bot-3:
@mitchute @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 15 days since this pull request was last updated.

src/Transition/VCompareGlobals.f90
```diff
@@ -104,7 +104,7 @@ module VCompareGlobals
 LOGICAL, DIMENSION(:), ALLOCATABLE :: OTMVarCaution
 INTEGER NumRepVarNames
 
-LOGICAL :: MakingPretty=.false.
+LOGICAL :: MakingPretty=.true.
```
Myoldmopar: You wouldn't want to leave this in the merged version, right?

performance_tests/10zone1vav.idf
```diff
@@ -35,10 +35,12 @@
   Output:DebuggingData,0,0;
 
   ZoneCapacitanceMultiplier:ResearchSpecial,
+    1.,                      !- Name
```
Myoldmopar, line 38: Very interesting.  This object has \min-fields 6 in the IDD, which is why the transition cleanup forces the extra fields on the end.  If those fields really aren't needed, we would probably rather just change the IDD entry, then it wouldn't add all these unnecessary trailing fields.

testfiles/5ZoneSwimmingPool.idf
```diff
@@ -3513,20 +3513,20 @@
 
   Output:Variable,*,Indoor Pool Current Activity Factor,hourly;
 
-  Output:Variable,*,Indoor Pool Current Cover Factor,hourly;
-
+  Output:Variable,*,Indoor Pool Current Cover Factor,hourly;
```
Myoldmopar, line 3516: Hmm, what's going on here?  I can't see the diff.  Invalid character maybe?  Time to crack open the ol' hex editor.

Myoldmopar (changes requested):
Overall this seems really great.  Major cleanup.

I would like to resolve the issue of the Research Especial multiplier min fields topic before pushing toward a merge.  @mjwitte thoughts on that?  

Also, the making_pretty variable is still set to true.  Is that not adjustable via environment variable or anything?  You don't need to investigate that necessarily, but I think that we should probably put it back to false unless I'm missing something.

src/Transition/VCompareGlobals.f90
mitchute: Yes, that needs to be corrected.

mitchute:
@Myoldmopar it would be nice if we had an automated way of doing this. Mine works but, as you can see, from Sept (see issue #5860) to Feb, a lot of of cleanup was required. I actually should probably run it again...

performance_tests/10zone1vav.idf
mjwitte: min-fields isn't the issue here.  The files that are showing the diffs in ResearchSpecial weren't properly transitioned to v8.7  Name and Zone Name were new fields in v8.7.  Looking at 20zone1VAV, for example, in develop, the file says v8.7, but the new fields are not present:
```
  ZoneCapacitanceMultiplier:ResearchSpecial,
    1.,                      !- Temperature Capacity Multiplier
    1.,                      !- Humidity Capacity Multiplier
    1.,                      !- Carbon Dioxide Capacity Multiplier
    1.0;                     !- Generic Contaminant Capacity Multiplier
```
Probably need to go back and re-transition the performance test files from 8.6 to 8.7.
And why isn't ZoneCapacitanceMultiplier:ResearchSpecial throwing an error for Zone Name = "1."?

rraustad: So the example files (or some of them) weren't manually transitioned when these fields were added. Doesn't that mean that the diff's seen here were also present in the PR that added these fields? It also looks like the code does test for zone name so there should be a warning? Unless that zone name was in a zone list!

line 1600 ZoneTempPredictorCorrector:

    Item1 = FindItemInList( cAlphaArgs( 2 ), Zone );
    if ( Item1 == 0 && NumOfZoneLists > 0 ) ZLItem = FindItemInList( cAlphaArgs( 2 ), ZoneList );
    if ( Item1 > 0 ) {
        ZoneNum = Item1;
        Zone( ZoneNum ).FlagCustomizedZoneCap = true;
        Zone( ZoneNum ).ZoneVolCapMultpSens = rNumericArgs( 1 );
        Zone( ZoneNum ).ZoneVolCapMultpMoist = rNumericArgs( 2 );
        Zone( ZoneNum ).ZoneVolCapMultpCO2 = rNumericArgs( 3 );
        Zone( ZoneNum ).ZoneVolCapMultpGenContam = rNumericArgs( 4 );
    } else if ( ZLItem > 0 ) {
        for( int ZonePtrNum = 1; ZonePtrNum < ZoneList( ZLItem ).NumOfZones; ZonePtrNum++ ){
            ZoneNum = ZoneList( ZLItem ).Zone( ZonePtrNum );
            Zone( ZoneNum ).FlagCustomizedZoneCap = true;
            Zone( ZoneNum ).ZoneVolCapMultpSens = rNumericArgs( 1 );
            Zone( ZoneNum ).ZoneVolCapMultpMoist = rNumericArgs( 2 );
            Zone( ZoneNum ).ZoneVolCapMultpCO2 = rNumericArgs( 3 );
            Zone( ZoneNum ).ZoneVolCapMultpGenContam = rNumericArgs( 4 );
        }
    } else {
        ShowSevereError( cCurrentModuleObject + "=\"" + cAlphaArgs( 1 ) + "\" invalid " + cAlphaFieldNames( 2 ) + "=\"" + cAlphaArgs( 2 ) + "\" not found." );
        ErrorsFound = true;
    }

mjwitte: Interesting question.  Looks like these files aren't even running in the performance suite anymore.  I don't see 10zone1VAV and 20zone1VAV in the results list [here](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-d26f74b8a927051dad6e1810bcec27011a454872-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html).  Oh ,that's because they are commented out in the [CMakeLists](https://github.com/NREL/EnergyPlus/blob/8bdce3a6ba876eea980ab77cd6b4ceea36bfc7fa/performance_tests/CMakeLists.txt#L10).  Looking at the history for 8.6 and 8.7 the only change to these files was incrementing the version number, so who knows *what* version these files really are.

So, the bigger question - if we aren't running these files (quite a few actually) then perhaps we should move them to an inactive folder?

rraustad: Just tried to run 20zone1vav.idf and get this:

    ** Severe  ** GetNodeListsInput: NodeList="PVAV_SYS SUPPLY FAN UPSTREAM NODES" has duplicate nodes:
    **   ~~~   ** ...list item=5, "AUTOSIZE", duplicate list item=6, "AUTOSIZE".
    **   ~~~   ** ...list item=5, "AUTOSIZE", duplicate list item=7, "AUTOSIZE".
    **   ~~~   ** ...list item=5, "AUTOSIZE", duplicate list item=8, "AUTOSIZE".

    NodeList,
    PVAV_Sys Supply Fan Upstream Nodes,  !- Name
    COIL:Water:Cooling,      !- Node 1 Name
    PVAV_Sys Cooling Coil,   !- Node 2 Name
    OffFan,                  !- Node 3 Name
    0.022303881,             !- Node 4 Name
    autosize,                !- Node 5 Name
    autosize,                !- Node 6 Name
    autosize,                !- Node 7 Name
    autosize,                !- Node 8 Name
    autosize,                !- Node 9 Name
    autosize,                !- Node 10 Name
    PVAV_Sys Cooling Coil Water Inlet Node,  !- Node 11 Name
    PVAV_Sys Cooling Coil Water Outlet Node,  !- Node 12 Name
    PVAV_Sys Mixed Air Outlet,  !- Node 13 Name
    PVAV_Sys Cooling Coil Outlet,  !- Node 14 Name
    SIMPLEANALYSIS,          !- Node 15 Name
    CROSSFLOW;               !- Node 16 Name

rraustad: Well, the error does show so not sure what happened. There are also many warnings and severes in this file, maybe the transition will eliminate some or all of them. I still think if there was a zone list in the file and the zone name was not found there would be no warning.

    ** Severe  ** ZoneCapacitanceMultiplier:ResearchSpecial="1." invalid Zone or ZoneList Name="1." not found.
    ** Severe  ** GetStagedDualSetpoint: Errors with invalid names in ZoneControl:Thermostat:StagedDualSetpoint objects.
    **   ~~~   ** ...These will not be read in.  Other errors may occur.
    **  Fatal  ** Errors getting Zone Control input data.  Preceding condition(s) cause termination.

The node list should be:

    NodeList,
    PVAV_Sys Supply Fan Upstream Nodes,  !- Name
    PVAV_Sys Mixed Air Outlet,  !- Node 1 Name
    PVAV_Sys Cooling Coil Outlet;  !- Node 2 Name

rraustad: I take it back, if there were a zone list in the file, and that zone list was NOT named "1.", there should also be a warning.

nrel-bot-3:
@mitchute @lgentile it has been 33 days since this pull request was last updated.

testfiles/WWHPSimpleAuto.idf
```diff
@@ -26,7 +26,7 @@
 !           tries to meet the load.
 ! SolDis=FullInteriorAndExterior, Aniso, Detailed Interior and Exterior Convection
 
-  Version,8.7;
+  Version,8.6;
```
mitchute: @Myoldmopar I ran this again, supposedly make-pretty transitioning to 8.8, which is what I'm seeing everywhere else. Is there any reason this file should be moved back to 8.6?

testfiles/5ZoneSwimmingPool.idf
mitchute: @Myoldmopar compared the hex versions of this file and am not seeing any diffs. Suggestions?

Myoldmopar: If the hex codes look the same then I think we're ok.

Myoldmopar (approved):

Myoldmopar:
I resolved the conflicts using the online editor.  The PCM conflict was obvious, I had added new fields while your branch had added missing units.  The other two were Fan:SystemModel conflicts, and I couldn't find any meaningful conflicts in the input parameters themselves, so I think maybe it was just both cleaning up something.  I kept the ones from this branch, but it'll be good to verify the diffs make sense.  Just checked the raw diff and it seems OK to me.  Let's see what CI thinks now that develop is once again pulled in here.

Myoldmopar:
This looks great.  Going in.  If it ends up conflicting others then I'll have resolve them there ( @mjwitte ).  It would just be a matter of taking their version of the conflict and then cleaning up any indentation issues that this branch cleaned up.

Issue #5906, opened 2016-10-12
Adaptive convection resulting in extremely low zone temperature which causes fatal error


mjwitte:
## Issue overview

This user file fails due to a very cold zone air temperature.

```
   ** Warning ** Temperature out of range [-100. to 200.] (PsyPsatFnTemp)
   **   ~~~   **  Routine=CorrectZoneHumRat, Environment=RUN PERIOD 1, at Simulation time=01/12 00:35 - 00:40
   **   ~~~   **  Input Temperature=-263.23
   ** Warning ** Calculated Relative Humidity out of range (PsyRhFnTdbWPb) 
   **   ~~~   **  Routine=CorrectZoneAirTemp, Environment=RUN PERIOD 1, at Simulation time=01/12 00:35 - 00:40
   **   ~~~   **  Dry-Bulb= -263.23 Humidity Ratio= 1.000E-005 Calculated Relative Humidity [%]= 93059.94
   **   ~~~   ** Relative Humidity being reset to 100.0%
   ** Severe  ** PsyRhoAirFnPbTdbW: RhoAir (Density of Air) is calculated <= 0 [-6.29394].
   **   ~~~   ** pb =[99033.33], tdb=[-327.74], w=[2.6406651E-003].
   **   ~~~   **  Routine=CorrectZoneHumRat, Environment=RUN PERIOD 1, at Simulation time=01/16 23:35 - 23:40
   **  Fatal  ** Program terminates due to preceding condition.
```

Running this in the debugger shows that the error is in  in zone 10 (ZoneWest) and the cold temperature is calculated in ZoneTempPredictorCorrector::CorrectZoneAirTemp

```
                TempDepCoef = SumHA + SumMCp + SumSysMCp;
                TempIndCoef = SumIntGain + SumHATsurf - SumHATref + SumMCpT + SumSysMCpT + ( NonAirSystemResponse( ZoneNum ) / ZoneMult + SysDepZoneLoadsLagged( ZoneNum ) );
```

At the time of the error:
TempDepCoef = 0
TempIndCoef = -395
SumHATsurf = 6481
SumHATref = 6876
and all the other terms above are zero.

Turning off interior convection = AdaptiveConvectionAlgorithm gets rid of the error and the file runs to completion, so it _appears_ that the problem is in adaptive convection.
### Details

Some additional details for this issue (if relevant):
- Platform Win64
- Version of EnergyPlus 8.6.0
- helpdesk ticket number 11490
### Checklist

[...]

PR #5993, opened 2017-01-27
Prevent crash in ZoneHVAC:LowTemperatureRadiant:* when the system is heat-only or cool-only

lgu1234:
Pull request overview
---------------------
Addresses #5905.  For ZoneHVAC:LowTemperatureRadiant:* to prevent a crash, the OperationMode change requires one more condition that a heating or cooling system should be available for both constant flow and variable flow systems.

No crash in debug mode.
 
### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [x] Code style (parentheses padding, variable names)
 - [x] Functional code review (it has to work!)
 - [x] If defect, results of running current develop vs this branch should exhibit the fix
 - [x] CI status: all green or justified
 - [x] Performance: CI Linux results include performance check -- verify this
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
   - [x] If any virtual classes, ensure virtual destructor included, other things
 - n/a IDD changes:
 - n/a If new idf included, locally check the err file and other outputs
 - n/a Documentation changes in place
 - n/a Changed docs build successfully
 - n/a ExpandObjects changes?
 - n/a If output changes, including tabular output structure, add to output rules file for interfaces

PR #6026, opened 2017-02-21
Corrected surface convection calculation with AirRef ZoneSupplyAirTemp

### LINKED ISSUE #5905: ZoneHVAC:LowTemperatureRadiant:VariableFlow array bounds error
## Issue overview

While debugging for a different issue, this use file tripped on an array bounds error in the ZoneHVAC:LowTemperatureRadiant:VariableFlow model.  The first radiant system is heating-only, and `HydrRadSys( RadSysNum ).CoolingSystem=false` as it should, but it's getting to [this block](https://github.com/NREL/EnergyPlus/blob/198c6a3cffe4df717545e75a97f01510aa6a1875/src/EnergyPlus/LowTempRadiantSystem.cc#L5102-L5106) which then tries to access data for Node(0).  `OperatingMode` shouldn't equal `CoolingMode` at this point if the system can't cool.

Need to check logic for both cooling and heating operation for both the variable flow and constant flow radiant systems to be sure that CoolingMode is never set when CoolingSystem=false, likewise for heating.
### Details

- Win64
- Version of EnergyPlus 8.6.0
- helpdesk ticket number 11490 (but this is not related to the user's original question)
### Checklist

- [x] Defect file added to EnergyPlusDevSupport\DefectFiles

lgu1234:
Pull request overview
---------------------
Addresses #5906. The changes in ZoneTempPredictorCorrector corrected surface convection load calculation when TAirRef = ZoneSupplyAirTemp. The existing surface load calculation was treated similar to internal gain. When the supply air temperature is higher than the surface temperature, the negative item made zone air temperature drops at every time step. Eventually, zone air temperature was very low to cause severe errors in psy functions. 

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - n/a IDD changes:
 - n/a Changed docs build successfully
 - n/a ExpandObjects changes?


mjwitte:
@lgu1234 Does this PR address #5905 *and* #5906?

lgu1234:
@mjwitte Yes. It covers both.

mjwitte:
Now I see there is PR #5993 which fixes #5905.  So, once that one drops in, then this one will fix only #5906.

lgu1234:
@mjwitte The main reason to use the fix of #5905 is a pre-requirement to fix #5906, as you mentioned.

mjwitte:
See [comments](https://github.com/NREL/EnergyPlus/issues/5906#issuecomment-283092813) in issue thread for 5906.

PR #6027, opened 2017-02-22
Better fallthrough catch; Catch when ZoneNum is 0 for air velocity ca…

Myoldmopar:
Tested locally, airflow network file now passes.  Merging immediately


mjwitte:
@Myoldmopar @mitchute So, it looks like Mac (at least) does not like the better fallthrough catch.  Getting [build warnings](https://github.com/NREL/EnergyPlus/blob/7ad909b66046c0918b8c482011a34fdc51ad6364/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L5444-L5445) for latest develop.  Would it be ok to simply intialize to zero [here](https://github.com/NREL/EnergyPlus/blob/7ad909b66046c0918b8c482011a34fdc51ad6364/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L5444-L5445)?

mitchute:
@mjwitte fine by me.

mjwitte:
@mitchute Now that the new Attic test file is actually running, it's [failing](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-7ad909b66046c0918b8c482011a34fdc51ad6364-x86_64-MacOS-10.9-clang.html).  Can you take a look and make a PR to fix?  And you could add the initializations there.

mbadams5:
`assert( false );` will only work in debug builds. In release builds, it doesn't exist and hence the warnings about uninitialized variables.

PR #6039, opened 2017-03-02
New error for Output:Table:Monthly and Annual don't have proper order for aggregation types

### LINKED ISSUE #5911: Add error message when "hoursshown" aggregation is used in first column of Output:Table:Annual and Output:Table:Monthly
When the advanced aggregation types are used in Output:Table:Monthly or Output:Table:Annual as the first column, an error message should be generated saying that the table results are invalid. The advanced aggregation types depend on a previous column being either Maximum, Minimum, or Hours---. The advanced aggregation types are ValueWhenMaximumOrMinimum, SumOrAverageDuringHoursShown, MaximumDuringHoursShown, MinimumDuringHoursShown. The error message should be smart to detect whenever an advanced aggregation types are used without a previously appropriate aggregation type

To reproduce the problem just take a file using these advanced aggregation types and move them to the first column of the Output:Table:Monthly out Output:Table:Annual reports.

JasonGlazer:
Add new error message for Output:Table:Monthly and Output:Table:Annual when the advanced aggregation types (such as MaximumDuringHoursShown and SumOrAverageDuringHoursShown) are used when the corresponding aggregation type is not used in an earlier column (such as Maximum or HoursNonZero). 

This resolves #5911

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
I added a defect file to show the new warning messages. See [here](https://github.com/NREL/EnergyPlusDevSupport/tree/master/DefectFiles/5000s/5911).

tst/EnergyPlus/unit/OutputReportTabularAnnual.unit.cc
```diff
@@ -272,3 +272,40 @@ TEST_F(EnergyPlusFixture, OutputReportTabularAnnual_columnHeadersToTitleCase)
...
+		"Test Report, !- Name",
+		", !- Filter",
+		", !- Schedule Name",
+		"Electricity:Facility, !- Variable or Meter 2 Name",
+		"SumOrAverageDuringHoursShown, !- Aggregation Type for Variable or Meter 2",
+		", !- field Digits After Decimal 2",
+		"Misc Facility Electric Energy, !- Variable or Meter 3 Name",
+		"SumOrAverage, !- Aggregation Type for Variable or Meter 3",
+		"0; !- field Digits After Decimal 3",
+	} );
```
Myoldmopar, line 289: Do you think it's worth adding a test for `Output:Table:Monthly`?  It's hard to tell from the diff exactly what functional changes were made to the code, so the more unit tests the better.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -6395,244 +6399,272 @@ namespace OutputReportTabular {
...
+						sqlite->createSQLiteTabularDataRecords( tableBody, rowHead, columnHead, MonthlyInput( iInput ).name, MonthlyTables( curTable ).keyValue, "Custom Monthly Report" );
+					}
+				} // if ( missingHourAggError || missingMaxOrMinError )
 			} //jTables
+			if ( missingMaxOrMinError ) {
+				ShowWarningError( "The Output:Table:Monthly report named=\"" + MonthlyInput( iInput ).name + "\" has a valueWhenMaxMin aggregation type for a column without a previous column that uses either the minimum or maximum aggregation types. The report will not be generated." );
+			}
+			if ( missingHourAggError ) {
+				ShowWarningError( "The Output:Table:Monthly report named=\"" + MonthlyInput( iInput ).name + "\" has a --DuringHoursShown aggregation type for a column without a previous field that uses one of the Hour-- aggregation types. The report will not be generated." );
+			}
```
Myoldmopar: If we aren't generating the report, is it worth just shutting it down with a fatal here?  I'd hate for a user to run an hour long simulation and quietly not get the report they were expecting.  Seems severe enough here to call fatal to me, but I don't feel _that_ strongly about it.

Myoldmopar (commented):
Made a few comments here.  Nothing major but I'd like to hear your thoughts on the two comments before continuing on this.

tst/EnergyPlus/unit/OutputReportTabularAnnual.unit.cc
JasonGlazer: Sure, that is a good idea.

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: I don't feel strongly about it either. It seems like most of the messaging related to the monthly reporting were warnings but I can make this one a fatal if you want.

JasonGlazer: Looking at the code, the testing that I added occurs during the writing of the monthly and annual tables near the end of the simulation. I probably should put them at the beginning just after getinput.

JasonGlazer:
@Myoldmopar I made the changes you suggested. It ended up moving around a bunch of code but the checking algorithm is the same it is just being used at a different time.

JasonGlazer:
Merged only hoping that the CI issues were a fluke since I could not reproduce them locally.

JasonGlazer:
Turns out the reason I could not replicate the issue was that I wasn't forcing the design day runs to occur like the CI does.

mjwitte:
Duh - of course!

JasonGlazer:
@Myoldmopar this is finally ready for review again. I added a unit test and moved this so that it was at the beginning of the simulation and threw a fatal so that it would not waste the users time as you suggested.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -695,12 +695,16 @@ namespace OutputReportTabular {
...
 			GetInputTabularTimeBins();
 			GetInputTabularStyle();
 			GetInputOutputTableSummaryReports();
 			// noel -- noticed this was called once and very slow -- sped up a little by caching keys
 			InitializeTabularMonthly();
-			GetInputFuelAndPollutionFactors();
+			if ( isInvalidAggregationOrder( ) ) {
+				ShowFatalError( "OutputReportTabular: Invalid aggregations detected, no simulation performed." );
+			}
+			GetInputFuelAndPollutionFactors( );
```
Myoldmopar, line 707: I'm not going to sweat it here, but it looks like your visual studio is set up to add blank padding for function calls with no args.

src/EnergyPlus/OutputReportTabularAnnual.cc
```diff
@@ -246,6 +249,65 @@ namespace EnergyPlus {
...
+			}
+			for ( annualTableIt = annualTables.begin( ); annualTableIt != annualTables.end( ); ++annualTableIt ) {
+				if ( annualTableIt->invalidAggregationOrder( ) ) {
+					invalidAggregationOrderFound = true;
+				}
+			}
+			if ( invalidAggregationOrderFound ) {
+				ShowFatalError( "OutputReportTabularAnnual: Invalid aggregations detected, no simulation performed." );
+			}
+		}
```
Myoldmopar, line 268: If this were in a more performance-centric section of the code, I'd push on why you even need the intermediate variable and why not just call the fatal from the loop above.  But I won't now.

Myoldmopar (approved):

src/EnergyPlus/OutputReportTabular.cc
JasonGlazer: Thanks

src/EnergyPlus/OutputReportTabularAnnual.cc
JasonGlazer: Just so all instances of this issues are reported in the error file prior to the fatal.

JasonGlazer:
@Myoldmopar I assume that I should make a new pull request when I re-fix this issue? I can continue using the same branch or should I make a new one of those also?

Myoldmopar:
As long as you pull it up to the latest develop then it doesn't matter either way.  Same branch is fine.

JasonGlazer:
I tried with the existing branch but it seemed to be carrying along some baggage of other peoples changes so I just created a new branch instead. See #6101

Issue #6046, opened 2017-03-06
DX Coil rating for bypass factor should use volume converted to standard mass


rraustad:
Issue overview
--------------
Rated performance of HVAC coils should not change with geography. Performance of these coils will change with changes in altitude/pressure, but the rated performance of a coil should be standardized so that a specific coil can be autosized regardless of the location.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6048

PR #6034, opened 2017-02-24
Improved sizing for Coil:Cooling:DX bypass factor

rraustad:
Pull request overview
---------------------
Manually sized DX coils can fail sizing of bypass factor.
### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Code style (parentheses padding, variable names)
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check -- verify this
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add rules to spreadsheet
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Documentation changes in place
 - [ ] Changed docs build successfully
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #6048, opened 2017-03-06
DX coil calculation of bypass factor should use volume converted to standard mass

rraustad:
Pull request overview
---------------------
The DX coil model uses an internal variable to represent how close the supply air temperature approaches the coil surface temperature, or apparatus dew point. This internal variable is called the rated coil bypass factor (Rated CBF) and was calculated based on rated inlet air conditions at the operating air mass flow rate (volume converted to mass at current barometric pressure). This calculation was changed to convert volume to mass flow rate at standard barometric pressure (sea level) to keep all rated information at standard temperature and pressure. Very little change in energy use and performance is expected with this subtle correction. Addresses issue #6046.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - n/a IDD changes:
 - n/a ExpandObjects changes?


rraustad:
1ZoneDataCenterCRAC_wPumpedDXCoolingCoil now shows this warning which shows why it failed at lower altitudes. The volume flow rate at the specified total cooling capacity pushed the outlet condition over the saturation curve when the barometric pressure was higher than Denver. As I recall, the air mass flow rate was 7.something before. I'll change the SHR to 0.83 for this example file.

    ** Severe  ** For object = Coil:Cooling:DX:SingleSpeed, name = "MAIN COOLING COIL 1"
    **   ~~~   ** Calculated outlet air relative humidity greater than 1. The combination of
    **   ~~~   ** rated air volume flow rate, total cooling capacity and sensible heat ratio yields coil exiting
    **   ~~~   ** air conditions above the saturation curve. Possible fixes are to reduce the rated total cooling
    **   ~~~   ** capacity, increase the rated air volume flow rate, or reduce the rated sensible heat ratio for this coil.
    **   ~~~   ** If autosizing, it is recommended that all three of these values be autosized.
    **   ~~~   ** ...Inputs used for calculating cooling coil bypass factor.
    **   ~~~   ** ...Inlet Air Temperature     = 26.67 C
    **   ~~~   ** ...Outlet Air Temperature    = 14.15 C
    **   ~~~   ** ...Inlet Air Humidity Ratio  = 1.125000E-002 kgWater/kgDryAir
    **   ~~~   ** ...Outlet Air Humidity Ratio = 1.036286E-002 kgWater/kgDryAir
    **   ~~~   ** ...Total Cooling Capacity used in calculation = 148300.00 W
    **   ~~~   ** ...Air Mass Flow Rate used in calculation     = 9.831341 kg/s
    **   ~~~   ** ...Air Volume Flow Rate used in calculation   = 8.500000 m3/s
    **   ~~~   **  During Warmup, Environment=DENVER CENTENNIAL ANN HTG 99.6% CONDNS DB, at Simulation time=12/21 00:00 - 00:10
    **   ~~~   ** CalcCBF: SHR adjusted to achieve valid outlet air properties and the simulation continues.
    **   ~~~   ** CalcCBF: initial SHR = 0.85000
    **   ~~~   ** CalcCBF: adjusted SHR = 0.83022

mjwitte:
Can't we just autosize the SHR now with this fix?

rraustad:
Probably, should I go ahead and autosize the entire coil?

    ** Severe  ** For autosizing of Coil:Cooling:DX:SingleSpeed MAIN COOLING COIL 1, a system sizing run must be done.
    **   ~~~   ** No "Sizing:System" objects were entered.
    **   ~~~   ** The "SimulationControl" object did not have the field "Do System Sizing Calculation" set to Yes.

mjwitte:
Not sure @EnergyArchmage built this file originally.

rraustad:
Diff's are misleading. Changing how the CBF is calculated changes the power needed to meet the same sensible load. It's usually a very small change. These 2 files show little difference.

cbfusingvolume

rraustad:
Annual simulation for DXSystemAuto shows little change in energy use, and it's all in the DX coil.

dxsystemauto

src/EnergyPlus/DXCoils.cc
```diff
@@ -9922,8 +9921,11 @@ Label50: ;
 		Real64 Slope( 0.0 ); // Calculated Slope used while hunting for Tadp
 		Real64 Tolerance; // Convergence tolerance for CBF calculations
 		Real64 HTinHumRatOut; // Air enthalpy at inlet air temp and outlet air humidity ratio [J/kg]
+		Real64 AirMassFlowRate; // the standard air mass flow rate at the given capacity [kg/s]
+		Real64 adjustedSHR; // SHR calculated using adjusted outlet air properties []
 		static bool CBFErrors( false ); // Set to true if errors in CBF calculation, fatal at end of routine
 
+		AirMassFlowRate = AirVolFlowRate * PsyRhoAirFnPbTdbW( StdPressureSeaLevel, InletAirTemp, InletAirHumRat, RoutineName );
```
rraustad, line 9548: Fundamental change is right here. Pass in volume and convert to standard mass flow. Function used to pass in mass flow rate, sometimes at STP and other time at elevation. Now it's consistent. Changing back to how it's done now simply means changing StdPressureSeaLevel to StdBaroPress, not that we'd ever do that.

src/EnergyPlus/DXCoils.cc
```diff
@@ -9948,14 +9950,24 @@ Label50: ;
...
 			if ( TotCap > 0.0 ) {
-				if ( ( ( MinRatedVolFlowPerRatedTotCap( DXCT ) - AirMassFlowRate / PsyRhoAirFnPbTdbW( BaroPress, InletAirTemp, InletAirHumRat, RoutineName ) / TotCap ) > SmallDifferenceTest ) || ( ( AirMassFlowRate / PsyRhoAirFnPbTdbW( BaroPress, InletAirTemp, InletAirHumRat, RoutineName ) / TotCap - MaxRatedVolFlowPerRatedTotCap( DXCT ) ) > SmallDifferenceTest ) ) {
-					ShowContinueError( "...Air Volume Flow Rate per Watt of Rated Cooling Capacity is also out of bounds at = " + RoundSigDigits( AirMassFlowRate / PsyRhoAirFnPbTdbW( BaroPress, InletAirTemp, InletAirHumRat, RoutineName ) / TotCap, 7 ) + " m3/s/W" );
+				if ( ( ( MinRatedVolFlowPerRatedTotCap( DXCT ) - AirVolFlowRate / TotCap ) > SmallDifferenceTest ) || ( ( AirVolFlowRate / TotCap - MaxRatedVolFlowPerRatedTotCap( DXCT ) ) > SmallDifferenceTest ) ) {
+					ShowContinueError( "...Air Volume Flow Rate per Watt of Rated Cooling Capacity is also out of bounds at = " + RoundSigDigits( AirVolFlowRate / TotCap, 7 ) + " m3/s/W" );
 				}
 			}
 			ShowContinueErrorTimeStamp( "" );
-			ShowFatalError( "Check and revise the input data for this coil before rerunning the simulation." );
```
rraustad: Same changes as #6034, that PR should be merged first.

tst/EnergyPlus/unit/DXCoils.unit.cc
```diff
@@ -1407,17 +1417,16 @@ namespace EnergyPlus {
 		Real64 const RatedInletAirHumRat( 0.01125 ); // Humidity ratio corresponding to 80F dry bulb/67F wet bulb
 		std::string const CallingRoutine( "DXCoil_ValidateADPFunction" );
 
-		Real64 DesMassFlow = DXCoil( 1 ).RatedAirVolFlowRate( 1 ) * PsyRhoAirFnPbTdbW( StdBaroPress, RatedInletAirTemp, RatedInletAirHumRat, CallingRoutine );
-		Real64 CBF_calculated = CalcCBF( DXCoil( 1 ).DXCoilType, DXCoil( 1 ).Name, RatedInletAirTemp, RatedInletAirHumRat, DXCoil( 1 ).RatedTotCap( 1 ), DesMassFlow, DXCoil( 1 ).RatedSHR( 1 ), true );
+		Real64 CBF_calculated = CalcCBF( DXCoil( 1 ).DXCoilType, DXCoil( 1 ).Name, RatedInletAirTemp, RatedInletAirHumRat, DXCoil( 1 ).RatedTotCap( 1 ), DXCoil( 1 ).RatedAirVolFlowRate( 1 ), DXCoil( 1 ).RatedSHR( 1 ), true );
 
-		EXPECT_NEAR( 0.747472, DXCoil( 1 ).RatedSHR( 1 ), 0.0000001 );
-		EXPECT_NEAR( 0.1012203, CBF_calculated, 0.0000001 );
```
rraustad, line 1484: Don't particularly like this change in BF from 0.1 to 0.0003, but it's all in how you select total capacity and air volume flow rate. We don't restrict the users choices right up to the saturation curve (i.e., BF=0.000001 is OK and I have seen 1E-6 before) so this change is arbitrary given the total capacity and volume flow rate selected here since barometric pressure was changed.

rraustad:
Here's a comparison of an example file with big table diffs. The only difference is in the coil total/sensible capacity output and associated power with a little change in fan power since PLR changed slightly. Annual simulation results are below this figure.

coolingcoilfreezeprevention_ratedcbf

coolingcoilfreezeprevention_ratedcbf_energy

rraustad:
Some example files show failures in the rating procedure. I expect it has to do with choice of barometric pressure in some calculations.

Fault_FoulingAirFilter_RefBldgMediumOfficeNew2004:

    ! &lt;VAV DX Cooling Coil Standard Rating Information>, DX Coil Type, DX Coil Name, Fan Type, Fan Name, Standard Net Cooling Capacity {W}, Standard Net Cooling Capacity {Btu/h}, IEER {Btu/W-h}, COP 100% Capacity {W/W}, COP 75% Capacity {W/W}, COP 50% Capacity {W/W}, COP 25% Capacity {W/W}, EER 100% Capacity {Btu/W-h}, EER 75% Capacity {Btu/W-h}, EER 50% Capacity {Btu/W-h}, EER 25% Capacity {Btu/W-h}, Supply Air Flow 100% {kg/s}, Supply Air Flow 75% {kg/s},Supply Air Flow 50% {kg/s},Supply Air Flow 25% {kg/s}
    - VAV DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:TwoSpeed,VAV_1_COOLC DXCOIL,Fan:VariableVolume,VAV_1_FAN,157344.46,536881.58,12.64,2.77,3.44,4.16,4.27,9.45,11.75,14.20,14.57,7.1339,5.6668,3.7697,1.7215,
    + VAV DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:TwoSpeed,VAV_1_COOLC DXCOIL,Fan:VariableVolume,VAV_1_FAN,157344.46,536881.58,-228.52,2.77,3.43,-999.00,4.28,9.45,11.71,-999.00,14.60,7.1339,5.7491,-999.0000,1.6121,

Was hitting max iterations at 500. I bumped it up to 1000 and the warnings are gone.

     Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size High Speed Rated Sensible Heat Ratio, 0.73733
     Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size Low Speed Gross Rated Sensible Heat Ratio, 0.73733
     VAV DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:TwoSpeed,VAV_1_COOLC DXCOIL,Fan:VariableVolume,VAV_1_FAN,139450.51,475824.88,12.96,2.92,3.57,4.22,4.28,9.95,12.17,14.39,14.59,7.1930,5.7997,3.8934,1.6258,

rraustad:
The ratings we report for DX coils should not change with geography. What we are doing is reporting how the coil performance changes with geography, which is OK as long as the user understands. What I tried to do here is calculate a "rated" bypass factor for a coil. This should be constant for a specific coil. Then I think about these rating reports and wonder why they should change with geography. If I pick a coil out of a catalog, should the ratings change with geography? Makes me think we should do something "standard" for sizing and then allow that coil to operate differently during the simulation.

src/EnergyPlus/DXCoils.cc
```diff
@@ -12148,7 +12155,7 @@ Label50: ;
 		Real64 PartLoadAirMassFlowRate;
 		Real64 AirMassFlowRatio;
 		static Real64 AccuracyTolerance( 0.2 ); // tolerance in AHRI 340/360 Table 6 note 1
-		static int MaximumIterations( 500 );
+		static int MaximumIterations( 1000 );
```
mjwitte, line 11643: Why does this have to work so hard?  Even 500 seems unnecessary.

mjwitte (commented):
Code changes look OK.  Are there any defect files for this?

src/EnergyPlus/DXCoils.cc
rraustad: What I saw in the debugger was that in RegulaFalsi the lower bound of mass flow rate at 0.01 * full mass flow gave a supply air temperature of -1000 and full mass flow rate showed about 14C SAT. With a target SAT of about 12C, RegulaFalsi thought the solution was very near full mass flow rate and took very small steps towards the final answer. It was getting close at 500 iterations but needed more to get there. Could probably do a better job of selecting the bounds for the RegulaFalsi call.

rraustad:
No specific defect file. The example file 1ZoneDataCenterCRAC_withPumpedDXCoolingCoil was the original file that showed problems with geography.

mjwitte:
Will run that locally.  That file isn't showing any error diffs on CI.

The big error dump that comes out now seems overkill.  The SHR gets adjusted and the simulation proceeds.  Seems like the whole thing should be just a warning and maybe just need that last part:
```
** Warning ** For object = Coil:Cooling:DX:SingleSpeed, name = "MAIN COOLING COIL 1"
**   ~~~   ** CalcCBF: SHR adjusted to achieve valid outlet air properties and the simulation continues.
**   ~~~   ** CalcCBF: initial SHR = 0.85000
**   ~~~   ** CalcCBF: adjusted SHR = 0.83022
```

Also, did you notice that _SmallOffice_Dulles now fails warmup convergence?

rraustad:
If we agree that the existing warning is to too much info then I can remove. I didn't want to change too much from what we were doing in the past while I get a handle on what is needed to move forward.
I didn't notice _SmallOffice_Dulles.

mjwitte:
Sorry, I'm not following what prompted this fix?   
1ZoneDataCenterCRAC_withPumpedDXCoolingCoil (as it was before) has no coil error before this fix, but the error came up after this fix.   1ZoneDataCenterCRAC_withPumpedDXCoolingCoil as modified in this branch, has no coil error before or after this fix.

Also, DOAToVRF and RefBldgSmallOfficeNew2004_Chicago now have these warnings where they did not before.

rraustad:
It's OK to be hesitant here, it's a model correction. 

What prompted the change was that the 1ZoneDataCenter file has rated CBF warnings in cities other than Denver. It made me think of a catalog coil where the ratings for that coil are the same regardless of where you are. It's 3-tons rated, has xxx rated "volume" airflow, and has a specific rated SHR. If I enter that coil into E+, should the rated SHR of that coil change with geography? I think it should not but it does. The performance will certainly change with geography (change in air mass flow rate due to altitude) but ratings should not. The DX coil model already accounts for changes in mass flow rate through the coil, whether it be a result of altitude or fan speed. Next up is rated SHR. If someone operates a DX coil at off-design air flow rate does the rated SHR change? It shouldn't but I have to understand how this rating is used within the model. Also think of those users trying to set up a baseline system with 13 SEER. Right now they would have to do this in a city at sea level.

AHRI 210/240:

2.3.1.b. Express the Cooling Full-load Air Volume Rate, the Cooling Minimum Air Volume Rate, and the Cooling Intermediate Air Volume Rate in terms of standard air.
3.13 Standard Air. Air weighing 0.075 lb/ft3 [1.2 kg/m3] which approximates dry air at 70°F [21°C] and at a barometric pressure of 29.92 in Hg [101.3 kPa].

rraustad:
Interesting adjustment factor table from Trane. 

 - Air density adjustment doesn't seem to be wrt sea level and references "leaving air temp"
 - cooling capacity correction is only 4% at 1 mile, would have thought it was more.
 - SHR correction is significant at 10% for 1 mile high resulting in a shift from sensible to latent with little change in total capacity. Compressor still provides nearly the same capacity while the air side is affected by the change in air mass flow rate across the coil.

dxcoilaltitudeadjustmentfactors

nrel-bot-3:

nrel-bot-3:

Myoldmopar:
This is assigned to me in Pivotal right now, but after reading the thread, it feels unfinished.  Do we need to have a brief meeting of the minds to resolve this to completion?

rraustad:
I think this is done. It wouldn't hurt to talk about it. There will be a few ripple effects occur I'm sure. For example, the DOAToVRF does have the warning now. The reason is that mass flow rate based on StdBaroPress was used before and now StdPressureSeaLevel is used to convert volume to mass in CalcCBF.

In this file:
StdBaroPress = 101192.9
StdPressureSeaLevel = 101325.0

So what happens is the you get a slightly higher mass flow rate used in the rated CBF calculation. This pushes the outlet condition just over the saturation curve, and the new code pulls it back. The SHR changes from 0.8 to 0.799. DOAToVRF is an autosized file so the user has no control over this and we may need to figure out how to correct this autosizing issue. 

I see in the CoolingCapacitySizing section that StdBaroPress is used. This means that the sized capacity is not really "rated" it's adjusted based on altitude. The same pressure is used to calc the WB temp, which is used to ride the performance curve. Not sure what affect this has (i.e., higher or lower capacity result). Also notice at the bottom of this code that StdRhoAir is used for fan coils and ventilators but not for other coil types (e.g., DXCoils)

    rhoair = PsyRhoAirFnPbTdbW ( StdBaroPress, CoilInTemp, CoilInHumRat, CallingRoutine );
    CoilInEnth = PsyHFnTdbW ( CoilInTemp, CoilInHumRat );
    CoilInWetBulb = PsyTwbFnTdbWPb ( CoilInTemp, CoilInHumRat, StdBaroPress, CallingRoutine );
    CoilOutEnth = PsyHFnTdbW ( CoilOutTemp, CoilOutHumRat );
    if ( DataTotCapCurveIndex > 0 ) {
        TotCapTempModFac = CurveValue ( DataTotCapCurveIndex, CoilInWetBulb, OutTemp );
    } else if ( DataTotCapCurveValue > 0 ) {
        TotCapTempModFac = DataTotCapCurveValue;
    } else {
        TotCapTempModFac = 1.0;
    }
    if (ZoneEqFanCoil) {
        PeakCoilLoad = max( 0.0, ( StdRhoAir * DesVolFlow * ( CoilInEnth - CoilOutEnth ) ) );
    } else if ( ZoneEqUnitVent ) {
        PeakCoilLoad = max( 0.0, ( StdRhoAir * DesVolFlow * ( CoilInEnth - CoilOutEnth ) ) );
    } else {
        PeakCoilLoad = max( 0.0, ( rhoair * DesVolFlow * ( CoilInEnth - CoilOutEnth ) ) );
    }

Myoldmopar:
Sounds good.  I'm not worried about the `DOAToVRF` warning; it makes sense.  To keep our example files clean, we _could_ change climates, and that could get rid of the warning, right?  I'm not saying we need to, just trying to understand the implications of leaving the warning in there.  If I'm right in that thinking, I'd like to know if @mjwitte would prefer to make example file/weather changes to get rid of the warning or not.

mjwitte:
@rraustad @Myoldmopar I'll yield to your judgement here.  We prefer fewer warnings in example files so fixing it is good.  I forget the details of this change.  Is it not possible to eliminate the warning without changing climates?

Myoldmopar:
I'm inferring that changing the climate will remove the warning because we could get rid of the deviation between `StdPressureSeaLevel` and `StdBaroPress`.  If this is correct, I don't mind tossing in some different design days and location info, and updating CI to use a different epw.  Let's let @rraustad verify that my inferring is correct.

mjwitte:
But does that mean VRF will always generate a warning whenever it's modeled in a location that's not at sea level?  That doesn't seem right.  Is this an input problem or does the VRF model need to use the other pressure?

Myoldmopar:
I think it means it will generate a warning when you are in a location that's not at sea level with inputs that put you outside of the saturation curve.  Which it should.  I don't think that it will automatically warn for a non-sealevel climate though.

mjwitte:
So, can the inputs be changed slightly, or is that a bunch of work?

Myoldmopar:
And I'll officially defer to @rraustad instead of inferring/assuming on that one.  I can look into it if needed but I suspect he already knows the answer.

rraustad:
It's an air flow vs capacity issue. Since I can't really change zone load I lowered the SAT in the zone sizing object, from 12.5C to 12C. This lowered the zone cooling air flow rate and eliminated the warning.

    Sizing:Zone,
     SupplyAirTemperature,    !- Zone Cooling Design Supply Air Temperature Input Method
     12.0,                    !- Zone Cooling Design Supply Air Temperature {C}

rraustad:
Similar for RefBldgSmallOfficeNew2004_Chicago, lowered SAT in 2 places, from 14C to 13.9C. So how to move forward? It would look odd to change specific zones for one SAT while other zones used a different input.

    Sizing:Zone,
     Perimeter_ZN_1,          !- Zone or ZoneList Name
     SupplyAirTemperature,    !- Zone Cooling Design Supply Air Temperature Input Method
     13.9000,                 !- Zone Cooling Design Supply Air Temperature {C}

    Sizing:Zone,
     Perimeter_ZN_3,          !- Zone or ZoneList Name
     SupplyAirTemperature,    !- Zone Cooling Design Supply Air Temperature Input Method
     13.9000,                 !- Zone Cooling Design Supply Air Temperature {C}

mjwitte:
Not odd if some zones have different equipment which would normally deliver a different supply temp.

Myoldmopar:
Yeah, I say proceed with the updated temps as you've done.  The only better option I see is adjusting the SAT of all the zones down to 13.9, since I assume that wouldn't cause any warnings since you are moving the SAT down across the board.  If that would help you sleep better at night @rraustad feel free to make that change.  Otherwise this can go in methinks.

rraustad:
This one can go. I looked at _SmallOffice_Dulles to see why the convergence failed but it passes here in 6 warmup days. 

LOCATION,Washington Dc Dulles IntL Ar,VA,USA,TMY3

I thought it had to do with CoolReheat being used for the air loops but can't investigate that further.

Issue #6043, opened 2017-03-06
Components should all use the final zone sizing array


rraustad:
Issue overview
--------------
Searching the code for CalcFinalZoneSizing shows multiple occurrences in equipment model sizing routines. This is an intermediate array which is copied to FinalZoneSizing at the end of the sizing calcs. Further updates to this array occurs after these arrays are set equal. So any component model using CalcFinalZoneSizing is not using the most updated data.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
The question here is which sizing array to use for equipment models.

The eio reports the design load and user specified loads:

    ! <Zone Sizing Information>, Zone Name, Load Type, Calc Des Load {W}, User Des Load {W}, Calc Des Air Flow Rate {m3/s}, User Des Air Flow Rate {m3/s}, Design Day Name, Date/Time of Peak, Temperature at Peak {C}, Humidity Ratio at Peak {kgWater/kgDryAir}, Floor Area {m2}, # Occupants, Calc Outdoor Air Flow Rate {m3/s}, Calc DOAS Heat Addition Rate {W}
    Zone Sizing Information, SPACE1-1, Cooling, 2462.87170, 3201.73321, 0.20717, 0.26932, CHICAGO ANN CLG 1% CONDNS DB=>MWB, 7/21 16:00:00, 30.85800, 1.45947E-002, 99.16000, 11.00000, 0.10384, 0.00000
    Zone Sizing Information, SPACE1-1, Heating, 2637.54214, 3428.80478, 7.66703E-002, 0.10384, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 8.38927E-004, 99.16000, 11.00000, 0.10384, 0.00000

It's easy to see where these data come from here:

    ReportZoneSizing( FinalZoneSizing( CtrlZoneNum ).ZoneName, "Cooling", CalcFinalZoneSizing( CtrlZoneNum ).DesCoolLoad,
[...]

PR #6041, opened 2017-03-03
Terminal unit reheat coils size to zero when zone has no design heating load #6006

rraustad:
Pull request overview
---------------------
Reheat coils can size to 0 capacity when zone heating load = 0. Update sizing routine to include some form of sizing when zone loads do not exist. This corrects issue #6006. Some differences are expected in existing input files.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [x] Code style (parentheses padding, variable names)
 - [x] Functional code review (it has to work!)
 - [x] If defect, results of running current develop vs this branch should exhibit the fix
 - [x] CI status: all green or justified
 - [x] Performance: CI Linux results include performance check -- verify this
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
   - [x] If any virtual classes, ensure virtual destructor included, other things
 - n/a IDD changes:
 - n/a If new idf included, locally check the err file and other outputs
 - n/a Documentation changes in place
 - n/a Changed docs build successfully
 - n/a ExpandObjects changes?
 - n/a If output changes, including tabular output structure, add to output rules file for interfaces

Issue #6006, opened 2017-02-10
Reheat coils size to zero when zone has no design heating load - because ZoneTempAtHeatPeak is zero


mjwitte:
Issue overview
--------------
User complained that files moved up from v8.0 to v8.6 have much smaller reheat coil sizes and many unmet heating hours.  Running the file in the debugger reveals that ZoneDesTemp = CalcFinalZoneSizing( CurZoneEqNum ).ZoneTempAtHeatPeak = 0.0 for a zone which doesn't log a design heating load.  This breaks the reheat load adjustement in SingleDuct::SizeSys resulting in a negative DesCoilLoad which leads to a zero size reheat coil.
```
							CoilInTemp = TermUnitFinalZoneSizing( CurZoneEqNum ).DesHeatCoilInTempTU;
							DesMassFlow = StdRhoAir * TermUnitSizing( CurZoneEqNum ).AirVolFlow;
							DesZoneHeatLoad = CalcFinalZoneSizing( CurZoneEqNum ).DesHeatLoad * CalcFinalZoneSizing( CurZoneEqNum ).HeatSizingFactor;
							ZoneDesTemp = CalcFinalZoneSizing( CurZoneEqNum ).ZoneTempAtHeatPeak;
							ZoneDesHumRat = CalcFinalZoneSizing( CurZoneEqNum ).ZoneHumRatAtHeatPeak;
							// the coil load is the zone design heating load plus (or minus!) the reheat load
							DesCoilLoad = DesZoneHeatLoad + PsyCpAirFnWTdb( ZoneDesHumRat, 0.5 * ( CoilInTemp + ZoneDesTemp ) ) * DesMassFlow * ( ZoneDesTemp - CoilInTemp );
```

Also, the sizing outputs should report "Design Size Reheat Coil Sizing **Outlet** Air Temperature [C]"

### Details
Some additional details for this issue (if relevant):
 - Platform - Any
 - Version of EnergyPlus v8.6.0
 - Direct e-mail

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
From sizing call, options to solve:
[...]

PR #6052, opened 2017-03-08
Remove all reference to CalcFinalZoneSizing in zone equipment models.

rraustad:
Pull request overview
---------------------
The intermediate sizing array CalcFinalZoneSizing is used intermittently throughout the zone equipment models. References to other sizing data use the FinalZoneSizing array. This change removes the intermediate array from the zone equipment models so that all models use only the FinalZoneSizing array. This should avoid confusion as to what data is used for equipment sizing. This addresses issue #6043. 

Internal use only, Do Not Publish.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - n/a IDD changes:
 - n/a Changed docs build successfully
 - n/a ExpandObjects changes?


rraustad:
Well I wasn't expecting this. Will have to check the unit tests. The one eio failure is for:

    ZoneHVAC:LowTemperatureRadiant:Electric,
     Zone 1 Radiant Floor,    !- Name
     RadiantPanelAvailSched,  !- Availability Schedule Name
     West Zone,               !- Zone Name
     Zn001:Flr001,            !- Surface Name or Radiant Surface Group Name
     HeatingDesignCapacity,   !- Heating Design Capacity Method
     autosize,                !- Heating Design Capacity {W}

So how does an autosized field come out perfectly at 10000.00000 ??

    - Component Sizing Information, ZoneHVAC:LowTemperatureRadiant:Electric, ZONE 1 RADIANT FLOOR, Design Size Heating Design Capacity [W], 10000.00000
    + Component Sizing Information, ZoneHVAC:LowTemperatureRadiant:Electric, ZONE 1 RADIANT FLOOR, Design Size Heating Design Capacity [W], 17768.37980

src/EnergyPlus/LowTempRadiantSystem.cc
```diff
@@ -2078,8 +2078,8 @@ namespace LowTempRadiantSystem {
 							if ( ZoneSizingRunDone ) {
 								CheckZoneSizing( CompType, CompName );
 								SizingMethod = AutoCalculateSizing;
-								DataConstantUsedForSizing = CalcFinalZoneSizing( CurZoneEqNum ).DesHeatLoad;
-								DataFractionUsedForSizing = CalcFinalZoneSizing( CurZoneEqNum ).HeatSizingFactor;
+								DataConstantUsedForSizing = FinalZoneSizing( CurZoneEqNum ).NonAirSysDesHeatLoad;
```
rraustad, line 2081: These should be DataFractionUsedForSizing = 1.0 now

rraustad:
Looks like it is exactly 10000.0. Setting DataFractionUsedForSizing = 1.0 gets it back to where it was.

  Sizing:Zone,
    1.7768379797679387,      !- Zone Heating Sizing Factor

rraustad:
Looks good up to this point. The final change is going to cause diff's (same as or worse than #6041). I'd rather wait until #6006 is merged in via #6041 to minimize or eliminate diff's. OR I can do it now to see the diff's and do it again after #6041 is merged to see if they go away.

rraustad:
This is the only remaining zone equipment model that accesses CalcFinalZoneEquipment. There are still 2 unit tests using this array (OK since we still need unit tests on this array), and I see lot's of lines in OutputReportTabular (won't touch this for this effort).

 SingleDuct.cc - these will cause diff's, but hoping for minimum or no diff's after #6041 is merged.

    (2362): ZoneDesTemp = CalcFinalZoneSizing( CurZoneEqNum ).ZoneTempAtHeatPeak;
    (2363): ZoneDesHumRat = CalcFinalZoneSizing( CurZoneEqNum ).ZoneHumRatAtHeatPeak;
    (2434): ZoneDesTemp = CalcFinalZoneSizing( CurZoneEqNum ).ZoneTempAtHeatPeak;
    (2435): ZoneDesHumRat = CalcFinalZoneSizing( CurZoneEqNum ).ZoneHumRatAtHeatPeak;

mjwitte:
@rraustad Now that #6041 is merged, you can make the final change here.

src/EnergyPlus/DataSizing.hh
```diff
@@ -451,14 +451,17 @@ namespace DataSizing {
 		bool EMSOverrideDesCoolMassOn; // true if EMS is acting on this structure
 		Real64 EMSValueDesCoolMassFlow; // Value EMS directing to use for Design Cooling air mass flow [kg/s]
 		Real64 DesHeatLoad; // zone design heating load [W]
+		Real64 NonAirSysDesHeatLoad; // same as zone design heating load without special multipliers from Sizing:Zone object [W]
```
mjwitte: I don't think this is the correct description.  This *should* include any sizing factor fro sizing:zone and/or sizing:parameters.  What it doesn't include is any scaling dues to airflow rates which could come from user-specified flow rates in sizing:zone or sizing:system, and/or sizing on ventilation requirement.  Suggest:
```
Real64 DesHeatLoad; // zone design heating load including sizing factor and scaled to match airflow sizing
Real64 NonAirSysDesHeatLoad; // base zone design heating load including sizing factor
```
Similar for cooling.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -2609,10 +2610,12 @@ namespace ZoneEquipmentManager {
 				z.OutHumRatAtHeatPeak = c.OutHumRatAtHeatPeak;
 				z.TimeStepNumAtHeatMax = c.TimeStepNumAtHeatMax;
 				z.DesHeatVolFlow = c.DesHeatVolFlow;
+				z.NonAirSysDesHeatVolFlow = c.DesHeatVolFlow;
 				z.DesHeatCoilInTemp = c.DesHeatCoilInTemp;
 				z.DesHeatCoilInHumRat = c.DesHeatCoilInHumRat;
 
 				z.DesCoolLoad = c.DesCoolLoad;
+				z.NonAirSysDesCoolLoad = c.DesCoolLoad;
```
mjwitte, line 2618: Should we be adding a NonAirSysDesCoolVolFlow just for completeness?  I'm guessing you didn't add it because it's not used presently by any components

rraustad: Correct, I didn't add the CoolVol variable because it would not be used anywhere. I'll circle back when I see what the CI reports to fix the description. I'm on the fence about adding in the unused variable.

mjwitte: That's fine.  How about a comment in the header file where the CoolVol would be declared that says it isn't currently needed?

mjwitte:
@rraustad CI came back clean.  If you can push those comment changes up, then this can drop in.

rraustad:
Went ahead and added in NonAirSysDesCoolVolFlow since it only took a few lines. Now developers won't have to search to find where to add it.

Issue #5501, opened 2016-02-22
User file not calculating zone volumes correctly


mjwitte:
Helpdesk ticket 11140
User expected similar loads from four similar zones, but they varied signficantly.  The four zones have similar floor area and the same ceiling height but greatly different zone volumes:

![image](https://cloud.githubusercontent.com/assets/4991075/13236166/5aa0737a-d988-11e5-8d5e-975485bdd074.png)

The difference in volumes cause the infiltration and ventilation flows to be different, because both were specified in ACH.

Workaround is to put hard values for Volume in the Zone objects.

mitchute:
@Myoldmopar I had a minute to look at this.

The volume is being calculated incorrectly by [CalcPolyhedronVolume](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SurfaceGeometry.cc#L8110). 

The building from the user's file is shown here.
![capture](https://cloud.githubusercontent.com/assets/2985979/13475340/df1fec06-e087-11e5-8490-ea984c3b50cb.PNG)

I'm not seeing anything that is immediately apparent that is wrong. Are there some assumptions, like the zone be convex, for `CalcPolyhedronVolume` to give accurate results? Either way, at the very least, I don't see how the the minimum volume limit is being enforced. It looks to me like `CalcVolume` always comes out of `CalcPolyhedronVolume` with a non-zero value. Cleaning up [this if statement](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SurfaceGeometry.cc#L8124) could at least clear up that problem. As far as `CalcPolyhedronVolume` over predicting volumes, I am going to need to dig in and understand the background on when the method `CalcPolyhedronVolume` uses is valid.

JasonGlazer:
Updated defect file to EnergyPlus 8.6 in DevSupport repo, defect still present.

Myoldmopar:
Closed via #6056

PR #6056, opened 2017-03-09
Add fixes to zone volume calcs including for non-convex zones

JasonGlazer:
Address zone volume calculations related to non-convex zones to address #5501.
Fix example file geometry issues to address #6161.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
The CI shows lots of files that have different calculated zone volumes now.

JasonGlazer:
@mitchute and @mjwitte the 5501 test file was missing some walls in four of the five zones and that meant that the zones were not full enclosed, which means the algorithm for calculating the volume was not going to work. After I added the walls, it calculated the proper volume in both the 8.6 release and with my current changes. You can see the updated file in the dev support directory:

https://github.com/NREL/EnergyPlusDevSupport/tree/master/DefectFiles/5000s/5501

To me, this implies that the algorithm should not be used unless the zone is fully enclosed. I don't think that we have anything that checks if a zone is fully enclosed in the code, do you think we should add one?

JasonGlazer:
Trying to understand the reason for the diffs shown in the CI, I have removed all the changes except for the lines:

```
			if ( CalcVolume > 0.0) {
				TempVolume = CalcVolume;
			} else {
				TempVolume = MinimumVolume;
			}

```
to

```
          		if ( CalcVolume > MinimumVolume ) {
				TempVolume = CalcVolume;
			} else {
				TempVolume = MinimumVolume;
			}

```


This is clearly a fix that needs to be made but may be responsible for most of the diffs that the CI shows.

JasonGlazer:
@mitchute @Myoldmopar @mjwitte 

The CalcVolume > MinimumVolume change did seem to cause almost all of the diffs but I am not sure that it is a good change. Even though it seem that the intention of the code around this would make this change reasonable, the computation of the MinimumVolume to be the FloorArea * CeilingHeight makes all attic-type zones have the wrong volume calculation since for attics it should usually be half that. 

In general, it seems like a completely different approach should be used that does not use a "minimum volume" metaphor. If the zone is fully enclosed (something we don't currently test for but I can add a function to test) then it uses the CalcPolyhedronVolume method that currently works. If it is not fully enclosed, then if the ceiling and floor are the same area assume it is a prism and use floor area time ceiling height. If it does not have both a floor and ceiling defined and all the angles are 90 degree, use the floor or ceiling area times the ceiling height. If non of these approaches can be used, throw a severe error. If the user provides a volume, than this can all be ignored but a warning is provided if it is more than 5% different than the calculated value.

What do you think?

JasonGlazer:
One thing to add, the impact of the zone volume computation cascades into lots of other results, so that is why I'm suggesting a severe error rather than a warning.

mjwitte:
Regarding attics, what is the ceiling height?  Seems to me that it should be the average height of the roof surfaces which should give the correct volume when multiplied by the floor area.

JasonGlazer:
I don't think the way the averaging is done for ceiling height that it would work for an attic under a hip roof.

JasonGlazer:
The previous commits do not pass all unit tests. Some unit tests are causing errors in the zone volume calculations even when they are focused on HVAC issues just because the developer included full IDF descriptions of the zones. 

This was pushed to see if the results of the CI tests show a consistent problem that can be fixed.

nrel-bot-3:

nrel-bot-3:
@JasonGlazer @lgentile it has been 15 days since this pull request was last updated.

JasonGlazer:
@mjwitte I ended up adding a new warning message but only when DisplayExtraWarnings is used that looks like:

```
   ** Warning ** The Zone="ZONE ONE" is not fully enclosed. To be fully enclosed, each edge of a surface must also be an edge on one other surface.
   **   ~~~   **   The zone volume was calculated using the floor area times ceiling height method where the floor and ceiling are the same except for the z-coordinates.
   **   ~~~   **   The surface    "ZN001:WALL002" has an edge that is either not an edge on another surface or is an edge on three or more surfaces: 
   **   ~~~   **     Vertex start { 15.2400, 15.2400, 4.5720}
   **   ~~~   **     Vertex end   { 15.2400, 0.0000, 3.0000}
   **   ~~~   **   The surface    "ZN001:CEILING" has an edge that is either not an edge on another surface or is an edge on three or more surfaces: 
   **   ~~~   **     Vertex start { 15.2400, 0.0000, 4.5720}
   **   ~~~   **     Vertex end   { 15.2400, 15.2400, 4.5720}
```
The line with "The zone volume was calculated..." changes depending on the algorithm that was used to estimate the zone volume. Does this approach make sense?

mjwitte:
@JasonGlazer This is good.  Please add the routine name: to the beginning of the first line.  

Also, if you want, you could have a summary warning that always appears with something like "Warning there 3 zone which are not fully enclosed.  Use Output:Diagnostics:DisplayExtrawarnings for more details."  This would follow the pattern of 
```
   ** Warning ** CheckUsedConstructions: There are 5 nominally unused constructions in input.
   **   ~~~   ** For explicit details on each unused construction, use Output:Diagnostics,DisplayExtraWarnings;
```

src/EnergyPlus/DataVectorTypes.hh
```diff
@@ -88,6 +88,13 @@ namespace DataVectorTypes {
 	typedef  ObjexxFCL::Vector3< Real64 >  Vector;
 	typedef  ObjexxFCL::Vector2< Real64 >  Vector_2d;
 
+	struct Vector2dCount: Vector_2d {
+		int count;
+		Vector2dCount():
+			count( 0 )
+		{}
+	};
+
```
Myoldmopar, line 97: So, is this entire struct just an integer?  If you want a different type name for some reason, you could typedef an int to something else.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8737,6 +8803,605 @@ namespace SurfaceGeometry {
...
+			int firstSurfNum;
+			EdgeByPts( ):
+				start( 0 ),
+				end( 0 ),
+				count( 0 ),
+				firstSurfNum( 0 )
+			{}
+		};
+		std::vector<EdgeByPts> uniqueEdges;
+		uniqueEdges.reserve( zonePoly.NumSurfaceFaces * 6 );
```
Myoldmopar, line 8906: :+1:

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8737,6 +8803,605 @@ namespace SurfaceGeometry {
...
+	// test if two points on a plane are in the same position based on a small tolerance (based on Vector2dCount comparison)
+	bool
+		isAlmostEqual2dPt(
+			DataVectorTypes::Vector2dCount v1,
+			DataVectorTypes::Vector2dCount v2
+		)
+	{
+		// J. Glazer - March 2017
+
+		Real64 tol = 0.0254; //  2.54 cm = 1 inch
```
Myoldmopar: I'm curious how you chose 1 inch on the 2d points, and 0.5 inch on the 3d function.  I don't really care which you use, just curious about it.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8737,6 +8803,605 @@ namespace SurfaceGeometry {
...
+		Real64 tol = 0.0254; //  2.54 cm = 1 inch 
+		return ( ( abs( v1.x - v2.x ) < tol ) && ( abs( v1.y - v2.y ) < tol ) );
+	}
+
+	// test if two points on a plane are in the same position based on a small tolerance (based on Vector2dCount comparison)
+	bool
+		isAlmostEqual2dPt(
+			DataVectorTypes::Vector2dCount v1,
+			DataVectorTypes::Vector2dCount v2
+		)
```
Myoldmopar: Some indentation went wonky

src/EnergyPlus/Vectors.cc
```diff
@@ -815,10 +815,9 @@ namespace Vectors {
 
 	}
 
-	void
+	Real64
```
Myoldmopar, line 818: :+1:

Myoldmopar (changes requested):
This is gobs of effort.  Thanks @JasonGlazer.  There is one whitespace thing that should probably get fixed (hence my "reject" here); I can do that fix.  I had a comment about the `Vector2dCount` struct that I may investigate myself also.  I would really love if @mbadams5 could give it a quick look over at your stuff from a performance view, as performance went down very slightly, and he may be able to clue in to an issue. 
 Or it may just be that the added expense in doing these calculations is something we'll just have to eat for this PR.  Overall this looks really good to me.

src/EnergyPlus/DataVectorTypes.hh
JasonGlazer: No, it is inheriting from vector_2d so it is more than just a int.

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: They were both 1inch and had to fix the 3d function to 0.5 inch because it was being used and failing on a wall that was supposed to be 1 inch tall (a weird edge to an attic). I should go back and change the 2d compare to 0.5 inch also

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: ok

JasonGlazer:
@Myoldmopar I addressed the issue from the review and merged with latest develop. I want to check the CI results one last time before we merge this in.

JasonGlazer:
A summary of the CI fails and warnings:

 - MacOS – 47 fails, 9 warnings
 - Ubuntu – 48 fails, 9 warnings
 - Windows i386 – 47 fails, 9 warnings
 - Windows Win64 - 47 fails, 9 warnings

Based on this examined the Ubuntu fails.

Fails from Ubuntu

 - 1ZoneUncontrolled3SurfaceZone - This is a test file specifically for a non-enclosed zone. OK
 - 1ZoneUncontrolled_DD2009 – Fixed floor at ceiling height previously. New volume consistent. OK
 - 1ZoneUncontrolled_DDChanges – Added floor to fully enclose. Removed warnings. New volume consistent. OK.
 - 5ZoneSteamBaseboard - Fixed several surfaces. Added several to fully enclose. New volume consistent. OK.
 - AirflowNetwork_Multizone_HorizontalOpening – Fixed surface in wrong zone to fully enclose. New volume consistent. OK.
 - CmplxGlz_SmOff_IntExtShading - Added missing interior walls and basement walls. Changed to autocalculating height and volume. New volume consistent. OK.
 - CooltowerSimpleTest – Fixed several zones and coordinates. New volume consistent. OK
 - CooltowerSimpleTestwithVentilation - Fixed several zones and coordinates. New volume consistent. OK
 - DElight-Detailed-Comparison - Fixed a few coordinates that were probably entered wrong. Removed unused constructions. OK.
 - DElightCFSLightShelf - Fixed a few coordinates that were probably entered wrong. Removed unused constructions. OK.
 - DElightCFSWindow - Fixed a few coordinates that were probably entered wrong. Removed unused constructions. OK.
 - DaylightingDeviceShelf - Not fully enclosed due to interior shelf surface. No change made OK.
 - GeneratorswithPV - Some warnings fixed. Added several surfaces to fully enclose zone. New volume consistent. OK
 - GeometryTest – Fixed some coordinates to fully enclose zone. OK
 - GroundTempOSCCompactSched - Fixed several surfaces to fully enclose zone. Changed to using upper left corner starting point. OK
 - HP_wICSSolarCollector - Fixed wall attached to wrong zone, missing another wall, roof angle wrong. New volume consistent. OK
 - HeatPumpSecondaryCoil – Fixed an outside wall was in wrong location in east zone and was directly on top of an interior wall.  New volume consistent. OK
 - LBuilding-G000 – Added missing internal walls. Removed unused constructions. New volume consistent. OK
 - LBuilding-G090 – Added missing internal walls. Removed unused constructions. New volume consistent. OK
 - LBuilding-G180 – Added missing internal walls. Removed unused constructions. New volume consistent. OK
 - LBuilding-G270 – Added missing internal walls. Removed unused constructions. New volume consistent. OK
 - LBuildingAppGRotPar – Added missing internal walls. Removed unused constructions. New volume consistent. OK
 - MovableExtInsulationSimple – Added missing roof and floor. New volume consistent. OK
 - MovableIntInsulationLights – Added missing roof and floor. New volume consistent. OK
 - MovableIntInsulationLightsLowE – Added missing roof and floor. New volume consistent. OK
 - MovableIntInsulationSimple – Added missing roof and floor. New volume consistent. OK
 - PipingSystem_Underground_FHX - Fixed roof and intermediate floor which were not right size. Also fixed other wall coordinates not starting at upper left and counter clockwise. New volume consistent. OK
 - PipingSystem_Underground_FHX – this test seems to be getting repeated.
 - QTFtest - Added missing roof and floor. New volume consistent. OK
 - RadLoTempHydrMulti10 – Added missing floor and roof as well as fixing some coordinates to fully enclose zone. OK
 - RefBldgWarehouseNew2004_Chicago – Changed several walls to interior instead of exterior to indicate as fully enclosed. New volume consistent. OK
 - RefrigeratedWarehouse – Removed an extra wall that was already being mirrored between zones. New volume consistent. OK
 - SmOffPSZ – Added missing interior walls and basement walls. Changed to autocalculating height and volume. New volume consistent. OK
 - SmOffPSZ-MultiModeDX – Added missing interior walls and basement walls. Changed to autocalculating height and volume. New volume consistent. OK
 - SmOffPSZ_OnOffStagedControl – Added missing interior walls and basement walls. Changed to autocalculating height and volume. New volume consistent. OK
 - SolarShadingTest – Fixed incorrect coordinates to fully enclose zone. New volume consistent. OK
 - SolarShadingTest_SQL – Fixed incorrect coordinates to fully enclose zone. New volume consistent. OK
 - SteamSystemAutoSize – Fixed several zone walls that were missing and were parts of surfaces from other zones.  New volume consistent. OK
 - SupplyPlenumVAV – Fixed the order of corners for some surfaces to fully enclose zone. OK
 - TransparentInsulationSimple - Added missing roof and floor. New volume consistent. OK
 - VentilationSimpleTest - Fixed several zones and coordinates. New volume consistent. OK
 - ZoneCoupledGroundHTBasement – Fixed incorrect order of vertices so that now fully enclosed zone. No change in calculated volume. OK
 - ZoneCoupledKivaBasement – Fixed incorrect order of vertices so that now fully enclosed zone. No change in calculated volume. OK
 - _CTFTestsPart1 – Fixed that 10 zones all exactly overlapping and none have full floors or ceilings.  Also fixed coordinates on walls. New volume consistent. OK
 - _CTFTestsPart2 – Fixed that 9 zones all exactly overlapping and none have full floors or ceilings.  Also fixed coordinates on walls. New volume consistent. OK
 - _SingleZoneTestCTFTwoDD – Raised roof which was at floor level to make fully enclosed zone. No change in calculated volume. OK
 - _SingleZoneTestCondFDTwoDD – Raised roof which was at floor level to make fully enclosed zone. No change in calculated volume. OK
 - _SmallOffice_Dulles - Fixed unenclosed geometry warning by moving ceiling and roof to line up with top of the walls. No change in calculated volume. OK

Warnings from Ubuntu:

 - AdultEducationCenter – Not enclosed warning is issued because floor and roof of irregularly shaped zone is simply a square that does not meet all edges. The zone volume does not change. Not going to try to fix the complicated roof needed. OK
 - EarthTubeSimpleTest - Fixed several zones and coordinates. New volume consistent. OK
 - RefBldgHospitalNew2004_Chicago – Fixed some coordinates that had digits reversed to fully enclose zones. New volume consistent. OK
 - UserDefinedRoomAirPatterns - New unenclosed warning and but is an odd case that is not caught be current algorithm for an enclosed zone even though it is enclosed OK
 - UserInputViewFactorFile-LshapedZone – A new not fully enclosed warning message shows. This test file is looking at two identical L shaped zones that both are missing a wall and comparing the user input view factors with calculated view factors. I think if I add a wall to fix these they will mess up the view factors so I am not doing anything. OK
 - _1Zone_Heavy_AdiabaticX2 - A new not fully enclosed warning message shows.  No volume changes just warmup convergence. The file is intended to have an adiabatic interior wall that messes up the full enclosed since it is within the zone and touching the edges. OK
 - _1Zone_Heavy_SelfRef - A new not fully enclosed warning message shows.  No volume changes just warmup convergence. The file is intended to have an adiabatic interior wall that messes up the full enclosed since it is within the zone and touching the edges. OK
 - _FuelCellTest200 - A new not fully enclosed warning message shows.  The geometry for this example file consists of a single wall, weird but that is the intention. OK
 - _MicroCHPTest301 - A new not fully enclosed warning message shows.  The geometry for this example file consists of a single wall, weird but that is the intention. OK

Everything looks ok so this is ready for final review and (hopefully) merging.

src/EnergyPlus/DataVectorTypes.hh
Myoldmopar: Someone must've hacked my account when making that comment, cause I'd never miss something like that.  

:smirk: 

Got it, thanks!

Myoldmopar (approved):

Myoldmopar:
Yeah, the thorough level is off the chart.  Thanks for this effort @JasonGlazer, merging this now.

PR #6080, opened 2017-03-23
Prep for 8.7 Release

Myoldmopar:
Known tasks:

- [ ] VS 2015 optimizer issue per @DeadParrot 
- [ ] Known transition issues
- [ ] Change file versions
- [ ] #6081 
...


Myoldmopar:
@DeadParrot the Windows builds are failing miserably.  I was trying to avoid having to pull up my VM to test that compiler flag out.  Do you have any other tips about that compiler flag with the way I added it here?  I know you don't use CMake much, but it's pretty straightforward.  I'm so confused why the compiler flag starts with a dash and not a forward slash on windows...

DeadParrot:
@Myoldmopar the `-` vs `/` shouldn't matter with VC options and MS's [own page](https://blogs.msdn.microsoft.com/vcblog/2016/05/04/new-code-optimizer/) shows it with `-`. The only reason I can think for it causing the build to fail would be if the VC++ 2015 wasn't on Update 3 since the new optimizer only appeared there. If the failures are due to small solution diffs then that is probably OK. If the tests just aren't building but the compiler is on Update 3 then maybe try the hotfix instead of the option.

Myoldmopar:
OK, merging this into develop and getting it out the door.

PR #6083, opened 2017-03-29
Only perform CTF calculations for constructions used by surfaces with CTF heat balance algorithm

nealkruis:
Pull request overview
---------------------
This follows on the recent addition of Kiva to perform ground heat transfer calculations. This pull request accomplishes two objectives:

1. It prevents CTF calculations from being performed unless a construction is used by a surface with the CTF heat balance algorithm.
2. It removes IDD restrictions on material thickness that were specific to CTF limitations. If a material is thicker than 3 meters and used in a construction referenced by a surface with the CTF heat balance algorithm then a warning is issued.

Work completed for:

- Subcontract AEU-7-70112-01: “2017 EnergyPlus Foundation Heat Transfer Algorithm Model Development”

### Work Checklist

### Review Checklist
 - C++ checks:
   - [ ] Argument types


nealkruis:
@mjwitte @EnergyArchmage I'd appreciate your review of this PR. The one failed regression test is related to TDD domes not overriding the default `HeatTransferAlgorithm = HeatTransferModel_CTF` [here](https://github.com/NREL/EnergyPlus/blob/78a111df4adc0ac10b929e2abd28ac9391d46bfa/src/EnergyPlus/SurfaceGeometry.cc#L6385) (originally [here](https://github.com/NREL/EnergyPlusArchive/blame/f01340edc01971faed60853d4c021180ee0fa923/SourceCode/SurfaceGeometry.f90#L6699)). Fixed in this PR [here](https://github.com/NREL/EnergyPlus/pull/6083/files#diff-ff08ad5fe0fccbde744effe7442491b1R6384).

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -530,13 +530,18 @@ namespace HeatBalanceManager {
...
 					CNum = FindItemInList( cAlphaArgs( 4 ), Construct );
+				} else {
+					CNum = FindItemInList( cAlphaArgs( 2 ), Construct );
 				}
 				if ( CNum == 0 ) continue;
 				Construct( CNum ).IsUsed = true;
+				if ( ONum == 4 || ONum == 6 ) {
+					// GroundHeatExchanger:Surface or EnergyManagementSystem:ConstructionIndexVariable
+					// Include all EMS constructions since they can potentially be used by a CTF surface
+					Construct( CNum ).IsUsedCTF = true;
```
EnergyArchmage: The integration test failure in EMSThermochromicWindow is coming from this I think.  The EMS construction actuator can work on windows as well as opaque.  By setting IsUsedCTF to true here, too much code runs in InitConductionTransferFunctions for window constructions and severe errors and fatal.  I suspect that the logic in this PR will have to be revised to retain use of TypeIsWindow along with the new IsUsedCTF.

src/EnergyPlus/ConductionTransferFunctionCalc.cc
```diff
@@ -385,6 +387,10 @@ namespace ConductionTransferFunctionCalc {
 						}
 					}
 				}
+				if ( Material( CurrentLayer ).Thickness > 3.0 ) {
+					ShowWarningError( "InitConductionTransferFunctions: Found Material that is possibly too thick for CTF calculation" );
```
EnergyArchmage: I vaguely remember that the 3 m limit is there because of serious problems in CTF's FD.  I worry about letting it run with just a warning, whereas the previous IDD limit was a hard stop.  I think it would be safer to severe-> fatal here to avoid hard crashes when folks want to model an old castle or cave.

nealkruis:
Thanks @EnergyArchmage. Just getting around to checking the build results now. I've addressed both of your comments.

EnergyArchmage:
@nealkruis it looks like the TDD *pipe* is getting missed.  The TDD pipe is sort of a surface that uses CTF but whose surface geometry is coming from DaylightingDevice:Tubular.  I see you made a correction for the TDD's dome, but the TDD pipe got broken so some of the CI diffs in DaylightingDeviceTubular.idf are a problem.

nealkruis:
@EnergyArchmage I looked into this a bit. The Pipe construction isn't used for conduction (that I can see--please correct me if I'm wrong). The construction is only used to get the specular properties of the interior surface.

EnergyArchmage:
Okay, @nealkruis is correct about the pipe heat transfer to transition zone.  I have reviewed this and all is okay, merging.

PR #6085, opened 2017-03-30
Revert "New error for Output:Table:Monthly and Annual don't have proper order for aggregation types"

Myoldmopar:
Reverts NREL/EnergyPlus#6039

We may not need this at all, but I'm going to go ahead and get things moving just in case.


Myoldmopar:
@JasonGlazer I'm going to merge this in because this will likely be problematic for lots of people.  We can include this fix if we release a bug fix version of 8.7 later, but since I doubt we'll come up with a fix in the next hour or so, I think the best path is to just revert the error back out.

JasonGlazer:
A prudent decision. 

I looked at the code and I haven't yet figured out why it is not working. I think it might be in the GetInput for the monthly reports.

PR #6087, opened 2017-04-03
Enhanced Phase Change Material Simulation

Myoldmopar:
Pull request overview
---------------------
Still work to do here.
 - [x] Update IDD fields with descriptions, defaults, units, etc.
 - [x] Add error handling to input processor
 - [ ] Tentatively add zone level enthalpy outputs
 - [x] Remove any temporary outputs, etc.
 - [ ] Final review


nrel-bot-3:

Myoldmopar:
@mjwitte @JasonGlazer I'm going to create a tag of this, and if you happen to have a chance sometime soon, could you test out the Windows build?  I've been unable to replicate issues that the key user on this PR is seeing.

Myoldmopar:
Addressing review comments from long ago.

@JasonGlazer I removed the second latent heat term, there is now one latent term defining the energy change from solid to melted or vice-versa.  The path to get from one state to the other is different, and allows hysteresis, but this makes much more sense, energy-wise.

@michael-okeefe I worked with Kosny in the past, and he was always interested in seeing the PCM model improved in E+ to capture more physical phenomena, which this does.  This model is indeed a hT curve, so it simply fits within the existing finite difference simulation model as a new way to calculate Cp.  This is implemented as an alternative modeling approach, so that users can input data based on the amount of input data they have:

- No PCM data: just use constant properties
- Some critical parameters: use this new model and it will create an hT curve for you
- Full experimental test data: use the tabular input for PCM data to define the detailed hT curve

This could be expanded in the future, I could see adding in the hysteresis effect to the tabular test data input method, but that is well outside of the scope of this project.

I have received word from the SBV PI that at this point, the code is behaving as expected, and I think I am done.  I'll be looking for a reviewer to look this over before they/I merge it.  I'll ping relevant folks.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -5191,11 +5191,13 @@ CalcHeatBalanceInsideSurf( Optional_int_const ZoneToResimulate ) // if passed in
 	int OtherSideZoneNum; // Zone Number index for other side of an interzone partition HAMT
 	static int WarmupSurfTemp;
 	static int TimeStepInDay( 0 ); // time step number
+	static Array1D< Real64 > SurfaceEnthalpyRead; // ! True if EnthalpyM and EnthalpyH are read
```
mbadams5: Why create a new static function variable? It doesn't seem to cause unit testing issues in this PR but could in the future.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
```diff
@@ -0,0 +1,344 @@
...
+
+namespace EnergyPlus {
+
+namespace HysteresisPhaseChange {
+
+	bool getHysteresisModels( true );
+	int numHysteresisModels = 0;
+	std::vector< HysteresisPhaseChange > hysteresisPhaseChangeModels;
+
+	HysteresisPhaseChange * HysteresisPhaseChange::factory( const std::string objectName ) {
```
mbadams5: Should pass by reference.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
```diff
@@ -0,0 +1,344 @@
...
+		// because of the passive linking between materials and material property objects,
+		// we don't know ahead of time for sure whether we will have a material property
+		// so we can't return fatal here if it isn't found, just leave it null
+		return nullptr;
+	}
+
+	Real64 HysteresisPhaseChange::getEnthalpy( Real64 T, Real64 Tc, Real64 tau1, Real64 tau2 ) {
+		// Looks up the enthalpy on the characteristic curve defined by the parameters Tc, tau1, and tau2,
+		// and the position on that curve defined by T.
+		Real64 eta1 = ( this->totalLatentHeat / 2 ) * exp( -2 * abs( T - Tc ) / tau1 );
```
mbadams5: Can `tau1` or `tau2` be zero? If so, might need to protect against divide by zero.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
```diff
@@ -0,0 +1,344 @@
...
+			this->enthOld = this->getEnthalpy(prevTempTD, Tc, Tau1, Tau2 );
+			this->enthNew = this->getEnthalpy(updatedTempTDT, Tc, Tau1, Tau2 );
+		} else {
+			if ( prevPhaseChangeState == PhaseChangeStates::FREEZING && phaseChangeState == PhaseChangeStates::TRANSITION ) {
+				this->enthRev = this->getEnthalpy(phaseChangeTempReverse, this->peakTempFreezing, this->deltaTempFreezingLow, this->deltaTempFreezingHigh );
+				this->enthNew = ( this->specHeatTransition * updatedTempTDT ) + ( this->enthOld - ( this->specHeatTransition * prevTempTD) );
+				this->enthalpyM = this->getEnthalpy(updatedTempTDT, this->peakTempMelting, this->deltaTempMeltingLow, this->deltaTempMeltingHigh );
+				this->enthalpyF = this->getEnthalpy(updatedTempTDT, this->peakTempFreezing, this->deltaTempFreezingLow, this->deltaTempFreezingHigh );
+				if ( this->enthNew < this->enthRev && this->enthNew >= this->enthalpyF && updatedTempTDT <= prevTempTD) {
+					phaseChangeState = 1;
```
mbadams5: Should probably be consistent and use `PhaseChangeStates::FREEZING`. The same is true for all other instances where you directly use the number value instead of the "enum".

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
```diff
@@ -0,0 +1,344 @@
...
+		//	Tau1                ! Width of Melting Zone low
+		//	Tau2                ! Width of Melting Zone high
+		//	EnthalpyOld         ! Previos Timestep Nodal Enthalpy
+		//	EnthalpyNew         ! Current Timestep Nodal Enthalpy
+
+        Real64 Cp1 = this->specificHeatSolid;
+        Real64 Cp2 = this->specificHeatLiquid;
+        Real64 T = temperatureCurrent;
+
+        Real64 DEta1 = - ( this->totalLatentHeat * ( T - criticalTemperature ) * exp( -2 * abs( T - criticalTemperature ) / tau1 ) ) / ( tau1 * abs( T - criticalTemperature ) );
```
mbadams5: Same comment about `tau1` and `tau2` and divide by zero. Also can `abs( T - criticalTemperature )` ever be zero? It appears so due to ` } else if ( T == criticalTemperature ) {` below...

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.hh
```diff
@@ -0,0 +1,125 @@
...
+		~HysteresisPhaseChange()
+		{}
+
+		static
+		HysteresisPhaseChange *
+		factory( const std::string objectName );
+
+		Real64 getEnthalpy( Real64 T, Real64 Tc, Real64 tau1, Real64 tau2 );
+
+		Real64 getCurrentSpecificHeat( Real64, Real64, Real64, int, int & );
```
mbadams5: I know you can do this, but I personally don't particularly like it without variable names.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
```diff
@@ -0,0 +1,344 @@
...
+			// the input processor validates the numeric inputs based on the IDD definition
+			// still validate the name to make sure there aren't any duplicates or blanks
+			// blanks are easy: fatal if blank
+			if ( DataIPShortCuts::lAlphaFieldBlanks[0] ) {
+				ShowFatalError( "Invalid input for " + DataIPShortCuts::cCurrentModuleObject + " object: Name cannot be blank" );
+			}
+
+			// we just need to loop over the existing vector elements to check for duplicates since we haven't add this one yet
+			for ( auto & existingHysteresisModel : hysteresisPhaseChangeModels ) {
+				if ( DataIPShortCuts::cAlphaArgs( 1 ) == existingHysteresisModel.name ) {
```
mbadams5: Adding to then comparing to a `std::set`/`std::unordered_set` or something would be better here. That would change it from `O(n^2)` to `O(n log n)`/`O(n)` but i'm fine with this as is for now.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.hh
```diff
@@ -0,0 +1,125 @@
...
+// POSSIBILITY OF SUCH DAMAGE.
+
+#ifndef HYSTERESISPHASECHANGE_HH_INCLUDED
+#define HYSTERESISPHASECHANGE_HH_INCLUDED
+
+#include <vector>
+
+#include <EnergyPlus.hh>
+#include <DataGlobals.hh>
+#include <OutputProcessor.hh>
```
mbadams5: Why do you need `OutputProcessor.hh` and `DataGlobals.hh` includes in the header file? Could these move to the implementation file? If includes are not needed in the header file it is better to only include them in the implementation file.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
```diff
@@ -0,0 +1,677 @@
...
+{
+	Real64 previousTemperature;
+	Real64 updatedTemperature;
+	Real64 temperatureReverse;
+	int previousPhaseChangeState;
+	int expectedUpdatedPhaseChangeState;
+	Real64 expectedSpecificHeat;
+
+	GetSpecHeatArgs(Real64 _previousTemperature, Real64 _updatedTemperature, Real64 _temperatureReverse,
+					int _previousPhaseChangeState, int _expectedUpdatedPhaseChangeState, Real64 _expectedSpecificHeat) :
```
mbadams5: Looks like formatting issue.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
```diff
@@ -0,0 +1,677 @@
...
+	code_file << "}\n";
+	code_file.close();
+}
+
+TEST_F(HysteresisTest, GenerateTestData)
+{
+	//generate_test_data(this->ModelA, "StraightUpCurve", 14, 26, 0.5, 2);
+	//generate_test_data(this->ModelA, "StraightDownCurve", 30, 15, -0.5, -2);
+	//generate_test_data_two_directions(this->ModelA, "IntoMeltingAndBackDown", 14, 0.5, 20, -0.5, 14, 2);
+	generate_test_data_two_directions(this->ModelA, "CompletelyThroughMeltingAndBackDown", 14, 0.5, 30, -0.5, 14, 2);
```
mbadams5: Do these generate calls still need to be generated by the unit test? Also do they still need to be included with this PR?

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
```diff
@@ -0,0 +1,677 @@
...
+#include <gtest/gtest.h>
+#include <EnergyPlus/PhaseChangeModeling/HysteresisModel.hh>
+
+using namespace EnergyPlus;
+
+// A new one of these is created for each test
+class HysteresisTest : public testing::Test
+{
+public:
+	HysteresisPhaseChange::HysteresisPhaseChange ModelA;
```
mbadams5, line 59: I assume this is getting generated new for each unit test?

mbadams5 (changes requested):

src/EnergyPlus/HeatBalanceSurfaceManager.cc
Myoldmopar: Moving to namespace variable and adding to `clear_state`.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
Myoldmopar: Doing it!

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
Myoldmopar: tau1 and tau2 will always be assigned a meaningful value from the input object.  The input object includes \minimum>0 so these values should be guaranteed to always have meaningful, >0, values.  I am cleaning up the logic at the top of this function while addressing this.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
Myoldmopar: :+1:

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
Myoldmopar: Reworked that function so that the intermediate variables are only calculated when valid, good to go here now.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.cc
Myoldmopar: OK, I'll pass on making changes for now.  Even for PCM heavy files, this will only be looping over like 5-10 objects, once during setup.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.hh
Myoldmopar: Turns out I didn't need them, removed.

src/EnergyPlus/PhaseChangeModeling/HysteresisModel.hh
Myoldmopar: I don't like it either.  I was doing it early on when the argument list was fluid and changing a lot, and forgot to go back and add them in.  I fixed it.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
Myoldmopar: On each unit test, it should be calling setUp(), right?  And setUp should make sure it is fully populated no matter which unit test is calling it.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
Myoldmopar: Took out the wrapping.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
Myoldmopar: Accidental leftover, took all that stuff out.

Myoldmopar:
@mbadams5 All review comments addressed via 1686368.  Unless you have a remaining issue with the unit test `setUp` stuff, this is complete.  If you are satisfied and want to merge it sometime, feel free, otherwise I'll do it after CI comes back happy.

Myoldmopar:
And thanks for the thorough review!

mbadams5 (approved):

Myoldmopar:
image
:+1: :+1: :+1:

Issue #6088, opened 2017-04-04
User file showing fatal error when sizing hot water coil UA


rraustad:
Issue overview
--------------
Very low zone loads cause the hot water coil air flow rate used to size the coil UA to be lower than SmallAirVolFlow (0.001) which results in the sizing routine setting the air flow rate to 0.

    if ( IsAutoSize || SizingDesRunThisAirSys || SizingDesRunThisZone ) {
        if ( AutosizeDes < SmallAirVolFlow && bCheckForZero) {
            AutosizeDes = 0.0;
        }
    }

Suggest adding the Boolean that avoids the low air flow check (2 places):

    } else if ( SizingType == HeatingAirflowUASizing ) {
        if ( TermUnitSingDuct || TermUnitPIU || TermUnitIU ) {
            AutosizeDes = StdRhoAir * TermUnitSizing( CurZoneEqNum ).AirVolFlow * TermUnitSizing(CurZoneEqNum ).ReheatAirFlowMult;
        } else if ( ZoneEqFanCoil ) {
            AutosizeDes = StdRhoAir * FinalZoneSizing( CurZoneEqNum ).DesHeatVolFlow;
        } else {
            AutosizeDes = FinalZoneSizing( CurZoneEqNum ).DesHeatMassFlow;
        }
        bCheckForZero = false;

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (E+ Dev support defect file folder 6088)
 - [x] Ticket added to Pivotal for defect (development team task)
 - [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Tampa weather (Miami shows issue)

rraustad:
Made suggested correction above and user file runs to completion. However, coils sizes are very small (30W) and air flow used during sizing is also very small (0.0009 m3/s or 2 CFM). Coil UA sizes out to 1W/K, seems very unreasonable.

[...]

PR #6089, opened 2017-04-04
Change HW coil UA sizing to use UA = 1 when loads are very small

rraustad:
Pull request overview
---------------------
Very low loads can result in a calculated HW coil air flow rate that is less than 0.001 m3/s which makes it difficult or impossible to calculate a coil UA value. When loads are very small, set the coil UA to 1 so that the simulation can proceed. This low UA value is representative of a very small coil as calculated by the simulation sizing routine. This change corrects #6088. Minimal or no differences in results are expected. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ReportSizingManager.cc
+++ b/src/EnergyPlus/ReportSizingManager.cc
@@ -1147,7 +1147,6 @@ namespace ReportSizingManager {
 					} else {
 						AutosizeDes = FinalZoneSizing( CurZoneEqNum ).DesHeatMassFlow;
 					}
-					bCheckForZero = false;
 				} else if (SizingType == WaterHeatingCoilUASizing) {
 					if ( DataCapacityUsedForSizing > 0.0 ) {
 						Par( 1 ) = DataCapacityUsedForSizing;
@@ -1951,7 +1950,6 @@ namespace ReportSizingManager {
 						}
 					}
 					AutosizeDes *= StdRhoAir;
-					bCheckForZero = false;
 				} else if (SizingType == WaterHeatingCoilUASizing) {
 					if ( DataCapacityUsedForSizing >= SmallLoad ) {
 						Par( 1 ) = DataCapacityUsedForSizing;
--- a/src/EnergyPlus/ReportSizingManager.cc
+++ b/src/EnergyPlus/ReportSizingManager.cc
@@ -1148,7 +1148,7 @@ namespace ReportSizingManager {
 						AutosizeDes = FinalZoneSizing( CurZoneEqNum ).DesHeatMassFlow;
 					}
 				} else if (SizingType == WaterHeatingCoilUASizing) {
-					if ( DataCapacityUsedForSizing > 0.0 ) {
+					if ( DataCapacityUsedForSizing > 0.0 && DataWaterFlowUsedForSizing > 0.0 && DataFlowUsedForSizing > 0.0 ) {
 						Par( 1 ) = DataCapacityUsedForSizing;
 						Par( 2 ) = double( DataCoilNum );
 						Par( 3 ) = double( DataFanOpMode ); //fan operating mode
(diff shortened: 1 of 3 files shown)


rraustad:
Needs unit test.

nrel-bot-3:

rraustad:
Results from defect file show UA is set to 1 when loads are very small. First coil below is just larger than limit on air flow of 0.001 based on coil capacity. Second coil hits this limit and sets coil UA = 1.

    Component Sizing Information, Coil:Heating:Water, TZ15-34 ZONE CV HW RHT COIL, Design Size Design Coil Load [W], 30.77731
    Component Sizing Information, Coil:Heating:Water, TZ15-34 ZONE CV HW RHT COIL, Design Size Maximum Water Flow Rate [m3/s], 3.67747E-007
    Component Sizing Information, Coil:Heating:Water, TZ15-34 ZONE CV HW RHT COIL, Design Size U-Factor Times Area Value [W/K], 1.16591

    Component Sizing Information, Coil:Heating:Water, TZ15-41 ZONE CV HW RHT COIL, Design Size Design Coil Load [W], 26.63090
    Component Sizing Information, Coil:Heating:Water, TZ15-41 ZONE CV HW RHT COIL, Design Size Maximum Water Flow Rate [m3/s], 3.18203E-007
    Component Sizing Information, Coil:Heating:Water, TZ15-41 ZONE CV HW RHT COIL, Design Size U-Factor Times Area Value [W/K], 1.00000

rraustad:
In this branch there is a big diff for DElight-Detailed-Comparison but the only thing I see different is the warmup loads convergence. I don't see how that can cause a big diff unless some piece of equipment now turns on and off at a slightly different time. 

When I run this file locally, I get 0 diff's.

    DElight-Detailed-Comparison.csv
    vs
    DElight-Detailed-Comparison_new.csv
    All Records are Equal.    864 records compared
    * * * * * * * * * * * * 
    DElight-Detailed-ComparisonMeter.csv
    vs
    DElight-Detailed-Comparison_newMeter.csv
    All Records are Equal.    3 records compared

rraustad:
@larryscheier the eio results above show a HW coil with 30 W capacity and 0.0000003677 m3/s water flow rate or 0.006 gpm. What does Trane do regarding coil selection for coils such as this one. Does Trane just provide calculations made by the software and the user then decides what size coil to install? If so, then if there are many of these small coils, and the user decides to install e.g.,  500 W coils with 2 gpm, then the plant flow rate and pump would need to be upsized from what the simulation reported. That sounds like a rather tedious post-processing task (i.e., translating simulation output to plant schedules on mech drawings). Alternately, if the loads are really that small then the plant and pump flow rate probably wouldn't increase much even when larger coils are installed. I'm talking out load here, those small sizes just bother me and I can't let it go.

larryscheier:
@EnergyArchmage @nagappanchidambaram  "Does Trane just provide calculations made by the software and the user then decides what size coil to install?" Yes, Trace 700 will report and use during the simulation the small autosized HW coil sizes as if those small capacity coils were available. In reality, the user will look at our sizing report and decide whether or not to install such a small coil or they will put in the smallest size available. I don't ever see users actually trying to override these small coils because, as you say, that would be a pain. However, third party UI software developers can automate this task (reviewing autosized coil sizes after initial sizing run and replace with the closest actual size) so this is not a task I expect E+ engine to perform. I would avoid fixes that just set the UA coil size to zero because this could lead to unmet hours.

nrel-bot-3:

nrel-bot-3:

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1951,7 +1951,7 @@ namespace ReportSizingManager {
 					}
 					AutosizeDes *= StdRhoAir;
 				} else if (SizingType == WaterHeatingCoilUASizing) {
-					if ( DataCapacityUsedForSizing >= SmallLoad ) {
+					if( DataCapacityUsedForSizing >= SmallLoad && DataWaterFlowUsedForSizing > 0.0 && DataFlowUsedForSizing > 0.0 ) {
```
Myoldmopar, line 1954: Space between if and open parenthesis got lost, but otherwise these code changes look good.  Diffs in DElight files do not bother me.  I'm OK merging this.  Any hold ups from anyone else?

Myoldmopar (approved):

Issue #6105, opened 2017-04-19
HW supplemental heating coil in AirloopHVAC UnitarySystem not controlling to set point


rraustad:
Issue overview
--------------
User file uses HW supplemental heating coil in AirloopHVAC:UnitarySystem with Setpoint control. HW coil outlet temperature overshoots set point at times and undershoots other times. The Coil:Heating:WaterToAirHeatPump:VariableSpeedEquationFit also appears to have control issues.

Found that CW and HW coils also exhibit poor set point based control. Added new defect file 6105-HVACTemplate-5ZoneUnitarySystem_expidf.idf

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (V8.7)
 - Unmethours link or helpdesk ticket number - HelpDesk  ticket 11986

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (GitHub - EnergyPlusDevSupport\DefectFiles) - Burlington VT weather
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Results show poor control of HW supplemental coil.

Supp coil inlet air node = Unitary HC Air Outlet
Supp coil outlet air node = Unitary Air Outlet Node

See x-axis 400 - 600. HW coil inlet node air temp (dark blue dash line) is at the SP (grey line) yet coil is fully on providing a high outlet temperature (orange). Also appears that HW supp coil water outlet node mass flow rate (yellow) does not represent coil heating rate (light blue) (i.e., why 0 at 500-550 when heating coil heating rate is non-zero)?

![6105-hwsuppcontrol](https://cloud.githubusercontent.com/assets/5778156/25183948/f9de786a-24e6-11e7-9fc3-3b48718a54af.png)

rraustad:
A few changes and this looks much better but the water side flow rate still does not match the heating coil heating rate. Still looking and suspect that the SetComponentFlowRate is not called when needed.

![6105-hwsuppcontrol_nearlyfixed](https://cloud.githubusercontent.com/assets/5778156/25188505/49971eea-24f4-11e7-936e-30df28f878cb.png)

rraustad:
[...]

PR #6107, opened 2017-04-21
Correct poor operation of water coils in UnitarySystems using set point control

rraustad:
Pull request overview
---------------------
Use of water coils or Coil:*:WaterToAirHeatPump:EquationFit in AirloopHVAC:UnitarySystem shows poor control when ControlType=SetPoint. Revised calculation methodology to allow proper control of coil to a set point temperature. Also allow non-heat pump systems to control both cooling and heating coil to meet their respective set point temperature at the same time. This corrects issue #6105.

No known workaround.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes:
   - n/a If transition, add rules to spreadsheet
 - n/a ExpandObjects changes?


rraustad:
This defect file runs slowly so a new defect file was created from HVACTemplate-5ZoneUnitarySystem where Sys 5 was changed to Setpoint based control. Poor control on those coils as well. This graph created using PRT_Porto weather (since that's what I had set up at the time).

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -963,6 +969,12 @@ namespace HVACUnitarySystem {
 		UnitarySystem( UnitarySysNum ).DehumidInducedHeatingDemandRate = 0.0;
 		UnitarySystem( UnitarySysNum ).CoolCoilWaterFlowRatio = 0.0;
 		UnitarySystem( UnitarySysNum ).HeatCoilWaterFlowRatio = 0.0;
+		if( UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode > 0 ) Node( UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode ).MassFlowRate = 0.0;
```
rraustad: These can be deleted now.

rraustad: Interesting, if I delete these I don't get the control I had before. Also, we are not supposed to use Node().MassFlowRate on left side of equation so I'll have to switch to using SetComponentFlowRate.

rraustad:
I still have good results with this last change.

6105_fixed

rraustad:
2 files show big diffs:
EnergyPlusFixture.TestExternalNodes (???)
DElight-Detailed-Comparison (only diff in eio is warmup convergence)

4 files show small diff's in eios as warmup convergence only.
5Zone_Unitary_VSDesuperheatWaterHeater
5Zone_Unitary_VSDesuperheater
DElightCFSLightShelf
UnitarySystem_WaterCoils_wMultiSpeedFan

2 of these show error messages were removed from simulation results:
5Zone_Unitary_VSDesuperheatWaterHeater
5Zone_Unitary_VSDesuperheater

    -   ** Warning ** AirLoopHVAC:UnitarySystem - sensible part-load ratio calculation failed: part-load ratio limits exceeded, for unit = GASHEAT DXAC FURNACE 1
    -   **   ~~~   ** Estimated part-load ratio = 0.602
    -   **   ~~~   ** The estimated part-load ratio will be used and the simulation continues. Occurrence info:
    -   **   ~~~   **  Environment=CHICAGO ANN CLG .4% CONDNS WB=>MDB, at Simulation time=07/21 00:00 - 00:15

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -857,67 +858,63 @@ namespace HVACUnitarySystem {
...
-				mdot = min( Node( UnitarySystem( UnitarySysNum ).CoolCoilFluidOutletNodeNum ).MassFlowRateMaxAvail, UnitarySystem( UnitarySysNum ).MaxCoolCoilFluidFlow );
-				Node( UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode ).MassFlowRate = mdot;
+
+				//     set water-side mass flow rates
+				mdot = UnitarySystem( UnitarySysNum ).MaxCoolCoilFluidFlow;
+				SetComponentFlowRate( mdot, UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode, UnitarySystem( UnitarySysNum ).CoolCoilFluidOutletNodeNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopSide, UnitarySystem( UnitarySysNum ).CoolCoilBranchNum, UnitarySystem( UnitarySysNum ).CoolCoilCompNum );
 				//     simulate water coil to find operating capacity
 				SimulateWaterCoilComponents( UnitarySystem( UnitarySysNum ).CoolingCoilName, FirstHVACIteration, UnitarySystem( UnitarySysNum ).CoolingCoilIndex, InitUnitarySystemsQActual );
 				UnitarySystem( UnitarySysNum ).DesignCoolingCapacity = InitUnitarySystemsQActual;
+
```
rraustad, line 864: The above test is in Init and looks for the maximum capacity of fluid coils for use in the model. Changed from using Node().MassFlowRate on left side of equation to using SetComponentFlow rate calls. Similar in next 4 IF blocks.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9221,7 +9236,10 @@ namespace HVACUnitarySystem {
 
 					} else if ( ( CoilType_Num == Coil_CoolingWater ) || ( CoilType_Num == Coil_CoolingWaterDetailed ) ) { // COIL:COOLING:WATER
 
-						SimWaterCoils( UnitarySysNum, FirstHVACIteration, PartLoadFrac, CoolingCoil );
+						mdot = UnitarySystem( UnitarySysNum ).MaxCoolCoilFluidFlow;
+						SetComponentFlowRate( mdot, UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode, UnitarySystem( UnitarySysNum ).CoolCoilFluidOutletNodeNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopSide, UnitarySystem( UnitarySysNum ).CoolCoilBranchNum, UnitarySystem( UnitarySysNum ).CoolCoilCompNum );
+
+						SimulateWaterCoilComponents( CompName, FirstHVACIteration, UnitarySystem( UnitarySysNum ).CoolingCoilIndex, _, UnitarySystem( UnitarySysNum ).FanOpMode, PartLoadFrac );
```
rraustad, line 9253: This code is in the calc routine to test output at PLR=1. Note the water flow rate was never set at max to test outlet air temp vs set point temp.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9119,7 +9134,7 @@ namespace HVACUnitarySystem {
 					UnitarySystem( UnitarySysNum ).CompPartLoadRatio = PartLoadFrac;
 				} else if ( ( CoilType_Num == Coil_CoolingWater ) || ( CoilType_Num == Coil_CoolingWaterDetailed ) ) { // COIL:COOLING:WATER
 
-					SimWaterCoils( UnitarySysNum, FirstHVACIteration, PartLoadFrac, CoolingCoil );
+					SimulateWaterCoilComponents( CompName, FirstHVACIteration, UnitarySystem( UnitarySysNum ).CoolingCoilIndex, _, UnitarySystem( UnitarySysNum ).FanOpMode, PartLoadFrac );
```
rraustad, line 9148: This code is in calc routine to test output at PLR=0. Got rid of intermediate subroutine and simply called water coil directly.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9791,9 +9811,17 @@ namespace HVACUnitarySystem {
...
+										} else {
+											SolveRegulaFalsi( HumRatAcc, MaxIte, SolFla, SpeedRatio, DXCoilVarSpeedHumRatResidual, 0.0, 1.0, Par );
+										}
 									} else {
-										SpeedRatio = 1.0;
+										if( SpeedNum == 1 ) {
+											CycRatio = 1.0;
+										} else {
+											SpeedRatio = 1.0;
+										}
```
rraustad, line 9835: Also noticed some warnings from water to air HP and corrected those as well. Different function needs to be called for speed = 1 vs speed > 1.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9923,6 +9951,11 @@ namespace HVACUnitarySystem {
 
 		LoopDXCoilRTF = max( LoopDXCoilRTF, LoopDXCoilMaxRTFSave );
 
+		if ( UnitarySystem( UnitarySysNum ).CoolingCoilType_Num == Coil_CoolingWater || UnitarySystem( UnitarySysNum ).CoolingCoilType_Num == Coil_CoolingWaterDetailed ) {
+			mdot = PartLoadFrac * UnitarySystem( UnitarySysNum ).MaxCoolCoilFluidFlow;
+			SetComponentFlowRate( mdot, UnitarySystem( UnitarySysNum ).CoolCoilFluidInletNode, UnitarySystem( UnitarySysNum ).CoolCoilFluidOutletNodeNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopNum, UnitarySystem( UnitarySysNum ).CoolCoilLoopSide, UnitarySystem( UnitarySysNum ).CoolCoilBranchNum, UnitarySystem( UnitarySysNum ).CoolCoilCompNum );
+		}
+
```
rraustad, line 9969: And finally, at end of SP subroutines, set final fluid flow rate using SetComponentFlowRate. 3 places, here and at end of heating and supp heating routines.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -126,6 +126,422 @@ using WaterCoils::WaterCoil_Cooling;
 using WaterCoils::SimpleAnalysis;
 using General::TrimSigDigits;
 
+class ZoneUnitarySystemTest: public EnergyPlusFixture {
```
rraustad, line 127: Setting up a class so unit tests for UnitarySystem would not use as many IDF objects. I'll circle back and clean up other unit tests as this method progresses.

rraustad:
@Myoldmopar @mjwitte this PR has a revision to the UnitarySystem unit test that I would like to start using. I'd like to get this one into develop when one of you has some spare review time available.

Myoldmopar:
Are you saying you'd like this whole PR to be put on a rush review?  Or that you'd like to get the unit test into develop right away on its own?

rraustad:
I'm interested in the unit test method. It's easier and cleaner. Only when you have time though. I can't do any other UnitarySystem unit tests like this until this gets into develop.

Myoldmopar:
I mean, I don't mind just creating a branch off of develop right now, pasting in the new unitary system unit test file as it is in this branch, making a pull request, and getting it into develop.  But if you expect to be making any more changes to the file in this branch, it could hit conflicts when you pull develop back in.  Probably not hard to resolve, but still annoying.

rraustad:
No special treatment, just pointing to this PR as a potential candidate for dropping.

Myoldmopar:
Fair enough.  OK, so this PR should just drop sooner than others if possible.  Cool.

mjwitte (commented):
Tested both defect files before and after.  

HW coil control looks good for the HVACTemplate test file, but there are times where it appears that the cooling coil is overshooting?  7/7 9:00-21:00.  I ran this with Chicago weather.

The other test file is still a mess.  Many hours overheating, and now the simulation fails with an overheated plant loop.  There are times where the HW coil is still overheating, e.g. on 04/01, it's on setpoint until 11:30, then it puts out 38C for the rest of the day while the setpoint is only 18-20C.

mjwitte:
p.s. I merged in develop to make this branch current.

rraustad:
Well, not sure what to say except sorry for wasting your time. I ran the HVACTemplate file again and do not see where there is an overshoot with the CW coil. This result looks pretty much what I remember and same as what I show above. With this result I had thought I corrected the problem. I will look at the defect file next.

6105_verify

mjwitte:
I was looking at "Sys 5 HW ChW System Heating Coil Outlet" the system outlet node.  I see the problem, the cooling coil outlet is on a SetpointManager:MixedAir even though this system is a blowthru system. So that explains that.

rraustad:
I think you just highlighted a different issue. Looking at the system outlet node, the HW coil undershoots the SP. This is probably due to water coil controllers being called before the coil is simulated? I don't get it though. The HVACTemplate example file has fan, then cooling coil, then heating coil. Why doesn't the HW coil respond to the CW coil outlet node temp, which is below the HW coil SP (col CN vs CO)? It doesn't matter how the SP got there, the coil should control to the SP. On successive iterations, the HW coil controller should see that temp as the inlet to the HW coil ? The HW coil water flow is 0 here (col CR and CS). Since the water coils are controlling better now (previous figure for HW coil control), I don't think it's the UnitarySystem causing this but can't be sure at this point. Probably a separate issue. On to the defect file.

6105_hw_cw_controllerissue

rraustad:
Well, it's not the controllers since this is a UnitarySystem and doesn't use controllers.

rraustad:
Looking at the defect file I don't see any problems other than the defect file needed a bypass pipe in the HW loop (used to plot results below) and that the plant sometimes chokes flow to the VS heating coil. From these results you can see that the output air SP is mostly met (right figures) and each coil operates to meet their SP. For times when it appears that the VS heating coil (left graphs) is not doing it's job, the plant has choked water flow to 0 and the supplemental coil picks up the load (these 2 coils are on different plant loops, VS heat on GSHP plant and supp heat coil on HW plant). I don't see anything when single stepping that makes me think I missed something (except that when plant chokes flow the VS coil speed goes to max before exiting loop since it can't meet the SP at any speed).

suppheaterworks

rraustad:
I added the defect file *_wBypassBranch to the defect file folder.

rraustad:
Regarding the lack of HW coil control in previous comment, the routines check to see if the other coil PLR = 0 before allowing operation. This means they can never operate at the same time. Since the cooling coil is first, it won. For load based control this makes sense, but for SP control I don't think it makes as much sense. What if the user is somehow trying to overdehumidify using SP's?

Should I remove the check for other coil PLR = 0? These routines are unique to SP based control.

    // IF DXCoolingSystem is scheduled on and there is flow
    if ( ( GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).SysAvailSchedPtr ) > 0.0 ) && GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).CoolingCoilAvailSchPtr ) > 0.0 && ( Node( InletNode ).MassFlowRate > MinAirMassFlow ) && UnitarySystem( UnitarySysNum ).HeatingPartLoadFrac == 0.0 ) {


    // IF DXHeatingSystem is scheduled on and there is flow
    if ( GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).SysAvailSchedPtr ) > 0.0 && GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).HeatingCoilAvailSchPtr ) > 0.0 && Node( InletNode ).MassFlowRate > MinAirMassFlow && UnitarySystem( UnitarySysNum ).CoolingPartLoadFrac == 0.0 ) {

mjwitte:
If the heating coil and the cooling coil are part of the same heat pump, then you need this check.  But if either coil is non-dx, then you shouldn't lock out the other coil.

mjwitte:
Just retested with the version with the bypass pipe.  Wow - big difference.  Without the bypass pipe, the supp heat coil goes wild and puts out 38C outlet temps - should this be happening?  With the bypass pipe, it's off and the supply air temp is where it should be.

rraustad:
With no bypass pipe the plant can force water through a branch when it is not requested, or restrict when it is requested. If this is a water coil and air is passing through the coil, what should the model do? I say it should transfer heat from water to air, which it does. The water coil model cares only about the water side flow rate and then passes heat to the air stream if water flow > 0.

I'll see if there is an easy way to check for other coil PLR = 0 only when it's a HP.

rraustad:
Non-HP systems now allow both main coils to operate simultaneously.

6105_cwhwcoilsoperatetogether

rraustad:
The files with diff's now are due to the use of non-HP type equipment in the UnitarySystem model and that now E+ allows both coils to operate simultaneously.

One has this:

  AirLoopHVAC:UnitarySystem,
    Coil:Heating:Fuel,       !- Heating Coil Object Type
    Main Heating Coil 1,  !- Heating Coil Name
    CoilSystem:Cooling:DX:HeatExchangerAssisted,  !- Cooling Coil Object Type
    Heat Exchanger Assisted Cooling Coil 1,      !- Cooling Coil Name

The other two have:

  AirLoopHVAC:UnitarySystem,
    Coil:Heating:Fuel,       !- Heating Coil Object Type
    Main Heating Coil 1,  !- Heating Coil Name
    Coil:Cooling:DX:VariableSpeed,  !- Cooling Coil Object Type
    Main Cooling Coil 1,      !- Cooling Coil Name

I should probably look at a before and after for one of these to see what the diff's are.

rraustad:
Well, these files are doing what they did before but don't ask me what that is since the system outlet temp isn't even close to what it should be. I'm not sure what these example files are supposed to represent.

Header for 5Zone_Unitary_HXAssistedCoil:

    ! Highlights:              packaged VAV system with setpoint-based Unitary system using variable speed air to air cooling coil model
    !                          desuperheater for supplemental coil used for humidity control. 


6105_5zone_unitary_hxassistedcoil

If I zoom in to see what is different it looks like the main heating coil is on now instead of the reheat coil. This makes sense since both main coils are allowed to operate simultaneously.

6105_heatcoilswitch

rraustad:
OK @mjwitte. Add a new issue to fix these example files to run better?

mjwitte:
Hmm, for this case (dehumidification with desuperheat available) you want the desuperheat to do what it can before activating the heating coil.  Should the controls be smart enough to handle this?  Maybe.  Wait - this file doesn't actually have a desuperheater coil - comments need to be updated.

But the UnitarySystem docs talk about a reheat coil.  But the field name is "Supplemental Heating Coil Object Type".  In the I/O Ref "Supplemental Heating Coil Object Name" speaks of a reheat coil - but not a supplemental coil.  IDD notes and I/O Ref need some cleanup here.  If I understand correctly, this coil is really "Supplemental Heating or Reheat Coil"?

Back to the original question - if we are saying that reheat is done by the third coil, then the main heating coil should *not* operate simultaneously with the cooling coil, only the supplemental/reheat coil should operate.

rraustad:
The names of the coils don't reflect their use. The choice of the type of 2nd heating coil depends on the intended function. It would be a reheat coil during dehumidification or a supplemental coil if the main heating coil didn't have the capacity or was off for some reason.

Since this discussion involves set point based control, the model doesn't know why the set point was placed at the coil outlet, it just tries to meet that SP. So I think you had it right the first time. For SP control, the coils always try to meet their respective SP unless it's a HP. For load based control, only one of the main coils will operate. I guess there is a special case when there is a heating and dehumidification load. For that case, I can see all 3 coils operating. Not sure that works in current code.

mjwitte:
But if the reheat coil really *is* a desuperheater, then you don't want the heating coil doing the reheat - you want the desuperheater to do it.  There has to be a way to make that happen.

rraustad:
I'll think about how to prioritize heating mode when a desuperheater is used. Sounds like a different issue, and that I don't have a good solution at the moment.

mjwitte:
How are the two heating coils treated during load-based control?  Is the main heating coil ever used for reheat in that case?  I would think if <dehumidification control = coolreheat> and <cooling coil is operating> then main heating coil is off.  Maybe that's too simplistic.

rraustad:
For load control the main heating coil would try to meet the heating load. The main heating coil does not know of a reheat or supplemental heating load. That is calculated after the main heating coil is simulated. If there were remaining load, the next coil would pick that up. How to prioritize the second of these two coils eludes me.

rraustad:
I guess, if a desuperheater is used, I could model the main heating coil with 0 load then let the desuperheater try to meet all the load, then calculate what the main heating coil should do and then simulate each with those new coil load calculations. Probably not a horrible solution.

rraustad:
Complicated for constant fan mode. Heating load means Tstat requires heating, or if constant fan, intake of OA overshoots the Tstat heating SP and heat is required (even if the Tstat thinks a cooling load exists or is in deadband).
If there is a heating load, the main heating coil operates, same for cooling. If there is a cooling load and dehumidification, then the main cooling coil and sup coil should run. If there is a heating load and dehumidification, then all 3 can run, IF this code handles that correctly.
Yes the main heating coil is supposed to run only for heating load sake. The supp coil picks up all excess heating (supp) and reheat loads.

rraustad:
Just pushed a fix to the docs per your 1st comment 9 days ago. CI should be running soon.

nrel-bot-3:
@rraustad @lgentile it has been 18 days since this pull request was last updated.

rraustad:
Reminder that this blocks the "Increase speeds in multi speed object" effort. Waiting on this new unit test to move forward on other UnitarySystem unit tests.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9080,10 +9095,12 @@ namespace HVACUnitarySystem {
...
 		// IF DXCoolingSystem is scheduled on and there is flow
-		if ( ( GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).SysAvailSchedPtr ) > 0.0 ) && GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).CoolingCoilAvailSchPtr ) > 0.0 && ( Node( InletNode ).MassFlowRate > MinAirMassFlow ) && UnitarySystem( UnitarySysNum ).HeatingPartLoadFrac == 0.0 ) {
+		if ( ( GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).SysAvailSchedPtr ) > 0.0 ) && GetCurrentScheduleValue( UnitarySystem( UnitarySysNum ).CoolingCoilAvailSchPtr ) > 0.0 && ( Node( InletNode ).MassFlowRate > MinAirMassFlow ) ) {
 
-			// Determine if there is a sensible load on this system (aren't the first 2 tests redundant?)
-			if ( ( Node( InletNode ).Temp > DesOutTemp ) && ( std::abs( Node( InletNode ).Temp - DesOutTemp ) > TempControlTol ) ) SensibleLoad = true;
+			// Determine if there is a sensible load on this system
+			if ( Node( InletNode ).Temp - DesOutTemp > TempControlTol ) SensibleLoad = true;
+			// if a heat pump and other coil is on, disable this coil
+			if ( UnitarySystem( UnitarySysNum ).HeatPump && UnitarySystem( UnitarySysNum ).HeatingPartLoadFrac > 0.0 ) SensibleLoad = false;
```
rraustad, line 9099: over specified test for sensible load. simplified to be just if temp - SP > tol then coil is on (cooling). Heating is reversed (DesOutTemp- temp).

rraustad:
Drop this in.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -819,10 +820,10 @@ namespace HVACUnitarySystem {
...
 				}
 			}
-			if ( UnitarySystem( UnitarySysNum ).HeatingCoilType_Num == Coil_HeatingGasOrOtherFuel || UnitarySystem( UnitarySysNum ).HeatingCoilType_Num == Coil_HeatingElectric ) {
+//			if ( UnitarySystem( UnitarySysNum ).HeatingCoilType_Num == Coil_HeatingGasOrOtherFuel || UnitarySystem( UnitarySysNum ).HeatingCoilType_Num == Coil_HeatingElectric ) {
 //				SimulateHeatingCoilComponents( UnitarySystem( UnitarySysNum ).HeatingCoilName, FirstHVACIteration, 1.0, UnitarySystem( UnitarySysNum ).HeatingCoilIndex, _, _, UnitarySystem( UnitarySysNum ).FanOpMode, 1.0 );
 //				UnitarySystem( UnitarySysNum ).DesignHeatingCapacity = GetHeatingCoilCapacity( cAllCoilTypes( UnitarySystem( UnitarySysNum ).HeatingCoilType_Num ), UnitarySystem( UnitarySysNum ).HeatingCoilName, InitUnitarySystemsErrFlag );
-			}
+//			}
```
mjwitte: @rraustad Are we ready to delete these commented lines?

mjwitte (commented):
This is working much better than before.  I've added a new flavor of the defect file.  There's a few places that look odd.

01/01 - UNITARY CC:Cooling Coil Sensible Cooling Rate [W](Each Call) is always zero as expected, but there are occasional non-zero values for UNITARY CC WATER INLET:System Node Mass Flow Rate [kg/s](Each Call)  This is happening in v8.7 without this fix, so maybe that's a separate issue.  Similar behavior is seen on the UNITARY HC WATER INLET in July, see 07/04 for example.

04/06 06:00-06:18  Heating output seems to randomly flip between the watertoair heatpump heating coil UNITARY HC:Heating Coil Heating Rate [W](Each Call) and the HW supp heating coil UNITARY SUPP HEAT:Heating Coil Heating Rate [W](Each Call).  The supp heat coil is controlling well and pickup up the load, but why is the main heating coil not running?

mjwitte:
@rraustad And be sure to pull this branch first - I merged it up and fixed a tiny typo.

rraustad:
Verified there is water flow rate when cooling coil is off (4/1 around 4:18 AM). Stepped all the way through UnitarySystem and see water flow rate = 0 since coil is off. Then step by ManagePlantLoops in SimSelectedEquipment and boom, cooling coil water flow rate is non-zero. FlowLock = 1 at this point (same for all 3 plant loops) so I guess it's just passing water through the coil even though it is off. I don't see a bypass pipe in this loop so you get what you get. Will look at random flip between main heating coil and supp heating coil next.

    BranchList,
      GSHP Demand Side Branch List,  !- Name

    int const FlowUnlocked( 0 ); // components request flow
    int const FlowLocked( 1 ); // components take their inlet flow

rraustad:
RE: Flip flopping heating coils. The plant is shutting down flow to the Unitary HC and the supp coil is picking up the load. At this time, the Unitary HC goes to speed = 10 and the outlet air temp never changes. I stepped into VS coils and see that when SetComponentFlowRate is called (line 2965) that VarSpeedCoil( DXCoilNum ).WaterMassFlowRate went from about 24 to 0, and so the supp coil needed to turn on. Why this happens is a different question.

    if ( ( VarSpeedCoil( DXCoilNum ).VSCoilTypeOfNum == TypeOf_CoilVSWAHPHeatingEquationFit ) || ( VarSpeedCoil( DXCoilNum ).VSCoilTypeOfNum == TypeOf_CoilVSWAHPCoolingEquationFit ) ) {
        SetComponentFlowRate( VarSpeedCoil( DXCoilNum ).WaterMassFlowRate, VarSpeedCoil( DXCoilNum ).WaterInletNodeNum, VarSpeedCoil( DXCoilNum ).WaterOutletNodeNum, VarSpeedCoil( DXCoilNum ).LoopNum, VarSpeedCoil( DXCoilNum ).LoopSide, VarSpeedCoil( DXCoilNum ).BranchNum, VarSpeedCoil( DXCoilNum ).CompNum );

rraustad:
The only thing left is the commented out code that is no longer used (lines 823 - 826). I do see some things that look odd, for example, the unitary fan part-load ratio always = 0. This is because CompOnMassFlowRate = 0 for set point based systems (i.e., SetOnOffMassFlowRate never called). This was an over sight. Also, at the beginning of ReportUnitarySystem, many variables are reset to 0 (e.g., PLR, FanPLR, CycRatio, SpeedRatio). These should always be reset to the end result and it appears some coil types are missing. What I see is that the VS DX coil doesn't reset CycRatio and SpeedRatio, which is OK since that coil model doesn't have these report variables, but I think this type of result is necessary for variable speed coils. These changes are beyond the scope if this issue.

mjwitte:
@rraustad OK - added a bypass pipe on the demand side of the GSHP loop and the strange operation is gone (no flow when coil load is zero, and no flip-flopping between heat pump heating coil and supp heating coil).  If you think we need a followup issue on the things you noticed above, please post it.  I'll let CI finish and then drop this in.

rraustad:
Issue #6201 posted.

mjwitte:
Going ahead and merging this, but the WinCI results don't make any sense compared to the other platforms.  @Myoldmopar @jasondegraw

Issue #6020, opened 2017-02-17
Incorrect distribution of DOA to primary air inlet nodes of terminal air mixers 


Nigusse:
Issue overview

DOA primary air from terminal air mixer is evenly distributed among all the zones, regardless of the amount of design outdoor air specified for each zone. This was reported for FanCoil unit example file DOAToFanCoilInlet.idf.

### Details
Some additional details for this issue (if relevant):
- Platform Win64
- Version of EnergyPlus 8.6, 8.5

Further testing showed the same issue in PTAC, PTHP, WSHP, VRF, and UnitarySystem connected to DOA example files.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

EnergyArchmage:
Currently the whole-air-system OA flow rate comes from the sum of the OA design rates for each of the zones attached.  Then that whole-air-system flow is evenly divided between the zones attached, which will nearly always be incorrect. 

I suggest that the best solution to this issue is to add a new input field for the name of a DesignSpecification:OutdoorAir to the AirTerminal:SingleDuct:Mixer object.  This allows for DCV as well as correctly distributing OA flow among the different zones.

Nigusse:
The AirTerminal:SingleDuct:Mixer object implementation use all the calculations in each zoneHVAc equipment.  We can use the OA amount from ZoneHVAC equipment determined from DesignSpecification:OutdoorAir object.

mjwitte:
Also check CO2 concentration at mixer outlet (and other contaminants and OA mass flow).  Ref:  Helpdesk ticket 12014.

EnergyArchmage:
But the ZoneHVAC equipment do not all have a reference to a DesignSpecification:OutdoorAir object, so how can it be used.  Are you proposing to take it from the zone sizing object?  

The I/O reference specifically says to set the outdoor air flow rates to zero in the ZoneHVAC when using the mixer.

mjwitte:
I think the ZoneHVAC equipment should have a non-zero OA flow rate (set either directly or autosized) and that should be the flow request on the mixer terminal unit.  That should flow upstream to the DOAS just like it would from any other standard terminal unit.

Nigusse:
[...]

PR #6117, opened 2017-05-05
Correct distribution of DOA to primary air inlet nodes of ZoneHVAC

Nigusse:
Issue  #6020

***This description needs to change to match the final solution***
Now the OA is correctly distributed to air terminal mixer and zones per the amount specified in OA Design Specification object. This issue has been addressed for FanCoil, WSHP, PTAC, PTHP and VRF units. 

Pull request overview
---------------------
This is a defect. Diffs are expected because, now a different amount of OA is used instead of zero. Diffs are expected in EIO, audit and csv output files. Idd has been cleaned. Notes has been removed from OA input fields for each object.   And I/O reference document has also been updated.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:


Nigusse:
Now DOA is correctly distributed to the terminal air mixers connected to the air loop per the zone outdoor air design specification. Example files that use terminal air mixer objects are expected to show diffs. Some unit test files that failed due to this fix has also been addressed. Unit tests are going to be added.

Nigusse:
Added CO2 and general contaminant update equation to air terminal mixer model.

nrel-bot-3:
@Nigusse @lgentile it has been 15 days since this pull request was last updated.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -3690,32 +3690,32 @@ namespace HVACVariableRefrigerantFlow {
...
+			//	}
+			//	//   check that OA flow in no cooling and no heating must be set to zero when connected to DOAS
+			//	if ( VRFTU( VRFTUNum ).NoCoolHeatOutAirVolFlow != 0 ) {
+			//		ShowWarningError( cCurrentModuleObject + " = " + VRFTU( VRFTUNum ).Name );
+			//		ShowContinueError( ".. " + cNumericFieldNames( 7 ) + " must be zero when " + cCurrentModuleObject );
+			//		ShowContinueError( "..object is connected to central dedicated outdoor air system via AirTerminal:SingleDuct:Mixer" );
+			//		ShowContinueError( ".. " + cNumericFieldNames( 7 ) + " is set to 0 and simulation continues." );
+			//		VRFTU( VRFTUNum ).NoCoolHeatOutAirVolFlow = 0;
+			//	}
+			//}
```
Myoldmopar: This big comment should probably just be removed right?

src/EnergyPlus/PackagedTerminalHeatPump.cc
```diff
@@ -1327,32 +1327,32 @@ namespace PackagedTerminalHeatPump {
...
+			//	}
+			//	//   check that OA flow in no cooling and no heating must be set to zero when connected to DOAS
+			//	if ( PTUnit( PTUnitNum ).NoCoolHeatOutAirVolFlow != 0.0 ) {
+			//		ShowWarningError( CurrentModuleObject + " = " + PTUnit( PTUnitNum ).Name );
+			//		ShowContinueError( ".. " + cNumericFields( 6 ) + " must be zero when " + CurrentModuleObject );
+			//		ShowContinueError( "..object is connected to central dedicated outdoor air system via AirTerminal:SingleDuct:Mixer" );
+			//		ShowContinueError( ".. " + cNumericFields( 6 ) + " is set to 0 and simulation continues." );
+			//		PTUnit( PTUnitNum ).NoCoolHeatOutAirVolFlow = 0;
+			//	}
+			//}
```
Myoldmopar: It's great if this isn't needed, just go ahead and remove it.

testfiles/DOAToFanCoilInlet.idf
```diff
@@ -2127,7 +2127,7 @@
     autosize,                !- Maximum Supply Air Flow Rate {m3/s}
     ,                        !- Low Speed Supply Air Flow Ratio
     ,                        !- Medium Speed Supply Air Flow Ratio
-    0,                       !- Maximum Outdoor Air Flow Rate {m3/s}
+    autosize,                !- Maximum Outdoor Air Flow Rate {m3/s}
```
Myoldmopar, line 2130: Transition rule needed?

testfiles/DOAToPTHP.idf
```diff
@@ -157,7 +157,7 @@
 
  ! Miami Intl Ap FL USA Annual Heating Design Conditions Wind Speed=3.8m/s Wind Dir=340
  ! Coldest Month=January
- ! Miami Intl Ap FL USA Annual Heating 99.6%, MaxDB=8.7�C
+ ! Miami Intl Ap FL USA Annual Heating 99.6%, MaxDB=8.7?C
```
Myoldmopar, line 160: Not sure if the black invalid character or the plain question mark is better here.  Actually, leaving the bad character in there is a better option because it's easier to search and replace later.  I can try to undo these changes to the files if you need help @lgu1234, since I assume you didn't manually make this change, your editor did.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -5003,15 +5004,35 @@ namespace SingleDuct {
...
 		// SUBROUTINE LOCAL VARIABLE DECLARATIONS:
 		int MixedAirOutNode;
+		int PriInNode;
+		int SecInNode;
 
+		PriInNode = SysATMixer( SysNum ).PriInNode;
+		SecInNode = SysATMixer( SysNum ).SecInNode;
 		MixedAirOutNode = SysATMixer( SysNum ).MixedAirOutNode;
-
+
```
Myoldmopar: Moving forward, it's fine to just define the variables on the same line as the assignment:

```
int PriInNode = SysATMixer( SysNum ).PriInNode;
```

No big deal, just feel free to do that in the future.

There is, however, this stray chunk of white space on line 5013 that might as well go away now.

src/EnergyPlus/PackagedTerminalHeatPump.cc
```diff
@@ -1956,32 +1956,32 @@ namespace PackagedTerminalHeatPump {
...
+			//	}
+			//	//  check that OA flow in no cooling and no heating must be set to zero when connected to DOAS
+			//	if ( PTUnit( PTUnitNum ).NoCoolHeatOutAirVolFlow != 0.0 ) {
+			//		ShowWarningError( CurrentModuleObject + " = " + PTUnit( PTUnitNum ).Name );
+			//		ShowContinueError( ".. " + cNumericFields( 6 ) + " must be zero when " + CurrentModuleObject );
+			//		ShowContinueError( "..object is connected to central dedicated outdoor air system via AirTerminal:SingleDuct:Mixer" );
+			//		ShowContinueError( ".. " + cNumericFields( 6 ) + " is set to 0 and simulation continues." );
+			//		PTUnit( PTUnitNum ).NoCoolHeatOutAirVolFlow = 0;
+			//	}
+			//}
```
Myoldmopar: Same comment, toss these away if they aren't needed.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -9530,6 +9548,19 @@ namespace HVACVariableRefrigerantFlow {
...
+				//Node( VRFTU( VRFTUNum ).ATMixerPriNode ).MassFlowRate = min( Node( VRFTU( VRFTUNum ).ATMixerPriNode ).MassFlowRateMaxAvail, Node( VRFTU( VRFTUNum ).VRFTUInletNodeNum ).MassFlowRate );
+				SimATMixer( VRFTU( VRFTUNum ).ATMixerName, FirstHVACIteration, VRFTU( VRFTUNum ).ATMixerIndex );
+				Tin = Node( VRFTU( VRFTUNum ).ATMixerOutNode ).Temp;
+				Win = Node( VRFTU( VRFTUNum ).ATMixerOutNode ).HumRat;
+			} else {
+				Tin = Node( VRFTU( VRFTUNum ).ATMixerSecNode ).Temp;
+				Win = Node( VRFTU( VRFTUNum ).ATMixerSecNode ).HumRat;
+			}
+		}
+
```
Myoldmopar, line 9677: The changes in this file are just another reminder of how many _gotchas_ (potential future bugs) there in in the codebase.  Your code is fine @lgu1234, I'm just lamenting for a moment :laughing:

Myoldmopar (changes requested):
There are some minor changes, code cleanup mostly.  I had a question regarding the need for a transition rule since you manually went in and updated the IDFs.  I also made a comment about the degree symbol in the IDFs that we need to resolve.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Nigusse: Yes, the commented out section should be removed.

Nigusse (commented):
@Myoldmopar Thank you for the review. I am addressing your comments.

Myoldmopar:
OK, so a few things:
 - I merged develop in here, so pull before doing anything else
 - There are build warnings, just unused variables, but I didn't resolve those yet
 - I was about to start making the transition program, and found I hadn't started it for 8.8 yet, so I am going to go do that in a separate PR and then I'll merge it into this branch when ready
 - But now for the real concern.  We are actually OK with forcing a file that wasn't autosized before into being autosize based on them having a blank or zero in this field?  (https://github.com/NREL/EnergyPlus/pull/6117/files#diff-84d66175020363cd70e1e9161a145675R11)  I feel very uneasy about doing that.  What if the user doesn't have design/sizing objects in their file, won't this be a big annoyance?  Maybe I'm missing something, please clarify.

Nigusse:
@Myoldmopar I did have the same concern as yours but forgot to mention it. Probably the best approach is not to provide a transition. Let the user file run with whatever values specified in those fields.

Myoldmopar:
I agree.  @mjwitte ?

mjwitte:
Not sure yet - I need some time to look this over and understand what the changes are.  BTW - this branch is conflicted now.

Myoldmopar:
Yeah the conflict will go away if we decide we don't actually need the transition rule anyway.

mjwitte:
@Nigusse @Myoldmopar The fundamental question is which device should be controlling the outdoor air flow.  In the original implementation, the DOAS controlled the flow to the zone unit (even though it divided up the flow incorrectly).  This fix flips that around (if I'm following correctly) so now the zone unit sets the flow rate on the OA inlet node which is one of the mixer nodes.  

* If we don't transition these to autosize, then we end up with zero OA flow and the DOAS doesn't run?
* What happens if the DOAS can't provide all the requested airflow?  Max iterations exceeded?  Or just starve the OA flow?  If it's starved, then what to the flows in the zone unit look like - just a smaller OA fraction?  
* What happens when the DOAS air handler is off but the zone unit is running?
* What happens when the DOAS air handler is on but the zone unit is off?  In a real system, for the outlet mixer, the OA would go into the zone.  For the inlet mixer, the OA proably would go to the zone?  But with the current code, I'm guessing there is no flow at the mixer if the zone unit is off.  Which may simply shut off the DOAS (whether that's the user's intent or not) or go into max iteration land.
* There is a pending new feature request for the OA flow rate to be controllable for this configuration (DCV, time-of-day, etc.).  That may influence how we do this fix.

One option to consider is to add a new input to the AirTerminal:SingleDuct:Mixer, "Design Specification Outdoor Air Object Name".  That would allow the mixer to set the OA flow rate (fixed for now, but variable with the above new feature).  This would be similar to what's done with AirTerminal:DualDuct:VAV:OutdoorAir.  But if we do this, then the zone units need to do the right thing with the OA flow (even when they are off).

testfiles/DOAToPTAC.idf
```diff
@@ -1910,13 +1910,13 @@
   Sizing:Zone,
     SPACE1-1,                !- Zone or ZoneList Name
     SupplyAirTemperature,    !- Zone Cooling Design Supply Air Temperature Input Method
-    13.8,                    !- Zone Cooling Design Supply Air Temperature {C}
+    15.5,                    !- Zone Cooling Design Supply Air Temperature {C}
```
mjwitte, line 1913: @Nigusse Curious why this change and others like it?

Nigusse: Because of changes in the amount of outdoor air (primary air), there were warning error messages that are related to capacity to flow ratio out of range.

Nigusse:
I agree that now the OA amount is set by the Zone equipment and will be the design OA flow. It is only the zone equipment that has the information about the amount of OA. And the OA amount can be limited by the supply air flow of the parent object in the case of inlet side mixer or will be the maximum amount if it is supply side mixer.

(1) Yes it can result to zero outdoor air flow if we do not transition when the flow rates are set to zero. But the flip side is that if we transition to autosize and zone sizing is set to No, then we will have a different issue. Probably the best option is to provide only warning message during transition.

(2) The model does not control the OA amount directly, but it will be proportional to the supply air or will be the maximum design amount. The terminal mixer does not do any control except just calculating the mixed air properties based on the flow rates set by the zone equipment.

(3) I believe that if a zone equipment is off, then the terminal air mixer primary air flow rate is set to zero as well. If all zone equipment connected to DOAS are off, then there will not be any flow set to DOAS. I agree this may cause max iteration issue but I have not tested it.

(4) We set a plan to expand the original implementation to other Zone Equipment. I learned about OA control issues when I did Unit Ventilator because I had to allow controlling the OA amount for different OA control options available.

mjwitte:
@Nigusse Thanks for the responses.  Then I'm OK with removing the transition rule for this

*But* I would like to see a warning from the zone equipment if it is connected to a mixer (ATMixerExists is true) and the design OA flow rate is zero.

Nigusse:
@mjwitte @Myoldmopar Will add a warning message in the get input routine for each of the zone HVAC equipment connected to DOAS.

Nigusse:
@mjwitte @Myoldmopar Below is a sample warning error messege when ATMixerExists is true and the design OA flow rate is zero.

   ** Warning ** GetPTUnit: ZoneHVAC:PackagedTerminalAirConditioner="SPACE1-1 PTAC" input value to Cooling Outdoor Air Flow Rate = 0.0 
   **   ~~~   ** Cooling Outdoor Air Flow Rate must be > 0 or set to autosize when ZoneHVAC:PackagedTerminalAirConditioner is connected to DOAS. 
   ** Warning ** GetPTUnit: ZoneHVAC:PackagedTerminalAirConditioner="SPACE1-1 PTAC" input value to Heating Outdoor Air Flow Rate = 0.0 
   **   ~~~   ** Heating Outdoor Air Flow Rate must be > 0 or set to autosize when ZoneHVAC:PackagedTerminalAirConditioner is connected to DOAS. 
   ** Warning ** GetPTUnit: ZoneHVAC:PackagedTerminalAirConditioner="SPACE1-1 PTAC" input value to No Load Outdoor Air Flow Rate = 0.0 
   **   ~~~   ** No Load Outdoor Air Flow Rate must be > 0 or set to autosize when ZoneHVAC:PackagedTerminalAirConditioner is connected to DOAS.

mjwitte:
@Nigusse The warning is a good start.  I think it needs to reference the mixer object name (assuming that's available, or at least the mixer object type).  Something like:
** Warning ** GetPTUnit: ZoneHVAC:PackagedTerminalAirConditioner="SPACE1-1 PTAC" input value to Cooling Outdoor Air Flow Rate = 0.0
** ~~~ ** and is connected to  AirTerminal:SingleDuct:Mixer = "SPACE1-1 DOAS MIXER".  Cooling Outdoor Air Flow Rate must be > 0 or set to autosize to get flow from the DOAS.

Maybe you can word this better.

Nigusse:
Added warning message Zone HVAC equipment: PTAC, PTHP, WSHP, VRF, and FanCoil when the OA amount is set to zero and terminal air mixer exists.  Resolved the merge conflict and did some cleanup. This is ready for review.

mjwitte:
@Nigusse @Myoldmopar Before we merge this, I'm going to reopen some discussion on who sets the OA flow rate here - sorry.  

Prior to this fix, the air loop connected to the AirTerminal:SingleDuct:Mixer set the OA flow rate at each mixer based on the total system flow divided by the number of mixers.

This fix flips things around so that the OA flow rate set by the zone equipment (e.g. fan coil unit) is the OA flow rate at the AirTerminal:SingleDuct:Mixer object, and the airloop flow will be the sum of the mixer flows.

There is other work in discussion which proposes to add variable OA flow controls for this system configuration.  This raises the question of whether the mixer should be the one controlling the OA flow, instead of the zone equipment.

Myoldmopar:
Alright.  @mjwitte given that there is other work in discussion, would it be unfortunately best to not make this change now?  So we avoid going in one direction just to flip-flop back the next release?  There are still a few weeks before IO freeze, so we do have time to chat.  Maybe our planning meeting this week will help clarify what's coming in OA land.  Maybe we need to utilize the sizing call this week to really dive in and decide?

If we do choose to get this in, it's conflicted yet again, and this time it's not just the transition rules file, but something in the code itself.  There's also a build warning for a simple unused parameter that should be cleaned up.

mjwitte:
@Nigusse @lgu1234 Does #6223 make this PR obsolete?  If not, which one should drop in first?

Nigusse:
@mjwitte  @lgu1234 Yes, this PR is going to be obsolete, no need to update it. Most of the changes were flipping inputs data from the air loop to zone equipment. Gu has moved any relevant stuff to #6223.

mjwitte:
Superseded by #6223.

rraustad:
What's the $ value of a rejected effort?

Issue #6118, opened 2017-05-08
VRF Sizing Problem and PLR error


mjwitte:
Issue overview
--------------
User file with VRF has some problems:

1.  Hard flow rates specified in the VRF terminal unit do not get used in sizing the VRF coils - seems that they should.
2.  ~~With the terminal unit flow rates autosized, the simulation throws errors like this:~~ (in v8.6, not in v8.7)
```
   ** Warning ** ZoneHVAC:TerminalUnit:VariableRefrigerantFlow "THERMAL ZONE 1-6|VTU-1"
   **   ~~~   ** Terminal unit part-load ratio calculation failed: PLR limits of 0 to 1 exceeded
   **   ~~~   ** Please fill out a bug report and forward to the EnergyPlus support group.
```
3. And then it fails with a fatal error:
```
   ** Severe  ** PsyRhoAirFnPbTdbW: RhoAir (Density of Air) is calculated <= 0 [-410.64918].
   **   ~~~   ** pb =[99700.00], tdb=[-274.00], w=[9.9888827E-006].
   **   ~~~   **  Routine=Unknown, Environment=ANNUAL RUN FOR 2013, at Simulation time=09/05 07:00 - 07:01
   **  Fatal  ** Program terminates due to preceding condition.
```

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.7.0
 - Helpdesk ticket number 11995

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added to DevSupport\DefectFiles.  Use USA_CA_San.Diego-Montgomery.Field.722903_TMY3.epw
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
So what should we do here? Limit OA flow to max of cooling/heating? Raise max cooling/heating to allow user specified OA? Let sizing run as is and limit OA flow in calc routine? I guess it makes most sense to figure this out in the sizing routine so user can be warned of the change. I am leaning towards limiting OA flow to max of cooling/heating and if not autosized, warn user.

    ZoneHVAC:TerminalUnit:VariableRefrigerantFlow,
      Thermal Zone 1-3|VTU-1,  !- Zone Terminal Unit Name
      autosize,                !- Cooling Supply Air Flow Rate {m3/s}
[...]

PR #6121, opened 2017-05-09
Test that DX coil performance curves are near 1.0 at rated condition and VRF coils use terminal unit flow rate for sizing

rraustad:
Pull request overview
---------------------
Input files with bad performance curves for DX coils can cause odd results (e.g., very cold zone temperatures). Add test of performance curves and report warning when curve result is not near 1.0 at reference condition. Additionally, the VRF TU allows input for air flow rate, this value should also be used to size the VRF DX coils. This corrects issue #6118.

Note: now that performance curves are tested, some existing files may warn that performance curves are not normalized to 1 at the rating condition. This is not a cause for concern if the performance curves were normalized to an off-design rating point.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes: n/a


rraustad:
New warning message when performance curves are out of bounds:

    ** Warning ** GetDXCoils: Coil:Cooling:DX:VariableRefrigerantFlow="THERMAL ZONE 1-3|VTU-1|VRF COOLING COIL"
    **   ~~~   ** ...Cooling Capacity Ratio Modifier Function of Temperature Curve Name output is not equal to 1.0 (+ or - 10%) at reference conditions.
    **   ~~~   ** ...Curve output at reference conditions = 23.000

This CapFT curve multiplied the DX coil capacity 23 fold and resulted in outlet air temps of -272 C.

Existing warning message attempted to direct the user to this coil but it really did not help diagnose the problem.

    ** Warning ** Coil:Cooling:DX:VariableRefrigerantFlow "THERMAL ZONE 1-3|VTU-1|VRF COOLING COIL" - Full load outlet air dry-bulb temperature < 2C. This indicates the possibility of coil frost/freeze. Outlet temperature = -17.77 C.
    **   ~~~   **  ...Occurrence info = SAN DIEGO MONTGOMER ANN CLG 1% CONDNS DB=>MWB, 08/21 07:00 - 07:01
    **   ~~~   ** ... Possible reasons for low outlet air dry-bulb temperatures are: This DX coil
    **   ~~~   **    1) may have a low inlet air dry-bulb temperature. Inlet air temperature = 22.830 C.
    **   ~~~   **    2) may have a low air flow rate per watt of cooling capacity. Check inputs.
    **   ~~~   **    3) is used as part of a HX assisted cooling coil which uses a high sensible effectiveness. Check inputs.

Not sure what else if anything can be done for this type of input error.

rraustad:
Now there are lots of these. Either we never knew we were using a bad curve, or the curve is good and something else is amiss.

    +   ** Warning ** GetDXCoils: Coil:Heating:DX:SingleSpeed="HEAT PUMP DX HEATING COIL 1"
    +   **   ~~~   ** ...Energy Input Ratio Function of Temperature Curve Name output is not equal to 1.0 (+ or - 10%) at reference conditions.
    +   **   ~~~   ** ...Curve output at reference conditions = 0.819

rraustad:
Something was amiss. It's not the curves it's the programming. Cubic and Quadratic are only a function of OAT. Copy-paste and forgot to remove Indoor T and the curve manager used that value.

    EIR = DXCoil( DXCoilNum ).RatedEIR( Mode ) * EIRTempModFac * EIRFlowModFac;

    if ( ( DXCoil( DXCoilNum ).EIRTempModFacCurveType( 1 ) == Quadratic ) || ( DXCoil( DXCoilNum ).EIRTempModFacCurveType( 1 ) == Cubic ) ) {
        EIRTempModFac = CurveValue( DXCoil( DXCoilNum ).EIRFTemp( Mode ), OutdoorDryBulb );
    } else if ( DXCoil( DXCoilNum ).EIRTempModFacCurveType( 1 ) == BiQuadratic ) {
        EIRTempModFac = CurveValue( DXCoil( DXCoilNum ).EIRFTemp( Mode ), InletAirDryBulbTemp, OutdoorDryBulb );
    } else {


    if ( SELECT_CASE_var == "QUADRATIC" ) {
        CurveVal = CurveValue( DXCoil( DXCoilNum ).CCapFTemp( 1 )), RatedInletAirTempHeat, RatedOutdoorAirTempHeat );
    } else if ( SELECT_CASE_var == "BIQUADRATIC" ) {
        CurveVal = CurveValue( DXCoil( DXCoilNum ).CCapFTemp( 1 ), RatedInletAirTempHeat, RatedOutdoorAirTempHeat );
    } else if ( SELECT_CASE_var == "CUBIC" ) {
        CurveVal = CurveValue( DXCoil( DXCoilNum ).CCapFTemp( 1 )), RatedInletAirTempHeat, RatedOutdoorAirTempHeat );

dxheatperfcurves

src/EnergyPlus/DXCoils.cc
```diff
@@ -2283,10 +2287,10 @@ namespace DXCoils {
 				{ auto const SELECT_CASE_var( GetCurveType( DXCoil( DXCoilNum ).PLFFPLR( 1 ) ) );
 
 				if ( SELECT_CASE_var == "QUADRATIC" ) {
-					// curve limits at rating point tested below
+					// curve ouput tested below
```
rraustad: sowy about my lisp, I ouput incowectly

rraustad:
I guess it just figures that fixing this VRF problem would highlight issues with widely used performance curves ONLY IN VRF input files!

rraustad:
Well, I see where the VRF TU CapFT curves came from and the curve coefficients match the spreadsheet. What I don't understand is why Mitsubishi used 16C as the reference point (Figure 1 [here](http://www.fsec.ucf.edu/en/Publications/pdf/FSEC-CR-1910-12.pdf)). The Daikin table data in Figure 2 shows 19C as the reference point with 33.5 kW cooling capacity reference (grey highlight in Fig 2). To fix this I would need to go back and adjust the performance curves to use the reference capacity at 19.4C IWB.

vrftucapft

mjwitte:
@rraustad So, what's the plan here.  Are we going to leave these curves as-is, knowing that the rated capacity is off a bit?  Is there not a standard rating condition for VRF units?

src/EnergyPlus/DXCoils.cc
```diff
@@ -3942,7 +4142,7 @@ namespace DXCoils {
 						CurveVal = CurveValue( DXCoil( DXCoilNum ).MSCCapFTemp( I ), RatedInletWetBulbTemp, RatedOutdoorAirTemp );
 						if ( CurveVal > 1.10 || CurveVal < 0.90 ) {
 							ShowWarningError( RoutineName + CurrentModuleObject + "=\"" + DXCoil( DXCoilNum ).Name + "\", curve values" );
-							ShowContinueError( "..." + cAlphaFields( 13 + ( I - 1 ) * 6 ) + " output is not equal to 1.0 (+ or - 10%) at rated conditions." );
+							ShowContinueError( "..." + cAlphaFields( 13 + ( I - 1 ) * 6 ) + " output is not equal to 1.0 (+ or - 10%) at reference conditions." );
```
mjwitte: @rraustad Why is this one reference instead of rated?  It's using fixed Rated temps in the curve call.

src/EnergyPlus/DXCoils.cc
```diff
@@ -3972,6 +4172,7 @@ namespace DXCoils {
 						CurveVal = CurveValue( DXCoil( DXCoilNum ).MSCCapFFlow( I ), 1.0 );
 						if ( CurveVal > 1.10 || CurveVal < 0.90 ) {
 							ShowWarningError( RoutineName + CurrentModuleObject + "=\"" + DXCoil( DXCoilNum ).Name + "\", curve values" );
+							ShowContinueError( cAlphaFields( 14 + ( I - 1 ) * 6 ) + " = " + Alphas( 14 + ( I - 1 ) * 6 ) );
```
mjwitte: Field name is redundant here and in the next line.  This line is nice, but then I'd remove the field name in the next line.

There's so many of these, all using about six lines each.  Seems ripe for a separate function.
```
CheckCuveAtRatedConditions( RoutineName + CurrentModuleObject, DXCoil( DXCoilNum ).Name, DXCoil( DXCoilNum ).MSCCapFFFlow, cAlphaFields( . . . ), X1, X2-optional, X3-optional );
```

src/EnergyPlus/DXCoils.cc
```diff
@@ -12836,6 +13108,71 @@ Label50: ;
...
+		// SUBROUTINE INFORMATION:
+		//       AUTHOR         Richard Raustad
+		//       DATE WRITTEN   May 2017
+
+		// PURPOSE OF THIS SUBROUTINE:
+		// This subroutine gets a name for a given DX Coil -- issues error message if that
+		// DX Coil is not a legal DX Coil.
+
+		// METHODOLOGY EMPLOYED:
+		// na
```
mjwitte: Remove this and other unused comment blocks below.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -5408,7 +5529,10 @@ namespace HVACVariableRefrigerantFlow {
 		}
 
 		if ( CurZoneEqNum > 0 ) {
-			ZoneEqSizing( CurZoneEqNum ).AirVolFlow = max( VRFTU( VRFTUNum ).MaxCoolAirVolFlow, VRFTU( VRFTUNum ).MaxHeatAirVolFlow );
+			ZoneEqSizing( CurZoneEqNum ).CoolingAirFlow = true;
+			ZoneEqSizing( CurZoneEqNum ).CoolingAirVolFlow = VRFTU( VRFTUNum ).MaxCoolAirVolFlow;
+			ZoneEqSizing( CurZoneEqNum ).HeatingAirFlow = true;
+			ZoneEqSizing( CurZoneEqNum ).HeatingAirVolFlow = VRFTU( VRFTUNum ).MaxHeatAirVolFlow;
```
mjwitte, line 5510: Looks like there are more changes here than just adding a bunch of curve value checks?  Please describe this and any other changes in the PR synopsis.

mjwitte: And change the title.

rraustad: Changes in response to: MJWitte [wrote](https://github.com/NREL/EnergyPlus/issues/6118) - Ok, but what about item 1 above, the coil sizing when there's a hard flow rate in the terminal unit?

src/EnergyPlus/DXCoils.cc
rraustad: Existing checks of curves used "reference" in the message. I used "rated". I searched and replaced all with "rated".

rraustad:
Added new function with output as:

    ** Warning ** GetDXCoils: Coil:Cooling:DX:VariableRefrigerantFlow="TU1 VRF DX COOLING COIL curve values
    **   ~~~   ** ... Cooling Capacity Ratio Modifier Function of Temperature Curve Name = VRFTUCOOLCAPFT output is not equal to 1.0 (+ or - 10%) at rated conditions.
    **   ~~~   ** ... Curve output at rated conditions = 1.136

Old warning shows:

    ** Warning ** GetDXCoils: Coil:Cooling:DX:VariableRefrigerantFlow="TU1 VRF DX COOLING COIL"
    **   ~~~   ** ...Cooling Capacity Ratio Modifier Function of Temperature Curve Name output is not equal to 1.0 (+ or - 10%) at rated conditions.
    **   ~~~   ** ...Curve output at rated conditions = 1.136

rraustad:
@Myoldmopar what do I need to do to get rid of optional "_" in my call? CurveValue call doesn't need it and can take up to 4 additional arguments? Maybe you need to wait to see my new function?

    if ( SELECT_CASE_var == "CUBIC" ) {
        checkCurveValue( RoutineName + CurrentModuleObject, DXCoil( DXCoilNum ).Name, DXCoil( DXCoilNum ).HCOPFAirFlow, cAlphaFields( 14 ), Alphas( 14 ), 1.0, _, _, _, _ );
    } else if ( SELECT_CASE_var == "QUADRATIC" ) {
        HeatCOPFAirFlow = CurveValue( DXCoil( DXCoilNum ).HCOPFAirFlow, 1.0 );

rraustad:
I addressed all comments, thank you Mike.

Re the new errors in VRF example files, those are valid curves just set up at off-design conditions. So I am not sure what to do about that. Add a disclaimer somewhere about off-design curves, re-engineer the example file curves?
That VRF system was deployed before AHRI 1230, but manufacturers were supposed to be using AHRI 210/240 at that time, so I am still confused why the rating point is at 16C and not 19.4C. I could fix the example file but that would take some time.

src/EnergyPlus/DXCoils.cc
```diff
@@ -3177,30 +3068,17 @@ namespace DXCoils {
 					// Verify Curve Object, only legal types are BiQuadratic or Cubic
 					{ auto const SELECT_CASE_var( GetCurveType( DXCoil( DXCoilNum ).HCOPFTemp ) );
 
-					if ( SELECT_CASE_var == "BIQUADRATIC" ) {
+					if ( SELECT_CASE_var == "BIQUADRATIC" || SELECT_CASE_var == "CUBIC" ) {
```
rraustad: oops! curve type is needed here

src/EnergyPlus/DXCoils.cc
```diff
@@ -12785,53 +12237,77 @@ Label50: ;
...
+		using InputProcessor::FindItemInList;
 
-		// SUBROUTINE LOCAL VARIABLE DECLARATIONS:
-		// na
 		if ( GetCoilsInputFlag ) {
 			GetDXCoils();
 			GetCoilsInputFlag = false;
 		}
 
-		DXCoilIndex = FindItemInList( DXCoilName, DXCoil );
```
rraustad, line 12823: Git just confused here, this line wasn't deleted

src/EnergyPlus/DXCoils.cc
```diff
@@ -12785,53 +12237,77 @@ Label50: ;
...
 		if ( DXCoilIndex == 0 ) {
 			if ( present( SuppressWarning ) ) {
 				//     No warning printed if only searching for the existence of a DX Coil
 			} else {
 				if ( present( ThisObjectType ) ) {
-					ShowSevereError( ThisObjectType + ", GetDXCoilIndex: DX Coil not found=" + DXCoilName );
+					ShowSevereError( ThisObjectType + ", GetDXCoilIndex: DX Coil not found " );
 				} else {
-					ShowSevereError( "GetDXCoilIndex: DX Coil not found=" + DXCoilName );
+					ShowSevereError( "GetDXCoilIndex: DX Coil not found " );
```
rraustad, line 12305: same here, similar lines in function GetDXCoilIndex and GetDXCoilName are getting munged.

src/EnergyPlus/CurveManager.hh
```diff
@@ -571,6 +571,20 @@ namespace CurveManager {
...
 
+	void
+	checkCurveValue(
+		std::string const callingRoutineObj, // calling routine with object type
+		std::string const objectName,     // parent object where curve is used
+		int const curveIndex,             // index to curve object
+		std::string const cFieldName,     // object field name
+		std::string const cFieldValue,    // user input curve name
+		Real64 const Var1,                // required 1st independent variable
+		Optional <Real64 const > Var2,    // 2nd independent variable
```
mjwitte: @rraustad I think you can avoid the need for trailing "_" in the calls by providing a default here
`Optional <Real64 const > Var2 = "_", // 2nd independent variable`
Like [CurveManager.hh](https://github.com/NREL/EnergyPlus/blob/173bb58bcc95ec005c91946ecee8f573224904fd/src/EnergyPlus/CurveManager.hh#L412-L418)

src/EnergyPlus/CurveManager.cc
```diff
@@ -6379,6 +6379,42 @@ Label999: ;
 		return CurveOrTableObjectTypeNum;
 	}
 
+	void
+	checkCurveIsNormalizedToOne( 
+		std::string const callingRoutineObj, // calling routine with object type
+		std::string const objectName,        // parent object where curve is used
+		int const curveIndex,                // index to curve object
+		std::string const cFieldName,        // object field name
+		std::string const cFieldValue,       // user input curve name
```
mjwitte, line 6388: This argument isn't really needed, since the curve name is known from the index, but probably not worth changing all of these now.  Otherwise, much better with this new function.

rraustad: Good point, it's only 68 places so if I do need another change, this one should be included.

mjwitte: **Only** 68?

Myoldmopar:
> @Myoldmopar what do I need to do to get rid of optional "_" in my call? CurveValue call doesn't need it and can take up to 4 additional arguments? Maybe you need to wait to see my new function?

Maybe it's just getting late, but I think you should be able to just not pass them in if they have default values in the function signature.

Issue #6092, opened 2017-04-06
Same idf file can produce different result - related to Fan:SystemModel


mjwitte:
Issue overview
--------------
User complained that two identical files were producing a different sizing result.  At first, could not reproduce this, but after repeated attempts, saw two different results show up for the same idf file.  Diffing the err file shows one additional error:
```
   ** Warning ** SizeUnitarySystem: Potential issue with equipment sizing for AirLoopHVAC:UnitarySystem CORE_RETAIL ASHP UNITARY
   **   ~~~   ** ...Rated Total Cooling Capacity = 0.00 [W]
   **   ~~~   ** ...Air flow rate used for sizing = 5.95662 [m3/s]
   **   ~~~   ** ...Outdoor air fraction used for sizing = 0.40
   **   ~~~   ** ...Coil inlet air temperature used for sizing = 29.11 [C]
   **   ~~~   ** ...Coil outlet air temperature used for sizing = 12.77 [C]
```
Diffing the eio files, show differences in component sizing for 
```
Coil:Cooling:DX:MultiSpeed, CORE_RETAIL ASHP COOLING COIL (cooling capacities and SHRs, no diffs in airflow rates)
AirLoopHVAC:UnitarySystem, CORE_RETAIL ASHP UNITARY (design size cooling and heating capacities, no diffs in airflow rates)
Coil:Heating:DX:MultiSpeed, CORE_RETAIL ASHP HEATING COIL (heating capacities, no diffs in airflow rates)
Coil:Heating:Electric, CORE_RETAIL ASHP SUPP HEATING COIL (heating capacity)
```

There are also diffs in the DX Heating Coil Standard Rating Information for the CORE_RETAIL coils which is to be expected.  But there is also a diff in the DX ratings for one of the other coils even though none of it's component sizing results are different:
```
! <DX Cooling Coil Standard Rating Information>, Component Type, Component Name, Standard Rating (Net) Cooling Capacity {W}, Standard Rated Net COP {W/W}, EER {Btu/W-h}, SEER {Btu/W-h}, IEER {Btu/W-h}
 DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:MultiSpeed, BACK_SPACE ASHP COOLING COIL, 35028.7,  ,  , 59.67,  
vs.
 DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:MultiSpeed, BACK_SPACE ASHP COOLING COIL, 35028.7,  ,  , 70.71,  
[...]

PR #6126, opened 2017-05-11
Fix seemingly random sizing results problem related to AirloopHVAC:UnitarySystem with Fan:SystemModel

mjwitte:
Pull request overview
---------------------
Addresses #6092 where the same file could produce different sizing results.  Problem was related to an array bounds violation when calculating the fan design heat gain when using Fan:SystemModel.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ReportSizingManager.cc
+++ b/src/EnergyPlus/ReportSizingManager.cc
@@ -1696,8 +1696,6 @@ namespace ReportSizingManager {
 							}
 							SupFanNum = PrimaryAirSystem( CurSysNum ).SupFanNum;
 							RetFanNum = PrimaryAirSystem( CurSysNum ).RetFanNum;
-							PsyCpAirFnWTdb( CoilOutHumRat, 0.5 * ( CoilInTemp + CoilOutTemp ) ); //? what is this doing, a function with no left hand side?
-							FanCoolLoad = FanDesHeatGain( SupFanNum, DesVolFlow ) + ( 1.0 - OutAirFrac ) * FanDesHeatGain( RetFanNum, DesVolFlow );
 							switch ( PrimaryAirSystem( CurSysNum ).supFanModelTypeEnum ){
 								case DataAirSystems::structArrayLegacyFanModels: {
 									FanCoolLoad = FanDesHeatGain( PrimaryAirSystem( CurSysNum ).SupFanNum, DesVolFlow );
--- a/src/EnergyPlus/HVACUnitarySystem.cc
+++ b/src/EnergyPlus/HVACUnitarySystem.cc
@@ -1884,7 +1884,13 @@ namespace HVACUnitarySystem {
 		SupFanNum = 0;
 
 		if ( CurSysNum > 0 && CurOASysNum == 0 && UnitarySystem( UnitarySysNum ).FanExists ) {
-			PrimaryAirSystem( CurSysNum ).SupFanNum = UnitarySystem( UnitarySysNum ).FanIndex;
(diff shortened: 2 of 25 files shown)


EnergyArchmage (approved):
looks good to me.  good catch.   Sorry this didn't get done right.

mjwitte:
@EnergyArchmage No big deal.  I'm just surprised that the example file(s) using Fan:SystemModel didn't fail in CI debug builds.  Maybe something to do with the mix of fan types and how many.  Anyways, jumped right out when running the defect file with a debug build.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1696,8 +1696,6 @@ namespace ReportSizingManager {
 							}
 							SupFanNum = PrimaryAirSystem( CurSysNum ).SupFanNum;
```
mjwitte, line 1697: The problem is deeper than this.  When the defect file fails here, CurSysNum=3 and SupFanNum = 2.  That shouldn't be possible, because the defect file has four airloops with AirloopHVAC:UnitarySystem which all use Fan:SystemModel.  It also has one ZoneHVAC:UnitHeater which uses Fan:ConstantVolume. That's why SupFanNum=2 exceeds the array bounds on Fan.  Earlier calls here with SupFanNum=0 are fine, because FanDesHeat just sets the fan heat to zero and returns.

So, anyway, SupFanNum should be zero here.  Need to track down why it isn't.

EnergyArchmage: note that for Fan:SystemModel, the indexing is zero-based, but of course for the old fans the indexing is 1-based.

mjwitte: SimAirServingZones::InitAirLoops sets supFanModelTypeEnum, and SupFanNum or supFanVecIndex when the supply fan is directly on the main branch [here](https://github.com/NREL/EnergyPlus/blob/173bb58bcc95ec005c91946ecee8f573224904fd/src/EnergyPlus/SimAirServingZones.cc#L1866-L1939).  (similar for return fan)

If the fan is a child of a unitary system, it is ignored here.  Only the unitary system tries to set this [here](https://github.com/NREL/EnergyPlus/blob/173bb58bcc95ec005c91946ecee8f573224904fd/src/EnergyPlus/HVACUnitarySystem.cc#L1886-L1888) in HVACUnitarySystem::SizeUnitarySystem, but it needs to be sensitive to the fan object type.  I can add that here, but there's still a question then about other unitarysystems.

The I/O Ref (v8.7 p. 645) says "Central supply and return fan temperature rise is taken into account in sizing the central cooling coils."  But it appears that is true only when the fan is directly on the branch or if the fan is inside an AirloopHVAC:UnitarySystem.  Do we need to add a similar block of code to other unitarysystem types?  And I wonder if the code in InitAirLoops gets the right fan for dual-fan DDVAV systems?  I guess this part should be a new issue.  Fix the unitary system here and leave the rest.

mjwitte: Forgot to include the link before to the code in [HVACUnitarySystem](https://github.com/NREL/EnergyPlus/blob/173bb58bcc95ec005c91946ecee8f573224904fd/src/EnergyPlus/HVACUnitarySystem.cc#L1886-L1888) that currently sets PrimaryAirLoop.SupFanNum.  This just needs to check the fan type and then set PrimaryAirLoop.supFanModelTypeEnum and either supFanVecIndex or SupFanNum.  Will push that up shortly.

nrel-bot-2:

nrel-bot-3:

Issue #5835, opened 2016-08-23
Sizing:System with FlowPerCoolingCapacity and CoolingDesignCapacity not working as expected


mjwitte:
## Issue overview

User file Sizing:System object has these inputs for system airflow and coil capacities:

```
    FlowPerCoolingCapacity,  !- Cooling Supply Air Flow Rate Method
    0,                       !- Cooling Supply Air Flow Rate {m3/s}
    0.00006041,              !- Cooling Supply Air Flow Rate Per Unit Cooling Capacity {m3/s-W}
    FlowPerHeatingCapacity,  !- Heating Supply Air Flow Rate Method
    0,                       !- Heating Supply Air Flow Rate {m3/s}
    0.00006041,              !- Heating Supply Air Flow Rate Per Unit Heating Capacity {m3/s-W}
    ZoneSum,                 !- System Outdoor Air Method
    1.0,                     !- Zone Maximum Outdoor Air Fraction {dimensionless}
    CoolingDesignCapacity,   !- Cooling Design Capacity Method
    12500,                   !- Cooling Design Capacity {W}
    HeatingDesignCapacity,   !- Heating Design Capacity Method
    14400,                   !- Heating Design Capacity {W}
    OnOff;                   !- Central Cooling Capacity Control Method

```

Expected airflow rate for this system is 0.00006041_14400=0.8699 for heating and 0.00006041_12500=0.755124 for cooling, but the eio reports:

```
 System Sizing Information, AC01-02, Calculated Cooling Design Air Flow Rate [m3/s], 0.23372
 System Sizing Information, AC01-02, User Cooling Design Air Flow Rate [m3/s], 0.19308
 System Sizing Information, AC01-02, Calculated Heating Design Air Flow Rate [m3/s], 0.24154
 System Sizing Information, AC01-02, User Heating Design Air Flow Rate [m3/s], 0.24722
```

Appears the the FlowPer...Capacity is not checking if the user also specified a hard value for heating/cooling capacity.
### Details

Some additional details for this issue (if relevant):
- Platform ()
- Version of EnergyPlus v8.5.0, didn't recheck with current develop, but expect the same
- Helpdesk ticket number11503
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [ ] Ticket added to Pivotal for defect (development team task)
[...]

PR #6127, opened 2017-05-12
Sizing:System with FlowPerCoolingCapacity and CoolingDesignCapacity not working as expected

Nigusse:
Issue #5835

System cooling air flow rate sizing using FlowPerCoolingCapacity method uses the Cooling Design Capacity user specified hard sized value; otherwise, defaults to auto-sized cooling capacity for scaling the system air flow rate.

Pull request overview
---------------------
The code change allows to check if there is a user specified hard sized cooling/heating capacity for scaling the system cooling/heating air flow rate. The unit test and defect file demonstrate the same.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)


Nigusse:
I see unhandled Failure for one of the CI machines [here](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-1dfb716194a4e8eb647ff94112c9fd2f20548c32-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html). This popped up after I initialized one variable. I am not sure what is causing it? @Myoldmopar @mjwitte

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -6314,28 +6315,33 @@ namespace SimAirServingZones {
 				CalcSysSizing( AirLoopNum ).InpDesCoolAirFlow = CalcSysSizing( AirLoopNum ).DesCoolVolFlow * FractionOfAutosize;
 				FinalSysSizing( AirLoopNum ).InpDesCoolAirFlow = FinalSysSizing( AirLoopNum ).DesCoolVolFlow * FractionOfAutosize;
 			} else if ( SELECT_CASE_var == FlowPerCoolingCapacity ) {
-				if ( FinalSysSizing( AirLoopNum ).CoolingCapMethod == FractionOfAutosizedCoolingCapacity ) {
+				if ( FinalSysSizing( AirLoopNum ).CoolingCapMethod == FractionOfAutosizedCoolingCapacity || FinalSysSizing( AirLoopNum ).ScaledCoolingCapacity == DataSizing::AutoSize ) {
```
rraustad: Do you even need the OR here? Just check if ScaledCoolingCapacity == AutoSize inside the IF and set to 1 if needed.

Nigusse: Yes. The capacity can be different from Autosized.  If the cooling capacity sizing method is FractionOfAutosizedCoolingCapacity, the ".ScaledCoolingCapacity" variable is filled with the fraction value. not "Autosized".

rraustad: Scaled cooling capacity is not an autosizeable field. Regardless, if CoolingCapMethod == FractionOfAutosizedCoolingCapacity you will execute the code, regardless of the ScaledCoolingCapacity value.

    N23, \field Fraction of Autosized Cooling Design Capacity
      \type real
      \minimum 0.0
      \note Enter the fraction of auto-sized cooling design capacity. Required field when the cooling
      \note design capacity method field is FractionOfAutosizedCoolingCapacity.

Nigusse: You are right. I do not need to check that.

Nigusse: Get input already checks the value of ".ScaledHeatingCapacity" and for this capacity sizing method it will not be set to autosize. I will make change if needed.

Nigusse: After reviewing the code, the check for ScaledCoolingCapacity == AutoSize is needed because this variable is set to Autosize if the cooling capacity is Autosized. The same is true for heating.

Nigusse:
Added missing flow per capacity flow sizing method when the capacity is sized using capacity per floor area.

nrel-bot-3:

nrel-bot-3:
@Nigusse @lgentile it has been 17 days since this pull request was last updated.

nrel-bot-3:

nrel-bot-3:

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -6355,28 +6363,35 @@ namespace SimAirServingZones {
...
+					CpAirStd = PsyCpAirFnWTdb( constant_zero, constant_twenty );
+					AutosizedCapacity = StdRhoAir * FinalSysSizing( AirLoopNum ).DesHeatVolFlow * CpAirStd * ( CoilOutTemp - CoilInTemp );
+					TempSize = FinalSysSizing( AirLoopNum ).FlowPerHeatingCapacity * AutosizedCapacity * FractionOfAutosize;
+				} else if ( FinalSysSizing( AirLoopNum ).HeatingCapMethod == HeatingDesignCapacity && FinalSysSizing( AirLoopNum ).ScaledHeatingCapacity > 0 ) {
+					TempSize = FinalSysSizing( AirLoopNum ).FlowPerHeatingCapacity * FinalSysSizing( AirLoopNum ).ScaledHeatingCapacity;
+				} else if ( FinalSysSizing( AirLoopNum ).HeatingCapMethod == CapacityPerFloorArea ) {
+					TempSize = FinalSysSizing( AirLoopNum ).FlowPerHeatingCapacity * FinalSysSizing( AirLoopNum ).ScaledHeatingCapacity * FinalSysSizing( AirLoopNum ).FloorAreaOnAirLoopCooled;
 				} else {
-					CoilInTemp = OutAirFrac * FinalSysSizing( AirLoopNum ).HeatOutTemp + ( 1.0 - OutAirFrac ) * FinalSysSizing( AirLoopNum ).HeatRetTemp;
+					// incorrect input
```
rraustad: Is it incorrect input? or should nothing be done? I assume GetInput will catch incorrect in;puts. So if the input is not what you check above then there is nothing else to do (except that these tests are && with other tests). So is everything caught here? In other words, will the else ever be executed? And what if it is?

    A10, \field Heating Design Capacity Method
      \type choice
      \key None
      \key HeatingDesignCapacity
      \key CapacityPerFloorArea
      \key FractionOfAutosizedHeatingCapacity

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -6355,28 +6363,35 @@ namespace SimAirServingZones {
 				CalcSysSizing( AirLoopNum ).InpDesHeatAirFlow = CalcSysSizing( AirLoopNum ).DesHeatVolFlow * FractionOfAutosize;
 				FinalSysSizing( AirLoopNum ).InpDesHeatAirFlow = FinalSysSizing( AirLoopNum ).DesHeatVolFlow * FractionOfAutosize;
 			} else if ( SELECT_CASE_var == FlowPerHeatingCapacity ) {
-				if ( FinalSysSizing( AirLoopNum ).HeatingCapMethod == FractionOfAutosizedHeatingCapacity ) {
-					FractionOfAutosize = FinalSysSizing( AirLoopNum ).ScaledHeatingCapacity;
-				}
-				if ( FinalSysSizing( AirLoopNum ).HeatOAOption == MinOA ) {
-					if ( FinalSysSizing( AirLoopNum ).DesHeatVolFlow > 0.0 ) {
-						OutAirFrac = FinalSysSizing( AirLoopNum ).DesOutAirVolFlow / FinalSysSizing( AirLoopNum ).DesHeatVolFlow;
+				if ( FinalSysSizing( AirLoopNum ).HeatingCapMethod == FractionOfAutosizedHeatingCapacity || FinalSysSizing( AirLoopNum ).ScaledHeatingCapacity == DataSizing::AutoSize ) {
```
rraustad: This is confusing. There are 2 tests here. One or the other or both could be true. If HeatingCapMethod != FractionOfAutosizedHeatingCapacity and ScaledHeatingCapacity == AutoSize how do you know what to do? What if HeatingCapMethod == CapacityPerFloorArea, these calculations don't look correct?

rraustad: It would be clearer if there were 3 tests, one for each input type. Then inside each test, you could test for other constraints and act accordingly.

src/EnergyPlus/SimAirServingZones.cc
Nigusse: Removed the else case and and added cases that were not captured previously.

src/EnergyPlus/SimAirServingZones.cc
Nigusse: Added a separate case for ScaledHeatingCapacity == AutoSize.

Nigusse:
@rraustad Addressed your review comments and merged in develop to the branch.

rraustad (approved):
Approved upon CI finishing.

rraustad:
Diff's are "invalid string position" and warmup convergence diff's.

rraustad:
CI reports no diff's except those that continuously show errors in DELight and Fin example files.

rraustad:
Reviewed and merged.

Issue #6119, opened 2017-05-08
Zone exhaust fan impacts zone air balance incorrectly


mjwitte:
Issue overview
--------------
Fan:ZoneExhaust should have no impact on the conditions or load in a zone that is served only by zone equipment (unless used with AirflowNetwork or ZoneAirMassFlowConservation).  The exhaust fan should simply consume electricity, and it is assumed that the user has specified infiltration or interzone mixing to offset the exhaust flow.

A user reports that a single zone model with four pipe fan coil (no outdoor air flow) and zone exhaust fan does not produce reasonable results.  With zone equipment, one would expect that the "balanced flow" schedule for the exhaust fan would have no impact, but it changes the results significantly.  There are three test files, one with no exhaust fan, one with exhaust fan and balanced flow=0, and another with exhaust fan and balanced flow = 1.  These *should* all produce the same results (other than the exhaust fan electricity), but they do not.

### Workaround ###
Delete the exhaust fan and use ElectricEquipment with 1.0 Fraction Lost to model the fan power.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.7.0
 - Helpdesk ticket number 11987

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added DevSupport\DefectFiles use Sacramento weather.
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

lgu1234:
@mjwitte I tested two files with an exhaust: one with exhaust fan and balanced flow=0, and another with exhaust fan and balanced flow = 1. When the schedule value > 0, the fan module calculates:

				if ( Fan( FanNum ).BalancedFractSchedNum > 0 ) {
					BalancedExhMassFlow = UnbalExhMassFlow * GetCurrentScheduleValue( Fan( FanNum ).BalancedFractSchedNum );
					UnbalExhMassFlow = UnbalExhMassFlow - BalancedExhMassFlow;

When BalancedExhMassFlow > 0, it becomes a part of ExhMassFlowRate in the CorrectZoneHumRat function of the ZoneTempPredictorCorrector
[...]

PR #6133, opened 2017-05-17
Corrected exhaust mass flow handling in CorrectZoneHumRat

lgu1234:
Pull request overview
---------------------
Addresses #6119.

The exhaust mass flow handling is corrected in CorrectZoneHumRat in ZoneTempPredictorCorrector.

Results from 3 test files have tiny differences.

Differences are expected in 5 example files due to column mismatch:

5ZoneAirCooled_ZoneAirMassFlowBalance
5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized
5ZoneEndUses
5ZoneGeometryTransform.
AdaptiveThermostatASH55_5Zone_Miami

After making column matched, results from above 5 example files are identical.
 
### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes n/a


lgu1234:
@mjwitte I just checked CI machine results and found 6 additional example files failed. I only expect failure of 5 example files due to column mismatch. Then I checked my results obtained from my local machine. There are no differences in 6 example files.

lgu1234:
@mjwitte Here are 6 additional file names:

DOASDualDuctSchool.csv

HospitalBaseline.csv

HospitalBaselineReheatReportEMS.csv

HospitalLowEnergy.csv
	
HybridVentilationControlGlobalSimple.csv

UserDefinedRoomAirPatterns.csv

nrel-bot-3:
@lgu1234 @lgentile it has been 16 days since this pull request was last updated.

nrel-bot-2:
@lgu1234 @lgentile it has been 15 days since this pull request was last updated.

nrel-bot-3:

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -5017,10 +5022,25 @@ namespace ZoneTempPredictorCorrector {
...
+					ExhMassFlowRate += Node( ZoneEquipConfig( ZoneEquipConfigNum ).ExhaustNode( NodeNum ) ).MassFlowRate / ZoneMult;
+				}
 			} // NodeNum
-			ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExhBalanced; // Balanced exhaust flow assumes there are other flows providing makeup air such as mixing or infiltration, so subtract it here
+			if ( !ExhaustNodeFlag ) {
+				ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExhBalanced; // Balanced exhaust flow assumes there are other flows providing makeup air such as mixing or infiltration, so subtract it here
+			}
 
 			if ( ZoneEquipConfig( ZoneEquipConfigNum ).ReturnAirNode > 0 ) {
 				TotExitMassFlowRate = ExhMassFlowRate + Node( ZoneEquipConfig( ZoneEquipConfigNum ).ReturnAirNode ).MassFlowRate / ZoneMult;
```
mjwitte: @lgu1234 I am not comfortable with simply ignoring the exhaust mass flow for zones without an air loop.  If this is the correct fix, then it seems the same changes are needed in [CorrectZoneContaminants](https://github.com/NREL/EnergyPlus/blob/9a786ca0a17b0eb97c4641f31663b44367a3d54f/src/EnergyPlus/ZoneContaminantPredictorCorrector.cc#L2141-L2147)  Hmmm, CorrectZoneContaminants doesn't have the adjustment for balanced exhaust either.

I question both of these functions in their use of TotExitMassFlowRate.  [CorrectZoneAirTemp](https://github.com/NREL/EnergyPlus/blob/9a786ca0a17b0eb97c4641f31663b44367a3d54f/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L4295-L4302) does not appear to use the leaving mass flow rate.  So, why is it part of the solution in CorrectZoneHumRat and CorrectZoneContaminants?  We assume conservation of mass within a zone, so only the incoming flows should be important.

lgu1234: @miwitte I think the fix is correct. You can see testing results to show correct outputs.

You are right, CorrectZoneContaminants should be modified in the same way. I think the fix should be verified first. If accepted, I am going to do the same way in CorrectZoneContaminants.

I think you also notice CorrectZoneAirTemp uses different approach to handle zone balance components and does not have this item. The component process is accomplished in the CalcZoneSums. I can not tell why from my memory. I notice the difference before this issue. I plan to create a new function to handle moisture component loads, similar to CalcZoneSums. It will take time and needs detailed consideration. I hope this will be a new feature for FY18 to ensure all component loads are mapped correctly in all zone variables, such as temperature, moisture, CO2, and contaminant.

mjwitte: @lgu1234 I ran the three defect files. The results are different from plain v8.7, but the three files still produce results that are different from each other.

If this fix proves to be correct, you don't need to add the extra loop here.  We already have a variable for ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExh which is the sum of the flows into all exhaust fans in the zone.  So, I think the equivalent to your fix would be to revert to original code and simply replace 
```
ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExhBalanced;
```
with
```
ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExh;
```
This would exclude all exhaust fan mass flows from ExhMassFlowRate regardless of balanced or unbalanced.

mjwitte:
@lgu1234 Re-ran the defect files and they show the correct results now.  My mistake earlier.  I still think this can be written more efficiently by using ZoneEquipConfig.ZoneExh directly instead of adding a new loop.

src/EnergyPlus/DataZoneEquipment.cc
```diff
@@ -635,6 +637,9 @@ namespace DataZoneEquipment {
 
 					} else if ( SELECT_CASE_var == "FAN:ZONEEXHAUST" ) {
 						ZoneEquipList( ControlledZoneNum ).EquipType_Num( ZoneEquipTypeNum ) = ZoneExhaustFan_Num;
+						ZoneEquipList( ControlledZoneNum ).EquipData( ZoneEquipTypeNum ).NumInlets = 1;
+						ZoneEquipList( ControlledZoneNum ).EquipData( ZoneEquipTypeNum ).InletNodeNums.allocate( ZoneEquipList( ControlledZoneNum ).EquipData( ZoneEquipTypeNum ).NumInlets = 1 );
+						ZoneEquipList( ControlledZoneNum ).EquipData( ZoneEquipTypeNum ).InletNodeNums( 1 ) = GetFanInletNode( "FAN:ZONEEXHAUST", ZoneEquipList( ControlledZoneNum ).EquipName( ZoneEquipTypeNum ), GetZoneEquipmentDataErrorsFound );
```
mjwitte: @lgu1234 I think the changes in this file are no longer needed.  And some related changes in other files can be removed.

lgu1234: @miwitte OK. I ma going to remove them.

tst/EnergyPlus/unit/ZoneTempPredictorCorrector.unit.cc
```diff
@@ -213,7 +215,7 @@ TEST_F( EnergyPlusFixture, ZoneTempPredictorCorrector_CorrectZoneHumRatTest )
 	MDotOA( 1 ) = 0.0;
 
 	CorrectZoneHumRat( 1, controlledZoneEquipConfigNums );
-	EXPECT_NEAR( 0.008, Node( 5 ).HumRat, 0.00001 );
+	EXPECT_NEAR( 0.0081218, Node( 5 ).HumRat, 0.00001 );
```
mjwitte: @lgu1234 Looking again at this test, if the zone starts at 0.008, and the incoming flows are 0.008, how is it possible for the zone to end at something 0.0081218?  It must be 0.008.  Same with the test below.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
```diff
@@ -2141,6 +2141,11 @@ namespace ZoneContaminantPredictorCorrector {
 				for ( NodeNum = 1; NodeNum <= ZoneEquipConfig( ZoneEquipConfigNum ).NumExhaustNodes; ++NodeNum ) {
 					ExhMassFlowRate += Node( ZoneEquipConfig( ZoneEquipConfigNum ).ExhaustNode( NodeNum ) ).MassFlowRate / ZoneMult;
 				} // NodeNum
+				if ( !( ZoneEquipConfig( ZoneEquipConfigNum ).AirLoopNum > 0 || ZoneAirMassFlow.EnforceZoneMassBalance || AirflowNetworkNumOfExhFan > 0 ) ) {
+					ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExh / ZoneMult; // Exclude all exhaust fan flow here
+				} else {
+					ExhMassFlowRate -= ZoneEquipConfig( ZoneEquipConfigNum ).ZoneExhBalanced / ZoneMult; // Balanced exhaust flow assumes there are other flows providing makeup air such as mixing or infiltration, so subtract it here
+				}
```
mjwitte: @lgu1234 Don't forget about contaminant corrector.

lgu1234: @mjwitte Sorry. I forgot it. I am going to make the same changes in the contaminant corrector.

mjwitte:
@lgu1234 So, lots of diffs from this.  Not surprised.  Have you looked closely at any of them?

lgu1234:
@mjwitte Before today's revision, there are 12 files with differences. Today's revision causes differences in 20 files. I am not surprised. I have not looked any files yet. The change I made today is fundamentally correct, because the mass flow rate used to calculate MoistureMassFlowRate in B should be the same used in A. The TotExitMassFlowRate may not be equal to the ZoneMassFlowRate sometimes.

Therefore, the first step is to verify my changes to ensure they are correct.

mjwitte:
@lgu1234 I tested the latest code with the defect files and with three variations on HVACTemplate-PTAC-DOAS and the results are identical with no exhaust fan, balanced exhaust fan, and unbalanced exhaust fan.  I can look at a couple of the diff files tomorrow just to be sure the changes look reasonable.

lgu1234:
@mjwitte I tested example files with differences on my local computer and took quick look at files using AirflowNetwork models. I think differences are reasonable so far.

lgu1234:
@mjwitte I also looked at results from two files with 5ZoneAirCooled_ZoneAirMassFlowBalance*. Results are reasonable. Then I asked Bereket to check results, since he is the person who developed the ZoneAirMassFlowBalance feature. He checked results from 3 runs: Ver. 8.7, Existing Develop, and My changes. The mass conservation is broken for all 3 runs.

mjwitte:
@lgu1234   I've tested annual runs 5ZoneAirCooled, HVACTemplate-5ZonePTAC-DOAS, and RefBldgHospitalNew2004_Chicago with current develop vs this branch.  As expected 5ZoneAirCooled is identical (no exhaust nodes in any zones).  HVACTemplate-5ZonePTAC-DOAS  (no exhaust fans, but exhaust nodes into the PTAC) and RefBldgHospitalNew2004_Chicago (zone equipment and exhaust fans) are also identical.  Also tested RefBldgSecondarySchoolNew2004_Chicago.  There are small differences, but nothing troubling (<0.03% change in annual energy use).  So, I'm ok to merge this.  Do you agree?

@Nigusse will post a new issue for the mass balance problems?

lgu1234:
@mjwitte Great to know testing results are OK. I agree. Please merge this.

Issue #6136, opened 2017-05-18
FourPipeBeamLargeOffice crashes when system sizing is turned off


mitchute:
I found this while working on #5120. 

Four pipe beam relies on system sizing being run before it is initialized. Error messages need to be added so it fails gracefully.

### Checklist
 - [x] Defect file added (FourPipeBeamLargeOffice.idf)
 - [x] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6137

PR #6137, opened 2017-05-18
Fix four pipe beam crash when system sizing is off

mitchute:
Pull request overview
---------------------
Addresses #6136. This adds an error message to the four pipe beam init so it will fail gracefully when system sizing is off.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2:
@mitchute @lgentile it has been 14 days since this pull request was last updated.

src/EnergyPlus/HVACFourPipeBeam.cc
```diff
@@ -771,22 +771,27 @@ namespace FourPipeBeam {
...
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).DesCoolVolFlow
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeCoolVDot );
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).DesHeatVolFlow
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeHeatVDot );
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).MassFlowAtCoolPeak
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeCoolVDot ) * DataEnvironment::StdRhoAir;
-
-			ReportSizingOutput( this->unitType, this->name, "AirLoopHVAC Design Supply Air Flow Rate Adjustment [m3/s]",
-								( this->vDotDesignPrimAir - originalTermUnitSizeMaxVDot ) );
+			if ( DataSizing::FinalSysSizing.size() > 0 ) {
```
Myoldmopar: I think you could get this from `DataSizing::SysSizingRunDone`.  @mjwitte that makes sense, right?

I'd **like** to consider it more robust in case those arrays ever get restructured.  In reality, it's probably of equal robustness, but from a code readability standpoint I'd prefer checking the flag.

src/EnergyPlus/HVACFourPipeBeam.cc
```diff
@@ -91,7 +91,7 @@ namespace FourPipeBeam {
 	///// Note use of shared_ptr here is not a good pattern, not to be replicated without further discussion.
 	std::shared_ptr< AirTerminalUnit >
 	HVACFourPipeBeam::fourPipeBeamFactory(
-		int EP_UNUSED(objectType),
+		int EP_UNUSED( objectType ),
```
Myoldmopar, line 94: Thanks for all the whitespace cleanup.  And thanks to GitHub for the `?w=1` query parameter on the URL so I can see the core changes easily :+1:.

Myoldmopar (changes requested):
Hey, let's figure out if we can use a flag for checking the system sizing instead of that one array size and then get this in.

src/EnergyPlus/HVACFourPipeBeam.cc
mjwitte: @mitchute @Myoldmopar @EnergyArchmage I would think that checking SysSizingRunDone should be adequate.  

But there's really no way to use this component without system sizing?  Is that clear in the docs?  Is there a way around it using some assumptions?  This is the only component like that (I think).

src/EnergyPlus/HVACFourPipeBeam.cc
```diff
@@ -771,22 +771,27 @@ namespace FourPipeBeam {
...
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).DesCoolVolFlow
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeCoolVDot );
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).DesHeatVolFlow
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeHeatVDot );
-			DataSizing::FinalSysSizing( DataZoneEquipment::ZoneEquipConfig(CurZoneEqNum).AirLoopNum ).MassFlowAtCoolPeak
-				+= ( this->vDotDesignPrimAir - originalTermUnitSizeCoolVDot ) * DataEnvironment::StdRhoAir;
-
-			ReportSizingOutput( this->unitType, this->name, "AirLoopHVAC Design Supply Air Flow Rate Adjustment [m3/s]",
-								( this->vDotDesignPrimAir - originalTermUnitSizeMaxVDot ) );
+			if ( DataSizing::SysSizingRunDone ) {
```
mitchute, line 778: @Myoldmopar this has been updated.

Myoldmopar (approved):

Myoldmopar:
Hmm.  Weird voodoo in the Linux build over here too, @lgu1234.  It looks like it is nothing you're doing over in #6114.

Anyway, it's unrelated to this work, so I'm inclined to merge it in.  Hopefully it won't show up in develop, but if it does...there will be a greater urgency to address it :+1:

Issue #6082, opened 2017-03-24
SolarCollector:UnglazedTranspired setting exterior temp to 0 when no flow


rmcclung:
When using SolarCollectro:UnglazedTranspired to preheat out door air for an air handler, and that air handler has no OA flow (at night, for example), the Solar Collector Air temperatures are set to zero, which sets the temperature of the associated wall surface to zero instead of the outdoor air temp, which increases/decreases heat loss through the assembly.

*See #6139 for more details. The heat transfer was correct but the zero output was misleading.*

EnergyArchmage:
The surface heat transfer should be switching to use the passive model when flow is zero.

mjwitte:
@rmcclung Do you have a test file that shows a problem here?  There is a pending change to the code, #6139, that sets the leaving air temperature from the collector to the average ambient instead of zero when the collector is off.  But this isn't changing any surface temperatures, because the passive baffle code was already using the ambient conditions.  I just want to be sure the fix is covering everything it needs to.
(fyi, @Nigusse )

rmcclung:
Hi Micheal,

Unfortunately, I’ve made a lot of changes to that file since, I can’t seem to reproduce the problem anymore.

Ruth

mjwitte:
Closed via #6139

PR #6139, opened 2017-05-18
Fixed SolarCollector:UnglazedTranspired reporting for leaving air temperature when no flow

Nigusse:
Pull request overview
---------------------
Addresses Issue #6082.
Previously when there was no airflow in the collector, the reported value for "Solar Collector Leaving Air Temperature" was zero.  Now it is reported as the average ambient temperature for the collector surface(s).  The passive collector heat balance remains unchanged.  This was just a reporting change.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - [x] Doc changes, build ok.


Nigusse:
Diffs was expected in example file, TranspiredCollectors.idf but output variable Solar Collector Leaving Air Temperature [C] only showed diffs. The diffs occurred because of changing the initialization of variable UTSC.TairHX to ambient temperature instead of zero.  This diff occurred only when the collector is off; otherwise, the UTSC.TairHX was already set to ambient temperature when the collector is active. 

Initialization of variables UTSC.Tplen and UTSC.Tcoll were changed from zeros to their saved previous time step values. But the later changes did not bring any diffs in the result to a surprise because these two variables were already re-set to their saved previous time step values in calc routines just before their use. So, changing the initialization of UTSC.Tplen and UTSC.Tcoll in InitTranspiredCollector routine won't make any diffs.

src/EnergyPlus/TranspiredCollector.cc
```diff
@@ -1112,6 +1131,7 @@ namespace TranspiredCollector {
 		// USE STATEMENTS:
 
 		// Using/Aliasing
+		using DataEnvironment::IsRain;
```
mjwitte: Unused.

src/EnergyPlus/TranspiredCollector.cc
```diff
@@ -1172,7 +1192,7 @@ namespace TranspiredCollector {
 		UTSC( UTSCNum ).Tcoll = TmpTscoll;
 		UTSC( UTSCNum ).HrPlen = HrPlen;
 		UTSC( UTSCNum ).HcPlen = HcPlen;
-		UTSC( UTSCNum ).TairHX = 0.0;
+		UTSC( UTSCNum ).TairHX = Tamb;
```
mjwitte, line 1191: OK here, but a few lines further down we have `UTSC( UTSCNum ).SupOutEnth = OutEnthalpy` which is grabbing from DataEvnironment.  Shouldn't we calculate an` OutEnthalpyAmb` using `Tamb` and `OutHumRatAmb`?

mjwitte:
Tested with example file, changes look reasonable.  Although the end result is only a reporting change in the Solar Collector Leaving Air Temperature which was zero before.  There is no change in surface temperature, because it appears that CalcPassiveExteriorBaffleGap was already doing the right thing.  @EnergyArchmage would appreciate your review.

EnergyArchmage:
The code changes look fine.  But this seems like a lot of trouble for a non-issue.  The system node conditions don't matter when there is no air system flow.   The original post was not correct about a zero affecting surface temperatures.  There are many places in EnergyPlus where system node conditions do not make sense during zero flow.

mjwitte:
I agree that this wasn't as serious as first presented, but we do put non-zero properties on system nodes in general even when there is no flow.  Looking deeper at this, I see that the system node temp and humidity for the nodes leaving the collectors have not changed  (they weren't zero before), only the reporting for Solar Collector Leaving Air Temperature.  But the System Node Temperaure at the corresponding node doesn't match.  They are close, and follow a similar pattern, but they always differ by as much as 2+C.  Should they be the same?

EnergyArchmage:
oh right, no they shouldn't be the same.  I was thinking wrong, the term HX leaving air is coming off the transpired collector into the gap air balance and is not a system node.  It is an intermediate model value of interest to solar collector modelling.  That term doesn't exist in the passive case so it is not clear what it should it be.  maybe -9999.0

Nigusse:
The reason that I assigned to OA temperature is that when there is no "air flow" (in the passive case) the temperatures that is more representative (or closer) is that of inlet condition. This is analogous to setting outlet conditions of air and water coils to its inlet condition when there is no flow. What it means is that when there is no flow the inlet conditions does not change. Is it it more realistic to setting to arbitrary value of -9999.0?

mjwitte:
Oh, I see, the system node temp matches the collector "Plenum Air Temperature".  So, even if we wanted to calculate the collector air temperature during no flow, the value being computed in this fix, average of the outdoor air temps, isn't correct.  If the surface is in the sun, the temperature inside would be warmer than outdoor air.

We can leave it as-is, reporting zeros and update the I/O ref description to state that it reports zero when there is no flow.
Or we can change to report -999 and update the I/O ref description.
Or set this equal to the Collector Surface Temperature instead - and update the I/O ref.

mjwitte:
@Nigusse Well, setting it to the inlet condition is valid as well.  If @EnergyArchmage we can keep this, but we still should explain it in the I/O Ref description for Solar Collector Leaving Air Temperature.

Nigusse:
@mjwitte @EnergyArchmage Let me know what value to set to and I will update the I/O reference accordingly.

EnergyArchmage:
I suppose that the best value would be the plenum air temperature modeled for the passive gap.  The I/O should explain that there is a difference in the model for the value depending on with and with/out flow.

Nigusse:
@mjwitte @EnergyArchmage I looked the code again. Setting UTSC( UTSCNum ).TairHX (which is the collector outlet air temperature) to Tamb is correct instead of setting it to zero because even for passive case Tamb is not actually passed to CalcPassiveExteriorBaffleGap routine but Tamb (set locally) is used when called by CalcPassiveTranspiredCollector.  The Engineering reference equation 19.307 (V8.7) also confirms that in passive case the pelnum entering air temperature is Tamb. This is the "CalcPassiveExteriorBaffleGap" model assumption. Setting TairHX = 0 is the issue requested here. 

But what Brent is suggesting, which I agree with, is that for passive case the UTSC( UTSCNum ).SupOutTemp should be set to TmpTaPlen instead of Tamb. This is not what the user requested in this issue but it can be addressed.

Nigusse:
Now the UTSC( UTSCNum ).SupOutTemp  is set to TmpTaPlen and UTSC( UTSCNum ).TairHX is set to Tamb when the transpired collector is in passive operation mode. Also updated the I/O ref documentation accordingly.

src/EnergyPlus/TranspiredCollector.cc
```diff
@@ -1172,14 +1188,14 @@ namespace TranspiredCollector {
...
 		UTSC( UTSCNum ).InletMDot = 0.0;
 		UTSC( UTSCNum ).InletTempDB = Tamb;
 		UTSC( UTSCNum ).Vsuction = 0.0;
 		UTSC( UTSCNum ).PlenumVelocity = 0.0;
-		UTSC( UTSCNum ).SupOutTemp = Tamb;
+		UTSC( UTSCNum ).SupOutTemp = TmpTaPlen;
 		UTSC( UTSCNum ).SupOutHumRat = OutHumRatAmb;
-		UTSC( UTSCNum ).SupOutEnth = OutEnthalpy;
+		UTSC( UTSCNum ).SupOutEnth = PsyHFnTdbW( Tamb, OutHumRatAmb );
```
mjwitte: I think this needs to be TmpTaPlen now instead of Tamb?  And is OutHumRatAmb appropriate?  I suppose that makes sense here.

Issue #6112, opened 2017-05-02
Missing output row in SQL database


shorowit:
E+ 8.7 on Windows

The E+ html file includes 21 rows in the Component Sizing Information table, where row 20 is:
`20 	Boiler:HotWater 	RESIDENTIAL BOILER ELECTRIC 	User-Specified Nominal Capacity [W] 	26352.97405`
whereas the SQL file only has 20 rows and is missing this output.

IDF file attached.

[in.idf.txt](https://github.com/NREL/EnergyPlus/files/970829/in.idf.txt)

mjwitte:
@shorowit @macumber Actually, there are lots of missing entries in the sql component sizing table.  This particular file only has one missing value, but most files that have user-specified sizes will have dual reporting for nearly every autosized value (design size and user-specified).  For example, 5ZoneAutoDXVAV, the first part of the component sizing table looks like this:
![image](https://user-images.githubusercontent.com/4991075/26946677-baa941e8-4c55-11e7-98ac-c1ff211c34d4.png)

Currently, only the 5 "Design Size" columns of this table are being written to sql.  With the fix, all of the Design Size and User-Specified values will be written to sql.  Is this going to cause problems in reading this table?

macumber:
@DavidGoldwasser @asparke2

shorowit:
I can't speak for others but it won't be a problem for me.

DavidGoldwasser:
Shouldn't cause an issue for me unless the names of existing columns change, if that happens and I know it, I can update the reporting measure if needed.

mjwitte:
@DavidGoldwasser OK, will merge this then.  Just to be clear, all existing rows in this table will still be there, but the additional user-defined values that will now appear in the table are the ones to grab if you're trying to report the sizing that was actually used in the simulation.  I will add an explanation of this to the output changes doc.

asparke2:
As long as the existing column headings in the sql file don't change I won't have any issue.  If they do change, it would be nice to have it documented so I can preemptively correct any code that uses those names.

mjwitte:
@asparke2 This table is organized differently in the sql file than it appears in the table file.  The sql file has a separate record for every value in the table, so the generic sql column headings won't change, but there will be additional records that were previously omitted.
![image](https://user-images.githubusercontent.com/4991075/26983820-b10182a6-4d02-11e7-9b60-812c4e91d090.png)
[...]

PR #6147, opened 2017-05-23
Addressed missing component sizing output row in SQL database 

Nigusse:
Issue #6112

Addressed Report Sizing Output does not always report/tabulate user specified sizing variable to sql output file.

Pull request overview
---------------------
This is a defect. Report Sizing Output does not always report/tabulate user specified sizing variable to sql output file. This fix may produce diffs in sql output file. Defect file has been added.

### Work Checklist
  
### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes n/a

### DIFF
--- a/src/EnergyPlus/ReportSizingManager.cc
+++ b/src/EnergyPlus/ReportSizingManager.cc
@@ -185,7 +185,9 @@ namespace ReportSizingManager {
 
 		// add to SQL output
 		if ( sqlite ) sqlite->addSQLiteComponentSizingRecord( CompType, CompName, VarDesc, VarValue );
-
+		if ( present( UsrDesc ) && present( UsrValue ) ) {
+			if ( sqlite ) sqlite->addSQLiteComponentSizingRecord( CompType, CompName, UsrDesc, UsrValue );
+		}
 	}
 
 	void
--- a/tst/EnergyPlus/unit/ReportSizingManager.unit.cc
+++ b/tst/EnergyPlus/unit/ReportSizingManager.unit.cc
@@ -53,10 +53,8 @@
 #include <ObjexxFCL/gio.hh>
 
 #include "Fixtures/EnergyPlusFixture.hh"
-#include "Fixtures/SQLiteFixture.hh"
 
 // EnergyPlus Headers
-#include <EnergyPlus/Boilers.hh>
 #include <EnergyPlus/DataEnvironment.hh>
 #include <EnergyPlus/DataPrecisionGlobals.hh>
 #include <EnergyPlus/DataSizing.hh>
@@ -67,11 +65,9 @@
 #include <EnergyPlus/DataZoneEquipment.hh>
 #include <EnergyPlus/Psychrometrics.hh>
 #include <EnergyPlus/ReportSizingManager.hh>
-//#include <SQLiteProcedures.hh>;
 
 using namespace EnergyPlus;
 using namespace ObjexxFCL;
-using namespace EnergyPlus::Boilers;
 using namespace EnergyPlus::DataAirSystems;
 using namespace EnergyPlus::DataGlobals;
 using namespace EnergyPlus::DataEnvironment;
@@ -81,7 +77,6 @@ using namespace EnergyPlus::DataZoneEquipment;
 using namespace EnergyPlus::Fans;
 using namespace EnergyPlus::Psychrometrics;
 using namespace EnergyPlus::ReportSizingManager;
-//using namespace EnergyPlus::SQLiteProcedures;
 
 TEST_F( EnergyPlusFixture, ReportSizingManager_GetCoilDesFlowT )
 {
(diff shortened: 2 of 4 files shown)


tst/EnergyPlus/unit/ReportSizingManager.unit.cc
```diff
@@ -353,3 +353,25 @@ TEST_F( EnergyPlusFixture, ReportSizingManager_RequestSizingZone ) {
...
+	// input values
+	CompType = "BOILER:HOTWATER";
+	CompName = "RESIDENTIAL BOILER ELECTRIC";
+	VarDesc = "Design Size Nominal Capacity[W]";
+	VarValue = 105977.98934;
+	UsrDesc = "User-Specified Nominal Capacity[W]";
+	UsrValue = 26352.97405;
+
+	// boiler hot water autosizing and userspecified nominal capacity reporting
+	ReportSizingManager::ReportSizingOutput( CompType, CompName, VarDesc, VarValue, UsrDesc, UsrValue );
```
mjwitte: This unit test doesn't really check anything.  Perhaps you can follow the patterns in SQLite.unit.cc to actually test what was written?

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -185,7 +185,9 @@ namespace ReportSizingManager {
 
 		// add to SQL output
 		if ( sqlite ) sqlite->addSQLiteComponentSizingRecord( CompType, CompName, VarDesc, VarValue );
-
+		if ( present( UsrDesc ) && present( UsrValue ) ) {
+			if ( sqlite ) sqlite->addSQLiteComponentSizingRecord( CompType, CompName, UsrDesc, UsrValue );
```
mjwitte, line 189: This works as expected, for most files it will results in many more entries for this table in sql when design size and user-specified values are present.  For example file 5ZoneAutoDXVAV, this table in sql grows from 41 rows to 53 rows.  Checking with @shorowit and @macumber to be sure this won't break things that read this output.

tst/EnergyPlus/unit/ReportSizingManager.unit.cc
Nigusse: @mjwitte Thanks. I will add a unit test to SQLite.unit.cc.

mjwitte: My intent here was to add to the unit test in reportsizingmanager.unit.  Would it work to combine the last half of what you added to sqlite.unit with what you have here?  As it is now, neither test is really checking to be sure that reportsizingmanager wrote both sets of information to the sql output.

Also, please add a description of this change to \src\Transition\OutputRulesFiles\OutputChanges8-7-0-to-8-7-0.md  This will be a new file.  You can use the 8-6-0-to-8-7-0 as a starting point.

Nigusse: Not sure if it can be done. I tried a similar approach in my original unit test creation but unsuccessful. I will try a second attempt.

mjwitte: Oh, ok.  Maybe @mbadams5 can help here.  The call to ReportSizingManager in this unit test creates some sql records.  How can this test check to see if those records are actually written as expected?

mjwitte: @Nigusse Of course, this test will need to set the sqlite flag to true.  But then there may need to be some more effort to set up the sqlite output file.  So, maybe this should be using SQLiteFixture here since it's not really relying on any general EnergyPlusFixture stuff.

mbadams5: First, `SQLiteFixture` derives from `EnergyPlusFixture` so it'll have everything it has and more.

Next, if you look in OutputProcessor.unit.cc there are examples of how to call functions that write to sqlite within them. [For example...](https://github.com/NREL/EnergyPlus/blob/develop/tst/EnergyPlus/unit/OutputProcessor.unit.cc#L1001-L1003)

The key take away from that is that you need to move the unit test's sqlite object to the one in the EnergyPlus namespace and back again.

Once you have done this, you will be able to query the sqlite records just like your SQLite unit test you created (and thus combine the two unit tests).

Nigusse: @mbadams5 Thanks. Combined the unit test based on the example file. @mjwitte Also added new out put change file \src\Transition\OutputRulesFiles\OutputChanges8-7-0-to-8-8-0.md.

mjwitte:
@Nigusse Combined unit test looks good.  I made some edits to the output changes description.  Will wait for one or two CI reports then merge.

Nigusse:
@mjwitte Thanks for the review and following up.

PR #6157, opened 2017-06-01
Output units refactoring

JasonGlazer:
Refactor the SetUpOutputVariable call and all related functions, so that units are stored in a separate data field from the variable name, and passed as an integer or enum value instead of a string. The corresponding string in International System of Units (SI units), string in Imperial Units (IP units), and conversion shall all be predefined. 

The unit type designator may be implemented to have multiple "flavors" for the same unit (for example, Watts), such as electricW, refrigerationW, etc. for cases where there is more than one IP unit for a given SI unit.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
@Myoldmopar I have some small diffs in the pull request but I have made *no* code changes and the code is consistent with develop. Weird.

mjwitte:
@JasonGlazer These delight files have been showing diffs like this on every branch for a few weeks - haven't figure out why yet.

mjwitte:
And only on Windows

nrel-bot-3:

JasonGlazer:
Looking over the code change again, there should be no differences at all. I'm going to remerge and see if it was some CI timing issue.

JasonGlazer:
DElight cases failing. I am ignoring these since they seem to be happening on the Windows CI tests lately at random.

JasonGlazer:
I was hoping for zero diffs but found a few that seem unrelated to the changes made:

- regression.DElight-Detailed-Comparison shows warmup convergence diffs in the EIO file and HTML files under Win64
- AirloopHVAC_VentilationRateProcedure, AirloopHVAC_ZoneSumTest, and FinAndOverhangCount unit tests show an "invalid string position" error under i386 but this cannot be reproduced locally when running these tests in the full suite of unit tests or run individually.

Given that these diffs have been seen on other pull requests that are unrelated to this, I am going to assume this is no significant diffs.

jasondegraw:
@JasonGlazer Could you rebuild the unit tests in Release mode and try that? I have not been able to reproduce all of the unit test failures locally, but I can reproduce some in Release mode.

JasonGlazer:
@jasondegraw Interesting, when ran the unit tests in Release mode each of those three unit tests still passed when run individually. When I ran the entire set of unit tests it tripped on AirTerminalSingleDuctMixer_SimPTAC_HeatingCoilTest

jasondegraw:
@JasonGlazer Yes, that's what I'm seeing as well (both on Windows and Linux). The DElight issues are still a bit of a mystery.

mjwitte:
So, maybe there's a clear_state omission in the test prior to AirTerminalSingleDuctMixer_SimPTAC_HeatingCoilTest?

jasondegraw:
That is entirely possible, but we probably need to start with this unit test. Writing an email now...

JasonGlazer:
I just checked and the same unit test fails when running the develop branch.

JasonGlazer:
Those three other unit test failures showing on the i386 runs still do not show up when run locally.

jasondegraw:
@JasonGlazer Bereket's branch fixes the problems that are reproducible that I know of on the Win64 side. I'll try to see what is going on with the i386 side of things if I can get the CI to stop deleting things out from under me. I'm pretty sure this is a develop problem and not a problem on this branch.

This might not be the best place for this discussion, but one test that I am seeing fail intermittently is FinAndOverhangCount, which it looks like you wrote. Is there any file IO going on in that test? File IO seems to be a common link between a number of the failing tests.

JasonGlazer:
@jasondegraw FixAndOverhangCount doesn't have any I/O but it is one of the unit tests that has a long string of IDF and uses process_idf() and calls ManageSimulation(). Other unit tests do the same so I'm not sure what is up.

nrel-bot-3:
@JasonGlazer @lgentile it has been 16 days since this pull request was last updated.

JasonGlazer:
MacOS and Linux both show the same 11 files with RDD diffs. From MacOS:

5ZoneSteamBaseboard
- -HVAC,Average,Heating Coil Steam Mass Flow Rate [Kg/s]
- +HVAC,Average,Heating Coil Steam Mass Flow Rate [kg/s]
- Case change from Kg/s to kg/s.

CommonPipe_Pri-Sec
- Same as 5ZoneSteamBaseboard

EMSConstantVolumePurchasedAir
- -Output:Variable,*,Erl Zn001:Wall001:Win001 Incident Angle,hourly; !- Zone Average [DEG]
- +Output:Variable,*,Erl Zn001:Wall001:Win001 Incident Angle,hourly; !- Zone Average [deg]
- Case change for DEG to deg

EMSWindowShadeControl
- Same as EMSConstantVolumePurchasedAir

HospitalBaselineReheatReportEMS
- -Output:Variable,*,Flr_1_Cafe_DT2,hourly; !- Zone Average [DELTAC]
- +Output:Variable,*,Flr_1_Cafe_DT2,hourly; !- Zone Average [deltaC]
- Case change for DELTAC to deltaC

PackagedTerminalAirConditioner
- Same as 5ZoneSteamBaseboard

PackagedTerminalAirConditionerVSAS
- Same as 5ZoneSteamBaseboard

RefrigeratedWarehouse
- -HVAC,Average,Refrigeration Zone Air Chiller Frost Accumulation Mass [Kg]
- +HVAC,Average,Refrigeration Zone Air Chiller Frost Accumulation Mass [kg]
- Case change from Kg/s to kg/s.

SteamSystemAutoSize
- Same as 5ZoneSteamBaseboard

TermReheatPri-SecLoop
- Same as 5ZoneSteamBaseboard

VariableRefrigerantFlow_FluidTCtrl_HR_5Zone
- -HVAC,Average,VRF Heat Pump Simultaneous Cooling and Heating Efficiency [Btu/h/W]
- +HVAC,Average,VRF Heat Pump Simultaneous Cooling and Heating Efficiency [Btu/h-W]
- Correct mistyped units from Btu/h/W to Btu/h-W

Overall, these are all either case changes or in one case fixes to the units. Otherwise no diffs.

I will still be changing all Unit::deg to Unit::degree, and Unit::ACH to Unit::ach which will causes lots of other RDD diffs so this is probably the build with the fewest CI diffs. In addition, some other kg/s cases will be show up because of an extra enum covering them.

Myoldmopar:
Thanks @JasonGlazer for the layout of the current diffs.  It will be very easy to verify and get it merged in once you turn back on all the bigger diffs.

mjwitte:
> I will still be changing all Unit::deg to Unit::degree

@JasonGlazer We want the reverse here - change degree to deg.

JasonGlazer:
@Myoldmopar @mjwitte I just did the documentation changes as well as the final code changes. Remember these code changes will result in more RDD diffs because they removed some redundant units that I was leaving just to reduce the number of diffs.

doc/module-developer/src/output/how-do-i-output-my-variables.tex
```diff
@@ -6,35 +6,34 @@ \section{How Do I Output My Variables?}\label{how-do-i-output-my-variables}
...
-~~~~~~~~~~~~~~~~~~~~~~~~ ResourceTypeKey,EndUseKey,GroupKey)
 
 Interface statements allow for the same call to be used for either real or integer ``ActualVariable'' variables.~ A few examples from EnergyPlus and then we will define the arguments:
 
-CALL SetupOutputVariable(`Site Outdoor Drybulb Temperature {[}C{]}', \&
+\begin{lstlisting}
 
-~~~~~~~ OutDryBulbTemp,`Zone',`Average',`Environment')
+SetupOutputVariable( "Site Outdoor Air Drybulb Temperature", OutputProcessor::Unit::C, OutDryBulbTemp, 
+  "Zone", "Average", "Environment" );
```
Myoldmopar, line 23: OMG `OutputProcessor::Unit::C` is so beautiful.

doc/module-developer/src/output/output-variable-dos-and-don-ts.tex
```diff
@@ -31,7 +31,7 @@ \subsection{Output Variable Naming Conventions}\label{output-variable-naming-con
   Output variable names shall always be less than 100 characters long.
 \end{itemize}
 
-An output variable name should be constructed from up to six separate name elements.~ The following diagram shows the elements and the order that they are to be used in the name.~ The first three are each somewhat optional and are used as needed to identify the type of engineering model, object, or device being reported on.~ The last three are nearly always needed. Units are always needed (as described below.
+An output variable name should be constructed from up to five separate name elements.~ The following diagram shows the elements and the order that they are to be used in the name.~ The first three are each somewhat optional and are used as needed to identify the type of engineering model, object, or device being reported on.~ The last two are nearly always needed. .
```
Myoldmopar, line 34: Stray period at the end of this paragraph; fix it if more commits are needed, not worth fixing if there's no other commits.

doc/module-developer/src/output/output-variable-dos-and-don-ts.tex
```diff
@@ -49,11 +49,9 @@ \subsection{Output Variable Naming Conventions}\label{output-variable-naming-con
...
+In some situations, a model may lend itself to generating a series of related output variables that depend on input.~ For example a surface heat transfer model may have results for numerous individual nodes or cells across the thickness of the surfaces.~ The number of nodes in the model will vary at runtime depending on the Construction.~ When this is the case, the standard approach used in EnergyPlus is to produce a series of output variables by enumerating the output variable name (rather than the key value) to generate a specific name for each instance.~ The generated name element for each instance in the series is appended to the end of the name and should end in a number. For example, the third node in a stratified tank model is called ``Chilled Water Thermal Storage Temperature Node 3 {[}C{]}.''~ This is done to facilitate wildcarding when postprocessing output.
```
Myoldmopar, line 52: For reference, there's no need to keep using the tildes at the end of each sentence in future code.  It's valid LaTeX, and part of the translation, but not necessary.

Myoldmopar (approved):
Major effort, I'm pumped about getting rid of the ambiguity when writing the variable name string in setup output variable calls.  This is much more formal, standardized, and long desired.  So thanks.

The code changes look great.  The CI results look fine.  I believe you have resolved the review comments, including the late decision to stick with "deg" not "degree" according to the enum in the OutputProcessor.

I have nothing further.  @mjwitte if you think it's good merge at will, otherwise I'll wait for comments and merge sometime tomorrow.

JasonGlazer:
I spot checked the huge number of RDD diffs and did not see anything I wasn't expecting. I think this is good to go.

mjwitte:
@JasonGlazer Looks good.

@Myoldmopar Because of this, but not part of this, I noticed . . .

I wasn't expecting the new [degree/deg] variables to show up in every rdd file.  I thought they would only appear if the new SurfaceProperty objects were used.  I guess the surface and zone wind direction make sense and have values regardless of the SurfaceProperty objects.

But I would suggest that System Node Wind Direction and Wind Speed should only be reported for outdoor air nodes.  

** Either way, the Wind Speed is wrong for outdoor air nodes with a height specified - it should vary with height, but a test with nodes at 1m and 100m show the same wind speed.  I'll post a new issue for that after I hear back from you on whether to report the wind data for all nodes.

Also, I was surprised to see Surface Outside Normal Azimuth Angle even when the new rotation object isn't used.  Not a strong feeling on this.

** But I think the units are wrong for this - it says it's [rad] but the values sure look like [deg].

Myoldmopar:
@mjwitte Are you proposing a follow-up issue to tackle those cleanups?  And have this go in?  I can open an issue with the following:

- Only report these for outdoor air nodes:
  - System Node Wind Direction
  - System Node WInd Speed
  - Surface Outside Normal Azimuth Angle   -- I think this can stay in actually even for static surfaces.
- Fix units on Surface Outside Normal Azimuth Angle
- Wind Speed doesn't vary with height -- did you verify this with a test file that includes the height variation input object?

Myoldmopar:
Alright, I'm working on those followup things with an issue (#6304).  I'll merge this in now.

JasonGlazer:
@mjwitte and @Myoldmopar I was thinking about the SetupOutputVariable calls and I think we should think about replacing the rest of the string arguments that are essentially other enumerations:

```
void
SetupOutputVariable(
	std::string const & VariableName, // String Name of variable (with units)
	OutputProcessor::Unit const & VariableUnit, // Actual units corresponding to the actual variable
	Real64 & ActualVariable, // Actual Variable, used to set up pointer
	std::string const & IndexTypeKey, // Zone, HeatBalance=1, HVAC, System, Plant=2
	std::string const & VariableTypeKey, // State, Average=1, NonState, Sum=2
	std::string const & KeyedValue, // Associated Key for this variable
	Optional_string_const ReportFreq = _, // Internal use -- causes reporting at this freqency
	Optional_string_const ResourceTypeKey = _, // Meter Resource Type (Electricity, Gas, etc)
	Optional_string_const EndUseKey = _, // Meter End Use Key (Lights, Heating, Cooling, etc)
	Optional_string_const EndUseSubKey = _, // Meter End Use Sub Key (General Lights, Task Lights, etc)
	Optional_string_const GroupKey = _, // Meter Super Group Key (Building, System, Plant)
	Optional_string_const ZoneKey = _, // Meter Zone Key (zone name)
	Optional_int_const ZoneMult = _, // Zone Multiplier, defaults to 1
	Optional_int_const ZoneListMult = _, // Zone List Multiplier, defaults to 1
	Optional_int_const indexGroupKey = _, // Group identifier for SQL output
	Optional_string_const customUnitName = _ // the custom name for the units from EMS definition of units
);

```

Which includes the IndexTypeKey and VariableTypeKey and possibly ReportFreq, ResourceTypeKey, EndUseKey, GroupKey.

What do you think?

Myoldmopar:
I would :heart: this.

JasonGlazer:
I just added a new issue related to this #6328

JasonGlazer:
I just added a new issue related to this #6328

Issue #6148, opened 2017-05-24
Problems with control of water coils in outdoor air stream


mjwitte:
Issue overview
--------------
This is related to #5695 which was addressed by #5973.

1. User file has hot water preheat coil in the outdoor air stream with economizer active.  Sometimes the coil controls well and at other times it does not.  If the economizer is turned off, then the coil controls exactly as it should.  I suspect these is a problem with the way that timing of the call to the coil controller.  For OA system controllers, this may need a different approach that what is used for controllers on the main air loop branch.

2. Also, I am not sure that the fix in #5973 is general enough.  It appears to disable lockout for all coils on a given airloop simply if these is a hot water coil in the outdoor air system.  If there are also chilled water coils, it seems this fix could prevent lockout when desired.  This needs review and may require additional variables to track lockout individually by coil or for the coil controller to check the lockout status the way that it is done for electric and gas coils.

3. User file has chilled water pre-cool coil in the outdoor air stream.  With differential dry-bulb economizer, coil operates as expected.  With differential enthalpy, it does not.  See ticket 12003 for more details.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus 8.7.0
 - helpdesk ticket number 12001 and 12003.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [x] Ticket added to Pivotal for defect (development team task)
 - [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
This looks promising. Basically what I did was figure out what the water flow rate needed to be at the same time the water coil is simulated. I also bypassed ControlCompOutput and found the answer using RegulaFalsi. This may be the start of a method to replace ControlCompOutput.

    SimOAComponent(
    ..
    } else if ( SELECT_CASE_var == WaterCoil_SimpleHeat ) { // 'Coil:Heating:Water')
        if ( Sim ) {
            if ( CompIndex > 0 ) {
                CalcHotWaterSimpleCoilWaterFlowRate( CompName, FirstHVACIteration, CompIndex );
                ControllerProps( WaterCoil( CompIndex ).ControllerIndex ).BypassControllerCalc = true;
            }
            SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
        }
        OAHeatingCoil = true;

[...]

Issue #5884, opened 2016-09-27
Repeated ControlCompOutput Max Iterations Exceeded for Rad/Conv HW Baseboard


RKStrand:
## Issue overview

When checking in the new simple chilled ceiling model, it was noted that the input files were resulting in over 30 ControlCompOutput errors for max iterations were exceeded for the annual run of the main simple chilled ceiling model.  The problem appears, however, to be completely unrelated to the chilled panel model though because when the panel is turned off the error messages do not disappear.  While the size of the loads for which the baseboard has trouble converging are all small, it seems like this should not be happening and needs to be investigated further to discover the problem and correct it as needed.  The problem can been seen in the 5ZoneCoolingPanelBaseboard.idf file in the test suite.
### Details

Some additional details for this issue (if relevant):
- Version of EnergyPlus: V8.6
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ x ] Defect file added (already existing in the test suite 5ZoneCoolingPanelBaseboard.idf)
- [x] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte Just ran this using release version 8.7...and I'm not seeing these error messages.  Did they magically disappear or am I missing something here?

RKStrand:
@mjwitte Nevermind...forgot that it was something that happened during an annual run.  I'm seeing the issue now.  On another note, I see that this is part of GitHub Issue #5823.  I think that is my mistake since based on the number the one you entered would have gone in first.  Could we maybe keep this as a separate issue and eliminate if from 5823? What do you think?

RKStrand:
@mjwitte Two things going on here.  First, the tolerance is set somewhat low and there are some cases when the flow rate is fairly low that are just barely missing the cut-off.  Second, the rest of the cases where max iterations is encountered, the flow rate is incredibly low--just above 10-9 kg/s which translates to roughly 1mg/hour.  That is really low and is a result of a choice in the code to set the tolerance for flow at 10-9.  So, many of these cases have such a low flow rate that it isn't even really possible.

So, here is what I am doing to resolve this:
1. Adjust the input file tolerance on not just the cooling panel file but all of the other files containing baseboard to 0.002 from 0.001.  The choice of 0.001 is not based on anything other than a guess so adjusting this probably won't have much of an impact.
[...]

PR #6176, opened 2017-06-14
DO NOT MERGE: Correct operational control of Outdoor Air System water coils

rraustad:
Pull request overview
---------------------
Water coils in the outdoor air system are locked out when economizer operation is active. This seems too restrictive since these coils are controlled by the outlet air temperature with respect to a user supplied set point temperature. This change addresses issue #6148.

The outdoor air coils are now allowed to operate regardless of economizer operation and it is the users responsibility to correctly control these coils. CHANGE THIS AND OTHER COMMENTS BASED ON FINAL ASSESSMENT.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/MixedAir.cc
+++ b/src/EnergyPlus/MixedAir.cc
@@ -64,6 +64,7 @@
 #include <DataEnvironment.hh>
 #include <DataHeatBalance.hh>
 #include <DataHeatBalFanSys.hh>
+#include <DataHVACControllers.hh>
 #include <DataIPShortCuts.hh>
 #include <DataLoopNode.hh>
 #include <DataPrecisionGlobals.hh>
@@ -81,6 +82,7 @@
 #include <GeneralRoutines.hh>
 #include <HeatingCoils.hh>
 #include <HeatRecovery.hh>
+#include <HVACControllers.hh>
 #include <HVACDXHeatPumpSystem.hh>
 #include <HVACDXSystem.hh>
 #include <HVACHXAssistedCoolingCoil.hh>
@@ -622,8 +624,17 @@ namespace MixedAir {
 		using HVACUnitarySystem::GetUnitarySystemOAHeatCoolCoil;
 		using HVACUnitarySystem::CheckUnitarySysCoilInOASysExists;
 		using Humidifiers::SimHumidifier;
+		using HVACControllers::ManageControllers;
+		using WaterCoils::WaterCoil;
+		using WaterCoils::CalcWaterCoilWaterFlowRate;
+		using HVACControllers::ControllerProps;
+
 		// Locals
 		// SUBROUTINE ARGUMENTS:
+		std::string const ControllerName( "DontCare" );
+		int AirLoopPass;
+		bool ControllerConvergedFlag;
+		bool IsUpToDateFlag;
 
 		// SUBROUTINE PARAMETER DEFINITIONS: None
 
@@ -638,6 +649,9 @@ namespace MixedAir {
 		OAHX = false;
 		Real64 AirloopPLR;
 		int FanOpMode;
+		AirLoopPass = 1;
+		ControllerConvergedFlag = false;
+		IsUpToDateFlag = false;
 
 		{ auto const SELECT_CASE_var( CompTypeNum );
 
@@ -672,11 +686,19 @@ namespace MixedAir {
 			// Coil Types
 		} else if ( SELECT_CASE_var == WaterCoil_Cooling ) { // 'Coil:Cooling:Water'
 			if ( Sim ) {
+				if ( CompIndex > 0 ) {
+					CalcWaterCoilWaterFlowRate( CompName, FirstHVACIteration, CompIndex );
+					ControllerProps( WaterCoil( CompIndex ).ControllerIndex ).BypassControllerCalc = true;
+				}
 				SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
 			}
 			OACoolingCoil = true;
 		} else if ( SELECT_CASE_var == WaterCoil_SimpleHeat ) { // 'Coil:Heating:Water')
 			if ( Sim ) {
+				if ( CompIndex > 0 ) {
+					CalcWaterCoilWaterFlowRate( CompName, FirstHVACIteration, CompIndex );
+					ControllerProps( WaterCoil( CompIndex ).ControllerIndex ).BypassControllerCalc = true;
+				}
 				SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
 			}
 			OAHeatingCoil = true;
@@ -687,6 +709,10 @@ namespace MixedAir {
 			OAHeatingCoil = true;
 		} else if ( SELECT_CASE_var == WaterCoil_DetailedCool ) { // 'Coil:Cooling:Water:DetailedGeometry'
 			if ( Sim ) {
+				if ( CompIndex > 0 ) {
+					CalcWaterCoilWaterFlowRate( CompName, FirstHVACIteration, CompIndex );
+					ControllerProps( WaterCoil( CompIndex ).ControllerIndex ).BypassControllerCalc = true;
+				}
 				SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
 			}
 			OACoolingCoil = true;
--- a/src/EnergyPlus/SimAirServingZones.cc
+++ b/src/EnergyPlus/SimAirServingZones.cc
@@ -1109,15 +1109,15 @@ namespace SimAirServingZones {
 								}
 							}
 						}
-						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
-						for (OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum) {
-							for (OACompNum = 1; OACompNum <= OutsideAirSys( OASysNum ).NumComponents; ++OACompNum) {
-								CompType = OutsideAirSys( AirSysNum ).ComponentType( OACompNum );
-								if (SameString( CompType, "Coil:Heating:Water" )) {
-									PrimaryAirSystem( AirSysNum ).CanBeLockedOutByEcono( OASysControllerNum ) = false;
-								}
-							}
-						}
+//						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
+//						for (OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum) {
+//							for (OACompNum = 1; OACompNum <= OutsideAirSys( OASysNum ).NumComponents; ++OACompNum) {
+//								CompType = OutsideAirSys( AirSysNum ).ComponentType( OACompNum );
+//								if (SameString( CompType, "Coil:Heating:Water" )) {
+//									PrimaryAirSystem( AirSysNum ).CanBeLockedOutByEcono( OASysControllerNum ) = false;
+//								}
+//							}
+//						}
 					}
 				}
 			}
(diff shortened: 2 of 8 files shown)


src/EnergyPlus/WaterCoils.cc
```diff
@@ -6377,6 +6382,185 @@ Label10: ;
...
+		mdotLow = mdot;
+		SimulateWaterCoilComponents( CoilName, FirstHVACIteration, CoilNum );
+
+		if ( WaterCoil( CoilNum ).WaterCoilType_Num == WaterCoil_SimpleHeating ) { // heating coil
+			deltaT = TempSetPoint - Node( AirOutletNode ).Temp; // positive result means coil is on
+		} else {
+			deltaT = Node( AirOutletNode ).Temp - TempSetPoint; // positive result means coil is on
+		}
+
+// *** NEED TO FIGURE OUT LOCKOUT LOGIC for non-OA system coils ( and OA system coils if they can be locked out ) *** and bypass next calculations if in lockout mode //
```
rraustad: @mjwitte when should OA System water coils, or any coil in the OA System for that matter, be locked out?

src/EnergyPlus/MixedAir.cc
```diff
@@ -672,11 +686,19 @@ namespace MixedAir {
 			// Coil Types
 		} else if ( SELECT_CASE_var == WaterCoil_Cooling ) { // 'Coil:Cooling:Water'
 			if ( Sim ) {
```
rraustad, line 688: switched from ControlCompOutput to RegulaFalsi for water coils in the outside air system

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2945,6 +2948,10 @@ namespace SimAirServingZones {
 			if ( QActual > 0.0 ) CoolingActive = true; // determine if coil is ON
 
 		} else if ( SELECT_CASE_var == WaterCoil_SimpleHeat ) { // 'Coil:Heating:Water'
```
rraustad, line 2950: and switched from ControlCompOutput to RegulaFalsi for water coils in the main air loop

src/EnergyPlus/WaterCoils.cc
```diff
@@ -993,6 +996,13 @@ namespace WaterCoils {
 			PlantLoopScanFlag( CoilNum ) = false;
 		}
 		if ( ! SysSizingCalc && MySizeFlag( CoilNum ) ) {
+
+			// get water coil controller index
+			errFlag = false;
+			GetControllerIndex( WaterCoil( CoilNum ).WaterInletNodeNum, WaterCoil( CoilNum ).ControllerIndex, errFlag );
+			GetControllerName( WaterCoil( CoilNum ).ControllerIndex, WaterCoil( CoilNum ).ControllerName, errFlag );
+
```
rraustad: I did this so I could call ManageControllers from where the coils are simulated. Don't need this anymore.

rraustad: Well, might need GetControllerIndex until ControlCompOutput is completely removed (needed to set bypass controller calcs flag).

rraustad:
@Myoldmopar @nealkruis can you fix this so when reviewing files that fail, when you back out you are at the file you started from? When I click "failed" in the Test Results, then click "Regression Results" for a specific file, and then review some err or eio files, when I click the back button twice I have to start all over. Sometimes while reviewing differences I forget where I'm at and can't remember which file I was reviewing so I can run that example file or which files I have already reviewed so I can continue my review. It would be nice to back out and be right where I started the review for a specific file. It would also be nice if the file name could be displayed on the differences page (err, eio, csv, etc.) so we can tell which file we are reviewing.

Myoldmopar:
@lefticus can you comment on the level of effort on these 2 requests?

 - From a single test run page, with the accordion list of pass/fail/etc., when you click on a file, and then come back to the accordion list, he would like to be in the same spot as before.  @rraustad in my reviewing, I've developed muscle memory to just open a new tab for each file I review, but I do understand your situation, as it took several frustrating attempts while developing this habit.
 - On the detail page where the uploaded s3 assets are posted, it would be nice to list the filename at the top of that page so you don't have to track which file you clicked on to get there.

Feel free to email me separate if you'd like to have a longer conversation than appropriate for commenting on this PR.

rraustad:
I think I have everything I need to eliminate HVACControllers (if we so choose) and I wanted to get this code over where I need it. Then I looked at it...and looked at it....and, huh?

Who remembers why this was added, the autosizing of the water controller accuracy. Other models typically use 0.001, but here that accuracy is calculated as a very small number. Do we really need to get that close? High air flow system would still be within 0.001 C of set point, this isn't good enough? This value is in degrees C. It doesn't say in the notes but the units for 2100 appear to be kg/m3. Doing the math on the first line of code below, the units for offset (accuracy) is WC/kg ??? 
Offset = ( 0.001 C / ( 2100 kg/m3 * m3 ) ) * ( W)
@EnergyArchmage @mjwitte @Myoldmopar

    Controller:WaterCoil,
       Autosize,                !- Controller Convergence Tolerance {deltaC}


    if ( ControllerProps( ControlNum ).Offset == AutoSize ) {
        // 2100 = 0.5 * 4.2 * 1000/1.2 * 1.2 where 0.5 is the ratio of chilled water delta T to supply air delta T,
        //   4.2 is the ratio of water density to air density, 1000/1.2 is the ratio of water specific heat to
        //   air specific heat, and 1.2 converts the result from air volumetric flow rate to air mass flow rate.
        //   The assumption is that a temperatute tolerance of 0.001 C is good for an air mass flow rate of 1 kg/s.
        //   So we divide .001 by the air mass flow rate estimated from the water volumetric flow rate to come up
        //   with a temperature tolerance that won't exceed the loop energy error tolerance (10 W).
        // Finally we need to take into account the fact that somebody might change the energy tolerance.
        ControllerProps( ControlNum ).Offset = ( 0.001 / ( 2100.0 * max( ControllerProps( ControlNum ).MaxVolFlowActuated, SmallWaterVolFlow ) ) ) * ( HVACEnergyToler / 10.0 );
        // do not let the controller tolerance exceed 1/10 of the loop temperature tolerance.
        ControllerProps( ControlNum ).Offset = min( 0.1 * HVACTemperatureToler, ControllerProps( ControlNum ).Offset );
        ReportSizingOutput( ControllerProps( ControlNum ).ControllerType, ControllerProps( ControlNum ).ControllerName, "Controller Convergence Tolerance", ControllerProps( ControlNum ).Offset );
    }


For the defect file:

     Component Sizing Information, Controller:WaterCoil, CONTROLLER WATER COIL 1, Controller Convergence Tolerance, 4.76190E-005
     Component Sizing Information, Controller:WaterCoil, CONTROLLER WATER COIL 2, Controller Convergence Tolerance, 9.42188E-005

rraustad:
Well this is interesting, 5ZoneAirCooledWithCoupledInGradeSlab has max iteration warnings. It's not that it can't find the answer, it's that RegulaFalsi approaches the answer in very small steps. So the 200 max iteration limit was insufficient (and excessive in my opinion). Now if RegulaFalsi could somehow switch between secant and bisection, then it could find the answer much more quickly. Probably within 10's of iterations. What does surprise me a little is that the water coil controller here has a 0.002 C tolerance and somehow ControlCompOutput is able to converge on that solution quicker than RegulaFalsi? It seems the relationship between water coil SAT and water flow rate is extremely non-linear and this confuses RegulaFalsi. In this case, at this condition, bisection does a better job?

rraustad:
@lefticus one more thing that would make reviewing really quick. If there were a master err and eio file that had all files with err differences included in that one file. We could open the master err and see all messages for all files (separated by file name of course). Same for eio's with eio diff's. Might be a big file sometimes but it would make review much faster.

rraustad:
I was wondering where these were lurking....

    if ( SELECT_CASE_var == TypeOf_Baseboard_Rad_Conv_Water ) { // 'ZoneHVAC:Baseboard:RadiantConvective:Water'
        ControlCompOutput( HWBaseboard( BaseboardNum ).EquipID, cCMO_BBRadiator_Water, BaseboardNum, FirstHVACIteration, QZnReq, HWBaseboard( BaseboardNum ).WaterInletNode, MaxWaterFlow, MinWaterFlow, HWBaseboard( BaseboardNum ).Offset, HWBaseboard( BaseboardNum ).ControlCompTypeNum, HWBaseboard( BaseboardNum ).CompErrIndex, _, _, _, _, _, HWBaseboard( BaseboardNum ).LoopNum, HWBaseboard( BaseboardNum ).LoopSideNum, HWBaseboard( BaseboardNum ).BranchNum );

HWBaseboardRadiator.cc
OutdoorAirUnit.cc
PoweredInductionUnits.cc
SingleDuct.cc
SteamBaseboardRadiator.cc
UnitHeater.cc
UnitVentilator.cc
VentilatedSlab.cc

rraustad:
I see this with TwoWayCommonPipe_Pri-Sec, on Mac but when run locally I don't see this error. Is Chicago weather used for the CI for this file? I'm using TMY3, is TMY2 used on the CI?

    Site:Location,
      CHICAGO_IL_USA TMY2-94846,  !- Name


    +   ** Warning ** CalcWaterCoilWaterFlowRate: Water coil control failed for Coil:Cooling:Water:DetailedGeometry="VAV SYS 1 COOLING COIL"
    +   **   ~~~   **  During Warmup, Environment=CHICAGO_IL_USA ANNUAL COOLING 1% DESIGN CONDITIONS DB/MCWB, at Simulation time=07/21 23:30 - 23:45
    +   **   ~~~   ** ...Bad water flow rate limits
    +   **   ~~~   ** ...Starting minimum part load ratio =0.438
    +   **   ~~~   ** ...Starting maximum part load ratio =0.876
    +   **   ~~~   ** ...Water water flow rate of [0.49218 (kg/s)] will be used and the simulation continues

rraustad:
Where do I find the 15, 30, and 45zoneVAV performance test files?

Nigusse:
look under \EnergyPlus\performance_tests.

mbadams5:
@rraustad For the past two performance enhancement commits, are you updating functionality/features or fixing bugs in those commits as well? If not, I would strongly suggest profiling to find your hotspots. The changes you've made, on the surface, appear to actually increase runtime.

rraustad:
@mbadams5 yes I see that runtime has doubled. It makes some sense but twice the runtime? I have been trying to look at the change in number of function calls and time. The existing method of water coil control does not account for changes in OA flow rate while the OA controller iterates on a solution. How the water coil controller could ever find a solution given those iterations is beyond me. I just got back from vacation and will look into this further.

mjwitte:
@rraustad Sounds like we need to separate this into two or more parts - one for the OA coil control issue and one for the more general coil control changes. The more general switch from controlcompoutput to regulafalsi (or whatever) should drop in first - then we can isolate the performance and diffs for just that.  Then come back and revisit this OA coil/economizer problem separately.  Referencing your [comment in 5884](https://github.com/NREL/EnergyPlus/issues/5884#issuecomment-311682072) and @RKStrand  earlier comment that it seems like we're distributing logic that was consolidated in ControlCompOutput.  Would it make sense to keep ControlCompOutput as a generic water flow control function and change it to use RegulaFalsi from there?

Referring to some earlier comments:

> Who remembers why this was added, the autosizing of the water controller accuracy. 

That was long long ago, but if I remember correctly, it was done to avoid user angst about what to enter, and it was supposed to be scaling this to line up with the max water flow rate(?).

> Now if RegulaFalsi could somehow switch between secant and bisection, then it could find the answer much more quickly

Isn't that what @lgu1234 is doing in #6104?

rraustad:
Duplicate branch to #6197. Only one of these two branches will merge to develop.

nrel-bot-3:

mjwitte:
@rraustad What's the status on this #6176 and the companion #6197?  Have you settled on which one to use?  If so, please close the other and update the pivotal ticket accordingly.

rraustad:
The PR superseded by #6197

Issue #6173, opened 2017-06-13
VAV airflow problems with other zone HVAC equipment


mjwitte:
Issue overview
--------------
Office building with two VAV systems serving 5 and 6 zones has ZoneHVAC:OutdoorAirUnit (or ideal loads) in several zones to represent a perimeter AHU.  Without the additional zone equipment, the VAV dampers modulate as expected, often at minimum during the heating season.  With the added zone equipment, the VAV dampers run wide open almost all of the time.  The zone equipment adds some heating or cooling to the space, so the VAV damper positions should change, but not go full open.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.7.0

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added EnergyPlusDevSupport\DefectFiles\6173 use Grand Rapids, MI weather
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6177

PR #6177, opened 2017-06-14
Issue warnings for unbalanced outdoor and exhaust airflows in ZoneHVAC:OutdoorAirUnit

lgu1234:
Pull request overview
---------------------
Addresses #6173.  For ZoneHVAC:OutdoorAirUnit, warnings are now throw if there is no exhaust fan, or if the exhaust flow does not balance the supply flow.
 
### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


nrel-bot-3:
@lgu1234 @lgentile it has been 17 days since this pull request was last updated.

nrel-bot-3:

nrel-bot-3:

src/EnergyPlus/SingleDuct.cc
```diff
@@ -1791,8 +1791,11 @@ namespace SingleDuct {
 		//Do a check and make sure that the max and min available(control) flow is
 		//  between the physical max and min while operating.
 		SysInlet( SysNum ).AirMassFlowRateMaxAvail = min( Sys( SysNum ).AirMassFlowRateMax, Node( InletNode ).MassFlowRateMaxAvail );
-		SysInlet( SysNum ).AirMassFlowRateMinAvail = min( max( Node( OutletNode ).MassFlowRateMin, Node( InletNode ).MassFlowRateMinAvail ), SysInlet( SysNum ).AirMassFlowRateMaxAvail );
-
+		if ( Sys( SysNum ).SysType_Num == SingleDuctConstVolReheat ) { // AirTerminal:SingleDuct:ConstantVolume:Reheat
+			SysInlet( SysNum ).AirMassFlowRateMinAvail = min( max( Node( OutletNode ).MassFlowRateMin, Node( InletNode ).MassFlowRateMinAvail ), SysInlet( SysNum ).AirMassFlowRateMaxAvail );
+		} else { // #6173
+			SysInlet( SysNum ).AirMassFlowRateMinAvail = min( Node( OutletNode ).MassFlowRateMin, SysInlet( SysNum ).AirMassFlowRateMaxAvail );
+		}
```
rraustad: @lgu1234 can you explain why inlet node MinAvail isn't used for the VAV terminal unit?

lgu1234: @rraustad The use of inlet node MinAvail will cause damper wide open. The revised code will force damper position at min as desired. Here are plots by running a test file of VAVplusZoneOAUnit.

Damper position from original run

old damper

Damper position from fix run

new damper

Zone air temperature from original run

old zonet

Zone air temperature from fix run
new zonet

mjwitte:
@rraustad I haven't looked closely at this yet (since it's non-IDD-changes).  It needs some scrutiny to be sure it's a general-enough solution to the problem.

rraustad:
Gu, I  believe you that this works I just need to understand why. I ran into the same problem with the SeriesPIU yesterday where MinAvail was equal to MaxAvail. So the question is, why is MinAvail getting set up to the max? In most cases, MinAvail should equal 0.

lgu1234:
@rraustad The MinAvail in my case is based on input:

AirTerminal:SingleDuct:VAV:NoReheat,
    E Lobby Core 6 VAV,      !- Name
    ,                        !- Availability Schedule Name
    E Lobby Core 6 Supply Inlet,  !- Air Outlet Node Name
    E Lobby Core 6 Zone Equip Inlet,  !- Air Inlet Node Name
    autosize,                !- Maximum Air Flow Rate {m3/s}
    Constant,                !- Zone Minimum Air Flow Input Method
    0.2;                     !- Constant Minimum Air Flow Fraction

My guess is that if no input, the default value is zero. I think it is an extreme case to set both MinAvila = MaxAvail.

src/EnergyPlus/SingleDuct.cc
rraustad: I realize that this change works in this case. But I also think the MinAvail is needed to ensure that other components provide a flow rate within constraints imposed elsewhere in the simulation. So, in order for this to move forward, the reason that MinAvail is higher than required here needs to be identified. Otherwise I am not willing to accept this change. Other reviewers might be willing to accept this change.

rraustad:
In SimAirServingZones, the main duct supply air flow rate set point is set.

    ZoneSideNodeNum = AirToZoneNodeInfo( AirLoopNum ).ZoneEquipSupplyNodeNum( OutNum );

    if ( ! FirstHVACIteration ) {
        MassFlowSet = Node( ZoneSideNodeNum ).MassFlowRate;  <-- demand side inlet
    } else { // first time through in each HVAC timestep, use loop design mass flow rates for required mass flows
        MassFlowSet = AirLoopFlow( AirLoopNum ).DesSupply;
    }
    // Need to make sure that flows are greater than zero
    if ( MassFlowSet >= 0.0 ) {
        Node( NodeNumOut ).MassFlowRateSetPoint = MassFlowSet;  <-- supply side outlet

Then in HVACManager the MinAvail or MaxAvail is adjusted to lock in the air flow rate.

    if ( Node( SupplyNode ).MassFlowRate > 0.0 ) {
        if ( ( Node( SupplyNode ).MassFlowRateSetPoint - Node( SupplyNode ).MassFlowRate ) > HVACFlowRateToler * 0.01 ) {
            FlowRatio = Node( SupplyNode ).MassFlowRate / Node( SupplyNode ).MassFlowRateSetPoint;
            for ( ZonesCooledIndex = 1; ZonesCooledIndex <= AirToZoneNodeInfo( AirLoopIndex ).NumZonesCooled; ++ZonesCooledIndex ) {
                TermInletNode = AirToZoneNodeInfo( AirLoopIndex ).TermUnitCoolInletNodes( ZonesCooledIndex );
                Node( TermInletNode ).MassFlowRateMaxAvail = Node( TermInletNode ).MassFlowRate * FlowRatio;
                Node( TermInletNode ).MassFlowRateMinAvail = min( Node( TermInletNode ).MassFlowRateMaxAvail, Node( TermInletNode ).MassFlowRateMinAvail );
            }
        }
        if ( ( Node( SupplyNode ).MassFlowRateSetPoint - Node( SupplyNode ).MassFlowRate ) < -HVACFlowRateToler * 0.01 ) {
            if ( Node( SupplyNode ).MassFlowRateSetPoint == 0.0 ) {
                for ( ZonesCooledIndex = 1; ZonesCooledIndex <= AirToZoneNodeInfo( AirLoopIndex ).NumZonesCooled; ++ZonesCooledIndex ) {
                    TermInletNode = AirToZoneNodeInfo( AirLoopIndex ).TermUnitCoolInletNodes( ZonesCooledIndex );
                    Node( TermInletNode ).MassFlowRateMaxAvail = Node( TermInletNode ).MassFlowRateMax;
                    Node( TermInletNode ).MassFlowRateMinAvail = Node( SupplyNode ).MassFlowRate / double( AirToZoneNodeInfo( AirLoopIndex ).NumZonesCooled );
                }
            } else {
                FlowRatio = Node( SupplyNode ).MassFlowRate / Node( SupplyNode ).MassFlowRateSetPoint;
                for ( ZonesCooledIndex = 1; ZonesCooledIndex <= AirToZoneNodeInfo( AirLoopIndex ).NumZonesCooled; ++ZonesCooledIndex ) {
                    TermInletNode = AirToZoneNodeInfo( AirLoopIndex ).TermUnitCoolInletNodes( ZonesCooledIndex );
                    Node( TermInletNode ).MassFlowRateMinAvail = Node( TermInletNode ).MassFlowRate * FlowRatio;
                    Node( TermInletNode ).MassFlowRateMaxAvail = max(Node(TermInletNode).MassFlowRateMaxAvail, Node(TermInletNode).MassFlowRateMinAvail);
                }
            }
        }
    }
So the question is, why is the supply air flow rate higher than what we believe the TU's should operate at?

rraustad:
This air handler has an OA system, a water coil, and a fan. So either the OA system is setting a high air flow rate or this mess with MinAvail is doing it.

    Branch,
    West VAV AHU Main Branch,!- Name
    ,                        !- Pressure Drop Curve Name
    AirLoopHVAC:OutdoorAirSystem,  !- Component 1 Object Type
    West VAV AHU OA System,  !- Component 1 Name
    West VAV AHU Air Loop Inlet,  !- Component 1 Inlet Node Name
    West VAV AHU Mixed Air Outlet,  !- Component 1 Outlet Node Name
    Coil:Cooling:Water,      !- Component 2 Object Type
    West VAV AHU Cooling Coil,  !- Component 2 Name
    West VAV AHU Mixed Air Outlet,  !- Component 2 Inlet Node Name
    West VAV AHU Cooling Coil Outlet,  !- Component 2 Outlet Node Name
    Fan:VariableVolume,      !- Component 3 Object Type
    West VAV AHU Supply Fan, !- Component 3 Name
    West VAV AHU Cooling Coil Outlet,  !- Component 3 Inlet Node Name
    West VAV AHU Supply Fan Outlet;  !- Component 3 Outlet Node Name

rraustad:
The number of iterations is much higher since there is no bounding of air flow through the TU's.

rraustad:
The fan does use MinAvail in the choice of mass flow rate. The fan inlet and outlet node flows do match, however the airloop supply side outlet and demand side inlet do not match.

    Fan:VariableVolume,
    West VAV AHU Supply Fan, !- Name
    West VAV AHU Cooling Coil Outlet,  !- Air Inlet Node Name
    West VAV AHU Supply Fan Outlet;  !- Air Outlet Node Name

    AirLoopHVAC,
    West VAV AHU,            !- Name
    West VAV AHU Controllers,!- Controller List Name
    West VAV AHU Availability Managers,  !- Availability Manager List Name
    autosize,                !- Design Supply Air Flow Rate {m3/s}
    West VAV AHU Branches,   !- Branch List Name
    West VAV AHU Air Loop Inlet,  !- Supply Side Inlet Node Name
    West VAV AHU Return Air Outlet,  !- Demand Side Outlet Node Name
    West VAV AHU Supply Path Inlet,  !- Demand Side Inlet Node Names
    West VAV AHU Supply Fan Outlet;  !- Supply Side Outlet Node Names

image

image

rraustad:
Been staring at this for a while now. The program is looking at the demand side inlet for the mass flow rate set point.

    ZoneSideNodeNum = AirToZoneNodeInfo( AirLoopNum ).ZoneEquipSupplyNodeNum( OutNum );

    if ( ! FirstHVACIteration ) {
        MassFlowSet = Node( ZoneSideNodeNum ).MassFlowRate;

    Node( NodeNumOut ).MassFlowRateSetPoint = MassFlowSet;

Then it looks at the ratio of air loop flow rate to the demand side inlet flow rate to adjust MinAvail and MaxAvail (see above).

    if ( ( Node( SupplyNode ).MassFlowRateSetPoint - Node( SupplyNode ).MassFlowRate ) < -HVACFlowRateToler * 0.01 ) {

        FlowRatio = Node( SupplyNode ).MassFlowRate / Node( SupplyNode ).MassFlowRateSetPoint;

If the demand side flow rate is passed to the inlet of the air loop, then why are these values different after the air loop simulated again? I see the TU's choosing their mass flow rate and this looks correct. I did not see where these sum of TU flow rates are passed to the demand side inlet. I guess I'd have to check that the sum of the TUs equals the demand side inlet when these tests are performed.

rraustad:
In this figure, this is the state of the air loop when these MinAvail MaxAvail adjustments are made.
Nodes:
[28] supply side outlet - WEST VAV AHU SUPPLY FAN OUTLET
[136] demand side inlet - WEST VAV AHU SUPPLY PATH INLET
[27] supply side inlet - WEST VAV AHU AIR LOOP INLET
[138] demand side outlet - WEST VAV AHYU RETURN AIR OUTLET

image

And then after the zone equipment was simulated. I don't know what to make of this, the demand side outlet is higher than the supply side outlet and I assume it's because the MinAvails were adjusted upwards. Its as if these tests are out of sync with the simulation.

image

rraustad:
I reverted Gu's change and moved ManageZoneEquipment to before the check on MassFlowRateSetPoint and changed the SP to the demand side outlet (which no longer exists with Mike's change). I did this to try to align the data before checking if the mass flow rate ratio is different (e.g., sim airloop AND sim zone equipment THEN set MassFlowRateSetPoint). The damper positions now appear correct but the supply side outlet and demand side inlet still don't match. Iterations are also way up again.

				if ( SimZoneEquipment ) {

					ManageZoneEquipment( FirstHVACIteration, SimZoneEquipment, SimAirLoops );
					SimElecCircuits = true; // If this was simulated there are possible electric changes that need to be simulated

					if ( ( IterAir == 1 ) && ( ! FlowMaxAvailAlreadyReset ) ) { // don't do reset if already done in FirstHVACIteration
						// ResetTerminalUnitFlowLimits(); // don't do reset at all - interferes with convergence and terminal unit flow controls
						FlowResolutionNeeded = true;
					} else {
						ResolveAirLoopFlowLimits();
						FlowResolutionNeeded = false;
					}


				ZoneSideNodeNum = AirToZoneNodeInfo( AirLoopNum ).ZoneEquipReturnNodeNum( OutNum );

				if ( ! FirstHVACIteration ) {
					MassFlowSet = Node( ZoneSideNodeNum ).MassFlowRate;
				} else { // first time through in each HVAC timestep, use loop design mass flow rates for required mass flows
					MassFlowSet = AirLoopFlow( AirLoopNum ).DesSupply;
				}

image

image

image

rraustad:
Here I changed the MassFlowRateSetPoint node to be the demand side outlet (instead of inlet) and the simulation keeps iterating without any real change. Min/Max Avail doesn't get updated because the MassFlowRateSetPoint is the same as the air loop mass flow rate. Also, the demand side inlet node keeps wanting to be 2.22 while the air loop wants 3.06. So something is causing the return node flow rate to be higher than what the TUs want. So this is the reason that MinAvail keeps increasing. The air loop is at a higher flow rate than the zone equipment and MinAvail must increase to converge.

image

rraustad:
I changed back to original code and ran through the same point in time. As the defect file shows, the air flow rates continue to increase throughout the iterations. Remember that the demand side outlet was 0.84 higher than the demand side inlet. Here the same difference is present. And as the simulation increases MinAvail to compensate, the demand side outlet also increased by the same 0.84, and the air loop operates 0.84 higher than the previous iteration and the cycle continues. Now the effort should be to determine why the demand side outlet is higher than the demand side inlet. The only way I can see that happening is if ZoneHVAC equipment have a higher supply air flow rate than inlet. In this case the zone return node would be higher than the sum of the TUs. If there are Zone Exhaust fans, the zone return would be lower than the sum of the TUs.

image

rraustad:
I continued with the iterations to see where it would stop/converge/hit max iterations, etc. The demand side air flow continued to increase with outlet 0.84 greater than inlet. Up to a point where the air loop limited air flow and the demand side inlet converged with the supply side outlet. However, notice that the demand side outlet is still 0.84 higher than the demand side inlet.

image

rraustad:
These last 3 snap shots of the debugger give us talking points of where and how to start fixing this. I think the first step is to figure out why the demand side outlet is higher than the inlet.

rraustad:
I didn't have to look very far. It's the zone OAUnits that are pumping air into the zones and that air is forced back to the air loop inlet. There is no exhaust fan in this object so it was intended to pressurize the zone.

  ZoneHVAC:OutdoorAirUnit,
    Middle AHU W-SW Perimeter 6,  !- Name
    Middle AHU Schedule,             !- Availability Schedule Name
    W-SW Perimeter 6,                !- Zone Name
    0.17934002841,                    !- Outdoor Air Flow Rate {m3/s}
    Middle AHU Schedule,           !- Outdoor Air Schedule Name
    Middle AHU W-SW Perimeter 6 Supply Fan,  !- Supply Fan Name
    BlowThrough,             !- Supply Fan Placement
    ,                        !- Exhaust Fan Name
    ,                        !- Exhaust Air Flow Rate {m3/s}
    ,                        !- Exhaust Air Schedule Name
    TemperatureControl,      !- Unit Control Type
    Middle AHU S Setpoint,              !- High Air Control Temperature Schedule Name
    Middle AHU S Setpoint,              !- Low Air Control Temperature Schedule Name
    Middle AHU W-SW Perimeter 6 OA Inlet Node,          !- Outdoor Air Node Name
    Middle AHU W-SW Perimeter 6 Supply Outlet Node,  !- Air Outlet Node Name
    Middle AHU W-SW Perimeter 6 Exhaust Node,  !- Air Inlet Node Name
    Middle AHU W-SW Perimeter 6 Supply Fan Outlet Node,   !- Supply FanOutlet Node Name
    Middle AHU W-SW Perimeter 6 Equipment;          !- Outdoor Air Unit List Name

image

rraustad:
From the IO ref for ZoneHVAC:OutdoorAirUnit:

    Zones may contain one or more zone outdoor air units. The user can choose to have each unit
    have its own exhaust fan or may choose to have only one unit have a exhaust fan that will handle
    the flow of all of the zone outdoor air units. Note that the user will need to account for any flow
    balance among the zone outdoor air units on the supply and exhaust sides.

Would it be easier to just have each unit account for their own exhaust air (i.e., exhaust air required)? Surely there needs to be a test for this as it impacts air loop performance.

lgu1234:
I modified the input file (VAVplusZoneOAUnit.idf) by adding exhaust fan in the ZoneHVAC:OutdoorAirUnit object:

  ZoneHVAC:OutdoorAirUnit,
    Middle AHU S-SW Perimeter 6,  !- Name
    Middle AHU Schedule,             !- Availability Schedule Name
    S-SW Perimeter 6,                !- Zone Name
    0.17934002841,                    !- Outdoor Air Flow Rate {m3/s}
    Middle AHU Schedule,           !- Outdoor Air Schedule Name
    Middle AHU S-SW Perimeter 6 Supply Fan,  !- Supply Fan Name
    BlowThrough,             !- Supply Fan Placement
    Middle AHU S-SW Perimeter 6 Exhaust Fan,  !- Exhaust Fan Name
    0.17934002841,                        !- Exhaust Air Flow Rate {m3/s}
    Middle AHU Schedule,                        !- Exhaust Air Schedule Name
    TemperatureControl,      !- Unit Control Type
    Middle AHU S Setpoint,              !- High Air Control Temperature Schedule Name
    Middle AHU S Setpoint,              !- Low Air Control Temperature Schedule Name
    Middle AHU S-SW Perimeter 6 OA Inlet Node,          !- Outdoor Air Node Name
    Middle AHU S-SW Perimeter 6 Supply Outlet Node,  !- Air Outlet Node Name
    Middle AHU S-SW Perimeter 6 Exhaust Node,  !- Air Inlet Node Name
    Middle AHU S-SW Perimeter 6 Supply Fan Outlet Node,   !- Supply FanOutlet Node Name
    Middle AHU S-SW Perimeter 6 Equipment;          !- Outdoor Air Unit List Name

Results are very meaningful. Here is a chart to show mass flow rates:

mass flow rate

Here is a chart to show damper position:

damperposition

Therefore, as long as we require exhaust fan inputs in the ZoneHVAC:OutdoorAirUnit objects, we should not have the problems.

lgu1234:
A new test file was uploaded in EnergyPlusDevSupport by adding exhaust fan objects, Node lists, and exhaust fan fields in ZoneHVAC:OutdoorAirUnit objects.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -1791,8 +1791,11 @@ namespace SingleDuct {
 		//Do a check and make sure that the max and min available(control) flow is
 		//  between the physical max and min while operating.
 		SysInlet( SysNum ).AirMassFlowRateMaxAvail = min( Sys( SysNum ).AirMassFlowRateMax, Node( InletNode ).MassFlowRateMaxAvail );
-		SysInlet( SysNum ).AirMassFlowRateMinAvail = min( max( Node( OutletNode ).MassFlowRateMin, Node( InletNode ).MassFlowRateMinAvail ), SysInlet( SysNum ).AirMassFlowRateMaxAvail );
-
+		if ( Sys( SysNum ).SysType_Num == SingleDuctConstVolReheat ) { // AirTerminal:SingleDuct:ConstantVolume:Reheat
```
rraustad: This change now needs to be reverted back to the original and a correction made to OAUnits. Either require exhaust air for each unit (requires doc and idd changes) or test that the sum of SA = EA for all OAUnits in each zone. A fatal error if SA != EA in either case. So if th is is to make V8.9 then the latter fix needs to be made.

mjwitte: @lgu1234 @rraustad Well, I learned something new.  I never realized that adding the OA unit without an exhaust fan (either in the unit or in the zone) could impact the air loop by creating extra return flow.  That shouldn't be able to drive the air loop to higher flow - but that's (kind of) a different issue.  I would suggest for this particular fix, that the OA units could check if there is no exhaust fan in the OA unit, and there is no zone exhaust fan, then issue a warning.  It might be what the user wants.  In this particular case, it is not what the user (moi) wanted.  And said user was very confused and frustrated by the odd result.  But I think I also have a case with ideal loads that shows similar symptoms.  I'll add that defect file shortly.

mjwitte: And why should excess return air from one or more zones be able to impact the system flow in an airloop?  That doesn't seem right.

lgu1234: @rraustad I agree that the change needs to be back to the original. 
@mjwitte I will add a check for exhaust fan and issue a warning if there is no exhaust fan. In addition, shall we also check exhaust fan flow rate to ensure the exhaust fan flow rate is balanced with OA rate?

mjwitte: Sure, I think that's useful.  As long as these are just warnings.

rraustad:
What I learned while investigating this is that ResolveAirLoopFlowLimits does two things. For constant volume systems it allows the air loop to operate at higher flow rate than the uncontrolled TUs are sized for (MaxAvail > Max) and it is also used to clamp MinAvail and MaxAvail (to a point where convergence is achieved) for simulations where system air flow rate continually changes for one reason or another.

src/EnergyPlus/SingleDuct.cc
mjwitte: So, there *is* already a defect file present for this with ideal loads plus VAV.  I re-ran that with all nodes showing flow rates - and sure enough - similar but not exactly the same - the exhaust nodes are connected properly but the exhaust node flow rates are zero.  That's an error in ideal loads.  It's up to you if you want to fix that here or make a second pull request.

mjwitte: Also, I wonder if airloops should be throwing a warning during the simulation if the return air flow is higher than the mixer air flow.  Check that in the OA mixer?  Not sure.

rraustad: For the OA units (or I guess any zone HVAC really but I haven't thought through those yet and whether they should ever be imbalanced), if the imbalance could somehow be communicated back to the OA mixer, then the extra air could be exhausted and problem solved because ResolveAirLoopFlowLimits looks at supply side outlet and demand side inlet. I'm also thinking that function assumes balanced flow, so if we balance one way or another that function won't bother us again.

lgu1234: @mjwitte I would like to make a new issue to deal with ideal load plus VAV case.

mjwitte: That's fine.

doc/input-output-reference/src/overview/group-zone-forced-air-units.tex
```diff
@@ -1616,7 +1616,9 @@ \subsection{ZoneHVAC:OutdoorAirUnit}\label{zonehvacoutdoorairunit}
 \caption{Zone Outdoor Air Unit Schematic \protect \label{fig:zone-outdoor-air-unit-schematic}}
 \end{figure}
 
-Zones may contain one or more zone outdoor air units.~ The user can choose to have each unit have its own exhaust fan or may choose to have only one unit have a exhaust fan that will handle the flow of all of the zone outdoor air units.~ Note that the user will need to account for any flow balance among the zone outdoor air units on the supply and exhaust sides.
+Zones may contain one or more zone outdoor air units.~ The user can choose to have each unit have its own exhaust fan or may choose to have only one unit have an exhaust fan that will handle the flow of all of the zone outdoor air units.~ Note that the user will need to account for any flow balance among the zone outdoor air units on the supply and exhaust sides.
+
+When an AirLoopHVAC also serves the same zone, an exhaust fan is strongly recommended to be input. In addition, the exhaust airflow rate and air schedule should be the same as outdoor air flow rate and outdoor air schedule, respectively. Otherwise, possible unbalanced flows will cause improper return airflow from the AirLoopHVAC.
```
rraustad: "exhaust fan is strongly recommended" ? What if the OAUnit has balanced flow? How about:

When an AirloopHVAC serves the same zone, it is strongly recommended that the outdoor and exhaust air flow rates are balanced.

lgu1234: @rraustad Thanks. Done.

rraustad (approved):
These changes have been discussed and implementation is consistent with design intent.

rraustad:
All diffs reviewed and nothing unexpected given these changes.

rraustad:
@mjwitte as a second reviewer I will wait for comments before merging.

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -553,6 +555,12 @@ namespace OutdoorAirUnit {
 
 			//N2
 			OutAirUnit( OAUnitNum ).ExtAirVolFlow = NumArray( 2 );
+			if ( OutAirUnit( OAUnitNum ).ExtFan ) {
+				if ( NumArray( 2 ) != NumArray( 1 ) ) {
+					ShowWarningError( CurrentModuleObject + "=\"" + cAlphaArgs( 1 ) + "\" unbalanbce inputs between " + cNumericFields( 1 ) + " and " + cNumericFields( 2 ) + " may cause unbalance mass flow rate between supply and exhaust." );
```
mjwitte: typo "unbalanbce"

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -1508,24 +1521,18 @@ namespace OutdoorAirUnit {
...
-			//! CALL ShowRecurringWarningErrorAtEnd(TRIM(OutAirUnit(OAUnitNum)%Name)//&
-			//! ':  Air mass balance is required by other outdoor air units, ZoneMixing, ZoneCrossMixing, or other air flow control inputs.'&
-			//!   , OutAirUnit(OAUnitNum)%UnBalancedErrIndex)
-			//      END IF
-			//    END IF
+			// Air mass balance check
+			if ( std::abs( OutAirUnit( OAUnitNum ).ExtAirMassFlow - OutAirUnit( OAUnitNum ).OutAirMassFlow ) > 0.001 ) {
+				OutAirUnit( OAUnitNum ).UnBalancedErrCount = OutAirUnit( OAUnitNum ).UnBalancedErrCount + 1;
+				if ( OutAirUnit( OAUnitNum ).UnBalancedErrCount == 1 ) {
+					ShowWarningError( "Air mass flow between zone supply and exhaust is not balanced" );
```
mjwitte: @lgu1234 Does this warning now happen for any unbalanced on a single OA unit?  Even if another unit balances everything?  Even if there is no airloop serving the same zone?  If so, then the docs should be changed to say this unit must be balanced.  Or perhaps better yet, the OA unit should be placing the correct flow on an exhaust node whether there is an exhaust fan or not.  But the exhaust node "Air Inlet Node Name" which is munged to "AirInlet Node Name" in the IDD is not a required field.  And this warning should probably not be triggered if zone mass conservation is active?  If we go ahead and add these changes, we still need  followup with a better fix after v8.8 release.

doc/input-output-reference/src/overview/group-zone-forced-air-units.tex
```diff
@@ -1616,7 +1616,9 @@ \subsection{ZoneHVAC:OutdoorAirUnit}\label{zonehvacoutdoorairunit}
...
 \end{figure}
 
-Zones may contain one or more zone outdoor air units.~ The user can choose to have each unit have its own exhaust fan or may choose to have only one unit have a exhaust fan that will handle the flow of all of the zone outdoor air units.~ Note that the user will need to account for any flow balance among the zone outdoor air units on the supply and exhaust sides.
+Zones may contain one or more zone outdoor air units.~ The user can choose to have each unit have its own exhaust fan or may choose to have only one unit have an exhaust fan that will handle the flow of all of the zone outdoor air units.~ Note that the user will need to account for any flow balance among the zone outdoor air units on the supply and exhaust sides.
+
+When an AirloopHVAC serves the same zone, it is strongly recommended that the outdoor and exhaust air flow rates are balanced. In addition, the exhaust airflow rate and air schedule should be the same as outdoor air flow rate and outdoor air schedule, respectively. Otherwise, possible unbalanced flows will cause improper return airflow from the AirLoopHVAC.
```
mjwitte: This explains things, which is better than nothing, but this really shouldn't be happening in the first place.  It's possible once the multiple airloop return flow calculations are finalized that the return flow will no longer be impacted by any imbalance in zone equipment flows, but I'm not 100% sure of that right now.

mjwitte:
@lgu1234 I've made some minor revisions to the error messages.

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -1508,24 +1523,18 @@ namespace OutdoorAirUnit {
...
+					ShowContinueError( "Occurs in ZoneHVAC:OutdoorAirUnit Object= " + OutAirUnit( OAUnitNum ).Name );
+					ShowContinueError( "Air mass balance is required by other outdoor air units: Fan:ZoneExhaust, ZoneMixing, ZoneCrossMixing, or other air flow control inputs." );
+					ShowContinueErrorTimeStamp( "The outdoor mass flow rate = " + General::RoundSigDigits( OutAirUnit( OAUnitNum ).OutAirMassFlow, 3 ) + " and the exhaust mass flow rate = " + General::RoundSigDigits( OutAirUnit( OAUnitNum ).ExtAirMassFlow, 3 ) + ".");
+				} else {
+					ShowRecurringWarningErrorAtEnd( OutAirUnit( OAUnitNum ).Name + ":  Air mass balance is required by other outdoor air units, Fan:ZoneExhaust, ZoneMixing, ZoneCrossMixing, or other air flow control inputs.", OutAirUnit( OAUnitNum ).UnBalancedErrIndex );
```
mjwitte: @lgu1234 @rraustad Since we don't know for sure here if other components are providing the flows to balance the zone, I think we should drop the recurring warning and state in the initial warning that this will only be reported once.  That's what the unbalanced airloop and zone warning do as well in CalcZoneMassBalance.

lgu1234: @mjwitte I am going to drop the recurring warnings.

tst/EnergyPlus/unit/OutdoorAirUnit.unit.cc
```diff
@@ -338,5 +338,19 @@ namespace EnergyPlus {
 		EXPECT_DOUBLE_EQ( EAFanPower, 75.0 );
 		EXPECT_DOUBLE_EQ( SAFanPower + EAFanPower, OutAirUnit( OAUnitNum ).ElecFanRate );
 
+		// #6173
```
Myoldmopar, line 341: We can always pass the blame back for what commit/PR added a block of code, so no need for comments like this.  The code looks good otherwise though, so unless there's a follow-up commit, this is OK.

Myoldmopar (approved):

Myoldmopar:
> The calculation of AirMassFlowRateMinAvail was changed to allow the minimum mass flow rate value based on user inputs for VAV single duct terminals.

It says this in the Pull Request description, so I was looking for the code changes to accompany this, and I don't see any.  Is the description out of date?

mjwitte:
Yes, the description is old - from the original attempt to fix this.

mjwitte:
@lgu1234 Our changes overlapped and the warning about blank exhaust fan name got in twice. I've removed the duplicated warning, and added a message to the unbalanced flow warning stating it is only reported once.  Also merged up to current develop.  If CI comes back clean, I will merge this.

mjwitte:
And I updated the pull request overview comment.

Myoldmopar:
Looks like a unit test failure still unless I'm missing something.  Once that is remedied then yeah this does look ready to go in.

mjwitte:
@Myoldmopar @lgu1234 Fixed the unit test that I broke yesterday, and fixed spelling of occurrence.

lgu1234:
@mjwitte Thanks a lot.

Myoldmopar:
Looks like this one will be clean now :crossed_fingers:

Myoldmopar:
CI is all done @mjwitte.  Merge this at will if you are happy with it.

Issue #6098, opened 2017-04-17
User file with large ZoneHVAC:CoolingPanel:RadiantConvective:Water crashes in UA calculation


mjwitte:
Issue overview
--------------
User file crashes with log of a negative number [here]() when calculating the UA for the radiant cooling panel.
```
			CoolingPanel( CoolingPanelNum ).UA = -MDotXCp * log( 1.0 - ( std::abs(Qrated) / ( MDotXCp * std::abs( Tinletr - Tzoner ) ) ) );
```
Qrated is 6498W and the denominator is only 3115.  This results in log(1.0 - 2.2) which fails.

The calculation needs to be protected from a bad log argument in general.
But this raises questions about the usefulness of the three Rated inputs in this object:
```
    16,                      !- Rated Inlet Water Temperature {C}
    28,                      !- Rated Inlet Space Temperature {C}
    0.063,                   !- Rated Water Mass Flow Rate {kg/s}
```
There is no corresponding input for a Rated Capacity, so the large actual capacity is compared to these Rated inputs to calculate the UA value.  Seems that the mass flow rate corresponding to the "Maximum Chilled Water Flow Rate" should be used here instead (which is more like 0.5).  

And the Rated inputs could possibly be unnecessary?

### Workaround
Enter a larger value for Rated Water Mass Flow Rate, check the eio output for the autosized Max Chilled Water Flow Rate and then use the mass flow equivalent as the input for Rate Water Mass Flow Rate.

### Details
Some additional details for this issue (if relevant):
 - Win10, 64-bit
 - Version of EnergyPlus 8.7.0
 - Helpdesk ticket number 11974

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added to EnergyPlusDevSupport\DefectFiles
 - [x] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte I'll start looking at this, hope to get to this over the next several days.  I have added a ticket to Pivotal (ID #147520311).  From an initial look at this, it seems that the equation used for the UA calculation assumes that the scaled capacity is the rated capacity which is not correct. I suggest that a check be put into the UA calculation to avoid an improper log calculation but will think more about the inputs and how to resolve the scaled vs. rated issue.

RKStrand:
[...]

Issue #5823, opened 2016-08-19
Followup for new cooling panel ZoneHVAC:CoolingPanel:RadiantConvective:Water


mjwitte:
## Issue overview

ZoneHVAC:CoolingPanel:RadiantConvective:Water was added in #5027.  Some outstanding issues remain:
- [x] Fix sizing - scalable sizing input fields were added to IDD and input processing in https://github.com/NREL/EnergyPlus/pull/5027/commits/7eb88d9c799aa3c640f81b1d006e781bf62e92ab but it doesn't appear that they are implemented in the sizing function for this component.  *The sizing function was added later in Sep 2016 with https://github.com/NREL/EnergyPlus/commit/3944eb16f10b7aa98ea06e52c6be3ae9eceeac11 @RKStrand tested the various options in Nov 2017 and this is working.*
- [x] Some minor OO restructuring - move functions specific to this object to be member functions of the CoolingPanelSimple struct
- [x] Review new doc sections one more time - some of the engineering ref equations need some fixing (sub/superscripting out of place), add the actual object name at the beginning of the engineering reference section so it can be found
- [x] Change the SizeCoolingPanelUA so rather than returning the error status via the reference it returns true on success and false on failure? It's a more "C++" way of doing this.
- ~Annual simulations of the new tests files have many "ControlCompOutput: Maximum iterations exceeded"errors~ (will be addressed in #5884)
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
- Unmethours link or helpdesk ticket number
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte Could you elaborate on the "fix sizing" and "minor OO restructuring" comments above?  Maybe this was already discussed and I just need to refresh my memory?

RKStrand:
@mjwitte Handling baseboard issues under GitHub Issue #5884 as this really has nothing to do with the new cooling panel model.  It affects all input files with baseboards.  Other test suite input files show the same behavior when an annual run is launched.

mjwitte:
[...]

PR #6181, opened 2017-06-21
Cooling Panel UA Calculation Bug for Large Capacity/Low Water Flow Rate

RKStrand:
Pull request overview
---------------------
This pull request includes changes to fix the defect User file with large ZoneHVAC:CoolingPanel:RadiantConvective:Water crashes in UA calculation (Issue #6098).  In the defect input file, the user had entered a large unit capacity but used a fairly small flow rate.  The result was that the panel would have a better than 100% efficiency for heat transfer and led to a hard crash of EnergyPlus.  The fix traps this and lets the user know how to correct the situation.  In addition to a new unit test, the I/O Reference was modified to clarify the relationship between different input parameters and also their intent.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #6098 
 
### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


doc/input-output-reference/src/overview/group-radiative-convective-units.tex
```diff
@@ -571,15 +571,15 @@ \subsubsection{Inputs}\label{inputs-3-028-1}
...
+This field is the rated standard water flow rate in kg/s that was used to obtain the design cooling capacity under rated temperature conditions.  This value along with the rated space and water inlet temperatures as well as the cooling capacity (whether autosized or entered by the user) to calculate the overall heat transfer coefficient (U-Value) that will be used to characterize the heat transfer between the water being circulated through the panel and the air in the space.  Users should note that if this value is too low in comparison to the cooling design capacity that the program will produce an error message and the user must adjust one of these two parameters or the rated temperature values.  The default value is 0.063kg/s (1 gpm).  If it is blank or zero, the default value is assumed.
```
jasondegraw: I think that perhaps the second sentence here needs a "is used" someplace.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,29 +1015,62 @@ namespace CoolingPanelSimple {
...
+		bool SizeUAErrorFlag = false;
+		SizeCoolingPanelUA( CoolingPanelNum, SizeUAErrorFlag );
+		if ( SizeUAErrorFlag ) ShowFatalError( "SizeCoolingPanelUA: Program terminated for previous conditions." );
+
+	}
+
+	void
+	SizeCoolingPanelUA(
+					 int const CoolingPanelNum,
+					 bool & SizeUAError
```
jasondegraw: Could we change this function so rather than returning the error status via the reference it returns true on success and false on failure? It's a more "C++" way of doing this.

jasondegraw (changes requested):
There are a couple small things that I picked on, but other than that it looks good.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: @jasondegraw Any objections to lumping this comment together with some other C++ modifications that have been requested?  That is already the subject for Followup for new cooling panel ZoneHVAC:CoolingPanel:RadiantConvective:Water (GitHub Issue #5823).

doc/input-output-reference/src/overview/group-radiative-convective-units.tex
RKStrand: @jasondegraw I agree...I thought I had commented on this earlier and it is not showing up here now.  Anyway, yes, I will correct this as soon as possible.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
jasondegraw: @RKStrand That sounds good to me.

RKStrand:
@jasondegraw @mjwitte Ok, I decided (after putting additional language in the other defect description) to just go ahead and fix this now.  So, the documentation was altered to fix the missing words and the code was updated as Jason requested.  So, hopefully this is ready to go.  Let me know if there is anything else that needs to be fixed/adjusted.  Thanks!

RKStrand:
@jasondegraw @mjwitte Ok, all of the review comments have been included and this is ready for final review.  I did end up correcting the documentation and making the OO change in the flag that Jason requested.  Thanks for your help with this!

src/EnergyPlus/ChilledCeilingPanelSimple.cc
jasondegraw: @RKStrand I think maybe I was unclear on what I was suggesting here, sorry if that was the case. What I was suggesting here was to change the function signature to `bool SizeCoolingPanelUA(int const CoolingPanelNum)` rather than use a reference.

RKStrand:
Jason:

Sorry for the confusion.  I want to make sure I understand what you are asking before I do anything else because I need to finish this and move on to other things.

If I am understanding you correctly, you want SizeCoolingPanelUA subroutine into a bool function that returns true or false based on the error flag status.  Is that correct?  If not, please clarify further.  If this is correct, then I wonder if that is really the right approach.  This subroutine does calculate a UA value which is stored in the data structure.  I always thought a function is a single purpose entity that cannot really do anything else.  But perhaps that is just my F90 background.  Can you comment on this?

Thanks,

Rick

jasondegraw:
@RKStrand Yes, that is what I am suggesting. It is typical in C++ code to see functions that can "fail" return a boolean status. That doesn't mean the function is doing more than one thing, though. It's just reporting the result of the function call in a better way.

RKStrand:
@jasondegraw OK, so even though the "function" is doing other calculations that will be stored in the data structure for the cooling panel, I can (should) define this as a bool function.  I will go ahead and do that and let you know when that is finished.  Hopefully it won't be too long.

RKStrand:
@jasondegraw I have committed some additional changes that should work and address the issue (I hope).  It seems to be working so if all the checks pass and you are happy with this, please complete the review and merge.  Thanks!

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,34 +1015,72 @@ namespace CoolingPanelSimple {
...
+		SizeCoolingPanelUASuccess = SizeCoolingPanelUA( CoolingPanelNum );
+		if ( ! SizeCoolingPanelUASuccess ) ShowFatalError( "SizeCoolingPanelUA: Program terminated for previous conditions." );
+
+	}
+
+	bool
+	SizeCoolingPanelUA(
+					 int const CoolingPanelNum
+					 )
+	{
```
rraustad: spacing severely messed up here

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,34 +1015,72 @@ namespace CoolingPanelSimple {
...
-				ShowFatalError( "SizeCoolingPanel: Program terminated for previous conditions." );
+				SizeCoolingPanelUA = false;
 			}
 		}
+	
+		return SizeCoolingPanelUA;
 		
 	}
-	
+
```
rraustad: Added a tab here?

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,34 +1015,72 @@ namespace CoolingPanelSimple {
...
+		Real64 Cp;
 		Real64 MDot;
 		Real64 MDotXCp;
 		Real64 Qrated;
 		Real64 Tinletr;
 		Real64 Tzoner;
+		Real64 RatCapToTheoMax; // Ratio of unit capacity to theoretical maximum output based on rated parameters
+
+		SizeCoolingPanelUA = true;
 		Cp = 4120.0; // Just an approximation, don't need to get an exact number
```
rraustad, line 1050: Why don't we need an exact number if a few lines below we test for RatCapToTheoMax < 1.0 ?? What if RatCapToTheoMax is 1.0001 because we weren't accurate on Cp?

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,34 +1015,72 @@ namespace CoolingPanelSimple {
...
 		MDot = CoolingPanel( CoolingPanelNum ).RatedWaterFlowRate;
 		MDotXCp = Cp * MDot;
 		Qrated = CoolingPanel( CoolingPanelNum ).ScaledCoolingCapacity;
 		Tinletr = CoolingPanel( CoolingPanelNum ).RatedWaterTemp;
 		Tzoner = CoolingPanel( CoolingPanelNum ).RatedZoneAirTemp;
+		RatCapToTheoMax = std::abs(Qrated) / ( MDotXCp * std::abs( Tinletr - Tzoner ) );
+		if (RatCapToTheoMax >= 1.0 ) {
```
rraustad: Why can't RatCapToTheoMax = 1?

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1015,34 +1015,72 @@ namespace CoolingPanelSimple {
...
+			ShowSevereError( "SizeCoolingPanelUA: Unit=[" + cCMO_CoolingPanel_Simple + ',' + CoolingPanel( CoolingPanelNum ).EquipID + "] has a cooling capacity that is greater than the maximum possible value." );
+			ShowContinueError( "The result of this is that a UA value is impossible to calculate." );
+			ShowContinueError( "Check the rated input for temperatures, flow, and capacity for this unit." );
+			ShowContinueError( "The ratio of the capacity to the rated theoretical maximum must be less than unity." );
+			ShowContinueError( "The most likely cause for this is probably either the capacity (whether autosized or hardwired) being too high or the rated flow being too low." );
+			SizeCoolingPanelUA = false;
```
rraustad: What if the water temp < zone temp? Just saying there could be more information here.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: @rraustad Well, I see your point that perhaps this needs a bit of tolerance in here.  This is mostly to trap the clearly illogical.  I will outline something in response to your next comment.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: Yes, this was messed up.  Fixed now.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: Also deleted the extra tab here and a couple other places in this new routine.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: @rraustad Two reasons. First because physically you can't get more heat transfer out of the fluid than physically possible.  Having a ratio of higher than 1 means that the exit temperature of the water leaving the panel would be warmer than the room temperature, violating the 2nd Law.  Second because in the code following this it would take the log of a zero or negative number which would cause a crash. Basically the user is setting up a cooling panel that is better than a perfect heat transfer device using rated input that does not make sense.  Admittedly, with a fixed Cp number this could result if the user is "flying to close to the sun".  I will adjust the code to be a little more gracious but also to limit the RatCaptoTheoMax number to something that will still allow a calculation.  So, it will only trap things when there is really a problem with the input.

doc/input-output-reference/src/overview/group-radiative-convective-units.tex
```diff
@@ -571,15 +571,15 @@ \subsubsection{Inputs}\label{inputs-3-028-1}
...
 
 \paragraph{Field: Cooling Design Capacity}\label{field-cooling-design-capacity-2}
 
-This field is for the cooling panel unit nominal cooling capacity in watts. This field can be autosized by EnergyPlus. When the Cooling Design Capacity Method is \emph{CoolingDesignCapacity}~ and this input is is blank, autosizing is assumed. Design day sizing run must be specified.
+This field is for the cooling panel unit nominal cooling capacity in watts at the rated flow rate of water through the unit (see Rated Water Mass Flow Rate field above). This field can be autosized by EnergyPlus. When the Cooling Design Capacity Method is \emph{CoolingDesignCapacity}~ and this input is is blank, autosizing is assumed. Design day sizing run must be specified.
```
rraustad: if another commit is required, this text could be corrected. "is is"

... and this input is is blank, autosizing is assumed.

RKStrand: @rraustad I see it and will correct it.  Need to make another commit to add things to the code per your other comments so this will get in.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: @rraustad I'll add some more text to the error message to help the user more.  By the way, right after this, there is an error check regarding the temperatures but I will add something to the error message you highlighted to discuss temperatures as well.

RKStrand:
@rraustad Ok, I have another commit that improves based on what you suggested and also fixes the "is is" doc problem (which actually showed up in several places--I did a Find and fixed them all in this file).  I also merged with develop and the test suite is running.  Hopefully this takes care of things and this is now ready to officially merge.  Thanks for reviewing this!

jasondegraw (approved):

jasondegraw:
@Myoldmopar @mjwitte I believe that @RKStrand has addressed all of the comments and this is ready to merge. I'm going to assign you instead of me because this PR has no input changes and can drop in whenever you think is best.

rraustad:
Diff's are either DELight, get_bucket - head_bucket, or invalid string position. This looks OK.

rraustad:
Reviewed, merged, and closed.

Issue #6182, opened 2017-06-21
Earth Tube One Per Zone Limit Check has Errors


RKStrand:
Issue overview
--------------
There is a limit of only one earth tube per zone in EnergyPlus.  (While this could be changed to allow multiple earth tubes per zone, that is not the subject of this issue and should be considered as a future minor enhancement.)  The code to check that there is only one earth tube per zone has several errors in it:

Code:
		// Check to make sure there is only on ventilation statement per zone
		for ( Loop = 1; Loop <= TotEarthTube; ++Loop ) {
			for ( Loop1 = Loop + 1; Loop1 <= TotEarthTube - 1; ++Loop1 ) {
				if ( EarthTubeSys( Loop ).ZonePtr == EarthTubeSys( Loop1 ).ZonePtr ) {
					ShowSevereError( cAlphaArgs( 1 ) + " is assigned to more than one " + cCurrentModuleObject );
					ShowContinueError( "Only one such assignment is allowed." );
					ErrorsFound = true;
				}
			}
		}

Errors:
- The outer loop should be to TotEarthTube -1 and the inner loop to TotEarthTube (the code as it currently stands could actually allow two earth tubes in a single zone but the second would overwrite the results of the first.)
- The comment at the beginning of the code snippet needs to be edited.
- The error message is very unclear and needs to be modified.

The condition that could trigger this problem is having the last earth tube be assigned to a zone that already has another earth tube.

The workaround for the user is to simply not have more than one earth tube per zone.

Validation should be straightforward--the code should not allow the last earth tube to be in a zone that has another earth tube already.

### Details
Some additional details for this issue (if relevant):
 - Platform (any)
 - Version of EnergyPlus (Version 8.7)
 - Unmethours link or helpdesk ticket number (none)

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
[...]

PR #6185, opened 2017-06-24
Earth Tube One Per Zone Limit Error

RKStrand:
Pull request overview
---------------------
This work fixes a problem in the earth tube summarized by GitHub Issue #6182.  In this defect, the code that checks to see that there is only one earth tube per zone is not working properly.  It is actually possible to put two earth tubes in a zone provided that the last earth tube in the input file gets assigned to a zone that already has an earth tube.  The problem is this shouldn't be allowed and then this basically overwrites the results of the first earth tube on a zone.  The problem only occurs when the second earth tube in a zone is the last earth tube in the input file, but still should not be allowed.  This work fixes this problem.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #6182.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/EarthTube.cc
+++ b/src/EnergyPlus/EarthTube.cc
@@ -237,7 +237,6 @@ namespace EarthTube {
 		int NumNumber;
 		int IOStat;
 		int Loop;
-		int Loop1;
 		Array1D_bool RepVarSet;
 
 		RepVarSet.dimension( NumOfZones, true );
@@ -416,10 +415,11 @@ namespace EarthTube {
 	
 	void
 	CheckEarthTubesInZones
-		( std::string const ZoneName, // name of zone for error reporting
-		  std::string const FieldName, // name of earth tube in input
-		  bool & ErrorsFound // Found a problem
-		)
+	(
+		std::string const ZoneName, // name of zone for error reporting
+		std::string const FieldName, // name of earth tube in input
+		bool & ErrorsFound // Found a problem
+	)
 	{
 
 		int Loop;
--- a/src/EnergyPlus/EarthTube.cc
+++ b/src/EnergyPlus/EarthTube.cc
@@ -405,16 +405,7 @@ namespace EarthTube {
 			}
 		}
 
-		// Check to make sure there is only on ventilation statement per zone
-		for ( Loop = 1; Loop <= TotEarthTube; ++Loop ) {
-			for ( Loop1 = Loop + 1; Loop1 <= TotEarthTube - 1; ++Loop1 ) {
-				if ( EarthTubeSys( Loop ).ZonePtr == EarthTubeSys( Loop1 ).ZonePtr ) {
-					ShowSevereError( cAlphaArgs( 1 ) + " is assigned to more than one " + cCurrentModuleObject );
-					ShowContinueError( "Only one such assignment is allowed." );
-					ErrorsFound = true;
-				}
-			}
-		}
+		CheckEarthTubesInZones( cAlphaArgs( 1 ), cCurrentModuleObject, ErrorsFound );
 
 		if ( ErrorsFound ) {
 			ShowFatalError( cCurrentModuleObject + ": Errors getting input.  Program terminates." );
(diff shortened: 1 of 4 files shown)


src/EnergyPlus/EarthTube.cc
```diff
@@ -405,23 +405,41 @@ namespace EarthTube {
...
+	
+	void
+	CheckEarthTubesInZones
+		( std::string const ZoneName, // name of zone for error reporting
+		  std::string const FieldName, // name of earth tube in input
+		  bool & ErrorsFound // Found a problem
+		)
+	{
+
+		int Loop;
```
Myoldmopar: Slowly coerce toward defining variables as they are used instead of at the top of the function.  I know it goes against your years of training, so no worries, just a note.

src/EnergyPlus/EarthTube.cc
```diff
@@ -405,23 +405,41 @@ namespace EarthTube {
...
+		int Loop1;
+		
+		// Check to make sure there is only one earth tube statement per zone
+		for ( Loop = 1; Loop <= TotEarthTube - 1; ++Loop ) {
+			for ( Loop1 = Loop + 1; Loop1 <= TotEarthTube; ++Loop1 ) {
```
Myoldmopar, line 430: Ahh I see the fix.  Weird.

src/EnergyPlus/EarthTube.cc
```diff
@@ -405,23 +405,41 @@ namespace EarthTube {
...
 
 		if ( ErrorsFound ) {
 			ShowFatalError( cCurrentModuleObject + ": Errors getting input.  Program terminates." );
 		}
 
 	}
 
+	
+	void
+	CheckEarthTubesInZones
```
Myoldmopar, line 417: Whitespace here looks close, but something a bit funny.  I'll fix it as I think there is one other thing I need to fix in this before merging.  Please hold...

src/EnergyPlus/EarthTube.cc
```diff
@@ -405,23 +405,41 @@ namespace EarthTube {
...
+		CheckEarthTubesInZones( cAlphaArgs( 1 ), cCurrentModuleObject, ErrorsFound );
 
 		if ( ErrorsFound ) {
 			ShowFatalError( cCurrentModuleObject + ": Errors getting input.  Program terminates." );
 		}
 
 	}
 
+	
+	void
```
Myoldmopar, line 416: :+1: for breaking up this functional section of code into a standalone function to allow for meaningful unit testing!

tst/EnergyPlus/unit/EarthTube.unit.cc
```diff
@@ -0,0 +1,103 @@
...
+	EarthTubeSys( 3 ).ZonePtr = 3;
+	
+	// First case--no conflicts, only one earth tube per zone (ErrorsFound = false)
+	CheckEarthTubesInZones( ZoneName, InputName, ErrorsFound );
+	EXPECT_EQ( ErrorsFound, false );
+
+	// Second case--conflict with the last earth tube and first (ErrorsFound = true)
+	EarthTubeSys( 3 ).ZonePtr = 1;
+	CheckEarthTubesInZones( ZoneName, InputName, ErrorsFound );
+	EXPECT_EQ( ErrorsFound, true );
```
Myoldmopar, line 96: Excellent.  OK, the only thing now is to add this unit test file to the CMakeLists.txt file so that the test actually gets built and run :upside_down_face:

Myoldmopar (changes requested):
A couple changes to be made, but I've made them locally and I'm building and testing.

Myoldmopar:
And now with 14a5623 in place, the unit test you made actually runs, and passes!  Merging.  Thanks!

RKStrand:
@Myoldmopar Awesome--thanks for cleaning things up and merging this, Edwin!

PR #6195, opened 2017-07-03
LWR between Buildings

xuanluo113:
Modeling Long-wave Radiant Heat Exchange between Buildings
---------------------
Thermal simulation at urban scale aims at understanding and efficiently directing energy flows between different subsystems, for example, how to reduce impact of nearby buildings on energy performance of a target building. Within common thermal network models, the longwave radiation heat fluxes on wall surfaces can be considered at different levels of detail. However, the long-wave radiant heat exchange between exterior surfaces of nearby buildings is over-simplified in EnergyPlus, causing under or over-estimate of exterior surface temperature. We propose to modify the heat balance of the building’s exterior surfaces to explicitly consider their long-wave radiant heat exchange with other exterior surfaces (which can include surfaces of shading objects if their temperatures are known) of the same buildings or nearby buildings. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


design/FY2017/NFP-Modeling LWR Heat Exchange between Buildings.md
```diff
@@ -0,0 +1,225 @@
...
+	       \object-list SurroundingSurfacesViewFactorNames
+	       \note 
+	  A11, \field Surrounding Surfaces Temperature Name
+	       \type object-list
+	       \object-list SurroundingSurfaceTemperatures
+	       \note 
+	  N2 , \field Number of Vertices
+	  …
+	
+	SurfaceProperty:ViewFactors,
```
macumber, line 99: Sometimes you will want view factors to surfaces in the input file and other times you will not.  Is it possible to use surface temperatures in the simulation for some view factors but override them for other surfaces (either shading surfaces or surfaces simply not in the simulation)?

macumber: Please coordinate with existing objects ZoneProperty:UserViewFactors:bySurfaceName and ComfortViewFactorAngles

macumber: Is there anyway to input/override the ground or sky temperature?

macumber: I don't like how Surrounding Surfaces View Factor Name and Surrounding Surfaces Temperature Name are currently factored out.  Right now they are both position based so you have to line up surface view factors and temperatures between the two objects.   This makes it hard to reuse these objects for other surfaces, unless every surface is listed.  I would consider combining SurfaceProperty:ViewFactors and SurfaceProperty:SurroundingSurfaceTemperatures into one object.  

I think my preference would be to use simulation surface temperatures for all surfaces and subsurfaces.  Then let the user specify a temperature schedule for all shading surfaces.  That way the user need only input the SurfaceProperty:ViewFactors object for each surface, temperatures would be calculated by the simulation or user input.  Users would have to enter at least a dummy shading surface for every shading surface that was used in a view factor.  I think it would also be useful to assign a shading surface (e.g. on the West facade of another building) to have the same outside temperature as a surface on the West facade of the building being simulated  (i.e. allow a shading surface to be assigned the temperature profile of another surface in the model).

macumber: I would suggest allowing E+ to calculate these view factors and output them in a way that would allow results to be cached between simulations.  It is possible these view factors could also be exported by the external shading calculation generator.

Myoldmopar: > Is there anyway to input/override the ground or sky temperature?

I think the ground/sky temperatures are overridden in the next object.  Which I was going to ask about as well.  What if the user just wants to use the current ground surface and sky temperatures already calculated within E+.  The fields should be optional so they only need to be modified if the user actually wants to input a schedule.  

> Right now they are both position based

I also dislike the having to line up the view factors and schedules between two different objects in the correct positions.  Once a person is doing very detailed urban modeling and they have to line up (not that they'll be doing it manually) 20 different surrounding surfaces, it will be annoying.  I'd be more in favor of like:

```
 SurfaceProperty:SingleSurroundingSurface,
   A1, Name
   N1, ViewFactor
   A2, Temperature Schedule
```

Then in the building surface object, you have one collection object:

```
 SurfaceProperty:SurroundingSurfaces
   A1, Name
   N1, Sky View Factor
   N2, Sky Temp Override Schedule
   N3, Ground View Factor
   N4, Ground Temp Override Schedule
   A2, <name of a SurfaceProperty:SingleSurroundingSurface instance>
   A3, <name of a SurfaceProperty:SingleSurroundingSurface instance>
   A4, <name of a SurfaceProperty:SingleSurroundingSurface instance>
 - etc
```

Then in the BuildingSurface:Detailed object, you only have to add one new field:

```
Name of a SurfaceProperty:SurroundingSurfaces object instance
```

and if that is blank, you aren't doing any fancy surrounding surface temp stuff.  If it is not blank, you'll be able to walk the structure to get everything you need.

It will also set this code up to be enhanced later because we can easily add data to the `SurfaceProperty:SingleSurroundingSurface` object without having to create entirely new objects and arrays.

xuanluo113:
**Michael J Witte <mjwitte@gard.com> Fri, Jul 7, 2017 at 3:06 PM**

The link between the surface and the SurfaceProperty:ViewFactors (and SurroundingSurfaceTemperatures) is overspecified as proposed.
The new fields in BuildingSurface:Detailed reference the names of the SurfaceProperty:* objects
A9, \field Calculate long-wave radiant heat exchange with surrounding surfaces
A10, \field Surrounding Surfaces View Factor Name
A11, \field Surrounding Surfaces Temperature Name
And the SurfaceProperty:* objects have a field that references the Surface Name they are connected to.  We only need one or the other.

The existing SurfaceProperty:* objects are almost all different in this regard, so I can't say one approach is preferred over the other.  Options:

1.  Keep the SurfaceProperty:* objects as-is, with a "Surface Name" field (or follow the pattern of SurfaceProperty:ConvectionCoefficients and make "Surface Name" be field A1) and do not add any new fields to BuildingSurface:Detailed.  As the SurfaceProperty:ViewFactors and SurroundingSurfaceTemperatures objects are processed, they can set an internal variable for each surface to calculate long-wave radiant heat exchange with surrounding surfaces.

Advantages:  No transition required on BuildingSurface:Detailed.
Disadvantages:  Every surface requires a unique pair of SurfaceProperty:ViewFactors and SurroundingSurfaceTemperatures objects.

2.  Keep the new fields in BuildingSurface:Detailed and delete the "Surface Name" field (A2) in the new SurfaceProperty:ViewFactors and SurroundingSurfaceTemperatures objects.

Advantages:  Surfaces which have a similar view of the surroundings could all share the same pair of SurfaceProperty:ViewFactors and SurroundingSurfaceTemperatures objects.  It wouldn't be unreasonable to assume all south-facing surfaces see the same surroundings, for instance.

Disadvantages:  BuildingSurface:Detailed requires a transition.

nrel-bot-3:
@hongtz68 @XuanLuoLBL @lgentile it has been 15 days since this pull request was last updated.

aaron-boranian:
Modeling longwave radiant heat exchange between building exterior surfaces and shades

Longwave radiant heat exchange is a crucial factor in the building exterior surface heat balance analysis. Currently EnergyPlus takes into account the longwave radiant heat exchange between building surface and the environment (i.e., ground, air, and sky), but ignores the heat exchange between the exterior surfaces/shades of adjoining buildings. Modeling such phenomena can lead to more accurate estimations of the exterior surface conditions, especially in urban canyon context, which is essential in predicting the building energy performance and investigating the coupling effect between buildings and microclimate.

Migrated from [UserVoice](http://energyplus.uservoice.com/forums/258860/suggestions/9674526) feedback from  @rongpengzhang 

**NOTE**: This seems to be related to PR #6227 would @XuanLuoLBL like to chime in?

mjwitte:
@aaron-boranian This was added to v8.8 in #6227 which superceded this PR.

PR #6197, opened 2017-07-03
Improved control of outdoor air system water coils

rraustad:
Pull request overview
---------------------
HVAC controllers were simulated prior to actual simulation of air loop and outdoor air system. This change moves the call to HVAC controllers to where water coils are simulated. This change corrects #6148. Control method for water coils in main air loop has not changed.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
Branch not ready for prime time. Testing speed only in this push.
Duplicate branch to #6176. Only one of these two branches will merge to develop.

src/EnergyPlus/HVACHXAssistedCoolingCoil.cc
```diff
@@ -831,6 +839,10 @@ namespace HVACHXAssistedCoolingCoil {
 			}
 
 			Error = CoilOutputTempLast - Node( HXAssistedCoil( HXAssistedCoilNum ).HXExhaustAirInletNodeNum ).Temp;
+			if ( Iter > 40 ) { // check for oscillation (one of these being negative and one positive) before hitting max iteration limit
```
rraustad, line 842: Saw oscillations while testing and added this to avoid max iter limit warnings. Nothing can be done if iterations oscillate back and forth with same result.

src/EnergyPlus/MixedAir.cc
```diff
@@ -672,12 +678,22 @@ namespace MixedAir {
 			// Coil Types
 		} else if ( SELECT_CASE_var == WaterCoil_Cooling ) { // 'Coil:Cooling:Water'
 			if ( Sim ) {
-				SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
```
rraustad, line 675: OA sys coils will now iterate in unison with HVAC controller

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2279,6 +2279,7 @@ namespace SimAirServingZones {
 			for ( AirLoopPass = 1; AirLoopPass <= 2; ++AirLoopPass ) {
 
 				SysReSim = false;
+				AirLoopControlInfo( AirLoopNum ).AirLoopPass = AirLoopPass; // save for use without passing as argument
```
rraustad, line 2315: save AirLoopPass instead of passing to many functions

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2631,6 +2634,8 @@ namespace SimAirServingZones {
 					// Re-evaluate air loop components with new actuated variables
 					++NumCalls;
 					SimAirLoopComponents( AirLoopNum, FirstHVACIteration );
+					// pass inner iteration loop (i.e., SolveWaterCoilController) result back to this loop
+					ControllerConvergedFlag = PrimaryAirSystem( AirLoopNum ).ControlConverged( AirLoopControlNum );
```
rraustad: Hard to pass back convergence of OA sys water coils. This method eased implementation.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2669,9 +2674,179 @@ namespace SimAirServingZones {
 	}
 
 	void
+	SolveWaterCoilController(
```
rraustad, line 2716: new function to iterate on OA sys water coils with HVAC controller

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2739,14 +2915,15 @@ namespace SimAirServingZones {
 		IterTot = 0;
 
 		AirLoopConvergedFlag = true;
+		BypassOAController = false; // not exactly sure of this but it seems all controllers need to be simulated -- don't bypass
```
rraustad, line 2954: new flag to tell HVAC controller to either simulate or bypass if this controller is in OA system

src/EnergyPlus/WaterCoils.cc
```diff
@@ -980,6 +983,11 @@ namespace WaterCoils {
 			MyCoilReportFlag = true;
 			InitWaterCoilOneTimeFlag = false;
 			PlantLoopScanFlag = true;
+
```
rraustad, line 986: water coils, and HXAssistedWaterCoils, now get controller name and index. Can't do this in GetInput for water coils since this will cause a recursive call to GetInput.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -375,6 +377,14 @@ namespace HVACControllers {
 				CheckEquipName( ControlNum ) = false;
 			}
 		}
+
```
rraustad, line 380: ManageContollers wil now return if controller is in OA system and instead be simulated in SolveWaterCoilController where controller will be simulated inside the iteration loop of OA sys water coils.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -2174,51 +2184,6 @@ namespace HVACControllers {
 	}
 
 	void
-	LimitController(
```
rraustad, line 2177: unused function deleted

src/EnergyPlus/HVACControllers.cc
```diff
@@ -3555,6 +3520,47 @@ Label100: ;
 	}
 
 	void
+	GetControllerNameAndIndex(
```
rraustad, line 3523: if controllers are to be simulated with water coils then water coils need to know which controller is attached.

src/EnergyPlus/HVACHXAssistedCoolingCoil.cc
```diff
@@ -290,23 +291,25 @@ namespace HVACHXAssistedCoolingCoil {
 		// REFERENCES:
 
 		// Using/Aliasing
```
rraustad, line 293: just rearranged these in alphabetical order.

rraustad:
Save a time consuming unit test this PR is ready for review.
Suggest this method over #6176 (faster for unknown reason).
See #6148 for additional discussion.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -289,11 +290,11 @@ namespace HVACControllers {
 		std::string const & ControllerName,
 		int & ControllerIndex,
 		bool const FirstHVACIteration,
-		int const EP_UNUSED( AirLoopNum ), // unused1208
-		int const AirLoopPass,
```
rraustad, line 293: Will now use AirLoopNum to pass AirLoopPass data instead of passing this variable as argument. Only needed for TraceController functions.

src/EnergyPlus/HVACHXAssistedCoolingCoil.cc
```diff
@@ -612,6 +616,8 @@ namespace HVACHXAssistedCoolingCoil {
 				//         Check node names in heat exchanger and coil objects for consistency
 				CoolingCoilErrFlag = false;
 				CoolingCoilInletNodeNum = GetWaterCoilInletNode( HXAssistedCoil( HXAssistedCoilNum ).CoolingCoilType, HXAssistedCoil( HXAssistedCoilNum ).CoolingCoilName, CoolingCoilErrFlag );
```
rraustad, line 618: New GetInput function to get controller name and index

rraustad:
1 file had err diff's, 5Zone_Unitary_HXAssistedCoil. Why the different weather file name? Looks like the new test for oscillations got rid of most of the warnings.

         ** Warning ** CoilSystem:Cooling:DX:HeatExchangerAssisted "HEAT EXCHANGER ASSISTED COOLING COIL 1" -- Exceeded max iterations (50) while calculating operating conditions.
    -   **   ~~~   **  During Warmup, Environment=CHICAGO ANN CLG .4% CONDNS WB=>MDB, at Simulation time=07/21 00:00 - 00:15
    +   **   ~~~   **  During Warmup, Environment=FEB 1 TO 5, at Simulation time=02/01 00:00 - 00:15
    
     ************* The following recurring error messages occurred.
     *************
     *************  ** Warning ** CoilSystem:Cooling:DX:HeatExchangerAssisted "HEAT EXCHANGER ASSISTED COOLING COIL 1" -- Exceeded max iterations error continues...
    -   *************  **   ~~~   **   This error occurred 70 total times;
    -   *************  **   ~~~   **   during Warmup 64 times;
    +   *************  **   ~~~   **   This error occurred 3 total times;
    +   *************  **   ~~~   **   during Warmup 3 times;

mjwitte:
Hmm, "Feb 1 to 5" is the name of a RunPeriod in this file.  I thought CI forced design day runs only.  @Myoldmopar ?

rraustad:
Looking at 5ZoneAirCooled big table diff (eplustbl.htm.absdiff) I see this. What does this mean? It looks like nonsense.

    72 | Coil:Cooling:Water vs Controller:WaterCoil | OA COOLING COIL 1 vs OA HC CONTROLLER 1 | Design Size Design Coil Load [W] vs Maximum Actuated Flow [m3/s] | 12219.7490287

mjwitte:
Looks like this is from the component sizing table, and the order of the rows changed.  Same as you're seeing in the eio diffs.

rraustad:
These 3 unit tests only fail on the i386-Windows-7-VisualStudio-14 CI and with the same message. Is this something to do with my changes or some other reason? These 3 unit tests run here individually on my home laptop (VS Community 2015 Update 3) in debug and release mode. All unit tests run to completion in debug build and stops at AirTerminalSingleDuctMixer_SimPTAC_HeatingCoilTest with release build. So I am not seeing this error.

    EnergyPlusFixture.TubularDaylightDiffuserCount
    EnergyPlusFixture.FinAndOverhangCount
    EnergyPlusFixture.AirloopHVAC_VentilationRateProcedure
    Running 1 test from 1 test case.[----------] Global test environment set-up.[----------]
    1 test from EnergyPlusFixture[ RUN ] EnergyPlusFixture.AirloopHVAC_VentilationRateProcedureunknown file:
    error: C++ exception with description "invalid string position" thrown in the test body.[ FAILED ]

rraustad:
The DElight files continue to show diff's (ESO big diff's) even though the diff's have nothing to do with HVAC systems. I ran in V8.7 and then in this branch and subtracted the results. There are diff's from V8.7. Then I did the same thing with develop vs this branch. No diff's! Regardless, why should a change to HVAC models (or most other changes we make to E+) affect the DElight model results for illuminance (i.e., why do these files keep showing up with diff's in this and other branches)?

delightresults

mjwitte:
@rraustad The DElight diffs have been happening on Windows CI for months now.  Nobody's figured out why yet.  Just ignore them.

nrel-bot-3:
@rraustad @lgentile it has been 20 days since this pull request was last updated.

rraustad:
Given that the change here is to simulate the HVAC controllers with the OA system coils, I thought I would compare the mixed air node temp and mass flow rate between V8.7 and this branch. I used 5ZoneAirCooled. These data look nearly identical. So this is comforting in that how we previously modeled the water coils (all controllers are modeled before the air loop is simulated) is no different than when water coil controllers are modeled when the OA coil is modeled (air loop coils are still modeled the same as before).

5zoneaircooled

tst/EnergyPlus/unit/OASystemHWPreheatCoil.unit.cc
```diff
@@ -2020,6 +2022,11 @@ namespace EnergyPlus {
...
 		EXPECT_NEAR( WaterCoil( 1 ).TotWaterHeatingCoilRate, WaterCoil( 1 ).InletAirMassFlowRate * CpAir * ( WaterCoil( 1 ).OutletAirTemp - WaterCoil( 1 ).InletAirTemp ), 1.0 ); 
 
+		// test that OA sys water coil bypasses normal controller calls before air loop simulation
+		EXPECT_EQ( "PREHEAT COIL CONTROLLER", HVACControllers::ControllerProps( 1 ).ControllerName );
+		EXPECT_TRUE( HVACControllers::ControllerProps( 1 ).BypassControllerCalc );
+		// test that water coil knows which controller controls the HW coil
+		EXPECT_EQ( WaterCoil( 1 ).ControllerIndex, 1 );
 	}
 
 }
```
rraustad: This is an existing unit test that works. So why does Git warn of the missing line at the end of a unit test?

Myoldmopar: "TL;DR: because source code files can be imported and you want to ensure you don't accidentally glue code together."

Git is showing the warning, but really most programming languages have a warning mechanism for this.  The reason is that C++, like most, have an import capability.  In C++ you can `#include` other C++ files.  And this include isn't magic, it simply brings in the source of the other file directly at the spot of the include statement.  The include doesn't append a newline at the end of it's import, it relies on the user to do what's right in that situation.  Every situation I've ever seen is that you'd want a newline to avoid accidentally gluing code together with imports.  Not sure if that makes super sense; it's getting late.  So if you want more info, a Google of "why do source code files need trailing newlines?" would probably be better than my response tonight.

rraustad:
Remaining tasks:
- [x] OASystemHWPreheatCoil.unit.cc requires a new line at end of file
- [x] plot OA system mass flow rate for 5ZoneAirCooled example file to look for diff's (this file has CW and HW coils in OA system with OA controller using no economizer and modulate flow inputs)
- [x] review defect file one last time to ensure OA sys water coil maintains set point temperature
- [x] remove line 1798 of HVACUnitarySystem as part of PR #6108 comment

rraustad:
The defect file does not appear to have changed and maintains the set point.

6148_1aug17

rraustad:
Control of OA system coils and OA flow rate have not changed in this example file. This is not the best test for this branch since controlling a water coil when air flow is not changing is a relatively simple task. This branch corrects an issue with variable outdoor air and controlling water coils in that type of environment.

5zoneaircooled_oaflowandtemp

rraustad:
Bottom line is that existing example files seem to control the same as before and the defect file now controls OA sys water coils to set point temperature.

nrel-bot-3:

rraustad:
Well this is interesting, I pulled in develop and was not asked to merge. I compiled and got these warnings. I think I've fixed them (since it compiles) but not sure why I had to.

    1>C:\Users\Richard\Documents\GitHub\EnergyPlusTeam\src\EnergyPlus\WaterCoils.cc(6415): error C2039: 'SolveRegulaFalsi': is not a member of 'EnergyPlus::General'
    1>  C:\Users\Richard\Documents\GitHub\EnergyPlusTeam\src\EnergyPlus\General.hh(66): note: see declaration of 'EnergyPlus::General'
    1>C:\Users\Richard\Documents\GitHub\EnergyPlusTeam\src\EnergyPlus\WaterCoils.cc(6415): error C2873: 'SolveRegulaFalsi': symbol cannot be used in a using-declaration
    1>C:\Users\Richard\Documents\GitHub\EnergyPlusTeam\src\EnergyPlus\WaterCoils.cc(6541): error C3861: 'SolveRegulaFalsi': identifier not found
    1>C:\Users\Richard\Documents\GitHub\EnergyPlusTeam\src\EnergyPlus\WaterCoils.cc(6600): error C3861: 'SolveRegulaFalsi': identifier not found

rraustad:
I see, I switched to the other parallel branch. 6148Problems...., It's up to date now. I'l try this again with the correct branch.

src/EnergyPlus/HVACControllers.cc
Myoldmopar: :+1:

src/EnergyPlus/HVACControllers.cc
```diff
@@ -289,11 +290,11 @@ namespace HVACControllers {
...
+		int const AirLoopNum,
```
Myoldmopar, line 293: Always nice to see lines of code going away.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -3555,6 +3520,47 @@ Label100: ;
...
+		// SUBROUTINE INFORMATION:
+		//       AUTHOR         Richard Raustad
+		//       DATE WRITTEN   June 2017
+
+		// PURPOSE OF THIS FUNCTION:
+		// This subroutine checks that the water inlet node number is matched by
+		// the actuator node number of some water coil and passed back controller name and index
+
+		// FUNCTION LOCAL VARIABLE DECLARATIONS:
+		int ControlNum;
```
Myoldmopar, line 3540: Could just define this inside the `for` loop below; no big deal for now though.

src/EnergyPlus/HVACHXAssistedCoolingCoil.cc
Myoldmopar: Did you figure out 40 was a good number?  Or just an educated guess?  I don't have a problem with it, just curious.

src/EnergyPlus/MixedAir.cc
```diff
@@ -672,12 +678,22 @@ namespace MixedAir {
...
 			OACoolingCoil = true;
 		} else if ( SELECT_CASE_var == WaterCoil_SimpleHeat ) { // 'Coil:Heating:Water')
 			if ( Sim ) {
-				SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
+				// get water coil and controller data if not called previously
+				if ( CompIndex == 0 ) SimulateWaterCoilComponents( CompName, FirstHVACIteration, CompIndex );
+				// iterate on OA sys controller and water coil at the same time
+				SolveWaterCoilController( FirstHVACIteration, AirLoopNum, CompName, CompIndex, WaterCoil( CompIndex ).ControllerName, WaterCoil( CompIndex ).ControllerIndex, false );
+				// set flag to tell HVAC controller it will be simulated only in SolveWaterCoilController()
+				ControllerProps( WaterCoil( CompIndex ).ControllerIndex ).BypassControllerCalc = true;
```
Myoldmopar, line 696: At first glance I thought, "we really need to wrap up all this repeated stuff into a single little worker function", but I noticed down a bit they aren't actually identical, so there would have to be logic in the little worker function, and there wouldn't be much to gain.

In the future, if the coils all share a base class, we can use polymorphism to collapse a bunch of these select case branches down, but that's not today, it's the future.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2663,7 +2668,13 @@ namespace SimAirServingZones {
 
 					// Re-evaluate air loop components with new actuated variables
 					++NumCalls;
+					// this call to SimAirLoopComponents will simulate the OA system and set the PrimaryAirSystem( AirLoopNum ).ControlConverged( AirLoopControlNum ) flag for controllers of water coils in the OA system
+					// for controllers not in the OA system, this flag is set above in this function
 					SimAirLoopComponents( AirLoopNum, FirstHVACIteration );
+					// pass convergence flag from OA system water coils (i.e., SolveWaterCoilController) back to this loop
+					// for future reference, the PrimaryAirSystem().ControlConverged flag is set while managing OA system water coils.
+					// If convergence is not achieved with OA system water coils, suspect how this flag is passed back here or why OA system coils do not converge
+					ControllerConvergedFlag = PrimaryAirSystem( AirLoopNum ).ControlConverged( AirLoopControlNum );
```
Myoldmopar, line 2677: :+1: for the future dev debugging tips here

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2702,9 +2713,176 @@ namespace SimAirServingZones {
...
+		// Total number of times SimAirLoopComponents() has been invoked
+
+		// SUBROUTINE PARAMETER DEFINITIONS:
+		// Maximum iterations of an air system/controllers simulation sequence
+		int const MaxIter( 50 );
+		static gio::Fmt fmtLD( "*" );
+
+		// SUBROUTINE LOCAL VARIABLE DEFINITIONS
+		// TRUE if controller supports speculative warm restart
+		bool AllowWarmRestartFlag;
```
Myoldmopar, line 2768: In the future, go ahead and initialize the variable right here where you declare them when possible.  No need to separate it out.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2702,9 +2713,176 @@ namespace SimAirServingZones {
...
+
+			if ( !ControllerConvergedFlag ) {
+				// Only check abnormal termination if not yet converged
+				// The iteration counter has been exceeded.
+				if ( Iter > MaxIter ) {
+
+					// The warning message will be suppressed during the warm up days.
+					if ( !WarmupFlag ) {
+						++ErrCount;
+						if ( ErrCount < 15 ) {
```
Myoldmopar, line 2837: 15. Again I'm just being curious if this was a trial and error value or if it just made sense to use this number.

tst/EnergyPlus/unit/OASystemHWPreheatCoil.unit.cc
```diff
@@ -2020,6 +2022,12 @@ namespace EnergyPlus {
 		CpAir = PsyCpAirFnWTdb( Node( AirInletNodeNum ).HumRat, Node( AirInletNodeNum ).Temp );
 		EXPECT_NEAR( WaterCoil( 1 ).TotWaterHeatingCoilRate, WaterCoil( 1 ).InletAirMassFlowRate * CpAir * ( WaterCoil( 1 ).OutletAirTemp - WaterCoil( 1 ).InletAirTemp ), 1.0 ); 
 
+		// test that OA sys water coil bypasses normal controller calls before air loop simulation
+		EXPECT_EQ( "PREHEAT COIL CONTROLLER", HVACControllers::ControllerProps( 1 ).ControllerName );
+		EXPECT_TRUE( HVACControllers::ControllerProps( 1 ).BypassControllerCalc );
```
Myoldmopar, line 2027: :+1:

Myoldmopar (approved):
Code changes look great.  I had a few comments, but found nothing wrong.  I'll pull this branch and run a couple tests and see how CI is doing with it.

src/EnergyPlus/HVACControllers.cc
rraustad: These old habits are hard to break, but slowly getting there. I do get a little confused when a loop variable is used multiple times and inside IF blocks where to initialize or to initialize multiple times.

src/EnergyPlus/HVACHXAssistedCoolingCoil.cc
rraustad: 40 is just a number far enough into the iteration count (max = 50) to let the solution converge before checking if there are oscillations since the solution could easily be positive and negative on successive iterations and still not have a problem converging. I just wanted a few iterations left before checking this.

src/EnergyPlus/SimAirServingZones.cc
rraustad: This new function is very similar to SolveAirLoopControllers, just focuses on a single air loop instead of looping through all of them (maybe could have done something trick with the existing function, I didn't go back once I got it working). 15 is also used there so I didn't change it. These warnings could be spread out over the year instead of successive starting at some point in time so we print X of them before handing over to ShowRecurringWarningErrorAtEnd.

Myoldmopar:
Build is fine.  One file has err diffs (5Zone_Unitary_HXAssistedCoil), so I ran it with develop and this branch and found the diffs are just that the max iterations for one of the coils went down from 70 to 3 (all during warmup on both develop and this branch) and some other small max iteration diffs.  Overall it was an improvement.  The performance of this branch does show a small regression of 0.15%, but I believe that was expected since we are coupling the iterations.  I'm OK with this since it provides the control we want.

I am inclined to merge this in.  Anything further @mjwitte @rraustad ?

rraustad:
@Myoldmopar @mjwitte nothing more on my end unless something needs to be changed. Thanks for the review.

mjwitte:
@Myoldmopar I haven't reviewed this in detail, but the overall concept sounds like a great step forward.   Have you confirmed this fixes the defect file?  Otherwise, I have not further comment.

Myoldmopar:
I ran the defect file in develop and this branch.  Error files fine in both.  I poked around at the preheat coil outputs and it seems to be under control.  There have been lots of results demonstrating the new code.  I think it's ready.

PR #6198, opened 2017-07-03
Enhancement for External Shading Calculation

xuanluo113:
Enhancement for External Shading Calculation
---------------------
External shading, i.e., shading of an exterior building surface, is closely connected with energy use in buildings for heating, cooling, and lighting and with the occupants’ visual and thermal comfort. Accurate calculation of the shading that building environment, building elements or shading devices may cast on the building envelope, plays a key role in thermal load calculations and building energy simulation. However, external shading calculation for a target building in an urban context can be considerably slow considering many surrounding surfaces may shade a target surface. Various external tools can be used to more efficiently pre-calculate the shading fractions for many target buildings in the urban context. We propose to add links to EnergyPlus to enable the import of external shading calculation results, and saving the EnergyPlus shading calculation results for reuse.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/design/FY2017/NFP-Enhancement for External Shading Calculation .md
+++ b/design/FY2017/NFP-Enhancement for External Shading Calculation .md
@@ -92,7 +92,7 @@ I've always viewed external (or cached) shading calculations as an all-or-nothin
 
 > Thanks for the great advice. We adopted the suggested approach.
 
-**Stuart Mentzer <Stuart_Mentzer@objexx.com> Thu, Jun 1, 2017**
+**Stuart Mentzer Thu, Jun 1, 2017**
 
 External shading would greatly benefit EnergyPlus for single building modeling as well as this is a big performance/scalability bottleneck. If a preferred external tool can be integrated into the "normal" EnergyPlus in place of or as an optional alternative to the current shading approach I know a few cloud providers who would be very pleased. (Perhaps they would even contribute to this work.)  I'm reaching out. If I hear anything I'll let you know.
 Is reflection also being considered for external treatment? If not the octree system I added for daylighting should be used for reflection to address scalability.
--- a/design/FY2017/NFP-Enhancement for External Shading Calculation .md
+++ b/design/FY2017/NFP-Enhancement for External Shading Calculation .md
@@ -1,0 +1,104 @@
+Enhancement for External Shading Calculation 
+================
+**Tianzhen Hong, Xuan Luo, LBNL**
+
+**Original: May 30, 2017**
+
+**Revised: June 27, 2017**
+
+# Justification for New Feature 
+
+External shading, i.e., shading of an exterior building surface, is closely connected with energy use in buildings for heating, cooling, and lighting and with the occupants’ visual and thermal comfort. Accurate calculation of the shading that building environment, building elements or shading devices may cast on the building envelope, plays a key role in thermal load calculations and building energy simulation. However, external shading calculation for a target building in an urban context can be considerably slow considering many surrounding surfaces may shade a target surface. Various external tools can be used to more efficiently pre-calculate the shading fractions for many target buildings in the urban context. We propose to add links to EnergyPlus to enable the import of external shading calculation results, and saving the EnergyPlus shading calculation results for reuse.  
+
+# Team Discussion so far
+Michael Witte and Stuart Mentzer provided suggestions on input details and external tool resources.
+
+# Overview 
+Solar shading affects energy use in a building by reducing the solar heat gains received by opaque surfaces or the transmitted and absorbed solar radiation of windows. External shading also influences daylighting level in a room and the view to the exterior. Accurate calculation of solar shading on building exterior surfaces is of great importance in whole building energy simulation. Modeling a building in the urban context may involve many shading surfaces from adjacent buildings for a target building, which can significantly slow down EnergyPlus simulations due to the currently implemented shading calculation algorithms. Furthermore, for urban building energy simulation which involves many buildings, there can be more efficient ways to do the shading calculations than the current way of doing shading calculations for every building separately in EnergyPlus. There is also the potential of using GPU and other parallel computing techniques for building shading calculations.
+
+Various simulation tools, such as Radiance and Daysim, employing state-of-the-art ray-tracing simulation techniques, can be used to pre-calculate the shading fractions for each exterior building surface. However, in the current implementation, EnergyPlus doesn’t allow overriding exterior shading calculations with external data either by EMS or co-simulation interface. Considering a large amount of information and the complexity of calculating the dynamic shading in EnergyPlus, allowing this external input would enable a huge potential to speed up EnergyPlus in urban-scale simulations. Another benefit is to enable reusing the shading results for parametric runs which usually do not change external shading. Adding this feature to EnergyPlus would also allow OpenStudio to run radiance simulations before the EnergyPlus simulation and reusing the radiance calculations for shading. 
+
+
+# Approaches 
+We would add an object **Shading:CalculationMethod** listed as follows. The object contains a mandatory **Field: External Shading Calculation Method**, which has two options namely **UseSchedules** and **InternalCalculation**. The field indicates whether the external shading is pre-calculated and imported as a schedule. If **UseSchedules** is chosen, the schedule field can be linked to a Schedule object **Schedule:File**. The name of the schedule is defined in the **Field: Surfaces External Shaded Fraction Schedule Name**. The **Schedule:File** object points to a CSV text file that has external shaded fraction scheduled values for each external surface considered in the calculation. The schedule values are listed in each column, and the header of the column points to the name of the corresponding exterior surface that receives the shading. For each column, the text contains 8760-8784 hours of surface external shaded fraction data, which is pre-calculated and stored in the text file. The Schedule object would be set up in the initialization, and the results can be loaded at runtime or overwritten through co-simulation or EMS code. The data pairs listed as extensible variables map each building exterior surface to a column in the schedule CSV file. During initializing, all columns are loaded at one, and each is saved corresponding to an exterior surface, and the data can be used in run time.
+
+	Shading:CalculationMethod,
+          \memo Used to specify the external shading calculation method used.
+          \extensible: -- duplicate last set of Surfaces – Schedule Mapping (last 2 fields), remembering to remove ; from "inner" fields.
+      A1, \field External Shading Calculation Method
+          \type choice
+          \key UseSchedules  
+          \key InternalCalculation
+          \default InternalCalculation
+          \note If UseSchedules is chosen, Surfaces External Shaded Fraction Schedule Name is required.
+      A2, \field Surfaces External Shaded Fraction Schedule Name
+          \type object-list
+          \object-list ScheduleNames
+          \note Used only if External Shading Calculation Method is UseSchedules.
+          \note Schedule values should be between 0 and 1.
+      N1, \field Number of Data Pairs of the Surface – Column in the schedule Mapping
+          \required-field
+          \type integer
+          \minimum 1
+          \maximum 500
+          \note The data pairs map an exterior surface to a column in the schedule file
+      A3, \field Surface 1 Name
+          \required-field
+          \type object-list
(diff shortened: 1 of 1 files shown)


Myoldmopar:
What is going on with all these external shading pull requests?  This is the fourth today...

xuanluo113:
@Myoldmopar 
Sorry I was trying to rename the old NFP pull request but it doesn't show my commits. I closed the old one and deleted the branch. This the the new one and the final version.

design/FY2017/NFP-Enhancement for External Shading Calculation .md
```diff
@@ -0,0 +1,104 @@
...
+================
+**Tianzhen Hong, Xuan Luo, LBNL**
+
+**Original: May 30, 2017**
+
+**Revised: June 27, 2017**
+
+# Justification for New Feature 
+
+External shading, i.e., shading of an exterior building surface, is closely connected with energy use in buildings for heating, cooling, and lighting and with the occupants’ visual and thermal comfort. Accurate calculation of the shading that building environment, building elements or shading devices may cast on the building envelope, plays a key role in thermal load calculations and building energy simulation. However, external shading calculation for a target building in an urban context can be considerably slow considering many surrounding surfaces may shade a target surface. Various external tools can be used to more efficiently pre-calculate the shading fractions for many target buildings in the urban context. We propose to add links to EnergyPlus to enable the import of external shading calculation results, and saving the EnergyPlus shading calculation results for reuse.
```
macumber, line 11: Also mountains and nearby geography can shade buildings

design/FY2017/NFP-Enhancement for External Shading Calculation .md
```diff
@@ -0,0 +1,104 @@
...
+
+# Team Discussion so far
+Michael Witte and Stuart Mentzer provided suggestions on input details and external tool resources.
+
+# Overview 
+Solar shading affects energy use in a building by reducing the solar heat gains received by opaque surfaces or the transmitted and absorbed solar radiation of windows. External shading also influences daylighting level in a room and the view to the exterior. Accurate calculation of solar shading on building exterior surfaces is of great importance in whole building energy simulation. Modeling a building in the urban context may involve many shading surfaces from adjacent buildings for a target building, which can significantly slow down EnergyPlus simulations due to the currently implemented shading calculation algorithms. Furthermore, for urban building energy simulation which involves many buildings, there can be more efficient ways to do the shading calculations than the current way of doing shading calculations for every building separately in EnergyPlus. There is also the potential of using GPU and other parallel computing techniques for building shading calculations.
```
macumber, line 17: GPU acceleration would be very useful but I think the main thrust of this work is caching previously computed values right?

design/FY2017/NFP-Enhancement for External Shading Calculation .md
```diff
@@ -0,0 +1,104 @@
...
...Calculation**. The field indicates whether the external shading is pre-calculated and imported as a schedule. If **UseSchedules** is chosen, the schedule field can be linked to a Schedule object **Schedule:File**. The name of the schedule is defined in the **Field: Surfaces External Shaded Fraction Schedule Name**. The **Schedule:File** object points to a CSV text file that has external shaded fraction scheduled values for each external surface considered in the calculation. The schedule values are listed in each column, and the header of the column points to the name of the corresponding exterior surface that receives the shading. For each column, the text contains 8760-8784 hours of surface external shaded fraction data, which is pre-calculated and stored in the text file. The Schedule object would be set up in the initialization, and the results can be loaded at runtime or overwritten through co-simulation or EMS code. The data pairs listed as extensible variables map each building exterior surface to a column in the schedule CSV file. During initializing, all columns are loaded at one, and each is saved corresponding to an exterior surface, and the data can be used in run time.
```
macumber, line 23: Why not add to existing ShadowCalculation object?  I suggest adding an option 'ExternalSchedules' to ShadowCalculation::Calculation Method.

macumber: Alternate method would be to add a field to each surface to point to a shading schedule similar to Transmittance Schedule Name of Shading:Site:Detailed. Have you compared pros and cons of these two approaches?

macumber: What happens if a schedule is not input for an exterior surface?  What happens if it is input for an interior surface?

macumber: When the shading calculation is 'UseSchedules' is there any need to read shading surfaces from the input file at all?

macumber: Is shading applied to diffuse and direct solar?  Do we need two schedules per surface, one for direct shading and one for diffuse?

design/FY2017/NFP-Enhancement for External Shading Calculation .md
```diff
@@ -0,0 +1,104 @@
...
+
+Various simulation tools, such as Radiance and Daysim, employing state-of-the-art ray-tracing simulation techniques, can be used to pre-calculate the shading fractions for each exterior building surface. However, in the current implementation, EnergyPlus doesn’t allow overriding exterior shading calculations with external data either by EMS or co-simulation interface. Considering a large amount of information and the complexity of calculating the dynamic shading in EnergyPlus, allowing this external input would enable a huge potential to speed up EnergyPlus in urban-scale simulations. Another benefit is to enable reusing the shading results for parametric runs which usually do not change external shading. Adding this feature to EnergyPlus would also allow OpenStudio to run radiance simulations before the EnergyPlus simulation and reusing the radiance calculations for shading.
```
macumber, line 19: Reusing shading calculation results between runs would require that both models have the same surfaces and subsurfaces with the same names.

design/FY2017/NFP-Enhancement for External Shading Calculation .md
macumber: Would EnergyPlus be able to output this Schedule:File for use in future runs?

macumber: Column headers with surface name are not required since column mapping is performed in input file.

macumber: Why require use of Schedule:File, allowing use of other schedule types might be useful.

macumber: Having this implemented per surface would allow EMS programs to implement shading calculations for a subset of surfaces and let remainder of surfaces use E+ method.

Myoldmopar:
It feels like there are going to be a lot of gotchas on this.  

- During development of a model, when you are modifying surfaces/geometry, you will have to avoid this because the schedule csv data could accidentally be used from a prior version where the shading values in the file no longer match up because your surfaces changed.  So you'll want to use it once you have your surface geometry nailed down.

- And obviously if you are manipulating the shading itself, you'll have to recalculate new shading values, so you have to have shading nailed down.

- And you will only be able to use it in one location, because if you start changing climates, the shading will change.  So you'll want to use it once you have your location selected.

- And at a single location, you will only want to use it for one orientation, because if someone is doing appendix G accreditation and they need to rotate their building 90 degrees 4 times, the values will no longer be valid.

So you can use this for one building configuration, in one location, in one orientation. once you have shading nailed down.

 ...

Even with that, I see some benefit in being able to call out to have another program do the more accurate shading calculation, but I can't see the benefit in caching.  Once you start doing large studies, there are so many things that can change from one run to the next that it seems like it would be exceedingly easy for us to let users accidentally run the wrong data and start getting bogus answers.

Convince me I'm wrong.  Assuming I am wrong and we proceed like this, I agree with @macumber on several of his comments.

macumber:
An idea (not sure its a good one) would be to compute a checksum on the input objects (e.g. surfaces, etc) that should not change when using the cached shading results.  That way you could say "the shading calcs in this object are valid if the checksum of the relevant objects is X", you would issue a warning if that was invalidated.

Myoldmopar:
> compute a checksum on the input objects (e.g. surfaces, etc) that should not change when using the cached shading results.

And a checksum on the entire contents of the weather file?

xuanluo113:
On 7/10/2017 7:03 AM, Jason Glazer wrote:
I think the approach of reading a bunch of hourly values via a CSV file seems like a generic function. I would suggest creating a new object called something like Schedule:File:Multiple and reference it. That way other objects in the future can reference it. The multipliers would be used from corresponding columns from the CSV file.

That would change the object to:

```
Shading:CalculationMethod,
      \memo Used to specify the external shading calculation method used.
      \extensible: -- duplicate last set of Surfaces – Schedule Mapping (last 2 fields), remembering to remove ; from "inner" fields.
  A1, \field External Shading Calculation Method
      \type choice
      \key UseSchedules  
      \key InternalCalculation
      \default InternalCalculation
      \note If UseSchedules is chosen, Surfaces External Shaded Fraction Schedule Name is required.
  A2, \field Surfaces External Shaded Fraction Schedule:File:Multiple Name
      \type object-list
      \object-list ScheduleFileMultipleNames
      \note Used only if External Shading Calculation Method is UseSchedules.
      \note Schedule values should be between 0 and 1.
  A3, \field Surface 1 Name
      \required-field
      \type object-list
      \object-list SurfaceNames
  A4, \field Surface 2 Name
      \required-field
      \type object-list
      \object-list SurfaceNames


See field for first and last column numbers in the following new object.

Schedule:File:Multiple,
 \min-fields 5
       \memo A Schedule:File points to a text computer file that has 8760-8784 hours of data.
  A1 , \field Name
       \required-field
       \type alpha
       \reference ScheduleFileMultipleNames
  A2 , \field Schedule Type Limits Name
       \type object-list
       \object-list ScheduleTypeLimitsNames
  A3 , \field File Name
       \required-field
       \retaincase
  N1 , \field First Column Number
       \required-field
       \type integer
       \minimum 1
  N2 , \field Last Column Number
       \required-field
       \type integer
       \minimum 1
  N2 , \field Rows to Skip at Top
       \required-field
       \type integer
       \minimum 0
  N3 , \field Number of Hours of Data
       \note 8760 hours does not account for leap years, 8784 does.
       \note should be either 8760 or 8784
       \default 8760
       \minimum 8760
       \maximum 8784
  A4 , \field Column Separator
       \type choice
       \key Comma
       \key Tab
       \key Fixed
       \key Semicolon
       \default Comma
  A5 , \field Interpolate to Timestep
       \note when the interval does not match the user specified timestep a "Yes" choice will average between the intervals request (to
       \note timestep resolution.  a "No" choice will use the interval value at the simulation timestep without regard to if it matches
       \note the boundary or not.
       \type choice
       \key Yes
       \key No
       \default No
  N4 ; \field Minutes per Item
       \note Must be evenly divisible into 60
       \type integer
       \minimum 1
       \maximum 60
```

This is more generic approach and disconnects reading a CSV file from shading and the user would only need to reference a single object.

Jason

design/FY2017/NFP-Enhancement for External Shading Calculation .md
amirroth: It's the same mechanism either way.

design/FY2017/NFP-Enhancement for External Shading Calculation .md
amirroth: And the same building location and orientation. Would need to add a checksum if we want to detect cached shading calcs automatically. And maybe also if we want to do this manually, just to avoid surprising results.

design/FY2017/NFP-Enhancement for External Shading Calculation .md
amirroth: I think the way to do this would be to load up schedules for all surfaces and then just ignore or override schedules for selected surfaces. The dominant use case will be external calculations for all surfaces so the input method should be tailored towards that I would think.

Another aspect of this is that for processing speed purposes the data should be probably be organized row-major with surfaces as columns and time steps as rows, so you can easily peel off the shading fractions for a given time step for all surfaces. That kind of organization is not really compatible with per surface shading-fraction schedule files.

Issue #6220, opened 2017-07-24
System output provided and single duct zone HVAC equipment load calculation use different cp air values


Nigusse:
Issue overview
--------------
System output provided variable "SysOutputProvided" calculated [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SingleDuct.cc#L2902-L2907) and
the load passed to reheat heating coil calculated [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L769-L772) use different cp value of supply air that results in remaining output required variable to have residual value updated [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L3687-L3689) and passed to the next priority zone HVAC equipment.  

The cp value of air based on the supply air condition entering the air terminal reheat coil and cp value of air determined using zone air condition can result in load difference of up to 15 Watts (based on defect file tested). This load difference cased by cp value calculation reference is significant enough to trigger the second priority zone HVAC equipment connected to a zone.

Suggestion: change the reheat coil load calculation for example [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SingleDuct.cc#L2902-L2907) to this form:
QZnReq = QZoneMax2 - MassFlow *  ( CpAirSys * SysInlet( SysNum ).AirTemp - CpAirZn * ZoneTemp );
CpAirSys = PsyCpAirFnWTdb( SysInlet( SysNum ).AirHumRat, SysInlet( SysNum ).AirTemp); 

Similar change may be required for gas reheat coils [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SingleDuct.cc#L2895-L2900) and steam heating coils [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SingleDuct.cc#L2909-L2914).

### Details
Some additional details for this issue (if relevant):
 - Win64
 - EnergyPlus Version 8.7
 - observed during investigation of issue #6189

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Closed via #6226.

PR #6226, opened 2017-07-27
Calculated system load and single duct zone HVAC equipment load calculation uses different cp air values

Nigusse:
Pull request overview
---------------------
As single duct HVAC equipment models meet a load, the calculation of the sensible load met does not match what is calculated by the equipment manager (due to use of different Cp values). This results in a very small remaining load that, if exists, other zone equipment must meet. Changing the calculation of sensible load met to use the same Cp values (in the Mdot * Cp * Delta T equation) corrects the intermittent need for other equipment to operate for very small loads. This change corrects Issue #6220.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)


Nigusse:
@mjwitte Addressed the issue for heating coils used in single duct VAV reheat object (SINGLE DUCT:VAV:REHEAT) objects. However, I also see the same issue in the following air terminal objects:
- SingleDuct:ConstantVolume:Reheat
- SINGLE DUCT:VAV:REHEAT:VS FAN
- SINGLE DUCT:VAVHEATANDCOOL:REHEAT

After reviewing the performance of this fix I will propagate the changes to the other air terminal objects.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -2747,7 +2747,7 @@ namespace SingleDuct {
 		UpdateSys( SysNum );
 
 		// At the current air mass flow rate, calculate heating coil load
-		QActualHeating = QToHeatSetPt - MassFlow * CpAirZn * ( SysInlet( SysNum ).AirTemp - ZoneTemp ); // reheat needed
+		QActualHeating = QToHeatSetPt - MassFlow * ( CpAirSysIn * SysInlet( SysNum ).AirTemp - CpAirZn * ZoneTemp ); // reheat needed
```
rraustad: This is very different from how coil models calculate Q. This is Mdot * ( delta CpT) whereas the coil models are Mdot * Cpin * dT. So why the difference? No reason to hold this up, just thinking globally.

rraustad:
@Nigusse please post a plot of results so reviewers have an easier time of understanding these changes.

src/EnergyPlus/SingleDuct.cc
rraustad: Would it be better to change the way RemainingOutputRequired is calculated instead of changing all the other places where capacity is calculated?

Nigusse:
Now the reheat coil of SINGLE DUCT:VAV:REHEAT object corrected heating load is calculated using an average CpAirAvg instead of CpAirZone shown [here](https://github.com/NREL/EnergyPlus/blob/149678381_Issue%236220/src/EnergyPlus/SingleDuct.cc#L2897-L2917). Also the SysOutputProvided by the terminal air single duct system uses an average CPAirAvg changed [here ](https://github.com/NREL/EnergyPlus/blob/149678381_Issue%236220/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L768-L772).  The SysOutputProvided change may impact all single duct air terminal objects with system output.

Nigusse:
@mjwitte @rraustad Diffs are expected simply due to changes how the CpAir values are calculated.  Very small zone air temperatures differences were observed as shown in the embedded plots below. 

zoneairtemp_ 6220

Also a unit heater which was previously on (before the fix) is now turned off. The CPAir value difference were the reason for triggering the unit heater heating coil on. Now when using an average CpAir those small loads due to Cp values diffs are gone and the unit heater is off. In this case the unit heater was a second priority equipment while the air distribution unit was the first priority. 

rheatcoilheatingrate_ 6220

In another zone the zone air temperatures difference with narrow vertical axis scale shows the diffs better. The corresponding reheat coil heating rate plots are also show below.

zoneairtemp2_ 6220

The corresponding reheat coil heating rate plot are shown below. The heating rate differences here are under 10 Watts.

rheatcoilheatingrate2_ 6220

Nigusse:
Added defect idf and result file.

rraustad:
I am starting to remember why and how we do things with the equipment models. The sensible load met is calculated as Mdot*DelatH at constant (minimum) humidity ratio. When the models iterate to meet a load (e.g., SensibleLoadToHeatingSP), the result is a zone temperature that is within 0.001 of the zone temperature set point. So in your plot above, if the temperature set point is 21.111 C, then this is about as close as you can get. In addition, if the sensible calculations are all done the same then there should not be a problem with RemainingSensibleLoad and other equipment trying to meet ones or tenths of watts.

    MinHumRat = min( Node( InletNode ).HumRat, Node( OutletNode ).HumRat );
    SensLoadMet = AirMassFlow * ( PsyHFnTdbW( Node( OutletNode ).Temp, MinHumRat ) - PsyHFnTdbW( Node( InletNode ).Temp, MinHumRat ) );

Nigusse:
@rraustad Did you see the second plot above? The unit heater was turned on for a small load brought about by diffs in the CpAir calculation. After the fix, the unit heater is now off. This was the main reason for posting this issue.

Nigusse:
@rraustad Yes the the heating thermostat setpoint for the this zone in the defect idf was 21.111 C. See the previous plot with heating setpoint temperature add below.

zoneairtemp3_ 6220


I have seen more tighter zone air temperature control after this fix in RefBldgLargeHotelNew2004_Chicago example file as well.

rraustad:
This issue is good for now. We discussed the differences in the way some equipment models use Qsens = MdotDeltaH at constant w, and some use Qsens = MdotCpDeltaT. I guess for now I don't care which method is used, or even if different models use different equations, as long as Corrector has the zone inlet node conditions needed to hit the Tstat set point temp. Knowing that the ADUs use a different equation than other equipment makes me think that this MIGHT cause problems in the future but I don't know at this point what those problems might be.

For now this issue is resolved. Waiting on CI.

rraustad:
- Noting first 20 diff's were warmup convergence which is expected with this type of change. Did not continue looking through remaining 130 files but did spot check a few others with same result.
- Noting 5Zone_Unitary_HXAssistedCoil dropped some warnings and added some warnings. This example file is ripe with warnings.

    -   **   ~~~   **  Routine=PsyWFnTdbTwbPb, Environment=CHICAGO ANN CLG .4% CONDNS WB=>MDB, at Simulation time=07/21 00:00 - 00:15
    -   **   ~~~   **  Input Temperature=-104.69
    +   **   ~~~   **  Routine=CalcMultiSpeedVarSpeedCoilCooling, Environment=CHICAGO ANN CLG .4% CONDNS WB=>MDB, at Simulation time=07/21 00:00 - 00:15
    +   **   ~~~   **  Input Temperature=-115.36

- Noting that HospitalLowEnergy has about 15% fewer SolveAirLoopControllers: Maximum iterations warnings
- Noting ReportDaylightFactors STILL HAS DX coil low air volume flow rate warnings which shouldn't be there.

I don't see anything that would cause pause considering this change.

rraustad:
Reviewed and merged.

rraustad:
@Nigusse did you still need to add a unit test for this change?

Myoldmopar:
Since we couldn't see the CI result badges above, it wasn't obvious that this PR introduced a build warning into the develop branch when it got merged in.  The warning is showing in [these results](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-b9cb942be372e765a93aa96cd02d72597ccab895-x86_64-MacOS-10.9-clang.html)

The build warning is confusing me though.  It is saying that [this line](https://github.com/NREL/EnergyPlus/blame/develop/src/EnergyPlus/SingleDuct.cc#L2617) has an `expression result unused` at column 126, which I don't really see.  Maybe it's just been a long day and I'll see it in the morning.  Does anyone else ( @jasondegraw @mbadams5 @mjwitte @rraustad @Nigusse ) see a problem with that line?

rraustad:
You believe the compiler is bug free? There is obviously nothing wrong there that we can see. I think it's the arguments that are tripping this so I suggest added parens and if that doesn't work a kludge on the next line of:

Real64 tempCp = CpAirAvg; // in case the use of CpAirAvg inside an IF block is causing this ... 

to see if either of these help.

rraustad:
Just a thought, our quality control is failing.

Myoldmopar:
Yeah, this one would've been caught except that the badges aren't showing up all the time.  I'm wondering if we're simply hitting a ratelimit on the website that generates those shields for us.

And if this turns out to be a compiler anomaly then, well, we can at least pass _some_ of the blame down the road...

Myoldmopar:
Also, I just built a release build of develop on Linux and it does not issue that warning for SingleDuct.  I can try my Mac tomorrow.

mbadams5:
I'm surprised this compiles... but apparently is valid C/C++...
```cpp
CpAirAvg = PsyCpAirFnWTdb( 0.5 * ( Node( ZoneNodeNum ).HumRat + SysInlet( SysNum ).AirHumRat), 0.5 * ( Node( ZoneNodeNum ).Temp, SysInlet( SysNum ).AirTemp ) );
```
Keen eyes would notice that the averaging of the temperatures is not using a `+` but rather `,`. Again, I don't know how this compiles but it is valid C/C++.

https://stackoverflow.com/questions/16563235/what-does-a-comma-separated-list-of-values-enclosed-in-parenthesis-mean-in-c-a
https://stackoverflow.com/questions/13209399/c-what-does-int-x-anyint1-anyint2-mean

So basically it is using the value from `SysInlet( SysNum ).AirTemp` and hence `Node( ZoneNodeNum ).Temp` is unused, and the cause of the warning.

rraustad:
Nice. Good eye.

    0.5 * ( Node( ZoneNodeNum ).Temp, SysInlet( SysNum ).AirTemp )

rraustad:
@Nigusse you want to fix this or should I since I merged a failed branch?

rraustad:
@Myoldmopar what do I do in this case? Re-open? push again? what?

Myoldmopar:
Usually in this case, I just make a branch called HotFix-Something to be clear this is a very minor quick fix that should go into develop right away.  I can just do that real quick.  I'll leave it open for CI to tell us how wrong we were, by showing us new diffs from it.

PR #6230, opened 2017-07-31
Decrease Documentation Margins

Myoldmopar:
Left, right, and bottom margins all set to 0.5".  Top was left at 1" because of the way the header is placed on the page.  Content fills page better.  Engineering reference went down about 150 pages, not that that was any motivation for this.

### DIFF
--- a/doc/header.tex
+++ b/doc/header.tex
@@ -3,7 +3,7 @@
 \usepackage{amssymb,amsmath}
 \usepackage{mathspec}
 \defaultfontfeatures{Ligatures=TeX,Scale=MatchLowercase}
-\usepackage[margin=1.0in]{geometry}
+\usepackage[margin=0.5in]{geometry}
 % fixltx2e has been merged into LaTeX2e proper
 % see: https://latex-project.org/ltnews/ltnews22.pdf
 % \usepackage{fixltx2e}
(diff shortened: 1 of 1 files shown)


mjwitte:
@Myoldmopar Hmmm, 4 days on and linux build is still pending.  Everything builds locally, but the 0.5 in margin looks too slim and it causes the page numbers at the bottom of main heading pages to almost fall off the bottom.  Suggest 0.75 in instead - just looks less intense.  Screen shot is 1in, 0.75 in, 0.5 in.
image

Myoldmopar:
I deleted the Linux build, let's see if there was just a glitch in The Matrix.

For the margins, yeah, 0.75 looks good.  I was more concerned about the side margins than the top/bottom anyway.  I'll make it 0.75 and wait to see if CI can do it.

michael-okeefe:
Looks good to me. I like the direction you're going with the 0.75" margins.

Myoldmopar:
@mjwitte CI doesn't seem to have had trouble with this branch, it seems to have just gone down entirely a few days back.  Windows and Linux both.  We'll get that box back up, but it'll have a lot to scrape through so I'm going to merge this to keep it from being in the way of other builds.

PR #6237, opened 2017-08-12
Add liquid/solid k and rho to PCM hysteresis model

Myoldmopar:
Pull request overview
---------------------
This is a followup to the hysteresis phase change model.  It was brought to my attention that if the hysteresis model allowed for direct input of the other physical properties (thermal conductivity and density) at the liquid and solid states, it would enhance the model's ability to match experimental data.

This change is minor, the fixed values of solid/liquid properties are used when outside of the transition region.  Within the transition region the average of the two is used.  This same behavior can be achieved with the tabular variable conductivity table input.  This change simply brings it directly into the object where the user is already modifying liquid/solid state parameters.

- [x] IDD
- [x] Example file
- [x] IORef
- [x] EngRef
- [x] Modify classes and GetInput
- [x] Modify calculation code
- [x] Unit tests


Myoldmopar:
@mjwitte This is a pretty isolated change to the new PCM model.  I know you don't have much review time, so I'm just giving a heads up about it going in soon in case you do want to look it over briefly.

tst/EnergyPlus/unit/PhaseChangeModeling/HysteresisModel.unit.cc
```diff
@@ -502,4 +506,16 @@ TEST_F(HysteresisTest, IntoMeltingAndBackDown)
...
+	EXPECT_NEAR(1.0, this->ModelA.getConductivity(19.0), 0.01);
+	EXPECT_NEAR(1.5, this->ModelA.getConductivity(21.5), 0.01);
+	EXPECT_NEAR(2.0, this->ModelA.getConductivity(24.0), 0.01);
+}
+
+TEST_F(HysteresisTest, TestVariableDensity) {
+	EXPECT_NEAR(3.0, this->ModelA.getDensity(19.0), 0.01);
+	EXPECT_NEAR(3.5, this->ModelA.getDensity(21.5), 0.01);
+	EXPECT_NEAR(4.0, this->ModelA.getDensity(24.0), 0.01);
 }
```
mjwitte: Need a newline at the end here.

mjwitte:
@Myoldmopar There is no transition info here yet.

Myoldmopar:
Thanks @mjwitte.  There shouldn't be any transition info here.  This was a new object just added last month, and just needed to add a couple more fields.  

I'll add the newline.

Thanks for looking this over.  If nothing else, I'll plan on merging later.

mjwitte:
@Myoldmopar Got it (the no transition thing).  Proceed.

Myoldmopar:
This is good and done; merging to get it out of the way for other PRs

PR #6248, opened 2017-08-16
Add air terminal unit sizing option, allow air loops with no return path, and partially allow multiple air loops serving the same zone

mjwitte:
Pull request overview
---------------------
This new feature 
1. Partially removes the limitation of one airloop (AirLoopHVAC) system per zone, allowing multiple air distribution units and multiple return air nodes in the same zone.
2. Adds an option to specify return air flow fraction (AirLoopHVAC field Design Return Air Flow Fraction of Supply Air Flow).
3. Adds options for user-control of sizing with DesignSpecification:AirTerminal:Sizing.
4. Allows an airloop with no return path.
5. Two new example files: 
5ZoneAirCooledWithDOASAirLoop (demostrates two air loops serving the same zones and DesignSpecification:AirTerminal:Sizing and no return path) and DOASDXCOIL_wADPBFMethod_NoReturnPath (demonstrates no return path, results identical to DOASDXCOIL_wADPBFMethod)
6. *Not working yet* Adds options to specify the return air node for heat gain from lights, airflow windows, and refrigeration cases.
7. *Not working yet* Manages allocation of return air flow to multiple airloops serving the same zone.

Note that there some limitations in the multiple air loop implementation related to room air models, output reporting, etc. <to be listed here>

See #6272 for part 2 of this feature implementation.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@mjwitte do you want me to pretty much just focus on this PR's changes?  Or do I need to compare/contrast what's here with what's in the other larger PR?

mjwitte:
Sorry for the confusion.  Focus on this one.  I'm going to close the other one for now because it broke things, and rethink how to deal with ZoneEquipConfig.AirloopNum.  This one needs some new example files yet, and engineering ref, but it include all the expected IDD changes.

rraustad:
Conflicts now with the Single Duct Cp calculation changes.

mjwitte:
Actually most of the conflicts came in with #6223.  Working on them along with other fixes.

rraustad:
SingleDuct.cc looks messy. I would have to print this out to merge this correctly. Good luck.

mjwitte:
eio diffs in ChangeoverBypassVAV, ChangeoverBypassVAV_AirToAir, ChangeoverBypassVAV_AirToAirHeatPump, ChangeoverBypassVAV_MaxTemp, and HVACTemplate-5ZonePackagedVAV are all like this:
```
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:HeatAndCool:Reheat, ZONE 1 VAV SYSTEM, Design Size Maximum Flow per Zone Floor Area during Reheat [m3/s-m2], 1.50220E-002
```
This is actually a fix, because previously `Sys( SysNum ).ZoneFloorArea` was zero in `SingleDuct::SizeSys` and now it's not.

src/EnergyPlus/DirectAirManager.cc
```diff
@@ -625,7 +634,7 @@ namespace DirectAirManager {
 		MaxAirVolFlowRateUser = 0.0;
 		SizingDesRunThisZone = false;
 
-		if ( CurZoneEqNum > 0 ) {
+		if ( ( CurZoneEqNum > 0 ) && ( CurTermUnitSizingNum > 0 ) ) {
```
rraustad, line 653: So what happens here if CurTermUnitSizingNum = 0? Sizing gets skipped? I assume it would never be 0 and this is just a safety check. CurZoneEqNum is no longer used, does it still need to be checked now? Better safe than sorry I guess.

src/EnergyPlus/HVACInterfaceManager.cc
```diff
@@ -147,34 +146,49 @@ namespace HVACInterfaceManager {
...
-		// FLOW:
+
+		if ( ( CalledFrom == CalledFromAirSystemDemandSide ) && ( OutletNode == 0 ) ) {
+			// Air loop has no return path - only check mass flow and then set return inlet node mass flow to sum of demand side inlet nodes
+			AirLoopConvergence( AirLoopNum ).HVACMassFlowNotConverged( 1 ) = false;
+			AirLoopConvergence( AirLoopNum ).HVACHumRatNotConverged( 1 ) = true;
+			AirLoopConvergence( AirLoopNum ).HVACTempNotConverged( 1 ) = true;
+			AirLoopConvergence( AirLoopNum ).HVACEnergyNotConverged( 1 ) = true;
+			AirLoopConvergence( AirLoopNum ).HVACEnthalpyNotConverged( 1 ) = true;
+			AirLoopConvergence( AirLoopNum ).HVACPressureNotConverged( 1 ) = true;
```
rraustad: I understand only checking for mass flow rate below and setting the flag to true if out of tolerance, but these other flags are initialized to true. Won't that cause reporting for these variables if max iterations are exceeded? Why not set the other flags to false to show they are converged (even though they are not checked below) because it doesn't matter for this case?

mjwitte: I may have gotten that backwards.

src/EnergyPlus/DirectAirManager.cc
mjwitte: CureZoneEqNum is still used a few lines down for `CheckThisZoneForSizing( CurZoneEqNum, SizingDesRunThisZone );`

src/EnergyPlus/HVACSingleDuctInduc.cc
```diff
@@ -764,7 +795,11 @@ namespace HVACSingleDuctInduc {
 				}
 			} else {
 				CheckZoneSizing( IndUnit( IUNum ).UnitType, IndUnit( IUNum ).Name );
-				MaxTotAirVolFlowDes = max( TermUnitFinalZoneSizing( CurZoneEqNum ).DesCoolVolFlow, TermUnitFinalZoneSizing( CurZoneEqNum ).DesHeatVolFlow );
+				if ( CurTermUnitSizingNum > 0 ) {
```
rraustad, line 798: From what I can tell, CurTermUnitSizingNum > 0 when the user enters a name for the DesignSpecification:AirTerminal:Sizing object. They don't have to use that object correct (i.e., blank input)? If that object isn't used then shouldn't sizing be done the same way as before? So what was MaxTotAirVolFlowDes before these changes? It wasn't equal to 0 was it (line 801 below)?

mjwitte: No, CurTermUnitSizingNum is *always* used.  If there is no DesignSpecification:AirTerminal:Sizing object then the flows etc in TermUnitFinalZoneSizing will be the same as before.  (FinalZoneSizing plus any system sizing adjustments).  If there is a DesignSpecification:AirTerminal:Sizing object then they get scaled by that also.

mjwitte: Maybe the checks for CurTermUnitSizingNum don't need to be there, but early on in this development, there were some timing issues of when that got initialized and these were added to protect against trying to access TermUnitFinalZoneSizing(0).

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -1133,6 +1133,21 @@ namespace InternalHeatGains {
...
+					if ( !lAlphaFieldBlanks( 7 ) ) {
+						if ( LightsObjects( Item ).ZoneListActive ) {
+							ShowSevereError( RoutineName + CurrentModuleObject + "=\"" + Lights( Loop ).Name + "\": " + cAlphaFieldNames( 7 ) + " must be blank when using a ZoneList." );
+							ErrorsFound = true;
+						} else {
+							retNodeName = AlphaName( 7 );
+						}
+					}
+					if ( Lights( Loop ).ZonePtr > 0 ) {
+						Lights( Loop ).ReturnNodePtr = DataZoneEquipment::GetReturnAirNodeForZone( Zone( Lights( Loop ).ZonePtr ).Name, retNodeName );
```
rraustad, line 1148: So ReturnNodePtr will equal 0 if there is no return used by HVAC? How will the lights know NOT to put heat gain to return when there is no HVAC connection?

mjwitte: The lights won't know.  There still has to be a zone return node and then later (I forget where - in corrector?) there's a decision made whether to put gains on the return node, or if there is zero flow or it's a cycling fan, then put it in the zone heat balance.  This is still currently on a zone-level basis - I need to redo that part to sort them out node-by-node and decide which gains go where.

mjwitte:
Diffs in 5ZoneCAV_MaxTemp were traced to the plenum leakage calcs (setting leakage to zero produce no diffs).  Found some fun things like negative plenum outlet flow rates early in the first environment when some of the ADUs weren't sized yet and node maxmassflow was getting set to a negative value which flowed into the plenum leakage and plenum flow calcs.  Added protection against negative values and removed some superfluous leakage calcs, so there may now be small diffs in more files with ADU leakage.

src/EnergyPlus/HVACSingleDuctInduc.cc
rraustad: Then MaxTotAirVolFlowDes always = max( TermUnitFinalZoneSizing( blah blah blah... regardless if CurTermUnitSizingNum > 0. So this is just a check to prevent an array access crash when CurTermUnitSizingNum = 0. OK then. The variable name confused me since it's not a ptr to the design spec object (which I assumed), it's an index to a TU in a list that is the number of total zone inlets nodes long. As long as the TermUnitFinalZoneSizing array is in that same order then it's OK. I guess if these weren't in the same order there would be massive diff's on the dashboard.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4057,12 +4089,17 @@ namespace SimAirServingZones {
 
 				for ( ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) {
 					CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum );
+					int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
```
rraustad, line 4093: So this is still CurTermUnitSizingNum right? The name was changed to protect the innocent.

    AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZoneInSysIndex ) = TermUnitSizingNumsCool( ZoneInSysIndex );
    TermUnitSizingNumsCool( NumZonesCool ) = ZoneEquipConfig( CtrlZoneNum ).AirDistUnitCool( ZoneInNum ).TermUnitSizingIndex;

mjwitte: <sigh> Yes, it's a bit twisted, but it's the only way I could get it to work with the scattered getinputs.  A refactor of all that would simplify this, but what I have now is the ZoneEquipConfig.AirDistUnitCool (and Heat).TermUnitSizingIndex are all preassigned when ZoneEquipConfig is set up.  Then any terminal unit that's connected to that AirDistUnitCool/Heat will grab and use that one.  There is a unique TermUnitSizingIndex for every zone inlet node - so some which are only used by zone equipment will never be used.

rraustad: OK by me, for now. We can change the variable name later. I'm just trying to find issues before this drops in.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4760,18 +4807,24 @@ namespace SimAirServingZones {
 
 				for ( ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) { // loop over cooled zones
 					CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum );
-					SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinCoolMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesCoolMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+					int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
+					Real64 coolFlowRatio = TermUnitSizing( TermUnitSizingIndex ).SpecDesSensCoolingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesCoolSATRatio ;
```
rraustad, line 4812: Potential divide by 0 here and for heatFlowRatio below?

    N2, \field Cooling Design Supply Air Temperature Difference Ratio
      \note This ratio adjusts the supply air temperature difference used to calculate 
      \note the cooling design supply air flow rate for this terminal unit.
      \type real
      \units dimensionless
      \minimum 0.0

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4760,18 +4807,24 @@ namespace SimAirServingZones {
...
+					SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinCoolMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesCoolMassFlow * coolFlowRatio / ( 1.0 + TermUnitSizing( TermUnitSizingIndex ).InducRat );
 				} // end of loop over cooled zones
 
 				if ( NumZonesHeated > 0 ) { // if there are zones supplied with central hot air
 					for ( ZonesHeatedNum = 1; ZonesHeatedNum <= NumZonesHeated; ++ZonesHeatedNum ) { // loop over heated zones
 						CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).HeatCtrlZoneNums( ZonesHeatedNum );
-						SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinHeatMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesHeatMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+						int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitHeatSizingIndex( ZonesHeatedNum );
+						Real64 heatFlowRatio = TermUnitSizing( TermUnitSizingIndex ).SpecDesSensHeatingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesHeatSATRatio ;
```
rraustad, line 4820: divide by 0?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4760,18 +4807,24 @@ namespace SimAirServingZones {
...
+						SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinHeatMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesHeatMassFlow * heatFlowRatio / ( 1.0 + TermUnitSizing( TermUnitSizingIndex ).InducRat );
 					} // end of loop over heated zones
 				} else { // otherwise use cool supply zones
 					for ( ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) { //loop over cooled zones
 						CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum );
-						SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinHeatMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesHeatMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+						int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
+						Real64 heatFlowRatio = TermUnitSizing( TermUnitSizingIndex ).SpecDesSensHeatingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesHeatSATRatio ;
```
rraustad, line 4827: divide by 0?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5707,18 +5781,20 @@ namespace SimAirServingZones {
...
 					// save the system cooling supply air temp
 					FinalZoneSizing( CtrlZoneNum ).DesCoolCoilInTempTU = CalcSysSizing( AirLoopNum ).CoolSupTemp;
 					// save the system cooling supply air hum rat
 					FinalZoneSizing( CtrlZoneNum ).DesCoolCoilInHumRatTU = CalcSysSizing( AirLoopNum ).CoolSupHumRat;
 					if ( FinalZoneSizing( CtrlZoneNum ).DesCoolMassFlow <= 0.0 ) continue;
-					CalcSysSizing( AirLoopNum ).NonCoinCoolMassFlow += FinalZoneSizing( CtrlZoneNum ).DesCoolMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-					SysCoolRetTemp += FinalZoneSizing( CtrlZoneNum ).ZoneRetTempAtCoolPeak * FinalZoneSizing( CtrlZoneNum ).DesCoolMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-					SysCoolRetHumRat += FinalZoneSizing( CtrlZoneNum ).ZoneHumRatAtCoolPeak * FinalZoneSizing( CtrlZoneNum ).DesCoolMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+					Real64 coolMassFlow = FinalZoneSizing( CtrlZoneNum ).DesCoolMassFlow * TermUnitSizing( TermUnitSizingIndex ).SpecDesSensCoolingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesCoolSATRatio ;
```
rraustad, line 5791: divide by 0?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5759,18 +5835,20 @@ namespace SimAirServingZones {
...
 						// save the system heating supply air temp
 						FinalZoneSizing( CtrlZoneNum ).DesHeatCoilInTempTU = CalcSysSizing( AirLoopNum ).HeatSupTemp;
 						// save the system heating supply air hum rat
 						FinalZoneSizing( CtrlZoneNum ).DesHeatCoilInHumRatTU = CalcSysSizing( AirLoopNum ).HeatSupHumRat;
 						if ( FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow <= 0.0 ) continue;
-						CalcSysSizing( AirLoopNum ).NonCoinHeatMassFlow += FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-						SysHeatRetTemp += FinalZoneSizing( CtrlZoneNum ).ZoneRetTempAtHeatPeak * FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-						SysHeatRetHumRat += FinalZoneSizing( CtrlZoneNum ).ZoneHumRatAtHeatPeak * FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow / ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+						Real64 heatMassFlow = FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow * TermUnitSizing( TermUnitSizingIndex ).SpecDesSensHeatingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesHeatSATRatio ;
```
rraustad, line 5845: divide by 0?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5796,18 +5874,20 @@ namespace SimAirServingZones {
...
```
rraustad, line 5884: divide by 0?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -6081,42 +6165,55 @@ namespace SimAirServingZones {
...
-								TermUnitFinalZoneSizing( CtrlZoneNum ).DesHeatVolFlow = FinalZoneSizing( CtrlZoneNum ).DesHeatVolFlow * ZoneOARatio * ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-								TermUnitFinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow = FinalZoneSizing( CtrlZoneNum ).DesHeatMassFlow * ZoneOARatio * ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-								TermUnitFinalZoneSizing( CtrlZoneNum ).DesHeatLoad = FinalZoneSizing( CtrlZoneNum ).DesHeatLoad * ZoneOARatio * ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-								TermUnitFinalZoneSizing( CtrlZoneNum ).HeatFlowSeq = FinalZoneSizing( CtrlZoneNum ).HeatFlowSeq * ZoneOARatio * ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
-								TermUnitFinalZoneSizing( CtrlZoneNum ).HeatLoadSeq = FinalZoneSizing( CtrlZoneNum ).HeatLoadSeq * ZoneOARatio * ( 1.0 + TermUnitSizing( CtrlZoneNum ).InducRat );
+								// size on ventilation load
+								ZoneOARatio = FinalZoneSizing( CtrlZoneNum ).MinOA * thisTUSizing.SpecMinOAFrac / max( FinalZoneSizing( CtrlZoneNum ).DesHeatVolFlow, FinalZoneSizing( CtrlZoneNum ).MinOA );
```
rraustad, line 6201: divide by 0? how many times have we seen zone air flow = 0 due to no load.

rraustad: Never mind, MinOA must be > 0

src/EnergyPlus/SimAirServingZones.cc
mjwitte: Change this to \minimum> 0.0 and the heating one.  Will add divide-by-zero protects in the code.

src/EnergyPlus/SimAirServingZones.cc
mjwitte: Well, you could have a zone with zero minOA.

rraustad: No, the IF test protects this.

rraustad:
Wowza. I made it all the way through and tried hard to break the logic. It seems sound. Save a review of the results then this is ready after @mjwitte's last push (e.g., minimum>0, divide by 0 protection, etc.).

rraustad:
Those buttons shouldn't be right next to each other.

mjwitte:
@rraustad Thanks - that's (somewhat) reassuring.  Still have some other kinks to work out - splitting GetUnitarySystemData into two parts now so that DOATToUnitarySystem will work.

rraustad:
Not "(somewhat)" reassuring, very reassuring. You have 2 diff's and an award winning record on warnings. This is good for this release.

mjwitte:
But I have one failing integration test, DOAToUnitarySystem, fails in InitAirLoops with 
```
   ** Warning ** GetAirPathData: AirLoopHVAC="DOAS" has no Controllers.
   ** Severe  ** Node SPACE1-1 AIR TERMINAL MIXER PRIMARY INLET connects to no component
```
Because, unitary system input hasn't been gotten yet, because it waits until after airloops are set up because it might exist as both zone equipment and on an airloop.  I tried splitting out the part of get input that checks the connections to zone and/or airloop and call that from the init, but then the unit tests break all over the place because the input isn't complete.   I think it would work if getinput was split out to a factory that create on at a time, but it's risky with the many many calls to getunitarysystem scattered about.

rraustad:
Let me ponder that.

mjwitte:
But maybe if I rearrange GetUnitarySystemInputData and check for zone equipment first, then skip the airloop stuff . . .

rraustad:
Sure begs for ZoneHVAC and AirloopHVAC counterparts ... If there are zone AND airloop objects, and GetInput gets both, then there is a problem. UNLESS all the problems are moved to Init? Or air loop systems return after GetInput, or ...

rraustad:
You need a test file with AirloopHVAC and ZoneHVAC to test any changes. I have them at work but not here. Just ask.

rraustad:
So this is the catch.

    // Airloop inputs are filled after zone equipment has simulated. Wait for air loop equipment to simulate. Zone equipment will not simulate the first few times through.
    // This is only a problem is unitary systems are used as zone AND airloop equipment in the same input file.
    if ( SimUnitarySystemZoneEquipTestFlag ) {
        if ( present(ZoneEquipment) && FirstHVACIteration ) {
            return;
        } else {
            SimUnitarySystemZoneEquipTestFlag = false;
        }
    }
    // Obtains and Allocates unitary system related parameters from input file
    if ( GetInputFlag ) {
        // Get the unitary system input
        GetUnitarySystemInput();
        GetInputFlag = false;
    }

So this returns on zone equipment but not air loop unless there is no zone equipment. What is needed for zone equipment?

rraustad:
No, it returns on all zone equipment and the first air loop equipment gets the input or on NOT FirstHVAC all UnitarySystem equipment is gotten.

rraustad:
How about this?

    // Airloop inputs are filled after zone equipment has simulated. Wait for air loop equipment to simulate. Zone equipment will not simulate the first few times through.
    // This is only a problem is unitary systems are used as zone AND airloop equipment in the same input file.
    if ( SimUnitarySystemZoneEquipTestFlag ) {
        if ( present(ZoneEquipment) && FirstHVACIteration ) {
           GetStuffINeed();
            return;
        } else {
            SimUnitarySystemZoneEquipTestFlag = false;
        }
    }
    // Obtains and Allocates unitary system related parameters from input file
    if ( GetInputFlag ) {
        // Get the unitary system input
        GetUnitarySystemInput();
        GetInputFlag = false;
    }

mjwitte:
Something like that.  I'm trying something now that deletes that return thing and sets a flag if there's unitary system that can't find it's home.  Then try again next time.  We'll see in the morning if it was successful.

src/EnergyPlus/HVACInterfaceManager.cc
```diff
@@ -147,34 +146,49 @@ namespace HVACInterfaceManager {
...
+
+			int totDemandSideMassFlow = 0.0;
+			int totDemandSideMinAvail = 0.0;
+			int totDemanSideMaxAvail = 0.0;
```
mjwitte: @rraustad Notice anything funny about these three lines?  Couldn't figure out why totDemandSideMassFlow was always zero even though Node( demInNode).MassFlowRate was not . . . doh!

mjwitte:
Added two new example files: 
5ZoneAirCooledWithDOASAirLoop (demostrates two air loops serving the same zones and DesignSpecification:AirTerminal:Sizing and no return path) and DOASDXCOIL_wADPBFMethod_NoReturnPath (demonstrates no return path, results identical to DOASDXCOIL_wADPBFMethod)

At this point, I'm done with phase 1.  Still have some work to do on allocation return flow when more than one return node in the zone, need engineering ref changes, and implement the things noted at the top as *not working yet*.  If reviewers see fit to merge this, then will need a followup issue to do final clean up next week.

Note that 5ZoneAirCooled will show an err diff - unused construction warning should be gone.

mjwitte:
@Myoldmopar Just noticed that the epw file for 5ZoneAirCooled in the cmake list is Golden CO but the file is Chicago.  And I copied that to the new example file here.  Any reason that's Golden?  I can fix it here.  for 5ZoneAirCooled but that will probably incur diffs.

mjwitte:
And yeah, there aren't any unit tests yet, but these changes do touch some existing ones.  Will add more next week

rraustad:
Might this be happening with the DElight files?

mjwitte:
@rraustad @Myoldmopar Working on resolving conflicts here.

rraustad:
Diff's from MacOS look OK. This is hopeful.

rraustad:
@mjwitte can you take a quick look at this. I took the new example file DOASDXCOIL_wADPBFMethod_NoReturnPath and compared it to DOASDXCOIL_wADPBFMethod. The last several columns do not line up, but those shown below do line up and show diff's. Can you explain?

This is the last few columns in the spreadsheet that have diff's. Diff's are *NoReturn - original exam;ple file. I'll email the spreadsheet separately.

image

rraustad:
Spreadsheet sent, please acknowledge.

rraustad:
OK, now I'm a little confused. The coils operate the exact same way and the air flow rate is the same. Yet the heating coil outlet temp is different? There is no return in one case which means the system entering air temp is different. So this might be doing what it should be doing but I would think the coils would operate differently. @mjwitte might be able to explain.

Original:

    AirLoopHVAC,
      DOAS,                    !- Name
      ,                        !- Controller List Name
      DOAS Availability Managers,  !- Availability Manager List Name
      autosize,                !- Design Supply Air Flow Rate {m3/s}
      DOAS Branches,           !- Branch List Name
      DOAS Air Loop Inlet,     !- Supply Side Inlet Node Name
      DOAS Return Air Outlet,  !- Demand Side Outlet Node Name
      DOAS Supply Path Inlet,  !- Demand Side Inlet Node Names
      DOAS Heating Coil Outlet;!- Supply Side Outlet Node Names


NoReturnPath:

    AirLoopHVAC,
      DOAS,                    !- Name
      ,                        !- Controller List Name
      DOAS Availability Managers,  !- Availability Manager List Name
      autosize,                !- Design Supply Air Flow Rate {m3/s}
      DOAS Branches,           !- Branch List Name
      DOAS Outdoor Air Inlet,  !- Supply Side Inlet Node Name
      ,                        !- Demand Side Outlet Node Name
      DOAS Supply Path Inlet,  !- Demand Side Inlet Node Names
      DOAS Heating Coil Outlet;!- Supply Side Outlet Node Names

noretrunpathdiffs

rraustad:
And the coils are fighting each other, but that is the same as the original file. This may be due to humidity control but if there is no return in one case then shouldn't the coil operation change?

    SetpointManager:MultiZone:Humidity:Maximum,
      Maximum MZone HumRatio setpoint,  !- Name
      DOAS,                    !- HVAC Air Loop Name
      0.003,                   !- Minimum Setpoint Humidity Ratio {kgWater/kgDryAir}
      0.015,                   !- Maximum Setpoint Humidity Ratio {kgWater/kgDryAir}
      DOAS Dehumidifier Setpoint Nodes;  !- Setpoint Node or NodeList Name

    NodeList,
      DOAS Dehumidifier Setpoint Nodes,  !- Name
      DOAS Cooling Coil Outlet;!- Node 1 Name

rraustad:
It appears in the figures above that when there is air flow, the results are identical. That's good, but I still can't figure out why since the inlet air temp is different, right?

mjwitte:
This should be 100% OA so the return air gets relieved and has no impact.

rraustad:
That's obvious. Why doesn't the coil operation change?

mjwitte:
The node name has changed but the conditions at the first coil inlet should be identical. In both cases it should be OA conditions plus maybe fan heat ( I forget where the fan is).


rraustad:
Note in this figure, the system inlet node temp is identical when air flow > 0. So OK, the same inlet temp exists and coil performance is the same. It's a blow thru fan and this temp is before the fan.

image

rraustad:
So the heating coil outlet temp is not different when air flow > 0. Since there is no OA mixer, the system inlet temp = OAT now when air flow = 0. Consider this reviewed and accepted.

rraustad:
This is a little weird. The compiler says this is unused:

line 543 DataZoneEquipment.
ZoneEquipList( ControlledZoneNum ).EquipIndex = 0;

This is a glitch not a show stopper.

rraustad:
No, it's the header file, saying ZoneEquipConfig isn't used.

  extern Array1D< EquipConfiguration > ZoneEquipConfig;

rraustad (approved):
I agree. Put this in place for future revisions.

rraustad:
Reviewed and merged.

rraustad:
I don't know what issue this relates to nor the Pivotal ticket.

PR #6252, opened 2017-08-17
Add a new field of Per Person Ventilation Rate Mode to AirTerminal:SingleDuct:Uncontrolled and AirTerminal:SingleDuct:Mixer

lgu1234:
Pull request overview
---------------------
A new field was added to enhance both terminal units types of AirTerminal:SingleDuct:Uncontrolled and AirTerminal:SingleDuct:Mixer. The ventilation rate determined by the Outdoor Air Flow per Person input in DesignSpecification:OutdoorAir can be varied by choosing either CurrentOccupancy or DesignOccupancy in the Per Person Ventilation Rate Mode field.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
So why are we adding a new field to zone equipment instead of adding the new field to the DesignSpecification:OutdoorAir object? Then everyone would have access to that information instead of just TUs.

mjwitte:
@rraustad Well, both approaches were discussed.  We opted for this one because it follow the precedent already present in AirTerminal:DuaDuct:VAV:OutdoorAir and it avoids confusion at the system OA controller level.  This clearly applies to just this terminal unit.  A bit ugly, yeah, but safe.

rraustad:
OK, that makes sense. But thinking out loud, if one TU wanted to use the people schedule, doesn't that mean the OA controller would also use the schedule? I guess not but it's a little strange for the OA controller doing something different than the TU. Or if one TU wanted to use max occupancy, and another wanted to use the schedule, then what would the OA controller need to do? How would the OA controller know how much OA to bring in? Yes, this can get very messy. So worst case, the OA controller could ask each TU how much OA it needed. I guess then OK for now until we get more comfortable with how this plays out.

rraustad:
There are no diff's with this change and the warning was already fixed. This appears to do what is expected.

rraustad:
Reviewed and merged.

PR #6255, opened 2017-08-18
Fix from comma operator to addition operator

Myoldmopar:
I'm sure there will be diffs, let's see how many...


Nigusse:
I will review the diffs and test the defect files locally.

Nigusse (approved):
This is good to go.

Nigusse:
I did review selected example files with diffs but I did not see any thing that concerns me. I am wondering how the compiler treated such code and passed it with warning message only @Myoldmopar .

Myoldmopar:
It's valid C++.  The only common usage I see is in a for loop:

```
for ( int i = 0; i < some_i_max; i++, j++ )
{
}
```

See the i++, j++ at the end?  That's the same operation as this.  It will evaluate each comma delimited expression, and then calmly return the result of the last one.

In our case, it was valid code, but the compiler could tell that we never used the first term, so it just calmly warned us.

Thanks for testing this.  Looks like one large diff and a bunch of small diffs.  Which I would expect.  I'll merge this.

PR #6256, opened 2017-08-18
Fixup idd units

Myoldmopar:
I accidentally brought in a bad unit in the PCM work.  This fixes that.

But not only that.  I also wrote a script to validate the units fields in the IDD against the units specified in the IDD header.  It illuminated a bunch more.  Almost entirely just capitalization.  So I cleaned them up.

I added this script to the repo, but I haven't integrated it with CI yet.  It will be straightforward to do so.  Then we'll get meaningful warnings anytime someone (me) tries to put in an invalid unit type.

### DIFF
--- a/idd/Energy+.idd.in
+++ b/idd/Energy+.idd.in
@@ -7049,7 +7049,7 @@ MaterialProperty:PhaseChangeHysteresis,
   N2 , \field Liquid State Thermal Conductivity
        \note The thermal conductivity used by this material when the material is fully liquid
        \required-field
-       \units W/mK
+       \units W/m-K
        \type real
        \minimum> 0
   N3 , \field Liquid State Density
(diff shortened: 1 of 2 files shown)


rraustad:
Good to go. @mjwitte ?

mjwitte:
@rraustad Looks OK to me.

Myoldmopar:
Thanks @mjwitte and @rraustad.  Will be great to get this integrated into CI so this happens automatically.  I'll merge it now.

rraustad:
OK, this is done as soon as CI finishes. My suggestion would be to include other checks for:

 - all objects have \min-fields
 - all objects have \memo
 - every field has a \type
 - defaults are consistent with \type real or \type int (e.g., 0 vs 0.0)
 - all \type choice fields have a default
 - no required-fields after min-fields (warning?)
 - no \default for required-fields
 - specific order of info (e.g., first is \type, then \units, \default, \minimum, \maximum, \note, etc.) ?
 - \note has 100 character limit to string length ?
 - no \minimum > (must be \minimum>)
 - no \maximum < (must be \maximum<)
 - probably others

Myoldmopar:
It's actually already done (merged).  I needed this out of the way quickly with other PRs needing to go in.  

Yes, I agree, that script could be updated to do lots more parsing and validation, with very little effort, so that the IDD stays pristine until we have the JDD.

Issue #6013, opened 2017-02-14
Zone warmup load convergence check errors for small loads


EnergyArchmage:
Issue overview
--------------
The calculations used to check for zone warmup load convergence, for both heating and cooling, have a flaw when loads are less than MinLoad.

For heating there is:

MaxHeatLoadZone( ZoneNum ) = std::abs( max( MaxHeatLoadZone( ZoneNum ), MinLoad ) );
WarmupConvergenceValues( ZoneNum ).TestMaxHeatLoadValue = std::abs( ( MaxHeatLoadZone( ZoneNum ) - MaxHeatLoadPrevDay( ZoneNum ) ) / MaxHeatLoadZone( ZoneNum ) );

But the test value calculation is using a MaxHeatLoadPrevDay that has not gotten the same treatment. need to add in 
MaxHeatLoadPrevDay( ZoneNum ) = std::abs( max( MaxHeatLoadPrevDay( ZoneNum ), MinLoad ) );

for example cooling test load should be like this.
					
MaxCoolLoadZone( ZoneNum ) = std::abs( max( MaxCoolLoadZone( ZoneNum ), MinLoad ) ); 
MaxCoolLoadPrevDay( ZoneNum ) = std::abs( max( MaxCoolLoadPrevDay( ZoneNum ), MinLoad ) ); 
WarmupConvergenceValues( ZoneNum ).TestMaxCoolLoadValue = std::abs( ( MaxCoolLoadZone( ZoneNum ) - MaxCoolLoadPrevDay( ZoneNum ) ) / MaxCoolLoadZone( ZoneNum ) );

Contribution from TRANE. 

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte I have assigned this to myself andwill take a look at this one as this is something I have worked on in the past and should be able to figure out what @EnergyArchmage is saying in the overview.

RKStrand:
[...]

PR #6262, opened 2017-08-22
Improve Zone Warmup Convergence Test for Small Zone Loads

RKStrand:
Pull request overview
---------------------
This pull request addresses the issue noted in GitHub Issue #6013.  That noted that the max and min load values for heating and cooling for the previous day were not being compared to the MinLoad value as the current day value was.  This could cause potential convergence issues.  While there was not an input file to show this, the unit tests created for this file tested all of the different possibilities to make sure this was fixed.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #6013 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HeatBalanceManager.cc
+++ b/src/EnergyPlus/HeatBalanceManager.cc
@@ -5291,6 +5291,7 @@ namespace HeatBalanceManager {
 
 				if ( MaxHeatLoadZone( ZoneNum ) > 1.0e-4 ) { // make sure load big enough to divide
 					MaxHeatLoadZone( ZoneNum ) = std::abs( max( MaxHeatLoadZone( ZoneNum ), MinLoad ) );
+					MaxHeatLoadPrevDay( ZoneNum ) = std::abs( max( MaxHeatLoadPrevDay( ZoneNum ), MinLoad ) );
 					WarmupConvergenceValues( ZoneNum ).TestMaxHeatLoadValue = std::abs( ( MaxHeatLoadZone( ZoneNum ) - MaxHeatLoadPrevDay( ZoneNum ) ) / MaxHeatLoadZone( ZoneNum ) );
 					if ( WarmupConvergenceValues( ZoneNum ).TestMaxHeatLoadValue <= LoadsConvergTol ) {
 						WarmupConvergenceValues( ZoneNum ).PassFlag( 3 ) = 2;
@@ -5304,6 +5305,7 @@ namespace HeatBalanceManager {
 
 				if ( MaxCoolLoadZone( ZoneNum ) > 1.0e-4 ) {
 					MaxCoolLoadZone( ZoneNum ) = std::abs( max( MaxCoolLoadZone( ZoneNum ), MinLoad ) );
+					MaxCoolLoadPrevDay( ZoneNum ) = std::abs( max( MaxCoolLoadPrevDay( ZoneNum ), MinLoad ) );
 					WarmupConvergenceValues( ZoneNum ).TestMaxCoolLoadValue = std::abs( ( MaxCoolLoadZone( ZoneNum ) - MaxCoolLoadPrevDay( ZoneNum ) ) / MaxCoolLoadZone( ZoneNum ) );
 					if ( WarmupConvergenceValues( ZoneNum ).TestMaxCoolLoadValue <= LoadsConvergTol ) {
 						WarmupConvergenceValues( ZoneNum ).PassFlag( 4 ) = 2;
--- a/tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
+++ b/tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
@@ -621,4 +621,87 @@ namespace EnergyPlus {
 		EXPECT_EQ( 0.4, Material( 1 ).InitMoisture ); // reset from 0.45 to 0.4 during get input
 
 	}
+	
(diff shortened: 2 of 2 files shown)


src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5291,6 +5291,7 @@ namespace HeatBalanceManager {
 
 				if ( MaxHeatLoadZone( ZoneNum ) > 1.0e-4 ) { // make sure load big enough to divide
 					MaxHeatLoadZone( ZoneNum ) = std::abs( max( MaxHeatLoadZone( ZoneNum ), MinLoad ) );
+					MaxHeatLoadPrevDay( ZoneNum ) = std::abs( max( MaxHeatLoadPrevDay( ZoneNum ), MinLoad ) );
```
rraustad, line 5294: But this value is set at the bottom of the function and includes the comparison to MinLoad. So what does this actually accomplish?

    // Transfer current daily max and min loads and temperatures to the
    // variables containing the last day's values
    MaxHeatLoadPrevDay( ZoneNum ) = MaxHeatLoadZone( ZoneNum );
    MaxCoolLoadPrevDay( ZoneNum ) = MaxCoolLoadZone( ZoneNum );
    MaxTempPrevDay( ZoneNum ) = MaxTempZone( ZoneNum );
    MinTempPrevDay( ZoneNum ) = MinTempZone( ZoneNum );

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
```diff
@@ -621,4 +621,87 @@ namespace EnergyPlus {
...
+		MaxCoolLoadPrevDay( 1 ) = 90.0;
+		CheckWarmupConvergence();
+		EXPECT_EQ( WarmupConvergenceValues( 1 ).PassFlag( 3 ), 2 );
+		EXPECT_EQ( WarmupConvergenceValues( 1 ).PassFlag( 4 ), 2 );
+		EXPECT_NEAR( WarmupConvergenceValues( 1 ).TestMaxHeatLoadValue, 0.0, 0.0001 );
+		EXPECT_NEAR( WarmupConvergenceValues( 1 ).TestMaxCoolLoadValue, 0.0, 0.0001 );
+		
+		// Test 2: Max Previous Day both less than MinLoad
+		MaxHeatLoadZone( 1 ) = 100.5;
+		MaxHeatLoadPrevDay( 1 ) = 90.0;
```
rraustad, line 663: How can MaxHeatLoadPrevDay = 90 when MinLoad = 100 ?

rraustad: To be clear, the first time through MaxHeatLoadPrevDay = 0, if MaxHeatLoadZone = 90, then MaxHeatLoadZone is set to 100 and then MaxHeatLoadPrevDay = MaxHeatLoadZone. The next time through, MaxHeatLoadPrevDay = 100 and doesn't need to be reevaluated. This function has to be called at least twice so again, what does the change to code accomplish?

src/EnergyPlus/HeatBalanceManager.cc
RKStrand: @rraustad Well, that's a good point and what you are saying makes sense.

@EnergyArchmage Based on what Rich is saying, is there really a need for this change as you outlined in the GitHub issue?  Is there some other scenario where the previous day's values are not being subjected to the MinLoad comparison already?

EnergyArchmage: It needs to be done before calculating the TestMaxHeatLoadValue test value.  as it is now the the difference can be calculated with inconsistent application of MinLoad in the two terms in the next line.  The reporting for warmup convergence can be wrong even though as Rich says the actual values are okay.

rraustad: @EnergyArchmage that doesn't make sense to me at this point without knowing more, but I don't see that this hurts anything other than runtime. I know you are trying to improve execution speed so I'll take your word for it at this point.

rraustad: @EnergyArchmage would it help to initialize previous day load variables to 100 (instead of this new line of code?) ? Were you seeing this on the first pass through, or each time through?

EnergyArchmage: I was seeing spurious "Fail" for convergence with zone loads between 0 and 100.  and extra warm up days because of it.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
RKStrand: @rraustad The unit tests were set-up to test that the code additions are actually going to do what they are designed to do--make sure that a previous day load of less than 100 gets reset to 100 for testing convergence.  I'm not sure how it could be happening that the previous load comes in less than 100 as you pointed out.  Since Brent was seeing this happen and suggested this correction, I went with his recommendation.

rraustad:
Reviewed:
- logic of additional math (no sure it's needed but doesn't hurt anything)
- unit test
- CI results all green (5ZoneAirCooledWithDOASAirLoop on the linux integration coverage tests is currently failing in develop)

Merged.

PR #6263, opened 2017-08-22
Use local nodes for air flow network calculation

xuanluo113:
Pull request overview
---------------------
1. Use local nodes for air flow network calculation
2. Tests and docs for local environmental data for urban context

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


xuanluo113:
@Myoldmopar 
Edwin, we added the feature back to be reviewed. The ticket is also for resolving the left issues for the Urban PR. 
@jasondegraw 
Jason, I don't know if you have time to review now but I was hoping to get some suggestions from you. The committed parts were mostly the original implementation, with some of your comments addressed - such as using the relative wind degree and making the local node interchangeable with a regular external node. 
Your question before on whether if it's the best logic was exactly where I found I may need help with for this part. I hesitated on the implementation to check whether a node is a regular external node or a local one in code. However, adding the link from surface - AirflowNetwork:MultiZone:Surface to local nodes - OutdoorAir:Node was kind of hard to do, so is adding new node types. I wonder if you have more thoughts on this version.

jasondegraw:
@XuanLuoLBL I'll try to get you review today, there are a few things that'll need some work. One other thing to do would be to add some unit tests that specifically test the new feature, I don't recall seeing much of that in the PR that got merged.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -1209,9 +1220,37 @@ namespace AirflowNetworkBalanceManager {
 			// Wind coefficient == Surface-Average does not need inputs of external nodes
 			CurrentModuleObject = "AirflowNetwork:MultiZone:ExternalNode";
 			AirflowNetworkNumOfExtNode = GetNumObjectsFound( CurrentModuleObject );
+			if ( AnyLocalEnvironmentsInModel ) {
+ 				CurrentModuleObject = "OutdoorAir:Node";
+ 				AirflowNetworkNumOfOutAirNode = GetNumObjectsFound( CurrentModuleObject );
+ 				AirflowNetworkNumOfExtNode += AirflowNetworkNumOfOutAirNode;
```
jasondegraw: I think that this calculation should go up at the top of the function where the other counts are computed.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -1209,9 +1220,37 @@ namespace AirflowNetworkBalanceManager {
...
+ 			}
 			if ( AirflowNetworkNumOfExtNode > 0 ) {
 				MultizoneExternalNodeData.allocate( AirflowNetworkNumOfExtNode );
-				for ( i = 1; i <= AirflowNetworkNumOfExtNode; ++i ) {
+				if ( AnyLocalEnvironmentsInModel ){
+ 					CurrentModuleObject = "OutdoorAir:Node";
+ 					for ( i = AirflowNetworkNumOfExtNode - AirflowNetworkNumOfOutAirNode + 1; i <= AirflowNetworkNumOfExtNode; ++i ) {
+ 						GetObjectItem( CurrentModuleObject, i - ( AirflowNetworkNumOfExtNode - AirflowNetworkNumOfOutAirNode ), Alphas, NumAlphas, Numbers, NumNumbers, IOStatus, lNumericBlanks, lAlphaBlanks, cAlphaFields, cNumericFields );
```
jasondegraw: I'm generally supportive of the idea here, where the external nodes are generated as needed for the outdoor air nodes. However, I'm not sure that putting this here is the best choice. 

@Myoldmopar @mjwitte @lgu1234 Any thoughts or feedback?

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4573,15 +4629,42 @@ namespace AirflowNetworkBalanceManager {
...
-						Vref = WindSpeedAt( MultizoneExternalNodeData( i ).height );
-						AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).curve,
-							Vref, AirflowNetworkNodeData( n ).NodeHeight, MultizoneExternalNodeData( i ).azimuth,
-							MultizoneExternalNodeData( i ).symmetricCurve, MultizoneExternalNodeData( i ).useRelativeAngle);
-						//AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).CPVNum, Vref, AirflowNetworkNodeData( n ).NodeHeight );
-					}
 					AirflowNetworkNodeSimu( n ).TZ = OutDryBulbTempAt( AirflowNetworkNodeData( n ).NodeHeight );
 					AirflowNetworkNodeSimu( n ).WZ = OutHumRat;
+					if ( i <= AirflowNetworkNumOfExtNode ) {
+						if ( MultizoneExternalNodeData( i ).OutAirNodeNum == 0 ) {
```
jasondegraw, line 4681: This seems like the right thing to do.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4573,15 +4629,42 @@ namespace AirflowNetworkBalanceManager {
...
+							LocalAzimuth = MultizoneExternalNodeData( i ).azimuth;
+							AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).curve,
+								MultizoneExternalNodeData(i).symmetricCurve, MultizoneExternalNodeData(i).useRelativeAngle,
+								LocalAzimuth, LocalWindSpeed, WindDir, LocalDryBulb, OutHumRat );
+						}
+						else {
+							// If and outdoor air node object is defined as the External Node Name in AirflowNetwork:MultiZone:Surface,
+							// the node object requires to define the Wind Pressure Coefficient Curve Name.
+							NodeNum = MultizoneExternalNodeData( i ).OutAirNodeNum;
+							if ( Node( NodeNum ).WindCoefCurveNum == 0 ) {
```
jasondegraw: This check looks like something that should be checked as the input is read in, rather than here. There are going to potentially be interactions with the surface average wind pressure coefficient calculation, though, so care will be needed.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4573,15 +4629,42 @@ namespace AirflowNetworkBalanceManager {
...
+								ShowSevereError("GetAirflowNetworkInput: AIRFLOWNETWORK:DISTRIBUTION:NODE: Invalid external node at " + AirflowNetworkNodeData( n ).Name + ": No wind coefficient is defined for airnode when used in the airflow network." );
```
jasondegraw: This error message should be revised so it is clearer that there's an issue with the wind pressure coefficient curve.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4573,15 +4629,42 @@ namespace AirflowNetworkBalanceManager {
...
+								ErrorsFound = true;								
+							}
+							else {
+								LocalWindSpeed = WindSpeedAtNode( NodeNum );
```
jasondegraw: I get what you're trying to do here, but if I understand the logic here, NodeNum is nonzero. In that case, the checks that these individual functions do are redundant. It'd be better just to do the direct assignment for all of these, e.g. `LocalWindSpeed = Node( NodeNum ).OutAirWindSpeed;`

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -5227,20 +5310,23 @@ namespace AirflowNetworkBalanceManager {
...
+	CalcWindPressure(
+		int const curve, // Curve index, change this to pointer after curve refactor
+		bool const symmetricCurve, // True if the curve is symmetric (0 to 180)
+		bool const relativeAngle, // True if the Cp curve angle is measured relative to the surface
+		Real64 const azimuth, // Azimuthal angle of surface
+		Real64 const windSpeed, // Wind velocity
+		Real64 const windDir, // Wind direction
+		Real64 const dryBulbTemp, // Air node dry bulb temperature
+		Real64 const humRat // Air node humidity ratio
+	)
```
jasondegraw, line 5359: This is good, eliminating the use of the globals inside the function makes things a bit more clear.

src/EnergyPlus/AirflowNetworkBalanceManager.hh
```diff
@@ -249,12 +249,14 @@ namespace AirflowNetworkBalanceManager {
...
-        bool const symmetricCurve, // True if the curve is symmetric (0 to 180)
-        bool const relativeAngle // True if the Cp curve angle is measured relative to the surface
+		int const curve, // Curve index, change this to pointer after curve refactor
+		bool const symmetricCurve, // True if the curve is symmetric (0 to 180)
+		bool const relativeAngle, // True if the Cp curve angle is measured relative to the surface
+		Real64 const azimuth, // Azimuthal angle of surface
+		Real64 const windSpeed, // Wind velocity
+		Real64 const windDir, // Wind direction
+		Real64 const dryBulbTemp, // Air node dry bulb temperature
+		Real64 const humRat // Air node humidity ratio
```
jasondegraw, line 259: Is there an indentation issue here? Looks like maybe two tabs?

src/EnergyPlus/DataEnvironment.cc
```diff
@@ -588,6 +590,161 @@ namespace DataEnvironment {
 
 	}
 
+	Real64
+	DryBulbTempAtNode( int NodeNum )  // OutdoorAir:Node Number
```
jasondegraw: This function and the others added in this file should be removed and direct assignment should be used.

src/EnergyPlus/DataLoopNode.hh
```diff
@@ -173,6 +173,10 @@ namespace DataLoopNode {
 		Real64 GenContamSetPoint; // {ppm}
 		bool SPMNodeWetBulbRepReq; // Set to true when node has SPM which follows wetbulb
 
+		int  WindCoefCurveNum; // Wind coefficient Curve ID, replace with pointer after curve refactor
```
jasondegraw: Does the node really need to know the pressure coefficient curve?  Maybe it does, but I have concerns that the logic that checks for this elsewhere is going to get tripped up by the automatic curve generation that takes place elsewhere.

tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
```diff
@@ -7940,6 +7961,685 @@ namespace EnergyPlus {
 		EXPECT_NEAR( cp105S*0.5*118.41123742118911, DataAirflowNetwork::AirflowNetworkNodeSimu( 5 ).PZ, 1e-13 );
 	}
 
+	TEST_F( EnergyPlusFixture, TestExternalNodesWithLocalAirNode ) {
```
jasondegraw, line 7963: Looks like this unit test isn't completely implemented yet, but this is definitely what we want to see.

jasondegraw (changes requested):
This is much better than before. The unit test needs to be more fully fleshed out and the added functions for local conditions (that just check if the nodenum is nonzero) should be removed. Also, I'm not sure that the input is handled in a way that is "standard", hopefully others will weigh in on that. The only remaining question I have is on compatibility with the SurfaceAverageCalculation setting in AFNSimulationControl. Do you want compatibility with automatically calculated Cp curves? It may be that the safest thing would be to disallow use of this feature with that option, but I do believe it will have to be addressed somehow.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
lgu1234: @XuanLuoLBL It may write more compact by using object names, instead of reassigning CurrentModuleObject:

AirflowNetworkNumOfExtNode = GetNumObjectsFound("AirflowNetwork:MultiZone:ExternalNode";);  
if ( AnyLocalEnvironmentsInModel )   {
				AirflowNetworkNumOfOutAirNode = GetNumObjectsFound( "OutdoorAir:Node");  
  				AirflowNetworkNumOfExtNode += AirflowNetworkNumOfOutAirNode;  
}

src/EnergyPlus/AirflowNetworkBalanceManager.cc
lgu1234: @XuanLuoLBL I think it may be better to place this section after reading CurrentModuleObject = "AirflowNetwork:MultiZone:ExternalNode"; Here is an order I would like to see

Read AirflowNetwork:MultiZone:ExternalNode to keep original order
Add a new section to read "OutdoorAir:Node"

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4230,16 +4271,20 @@ namespace AirflowNetworkBalanceManager {
 						if ( Contaminant.CO2Simulation ) AirflowNetworkNodeSimu( i ).CO2Z = OutdoorCO2;
 						if ( Contaminant.GenericContamSimulation ) AirflowNetworkNodeSimu( i ).GCZ = OutdoorGC;
 					}
+					//if ( AirflowNetworkNodeData( i ).OutAirNodeNum > 0 ) {				
+ 				//		AirflowNetworkNodeSimu( i ).TZ = DryBulbTempAtNode( AirflowNetworkNodeData( i ).OutAirNodeNum );
+ 				//		AirflowNetworkNodeSimu( i ).WZ = HumRatAtNode( AirflowNetworkNodeData( i ).OutAirNodeNum );
+ 				//	}
```
lgu1234: @XuanLuoLBL Remove commented lines.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -4573,15 +4629,42 @@ namespace AirflowNetworkBalanceManager {
...
+					if ( i <= AirflowNetworkNumOfExtNode ) {
+						if ( MultizoneExternalNodeData( i ).OutAirNodeNum == 0 ) {
+							LocalWindSpeed = WindSpeedAt( MultizoneExternalNodeData( i ).height );
+							LocalDryBulb = OutDryBulbTempAt( AirflowNetworkNodeData( n ).NodeHeight );
+							LocalAzimuth = MultizoneExternalNodeData( i ).azimuth;
+							AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).curve,
+								MultizoneExternalNodeData(i).symmetricCurve, MultizoneExternalNodeData(i).useRelativeAngle,
+								LocalAzimuth, LocalWindSpeed, WindDir, LocalDryBulb, OutHumRat );
+						}
+						else {
```
lgu1234: @XuanLuoLBL In order to be consistent with other code, you may move "else" after "}, such as:
if (  ) {
....
} else { // single line
....
}

src/EnergyPlus/DataEnvironment.cc
```diff
@@ -588,6 +590,161 @@ namespace DataEnvironment {
...
+		Real64 LocalDryBumpTemp; // Return result for function (m/s)
+
+		// Locals
+		// FUNCTION ARGUMENT DEFINITIONS:
+
+		if ( NodeNum == 0 ) {
+			LocalDryBumpTemp = OutDryBulbTemp;
+		}
+		else {
+			LocalDryBumpTemp = Node( NodeNum ).OutAirDryBulb;
```
lgu1234: @XuanLuoLBL Please  move "else" up a line. See the above comment.

src/EnergyPlus/DataEnvironment.cc
```diff
@@ -588,6 +590,161 @@ namespace DataEnvironment {
...
+
+	Real64
+	WetBulbTempAtNode( int NodeNum )  // OutdoorAir:Node Number
+	{
+
+		// FUNCTION INFORMATION:
+		//       AUTHOR         X Luo
+		//       DATE WRITTEN   August 2017
+		//       MODIFIED       na
+		//       RE-ENGINEERED  na
```
lgu1234: @XuanLuoLBL Remove comment lines with "NA".

src/EnergyPlus/DataEnvironment.cc
```diff
@@ -588,6 +590,161 @@ namespace DataEnvironment {
...
+		// Return value
+		Real64 LocalWetBumpTemp; // Return result for function (m/s)
+
+		// Locals
+		// FUNCTION ARGUMENT DEFINITIONS:
+
+		if ( NodeNum == 0 ) {
+			LocalWetBumpTemp = OutWetBulbTemp;
+		}
+		else {
```
lgu1234: @XuanLuoLBL See above of "else"

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: @jasondegraw The original calculation of the AirflowNetworkNumOfExtNode was actually here. The top of the function "sets the maximum numbers of input fields", and I don't really get what that does. Do I need to add to that?

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: @jasondegraw @lgu1234 I put it here as the original allocation of the external nodes was here, and I was hoping to allocate more (1) to the same structure, (2) under the same situation - AirflowNetworkSimu.iWPCCntr == iWPCCntr_Input. 

@lgu1234 I switched the order.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: This is only required in "OutdoorAir:Node" when linked to a "AirflowNetwork:MultiZone:Surface" and calculated in the AirflowNetwork. I removed the check to where reads"AirflowNetwork:MultiZone:Surface", which takes a ExternalNode field.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: Fixed.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: Removed those redundant functions.

src/EnergyPlus/AirflowNetworkBalanceManager.hh
xuanluo113: I think there supposed be two tabs here, as the other functions do in this header file.

src/EnergyPlus/DataEnvironment.cc
xuanluo113: Done.

src/EnergyPlus/DataLoopNode.hh
xuanluo113: You're right. It probably does not. I changed the logic a bit to read this in the structure MultiZoneExternalNode only and added some checks.

tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
xuanluo113: I modified it a bit. Any hint on what else is missing? Thanks.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: Fixed this.

xuanluo113:
@jasondegraw @lgu1234 
I addressed your comments at second last commits. I added another commit trying to make local nodes used if defined in several subroutines. Though I have two questions about the linkage which I'm not that familiar with and took a lot of time. I was hoping to get some help from you - Marked in the changed file.

@jasondegraw For your concern regarding the compatibility with the SurfaceAverageCalculation setting in AFNSimulationControl, the outdoor air nodes are not used in these cases, as no link is defined from a surface to a node. I made sure those are eliminated if defined in these cases.

@Myoldmopar Edwin, I'm soooo sorry that there seems to be one "\object-list" line I have to add in the idd, and it's for the alternative link to the air node in AirflowNetwork:Multizone:Surface. I hope that's fine.

tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
jasondegraw: Unless I've missed it, the unit test doesn't appear to test the creation of the outdoor air nodes. That's really what I'd like to see, even if it is only one.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
jasondegraw: No, I agree with @lgu1234 that doing away with the reassigning of CurrentModuleObject makes it easier to understand. Let's just do that.

jasondegraw:
@XuanLuoLBL Pretty much everything in my review has been addressed except the unit test and the SurfaceAverageCalculation compatibility. It'd be nice to see that the use of the outdoor air nodes in that unit test. On SurfaceAverageCalculation, could you point me to where that is being handled (either here on GitHub or via email)? I'm just not looking in the right places.

lgu1234:
@XuanLuoLBL Please give me a call for any linkage issues today.

xuanluo113:
@ALL who might be reviewing this follow up PR. I've addressed Jason and Lixing's comments on the airflow network part, added all 6 unit tests (for surface heat balance, zone heat balance, window heat balance, surface LWR, local air node in the network, and external shading import) and the documentation to the 3 new feature. 

The failed regression test was triggered by the eso big diffs. This is because I added the LWR from surrounding surfaces to the existing report variable - "Surface Outside Face Net Thermal Radiation Heat Gain Rate [W]"
`QdotRadOutRep( SurfNum ) = Surface( SurfNum ).Area * HExtSurf_fac + Surface( SurfNum ).Area * QRadLWOutSrdSurfsRep;`
, which I should have and didn't do at the last PR.

And I really can't figure out the unit test failure, which passes fine on my local branch. And if I edit the idf as the error message says here it won't pass on my machine with the error:
`  Actual: "   ** Warning ** GetShadowingInput: DetailedSkyDiffuseModeling is chosen but not needed as either the shading transmittance for shading devices does not change throughout the year\r\n   **   ~~~   **  or MinimalShadowing has been chosen.\r\n   **   ~~~   ** Simulation should be set to use SimpleSkyDiffuseModeling, but is left at Detailed for simulation.\r\n   **   ~~~   ** Choose SimpleSkyDiffuseModeling in the ShadowCalculation object to reduce computation time.\r\n"
Expected: expected_string
Which is: ""`
Any idea on this?

jasondegraw:
@XuanLuoLBL Not to be contrary, but I am still of the belief that SurfaceAverageCalculation needs to be disabled with this feature, which is why I haven't flipped my review status over to approved. Maybe you've addressed it and I just can't see it, if that is the case please point the code out to me.

xuanluo113:
@jasondegraw I guess I've been misunderstanding it. I was thinking there are no global flag or simulation settings to use the local air node in the airflow network. In the original design, only when "INPUT" is chosen, the external nodes are read and stored in the structure. So I did the same thing for the alternative local air node. So only when "INPUT" instead of "SurfaceAverageCalculation" && local node defined and linked to the external surface, the local nodes are read and used.
```
// *** Read AirflowNetwork external node
if ( AirflowNetworkSimu.iWPCCntr == iWPCCntr_Input ) {
    // Wind coefficient == Surface-Average does not need inputs of external nodes
    AirflowNetworkNumOfExtNode = GetNumObjectsFound( "AirflowNetwork:MultiZone:ExternalNode" );
    ...
```
If SurfaceAverageCalculation is chosen, this subroutine will be bypassed and no nodes will actually be read. And the calculation wouldn't consider it either. So I'm not very clear how to disable the SurfaceAverageCalculation choice here. Is that what you're concerning? Probably I should explain this in the documentation.

jasondegraw:
@XuanLuoLBL That is what happens when the input file is read, but later on external nodes and curves are generated for SurfaceAverageCalculation. Those generated nodes will use "regular" outside conditions in spite of user input to the contrary. With the present node/curve generation code, the best thing to do right now is to issue a fatal error if this feature is used with SurfaceAverageCalculation. I'll defer to @lgu1234, @mjwitte, and @Myoldmopar on the best way to do that, but I do believe it is necessary.

lgu1234:
@XuanLuoLBL I agree with @jasondegraw opinion. You may issue a fatal error with combination of SurfaceAverageCalculation and your local node for the time being before we find a good way to handle this type of scenario.

xuanluo113:
@jasondegraw @lgu1234 Probably I misunderstood the whole thing, but I don't think extra external node or curves are generated when we used SurfaceAverageCalculation. The external nodes are allocated and generated when **AirflowNetworkSimu.iWPCCntr == iWPCCntr_Input**. On the other hand, the NodeData itself does not store any curve data - if the node has extra coeff curve input it's only stored in the MultizoneExternalNode structure. 
```
// *** Read AirflowNetwork external node
 if ( AirflowNetworkSimu.iWPCCntr == iWPCCntr_Input ) {
   // Wind coefficient == Surface-Average does not need inputs of external node
    if ( AirflowNetworkNumOfExtNode > 0 ) {
        MultizoneExternalNodeData.allocate( AirflowNetworkNumOfExtNode ); 				
        CurrentModuleObject = "AirflowNetwork:MultiZone:ExternalNode";
        for ( i = 1; i <= AirflowNetworkNumOfExtNode - AirflowNetworkNumOfOutAirNode; ++i ) {
            GetObjectItem( CurrentModuleObject, i, Alphas, NumAlphas, Numbers,...
...
            MultizoneExternalNodeData( i ).ExtNum = AirflowNetworkNumOfZones + i; // External node number
           MultizoneExternalNodeData( i ).curve = CurveManager::GetCurveIndex( Alphas( 2 ) ); // Wind pressure curve
```
if SurfaceAverageCalculation is used
```
// Calculate CP values
if ( SameString( AirflowNetworkSimu.WPCCntr, "SurfaceAverageCalculation" ) ) {
    CalcWindPressureCoeffs();
```
The external nodes are allocated in CalcWindPressureCoeffs():
```
MultizoneExternalNodeData.allocate( AirflowNetworkNumOfExtSurfaces );
AirflowNetworkNumOfExtNode = AirflowNetworkNumOfExtSurfaces;
```
, and the curve is generated here. So I don't know how the user's extra input of external nodes will affect this.

xuanluo113:
@jasondegraw @lgu1234 
The question I have is that it seems all Node type data ( Node( i ) )would be considered as either distribution or zone node, and used in different system calculation - in my understanding by reading the code. So the `ValidateDistributionSystem()` function validates the inputs and connections of the distribution system. And before the validation, it eliminates all Node(i) that is not the node in the distribution system.
```
// Eliminate node not related to AirLoopHVAC
// Eliminate nodes with fluidtype = water
// Eliminate local external air node for network
```
So I just eliminate the extra local external air node here - This has to be done whichever method the users define - SurfaceAverage or Input.
I don't know if the handling is appropriate here.

xuanluo113:
@jasondegraw @lgu1234 
And more importantly, the OutdoorAir:Node can also be used in SurfaceProperty:LocalEnvironment and ZoneProperty:LocalEnvironment, to overwrite the environmental data in surface and zone heat balance calculation - and those seem to be completely separate processes. And we can actually use the Nodes when on airflownetwork is defined. I was thinking, if we don't allow the definition of any local node when using SurfaceAverageCalculation, shall we block the use of these properties as well?

jasondegraw:
@XuanLuoLBL Let me try again. Let's say that the input has **LocalAirNodeA** connected to **MultizoneSurfaceB** with **SurfaceAverageCalculation** specified. I would infer that the user's intent was for **MultizoneSurfaceB** to be subject to local conditions. `CalcWindPressureCoeffs` disregards this and generates **ExternalNodeC**  and attaches it to **MultizoneSurfaceB,** so the problem crops up when this code in "AirflowNetworkBalanceManager.cc" is executed:

```
if ( MultizoneExternalNodeData( i ).OutAirNodeNum == 0 ) {
	LocalWindSpeed = WindSpeedAt( MultizoneExternalNodeData( i ).height );
	LocalDryBulb = OutDryBulbTempAt( AirflowNetworkNodeData( n ).NodeHeight );
	LocalAzimuth = MultizoneExternalNodeData( i ).azimuth;
	AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).curve,
	MultizoneExternalNodeData(i).symmetricCurve, MultizoneExternalNodeData(i).useRelativeAngle,
		LocalAzimuth, LocalWindSpeed, WindDir, LocalDryBulb, OutHumRat );
} else {
	// If and outdoor air node object is defined as the External Node Name in AirflowNetwork:MultiZone:Surface,
	// the node object requires to define the Wind Pressure Coefficient Curve Name.
	NodeNum = MultizoneExternalNodeData( i ).OutAirNodeNum;
	LocalWindSpeed = Node ( ( NodeNum ) ).OutAirWindSpeed;
	LocalWindDir = Node ( ( NodeNum ) ).OutAirWindDir;
	LocalHumRat = Node ( ( NodeNum ) ).HumRat;
	LocalDryBulb = Node ( ( NodeNum ) ).OutAirDryBulb;
	LocalAzimuth = MultizoneExternalNodeData( i ).azimuth;
	AirflowNetworkNodeSimu( n ).PZ = CalcWindPressure( MultizoneExternalNodeData( i ).curve, 
		MultizoneExternalNodeData( i ).symmetricCurve, MultizoneExternalNodeData( i ).useRelativeAngle, 
		LocalAzimuth, LocalWindSpeed, LocalWindDir, LocalDryBulb, LocalHumRat );
	AirflowNetworkNodeSimu( n ).TZ = LocalDryBulb;
	AirflowNetworkNodeSimu( n ).WZ = LocalHumRat;
}
```
**ExternalNodeC** has no idea that **LocalAirNodeA** exists, the `if` branch gets taken, and non-local conditions are used. We could try to modify `CalcWindPressureCoeffs` so that it properly connects things up, but that code is already complicated enough and it'll take too much time to do that now. This exact situation occurred last release (it was me), that didn't work out well (I broke single sided natural ventilation). Modifying that function is not easy, we do need to fix that, but now is not the time.

If someone is going to go through the trouble of setting up local air nodes, I don't see it as a big deal to ask them to specify the wind pressure coefficient curves (especially since one curve can be used for all surfaces).

XuanLuoLBL:
@jasondegraw So what I should not allow is the input of LocalAirNodeA when connecting to MultizoneSurfaceB, in with SurfaceAverageCalculation specified? The connection is defined in:
```
AirflowNetwork:MultiZone:Surface,
...
  A3, \field External Node Name
      \type object-list
      \object-list ExternalNodeNames
      \object-list OutdoorAirNodeNames
      \note Used if Wind Pressure Coefficient Type = Input in the AirflowNetwork:SimulationControl object,
      \note otherwise this field may be left blank.
```
And in the code implementation, this input is just ignored when SurfaceAverageCalculation - no warnings or fatal errors for the original external nodes.
```
CurrentModuleObject = "AirflowNetwork:MultiZone:Surface";
AirflowNetworkNumOfSurfaces = GetNumObjectsFound( CurrentModuleObject );
if ( AirflowNetworkNumOfSurfaces > 0 ) {
...
    // either simple or detailed large opening, or crack
    MultizoneSurfaceData( i ).ExternalNodeName = Alphas( 3 ); // Name of external node, but not used at WPC="INPUT"
...
```
Shall I modify the logic here?

Originally I thought you were saying I should not allow any input of local OutdoorAir:Node if SurfaceAverageCalculation wind coeff curve, but what if I need to use these nodes elsewhere (such as in surface heat balance).

xuanluo113:
@jasondegraw So what I should not allow is the input of LocalAirNodeA when connecting to MultizoneSurfaceB, in with SurfaceAverageCalculation specified? The connection is defined in:
```
AirflowNetwork: MultiZone:Surface,
...
  A3, \field External Node Name
      \type object-list
      \object-list ExternalNodeNames
      \object-list OutdoorAirNodeNames
      \note Used if Wind Pressure Coefficient Type = Input in the AirflowNetwork:SimulationControl object,
      \note otherwise this field may be left blank.
```
And in the code implementation, this input is just ignored when SurfaceAverageCalculation - no warnings or fatal errors for the original external nodes.
```
CurrentModuleObject = "AirflowNetwork:MultiZone:Surface";
AirflowNetworkNumOfSurfaces = GetNumObjectsFound( CurrentModuleObject );
if ( AirflowNetworkNumOfSurfaces > 0 ) {
...
    // either simple or detailed large opening, or crack
    MultizoneSurfaceData( i ).ExternalNodeName = Alphas( 3 ); // Name of external node, but not used at WPC="INPUT"
...
```
So it doesn't show any error if I choose SurfaceAverageCalculation and still define the surface link and the original external node which would not be used either as :
```
  AirflowNetwork:SimulationControl,
    House AirflowNetwork,    !- Name
    MultizoneWithDistribution,  !- AirflowNetwork Control
    SurfaceAverageCalculation ,      !- Wind Pressure Coefficient Type
  AirflowNetwork:MultiZone:Surface,
    Garage:FrontDoor,        !- Surface Name
    CRcri,                   !- Leakage Component Name
    NFacade,                 !- External Node Name
    1.0;                     !- Window/Door Opening Factor, or Crack Factor {dimensionless}

  AirflowNetwork:MultiZone:ExternalNode,
    NFacade,                 !- Name
    1.524,                   !- External Node Height {m}
    NFacade_WPCValue;        !- Wind Pressure Coefficient Curve Name
```
So shall I modify the logic here?

Originally I thought you were saying I should not allow any input of local OutdoorAir:Node if SurfaceAverageCalculation, but what if I need to use these nodes elsewhere (such as in surface heat balance).

jasondegraw:
@XuanLuoLBL You are correct that no error or warning is issued for those situations, but those are different. Before the addition of your feature, **SurfaceAverageCalculation** unambiguously meant "ignore the external nodes". Now, with your feature, it could still mean that, but maybe it should mean "ignore the external nodes, but not the local air nodes". As @lgu1234 said in his earlier message we will need time to sort this out.

We should error out when local air nodes are found and **SurfaceAverageCalculation** is specified. This will require users to use **Input**. That is not an undue burden, and I don't think your feature requires the use of AirflowNetwork, does it?

xuanluo113:
@jasondegraw Thanks for the explainable. I added the error message. Would you mind checking out if that makes sense? Thanks very much.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -1244,12 +1275,56 @@ namespace AirflowNetworkBalanceManager {
...
+							ShowWarningError( RoutineName + CurrentModuleObject + " object, Invalid input " + cAlphaFields( 8 ) + " = " + Alphas( 8 ) );
+							ShowContinueError( "The default value is assigned as Absolute." );
+						}
+					}
+
+ 					MultizoneExternalNodeData( i ).Name = Alphas( 1 ) ; // Name of external node
+ 					NodeNum = GetOnlySingleNode( Alphas( 1 ) , ErrorsFound, CurrentModuleObject, "AirflowNetwork:Multizone:Surface", NodeType_Air, NodeConnectionType_Inlet, 1, ObjectIsParent );
+ 					MultizoneExternalNodeData( i ).OutAirNodeNum = NodeNum; // Name of outdoor air node
+ 					MultizoneExternalNodeData( i ).height = Node( NodeNum ).Height; // Nodal height
+ 					MultizoneExternalNodeData( i ).ExtNum = AirflowNetworkNumOfZones + i; // External node number
```
rraustad, line 1326: This looks wrong: ExtNum = AirflowNetworkNumOfZone + i. How does external node number relate to number of zones?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -441,30 +444,56 @@ namespace HeatBalanceSurfaceManager {
 			for ( SurfNum = 1; SurfNum <= TotSurfaces; ++SurfNum ) {
 				if ( Surface( SurfNum ).HasLinkedOutAirNode ) {
 					if ( Node( Surface( SurfNum ).LinkedOutAirNode ).OutAirDryBulbSchedNum != 0 ) {
-						Surface( SurfNum ).OutDryBulbTemp = GetCurrentScheduleValue( Node( Surface( SurfNum ).LinkedOutAirNode ).OutAirDryBulbSchedNum );
+						Surface( SurfNum ).OutDryBulbTemp = Node( Surface( SurfNum ).LinkedOutAirNode ).OutAirDryBulb;
```
rraustad: Do you still need to check for OutAirDryBuldSchedNum != 0 here (and 3 other places below)? If this LindedOutAirNode does not have a schedule then shouldn't Surface.OutDryBulbTemp = Node.OutAirDryBulb ?

src/EnergyPlus/OutAirNodeManager.cc
```diff
@@ -334,24 +336,29 @@ namespace OutAirNodeManager {
 				if ( NumNums > 0 ) Node( NodeNums( 1 ) ).Height = Numbers( 1 );
 
 				if ( NumAlphas > 1 ) {
+					AnyLocalEnvironmentsInModel = true;
+					Node( NodeNums( 1 ) ).IsLocalNode = true;
+				}
+
+				if ( NumAlphas > 1 && !lAlphaBlanks( 2 ) ) {
 					Node( NodeNums( 1 ) ).OutAirDryBulbSchedNum = GetScheduleIndex( Alphas( 2 ) );
```
rraustad, line 344: There's no warning here for when a schedule name is not found (i.e., OutAirDryBulbSchedNum = 0).

Myoldmopar (changes requested):
There has been extensive review so far, and thanks for that @jasondegraw and @rraustad.  I will look over it a bit more later, but for now I'll say one thing: the IDD change **must** be reverted back out of this branch.  Our IDD is frozen.  That cleanup can be added for the next release.

jasondegraw:
@Myoldmopar @rraustad could you review the error out on this feature with SurfaceAverageCalculation? It doesn't look right to me, but I'm not going to be able to look at it this week.

src/EnergyPlus/OutAirNodeManager.cc
rraustad: I read this wrong, the warnings were staring me in the face.

rraustad: OK, you just added these warnings. Good.

Myoldmopar (approved):

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -967,6 +982,17 @@ namespace AirflowNetworkBalanceManager {
...
+			for ( k = 1; k <= NumOfNodes; ++k ) {
+				if ( Node( k ).IsLocalNode ) {
+					ShowSevereError( RoutineName + "Invalid " + cAlphaFields( 3 ) + "=" + Alphas( 3 ) );
+					ShowContinueError( "A local air node is defined to INPUT the wind pressure coefficient curve, while Wind Pressure Coefficient Type is set to SurfaceAverageCalculation.");
+					ShowContinueError( "It requires  the Wind Pressure Coefficient Type be set to INPUT to use the local air node.");
+					ErrorsFound = true;
+					SimObjectError = true;
+					break;
+				}
+			}
```
Myoldmopar, line 994: @jasondegraw I mean this *seems* OK to me.  Inside the `SurfaceAverageCalculation` block, looping over all nodes and throwing if any of them are local nodes.  That's the gist of what error you were wanting, right?

Myoldmopar:
There is still a unit test failure here.  @XuanLuoLBL please address.  Once this unit test failure is fixed, CI will be all clean.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
jasondegraw: @Myoldmopar I guess, but it still doesn't look quite right to me. If it looks OK to you, it's probably OK, though, so I'll approve here in a second.

jasondegraw (approved):

src/EnergyPlus/AirflowNetworkBalanceManager.cc
Myoldmopar: Thanks @jasondegraw, I think the only issue holding this up now is the unit test failure.

Myoldmopar:
@XuanLuoLBL you just pulled master into this branch.  That's bad.  You need to revert that back out.  We work off of develop, not master.

Myoldmopar:
But the diffs on the PR files changed tab look fine.  What happened here?

xuanluo113:
@Myoldmopar Weird that the file changes are rendering the reverted commit. And actually commit "dcd5a01" doesn't show any file change at all.

My branch is up to date to develop right now. 

Got trapped by this and I'll never ever ever use that windows git console again...

Are things back to normal now?

mjwitte:
@XuanLuoLBL If I'm following this correctly, you haven't reverted the problem merge from master yet.  That was https://github.com/NREL/EnergyPlus/pull/6263/commits/2169d457856b9f766395f5d03916b381c5b7b32f  But don't do anything more until @Myoldmopar confirms.

Myoldmopar:
Correct; something is still wrong.

xuanluo113:
@mjwitte @Myoldmopar ` 2169d45 ` only added a blank space. And `dcd5a01` did nothing at all. And that's why it didn't bother me then.
And now when I wish to pull from origin/develop, it shows up-to-date, which is not...

xuanluo113:
@Myoldmopar I actually feel if I revert the reverted commit things gonna be fine. Or shall I checkout some specific commit?

Myoldmopar:
I will push up a different branch in a minute and we're closing this PR and opening a new one to merge from.

xuanluo113:
Okay thanks. Sorry about the mess.

Myoldmopar:
I still can't get it.  I'll have to try again tomorrow.

xuanluo113:
@Myoldmopar Can I just open up another branch and delete this one? I dare not do anything before you confirm now.

The unit test issue is killing me as well. It passes perfectly on my local machine every time from the very beginning and I just don't know what might trigger the difference. Tried lots of time.

Myoldmopar:
I just opened #6335 after what may have been a successful last try.  If that works, we'll just move forward with that one.  If not, then we'll just figure it out tomorrow.

xuanluo113:
@Myoldmopar Thanks.

PR #6265, opened 2017-08-23
Preparation for 8.8 Release

Myoldmopar:
Updates:

- [x] CMake Version numbers
- [x] IDF Version Numbers
- [x] Documentation Version IDs, Dates, and other Installation Rules
- [x] README file

This is a preliminary release prep pull request that should go in as soon as CI is clean.  Any IDF file missed by the version update script should show a diff on the initial check because the error file will say version mismatch.  Once any of those are cleaned up this should be merged in.

This does _not_ update the epupdate trigger file.


Myoldmopar:
The latest commit should fix the version issues, and I also made many improvements to the change_version script so there will be less issues next release.  I hope to take more of the manual effort from this PR and package it up.  At a minimum, the documentation title page version ID should be configured.

Myoldmopar:
Yay for less manual stuff, now the documentation title page is fully configured and doesn't need any handholding each release.  Now that the script automatically does _all_ the idfs, imfs, and unit test files, it's really just a matter of setting a few CMake things, running the change_version script, and updating the README. :+1:   I may go ahead and configure some things on the README also, we'll see.

Myoldmopar:
The one debug failure in the DOAS file is a crash [here]( https://github.com/NREL/EnergyPlus/blob/de27c9b91293dbee549db9fe65f8b5a96f295f83/src/EnergyPlus/SizingManager.cc#L3749) -- @mjwitte

Myoldmopar:
# Diffs and issues

## Windows 32
 - EIO diffs/Table big diffs: Version number change
 - 1 ESO big diff: DElight
 - 11 tests with ERR diffs: These are 11 test files that were still at 8.6.  So the error files said version not expected in those error files in develop, and now with this branch, they are clean.
 - Unit test failures: 3 with "invalid string position" (AirloopHVAC_VentilationRateProcedure, FinAndOverhangCount, TubularDaylightDiffuserCount)

## Windows 64
- Same as Win32 except the "invalid string position" failures aren't there but now here the DElight file shows MTR diffs too

## Mac
- Same as Win64 except the DElight failure isn't there

## Linux
- Just the expected 11 ERR diffs (cleanups)

## Linux-Debug
- Just the one crash mentioned [above](https://github.com/NREL/EnergyPlus/pull/6265#issuecomment-324601306)

As messy as this is, it is actually good to go in.

PR #6270, opened 2017-08-24
Remove problematic unit tests

Myoldmopar:
We are having spurious issues out of some unit tests in CI.  This is making it much harder to review actual test results.  The issues tend to happen on tests coming out of OutputReportTabular, and they are usually the ones with very long delimited input strings.  Two of the tests had very large input decks in the unit test, called ManageSimulation(), all to just check a couple indices.  

Let me be clear, there's **nothing** wrong with this approach.  I like that we're moving more toward  function-based unit testing, but even still, we handle cases like this all over the place in our unit tests.  

However, since our CI seems to be having trouble with them, and we cannot yet reproduce them locally, I'm pulling a few out.  If it cleans up the CI results, it will make our everyday reviewing **much** better!  We may lose a very very small fraction of a percent on unit test coverage, but it will help keep future defects from slipping in.


Myoldmopar:
And in anticipation of the other hot fix dropping in first, I pulled that fix in over here too.  To make sure this is as green as possible.

Myoldmopar:
And while I'm at it........I am going to comment out the DElight file runs from the tests.  When was the last time anyone saw a DElight diff and investigated?  It's been a long time since the diffs have been sketchy on them for so long.  Yeah, just like the other unit tests here, my plan is to figure it out, but until then, there is nothing gained from leaving them in and potentially reducing the focus on meaningful diffs.

mjwitte:
@Myoldmopar Since we've only seen spurious delight diffs on Windows, you *could* disable them only for Windows and keep them running for the others.

Myoldmopar:
That's a **great** idea!

Myoldmopar:
Yes.  That was a beautiful idea @mjwitte, thank you.  I just pushed the conditional version.  I built it on my Linux machine and ctest recognizes the 3 DElight integration tests and two unit tests, but it shouldn't on Windows.  w00t!

Issue #6276, opened 2017-08-26
Electric Baseboard Providing Negative Heating


RKStrand:
A problem was noted with electric radiant/convective baseboard units where there were strange occasional occurrences of negative heating rates being provided by the baseboard unit.  This should never happen.  While simply zeroing out negative loads would be one short-term solution, it is possible that the calculation of ZeroSourceSumHAT is not being updated properly.  If this is the source of the problem, then it could be an issue for other radiant/convective units that also use this variable.  So, this needs to be investigated for this model and then potentially looked at for other models as well.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus (if using an intermediate build, include SHA): V8.8
 - Rich Raustad identified the negative load issue
[BaseboardsOperatingWithNegativeLoadMet.idf.txt](https://github.com/NREL/EnergyPlus/files/1253787/BaseboardsOperatingWithNegativeLoadMet.idf.txt)

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ x ] Defect file added (attached to this issue)
 - [ x ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
This is ID#150572819 in PivotalTracker

RKStrand:
@rraustad @Myoldmopar @mjwitte This is now fixed and a pull request will be created shortly.  The defect file attached above does not report any negative baseboard heating after the fix.  While in the code, I also cleaned up the variable assignments for when the device is on and off (new routines) and eliminated extra comment lines there were basically blank holdovers from F90 days.

rraustad:
Closed via #6277.

PR #6277, opened 2017-08-29
Negative Heating Problem for Radiant-Convective Electric Baseboard

RKStrand:
Pull request overview
---------------------
This solves the issue noted in GitHub Issue #6276.  Negative baseboard heating can be reported without any heating from the electric radiant/convective baseboard unit.  The issue was resolved by trapping cases where the calculated LoadMet variable was negative, and then turning the unit off if resolution was not possible.  In addition, this fix standardized how variables were set when the unit was either on or off by creating new routines for each case.  Before this fix, not all of the same variables were being set which was also not correct.

### Work Checklist
   - Defect: This pull request repairs GitHub defect issue #6276 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ElectricBaseboardRadiator.cc
+++ b/src/EnergyPlus/ElectricBaseboardRadiator.cc
@@ -802,20 +802,20 @@ namespace ElectricBaseboardRadiator {
 				} else {
 
 					UpdateElectricBaseboardOn( AirOutletTemp, ElecBaseboard( BaseboardNum ).ElecUseRate, AirInletTemp, QBBCap, CapacitanceAir, Effic );
-					
+	
 				}
 
 			} else { // zero radiant fraction, no need of recalculation of heat balances
 
 				LoadMet = QBBCap;
 				UpdateElectricBaseboardOn( AirOutletTemp, ElecBaseboard( BaseboardNum ).ElecUseRate, AirInletTemp, QBBCap, CapacitanceAir, Effic );
-				
+
 			}
 
 		} else { // If there is an off condition the BB does nothing.
 
 			UpdateElectricBaseboardOff( LoadMet, QBBCap, RadHeat, QBBElecRadSource( BaseboardNum ), ElecBaseboard( BaseboardNum ).ElecUseRate, AirOutletTemp, AirInletTemp );
-			
+
 		}
 
 		// Assign calculated ones
@@ -842,16 +842,16 @@ namespace ElectricBaseboardRadiator {
 		// SUBROUTINE INFORMATION:
 		//       AUTHOR         Rick Strand
 		//       DATE WRITTEN   August 2017
-		
+
 		// PURPOSE OF THIS SUBROUTINE: Zero out appropriate system variables when it is off
-		
+
 		QBBCap = 0.0;
 		LoadMet = 0.0;
 		RadHeat = 0.0;
 		AirOutletTemp = AirInletTemp;
 		QBBElecRadSrc = 0.0;
 		ElecUseRate = 0.0;
-		
+
 	}
 
 	void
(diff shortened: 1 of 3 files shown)


rraustad:
This looks good. I found a few places where tabs were inserted on new lines. I corrected those. I ran the defect file and results are as expected. No negative loads and convective/radiant report variables are zero when baseboard is off.

6276elecbaseboard

src/EnergyPlus/ElectricBaseboardRadiator.cc
```diff
@@ -872,31 +770,52 @@ namespace ElectricBaseboardRadiator {
...
+					HeatBalanceSurfaceManager::CalcHeatBalanceInsideSurf( ZoneNum );
+					TempZeroSourceSumHATsurf = SumHATsurf( ZoneNum );
+					// Now, turn it back on:
+					QBBElecRadSource( BaseboardNum ) = RadHeat;
+					DistributeBBElecRadGains();
+					HeatBalanceSurfaceManager::CalcHeatBalanceOutsideSurf( ZoneNum );
+					HeatBalanceSurfaceManager::CalcHeatBalanceInsideSurf( ZoneNum );
+					// Recalculate LoadMet with new ZeroSource... term and see if it is positive now.  If not, shut it down.
+					LoadMet = ( SumHATsurf( ZoneNum ) - TempZeroSourceSumHATsurf ) + ( QBBCap * ElecBaseboard( BaseboardNum ).FracConvect ) + ( RadHeat * ElecBaseboard( BaseboardNum ).FracDistribPerson );
+					if ( LoadMet < 0.0 ) {
```
rraustad, line 795: Since I don't understand this I'll ask. What does it mean when TempZeroSourceSumHATsurf is different from ZeroSourceSumHATsurf( ZoneNum ) calculated at line 569 when FirstHVACIteration = true ? To me that means some model somewhere in code did not correctly account for the boundary condition that was present at the beginning of the time step (an energy leak).

rraustad: If there were a fast change in temperature, I assume from some other HVAC model running, wouldn't QZnReq reflect that capacity and update the load the baseboard needed to meet accordingly?

    QZnReq = ZoneSysEnergyDemand( ZoneNum ).RemainingOutputReqToHeatSP;

And isn't internal loads accounted for in predictor?

RKStrand: @rraustad I believe that internal loads are accounted for in predictor and if some other HVAC model running should update what is required from the baseboard in QZnReq.  In reality, QZnReq is the most important parameter here, not LoadMet.  LoadMet is helpful for perhaps figuring out what else is needed to get to the right setpoint temperature but that will all work out with iterations and getting to the right temperature.

As for the ZeroSourceSumHATsurf, what I was seeing was occasional anomalies where SumHATsurf was less than ZeroSourceSumHATsurf.  It seemed to be happening at times when there was a relative fast temperature change in the zone air.  The ZeroSource term is basically the best estimate at the extra convection from the walls to the zone air as a result of the radiant energy.  It's only an estimate to help the overall HVAC system get to the setpoint and isn't used in any actual heat balance.  So, even if something gets messed up in the ZeroSource term somehow, it won't result in any energy imbalances.  That's why I decided to simply give the baseboard one more shot at an accurate calculation if LoadMet is negative without messing with the value that is contained in ZeroSource.

If we really wanted to get more accurate numbers, we would always run any radiant type system first with no load and then with whatever we needed and that would be the best guess at LoadMet.  However, the radiant systems already cycle through the heat balances more than the forced air systems and yet another iteration through the expensive heat balances would simply slow things down more.  Does this help or did I make things more confusing with this post?

RKStrand:
@rraustad Thanks for fixing the tabs.  I do not understand why that happens.  Everything looks good on my computer as far as following the rules as far as I know.  I appreciate you just fixing them as that is helpful and probably faster anyway.

src/EnergyPlus/ElectricBaseboardRadiator.cc
rraustad: I just wanted to think through this one more time. If you are comfortable then that's good enough for me.

rraustad (approved):
Changes reviewed and discussed.

src/EnergyPlus/ElectricBaseboardRadiator.cc
RKStrand: @rraustad Totally understand.  It's one of those weird things with any of the systems that have a radiant component to them.  It's the best way that I know to get some sort of estimate as to what effect that radiant component will have on meeting the load.  It's not perfect but it avoids adding yet another layer of iteration in a section of code that has already been critiqued in the past as being slow.  So, it's a compromise but I am comfortable with it for now.

rraustad:
Merged and Closed

Issue #6274, opened 2017-08-25
User file fails with CalcMultiSpeedDXCoilCooling developer error


mjwitte:
Issue overview
--------------
   ** Severe  ** CalcMultiSpeedDXCoilCooling: Coil:Cooling:DX:MultiSpeed "ZONE RETAIL BACK_SPACE A  - STORY GROUND PSZ-VAV VAR SPD DX AC CLG COIL Developer error - inconsistent airflow rates.
   **   ~~~   ** When AirMassFlow > 0.0 and CycRatio > 0.0, then MSHPMassFlowRateHigh must also be > 0.0
   **   ~~~   **  During Warmup, Environment=DENVER INTL AP ANN CLG .4% CONDNS DB=>MWB, at Simulation time=07/21 00:00 - 00:15
   **   ~~~   ** AirMassFlow=9.388E-002,CycRatio=1.000, MSHPMassFlowRateHigh=0.000
   **  Fatal  ** Preceding condition(s) causes termination.

Workaround: If 2 AirloopHVAC:UnitarySystem object are in series on the main air loop branch, the UnitarySystem with the Coil:Cooling:DX:MultiSpeed object must be first in the series.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus Version 8.8.0-ddaa5689de
 - helpdesk ticket number 12173

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles\6274 run with Denver weather
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6280

PR #6280, opened 2017-08-29
AirloopHVAC:UnitarySystem fails sizing when Coil:Cooling:DX:MultiSpeed is used

rraustad:
Pull request overview
---------------------
When two AirloopHVAC:UnitarySystem are in series on the main air loop branch and the second system includes a Coil:Cooling:DX:MultiSpeed cooling coil, the cooling coil fails the sizing calculations. Corrects #6274.

Work around: place the Unitary System with the DX multispeed cooling coil upstream of the other system.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HVACUnitarySystem.cc
+++ b/src/EnergyPlus/HVACUnitarySystem.cc
@@ -1783,8 +1783,6 @@ namespace HVACUnitarySystem {
 		Real64 CoolCapAtPeak; // cooling capacity at peak [W]
 		Real64 HeatCapAtPeak; // heating capacity at peak [W]
 		std::string SystemType; // type of air loop equipment
-		Real64 OnOffAirFlowRatio; // used to pass to cooling coil for sizing
-		Real64 PartLoadRatio; // used to pass to cooling coil for sizing
 		bool TempCoolingLoad; // size cooling coils with a cooling load, save actual load
 		bool TempHeatingLoad; // save actual load
 		Real64 SysCoolingFlow; // individually sized cooling flow rate [m3/s]
--- a/src/EnergyPlus/HVACUnitarySystem.cc
+++ b/src/EnergyPlus/HVACUnitarySystem.cc
@@ -2241,7 +2241,7 @@ namespace HVACUnitarySystem {
 			}
 
 			// set the multi-speed high flow rate variable in case a non-zero air flow rate resides on the coil inlet during sizing (e.g., upstream system ran prior to this one)
-			MSHPMassFlowRateHigh = 1.0; // doesn't matter what this value is since only coil size is needed and CompOn = 0 here
+			MSHPMassFlowRateHigh = EqSizing.CoolingAirVolFlow * StdRhoAir; // doesn't matter what this value is since only coil size is needed and CompOn = 0 here
 			SimDXCoilMultiSpeed( BlankString, 1.0, 1.0, UnitarySystem( UnitarySysNum ).CoolingCoilIndex, 0, 0, 0 );
 			DXCoolCap = GetCoilCapacityByIndexType( UnitarySystem( UnitarySysNum ).CoolingCoilIndex, UnitarySystem( UnitarySysNum ).CoolingCoilType_Num, ErrFound );
 			EqSizing.DesCoolingLoad = DXCoolCap;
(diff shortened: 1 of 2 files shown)


src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2242,9 +2242,8 @@ namespace HVACUnitarySystem {
 				if ( ! allocated( UnitarySystem( UnitarySysNum ).MSCoolingSpeedRatio ) ) UnitarySystem( UnitarySysNum ).MSCoolingSpeedRatio.allocate( UnitarySystem( UnitarySysNum ).NumOfSpeedCooling );
 			}
 
-			OnOffAirFlowRatio = 1.0;
-			PartLoadRatio = 1.0;
-			SetOnOffMassFlowRate( UnitarySysNum, OnOffAirFlowRatio, PartLoadRatio );
+			// set the multi-speed high flow rate variable in case a non-zero air flow rate resides on the coil inlet during sizing (e.g., upstream system ran prior to this one)
+			MSHPMassFlowRateHigh = 1.0; // doesn't matter what this value is since only coil size is needed and CompOn = 0 here
```
rraustad: This was the problem. 2 unitary system on the main branch. The first sized OK and operated at some non-zero air flow rate. The second system tried to size and failed with the developer warning. The attempt to set MSHPMassFlowRateHigh by calling SetOnOffMassFlowRate did not work here since the air flow variables have not yet been sized/set. MSHPMassFlowRateHigh was equal to 0 here and thus the fatal.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9092,7 +9091,7 @@ namespace HVACUnitarySystem {
 		Real64 OutletHumRatDXCoil; // Actual outlet humidity ratio of the DX cooling coil
 		int SolFla; // Flag of solver, num iterations if >0, else error index
 		int SolFlaLat; // Flag of solver for dehumid calculations
-		Array1D< Real64 > Par( 8 ); // Parameter array passed to solver
+		Array1D< Real64 > Par( 10 ); // Parameter array passed to solver
```
rraustad, line 9092: The rest of the changes were due to failure of the unit test. In these routines, only the coil model is called (instead of the entire unitary system) to save time. When a blow through fan is used, the entire system needs to be modeled so fan heat is accounted for via the OnOffFanPartLoadFrac variable.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9500,19 +9504,12 @@ namespace HVACUnitarySystem {
 							if ( UnitarySystem( UnitarySysNum ).CoolingSpeedNum > 1.0 ) {
 								Par( 4 ) = CycRatio;
 								SolveRoot( Acc, MaxIte, SolFla, SpeedRatio, DXCoilVarSpeedResidual, 0.0, 1.0, Par );
-								UnitarySystem( UnitarySysNum ).CoolingCycRatio = SpeedRatio;
-								UnitarySystem( UnitarySysNum ).CoolingPartLoadFrac = SpeedRatio;
-								CalcPassiveSystem( UnitarySysNum, AirLoopNum, FirstHVACIteration );
```
rraustad, line 9505: The call the CalcPassiveSystem (a simple routine that just calls the child components in order) was moved to  the SolveRoot function and called each iteration. The unit test now hits the outlet node set point perfectly.

src/EnergyPlus/HVACUnitarySystem.cc
lgu1234: @rraustad I also checked sizing values in eio files from both runs and find only difference is cooling capacity:

The existing program is unable to size the cooling capacity and issued a fatal message.

The fixed program is able to generate the cooling capacity shown below: 
 Component Sizing Information, AirLoopHVAC:UnitarySystem, ZONE RETAIL BACK_SPACE A  - STORY GROUND PSZ-VAV UNITARY COOLING, Design Size Nominal Cooling Capacity [W], 3081.61625.

Therefore, the value of MSHPMassFlowRateHigh does not matter. However, in order to pass a error check, the value has to be > 0.0. It seems that "MSHPMassFlowRateHigh = 1.0" looks too arbitrary. Can we set up as design flow rate?

rraustad: I could set it to EqSizing.CoolingAirVolFlow if that would look better. The coil is not simulated here since CompOp = 0 and jumps past the code in DXCoil to where the coil is off. But I don't disagree that using 1.0 is misleading even though the comment says it doesn't matter what this value is. Since you are on this branch can you make that change?

lgu1234:
The fatal error has been fixed. A minor change will be uploaded. Will merge it as long as CI machine is green,

lgu1234:
Results are all green.

Issue #5120, opened 2015-08-18
Reverse DD failures


Myoldmopar:
These files show some diff on reverse design day testing.  It looks like there are some common themes, so a big chunk of them might be addressed with a small fix.  If we need to break these out into individual issues, that's fine.  This is more a placeholder to start investigating them.
- 1ZoneDataCenterCRAC_wPumpedDXCoolingCoil.idf
- 2ZoneDataCenterHVAC_wEconomizer.idf
- 5ZoneAirCooledWithCoupledInGradeSlab.idf
- 5ZoneCoolBeam.idf
- 5ZoneEngChill.idf
- 5ZoneIceStorage.idf
- DisplacementVent_1ZoneOffice.idf
- DOAToWaterToAirHPInlet.idf
- DOAToWaterToAirHPSupply.idf
- GSHP-GLHE.idf
- GSHPSimple-GLHE.idf
- GSHP-Slinky.idf
- HeatPumpSecondaryCoil.idf
- HeatPumpWaterToAirEquationFit.idf
- HeatPumpWaterToAir.idf
- HeatPumpWaterToAirWithAntifreezeAndLatentModel2.idf
- HeatPumpWaterToAirWithAntifreezeAndLatentModel.idf
- HeatPumpWaterToAirWithRHControl.idf
- PlateHeatExchanger.idf
- RefBldgLargeOfficeNew2004_Chicago.idf
- RefBldgLargeOfficeNew2004_Chicago-ReturnReset.idf
- RefBldgPrimarySchoolNew2004_Chicago.idf
- RefBldgSecondarySchoolNew2004_Chicago.idf
- UnitarySystem_VSHeatPumpWaterToAirEquationFit.idf
- VentilatedSlab_SeriesSlabs.idf
- VSHeatPumpWaterToAirEquationFit.idf
- ZoneCoupledGroundHTBasement.idf
- ZoneCoupledGroundHTSlabInGrade.idf
- ZoneCoupledGroundHTSlabOnGrade.idf

rraustad:
Anyone working on these? I'm looking around...

mjwitte:
"Not I," said the dog.

PR #6281, opened 2017-08-30
Move ResetNodeData to resolve reverse DD issue for IT equipment model and cleanup on 5ZoneCAV_MaxTemp

mjwitte:
Pull request overview
---------------------
ElectricEquipment:ITE:AirCooled uses a node temperature to compute the IT equipment power consumption and more.  On BeginEnvrnFlag, the node data needs to be reset before the IT equipment is computed.  So, `ResetNodeData` has been moved from `HVACManager` up into `SimulationManager`.  This should clear up revese design day issues (#5120) for 1ZoneDataCenterCRAC_wPumpedDXCoolingCoil and 2ZoneDataCenterHVAC_wEconomizer, and possibly others.

These changes caused a divide by zero in the EMS PUE calcs in 1ZoneDataCenterCRAC_wPumpedDXCoolingCoil.  Add div-by-zero protection in both 1ZoneDataCenterCRAC_wPumpedDXCoolingCoil and 2ZoneDataCenterHVAC_wEconomizer EMS programs.

5ZoneCAV_MaxTemp is almost all autosized, but there was a stray hard flow rate in one of the terminal units.  

Move SetOutAirNodes to be before emsCallFromBeginTimestepBeforePredictor, because original move of ResetNodeData caused some EMS files to show diffs.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HVACManager.cc
+++ b/src/EnergyPlus/HVACManager.cc
@@ -380,10 +380,10 @@ namespace HVACManager {
 		NumOfSysTimeSteps = 1;
 		FracTimeStepZone = TimeStepSys / TimeStepZone;
 
+		SetOutAirNodes();
+
 		bool anyEMSRan;
 		ManageEMS( emsCallFromBeginTimestepBeforePredictor, anyEMSRan ); //calling point
-
-		SetOutAirNodes();
 
 		ManageRefrigeratedCaseRacks();
 
--- a/src/EnergyPlus/HVACManager.cc
+++ b/src/EnergyPlus/HVACManager.cc
@@ -380,10 +380,10 @@ namespace HVACManager {
 		NumOfSysTimeSteps = 1;
 		FracTimeStepZone = TimeStepSys / TimeStepZone;
 
-		SetOutAirNodes();
-
 		bool anyEMSRan;
 		ManageEMS( emsCallFromBeginTimestepBeforePredictor, anyEMSRan ); //calling point
+
+		SetOutAirNodes();
 
 		ManageRefrigeratedCaseRacks();
 
--- a/src/EnergyPlus/HVACManager.cc
+++ b/src/EnergyPlus/HVACManager.cc
@@ -360,7 +360,6 @@ namespace HVACManager {
 		}
 
 		if ( BeginEnvrnFlag && MyEnvrnFlag ) {
-			ResetNodeData();
 			AirLoopsSimOnce = false;
 			MyEnvrnFlag = false;
 			InitVentReportFlag = true;
--- a/src/EnergyPlus/SimulationManager.cc
+++ b/src/EnergyPlus/SimulationManager.cc
@@ -499,6 +499,8 @@ namespace SimulationManager {
 				isFinalYear = true;
 			}
 
+			HVACManager::ResetNodeData(); // Reset here, because some zone calcs rely on node data (e.g. ZoneITEquip)
+
 			bool anyEMSRan;
 			ManageEMS( emsCallFromBeginNewEvironment, anyEMSRan ); // calling point
 
--- a/src/EnergyPlus/ZonePlenum.cc
+++ b/src/EnergyPlus/ZonePlenum.cc
@@ -766,6 +766,16 @@ namespace ZonePlenum {
 				Node( ZoneNodeNum ).HumRat = OutHumRat;
 				Node( ZoneNodeNum ).Enthalpy = PsyHFnTdbW( Node( ZoneNodeNum ).Temp, Node( ZoneNodeNum ).HumRat );
 
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneTemp = 20.0;
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneHumRat = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneEnthalpy = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletTemp = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletHumRat = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletEnthalpy = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletPressure = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRate = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRateMaxAvail = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRateMinAvail = 0.0;
 			}
 
 			InitAirZoneReturnPlenumEnvrnFlag = false;
@@ -881,6 +891,16 @@ namespace ZonePlenum {
 				Node( ZoneNodeNum ).HumRat = OutHumRat;
 				Node( ZoneNodeNum ).Enthalpy = PsyHFnTdbW( Node( ZoneNodeNum ).Temp, Node( ZoneNodeNum ).HumRat );
 
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneTemp = 20.0;
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneHumRat = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneEnthalpy = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletTemp = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletHumRat = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletEnthalpy = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletPressure = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRate = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRateMaxAvail = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRateMinAvail = 0.0;
 			}
 
 			MyEnvrnFlag = false;
--- a/testfiles/2ZoneDataCenterHVAC_wEconomizer.idf
+++ b/testfiles/2ZoneDataCenterHVAC_wEconomizer.idf
@@ -3096,7 +3096,11 @@
 
   EnergyManagementSystem:Program,
     CalculatePUE,            !- Name
-    set PUE_Value = Whole_Building_Power / IT_Equip_Power;  !- Program Line 1
+    IF IT_Equip_power > 0.0, !- Program Line 1
+    set PUE_Value = whole_building_power / IT_Equip_power,  !- Program Line 2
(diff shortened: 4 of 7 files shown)


rraustad:
Existing diff's look OK to me.

mjwitte:
I still need to add a unit test here.

mjwitte:
@Myoldmopar @rraustad @EnergyArchmage These are small, but somewhat fundamental changes, so I'd like some perspective on whether these changes look OK.

src/EnergyPlus/SimulationManager.cc
```diff
@@ -499,6 +499,8 @@ namespace SimulationManager {
 				isFinalYear = true;
 			}
 
+			HVACManager::ResetNodeData(); // Reset here, because some zone calcs rely on node data (e.g. ZoneITEquip)
+
```
rraustad, line 503: OK, ResetNodeData is called, literally, at the beginning of the next environment.

src/EnergyPlus/ZonePlenum.cc
```diff
@@ -766,6 +766,16 @@ namespace ZonePlenum {
...
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneTemp = 20.0;
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneHumRat = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).ZoneEnthalpy = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletTemp = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletHumRat = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletEnthalpy = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletPressure = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRate = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRateMaxAvail = 0.0;
+				ZoneRetPlenCond( PlenumZoneNum ).InletMassFlowRateMinAvail = 0.0;
```
rraustad, line 778: Well, I guess this can't hurt, to loop through all plenums and reset node data ONCE on the first BeginEnvrnFlag. InitAirZoneReturnPlenumEnvrnFlag is set to false directly after this loop so on next BeginEnvrnFlag, this won't exectue.

src/EnergyPlus/ZonePlenum.cc
```diff
@@ -881,6 +891,16 @@ namespace ZonePlenum {
...
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneTemp = 20.0;
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneHumRat = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).ZoneEnthalpy = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletTemp = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletHumRat = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletEnthalpy = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletPressure = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRate = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRateMaxAvail = 0.0;
+				ZoneSupPlenCond( PlenumZoneNum ).InletMassFlowRateMinAvail = 0.0;
```
rraustad, line 903: So here each plenum node data is reset individually each BeginEnvirnFlag. Would it be smarter to do the first reset EACH  BeginEnvrnFlag so that all plenums are reset at the same time in case something I can't think of needs all plenums to be reset at the same time? e.g.,

    if ( ! BeginEnvrnFlag ) {
        InitAirZoneReturnPlenumEnvrnFlag  = true;
     }

testfiles/2ZoneDataCenterHVAC_wEconomizer.idf
```diff
@@ -3096,7 +3096,11 @@
 
   EnergyManagementSystem:Program,
     CalculatePUE,            !- Name
-    set PUE_Value = Whole_Building_Power / IT_Equip_Power;  !- Program Line 1
+    IF IT_Equip_power > 0.0, !- Program Line 1
+    set PUE_Value = whole_building_power / IT_Equip_power,  !- Program Line 2
+    ELSE,                    !- Program Line 3
+    set PUE_Value = 0.0,     !- Program Line 4
+    ENDIF;                   !- Program Line 5
```
rraustad, line 3103: So this is the reason for a reverse DD issue? PUE_Value was retained for the next environment?

mjwitte: No, the ITE supply air inlet node temp was retained into the next environment, because IT equipment gets simulated in the heat balance, before HVACManager is called, which is where ResetNodeData used to be.

src/EnergyPlus/ZonePlenum.cc
mjwitte: My intent was to reset all plenums once at the beginning of each environment.   Will recheck that.

rraustad: This is just weird. At the top of this loop the node data is reset. That's what ResetNodeData is supposed to do, except the humrat probably isn't reset to outdoor humrat.

No show stoppers here.

rraustad: ResetNodeData sets everything to 0 so I wonder why this needs a special case?

mjwitte: @rraustad Looking at the whole init functions, for both supply and return plenums, the begin environment init happens once for each environment for all plenums.  If these were OO'd we'd initialize each one individually, but this should be OK for now.

rraustad:
This looks all good to me. My guess is that the ZoneRetPlenCond.* data doesn't even need to be reset but I could be wrong. It looks to me like the old method of resetting node information on BeginEnvrnFlag and now it's not needed. It doesn't hurt anything so I am OK with this.

rraustad:
I'm going to give @Myoldmopar @EnergyArchmage a chance to review if they'd like before I merge. I don't see anything wrong here. It will drop tomorrow at the latest.

testfiles/2ZoneDataCenterHVAC_wEconomizer.idf
rraustad: Then why was this changed?

mjwitte:
Regarding the existing node data resets in Init...Plenum, the plenum wants it's zone node temp to match what the zone are initialized to which is 20C and outdoor humrat.  But ResetNodeData sets everyone to 0C and 0kg/kg.  You may be right that the other plenum resets aren't needed, but at one point, I thought I saw some flow lingering in the plenum when it should have been zero.  So it seemed safer to do these to be sure.

rraustad:
Still gonna drop tomorrow.

testfiles/2ZoneDataCenterHVAC_wEconomizer.idf
mjwitte: 'cause it blew up with a zero for IT_Equip_Power after the other fixes went in.

src/EnergyPlus/HVACManager.cc
```diff
@@ -381,11 +380,11 @@ namespace HVACManager {
 		NumOfSysTimeSteps = 1;
 		FracTimeStepZone = TimeStepSys / TimeStepZone;
 
+		SetOutAirNodes();
+
```
EnergyArchmage: humm, there are 4 EMS actuators for overriding out  node conditions, with this change the calling point will no longer work for those.   I think this will be a problem

rraustad: So move it back after the ManageEMS. @mjwitte why was this moved?

mjwitte: @EnergyArchmage My (flawed) thinking here was if an outdoor air node is used as a sensor, it will see a zero is SetOutAirNodes hasn't happened yet.  But you're right, SetOutAirNodes won't respond to these until the next timestep (if at all?).  I'll put this one back.  

@rraustad hold on merging this.

But this makes me wonder if the timing of the new exterior conditions code in [InitHeatBalance](https://github.com/NREL/EnergyPlus/blob/001b3a8b6586a169ecaaab5666a8d77d53d15a18/src/EnergyPlus/HeatBalanceManager.cc#L4956-L4991) will work correctly - the only calling point I see before this is emsCallFromBeginNewEvironment.  The emsCallFromBeginTimestepBeforePredictor here seems too late to impact any surface heat balance - what am I missing here?  Seems like we need another calling point like emsCallFromBeginTimestepBeforeHeatBalance.

EnergyArchmage: in practice it often has been the case that emsCallFromBeginTimestepBeforePredictor doesn't work well, this InitHeatBalance timing may be the reason.  What does work, albeit with more lag, is to use the emsCallFromEndZoneTimestepAfterZoneReporting calling point to set things at the end of the previous timestep.  Might be another issue here, I don't think I ever looked at these new actuators before.

mjwitte: Those particular actuators in InitHeatBalance are very new.  And there's an additional call to SetOutAirNodes there.

mjwitte:
@Myoldmopar @mitchute It would be nice to confirm that this gets rid of reverseDD errors for the 1ZoneDataCenter and 2ZoneDataCenter files.  Who's set up to run these?

mitchute:
I can take a look.

mitchute:
Looks good! That cleans up those two files plus the Attic_ASTM_C1340 file.

mjwitte:
@mitchute Cool - thanks.

Myoldmopar:
:+1: :+1: :+1:

rraustad:
Diff's look the same. Still droppin' today.

mjwitte:
I would like to understand the diffs in the EMS files.

rraustad:
No eio changes except for warmup diff's. You initialized plenum temps on BeginEvrnFlag. The plenum temps changed slightly causing diff's in this and other files. At least that would be my guess. I typically don't investigate warmup diff's only.

mjwitte:
EMSDemandManager_LargeOffice has some big diffs in gas use.

rraustad:
That usually means boiler turned on at a different time. You looking at this or you want me to?

rraustad:
I'm looking, you can work on something else.

rraustad:
Well it's not on off timing. Just a change in zone temp that caused a change in boiler gas use of 4 W.

image

rraustad:
It's interesting that there are no diff's on Windows, yet diff's on Mac and Linux.

mjwitte:
I pretty much ignore windows CI at the moment.  I don't believe there's no diffs here.

rraustad:
Are you OK with the EMS files now? I have the results if you want to look at some other outputs.

mjwitte:
If you're ok with it, go ahead.

rraustad:
Reviewed, critiqued and adjusted based on comments. Diff's reviewed and seem appropriate for changes made.

Issue #6222, opened 2017-07-26
Include Surface PV in Multiplier


Myoldmopar:
Issue overview
--------------
A user was experiencing an issue with surface connected PV.  

As per the norm, they created a building model with a top/middle/bottom floor.  Then they planned to use a multiplier to multiply the middle floor to match a certain building size.

They connected PV panels to building surfaces.  They then noticed that when multipliers were in place, the PV production was not being multiplied.

It seems this should be possible since the PV is connected to surfaces which are connected to zones which are multiplied.  Let's at least have a discussion here to figure out the magnitude.

Two files are attached, an actual geometry file that includes the surfaces all generated for all floors, which takes a lot of input and time to run, and a file with a multiplier that _should_ produce the same amount of generation (or very close) after being multiplied.
[actual_geometry.idf.txt](https://github.com/NREL/EnergyPlus/files/1177334/actual_geometry.idf.txt)
[multiplier.idf.txt](https://github.com/NREL/EnergyPlus/files/1177335/multiplier.idf.txt)

lgentile:
I believe this is an issue that is blocking deliverables work on an NREL FY17 SBV project. Should be worked at with priority if at all possible.

RKStrand:
@Myoldmopar So, I ran both of the input files and looked at the code.  I'm actually getting similar results (differences in facility electric are 0.4%) and the code seems to indicate that the zone multiplier is being used.  Here is the code out of ReportPV subroutine:

```
if ( Surface( PVarray( PVnum ).SurfacePtr ).Zone != 0 ) { // might need to apply multiplier
	thisZone = Surface( PVarray( PVnum ).SurfacePtr ).Zone;
	PVarray( PVnum ).Report.DCEnergy *= ( Zone( thisZone ).Multiplier * Zone( thisZone ).ListMultiplier );
	PVarray( PVnum ).Report.DCPower *= ( Zone( thisZone ).Multiplier * Zone( thisZone ).ListMultiplier );
}
```

So, a multiplier is being applied as per this code and in this input file it seems to be doing what it is supposed to do.  Am I missing something?

Myoldmopar:
@RKStrand I'm actually able to reproduce the diffs, but it's not super straightforward.  First, I had to add relevant output variables (these idfs were part of a workflow that had the variables added later):

```
Output:Variable,*,Electric Load Center Produced Electric Power,hourly; !- HVAC Average [W]
[...]

PR #6283, opened 2017-08-31
PV Zone Multiplier Fix

RKStrand:
Pull request overview
---------------------
This work addresses GitHub Issue #6222 which showed that the zone multiplier was not being applied to the PV arrays.  This fixes the problem so that the zone multiplier is applied properly when PV surfaces are part of a zone.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #6222 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/Photovoltaics.cc
+++ b/src/EnergyPlus/Photovoltaics.cc
@@ -856,47 +856,45 @@ namespace Photovoltaics {
 		//       AUTHOR         B. Griffith
 		//       DATE WRITTEN   Jan. 2004
 		//       MODIFIED       B. Griffith, Aug. 2008
-		//       RE-ENGINEERED  na
 
 		// PURPOSE OF THIS SUBROUTINE:
 		// collect statements that assign to variables tied to output variables
 
-		// METHODOLOGY EMPLOYED:
-		// <description>
-
-		// REFERENCES:
-		// na
-
-		// USE STATEMENTS:
-		// na
 		// Using/Aliasing
 		using DataHeatBalance::Zone;
+		using DataGlobals::NumOfZones;
 		using DataSurfaces::Surface;
 		using DataHeatBalFanSys::QPVSysSource;
 		using TranspiredCollector::SetUTSCQdotSource;
-
-		// Locals
-		// SUBROUTINE ARGUMENT DEFINITIONS:
-
-		// SUBROUTINE PARAMETER DEFINITIONS:
-		// na
-
-		// INTERFACE BLOCK SPECIFICATIONS:
-		// na
-
-		// DERIVED TYPE DEFINITIONS:
-		// na
+		using InputProcessor::FindItemInList;
 
 		// SUBROUTINE LOCAL VARIABLE DECLARATIONS:
 		int thisZone; // working index for zones
+		int thisSurface; // index for PV surface
+		Array1D< bool > PVarrayZoneInitialized; // avoid doing the zone check too often
+		bool FirstTimeThru( true );
 
 		PVarray( PVnum ).Report.DCEnergy = PVarray( PVnum ).Report.DCPower * ( TimeStepSys * SecInHour );
 
+		thisSurface = PVarray( PVnum ).SurfacePtr;
+
+		if ( FirstTimeThru ) {
+			PVarrayZoneInitialized.dimension( NumPVs, false );
+			FirstTimeThru = false;
+		}
+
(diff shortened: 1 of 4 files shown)


RKStrand:
@Myoldmopar @lgentile Here is the pull request for this.  Results between the two input files noted in the GitHub Issue (#6222) now should identical results in the electric load center output.  I know this was a priority issue for OpenStudio, so I'm over notifying that this is ready to go to make sure it is not lost in all of the other things going on in the code right now.  I suggested Edwin as a reviewer, but that is just a suggestion.  Thanks!

lgentile:
Thank you @RKStrand ! This will help us move forward with our other OS-based project.

src/EnergyPlus/Photovoltaics.cc
```diff
@@ -856,55 +856,53 @@ namespace Photovoltaics {
...
-		// na
-
-		// DERIVED TYPE DEFINITIONS:
-		// na
+		using InputProcessor::FindItemInList;
 
 		// SUBROUTINE LOCAL VARIABLE DECLARATIONS:
 		int thisZone; // working index for zones
+		int thisSurface; // index for PV surface
+		Array1D< bool > PVarrayZoneInitialized; // avoid doing the zone check too often
```
Myoldmopar: Does this really persist through multiple calls into this function?  It looks like it's just a non-static local variable, so it will be cleaned back up each time this function is called.  And the `FirstTimeThru` variable is similar, it will just be true each time the function gets called again, right?

Here's what I'd like to propose.  Instead of adding your own new array here and first time flag, just add a single bool to the main PV array called `bool zoneInitialized;`.  Then you can let it do the allocations, you don't need an allocated flag, you don't need static variables, and you will actually have fewer code changes to accomplish this same task.  

If we kept it the way it currently is, you'd need to make the local variables static first, and then because they are static you'd need to clean them up in the `clear_state` function of this module.  (And if it doesn't have a `clear_state` you'd need to write one of your own and add it to the EnergyPlusFixture unit test class....)  In summary, avoid statics when possible! 😃 

Let me know if you have any issues with this.  The actual code fix here is great and I appreciate you getting it done so quickly.

RKStrand: @Myoldmopar I don't think that should be a problem.  I will work on that as soon as I can.  I was thinking keeping it local would be better, but if what you have proposed is better, I'm all for that.  Thanks for the review and suggestion.  And glad that I could help on this!

src/EnergyPlus/Photovoltaics.cc
```diff
@@ -856,55 +856,53 @@ namespace Photovoltaics {
...
+		thisSurface = PVarray( PVnum ).SurfacePtr;
+
+		if ( FirstTimeThru ) {
+			PVarrayZoneInitialized.dimension( NumPVs, false );
+			FirstTimeThru = false;
+		}
+
+		if ( ! PVarrayZoneInitialized( PVnum ) ) {
+			if ( Surface( thisSurface ).Zone == 0 ) { // might need to get the zone number from the name
+				Surface( thisSurface ).Zone = FindItemInList( Surface( thisSurface ).ZoneName, Zone, NumOfZones );
```
rraustad: Why isn't this done in GetInput?

RKStrand: @rraustad Good point.  I will change this so that it gets the Zone index directly during GetInput and store it in the local PV data structure.  That should then address what you and Edwin pointed out and still fix the issue.  Thanks!

RKStrand:
@Myoldmopar @rraustad Ok--I have changed the code to reflect both of your comments.  The zone number is now kept in the PV array data structure and it is filled during the GetInput phase rather than during report.  The whole thing is a lot cleaner now--thanks for your comments!  Please let me know if there are any other concerns.  If there aren't any, feel free to accept, merge, etc. at will.

Myoldmopar:
Thanks @RKStrand, I'm pulling and building locally real quick to verify and then I'll merge it in.

Myoldmopar:
Bazinga.  Works a champ.  Thanks @RKStrand.

@lgentile @rpg777 I'm merging this fix in now.  Let me know if you want a build of it right away and I'll help do that, otherwise it'll be available in 8.8 in a few weeks.

Issue #6152, opened 2017-05-24
Add thermostatic fan control to series fan-powered induction unit


mjwitte:
Issue overview
--------------
#5918 and #5879 added thermostatic fan control to parallel fan-powered induction units (AirTerminal:SingleDuct:ParallelPIU:Reheat).

*Addressed in #6285* - This suggestion is to add the same for series fan-powered inductions units (AirTerminal:SingleDuct:SeriesPIU:Reheat.

*Addressed in #6285* - For parallel PIUs, the thermostatic fan control is selected by entering zero for "Fan On Flow Fraction".  Series PIUs have no such field, so a new field will be needed, unless there is consensus that series PIU fans should always be thermostatically controlled and this is a bug-fix.  If a new field is added here, then the same field should probably be added to the parallel PIU.

User description of test file results
---------------------------------------
In the attached 5 zone VAV system example, during unoccupied period, when there is only one PIU zone call for heating in unoccupied period, all PIU fans activated, all zone heating baseboards turned off if PIU reheat coil can meet zone setpoint.

In the attached example file, I have baseboard heater for each zone to provide heating during unoccupied period, they are set to be always available. But for space 2-1, the baseboard heater is intentionally undersized (capacity 345 W) so that at times it will need the PIU to turn on to provide heating during unoccupied period.

Following are two examples.

Example 1: 
At HVAC timestep 441 (8:10), all PIUs are off. Space 2-1 temperature drops below heating setpoint (18.33 C) with baseboard heater at full capacity. All other zones are either at heating setpoint with baseboard heating or in deadband with baseboard at 0 W.

At the next HVAC timestep 442 (8:11:40), PIUs for space 1-1, 2-1 and 3-1 turned on with reheat coil on. The parallel PIU fan for Space 4-1 is off since it’s in deadband. For space 2-1, looks like the PIU heating is not enough to bring zone to heating setpoint, so the baseboard heater is still on. For Space 1-1 and 3-1, PIU reheat coil is enough to maintain zone at heating setpoint, so baseboard heaters are off. This is determined by the fact that we specify baseboard heater as 2nd heating equipment for the zone in the “ZoneHVAC:EquipmentList” object. However, in this situation, it is more logical to keep the baseboard heater running and use PIU as the 2nd heating equipment, since officially it is still in “unoccupied period”. But I understand this might be a separate issue. 

Example 2
[...]

PR #6285, opened 2017-09-01
Thermostatic fan control for series fan-powered induction units - and stabilize night cycle manager during warmup

mjwitte:
Pull request overview
---------------------
Addresses #6152 and fixes potential warmup convergence problems with night cycle managers.

For AirTerminal:SingleDuct:SeriesPIU:Reheat, and controls to run zone fan only if there is a heating load in the zone when the night cycle availability managers calls for CycleOnZoneFansOnly.  Otherwise, for the series PIU the zone fan always runs if there is primary air flow.

For AirTerminal:SingleDuct:ParallelPIU:Reheat, zone fan control logic has been cleaned up so that the fan runs whenever the primary air flow is below the Fan On Fraction or the zone requires reheat.  During night cycle, the fan runs only when there is a heating load in the zone when the night cycle availability managers calls for CycleOnZoneFansOnly.  The previous dependency on "Fan On Fraction" = 0.0 for thermostatic control has been removed.

The changes to the series PIU controls revealed a problem with AvailablityManager:NightCycle during simulation warmup.  If the night cycle manager initial start time move from one warmup day to the next, then convegence is never achieved.  To prevent this, the night cycle manager start/stop times are reset at the beginning of each warmup day.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/PoweredInductionUnits.cc
+++ b/src/EnergyPlus/PoweredInductionUnits.cc
@@ -1381,6 +1381,9 @@ namespace PoweredInductionUnits {
 		Real64 MinSteamFlow;
 		Real64 MaxSteamFlow;
 		Real64 mdot; // local plant fluid flow rate kg/s
+		// Initialize local fan flags to global system flags
+		bool PIUTurnFansOn = ( DataHVACGlobals::TurnFansOn || DataHVACGlobals::TurnZoneFansOnlyOn );  // If True, overrides fan schedule and cycles PIU fan on
+		bool PIUTurnFansOff = DataHVACGlobals::TurnFansOff; // If True, overrides fan schedule and PIUTurnFansOn and cycles PIU fan off
 
 		// FLOW
 
@@ -1422,7 +1425,13 @@ namespace PoweredInductionUnits {
 			if ( ! PriOn ) {
 				// no primary air flow
 				PriAirMassFlow = 0.0;
-				SecAirMassFlow = PIU( PIUNum ).MaxTotAirMassFlow;
+				// PIU fan off if there is no heating load, also reset fan flag if fan should be off
+				if ( QZnReq <= SmallLoad ) {
+					SecAirMassFlow = 0.0;
+					PIUTurnFansOn = false;
+				} else {
+					SecAirMassFlow = PIU( PIUNum ).MaxTotAirMassFlow;
+				}
 			} else if ( CurDeadBandOrSetback( ZoneNum ) || std::abs( QZnReq ) < SmallLoad ) {
 				// in deadband or very small load: set primary air flow to the minimum
 				PriAirMassFlow = PriAirMassFlowMin;
@@ -1439,9 +1448,9 @@ namespace PoweredInductionUnits {
 				Node( SecNode ).MassFlowRate = PIU( PIUNum ).MaxTotAirMassFlow;
 				SimAirMixer( PIU( PIUNum ).MixerName, PIU( PIUNum ).Mixer_Num ); // fire the mixer
 				if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SystemModelObject ) {
-					HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _,_,_,_ );
+					HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _, PIUTurnFansOn , PIUTurnFansOff ,_ );
 				} else if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SimpleConstVolume ) {
-					Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index ); // fire the fan
+					Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index, _, PIUTurnFansOn, PIUTurnFansOff ); // fire the fan
 				}
 
 				FanDeltaTemp = Node( HCoilInAirNode ).Temp - Node( SecNode ).Temp;
@@ -1473,9 +1482,9 @@ namespace PoweredInductionUnits {
 		SimAirMixer( PIU( PIUNum ).MixerName, PIU( PIUNum ).Mixer_Num );
 		// fire the fan
 		if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SystemModelObject ) {
-			HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _,_,_,_ );
+			HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _, PIUTurnFansOn, PIUTurnFansOff, _ );
 		} else if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SimpleConstVolume ) {
-			Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index ); // fire the fan
+			Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index, _, PIUTurnFansOn, PIUTurnFansOff ); // fire the fan
 		}
 		// check if heating coil is off
 		QActualHeating = QToHeatSetPt - Node( HCoilInAirNode ).MassFlowRate * CpAirZn * ( Node( HCoilInAirNode ).Temp - Node( ZoneNode ).Temp );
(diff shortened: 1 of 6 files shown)


mjwitte:
With latest version of defect file.  ParallelPIU fan operation is mostly unchanged. Changes in other zone temps result in the fan cycling at slightly different times in Space4-1 due to the zone hovering right near the heating setpoint.  Space 1-1 has a large negative other equipment load to force it to want heat almost all the time.
image
image

mjwitte:
For the series PIU, the fan operation is completely different.  Space 1-1 has zero baseboard heat, so the PIU fan covers all of the heating load, while Space 3-1 has baseboard heat operating first, so the baseboard covers all of the heating load.
image
image

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1381,6 +1381,9 @@ namespace PoweredInductionUnits {
 		Real64 MinSteamFlow;
 		Real64 MaxSteamFlow;
 		Real64 mdot; // local plant fluid flow rate kg/s
+		// Initialize local fan flags to global system flags
+		bool PIUTurnFansOn = ( DataHVACGlobals::TurnFansOn || DataHVACGlobals::TurnZoneFansOnlyOn );  // If True, overrides fan schedule and cycles PIU fan on
+		bool PIUTurnFansOff = DataHVACGlobals::TurnFansOff; // If True, overrides fan schedule and PIUTurnFansOn and cycles PIU fan off
```
rraustad, line 1386: So first we read the global fan control flags. These are set by avail managers.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1422,7 +1425,13 @@ namespace PoweredInductionUnits {
...
 				// no primary air flow
 				PriAirMassFlow = 0.0;
-				SecAirMassFlow = PIU( PIUNum ).MaxTotAirMassFlow;
+				// PIU fan off if there is no heating load, also reset fan flag if fan should be off
+				if ( QZnReq <= SmallLoad ) {
+					SecAirMassFlow = 0.0;
+					PIUTurnFansOn = false;
+				} else {
+					SecAirMassFlow = PIU( PIUNum ).MaxTotAirMassFlow;
+				}
```
rraustad, line 1434: Then we decide to override the global inside the model ?

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1439,9 +1448,9 @@ namespace PoweredInductionUnits {
 				Node( SecNode ).MassFlowRate = PIU( PIUNum ).MaxTotAirMassFlow;
 				SimAirMixer( PIU( PIUNum ).MixerName, PIU( PIUNum ).Mixer_Num ); // fire the mixer
 				if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SystemModelObject ) {
-					HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _,_,_,_ );
+					HVACFan::fanObjs[ PIU( PIUNum ).Fan_Index ]->simulate( _, PIUTurnFansOn , PIUTurnFansOff ,_ );
 				} else if ( PIU( PIUNum ).Fan_Num == DataHVACGlobals::FanType_SimpleConstVolume ) {
-					Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index ); // fire the fan
+					Fans::SimulateFanComponents( PIU( PIUNum ).FanName, FirstHVACIteration, PIU( PIUNum ).Fan_Index, _, PIUTurnFansOn, PIUTurnFansOff ); // fire the fan
```
rraustad, line 1453: And then pass the new fan control to the fan model.

rraustad: So who's in charge here, the avail managers or the PIU unit? I know why you did this I just need to think through how a PIU fan responds to avail managers.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1643,41 +1652,32 @@ namespace PoweredInductionUnits {
...
-			// Special controls if FanOnFlowFrac = 0.0
+			// Calculate if reheat is needed
 			bool ReheatRequired = false;
-			if (PIU(PIUNum).FanOnFlowFrac <= 0.0) {
-				// Calculate if reheat is needed
-				Real64 qMinPrimary = PriAirMassFlowMin * ( CpAirZn * min( -SmallTempDiff, ( Node( PriNode ).Temp - Node( ZoneNode ).Temp ) ) );
-				if (  qMinPrimary < QToHeatSetPt ) ReheatRequired = true;
-			}
+			Real64 qMinPrimary = PriAirMassFlowMin * ( CpAirZn * min( -SmallTempDiff, ( Node( PriNode ).Temp - Node( ZoneNode ).Temp ) ) );
+			if (  qMinPrimary < QToHeatSetPt ) ReheatRequired = true;
```
rraustad, line 1658: Here you just calculate Q regardless of flow frac. OK, it's what you do with this after that matters.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1643,41 +1652,32 @@ namespace PoweredInductionUnits {
...
-						SecAirMassFlow = PIU( PIUNum ).MaxSecAirMassFlow;
-						PIUTurnFansOn = ( DataHVACGlobals::TurnFansOn || DataHVACGlobals::TurnZoneFansOnlyOn );
-					}
+				// PIU fan off if there is no heating load, also reset fan flag if fan should be off
+				if ( QZnReq <= SmallLoad ) {
+					SecAirMassFlow = 0.0;
+					PIUTurnFansOn = false;
 				} else {
 					SecAirMassFlow = PIU( PIUNum ).MaxSecAirMassFlow;
+					PIUTurnFansOn = ( DataHVACGlobals::TurnFansOn || DataHVACGlobals::TurnZoneFansOnlyOn );
```
rraustad, line 1669: This looks like the load overrides the avail managers, that sounds backwards. If there is no load, turn off. If there is a load, and avail managers say to turn on, then turn on. I guess my question here would be why would the avail manager tell the fan to turn on if there were no load? Because I just wanted to...maybe?

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1643,41 +1652,32 @@ namespace PoweredInductionUnits {
...
 				}
 			} else if ( CurDeadBandOrSetback( ZoneNum ) || std::abs( QZnReq ) < SmallLoad ) {
 				// in deadband or very small load: set primary air flow to the minimum
 				PriAirMassFlow = PriAirMassFlowMin;
-				// PIU fan off if FanOnFlowFrac is 0.0 and reheat is not needed, also reset fan flag if fan should be off
-				if ( ( PIU( PIUNum ).FanOnFlowFrac <= 0.0 ) && ReheatRequired ) {
+				// PIU fan off if reheat is not needed, also reset fan flag if fan should be off
+				if ( ReheatRequired ) {
```
rraustad, line 1675: Now there is no decision, the choice was made prior to here. Just turn on if needed or turn off.

src/EnergyPlus/SystemAvailabilityManager.cc
```diff
@@ -2167,6 +2167,12 @@ namespace SystemAvailabilityManager {
 		static Array1D_bool ZoneCompNCControlType;
 		int CyclingRunTimeControlType; 
 
+		// reset start/stop times at beginning of each day during warmup to prevent non-convergence due to rotating start times
+		if ( WarmupFlag && BeginDayFlag ) {
```
rraustad, line 2171: So what does this do, just do it once?

src/EnergyPlus/PoweredInductionUnits.cc
mjwitte: Because the avail manager signal is global to all zone fans on the same airloop regardless of which zones may need heat.  So, it's like an "and" if the avail manger doesn't send a CycleOnZoneFansOnly, then the fan won't come on, regardless of the zone load.  But if the avail manager says go *and* the zone needs heat, then, yes, fire it up.  If you prefer that the availability manager be king - then we could change CycleOnZoneFansOnly to an array by zones and let the availability manager set this zone-by-zone.

src/EnergyPlus/PoweredInductionUnits.cc
```diff
@@ -1702,12 +1702,8 @@ namespace PoweredInductionUnits {
...
 				PriAirMassFlow = min( max( PriAirMassFlow, PriAirMassFlowMin ), PriAirMassFlowMax );
 				// check for fan on or off
-				if ( ( PriAirMassFlow > PIU( PIUNum ).FanOnAirMassFlow ) && ( PIU( PIUNum ).FanOnFlowFrac > 0.0 ) ) {
-					SecAirMassFlow = 0.0; // Fan is off; no secondary air
-					PIUTurnFansOn = false;
-				} else if ( ( PIU( PIUNum ).FanOnFlowFrac <= 0.0 ) && !ReheatRequired ) {
-					// if FanOnFlowFrac is 0, then fan does not run for cooling load unless reheat is required, also reset fan flag if fan should be off
-					SecAirMassFlow = 0.0; // Fan is off; no secondary air
+				if ( ( PriAirMassFlow > PIU( PIUNum ).FanOnAirMassFlow ) && !ReheatRequired ) {
+					SecAirMassFlow = 0.0; // Fan is off unless reheat is required; no secondary air; also reset fan flag
```
rraustad, line 1706: This just combines logic that does the same thing.

src/EnergyPlus/SystemAvailabilityManager.cc
mjwitte: It will reset once for each warmup day.  Keeps the start time from moving and messing up the convergence checks.

src/EnergyPlus/PoweredInductionUnits.cc
mjwitte: There's a subtle difference here.  Previoulsly FanOnFlowFrac has to be zero for the reheat check to happen.  Now that dependency is gone.  Plus if for some reason, the user sets FanOnFlowFrac to be less than the Primary flow min Frac (meaning the primary flow will never be low enough to trip on the fan) the fan will still come on if reheat is required now.

src/EnergyPlus/SystemAvailabilityManager.cc
rraustad: Finished review of code changes. The only question I have is who is king when controlling the fan? Avail manager or load?

rraustad: I'm just trying to think through what the avail managers should do. Lets take this one for example.

    AvailabilityManager:ScheduledOn

If I schedule that fan on should it turn on regardless of load?

If avail manager says to turn on fan, then the fan turns on. And if told to turn off, who cares what the load is ... turn off. I don't think any other model overrides the avail managers.

rraustad: I just looked at the title again. Thermostatic fan control ... so where is that choice? How do I choose Tstat fan control or ? I don't see any IDD changes.

src/EnergyPlus/PoweredInductionUnits.cc
rraustad: OK, so the PIUs don't have avail managers like the other zone equipment. So yes, it's a global flag.

    ZoneHVAC:FourPipeFanCoil,
       A15, \field Availability Manager List Name
          \note Enter the name of an AvailabilityManagerAssignmentList object.

But the revised documentation says:

    The secondary fan will run only when the Availability Schedule specified in the fan object is >0 or when it is overridden by an availability manager

I'm not saying it should or shouldn't act one way or another, and I think this is what the unit would do in reality, but what would a user say if they set an avail manager and the fan didn't turn on or off as they programmed? Maybe just a revision to the avail manager docs to say if no load for PIU then secondary fans won't turn on?

rraustad:
Can you add the avail manager signal to those plots above? They look good now, but I'm not sure what I would think if avail was high and fan was low.

mjwitte:
@rraustad There is no IDD choice for thermostatic control, because I thought we had concluded via e-mail discussion a couple of weeks ago that PIU fans are always tstat controlled.  I understand your concern about a fan disobeying an availability manager, but it's never been clear that a signal to cycle on zone fans should turn on every zone fan en masse. It's too late now to add a new field, so I propose redoing this with plan B where a night cycle manager with  Cycling Run Time Control Type = Thermostat or ThermostatWithMinimumRunTime will send zone-by-zone signals for CycleOnAnyZoneFansOnly.  If the control type is FixedRunTime, then all zone fans on the list would run regardless of load.
```
  AvailabilityManager:NightCycle,
    Reheat System 1 Avail,   !- Name
    ON,                      !- Applicability Schedule Name
    FanAndCoilAvailSched,    !- Fan Schedule Name
    CycleOnAnyCoolingOrHeatingZone,  !- Control Type
    1.0,                     !- Thermostat Tolerance {deltaC}
    **Thermostat,**            !- Cycling Run Time Control Type
    1800,                    !- Cycling Run Time {s}
    ,                        !- Control Zone or Zone List Name
    Reheat System 1 Cooling Zones,  !- Cooling Control Zone or Zone List Name
    East Zone,               !- Heating Control Zone or Zone List Name
    Reheat System 1 PIU Zones;  !- Heating Zone Fans Only Zone or Zone List Name
```

rraustad:
Plan B is fine. But I think regarding control of PIU, this should go in and Plan B can address anything else we need.

rraustad:
Review of the results above for before and after performance of PIUs show PIU fans running more than expected (before results). The basic PIU model needs to respond to zone loads according to basic model requirements (i.e., how and when secondary fan should operate). Advanced controls (i.e., availability managers) can interrupt the control sequence as needed. HOWEVER, E+ needs a working basic model. Only then can advanced model be added, Followup issue to investigate additional controls in #6289.

rraustad:
First step in moving PIU fan controls to a more representative model. Advanced controls to follow.

Issue #6199, opened 2017-07-05
IDF Editor cleanup for v8.8 release


mjwitte:
Issue overview
--------------
- [x] Changes in #6114 result in a new input unit type 
![image](https://user-images.githubusercontent.com/4991075/27887165-d57dc99e-61a4-11e7-95c2-e17c39bb5218.png)

### Details
Some additional details for this issue (if relevant):
 - Platform Windows
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
![image](https://user-images.githubusercontent.com/4991075/29461454-5f9804a0-83f1-11e7-9828-3bdf34b12dd0.png)

More units conflicts.  
@Myoldmopar The W/mK in the phase change PR should be W/m-K - need to fix the IDD for that.

Myoldmopar:
Easy enough, I'll push that fix shortly.

Myoldmopar:
Alright I pushed up my fix, as well as the start of a script that would do this validation for us in CI.  I guess it doesn't catch the "Unit Type for ... " types of fields, but it would be straightforward to add that in.

Myoldmopar:
Feel free to address the other unit issues on that same branch if you like.

mjwitte:
Nice

JasonGlazer:
@Myoldmopar what branch?

Myoldmopar:
It's already in develop, at `scripts/dev/validate_idd_units.py`

JasonGlazer:
Ok, thanks

mjwitte:
Closed via #6291

PR #6291, opened 2017-09-04
IDF Editor - add new Duplicate Object and Change function and support new units

JasonGlazer:
Pull request overview
---------------------
Address #6199 to support new types of units.

New 'Duplicate Object and Change' menu item under Edit and 'Dup Obj + Chg' toolbar item. This will duplicate the currently selected object and for each field in the object replace any text with new text. The original suggested text for change is based on the name of the object. This is especially useful for HVAC related objects when the node names are based on some of the same text as the name of the object.

### Review Checklist


mjwitte:
@JasonGlazer Duplicate and change object function works well.

For Table:TwoIndependentVariables, when Input Unit Type for Y = Wavelength, the units for Min/Max Value of Y should be micron, not dimensionless.

Conversion for W/(m3/s)-Pa to W/(ft3/s)-inH2O looks wrong.  The default of 1.66667 {W/(m3/s)-Pa } * 249.08891 {Pa/inH2O} * 0.028316846592 {m3/s per ft3/s} = 11.75571 {W/(ft3/s-inH2O}

JasonGlazer:
In the conversion, the IP unit is W/((ft3/min)-inH2O). I think you missed the seconds to minute conversion (it is CFM after all). That makes it 0.1959 which is what the conversion factor shows.

Also, that conversion factor was actually added back in March. When you posted the bug you must have been using IDF Editor 1.49 not 1.49a which included that conversion factor and was part of the 8.7 release. Either way, I appreciate you checking.

JasonGlazer:
@mjwitte I fixed the unitsBasedOnField issue with Wavelength.

mjwitte:
Oh - duh.  I need to make sure I have all of my working folders updated with the latest IDF Editor.  I do notice that the IDD conversion factor comment is missing.  I added it with ??? in #6297.

mjwitte:
Everything is good now.

Issue #5672, opened 2016-05-20
EAp2-4/5. Demand data taken from Design Days instead of from Annual Peak


ecoeficiente:
## Issue overview

This happens for Interior Lighting (not checked the rest of the rows) in EAp2-4/5 table:
- Electric demand value seems to be taken from Design Day at time of peak. At least it matches the value in _Peak Heating Sensible Heat Gain Components_ table. Depending on the schedule value at that time (if <1) it will not match the required demand value.
- According to LEED (from a review) demand should match Annual peak value. E+ result from _Anual and Peak Values table_ seems to fulfill this requirement.

It also happens for other rows as I find zero values for Demand when Electricy Use is not zero.

JasonGlazer:
@ecoeficiente do have a specific file that demonstrates this difference that I can use as a test file?

ecoeficiente:
Sorry for the delay, I've missed this among all Github emails I get... I got that in different LEED models which are quite big and not adequate for your tests.
I've randomly picked a file from the examples: _5ZoneBoilerOutsideAirReset.idf_, changing it to annual simulation and I get 9389.21W for space cooling electric demand in EAp2-4/5. That value matches _Demand End Use Components Summary_ Report which I think gathers the peak moment for each use (electricity, natural gas) and then the value for each category (e.g. cooling) which must not necessarily be the peak for that category.
If you go to _Energy Meters_ report you will see 31220.43W for cooling maximum value instead.

Note: I've done this with v8.5

JasonGlazer:
@ecoeficiente I am not getting those numbers, what weather file are you using?

ecoeficiente:
Yes, sorry. https://energyplus.net/weather-location/europe_wmo_region_6/ESP//ESP_Barcelona.081810_IWEC

JasonGlazer:
@ecoeficiente I cannot reproduce your numbers exactly with that file but I get in the EAp2-4/5. Performance Rating Method Compliance table for Space Cooling electric demand of 9319.11 and in the Energy Meters report in the Annual and Peak Values - Cooling a value of 31257.40 for all three "PlantLoop" numbers. I am not sure how these are inconsistent. I think you might be talking about the value in the Annual and Peak Values - Electricity table which for the Cooling:Electricity meter says 9327.56. That is slightly different than the EAp2-4/5 value but not unreasonable since the EAp2-4/5 value is from the peak demand for the entire building and the Cooling:Electricity is the peak for just that specific meter. 

[...]

Issue #5559, opened 2016-03-18
Add LEED reporting of peak end use and EFLH


JasonGlazer:
Add to the LEED report, the peak demand for each end use. It already reports it at the building peak but not individual peak. Also add the equivalent full load hours (EFLH) for each end use and the total peak by dividing the energy consumption by the individual peak. 

See helpdesk #9676

MatthewSteen:
#4588

mjwitte:
Closed via #5672

PR #6295, opened 2017-09-06
Fix EAp2-4/5 table to use individual peak demands instead of part of building peak demand

JasonGlazer:
Addresses #5672 

The LEED Summary report table EAp2-4/5 was reporting the demand for each end use at the time of the peak building demand. GBCI and users of LEED were expecting to see the individual peak demand for each end use separately (which might occur at different times). This fix changes to show the individual peak demands.

Also addresses remaining portion of #5559 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types


mjwitte:
I used RefBldgLargeHotelNew2004_Chicago.idf as a test file.  All of the end-use peaks agree with max values from meter output at the timestep frequency.  Is there a reason that the cells for demand are empty in the "Heating -- Other" row?
screenshot 2

JasonGlazer:
Interesting, I will take a look.

JasonGlazer:
@mjwitte this should address the missing demand values that you found.

mjwitte:
Missing values are there now.

PR #6296, opened 2017-09-06
Build zip and tar files for release

kbenne:
Pull request overview
---------------------
OpenStudio, and perhaps other software would like to make use of non
interactive "tarball" installers. This change is an attempt to request
zip / tar package files on release builds. Windows will receive zip files, 
Linux and Mac are configured for tar.gz.


Myoldmopar:
image

Looks tremendous!  Merging.

Issue #6266, opened 2017-08-23
Zone Volume < 0 fatal error is harsh - seems it should throw error but set it to some value like 10m3 and continue


mjwitte:
Issue overview
--------------
While testing a defect file, it now fails due to a fatal error.
```
   ** Severe  ** Indicated Zone Volume <= 0.0 for Zone=GARAGE PARKING
   ** Severe  ** All ZONE Volumes must be > 0.0
   **  Fatal  ** Errors found in Building Input, Program Stopped
```
There is the potential that many files that have less-than-perfect geometry will now fail where they ran before.  Seems the error should be severe or warning, but not fatal.  The zone volume could be set to something reasonable and let the simulation proceed.

This same file trips on this `assert( n == 4u );` in [DataSurfaces::Surface2D](https://github.com/NREL/EnergyPlus/blob/d945c222ed09925ab3f14e43ae76d990cbade53c/src/EnergyPlus/DataSurfaces.cc#L517).

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus Version 8.7.0-2dd86cbb4d

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

JasonGlazer:
@mjwitte please send me the file that causes this problem.

Myoldmopar:
Closed via #6299

PR #6299, opened 2017-09-08
Change fatal error to warning message when negative zone volume

JasonGlazer:
Previously a severe error and fatal error were thrown if a negative zone volume was reported. It is difficult to get a negative zone volume due to existing checks and when it occurs, something is seriously wrong with the geometry of the surfaces making up the zone. Now a warning message is issued with directions on how to get additional information and the zone volume is set to 10m3.

Note: in order to reproduce this error, I needed to temporarily change the code on lines 486 and 505 of SurfaceGeometry and then modify a zone to have a negative ceiling height and a positive floor area.

Addresses #6266 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/SurfaceGeometry.cc
+++ b/src/EnergyPlus/SurfaceGeometry.cc
@@ -9080,8 +9080,11 @@ namespace SurfaceGeometry {
 			}
 
 			if ( Zone( ZoneNum ).Volume <= 0.0 ) {
-				ShowSevereError( "Indicated Zone Volume <= 0.0 for Zone=" + Zone( ZoneNum ).Name );
-				ShowContinueError( "Zone Volume calculated was=" + RoundSigDigits( Zone( ZoneNum ).Volume, 2 ) );
+				ShowWarningError( "Indicated Zone Volume <= 0.0 for Zone=" + Zone( ZoneNum ).Name );
+				ShowContinueError( "The calculated Zone Volume was=" + RoundSigDigits( Zone( ZoneNum ).Volume, 2 ) );
+				ShowContinueError( "The simulation will continue with the Zone Volume set to 10.0 m3. ");
+				ShowContinueError( "...use Output:Diagnostics,DisplayExtraWarnings; to show more details on individual zones." );
+				Zone( ZoneNum ).Volume =  10.;
 			}
 
 			if ( ShowZoneSurfaces ) {
@@ -9129,15 +9132,6 @@ namespace SurfaceGeometry {
 			} else if ( countNotFullyEnclosedZones > 1 ) {
 				ShowWarningError( "CalculateZoneVolume: " + RoundSigDigits( countNotFullyEnclosedZones ) + " zones are not fully enclosed. For more details use:  Output:Diagnostics,DisplayExtrawarnings; " );
 			}
-		}
-
-		ErrorFlag = false;
-		for ( ZoneNum = 1; ZoneNum <= NumOfZones; ++ZoneNum ) {
-			if ( Zone( ZoneNum ).Volume <= 0.0 ) ErrorFlag = true;
-		}
-		if ( ErrorFlag ) {
-			ShowSevereError( "All ZONE Volumes must be > 0.0" );
-			ErrorsFound = true;
 		}
 
 	}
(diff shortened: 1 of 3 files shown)


JasonGlazer:
@mjwitte @Myoldmopar this one is ready to review.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -9080,8 +9078,11 @@ namespace SurfaceGeometry {
 			}
 
 			if ( Zone( ZoneNum ).Volume <= 0.0 ) {
-				ShowSevereError( "Indicated Zone Volume <= 0.0 for Zone=" + Zone( ZoneNum ).Name );
-				ShowContinueError( "Zone Volume calculated was=" + RoundSigDigits( Zone( ZoneNum ).Volume, 2 ) );
+				ShowWarningError( "Indicated Zone Volume <= 0.0 for Zone=" + Zone( ZoneNum ).Name );
+				ShowContinueError( "The calculated Zone Volume was=" + RoundSigDigits( Zone( ZoneNum ).Volume, 2 ) );
+				ShowContinueError( "The simulation will continue with the Zone Volume set to 10.0 m3. ");
+				ShowContinueError( "...use Output:Diagnostics,DisplayExtraWarnings; to show more details on individual zones." );
```
Myoldmopar, line 9084: I don't see where you added additional diagnostics; I'm assuming those are already there from before?

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -9080,8 +9078,11 @@ namespace SurfaceGeometry {
...
+				Zone( ZoneNum ).Volume =  10.;
```
Myoldmopar, line 9085: As much as I really don't like overwriting the zone volume to a magic number, I understand the rationale.  Shrug.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -2907,11 +2891,120 @@ TEST( SurfaceGeometryUnitTests, CalculateZoneVolume_BoxNoCeilingFloor_test )
...
+	EXPECT_NEAR( 0., SurfaceTmp( surfNum ).Vertex( 3 ).x, 0.001 );
+	EXPECT_NEAR( -5., SurfaceTmp( surfNum ).Vertex( 3 ).y, 0.001 );
+	EXPECT_NEAR( 0., SurfaceTmp( surfNum ).Vertex( 3 ).z, 0.001 );
+
+	EXPECT_NEAR( 0., SurfaceTmp( surfNum ).Vertex( 4 ).x, 0.001 );
+	EXPECT_NEAR( -5., SurfaceTmp( surfNum ).Vertex( 4 ).y, 0.001 );
+	EXPECT_NEAR( 3., SurfaceTmp( surfNum ).Vertex( 4 ).z, 0.001 );
+
+
+}
```
Myoldmopar, line 3009: And so you can't very well test the error handling because there's no way to inject a "valid negative zone volume" into the function you want to test?  Is that right?

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -506,7 +506,7 @@ namespace SurfaceGeometry {
 
 		}
 
-		CalculateZoneVolume( ErrorsFound, ZoneCeilingHeightEntered ); // Calculate Zone Volumes
+		CalculateZoneVolume( ZoneCeilingHeightEntered ); // Calculate Zone Volumes
```
Myoldmopar, line 509: One approach we'd like to move toward is returning the status as the function return value.  Nothing to do for this one, but just thinking about the future, have `calculateZoneVolume` return a boolean where zero is success and non-zero is failure.

Myoldmopar (approved):
Yeah, this one's a bit weird, but I get it.  I'm OK with these changes.

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: Yes, they existed already.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
JasonGlazer: So many tests to prevent getting negative zone volume now that it is very difficult to replicate the problem with unit tests without commenting out some of code. After spending some time I thought it made more sense to just increase unit coverage on the SurfaceGeometry file.

Issue #6106, opened 2017-04-20
Extra Output Variables in v8.7


jonwinkler:
Despite including a Key Value in certain Output:Variable objects, all corresponding output variables are being written (as if an * was included for the key value). This does not occur in v8.6.

This only occurs for certain outputs. See attached IDF and CSV files. For example, the Zone Mean Air Temperature is being written for all zones when it should only be writing it for 'living_1'. This is not happening for all output variables but I've discovered two so far (see output in attached ZIP file).

Attached is ZIP file containing two IDFs and two CSV files. A diff on the IDFs show they're the same file excluding the IDD changes from 8.6 to 8.7. The output from 8.7 has extra unwanted columns. This an issue when running 10,000's of simulations when output file size is a concern.

@nmerket 

[files.zip](https://github.com/NREL/EnergyPlus/files/944660/files.zip)

mjwitte:
@mbadams5

mbadams5:
At first glance, it seems like it is adding a * key value for certain output variables and the culprit is [here](https://github.com/NREL/EnergyPlus/blob/78a111df4adc0ac10b929e2abd28ac9391d46bfa/src/EnergyPlus/InputProcessor.cc#L5480). These are added due to the following in your input file which then causes all those keys to be added internally.
```
Output:Table:SummaryReports,
   AllSummaryAndMonthly;
```

What I haven't figured out yet, is why E+ 8.6 did not have this issue. The new code in E+ 8.7 is working properly, it just "thinks" there are extra Output:Variables added to the IDF. I'll keep looking into this more and try to find a solution.

mjwitte:
@JasonGlazer ?

mjwitte:
@mbadams5 I can't help thinking the problem is that the re2 matching is being applied at the wrong place and it's expanding in the wrong direction when it sees the "*" in `OutputVariablesForSimulation` rather than expanding based on `ReqRepVars`.

mbadams5:
@mjwitte I don't fully understand what you mean when you say `it's expanding in the wrong direction when it sees the "*" in OutputVariablesForSimulation rather than expanding based on ReqRepVars.`
The regex matching is being done in the same exact place as the previous linked list search and as far as I am aware is doing the exact same matching, but obviously something is different.

mjwitte:
@mbadams5 Neither do I - just fishing for an explanation.  I don't fully understand all the output setup functions, but what I see at this point are essentially three passes (figuring this out as I go):

[...]

PR #6303, opened 2017-09-10
Fix extra output variables in ESO/SQL when monthly tabular reports ask for same

JasonGlazer:
Addresses #6106 

When output variables have specific key values and those output variables are also part of monthly tabular reports, all key values were being put into the ESO/SQL (and normal CSV file). This fix makes it so only the specific key values requested will show up in the ESO/SQL file.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
@mbadams5 it looks like it was just a small fix, but it took me a while to figure it out. Could you please take a look and see if you think it would break something else? 

@mjwitte thanks for the detailed suggestions on the issue.

mbadams5:
@JasonGlazer I double checked the [original code back in 8.5.0](https://github.com/NREL/EnergyPlus/blob/c87e61b44bce627bc0fac6f0574d9b018015a3be/src/EnergyPlus/OutputProcessor.cc#L709) and your change is similar to the original functionality but with the expanded regex functionality. I don't remember why I used the `FindItemInVariableList` function there instead but this seems like the correct fix.

JasonGlazer:
@mbadams5 and the RE2 flag of:

`(?i) `

is correct and in the right place? I have not used RE2 before, and I just wanted to verify with you that is correct,

mbadams5:
Yes it is correct and in the right place. It is what I use for [case-insensitive search in another part of the code.](https://github.com/NREL/EnergyPlus/blob/91933e29e35a5e085d8e9ddfe8a7d9ba21939623/src/EnergyPlus/DataOutputs.cc#L121)

JasonGlazer:
Thanks.

JasonGlazer:
The ERR file differences were from IDF files explicitly requesting variables that were not available. These were fixed in the latest commits.

The AUD and MTD file differences are as expected and should reoccur.

JasonGlazer:
Looked at the CI results and the look ok to me.

5ZoneAirCooledWithDOASAirLoop
- AUD (fewer IDF lines, fewer fields with defaults, fewer considered output variables) OK
- Table (fewer IDF object sin object count table, fewer fields with defaults) OK

SurfaceTest_SurfLWR
- AUD (fewer IDF lines, fewer fields with defaults, fewer considered output variables) OK
- Table (one less IDF object in object count table, fewer fields with defaults) OK

5ZoneBoilerOutsideAirReset
- AUD (fewer total and actual RVariables) [ok]
- MTD (different numbered meters) [ok]

5ZoneWaterSystems
- AUD (fewer total and actual RVariables) [ok]
- MTD (different numbered meters) [ok]

SuperMarketDetailed_DesuperHeatingCoil
- AUD (fewer total and actual RVariables) [ok]
- MTD (different numbered meters) [ok]

_SingleZoneTestCondFDTwoDD
- AUD (fewer IDF lines, fewer fields with defaults, fewer considered output variables) OK

@mjwitte or @Myoldmopar please review this, I think it is ready to be merged.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -698,7 +698,7 @@ namespace OutputProcessor {
 		for ( Loop = MinIndx; Loop <= MaxIndx; ++Loop ) {
 			if ( ReqRepVars( Loop ).Key.empty() ) continue;
 			if ( ! SameString( ReqRepVars( Loop ).VarName, VariableName ) ) continue;
-			if ( ! DataOutputs::FindItemInVariableList( KeyedValue, VariableName ) ) continue;
+			if ( !RE2::FullMatch( KeyedValue, "(?i)" + ReqRepVars( Loop ).Key ) ) continue;
```
Myoldmopar, line 701: I'll rely on @mbadams5 agreeing with this change [here](https://github.com/NREL/EnergyPlus/pull/6303#issuecomment-329193945)

tst/EnergyPlus/unit/OutputProcessor.unit.cc
```diff
@@ -2291,6 +2292,122 @@ namespace EnergyPlus {
...
+			NumExtraVars = 0;
+			BuildKeyVarList( "LIVING1", "ZONE TOTAL INTERNAL LATENT GAIN RATE", 1, 2 );
+			EXPECT_EQ( 1, NumExtraVars );
+
+			NumExtraVars = 0;
+			BuildKeyVarList( "GARAGE", "ZONE TOTAL INTERNAL LATENT GAIN RATE", 1, 2 );
+			EXPECT_EQ( 0, NumExtraVars );
+
+
+		}
```
Myoldmopar, line 2407: Tests look good to me

Myoldmopar (approved):
Yep, this looks good.

Issue #6304, opened 2017-09-11
Some Node Reporting Issues


Myoldmopar:
Issue overview
--------------
Some things were noted during the review of #6157.

- Only report these for outdoor air nodes:
  - System Node Wind Direction
  - System Node WInd Speed
  - Surface Outside Normal Azimuth Angle   -- I think this can stay in actually even for static surfaces.
- Fix units on Surface Outside Normal Azimuth Angle
- Wind Speed doesn't vary with height -- did you verify this with a test file that includes the height variation input object?

Myoldmopar:
@mjwitte Outside Air looks to be a node connection type attribute, not a single node attribute.  That will make it a little more difficult to trim down to only report the wind direction/speed for outdoor air nodes.  I can easily trim it down to just air nodes so we don't get water nodes showing wind speed, because fluid type is an attribute on a single node.  What do you think?

Myoldmopar:
But really, once I pull wind speed and direction out for just air nodes, I should also pull dew point and wet bulb too.  This is feeling more like a "I shouldn't try to touch these at the last minute sort of thing."

mjwitte:
@Myoldmopar Perhaps you could move the SetUpOutput calls for wind speed and direction from the general node section to the getinputs for OutdoorAir:Node and OutdoorAir:NodeList and set them up there?  I agree - don't mess with the wet bulb etc.  That's been that way for a long time.  

Here's a test file.  It has 3 new outdoor air node objects and a new surface that's at about 100m names "ZN001:WALL001:100MUP" so you can see approximately what wind speed to expect at that height.  This also has output for Surface Outside Normal Azimuth Angle.
```
OutdoorAir:Node, Test OA Node;
OutdoorAir:Node, Test OA Node 100m, 100;
OutdoorAir:Node, Test OA Node 1m, 1;
```
[1ZoneDataCenterCRAC_wPumpedDXCoolingCoil-MoreOutput.idf.txt](https://github.com/NREL/EnergyPlus/files/1293258/1ZoneDataCenterCRAC_wPumpedDXCoolingCoil-MoreOutput.idf.txt)

mjwitte:
Closed via #6305

PR #6305, opened 2017-09-11
Fix Node and Reporting Issues

Myoldmopar:
Pull request overview
---------------------
Addresses #6304 

- Calculate outdoor air node wind speed as a function of height, just like DB/WB were already done.
- Change output for surface azimuth variable units to degrees


Myoldmopar:
@mjwitte I took the multistory file and was able to reproduce the same wind speed at each location.  With my fix they are now different, with a higher wind speed at the higher elevation:

image

mjwitte:
@Myoldmopar Marking this donotpublish since both of these fixes are for new features that haven't been released yet.  Code and unit test all look good.  Built and tested. Merging.

PR #6314, opened 2017-09-14
What if we add lcov ignore to fatal calls?

Myoldmopar:
This is a "what if" scenario.  We don't actively add unit/integration test support for the fatal error code paths.  There are probably a bunch in the code that really are "unreachable" assuming the IDD validation guards it.  I've long wanted to evaluate how much actual, meaningful code our tests are covering with the fatals in place.  I'll probably do a couple steps:
 - Just the ShowFatalError code lines
 - Also ShowSevereError code lines - since these _should_ result in fatal errors (no severe warnings, right?)
 - Then finally, try to match up the ShowContinueError lines that occur just before a ShowFatalError call.

It could certainly be that this doesn't ever get merged in, no problem.  However, it could be valuable.  If we can show that our integration test coverage is up in the 90%s once we ignore the fatal error paths, then it would be easy to target new integration tests to get it up to 100% of the meaningful code.  Which would be super cool.

### DIFF
--- a/src/EnergyPlus/AirflowNetworkBalanceManager.cc
+++ b/src/EnergyPlus/AirflowNetworkBalanceManager.cc
@@ -814,7 +814,7 @@ namespace AirflowNetworkBalanceManager {
 		}
 
 		if ( ErrorsFound ) {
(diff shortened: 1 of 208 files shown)


Myoldmopar:
Closing the PR but I'll leave the branch here to investigate this later.

Issue #6271, opened 2017-08-24
Fix Severe Errors and Warnings in Test Suite IDF Files


mitchute:
I ran the test suite files to get an idea how many severe errors or warnings are occurring currently in the test suite. The results can be found here: https://gist.github.com/mitchute/c4b16cc0807f8c291dddd09e7378181c

Summary:
40 Severe Errors
1487 Warnings

Disclaimer: I didn't run expand objects or any of the other preprocessors, so those files are likely not included in this list, although we may find that some go away once that is resolved. I also only passed the Chicago weather file, so there are also some spurious weather file warnings.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6316

Issue #6273, opened 2017-08-25
DOATToUnitVentilator node connection errors


mjwitte:
Issue overview
--------------
Not sure when this slipped in, but DOATToUnitVentilator is now throwing node connection errors.  Checked all of the DOATo* files and the others are OK.
```
   ** Severe  ** Node Connection Error, Node="ZONE1UNITVENTOAMIXEROUTLETNODE", The same node appears as a non-parent Inlet node more than once.
   **   ~~~   ** Reference Object=ZONEHVAC:UNITVENTILATOR-OA MIXER, Name=TU1
   **   ~~~   ** Reference Object=FAN:CONSTANTVOLUME, Name=ZONE1UNITVENTFAN
   ** Severe  ** Node Connection Error, Node="ZONE2UNITVENTOAMIXEROUTLETNODE", The same node appears as a non-parent Inlet node more than once.
   **   ~~~   ** Reference Object=FAN:CONSTANTVOLUME, Name=ZONE2UNITVENTFAN
   **   ~~~   ** Reference Object=ZONEHVAC:UNITVENTILATOR-OA MIXER, Name=TU2
   ** Severe  ** Node Connection Error, Node="ZONE3UNITVENTOAMIXEROUTLETNODE", The same node appears as a non-parent Inlet node more than once.
   **   ~~~   ** Reference Object=FAN:CONSTANTVOLUME, Name=ZONE3UNITVENTFAN
   **   ~~~   ** Reference Object=ZONEHVAC:UNITVENTILATOR-OA MIXER, Name=TU3
   ** Severe  ** Node Connection Error, Node="ZONE4UNITVENTAIRINLETNODE", The same node appears as a non-parent Inlet node more than once.
   **   ~~~   ** Reference Object=FAN:CONSTANTVOLUME, Name=ZONE4UNITVENTFAN
   **   ~~~   ** Reference Object=ZONEHVAC:UNITVENTILATOR-OA MIXER, Name=TU4
   ** Severe  ** Node Connection Error, Node="ZONE5UNITVENTAIRINLETNODE", The same node appears as a non-parent Inlet node more than once.
   **   ~~~   ** Reference Object=FAN:CONSTANTVOLUME, Name=ZONE5UNITVENTFAN
   **   ~~~   ** Reference Object=ZONEHVAC:UNITVENTILATOR-OA MIXER, Name=TU5
   ************* There were 5 node connection errors noted.
```

### Details
Some additional details for this issue (if relevant):
[...]

PR #6272, opened 2017-08-24
Allow Multiple Air Loops to One Thermal Zone - Part 2 and Fix Zone Air Mass Balance

mjwitte:
Pull request overview
---------------------
 Followup work to complete and clean up #6248. This pull request is focused on the simulation side of multiple airloops - getting rid of the one-return-node/one-airloopnum per zone mindset and revising the whole zone air mass balance to split thing up across multiple return nodes in a zone.

- [x] Address #6264 Followups from Multiple Airloops Part 1 (see checklist there)
- [x] Address #6273 DOATToUnitVentilator node connection errors (never released this way, error introduced in #6248)
- [x] Address #6219 Zone air mass balance is broken.
- [x] Finish implementing options to specify the return air node for heat gain from lights, airflow windows, and refrigeration cases.
- [x] Finish implementing allocation of return air flow to multiple airloops serving the same zone.
- [x] Engineering ref changes.

#### 6 files with big diffs *as of 2017-09-21*
5ZoneAirCooledWithDOASAirLoop and DOASDXCOIL_wADPBFMethod_NoReturnPath - big table diffs, because the outdoor air summary table is correct now (previously zeros)

5ZoneAirCooled_ZoneAirMassFlowBalance and 5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized - big eso diffs etc because zone air mass conservation is fixed here

DOAToWaterToAirHPInlet - These diffs should go away once #6333 drops in.

VSHeatPumpWaterHeater - big table and eio diffs because there is one less warmup day now *as of 09-25 there are now a few more warmup days - but no code changes here just merging in develop*

#### 3 files with err diffs *as of 2017-09-21*
DOAToUnitVentilator - Node connection errors are gone.

HospitalLowEnergy - One fewer SolveAirLoopControllers warning for DOAS1 and DOAS5.

VariableRefrigerantFlow_FluidTCtrl_HR_5Zone - New unbalanced air flow warnings - should go away when #6301 is fixed.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
[...]

PR #6316, opened 2017-09-15
Fix severe errors and some warnings in IDF files

mitchute:
Addresses #6271 

My goal was to tweak the inputs in the IDF files to make the severe errors and warnings go away, which was mostly accomplished with the exception of one minor code change. Regardless, the code change should have no effect on the (numerical) results.

**Severe Errors Remaining**
```DOAToUnitVentilator```
There are a few non-parent inlet node errors which are related to the ZoneHVAC:UnitVentilator object. Someone has [already thought about this](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/UnitVentilator.cc?utf8=%E2%9C%93#L490-L498), but it looks like some work remains left to be done to cleanup the error.

```EMSTestMathAndKill```
This file intentionally divides by zero using an EMS program. This sort of testing should be brought into a unit test, IMO. The same goes for ```1ZoneUncontrolled3SurfaceZone```, which FWIW only throws a warning for a 3 surface zone when it should probably severe and fatal.

```VSHeatPumpWaterToAirWithRHControl```
Heating loads don't converge after 40 days, and there looks to be some odd control behavior. Below is the space-1 air system heating rate for the 40 warmup days (with ReportDuringWarmup turned on).
image

```OutdoorAirUnit```
This file has two Coil:Cooling:Water objects; one fully autosized, the other with user defined numerical inputs. The autosized coil fails to converge on a UA value. I copied the numerical inputs over to the autosized coil and the severe error went away. This looks like there may need to be some investigation into what could be going wrong with autosizing.

So, I guess the question is, what needs to happen to resolve the remaining severe errors? For the "severe" errors which allow the simulation to continue, should they be changed to "warning"s? I think the sense that I've gotten from this group is that "severe" should lead to a fail, and not continue with the simulation. 

can-of-worms-2

**Remaining Warnings**
I cleaned up so, so many unused constructions... I also tweaked convergence tolerances, adjusted timesteps, and modified site locations in a few places to make warnings go away. There are lots of "calculated design load is zero" warnings, some "HVAC limits exceeded" warnings, "CHILLED WATER LOOP Demand Side is storing excess heat the majority of the time" warnings, and temperature or curve out of range warnings remaining.

I'm not sure if I will spend much more time cleaning up warnings at this point, but if there are suggestions on how to address the remaining severe errors, that would be helpful. Those, at least, should be addressed. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@mitchute DOAToUnitVentilator errors were noted in #6273 and the fix is waiting in #6272.  So don't worry about that one.

Myoldmopar:
@mitchute @mjwitte 

@lefticus just merged in some good changes to CI that should start taking effect soon.  One is making more attempts at pushing the files to s3, which will hopefully help the random-ish s3 failures.  I'd suggest over the weekend if CI still looks healthy to make a dummy commit to this branch to get all platforms to retry.

Myoldmopar:
@lefticus Did you expect the s3 CI changes to be running by yesterday?  Or is there something we still need to do on that?  I only ask because this branch hit the s3 errors on the latest run yesterday.  It seems to be a repeatable branch to get those errors, so it may be a spot to debug.

lefticus:
@Myoldmopar depends on if we are talking about documentation uploads or regression uploads. The doc stuff is done by decent_ci and should be fully deployed yes. If it's regression I think there's still an open PR for you with that work

lefticus:
@Myoldmopar look at the regression errors being generated on this build: http://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-0c736af1ca11d92f9d639de67236f3c6051eecfb-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html#build

lefticus:
@Myoldmopar @jasondegraw one last comment: I just looked through all of the builds with successful documentation builds (still don't know why the error is not being reported for doc uploads, but whatever) and see that the only linux system succeeding is `ubuntu-ci-1`. The `decent-ci-ubuntu-14-04` system must have a misconfigured S3 key setups?

Myoldmopar:
@mbadams5 Do you have a Linux box running CI?   ([reason I ask](https://github.com/NREL/EnergyPlus/pull/6316#issuecomment-330276253))

Myoldmopar:
So two left.  The `DOAToUnitVentilator` is being addressed elsewhere, and the `EMSTestMathAndKill` is intentional and fine.  For the remaining two:
 - The severe error in `VSHeatPumpWaterToAirWithRHControl` that appears to be a control problem.  @rraustad do you have any thoughts on that?  Seems like something up your alley.  It may be just an artifact of the control strategy and nothing actually wrong.  If we are going to let the simulation continue to run after this condition is encountered then I propose you change the severe to a warning and leave it as-is for now.
 - Fully autosized coil fails to converge on UA in `OutdoorAirUnit`.  Again, if we are going to use a default value once it can't converge, then I propose we throw it as a warning instead of a severe and carry on.  Changing the inputs is optional at this point, to me.

So I'd go ahead and change those two severes to warnings and try to get this merged in.  Anyone who wants to debate that those severes need to remain severe needs to be ready for me to add a `ShowFatalError` after the severe is thrown.

rraustad:
@Myoldmopar I'll take a quick look at VSHeatPumpWaterToAirWithRHControl.

rraustad:
@Myoldmopar  Re: control issues with VSHeatPumpWaterToAirWithRHControl. The cooling and heating coils are operating at the same time when dehumidification is needed. This is a separate issue.

Myoldmopar:
Thanks @rraustad, we'll leave it in there for now.  The only thing holding up this PR then is the fully autosized coil failing to converge on a UA value.  Do you have any thoughts off the top of your head about that one?

rraustad:
I thought we (I) had fixed that. There must still be a piece of the puzzle missing.

Myoldmopar:
@rraustad Oh, and is the cooling/heating coils both running issue already posted on Github?

@mitchute maybe to wrap this up, make sure these are all posted as Github issues and you'll be all set:
 - Autosized Coil in Example File Fails to Converge on UA
 - Cooling and Heating Simultaneous Operation in Example File Causes Severe Error  ( @rraustad may have already created one )
 - Suggestion: Create CI test to report a failure on any example file with a severe in the error file.  A whitelist can be created for things like EMSTestMathAndKill.  This would be a great way for reviewers to catch issues on new example files.

If those are all created and documented here, I think this can be merged in.  I don't want this to slip and have any conflicts (I don't think it would, but still...).

rraustad:
I don't think I've seen an issue where VS coils are operating at the same time. 

Regarding the cooling coil UA warnings for ` OutdoorAirUnit `, the autosized outlet air temp = 12C and the program resets it to inlet water temp + 0.5C. The difference between the outlet air temp and inlet water temp is too close for RootSolver to find an answer. I don't recall why + 0.5 was used or what a good GUESS is for design outlet air temp above inlet water temp. Obviously this adjustment is too small.

    ** Warning ** SizeWaterCoil: Coil="ZONE4OAUWCCOIL", Cooling Coil has leaving air temperature > entering water temperature.
    **   ~~~   **     Tair,out  =  12.000
    **   ~~~   **     Twater,in = 15.000
    **   ~~~   ** ....coil leaving air temperature will be reset to:
    **   ~~~   **     Tair,out = 15.500
    ** Warning ** In calculating the design coil UA for Coil:Cooling:Water ZONE4OAUWCCOIL
    **   ~~~   ** no apparatus dew-point can be found for the initial entering and leaving conditions;
    **   ~~~   ** the apparatus dew-point is below the coil design inlet water temperature;
    **   ~~~   ** the coil outlet design conditions will be changed to correct the problem.
    **   ~~~   ** The initial design conditions are: Tair,in = 27.6880
    **   ~~~   **                                    Wair,in = 1.309058E-002
    **   ~~~   **                                    Twater,in = 15.0000
    **   ~~~   **                                    Tair,out = 15.5000
    **   ~~~   **                                    Wair,out = 8.000000E-003
    **   ~~~   ** The revised design conditions are: Tair,out = 13.4252
    **   ~~~   **                                    Wair,out = 8.837592E-003
    ** Severe  ** Calculation of cooling coil design UA failed for coil ZONE4OAUWCCOIL
    **   ~~~   **   Bad starting values for UA
    **   ~~~   **  Coil design UA set to 3207.226733 [W/C]

Myoldmopar:
Thanks for another investigation @rraustad.  Sounds like that could definitely use its own issue to try to resolve on a better approach.  @mitchute go ahead and make any new tickets that need to be made and we'll get this one merged.

Myoldmopar:
Thanks for making those @mitchute.  This is good to go, with a much improved set of example files.

mitchute:
@Myoldmopar @rraustad thanks for looking this over!

Issue #5778, opened 2016-07-16
Design day temperature and other schedules do not interpolate to timestep


mjwitte:
## Issue overview

_From user:_
In the attached IDF file, I am using Schedule:Day:Interval objects to define summer designday drybulb, wetbulb, beam and diffuse solar radiation profiles. 

In the Schedule:Day:Interval objects, I have set the "Interpolate to Timestep" to "Yes". However, from the simulation result, I can see that the output values were not interpolated to each simulation time step (10 min), rather they are hourly step functions, see the attached plots.

![diffusesolar](https://cloud.githubusercontent.com/assets/4991075/16892089/32bde232-4acd-11e6-8461-799a5ce0bfa7.PNG)

![directsolar](https://cloud.githubusercontent.com/assets/4991075/16892090/38091d92-4acd-11e6-8318-cade192acf5a.PNG)

![tdb](https://cloud.githubusercontent.com/assets/4991075/16892092/3b617412-4acd-11e6-9701-cb453225a87c.PNG)

![twb](https://cloud.githubusercontent.com/assets/4991075/16892094/40cb22fe-4acd-11e6-8fa8-fddbd0ff9ba2.PNG)

### Workaround
Per comments in #6330, this is technically working as designed, but it will be changed to work as expected (after v8.8).  Until then, a workaround is to use minute-by-minute schedules using Schedule:Day:Interval, Schedule:Day:List, or Schedule:Compact.

### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus (v8.5.0)
- helpdesk ticket number 11368
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Comments from a developer emeritus:
I think there are two sort of problems here. (not sure I have the time to track them down). The schedule:interval is not designed to do what he wants. I tried putting in the Untils on the 1/2 hour but that didn't really help. I used schedule:hourly for some and that wasn't a solution either -- so basically, either the fill of the schedule values is not set up to "interpolate" over the timesteps like weather data is or the design day creation doesn't do that.

Myoldmopar:
Addressed via #6330

PR #6330, opened 2017-09-20
Fix interpolation on interval schedules

JasonGlazer:
Current code has an algorithm to interpolate on Schedule:Day:Interval only when the timesteps and the intervals of the schedules are inconsistent. This pull request addresses this described further in #5778 and interpolates between values when ever the interpolation choice is selected even if the timesteps and intervals are consistent. The approach used is to interpolate on a minute-by-minutes basis and then when the timestep values are chosen they are pulled from that specific minute.

The documentation previously described exactly what was done but the term "interpolation" was interpreted by users to mean that it always interpolated. Given that I am labeling this with both defect and new feature.

This change impacts the input and the documentation will need to be changed.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
Do you feel like this has enough time to get tests, docs, reviewed, and merged by tomorrow afternoon?  If so, cool.  If not, let's change the milestone.

JasonGlazer:
No, we should just change the milestone.

JasonGlazer:
@Myoldmopar and @mjwitte this is ready to be reviewed.

TiejunWu:
I downloaded the code and ran a test with a Schedule:Day:Interval object with the "Linear" interpolate mode. In the simulation, the Schedule:Day:Interval object is used as the drybulb temperature for a designday, I specified a 24 hour schedule with one value for each hour (see attached text file).

From the EnergyPlus output, I can see the following two issues.

1. Simulation timesteps fall in the first hour was not interpolated. I intended to have a periodic temperature profile. To get a periodic temperature profile after interpolation, the last hour value should be used together with the first hour value to interpolate for timesteps in the first hour. And the first hour value should be used together with the last hour value to interpolate for timesteps in the last hour.

2. The interpolation seem to have used the hourly value as the instant value at the end of the hour for the interpolation, rather it should be the average value for that hour. So that the the input hourly value should be considered the instant value at the half hour time. If we check the hourly average output from EnergyPlus for the linear interpolated profile, there would be significant deviation compared to the initial hourly schedule input. I am attaching a few PPT slides to highlight the difference.
[ScheduleDayIntervalObject.txt](https://github.com/NREL/EnergyPlus/files/1412052/ScheduleDayIntervalObject.txt)
[Interpolation of Schedule.pptx](https://github.com/NREL/EnergyPlus/files/1412428/Interpolation.of.Schedule.pptx)

mjwitte:
@TiejunWu I agree with comment 1 - first hour not interpolated.  @JasonGlazer Figuring out what to do at the beginning of the day can be tricky when there is a change in day type or no prior day to draw from - refer to the weather data interpolation method for some clues.

@TiejunWu Comment 2 is not so clear -  does the value in the schedule input at a given time represent an average or an instantaneous value.  With the weather data interpolation, this varies depending on the data type - average for solar data, instantaneous for state data such as temperature.  Schedules don't know what data type they will be used for, and the intervals used in the input may not be uniform.  When interpolating schedules, I think the inputs should represent the value at that time point and then interpolate between them.  If the desired values are the temperature at the half-hour, then they can be entered as such in the schedule object.

TiejunWu:
@mjwitte About comment 2, since in the Schedule:Day:Interval object we are using "Value Until Time" rather than  "Value At Time", it gives the impression that the schedule specifies a step function input in which the value is flat in the interval, in terms of a linear function, it represents the instantaneous value at the mid point of the interval.

JasonGlazer:
1. The first hour is tricky and I decided there was no "right" answer. The Schedule:Day:Interval is used by itself to describe the design day weather but can also be used as part of a chain of days in an annual schedule. Using the last hour value would really not be correct for the second case. For design days, I would suggest using the first timestep to be at 0:01 for the hour (or maybe even 0:00) and set the value you want to start.

2. I guess I disagree. I think it is easier for users to understand that the value they enter is the value at that time that they enter and is being linearly interpolated between those times, between those values.

nrel-bot-3:

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -690,28 +690,34 @@ namespace ScheduleManager {
...
 			// Depending on value of "Interpolate" field, the value for each time step in each hour gets processed:
-			if ( Alphas( 3 ) != "NO" && Alphas( 3 ) != "YES" ) {
+			if ( SameString( Alphas( 3 ), "NO" ) ) {
+				DaySchedule( Count ).IntervalInterpolated = ScheduleInterpolation::No;
+			} else if ( SameString( Alphas( 3 ), "AVERAGE" ) ) {
+				DaySchedule( Count ).IntervalInterpolated = ScheduleInterpolation::Average;
+			} else if ( SameString( Alphas( 3 ), "LINEAR" ) ) {
+				DaySchedule( Count ).IntervalInterpolated = ScheduleInterpolation::Linear;
+			} else {
+				DaySchedule( Count ).IntervalInterpolated = ScheduleInterpolation::No;
```
Myoldmopar: Is there a reason to assign here if you are aborting anyway?

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2991,15 +3020,16 @@ namespace ScheduleManager {
 
 		for ( Count = 1; Count <= NumUntils; ++Count ) {
 			Pos = index( Untils( Count ), "UNTIL" );
+			std::string curUntil = Untils( Count ); //debugging line
```
Myoldmopar: Still needed?

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -3030,15 +3060,35 @@ namespace ScheduleManager {
...
+				if ( totalMinutes == 0 ) totalMinutes = 1; // protect future division
+				if ( Count == 1 ) {
+					StartValue = Numbers( Count ); //assume first period is flat
+					EndValue = Numbers( Count );
+				} else {
+					StartValue = EndValue;
+					EndValue = Numbers( Count );
+				}
+				incrementPerMinute = ( EndValue - StartValue ) / totalMinutes;
+				curValue = StartValue + incrementPerMinute;
```
Myoldmopar, line 3070: Is this within a minute by minute loop?  Shouldn't curValue be equal to the StartValue plus the incrementPerMinute times the number of minutes?  I'll keep looking to figure out what's going on, but leaving this as a placeholder for now.

testfiles/Fault_HumidistatOffset_Supermarket.idf
```diff
@@ -596,7 +596,7 @@
     ON/OFF,                  !- Schedule Type Limits Name
     Through: 12/31,          !- Field 1
     For:AllDays,             !- Field 2
-    Interpolate:Yes,         !- Field 3
+    Interpolate:Average,         !- Field 3
```
Myoldmopar, line 599: This is interesting, I didn't know that `Interpolate:` was inside the `Through:` section so that you could presumably interpolate some sections and not others.  Weird.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -3030,15 +3060,35 @@ namespace ScheduleManager {
...
 					if ( SetMinuteValue( Min, SHr ) ) {
 						ShowSevereError( "ProcessScheduleInput: ProcessIntervalFields, Processing time fields, overlapping times detected, " + ErrContext + '=' + DayScheduleName );
 						ErrorsFound = true;
 						goto UntilLoop_exit;
 					}
-					MinuteValue( Min, SHr ) = Numbers( Count );
-					SetMinuteValue( Min, SHr ) = true;
+					if ( interpolationKind == ScheduleInterpolation::Linear ) {
+						MinuteValue( Min, EHr ) = curValue;
+						curValue += incrementPerMinute;
```
Myoldmopar, line 3082: I see that this block **is** inside a minute loop, so I get why it is only incrementing by incrementPerMinute each step.  Still don't see why it is doing it above.  Maybe it's just because of the diff view and I'm missing something not shown.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -3049,17 +3099,37 @@ namespace ScheduleManager {
...
-					SetMinuteValue( Min, SHr ) = true;
-				}
-				for ( Hr = SHr + 1; Hr <= EHr - 1; ++Hr ) {
-					MinuteValue( _, Hr ) = Numbers( Count );
-					SetMinuteValue( _, Hr ) = true;
-				}
-				for ( Min = 1; Min <= EMin; ++Min ) {
-					MinuteValue( Min, EHr ) = Numbers( Count );
-					SetMinuteValue( Min, EHr ) = true;
+				if ( interpolationKind == ScheduleInterpolation::Linear ) {
```
Myoldmopar, line 3098: Yeah, this seems OK.

Myoldmopar (commented):
I won't mark this approve or deny because I'd like to just discuss some of the comments I made.  Let me know if there is any confusion in them.

src/EnergyPlus/ScheduleManager.cc
JasonGlazer: Not really, just a programming habit to always set the variable when having a "case" structure.

src/EnergyPlus/ScheduleManager.cc
JasonGlazer: Nope. I will remove.

src/EnergyPlus/ScheduleManager.cc
JasonGlazer: This is just before the beginning of the minute by minute loop. It is just setting the initial value of the curValue which should be the value for minute 1.

src/EnergyPlus/ScheduleManager.cc
JasonGlazer: It is the value at the first minute. If curValue started as StartValue than minute zero (at the top of the hour) and minute one (one minute after the hour) would be the same value.

testfiles/Fault_HumidistatOffset_Supermarket.idf
JasonGlazer: Presumably. I did not test but each interpolate is associated with a day schedule.

JasonGlazer:
@Myoldmopar just a few code fixes as you suggested but also had some merge conflicts with transition that needed fixing. Assuming that the CI looks good, this should be ready to merge.

Myoldmopar:
Yep, this looks good now.  CI is failing on Win64, but it looks completely unrelated.  I'm going to have it rerun those and if they look clean this can go in.

JasonGlazer:
Sounds good.

Myoldmopar:
Well, the original Win64 issue is gone, though in the process, this branch became 29 commits behind develop, so now it shows diffs.  Unrelated to this though.  I'm merging.  Thanks.

Issue #6332, opened 2017-09-20
AirTerminal:SingleDuct:Mixer leaving air conditions invalid when zone equip flow < primary flow for InletSide


mjwitte:
Issue overview
--------------
DOAToWaterToAirHPInlet test file fails in branch #6272 with a negative density during warmup because the Space2-1 mixer outlet node temperature is -279C.  This is happening because the primary air mass flow rate is greater than the mixed air mass flow rate (entering the heat pump), so the mixed outlet enthalpy and humidity calculations are bogus here (more mass flow in the divisor than in the dividend (they should be equal).  

The same thing is happening in develop, just not to the same degree.  Why?  Because in develop, the return node max avail flow rate is set to the zone node max avail which is small because the heat pump is cycling.  In #6272 the return max avail is paired with the primarly inlet node max avail.

Looking at develop, with the defect test file that reports during warmup, you can see here that the space2-1 heat pump inlet node temp is 1207C
![image](https://user-images.githubusercontent.com/4991075/30656507-fae77a18-9df9-11e7-98b1-4815c6a8166e.png)

The proposed fix is to limit primary air flow to be <= the mixed air mass flow (which is set by the zone equipment).  

Will put this up shortly and see what happens to all of the DOATo* files.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus  8.8.0 914052f563

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6333

PR #6333, opened 2017-09-20
Fix DOASMixer primary flow problem and add error trap

mjwitte:
Pull request overview
---------------------
Addresses #6332 by limiting the DOAS mixer primary air flow to be no greater than the mixed air outlet mass flow rate which is set by the zone equipment.  This fixes a problem introduced by earlier DOAS mixer changes for v8.8 ( #6223 and #6252), so this doesn't need to be published.

**Question to reviewers:** Do you think the new fatal error is a good idea or should it just be a one-time warning?  If it happens, the user probably can't do anything to fix it.  Or maybe it should be a debug assert instead?

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types

### DIFF
--- a/src/EnergyPlus/SingleDuct.cc
+++ b/src/EnergyPlus/SingleDuct.cc
@@ -5067,7 +5067,9 @@ namespace SingleDuct {
 				Node( this->PriInNode ).MassFlowRate = max( Node( this->PriInNode ).MassFlowRate, Node( this->PriInNode ).MassFlowRateMin );
 			}
 		}
-
+		if ( this->MixerType == ATMixer_InletSide ) {
+				Node( this->PriInNode ).MassFlowRate = min( Node( this->PriInNode ).MassFlowRate, Node( this->MixedAirOutNode ).MassFlowRate );
+		}
 	}
 
 	void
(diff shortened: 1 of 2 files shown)


mjwitte:
Diffs as expected in DOAToUnitVentilator (variable OA flow controls with inlet side mixer) and DOAToWaterToAirHPInlet (cycling WSHP fans with inlet side mixer).

lgu1234:
@mjwitte @Nigusse  I tested all DOA* files. The fix worked. Then I commented a line and keep an error trap to run all DOA* files again:

		if ( this->MixerType == ATMixer_InletSide ) {
//				Node( this->PriInNode ).MassFlowRate = min( Node( this->PriInNode ).MassFlowRate, Node( this->MixedAirOutNode ).MassFlowRate );
		}
The following example files had a fatal error caught by the error trap:

DOAToFanCoilInlet
DOAToPTAC
DOAToUnitVentilator
DOAToWaterToAirHPInlet

because the mixed node has zero flow and the primary node has a small mass flow rate, except DOAToUnitVentilator.

I think the fatal error is good. I approve this PR.

I have a new question for future consideration. If a DOAs requires an amount of OA (mdot of Primary Node > 0), and a system is off due to no load (mdot of mixed node = 0), do we force the system to pass required OA without coil operation?

Nigusse:
@lgu1234 Yes for the option the primary air system to provide OA evenif the coil is off.

lgu1234 (approved):

PR #6335, opened 2017-09-20
Use local nodes for air flow network calculation 

Myoldmopar:
Pull request overview
---------------------
1. Use local nodes for air flow network calculation
2. Tests and docs for local environmental data for urban context


Myoldmopar:
It looks like https://github.com/NREL/EnergyPlus/commit/8746f3baa056ee89a2e4c37b3421a9cf79c12f55 did the trick.  Assuming it stays green as CI gets around to it, this will go in.

Myoldmopar:
@XuanLuoLBL can you update this PR?
 - [x] Copy the description and title from the old PR into this one
 - [x] Update all PR labels/milestones to be like the old one
 - [x] Reference the old one in this description to point back to the old one so that we can get back to the original conversation if needed.
 - [x] Close the other PR **without merging**.

xuanluo113:
@Myoldmopar Guess `8746f3b` works!

xuanluo113:
#6227 
The original feature PR - Improving Building Energy Simulation in an Urban Context
#6263 
The orignal feature update PR - Use local nodes for air flow network calculation

xuanluo113:
@Myoldmopar Done.

Myoldmopar:
This is good to go in.  Merging.

Issue #6327, opened 2017-09-20
VAV design minimum cooling air flow fraction > 1.0 when system adjusts zone flows


mjwitte:
Issue overview
--------------
- [x] Defect file for #6218 shows new warnings.  Not sure if this is the result of #6248 or #6140 .  Should this warning even be happening.  And if so, do fractions >1.0 make any sense?  And in any case, the units are wrong.
```
   ************* SizeHVACSingleDuct: Potential issue with equipment sizing for AirTerminal:SingleDuct:VAV:NoReheat = "CORE_BOTTOM ATU".
   **   ~~~   ** User-Specified Minimum Cooling Air Flow Fraction of 1.00000 [m3/s]
   **   ~~~   ** differs from Design Size Minimum Cooling Air Flow Fraction of 1.47043 [m3/s]
   **   ~~~   ** This may, or may not, indicate mismatched component sizes.
```
Pretty sure this came is as a result of #6140.

Preliminary look at this says it's because system sizing is using ventilation requirement  so the original cooling vol flow rate has been reduced. But then why is the DesCoolVolFlowMin > the OA requirement which should result in a 1.0 min frac.
One solution is in SingleDuct::SizeSys around line 2116, `MinAirFlowFracDes` should be limited to no more that 1.0.
That's probably useful, but either DesCoolVolFlowMin or Sys.MaxAirVolFlowRate is wrong at this point.

- [x] Also, noticed there are twice as many TermUnitSizing objects as zones when debugging. 
*That's correct.  There is one TermUnitsizing object for each zone inlet node, so if there is zone equipment there will be extra instances of TermUnitSizing.*

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus 8.8.0 9bd43c6688
### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added use the defect file for #6218
 - [x] Ticket added to Pivotal for defect (development team task)
 - [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6338

PR #6338, opened 2017-09-21
Fix VAV min flow frac > 1.0 and a bunch of termunitsizing issues related to multiple airloops

mjwitte:
Pull request overview
---------------------
1. Addresses #6327 to not allow design VAV min flow frac to be >1.0
2. Fix some places in size single duct to use TermUnitFinalZoneSizing as all terminal units should now in order to see the impact of any DesignSpecification:AirTerminal:Sizing inputs.
3. And make sure that VAV min flow is at least the MinOA requirement for the terminal unit, even if DesignSpecification:AirTerminal:Sizing sets the cooling flow to zero. 
4. Many changes throughout sizing functions to make sure that scaling of load vs OA or both flow through properly to system sizing.
5. Move Std 62.1 sizing to use TermUnitFinalZoneSizing.
6. Remove Std 62.1 system-level scaling of zone-level OA flow rates (Voz).
7. Report base Voz with no system adjustments for Zone Minimum Outdoor Air Flow Rates reported for zone sizing in eio output "Zone Sizing Information" and table output "HVAC Sizing Summary - Zone Sensible Cooling (and Heating) subtables.
8. Fix various sizing timing related issues in evap cooler and VRF TUControl.
9. Engineering Ref changes
- [ ] 10. Review VRF coil sizing report diffs noted below.
- [ ] 11. Later - Additional test files to demonstrate that scaling is working as it should.
- [x] 12. Addresses part of #6481 (#6500 addresses the rest)

### Explanation of Diffs
1. 5ZoneAirCooledWithDOASAirLoop
eio diffs: VAV SYS 1 is a recirculation system (no outdoor air) so it *should* have a zero system central design heating load, and the diffs for reheat coil sizing look good, too.
```
- System Sizing Information, VAV SYS 1, Cooling, Sensible, 13220.08, 0.80120, 0.80120, CHICAGO_IL_USA ANNUAL COOLING 1% DESIGN CONDITIONS DB/MCWB, 7/21 00:00:00
- System Sizing Information, VAV SYS 1, Heating, Sensible, 9065.05, 0.49859, 0.49859, CHICAGO_IL_USA ANNUAL HEATING 99% DESIGN CONDITIONS DB, 1/21 01:00:00
- System Sizing Information, DOAS, Cooling, Sensible, 5685.65, 0.16562, 0.26412, CHICAGO_IL_USA ANNUAL COOLING 1% DESIGN CONDITIONS DB/MCWB, 7/21 00:00:00
- System Sizing Information, DOAS, Heating, Sensible, 9206.40, 0.13374, 0.26412, CHICAGO_IL_USA ANNUAL HEATING 99% DESIGN CONDITIONS DB, 1/21 24:00:00
+ System Sizing Information, VAV SYS 1, Cooling, Sensible, 11617.36, 0.80120, 0.80120, CHICAGO_IL_USA ANNUAL COOLING 1% DESIGN CONDITIONS DB/MCWB, 7/21 00:00:00
+ System Sizing Information, VAV SYS 1, Heating, Sensible, 0.00, 0.49859, 0.49859, ,  00:00:00
+ System Sizing Information, DOAS, Cooling, Sensible, 5636.75, 0.26412, 0.26412, CHICAGO_IL_USA ANNUAL COOLING 1% DESIGN CONDITIONS DB/MCWB, 7/21 00:00:00
+ System Sizing Information, DOAS, Heating, Sensible, 9206.40, 0.26412, 0.26412, CHICAGO_IL_USA ANNUAL HEATING 99% DESIGN CONDITIONS DB, 1/21 24:00:00
- Component Sizing Information, AirTerminal:SingleDuct:VAV:NoReheat, SPACE1-1 DOAS AIR TERMINAL, Design Size Constant Minimum Air Flow Fraction, 1.34439
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:NoReheat, SPACE1-1 DOAS AIR TERMINAL, Design Size Constant Minimum Air Flow Fraction, 0.20000
- Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 1.32891E-004
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 1.17091E-004
- Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Reheat Coil Sizing Inlet Air Temperature [C], 12.20000
- Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Reheat Coil Sizing Inlet Air Humidity Ratio [kgWater/kgDryAir], 3.00000E-003
- Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size Rated Capacity [W], 6117.01392
- Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 1.32891E-004
- Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 141.32596
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Reheat Coil Sizing Inlet Air Temperature [C], 16.70000
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, SPACE1-1 VAV REHEAT, Design Size Reheat Coil Sizing Inlet Air Humidity Ratio [kgWater/kgDryAir], 8.00000E-003
+ Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size Rated Capacity [W], 5389.72480
+ Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 1.17091E-004
+ Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 129.77653
```

2. 5ZoneAirCooled_VRPSizing, 5ZoneAirCooled_VRPSizing_MaxZd
eio and table diffs for Zone-level OA flow rates per items 5 and 6 above.

3. 5ZoneVAV-ChilledWaterStorage-Mixed_DCV_MaxZd and 5ZoneVAV-ChilledWaterStorage-Mixed_DCV_MultiPath
eio, eso, and table diffs
Zone MinOA is no longer adjusted by Std62 system-level ratios so this changes the powered induction unit minimum primary air flow rate, and also the reheat coil size.
Annual run of 5ZoneVAV-ChilledWaterStorage-Mixed_DCV_MultiPath shows some changes in annual energy use.
2% more heating gas, 2% decrease in cooling electricity (as related electric). Small change in hours not met (1 more heating not met out of 200, 5 more cooling out of 335).

4. Fault_FoulingEvapCooler_StripMallZoneEvapCooler, StripMallZoneEvapCoolerAutosized, StripMallZoneEvapCooler
table diffs - order of rows in Energy Meters table.
err diffs - Reported severe errors are the same, but they are now counted as 2 severe errors during sizing (because zone equipment is called sooner now in cases without system sizing, if there *was* autosized system equipment in this file, then the errors would be during sizing both before and after this PR).
```
   ** Severe  ** InitZoneEvaporativeCoolerUnit: ZoneHVAC:EvaporativeCoolerUnit = SMSTORE6 ZONE EVAP UNIT
   **   ~~~   ** ...unit fan volumetric flow rate less than evaporative cooler unit design supply air flow rate.
   . . .
   Before:
   ************* EnergyPlus Warmup Error Summary. During Warmup: 0 Warning; 0 Severe Errors.
   ************* EnergyPlus Sizing Error Summary. During Sizing: 0 Warning; 0 Severe Errors.
   ************* EnergyPlus Completed Successfully-- 0 Warning; 2 Severe Errors; Elapsed Time=00hr 00min 28.24sec

   After:
   ************* EnergyPlus Warmup Error Summary. During Warmup: 0 Warning; 0 Severe Errors.
   ************* EnergyPlus Sizing Error Summary. During Sizing: 0 Warning; 2 Severe Errors.
   ************* EnergyPlus Completed Successfully-- 0 Warning; 2 Severe Errors; Elapsed Time=00hr 00min 27.79sec
```

5. House-2FurnaceAC*
eio diffs - just small warmup convergence diffs
table diffs - Std62.1 reports - These files have two furnaces, one sized to 75% of the total load, and the other to 25%. Before this PR,, any zone-level flow rates in the Std 62 reports were the same for both systems which was wrong. Now they are different as expected. System level numbers were correct before and have not changed.

6. MultiStory and MicroCogeneration
rdd/mtd/table diffs - order of rows changed in Energy Meters table report. No change in values. Due to changes in `ManageSizing` to call zone equipment sooner when there is no system sizing.

7. RadLoTempElecTermReheat and RadLoTempHydrTermReheat
table diffs - In the new coil sizing details reports - before this PR, the Terminal Unit Reheat Coil Multiplier column was all -9999. Now there are values on the order of 2 or 3 - assuming they are correct.

8.  RefBldgLargeHotelNew2004_Chicago
eio diffs - VAV min flow fraction is 1.0 now instead of 2+ - the whole reason this PR got started in the first place.
```
- Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, CORRIDOR_FLR_3 VAV BOX COMPONENT, Design Size Constant Minimum Air Flow Fraction, 2.28980
+ Component Sizing Information, AirTerminal:SingleDuct:VAV:Reheat, CORRIDOR_FLR_3 VAV BOX COMPONENT, Design Size Constant Minimum Air Flow Fraction, 1.00000
```
9. HVACTemplate-5ZoneVRF and VariableRefrigerantFlow_5Zone
err diffs - some warnings now counted as sizing warnings due to earlier execution of VRF equipment.

10. VariableRefrigerantFlow_FluidTCtrl_HR_5Zone
err diffs - Small change reported ODB in warnings
```
 ** Warning ** AirConditioner:VariableRefrigerantFlow "VRF HEAT PUMP".
 **   ~~~   ** ...InitVRF: VRF Heat Pump Min/Max Outdoor Temperature in Heat Recovery Mode Limits have been exceeded and VRF heat recovery is disabled.
-   **   ~~~   ** ... Outdoor Dry-Bulb Temperature                       = 27.605
+   **   ~~~   ** ... Outdoor Dry-Bulb Temperature                       = 27.706
    **   ~~~   ** ... Heat Recovery Minimum Outdoor Dry-Bulb Temperature = -20.000
    **   ~~~   ** ... Heat Recovery Maximum Outdoor Dry-Bulb Temperature = 26.000
    **   ~~~   ** ... Check VRF Heat Pump Min/Max Outdoor Temperature in Heat Recovery Mode limits.
-   **   ~~~   **  During Warmup, Environment=MIAMI INTL AP ANN CLG .4% CONDNS DB=>MWB, at Simulation time=07/21 00:00 - 00:15
+   **   ~~~   **  During Sizing, Environment=MIAMI INTL AP ANN CLG .4% CONDNS DB=>MWB, at Simulation time=07/21 24:00 - 24:15
```
table diffs - In the new Coil Sizing Details table, significant differences in VRF terminal unit coil capacity and leaving conditions at "rating conditions". 
**???? Problem ???**

11. DirectIndirectEvapCoolers, DirectIndirectEvapCoolersVSAS, HeatPumpWaterHeater, HeatPumpWaterHeaterStratified, RefBldgSecondarySchoolNew2004_Chicago, RefBldgSmallHotelNew2004_Chicago, and VSHeatPumpWaterHeater
eio diffs - change in time stamp for peak system heating load (no change in value)

12. LookupTables
eio diffs - just a change in order - table curve output shows before zone sizing output now.
table diffs - change in Design Size Constant Minimum Air Flow Fraction (no change in results, because there's a user-specified min frac)

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


src/EnergyPlus/SingleDuct.cc
```diff
@@ -2099,10 +2099,10 @@ namespace SingleDuct {
 		}
 		if ( Sys( SysNum ).ZoneMinAirFracMethod == ConstantMinFrac ) {
 			if ( ZoneSizingRunDone ) {
-				if ( CurZoneEqNum > 0 ) {
+				if ( CurTermUnitSizingNum > 0 ) {
```
Myoldmopar, line 2238: Based on the issue description, I guess this gets the index correctly due to the extra TermUnitSizing objects being stored.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -2099,10 +2099,10 @@ namespace SingleDuct {
...
 					// use the combined defaults or other user inputs stored in DesCoolVolFlowMin
 					if ( Sys( SysNum ).MaxAirVolFlowRate > 0.0 ) {
-						MinAirFlowFracDes = FinalZoneSizing( CurZoneEqNum ).DesCoolVolFlowMin / Sys( SysNum ).MaxAirVolFlowRate;
+						MinAirFlowFracDes = min( 1.0, TermUnitFinalZoneSizing( CurTermUnitSizingNum ).DesCoolVolFlowMin / Sys( SysNum ).MaxAirVolFlowRate );
```
Myoldmopar, line 2241: Makes sense to me.

src/EnergyPlus/SizingManager.cc
```diff
@@ -3768,6 +3768,9 @@ namespace SizingManager {
 					thisTUFZSizing.DesHeatOAFlowFrac = 0.0;
 				}
 				thisTUFZSizing.MinOA = thisFZSizing.MinOA * minOAFrac;
+
+				// Make sure cooling min flow is at least the minOA flow, regardless of the term unit cooling sizing ratio applied above
+				thisTUFZSizing.DesCoolVolFlowMin = max( thisTUFZSizing.DesCoolVolFlowMin, thisTUFZSizing.MinOA );
```
Myoldmopar: This seems like it stomps on things right at the end here.  Is there any chance this could result in a DesCoolVolFlowMin too high and have the same problem (ratio>1.0) downstream somewhere?

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2546,3 +2546,89 @@ TEST_F( EnergyPlusFixture, TerminalUnitMixerInitTest ) {
 	DataHeatBalance::ZoneIntGain.deallocate( );
 
 }
+
+TEST_F( EnergyPlusFixture, VAVReheatTerminal_SizeMinFrac ) {
+	std::string const idf_objects = delimited_string( {
+		"Version,8.4;",
```
Myoldmopar: Version 8.4?  :smiley:

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2546,3 +2546,89 @@ TEST_F( EnergyPlusFixture, TerminalUnitMixerInitTest ) {
...
+
+	int SysNum = 1;
+
+	// First test -  design min flow < max flow
+	ZoneSizingRunDone = true;
+	CurZoneEqNum = 1;
+	CurTermUnitSizingNum = 1;
+	DataSizing::TermUnitFinalZoneSizing( 1 ).DesCoolVolFlowMin = 0.5;
+	SingleDuct::SizeSys( SysNum );
+	EXPECT_EQ( 0.5, SingleDuct::Sys( SysNum ).ZoneMinAirFrac );
```
Myoldmopar, line 2711: Makes sense.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2546,3 +2546,89 @@ TEST_F( EnergyPlusFixture, TerminalUnitMixerInitTest ) {
...
+
+	// Second test -  design min flow > max flow
+	ZoneSizingRunDone = true;
+	CurZoneEqNum = 1;
+	CurTermUnitSizingNum = 1;
+	SingleDuct::Sys( SysNum ).ZoneMinAirFrac = AutoSize; // need to reset this so it sizes again
+	DataSizing::TermUnitFinalZoneSizing( 1 ).DesCoolVolFlowMin = 1.5;
+	SingleDuct::SizeSys( SysNum );
+	EXPECT_EQ( 1.0, SingleDuct::Sys( SysNum ).ZoneMinAirFrac );
```
Myoldmopar, line 2720: Excellent.  This unit test makes me feel very good about these changes.

Myoldmopar (approved):
:+1:

Could you comment on the one question I had about the final calculation that pulls the flow back up to at least minOA?

src/EnergyPlus/SizingManager.cc
mjwitte: Good question.  I added this, because the DOAS system in 5ZoneAirCooledWithDOASAirLoop came back with a zero design min frac because it's airterminal sizing spec is 0 for cooling 1.0 for outdoor air.  

One perspective is that the OA requirement should always be king at this point, so there should be two more lines here to make sure that DesHeatVolFlow and DesCoolVolFlow should be >=MinOA.  But then what about all the other flavors of that?  Will review the base zone sizing calcs to see which of these are checked against OA requirement in FinalZoneSizing.

mjwitte: The other perspective is that the equipment should sort this out and this line should go.  The autosized airflow for this terminal unit was OK even without this line.  It was just the autosized min frac that showed as zero.

tst/EnergyPlus/unit/SingleDuct.unit.cc
mjwitte: What? It works.  Will change if I change the other line.

src/EnergyPlus/SizingManager.cc
mjwitte: @Myoldmopar @rraustad Back to sizing basics.  The zone MinOA requirement impacts the cooling design flow rate during the base zone sizing calcs [here](https://github.com/NREL/EnergyPlus/blame/fc8cc8d371830483fd2306f8a43b3fe2dea1d615/src/EnergyPlus/ZoneEquipmentManager.cc#L2717-L2745).  You can see from the blame that this has been the case since way-back-when.  Interesting that this is only applied to cooling flow, and not heating, but that's how it is.

The way the air terminal sizing spec is working at the moment, the OA requirement will be the floor for the design cooling flow rate, even if a particular air terminal is supposed to deliver a fraction (or zero) of the OA.  That's true with or without the line  that was added here to adjust DesignCoolVolMinFlow.

I'm working on a fix for that now.  I'm going to save some info about the impact of MinOA (like maybe DesignCoolVolFlowNoOA) that honors all the minima *except* MinOA and then use it here with the ratios to set TermUnitFinalZoneSizing.DesignCoolVolFlow appropriately.  Not sure about all the other flavors of \*Cool\* here.

src/EnergyPlus/DataSizing.hh
```diff
@@ -455,6 +455,7 @@ namespace DataSizing {
 		bool EMSOverrideDesHeatMassOn; // true if EMS is acting on this structure
 		Real64 EMSValueDesHeatMassFlow; // Value EMS directing to use for Design Heating air mass flow [kg/s]
 		Real64 DesCoolMassFlow; // zone design cooling air mass flow rate [kg/s]
+		Real64 DesCoolMassFlowNoOA; // zone design cooling air mass flow rate with applying MinOA as a limit [kg/s]
```
mjwitte: Accidentally messed up this commit - it is a mix of new changes plus merging in develop.  Initial display of this commit shows the new changes mostly, plus conflict resolution changes.

Myoldmopar:
This seems good to go from my point of view.  You're done with it right @mjwitte?

mjwitte:
No, not quite done yet.  I don't like the diffs in 5ZoneCoolBeam, so I'll revert those changes and sort that out later.  Also, termunitsizing isn't flowing in correctly yet.

mjwitte:
With this commit, airflow rates are where they should be.  May still be an issue with system OA and reheat coil sizing.  Will check back later.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -2714,13 +2721,34 @@ namespace ZoneEquipmentManager {
...
+				}
+				for ( TimeStepIndex = 1; TimeStepIndex <= NumOfTimeStepInDay; ++TimeStepIndex ) {
+					FinalZoneSizing( CtrlZoneNum ).CoolFlowSeqNoOA( TimeStepIndex ) = FinalZoneSizing( CtrlZoneNum ).CoolFlowSeq( TimeStepIndex );
+					if ( MaxOfMinCoolMassFlowNoOA > FinalZoneSizing( CtrlZoneNum ).CoolFlowSeqNoOA( TimeStepIndex ) ) {
+						FinalZoneSizing( CtrlZoneNum ).CoolFlowSeqNoOA( TimeStepIndex ) = MaxOfMinCoolMassFlowNoOA;
+					}
+				}}
+
+				// Now make sure that the design cooling air flow rates are greater than or equal to the specified minimums including MinOA
+				{ Real64 MaxOfMinCoolVolFlow = 0.0; // max of the user specified design cooling minimum flows and min OA flow [m3/s]
```
rraustad, line 2792: What is the purpose of the open bracket here and close bracket at line 2773?

mjwitte: To limit the context of MaxOfMinCoolVolFlow and make sure it doesn't get confused with the other instance of MaxOfMinCoolVolFlow that was defined earlier without MinOA.

mjwitte:
Made progress on the reheat coil sizing and the overall airloop sizing, but it's still not right if the system flow rate is specified in sizing:system along with airterminal sizing factors.  Close.

src/EnergyPlus/SingleDuct.cc
rraustad: Why would this ever be > 1 ? There is something else going on.

src/EnergyPlus/DataSizing.cc
```diff
@@ -459,6 +459,40 @@ namespace DataSizing {
...
+		Real64 heatFlowRatio = 1.0;
+		if ( this->SpecDesHeatSATRatio > 0.0 ) {
+			heatFlowRatio = this->SpecDesSensHeatingFrac / this->SpecDesHeatSATRatio;
+		} else {
+			heatFlowRatio = this->SpecDesSensHeatingFrac;
+		}
+		Real64 adjustedFlow = heatFlowNoOA * heatFlowRatio + ( heatFlowWithOA - heatFlowNoOA ) * this->SpecMinOAFrac;
+		return adjustedFlow;
+	}
+
```
Myoldmopar, line 578: These methods are so nice.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4990,38 +4990,23 @@ namespace SimAirServingZones {
...
 					int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
-					Real64 coolFlowRatio = 1.0;
-					if ( TermUnitSizing( TermUnitSizingIndex ).SpecDesCoolSATRatio > 0.0 ) {
-						coolFlowRatio = TermUnitSizing( TermUnitSizingIndex ).SpecDesSensCoolingFrac / TermUnitSizing( TermUnitSizingIndex ).SpecDesCoolSATRatio;
-					} else {
-						coolFlowRatio = TermUnitSizing( TermUnitSizingIndex ).SpecDesSensCoolingFrac;
-					}
-					SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinCoolMassFlow += ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesCoolMassFlow * coolFlowRatio / ( 1.0 + TermUnitSizing( TermUnitSizingIndex ).InducRat );
+					Real64 adjCoolMassFlow = TermUnitSizing( TermUnitSizingIndex ).applyTermUnitSizingCoolFlow( ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesCoolMassFlow, ZoneSizing( CurOverallSimDay, CtrlZoneNum ).DesCoolMassFlowNoOA );
+					SysSizing( CurOverallSimDay, AirLoopNum ).NonCoinCoolMassFlow += adjCoolMassFlow / ( 1.0 + TermUnitSizing( TermUnitSizingIndex ).InducRat );
```
Myoldmopar, line 4950: :100:

mjwitte:
The VRP diffs were not expected here.  Investigating.

src/EnergyPlus/SingleDuct.cc
mjwitte: Valid point (why would this ever be > 1?  One example is a hard-sized value for one and autosize for the other that aren't consistent. For the fully autosized case, other fixes in this PR *should* have gotten rid of that, but I haven't confirmed that.

mjwitte:
@Myoldmopar  So, here's the status of this.  In the original implementation of DesignSpecification:AirTerminal:Sizing (back in #6248) the limited set of test files I was using convinced me that the new sizing was working, and it does indeed flow through and, say if you specify a system to serve 50% of the cooling load, the flows and coils come out 50% smaller.  Or, specifying one airloop to serve all of the heating/cooling (and no outdoor air) with a DOAS that serves all of the outdoor air and no heating/cooling - that works now.  

But, I hadn't tested other more complicated combinations with air terminal sizing, like VRP, specified flow in the sizing:system object, sizing on "ventilationrequirement", etc.  Those are what this PR is fixing.  And it's turning out to be a lot of code changes - lots of places where `FinalZoneSizing( CtrlZoneNum )` should be `TermUnitFinalZoneSizing( TermUnitSizingIndex )` which include the air terminal sizing factors.  Still working on this part.

So, the best path forward is probably to review #6272 and get that in.  And if this one proves to be too much to throw in, then I'll add a caveat to the docs explaining what work and what doesn't at this point.

Myoldmopar:
OK, I think given how far we are past the 11th hour, we should probably focus on your best path forward.  Can you add those caveats to the documentation over in #6272 and I'll work on getting that in.  Then we'll be done for 8.8.

nrel-bot-2:

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-2:

nrel-bot-3:

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4632,45 +4673,53 @@ namespace SimAirServingZones {
...
 							}
 						}
+						TermUnitFinalZoneSizing( TermUnitSizingIndex ).SupplyAirAdjustFactor = max( ClgSupplyAirAdjustFactor, HtgSupplyAirAdjustFactor );
+
+						// Because Std62 calcs are currently structured by zone rather than airterminal, need to push information back into FinalZoneSizing
+						FinalZoneSizing( CtrlZoneNum ).VozHtgByZone = TermUnitFinalZoneSizing( TermUnitSizingIndex ).VozHtgByZone;
+						FinalZoneSizing( CtrlZoneNum ).ZpzHtgByZone = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZpzHtgByZone;
+						FinalZoneSizing( CtrlZoneNum ).ZonePrimaryAirFractionHtg = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZonePrimaryAirFractionHtg;
+						FinalZoneSizing( CtrlZoneNum ).ZoneOAFracHeating = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZoneOAFracHeating;
+						FinalZoneSizing( CtrlZoneNum ).SupplyAirAdjustFactor = TermUnitFinalZoneSizing( TermUnitSizingIndex ).SupplyAirAdjustFactor;
```
rraustad: But what if there is 2 air loops serving the same zone ? The last one will set this value.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5342,18 +5382,19 @@ namespace SimAirServingZones {
...
-								Er = FinalZoneSizing( CtrlZoneNum ).ZoneSecondaryRecirculation;
-								Ep = FinalZoneSizing( CtrlZoneNum ).ZonePrimaryAirFraction;
-								ZoneOAFrac = FinalZoneSizing( CtrlZoneNum ).ZpzClgByZone;
-								ZoneEz = FinalZoneSizing( CtrlZoneNum ).ZoneADEffCooling;
-								VozClg = FinalZoneSizing( CtrlZoneNum ).VozClgByZone;
+								Er = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZoneSecondaryRecirculation;
+								Ep = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZonePrimaryAirFraction;
+								ZoneOAFrac = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZpzClgByZone;
+								ZoneEz = TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZoneADEffCooling;
+								VozClg = TermUnitFinalZoneSizing( TermUnitSizingIndex ).VozClgByZone;
```
rraustad, line 5329: These are also used to calculate a zone value, what if 2 air loops?

    Fa = Ep + ( 1.0 - Ep ) * Er;
    FaByZoneCool( CtrlZoneNum ) = Fa;

mjwitte: I know - the whole Std62 calc section doesn't work with more than one airloop/terminal unit in a zone. This needs a big redo but that may have to be yet another pull request.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5779,28 +5759,29 @@ namespace SimAirServingZones {
 						// and so the zone-level Voz must be prorated so their sum equals the system level Vot
 						Ratio = 1.0;
 						if ( VozSumClgBySys( AirLoopNum ) > 0 ) Ratio = VotClgBySys( AirLoopNum ) / VozSumClgBySys( AirLoopNum );
-						for ( ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) {
-							CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum );
-							FinalZoneSizing( CtrlZoneNum ).MinOA = Ratio * FinalZoneSizing( CtrlZoneNum ).VozClgByZone;
+						for ( int ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) {
+							int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
+							TermUnitFinalZoneSizing( TermUnitSizingIndex ).MinOA = Ratio * TermUnitFinalZoneSizing( TermUnitSizingIndex ).VozClgByZone;
 						}
```
mjwitte: @EnergyArchmage @rraustad Question:
The above lines adjust `.MinOA` by the ratio of VotClgBySys/VozSumBySys. This adjustment is made here in `UpdateSysSizing` but not redone in `ManageSystemVentilationAdjustments`. 

`FinalZoneSizing( CtrlZoneNum ).MinOA` is reported in the eio zone sizing data as the calculated outdoor air flow, and it's reported in the HVAC Sizing Summary Table as the Minimum Outdoor Air Flow Rate by zone.
Looking at current develop, 5ZoneAirCooled_VRPSizing, HVAC Sizing Summary, Space 1-1 is currently showing 0.142 for Minimum Outdoor Air Flow Rate. I think this is wrong, because I don't see 0.142 showing up anywhere in the Std 62 report, and the sum across all zones is only 0.621 while the system OA for cooling is 1.0498 for cooling and 0.3149 for heating.
It's 0.142 in v8.8.0 also.

My changes on this branch caused diffs in the HVAC Sizing Summary Minimum Outdoor Air Flow Rate because these changes leave `FinalZoneSizing( CtrlZoneNum ).MinOA` as it was (designspecOA amount divided by smaller of heating and cooling air distribution efficiency.). So, with this branch, it's reporting the same values as Voz-clg (which are the same as Voz-htg).

So, would you agree that we need to recalculate this in `ManageSystemVentilationAdjustments`?
And if there are more than one system serving a zone, then this should report the total OA from all systems to that zone? Oh, but what about zone equipment? That won't get captured here.

rraustad: I would agree that if the answer is wrong now then we need to correct it. @EnergyArchmage does this ring any bells?

EnergyArchmage: The intent in ManageSystemVentilationAdjustments is to only adjust *system* level VRP outcomes.  The zone level ventilation outcomes are already set and done and could not be redone.  Doing so would get circular and you'd have to iterate, so no I don't think it is possible to also recalculate.  The only zone level changes I knowingly made was to take out a Population term in the reported zone Voz values (in Std 62.1 tables) because that is how the standard defines Voz.  I took pains to not change ZoneSizing.MinOA.  

I never did fully understand the ratio and scaling adjustments done in the code you point to.  I think it is a mistake but changing zone-level outcomes was out of scope for that project and I didn't go that far.  For example the code comment below is just not the case.  
"//The system-level Vot will always be larger than the sum of the zone Voz
// and so the zone-level Voz must be prorated so their sum equals the system level Vot"
VRP has the advantage of Vot often comes out to be less than the zone sum (idea is that recirculation of somewhat unused fresh air (from non-critical zones) has a benefit that reduces the system level Vot compared to the simple zone sum. )  Sizing the terminal unit to max of heating and cooling makes a lot of sense but the ratio scaling done there doesn't make sense to me.  When Vot is lower, that will scale the TermUnitFinalZoneSizing.MinOA down, not up.  Also the system level Vot is not very accurate at this point in the calling order so doing powerful things with it is scary.

EnergyArchmage:
Looking at the diffs in 5ZoneAirCooled_VRPSizing between your branch and develop. 
I think you have fixed reporting problems and the diffs in the eio and html are expected.  

But the fact that FinalZoneSizing.MinOA values were not correct before, and now are, points out that errors still being made in 
https://github.com/NREL/EnergyPlus/blob/VAVDesignMinFlow6327/src/EnergyPlus/SimAirServingZones.cc#L5764 and https://github.com/NREL/EnergyPlus/blob/VAVDesignMinFlow6327/src/EnergyPlus/SimAirServingZones.cc#L5784

Where instead of FinalZoneSizing.MinOA the TermUnitFinalZoneSizing.MinOA values are getting skewed by the ratio.  I suspect that scaling should be removed.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5779,28 +5759,29 @@ namespace SimAirServingZones {
...
+						int NumZonesHeated = AirToZoneNodeInfo( AirLoopNum ).NumZonesHeated;
 						NumZonesForHtg = NumZonesHeated;
 						if ( NumZonesHeated == 0 ) NumZonesHeated = NumZonesCooled;
 
 						//The system-level Vot will always be larger than the sum of the zone Voz
 						// and so the zone-level Voz must be prorated so their sum equals the system level Vot
 						Ratio = 1.0;
 						if ( VozSumHtgBySys( AirLoopNum ) > 0 ) Ratio = VotHtgBySys( AirLoopNum ) / VozSumHtgBySys( AirLoopNum );
-						for ( ZonesHeatedNum = 1; ZonesHeatedNum <= NumZonesForHtg; ++ZonesHeatedNum ) {
+						for ( int ZonesHeatedNum = 1; ZonesHeatedNum <= NumZonesForHtg; ++ZonesHeatedNum ) {
```
EnergyArchmage: NumZoneForHtg in the for statement here might need to be NumZonesHeated

mjwitte:
@Myoldmopar @EnergyArchmage @rraustad OK, this is ready for review. Diffs are summarized/explained in the PR overview. Please look this over carefully - questions welcomed. I still need to look for any engineering ref changes that may be needed.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4370,101 +4334,117 @@ namespace SimAirServingZones {
...
-					CtrlZoneNum = AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum );
+				for ( int ZonesCooledNum = 1; ZonesCooledNum <= NumZonesCooled; ++ZonesCooledNum ) {
 					int TermUnitSizingIndex = AirToZoneNodeInfo( AirLoopNum ).TermUnitCoolSizingIndex( ZonesCooledNum );
 					if ( TermUnitSizingIndex == 0 ) {
-						ShowSevereError( "SetUpSysSizingArray: TermUnitSizingIndex = 0 for AirLoop=" + AirToZoneNodeInfo( AirLoopNum ).AirLoopName + ", Zone =" + Zone( AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum ) ).Name );
+						ShowSevereError( "SetUpSysSizingArray: TermUnitSizingIndex = 0 for AirLoop=" + AirToZoneNodeInfo( AirLoopNum ).AirLoopName + ", Zone =" + DataHeatBalance::Zone( AirToZoneNodeInfo( AirLoopNum ).CoolCtrlZoneNums( ZonesCooledNum ) ).Name );
 						ShowFatalError( "This is a defect. Please report this issue." );
 					}
 					if ( SysSizNum > 0 ) {
+						ZoneOAUnc = TermUnitFinalZoneSizing( TermUnitSizingIndex ).TotalOAFromPeople + TermUnitFinalZoneSizing( TermUnitSizingIndex ).TotalOAFromArea; // should not have diversity at this point
```
rraustad, line 4382: How are you splitting up people and area per terminal unit? line 4382 SimAirServingZones

src/EnergyPlus/SizingManager.cc
```diff
@@ -4235,20 +4333,28 @@ namespace SizingManager {
...
 			thisTUFZSizing = thisFZSizing;
+			thisTUFZSizing.ADUName = thisTUSizing.ADUName;
 
 			if( DataSizing::NumAirTerminalSizingSpec > 0 ) {
 				// Apply DesignSpecification:AirTerminal:Sizing adjustments - default ratios are 1.0
+				Real64 minOAFrac = thisTUSizing.SpecMinOAFrac;
+				// Outdoor air
+				thisTUFZSizing.MinOA = thisFZSizing.MinOA * minOAFrac;
+				thisTUFZSizing.TotalOAFromPeople = thisFZSizing.TotalOAFromPeople * minOAFrac;
+				thisTUFZSizing.TotalOAFromArea = thisFZSizing.TotalOAFromArea * minOAFrac;
```
rraustad, line 4355: Line 4355 SizingManager. So these are the total zone OA flow per person and zone, correct (from line 1640 of ZoneEquipmentManager)? And you move that data into each terminal unit sizing array which is then used for Std 62.1. Do you know the TU flow rate and the total zone flow rate here? You could use that ratio to allocate OA to each terminal unit. At the very least, you could divide by number of TU's per zone.

mjwitte: The user is supposed to manage OA allocation with DesignSpecification:AirTerminal:Sizing objects. I hadn't thought about trying to automatically allocate that. If the user puts more than one air terminal in the same zone without any DSATZ objects, then there will be excess OA. Flow rates are known here, so I suppose something could be done automatically for this case.

mjwitte: And just to clarify. TerminalUnitFinalZoneSizing is used for more than Std 62.1 calcs - it's used everywhere for terminal unit sizing, including .MinOA. So, then the question would be if we're going to automatically allocate OA for multiple terminal units without any DesignSpec AT Sizing objects, then do we automatically divide up the supply flow also? If there are no DSATZ objects, then the supply flow rate will be the same for every terminal unit in the same zone. So, that puts us back to dividing by the number of TUs per zone - but *only if nobody in that zone has a DesignSpec AT Sizing object.*

src/EnergyPlus/DataSizing.cc
EnergyArchmage: Yeah, but they are being added to DataSizing.cc what was a data-only file. I am not sure of the design policy but I would have expected such routines to be in SizingManager.cc.  I guess the thinking is that since they are implemented as structure methods then this is the right place.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4476,102 +4456,118 @@ namespace SimAirServingZones {
 					ZoneSA = 0.0;
 					ZonePA = 0.0;
 					Ep = 1.0;
-					if ( FinalZoneSizing( CtrlZoneNum ).ZoneSecondaryRecirculation > 0.0 ) { // multi-path system
+					if ( TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZoneSecondaryRecirculation > 0.0 ) { // multi-path system
 						//Vpz: "Primary" supply air from main air handler served by an oa mixer
-						ZonePA = FinalZoneSizing( CtrlZoneNum ).DesCoolVolFlow;
+						ZonePA = TermUnitFinalZoneSizing( TermUnitSizingIndex ).DesCoolVolFlow;
 						//Vdz: "Discharge" supply air delivered to zone by terminal unit
 						ZoneSA = max( TermUnitSizing( TermUnitSizingIndex ).AirVolFlow, ZonePA );
```
EnergyArchmage, line 4463: maybe this instance of TermUnitSizing should now be TermUnitFinalZoneSizing, line 4463 SimAirServingZones.cc

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4580,18 +4576,18 @@ namespace SimAirServingZones {
...
+								if ( TermUnitFinalZoneSizing( TermUnitSizingIndex ).ZoneSecondaryRecirculation > 0.0 || TermUnitFinalZoneSizing( TermUnitSizingIndex ).DesCoolVolFlowMin <= 0 ) {
 									// multi-path system or VAV Heating airflow max not defined
 									TermUnitSizing( TermUnitSizingIndex ).AirVolFlow *= HtgSupplyAirAdjustFactor;
 								} else {
 									// Single path; Use VAV Heating airflow max as the Vpz in the Zp = Voz / Vpz equations
-									FinalZoneSizing( CtrlZoneNum ).DesHeatVolFlow *= HtgSupplyAirAdjustFactor;
+									TermUnitFinalZoneSizing( TermUnitSizingIndex ).DesHeatVolFlow *= HtgSupplyAirAdjustFactor;
 									// Don't allow the design terminal airflow to be less than the design heating airflow
-									TermUnitSizing( TermUnitSizingIndex ).AirVolFlow = max( TermUnitSizing( TermUnitSizingIndex ).AirVolFlow, FinalZoneSizing( CtrlZoneNum ).DesHeatVolFlow );
+									TermUnitSizing( TermUnitSizingIndex ).AirVolFlow = max( TermUnitSizing( TermUnitSizingIndex ).AirVolFlow, TermUnitFinalZoneSizing( TermUnitSizingIndex ).DesHeatVolFlow );
```
EnergyArchmage, line 4590: More TermUnitSizing vs TermUnitFinalZoneSizing here.  And line 4607 below

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3866,72 +3864,38 @@ namespace SimAirServingZones {
...
-			DataSizing::VdzMinClgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VdzHtgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VdzMinHtgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::ZdzClgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::ZdzHtgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VpzClgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VpzMinClgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VpzHtgByZone.dimension( NumOfZones, 0.0 );
-			DataSizing::VpzMinHtgByZone.dimension( NumOfZones, 0.0 );
+			DataSizing::VdzClgByZone.dimension( NumAirTerminalUnits, 0.0 );
```
EnergyArchmage, line 3874: Unfortunate that all these array variables are named *ByZone when now they should be *ByTerminal or something.

EnergyArchmage:
Pretty massive and important set of changes.  I have reviewed the code changes and don't have any serious comments.  Diffs seem appropriate

I/O Ref entry for DesignSpecification:AirTerminal:Sizing lists some limitations that I think have been addressed here, so the documentation can be revised to remove some caveats there.

EnergyArchmage (approved):

EnergyArchmage:
As for item # 10 in PR description, I have no idea.  I am not up to speed with that VRF flavor.  Presumably it is related to changes at line 7753 and 7767 in HVACVariableRefrigerantFlow.cc  "Garate"  stuff.  seems like unrelated stuff crept in.

src/EnergyPlus/SimAirServingZones.cc
mjwitte: TermUnitSizing holds values that are set in the terminal units, e.g. [here for PIU](https://github.com/NREL/EnergyPlus/blob/VAVDesignMinFlow6327/src/EnergyPlus/PoweredInductionUnits.cc#L1084-L1091). The values *should be* based on TermUnitFinalZoneSizing with any scaling.

src/EnergyPlus/SimAirServingZones.cc
mjwitte: True, I could rename them if you think it's more clear.

Myoldmopar (approved):

Myoldmopar:
Looks good to me. I'm going to hold off on merging to let CI keep running on a few other branches, but this will go in within a couple hours.

PR #6355, opened 2017-10-05
Fix time-stepping and exposed perimeter logic for Kiva foundations

nealkruis:
Work completed under the Big Ladder Software Contribution License Agreement.

Pull request overview
---------------------
This addresses four issues related to Kiva. The first three issues occur when modeling Kiva foundations with exposed perimeter fraction less than 1.0:
1. Unnecessary limitation on total exposed perimeter input (can now equal zero).
1. Applies the correct weighting of exposed perimeter fraction to each wall section when multiple sections exist.
1. Use the individual wall constructions in each Kiva instance (previously just used the footing wall construction defined in the Foundation:Kiva object).

Some diffs are expected for files with basements/crawlspaces where the wall construction was different than the footing construction.

The final issue is related to the time-stepping in Kiva being locked into the zone timestep even when the kiva was being evaluated hourly. Now when evaluated hourly (as is the default) the ground will have a greater thermal mass effect. Big diffs are expected for all files where the [Simulation Timestep](http://bigladdersoftware.com/epx/docs/8-8/input-output-reference/group-advanced-surface-concepts.html#field-simulation-timestep) is set to "Hourly" or allowed to default.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HeatBalanceKivaManager.cc
+++ b/src/EnergyPlus/HeatBalanceKivaManager.cc
@@ -1067,7 +1067,7 @@ void KivaManager::calcKivaInstances()
 	for ( auto& kv : kivaInstances ) {
 		auto& grnd = kv.ground;
 		kv.setBoundaryConditions();
-		grnd.calculate( kv.bcs,DataGlobals::MinutesPerTimeStep*60. );
+		grnd.calculate( kv.bcs, timestep);
 		grnd.calculateSurfaceAverages();
 		kv.reportKivaSurfaces();
 		if ( DataEnvironment::Month == 1 && DataEnvironment::DayOfMonth == 1 && DataGlobals::HourOfDay == 1 && DataGlobals::TimeStep == 1 ) {
--- a/src/EnergyPlus/HeatBalanceKivaManager.cc
+++ b/src/EnergyPlus/HeatBalanceKivaManager.cc
@@ -637,46 +637,6 @@ bool KivaManager::setupKivaInstances()
 				}
 			}
 
-			// Get combinations of wall constructions and wall heights -- each different
-			// combination gets its own Kiva instance. Combination map points each set
-			// of construction and wall height to the associated exposed perimeter and
-			// list of wall surface numbers.
-			std::map<std::pair<int, Real64>,WallGroup> combinationMap;
-
-			if ( wallSurfaces.size() != 0 ) {
-				for ( auto& wl : wallSurfaces ) {
-
-					auto&v = Surfaces( wl ).Vertex;
-					// Enforce quadrilateralism
-					if ( v.size() != 4) {
-						ErrorsFound = true;
-						ShowSevereError( "Foundation:Kiva=\"" + foundationInputs[surface.OSCPtr].name + "\", only quadrilateral wall surfaces are allowed to reference Foundation Outside Boundary Conditions." );
-						ShowContinueError( "Surface=\"" + Surfaces( wl ).Name + "\", has " + General::TrimSigDigits( v.size() ) + " vertices." );
-					}
-
-					// sort vertices by Z-value
-					std::vector<int> zs = {0, 1, 2, 3};
-					sort( zs.begin(),zs.end(),[v]( int a, int b ){return v[a].z < v[b].z;} );
-
-					Real64 perimeter = distance( v[zs[0]], v[zs[1]] );
-
-					Real64 surfHeight = ( v[zs[2]].z + v[zs[2]].z )/2.0 - ( v[zs[0]].z + v[zs[1]].z )/2.0;
-					// round to avoid numerical precision differences
-					surfHeight = std::round( ( surfHeight ) * 1000.0 ) / 1000.0;
-
-					if ( combinationMap.count( {Surfaces( wl ).Construction, surfHeight}) == 0) {
-						// create new combination
-						std::vector<int> walls = {wl};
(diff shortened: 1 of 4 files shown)


nealkruis:
@mjwitte @Myoldmopar These diffs are small and expected. Let me know if you have any questions about these changes.

nealkruis:
@mjwitte @Myoldmopar I'm making a couple more changes here....please hold off before reviewing.

src/EnergyPlus/HeatBalanceKivaManager.cc
```diff
@@ -1067,7 +1067,7 @@ void KivaManager::calcKivaInstances()
 	for ( auto& kv : kivaInstances ) {
 		auto& grnd = kv.ground;
 		kv.setBoundaryConditions();
-		grnd.calculate( kv.bcs,DataGlobals::MinutesPerTimeStep*60. );
+		grnd.calculate( kv.bcs, timestep);
```
shorowit, line 1070: @nealkruis If I understand correctly, Kiva was always running with the E+ timestep even if "hourly" was specified?

nealkruis: @shorowit the calculation was still happening every hour, but the duration used to calculate the dynamics in the finite difference calculation was using the E+ timestep. Does that make sense?

For the current release (8.8), every model should specify "Timestep" for greatest accuracy.

nrel-bot-3:

nrel-bot-3:

Myoldmopar:
@nealkruis I don't have any problem with this going in.  @shorowit have you had a chance to verify the work here?  I only ask since you were the one person to comment on it.

There's some whitespace I'd like to get cleaned up, but I wasn't going to worry about it if this wasn't ready to go in for some other reason.

src/EnergyPlus/HeatBalanceKivaManager.cc
```diff
@@ -762,6 +722,46 @@ bool KivaManager::setupKivaInstances()
...
+			// Get combinations of wall constructions and wall heights -- each different
+			// combination gets its own Kiva instance. Combination map points each set
+			// of construction and wall height to the associated exposed perimeter and
+			// list of wall surface numbers.
+			std::map<std::pair<int, Real64>,WallGroup> combinationMap;
+
+			if ( wallSurfaces.size() != 0 ) {
+				for ( auto& wl : wallSurfaces ) {
+
+					auto&v = Surfaces( wl ).Vertex;
```
Myoldmopar, line 734: Whitespace here, and for the time being, we should probably go ahead and pad the angle brackets in the map declaration a few lines up, just to keep it uniform with existing code.

src/EnergyPlus/HeatBalanceKivaManager.cc
```diff
@@ -1132,6 +1153,7 @@ Real64 KivaManager::getValue(
 	Real64 h = 0.0;
 	Real64 q = 0.0;
 	Real64 Tz = DataHeatBalFanSys::MAT( DataSurfaces::Surface( surfNum ).Zone ) + DataGlobals::KelvinConv;
+	assert(surfaceMap[surfNum].size() > 0);
```
Myoldmopar, line 1156: Is there an actual code path that allows this to happen?  Should it be protected by something more graceful like a severe/fatal?

shorowit:
The timestep bugfix may be related to a Kiva issue I saw in 8.8. I will do some testing on a build that @nmerket will graciously provide to me.

shorowit:
Hmm, my timestep issue seems to be worse with this PR. My 60-min timestep test run still fails, but now my 10-min timestep test run (that works in v8.8) also fails. I'd like to have @nealkruis look at a couple of my files when he has the time.

Myoldmopar:
@shorowit @nealkruis This is an IDD changing PR, and we are basically at code freeze.  Has there been any progress?  If not I'll reassign this toward the next release milestone.

shorowit:
For what it's worth, the IDD change is not critical as we have a way to work around it. (Alternatively, it could go in by itself because it's very small.)

I'm not sure about the rest of the changes.

nealkruis:
The error @shorowit found is not related to this issue. It is separate and can likely be addressed in a separate pull request. The IDD change is minimal--very unlikely that it will affect anyone if it drops in at bug freeze.

@Myoldmopar your call as to how to handle this. I'd suggest we either merge this in as is and start a separate PR to address the other problem, or we keep it open and brush the minor IDD change under the rug until the bug freeze and get all issues addressed at once. It's also possible I'll have a fix before the IDD freeze on Friday.

src/EnergyPlus/HeatBalanceKivaManager.cc
nealkruis: I cannot conceive of a way that this will this assert will be triggered in EnergyPlus. Perhaps in a future unit test implementation, though.

Myoldmopar:
IDD is frozen, like hard freeze, once we get to IO freeze.  We don't even let comment changes in after that.  I'm OK letting this in now if it is ready to go, because the IDD changes won't get another chance.  My only real comments were whitespace related.  And at this point, CI burden is at a priority so I'm actually inclined to let it go in as is (since we will have a new format soon anyway).  I'll merge if I don't hear anything back from anyone soon.

Myoldmopar:
Alright I'm merging this.  @nealkruis please make a follow up issue to address any remaining pieces here that you and @shorowit need to resolve.

PR #6358, opened 2017-10-10
Fix outdoor air system index variable

nealkruis:
Pull request overview
---------------------
Fixes wrong index variable used to find heating coils that cannot be locked out by an economizer. Presents also as an array bounds assertion hard-crash. Problem introduced in #5973.
Also fixes incorrect logic which could result in disabling economizer lockout for all OA system chilled water coils if a single hot water coil was found in any OA system.

Initial work performed under the Big Ladder Contribution License Agreement. Additional changes completed by @mjwitte

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1144,7 +1144,7 @@ namespace SimAirServingZones {
 						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
 						for (OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum) {
 							for (OACompNum = 1; OACompNum <= OutsideAirSys( OASysNum ).NumComponents; ++OACompNum) {
-								CompType = OutsideAirSys( AirSysNum ).ComponentType( OACompNum );
+								CompType = OutsideAirSys( OASysNum ).ComponentType( OACompNum );
 								if (SameString( CompType, "Coil:Heating:Water" )) {
```
mjwitte: @nealkruis The fix looks good.  Would you be willing to fix the space-padding issues in this entire block as well?

nealkruis:
@mjwitte No problem. Did I miss anything?

mjwitte:
Looks good, thanks.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1142,10 +1142,10 @@ namespace SimAirServingZones {
 							}
 						}
 						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
-						for (OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum) {
-							for (OACompNum = 1; OACompNum <= OutsideAirSys( OASysNum ).NumComponents; ++OACompNum) {
-								CompType = OutsideAirSys( AirSysNum ).ComponentType( OACompNum );
-								if (SameString( CompType, "Coil:Heating:Water" )) {
+						for ( OASysNum = 1; OASysNum <= NumOASystems; ++OASysNum ) {
```
mjwitte: @nealkruis @Nigusse This for loop is wrong.  This check should be looking at just one particular OASys, not all of them.  If I'm following correctly, the current code will set `CanBeLockedOutByEcono` to false for every OA system water coil controller in all air loops if it finds even one hot water coil in one OA system anywhere in the model.  The for loop on OASysNum should be removed, and the loop over OACompNum should be moved up around line 1134 and should include a node match check.

mjwitte: And why does [line 1136](https://github.com/NREL/EnergyPlus/blob/7c3bbe48305fc421887df2fd5fb75cfb72e4075b/src/EnergyPlus/SimAirServingZones.cc#L1136) only check for "Coil:Cooling:Water:DetailedGeometry" and "Coil:Heating:Water"?  Coil:Cooling:Water can't be locked out?  Other heating coil types?

mjwitte: Duh - didn't scroll right enough.  Coil:Cooling:Water is included in line 1136, but what about other non-water coils.  And the comment above seems to indicate that the real fix to #5695 should be to lockout the electric and gas heating coils, not to let the hot water coil operate.  This whole section needs to be rethought carefully - when should OA coils be allowed to run?

nealkruis: @mjwitte It seems like there are two issues here:

1. Array bounds hard crash
2. Incorrect implementation of coil lockouts

Depending on the effort of <i>#</i>2, it might fit into this PR. If it's a larger effort, perhaps we should create a new issue and fix it on another branch.

mjwitte: Yeah, this PR should just fix the lockout under the original premise that heating coils in the OA system do not get locked out.  Then I'll consider posting a separate issue to see if this should be changed, or more likely, given more user control.

mjwitte: @nealkruis So, do you want to modify this to move the check up into the right loop?  Or shall I?

nealkruis: @mjwitte I think you have a better idea as to what exactly needs to be done here, but I'm happy to take a stab at it.

nrel-bot-3:
@nealkruis @lgentile it has been 16 days since this pull request was last updated.

Myoldmopar:
@nealkruis Are you able to push this PR along?  I'm trying to make a pass over stale PRs.  If this needs to get tabled for a while, I'd like to close it and reopen whenever the time is right rather than just have it sit and get CI comments on it.  Thanks!

mjwitte:
I'm working on this one. It goes deeper than originally thought and there's a second somewhat related issue. Stay tuned.

nrel-bot-2:

mjwitte:
This is ready for review now.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1123,33 +1121,43 @@ namespace SimAirServingZones {
...
-								if ( WaterCoilNodeNum == ActuatorNodeNum ) {
-									PrimaryAirSystem( AirSysNum ).CanBeLockedOutByEcono( OASysControllerNum ) = false;
+
+						bool nonLockoutCoilFound = false;
+						WaterCoilNodeNum = -1;
+						// added to fix bug issue #5695, if HW coil on outdoor air system, don't lock out during economizing
+						if ( OANum > 0 ) {
+							for ( int OACompNum = 1; OACompNum <= OutsideAirSys( OANum ).NumComponents; ++OACompNum ) {
+								CompType = OutsideAirSys( OANum ).ComponentType( OACompNum );
+								if ( SameString( CompType, "Coil:Heating:Water" ) ) {
```
Myoldmopar, line 1132: OK, this caught me off guard, but I guess this is just inside getinput.  And there's no chance the controls landscape could change *within* a simulation that would necessitate rechecking this; right?

Myoldmopar: And the below switch checks for several coil types, while this only checks one.  I guess that's because this is purely heating.  OK.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1337,9 +1345,9 @@ namespace SimAirServingZones {
 		}
 
 		OANum = GetNumOASystems();
-		for ( OASysNum = 1; OASysNum <= OANum; ++OASysNum ) {
+		for ( int OASysNum = 1; OASysNum <= OANum; ++OASysNum ) {
```
Myoldmopar, line 1348: :+1:

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1337,9 +1345,9 @@ namespace SimAirServingZones {
...
 			NumInList = GetOACompListNumber( OASysNum );
-			for ( OACompNum = 1; OACompNum <= NumInList; ++OACompNum ) {
+			for ( int OACompNum = 1; OACompNum <= NumInList; ++OACompNum ) {
```
Myoldmopar, line 1350: :+1:

Myoldmopar (approved):

Myoldmopar:
I can run some tests tomorrow, but the code changes look good.

rraustad:
Is there a defect file for this one? I ran #5695 but I assume that one worked before. 

I have reviewed and agree that this looks correct. During economizer operation OA heating coils are allowed to operate while OA water cooling coils are not. Any coils, heating or cooling, in the main air loop are allowed to operate.

What if there is a DX cooling coil in the OA system, seems like that coil type could operate during economizer while water cooling coils would not.

mjwitte:
@rraustad There's a defect file in DevSupport\DefectFiles\6358 - there was never a separate issue posted for this.
I tried to maintain the same lockout logic (which coil types can lockout when) as before, but was a little puzzled by the rules.  If we think that should change, then that should probably be a separate issue.

rraustad:
I sort of understand this to be non-controller actuated OA heating coils can operate while economizing (i.e., the elec heating coil that started all this). So this change means any OA heating coil can operate during economizer operation. Water cooling coils would be locked out. Since economizer is meant to reduce cooling loads, this sort of makes sense. The CanBeLockedOutByEcono flag is only used by the water coil controllers. We could use this same flag elsewhere if needed.

    ControllerConvergedFlag = false;
    // if the controller can be locked out by the economizer operation and the economizer is active, leave the controller inactive
    if ( AirLoopControlInfo( AirLoopNum ).EconoActive && PrimaryAirSystem( AirLoopNum ).CanBeLockedOutByEcono( AirLoopControlNum ) ) {
        ControllerConvergedFlag = true;
        continue;
    }

    // For each controller in sequence, iterate until convergence
    while ( ! ControllerConvergedFlag ) {

rraustad:
Defect file 6358 works as expected. I had to change weather to Burlington VT and min OAT for economizer = 1 C to get economizer to activate. The OA heating coil is active during economizer operation.

6358fixed

rraustad:
Both unit tests pass and test economizer lockout flag for cooling and heating coils in air loop vs OA system. This is ready, waiting for one more CI test.

PR #6359, opened 2017-10-11
Groundwork for glhe enhancements

mitchute:
There were a bunch of changes that I ran across while working on the glhe enhancement project that aren't necessarily glhe related. I've pulled those out here to facilitate review. I'm not expecting any diffs.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/Transition/CreateNewIDFUsingRulesV8_9_0.f90
```diff
@@ -0,0 +1,950 @@
...
+                CurArgs=CurVar
+                DO Arg=CurVar,1,-1
+                  IF (OutArgs(Arg) == Blank) THEN
+                    CurArgs=CurArgs-1
+                  ELSE
+                    EXIT
+                  ENDIF
+                ENDDO
+
+              CASE('WINDOWMATERIAL:BLIND:EQUIVALENTLAYER')
```
mjwitte: This case is leftover from the v8.8.0 transition.

tst/EnergyPlus/unit/KusudaAchenbachGroundTemperatureModel.unit.cc
```diff
@@ -58,11 +58,10 @@
 
 using namespace EnergyPlus;
 using namespace EnergyPlus::GroundTemperatureManager;
- 
+
 TEST_F( EnergyPlusFixture, KusudaAchenbachGroundTempModelTest1 )
 {
 	std::string const idf_objects = delimited_string({
-		"Version,8.4;",
```
mjwitte, line 65: Why delete these version statements?  Don't they help date the vintage of the idf snippet

doc/acknowledgments/CMakeLists.txt
```diff
@@ -16,8 +16,8 @@ SET( INCLUDED_TEX
 SET( INCLUDED_IMAGES
 )
 
-SET( SOURCE_FILENAME "acknowledgements" )
-SET( OUTPUT_FILENAME "Acknowledgements" )
+SET( SOURCE_FILENAME "acknowledgments" )
+SET( OUTPUT_FILENAME "Acknowledgments" )
```
mjwitte, line 20: I'm ok with fixing the spelling of the acknowledgments document name, but this will also require changes to the link in the [readme](https://github.com/NREL/EnergyPlus/blob/254f35681ee9ae1b80711faccca7049e8847c5c2/release/readme.html#L122) and a companion issue to change the Help menu links in EP-Launch and IDF Editor.  @jasondegraw has suggested eliminating this document altogether, but I don't know when that might happen.

mitchute: OK. I'll go ahead and create a separate issue for the other changes.

src/Transition/CreateNewIDFUsingRulesV8_9_0.f90
mitchute: Got it. Thanks.

tst/EnergyPlus/unit/KusudaAchenbachGroundTemperatureModel.unit.cc
mitchute: Don't the IDF objects in the unit tests have to undergo transition along with all of the IDF files to stay current with the code? To my mind, it doesn't seem necessary to keep the version number of the original unit test around even as the IDF snippets stay up to date with current code. Maybe we start adding a standard header to each test stating the author and version it when it was created? Or just rely on git blame to show the history. I don't have a horse in this race, so whatever is decided is fine with me.

Myoldmopar: The version numbers in the unit test snippet will almost always mean: "the oldest version of E+ that these inputs were valid, or the version the unit test was added."  As soon as the unit tests fail because the object changes, I'm sure the developer will probably update the version number.  But to what gain?  I don't really see the value of the version number in those.  As long as we are running the unit tests on each commit, we'll know immediately when one fails due to bad inputs so it should be quite obvious what needs to be updated to make it work again.

mbadams5: While I agree in principal, just having a failing unit test could be from any number of issues. Could be bad code, unit test code that needs to change, IDF snippet version needs to be updated, etc. In my opinion, having the Version number in the snippet provides a quick visual reference of what version the IDF snippet is last guaranteed to work. This can provide an easy reference to debug a failing unit test when it is the snippet that needs to be version translated. That said, as long as `process_idf()` is wrapped with `ASSERT_FALSE` then that should be enough to easily and quickly catch IDF snippet input issues.

mjwitte (commented):
Haha.  There's still 2 occurences of Acknowledgement - in the [title](https://github.com/NREL/EnergyPlus/blob/4f32e5141b7e182b9f10e7c408347dd36f537ed9/doc/acknowledgments/acknowledgments.tex#L3-L6).

mjwitte:
@Myoldmopar So, CI is still not posting built docs?

mjwitte (approved):
CI posted built docs this time! Looks good now.

PR #6372, opened 2017-10-24
System sizing adjustments for Std 61.2 VRP - fix and refactor

EnergyArchmage:
Pull request overview
---------------------
This PR is a contribution from Trane.  Set of code changes involves adding a new method for adjusting air system system sizing by mining data from each of the air terminal units attached to the system.  Standard 62.1 system level sizes and table reporting code has been refactored to correct for adjusted system sizes.  The central heating air flow ratio can now be autosized. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


design/FY2018/AirSystemSizingAdjust_Std62_1_refactor.md
```diff
@@ -0,0 +1,97 @@
...
+
+SizingManager.cc, new routine ManageSystemVentilationAdjustments() Redo Standard 62.1 VRP calculations using latest information on zone flows and report to tables.  Collect all summary table reporting one place with organized code.  
+
+SizingManager.cc, new routine DetermineSystemPopulationDiversity(). New routine to determine Pz sum, Ps, and D for each air system for Standard 62.1 calculations. Temporarily stop with system DoingSizing and do a side calculation that marches through all timesteps for each runperiod and sum up the concurrent max occupants in all the zones by air system.  If the input file has no RunPeriod objects, then use the design days.  This factors in all the scheduled detail for an accurate calculation of system population diversity, D.  Current code uses the schedule max for each people object but that cannot handle the coincident air system total unless all the schedules in the building have maximums at the same time of day.
```
mjwitte, line 51: Why rely on the runperiod or design day objects for the diversity calculation?  If the occupancy schedules are seasonal, this could result is different sizing for a full annual simulation vs an early run with selected run periods.  Seems this should always cycle through an entire year of schedules.

design/FY2018/AirSystemSizingAdjust_Std62_1_refactor.md
```diff
@@ -0,0 +1,97 @@
...
+
+DataSizing.cc/.hh add data variables for Standard 62.1 calculations, move variables buried in SimAirServingZones.cc subroutines and put into DataSizing.cc where the variables persist longer and can be accessed. 
+
+SizingManager.cc, new routine ManageSystemSizingAdjustments() - This routine adjusts system sizing outcomes based on how the zone air terminals finish out their sizing. The zone models are executed to trigger their sizing routines. Then the air terminal units data structures are scanned to sum design flow rates. Every air terminal connected to a particular air loop is summed for: 1) minimum heating flow rate, 2) maximum heating flow rate, 3) maximum flow rate, and 4) store zone level flow information for Standard 62.1 calculations, Vpz, Vpz_min, Vdz, and Vdz_min for both cooling and heating. The summed values are used to "Adjust" the system sizing results.  The corrected values are used to autosize the central heating flow ratio, if set to autosize by the user.
```
mjwitte, line 47: I haven't looked at the new code yet, so this may become obvious there, but I'm wondering how this timing fits in if Sizing:System specifies sizing based on VentilationRequirement or a hard value for Flow/System?  "This routine adjusts system sizing outcomes" - does this mean that the current sequence of zone sizing, system sizing, then call the terminal units is still in place, except for the 62.1 calcs?  Then the system sizing is adjusted?  What if the 62.1 calcs require that a particular zone have a higher supply flow (min or max) to meet the zone OA requirement? How does this get communicated back to the terminal unit?

src/EnergyPlus/AirTerminalUnit.hh
```diff
@@ -94,7 +94,8 @@ protected: // Creation
 		airInNodeNum( 0 ),
 		airOutNodeNum( 0 ),
 		zoneIndex( 0 ),
-		zoneNodeIndex( 0 )
+		zoneNodeIndex( 0 ),
+		airLoopNum( 0 )
```
mjwitte: I see I forgot this in #6248.  Also forgot to intialize ctrlZoneInNodeIndex here.

src/EnergyPlus/HVACCooledBeam.cc
```diff
@@ -567,6 +568,12 @@ namespace HVACCooledBeam {
 			OutWaterNode = CoolBeam( CBNum ).CWOutNode;
 			InitComponentNodes( 0.0, CoolBeam( CBNum ).MaxCoolWaterMassFlow, InWaterNode, OutWaterNode, CoolBeam( CBNum ).CWLoopNum, CoolBeam( CBNum ).CWLoopSideNum, CoolBeam( CBNum ).CWBranchNum, CoolBeam( CBNum ).CWCompNum );
 
+			if ( CoolBeam( CBNum ).AirLoopNum == 0 ) { // fill air loop index
+				if ( CoolBeam( CBNum ).CtrlZoneNum > 0 ) {
+					DataZoneEquipment::ZoneEquipConfig( CoolBeam( CBNum ).CtrlZoneNum ).InletNodeAirLoopNum( CoolBeam( CBNum ).AirOutNode );
```
mjwitte: This is not correct (here and in other terminal units).  ZoneEquipConfig.InletNodeAirLoopNum is not indexed by true node numbers.  It's indexed by the number of inlet nodes in the ZoneEquipConfig.  All of the air terminal units should know their ADUNum and the ADUs know their AirLoopNum, so this block shouldn't be necessary.  But there may be some timing issues of when the various terminal unit ADUNums get filled and when AirDistUnit.AirLoopNum gets filled.

mjwitte (commented):
@EnergyArchmage Some intial comments/questions.

design/FY2018/AirSystemSizingAdjust_Std62_1_refactor.md
mjwitte: @EnergyArchmage In ScheduleManager there are functions for ScheduleAnnualFullLoadHours and ScheduleHoursGT1perc which scan through a full year for a given schedule.  These just walk through weekdays, and appear to ignore special days - I don't know if that's required here but it could be added.  Seems like a similar approach could be used here.

mjwitte: I mean these walk through the days of the week, including weekends.

nrel-bot-3:

nrel-bot-3:

EnergyArchmage:
This comment provides discussion of integration test warning and failures for this pull request.  These are expected and usually widespread for this PR.

AUD Diffs in 389 files because of increases in value of  numColumnTag= because of added columns to some tabular summary reports.

Integration test "Failures" of the following are widespread.

AUD diffs
EIO Diffs: 
ESO big diffs
MTR Big Diffs
Table Big diffs


1. Widespread EIO and Table changes.

Additional EIO and summary table content is added for all test files with any AirLoopHVAC.  The following Component Sizing Information is added for each air loop
```
         Sum of Air Terminal Maximum Heating Flow Rates [m3/s]
         Sum of Air Terminal Minimum Heating Flow Rates [m3/s]
         Sum of Air Terminal Maximum Flow Rates [m3/s]
         Adjusted Heating Design Air Flow Rate [m3/s]
         Adjusted Cooling Design Air Flow Rate [m3/s]
         Adjusted Main Design Air Flow Rate [m3/s]
         Calculated Heating Air Flow Ratio []
         User Heating Air Flow Ratio []
```

The Standard 62.1 summary tables have been changed to increase the precision of reported values and extra columns are added to two tables.  

The Object count summary table shows diffs because number of autosizable fields increases for all files that have Sizing:System objects. 

2. Central Cooling Coil Sizing Changes
Files with big ESO and MTR diffs are often caused by central cooling coils getting sized to be larger because a larger design airflow rate is being used.   This leads to larger chilled water flow rates, changes to controller size,  larger chilled water plants etc and signficant diffferences in ESO and MTR results (and slight reductions in hours not met). HVAC energy increases from larger chilled water pumps.  It appears that, in the cases spot checked, the coils were previously being sized with too low of a flow rate and with the changes here are now being sized using the proper flow rate.  The design main flow rates did not change but the values used for cooling coil sizing did change.  These are all Noncoincident sizing method.  Did not find a good explanation for why the previous sizing calculations were not using the full flow rate.  Here are some examples of air flow rates from a handful of integration tests:
 
Filename   |                     Loop Design Supply Air Flow Rate  |     Previous CC Design Size Design Air Flow Rate     |      New CC Design Size Design Air Flow Rate
----------|---------------|------------------|-------------------
5ZoneDetailedIceStorage  |       1.36478               |                 1.18187                       |                         1.36478
5ZoneAirCooled     |             1.04982             |                   0.91307                       |                         1.04982
LgOffVAV             |           41.49864               |                33.64046                        |                       41.49864
WaterSideEconomizer_Integrated  |2.00084         |                       1.81002           |                                     2.00084

3. In RefBldgLargeHotelNew2004_Chicago there are some additional causes for differences in sizing.  The two DOAS air handlers have mixed hard sizes for the air loop design flow rate and autosized coils.  The code uses the hard size for air flow as the design size for the coil sizing which is a correction.  The main VAV WITH REHEAT air handler does have autosize air flow rate   the air handler FLR_3_DOAS_COOLC is sized slightly larger because the adjusted sizes are set upward to match the minimum outdoor air requirement of 2.3788 m3/s.


4. VRP corrections in test file 5ZoneAirCooled_VRPSizing show large differences in the minimum outside air flow, Vot.  The new code is producing lower values for Evz (by zone) because it is producing higher values for Zpz (=Zdz for single path).  The Xs values do not change.  The new code is using a different value for the minimum value for Vpz (or Vdz) when calculating zone outdoor air fraction Zpz = Voz/Vpz_min ( = Zdz = Voz/Vdz_min).  The new method of mining the data from the air terminal component models obtains the value directly from the size of the minimum flow at lowest (30%) damper operation.  The old/current method for Vpz_min is to use the value in FinalZoneSizing.DesCoolVolFlowMin, which in this file and the zones triggering min Evz it is being filled from FinalZoneSizing.DesCoolMinAirFlow2 which itself is based on a default value for flow per area of ZoneSizingInput.DesCoolMinAirFlowPerArea = 0.000762 m3/s-m2 (0.15 cfm/ft).  Tthis is for the input field called Cooling Minimum Air Flow per Zone Floor Area in the Sizing:Zone object and is used here because the cooling flow sizing uses DesignDayWithLimit.  So the old method for Vpz_min was based on a rule of thumb for a cooling flow rate to use as a lower limit for sizing the maximum box flow rate, to ensure that the box is large enough.  However during operation the box can turn down the flow and the VRP cooling calculations should use the minimum box flow for the final size of the VAV box and how the VAV box will actually operate.  The standard says "Vpz is the lowest zone primary airflow value expected" and the examples in "Single-Path Multiple-Zone System Design" (D. Stanke, ASHRAE Journal 2005) clearly uses the minimum damper flow for calculating Zp in Figure 4. The following tables highlight differences between current "master" and "dev" which has the changes in this PR. 

5ZoneAirCooled_VRPSizing comparison of zone VRP related terms.

Zone | master Zpz | dev Zpz |  master Evz |  dev Evz | master  Voz_clg  |    master  Vpz_min | dev  Voz_clg | dev  Vpz_min | 
--------|-------------|----------|---------------|----------|-------------------|-------------|------- |-------
1 |0.826|0.9349|0.426 |  0.3167 | 0.06244866   |   0.0756    |   0.062448667 |    0.066799562
2  |     0.734  |  0.4894  |0.518 |  0.7622 | 0.02759352     | 0.0376    |      0.027593528   |  0.056378826
3  |     0.753   | 0.8758|  0.499|   0.3758|  0.0553864    |    0.0735      |    0.0553864    |     0.063242153
4    |   0.599 |   0.3993|  0.653 |  0.8523 | 0.0248341    |    0.0415      |    0.024834175   |  0.062191524
5   |    0.74   |   1.0   |  0.512  | 0.2516  |0.1028594     |   0.1391     |     0.10285945   |    0.066332525

System level VRP terms for 5ZoneAirCooled_VRPSizing.

master Vps |  dev Vps |  master Xs | dev Xs|  master  Evz_min |  dev  Evz_min |  master Vou | dev Vou|       master  Vot |  dev Vot
------------| ---------------|---------------|--------------|-------|-----------|--------|-----|------|-----
 1.05  |  1.0498 | 0.252   | 0.2516 | 0.425 |      0.2516  |   0.2641   |  0.2641 |   0.6213  | 1.04982

Testing also reveals that there is problem with the VAV air terminal sizing in zone SPACE5-1. The minimum VAV flow of 0.066333 m3/s is less than the required outdoor air of 0.102859 m3/s.  This causes Zpz to be large but gets limited to 1.0 so that Evz = Xs which is extreme.  

The test file should be fixed to serve as a better example. The input file does not use the input field called Minimum Zone Ventilation Efficiency in the DesignSpecification:ZoneAirDistribution object, so the the Evz value is not limited and goes very low (would probably be improved example if it did use this minimum at say 0.65). 

The larger value of Vot increases the size of the outdoor air handler and affects the OA cooling and heating coils by a lot.

nrel-bot-3:
@EnergyArchmage @lgentile it has been 16 days since this pull request was last updated.

src/EnergyPlus/SizingManager.cc
```diff
@@ -770,6 +782,778 @@ namespace SizingManager {
...
+						FinalSysSizing( AirLoopNum ).DesCoolVolFlow =  max( airLoopHeatingMinimumFlowRateSum, FinalSysSizing( AirLoopNum ).DesCoolVolFlow );
+						FinalSysSizing( AirLoopNum ).MassFlowAtCoolPeak = FinalSysSizing( AirLoopNum ).DesCoolVolFlow * DataEnvironment::StdRhoAir ;
+					}
+				} else if ( FinalSysSizing( AirLoopNum ).SizingOption == Coincident ) {
+
+					if ( FinalSysSizing( AirLoopNum ).sysSizeCoolingDominant ) { // use minimum heating flow sum from air terminals 
+						// know that minimum heating flow is a hard minimum regardless of concurrence situation, so make sure that design is at least that high.
+						FinalSysSizing( AirLoopNum ).DesHeatVolFlow = max( airLoopHeatingMinimumFlowRateSum, FinalSysSizing( AirLoopNum ).DesHeatVolFlow );
+						FinalSysSizing( AirLoopNum ).DesMainVolFlow = max( airLoopHeatingMinimumFlowRateSum, FinalSysSizing( AirLoopNum ).DesMainVolFlow );
+						FinalSysSizing( AirLoopNum ).DesCoolVolFlow = FinalSysSizing( AirLoopNum ).DesMainVolFlow;
```
rraustad, line 1091: Cooling is different here than heating and main. Can't recall right now what Cooling and Heating is used for in the air loop simulation. Thought it was always main.

rraustad: That is, different compared to calculation in heating dominant.

src/EnergyPlus/SizingManager.cc
```diff
@@ -770,6 +782,778 @@ namespace SizingManager {
...
+				FinalSysSizing( AirLoopNum ).DesOutAirVolFlow = max( VotClgBySys( AirLoopNum ), VotHtgBySys( AirLoopNum ) );
+				//Fill Xs values for reporting
+				DataSizing::XsBySysCool( AirLoopNum ) = FinalSysSizing( AirLoopNum ).DesOutAirVolFlow / DataSizing::VpsClgBySys( AirLoopNum );
+				DataSizing::XsBySysHeat( AirLoopNum ) = FinalSysSizing( AirLoopNum ).DesOutAirVolFlow / DataSizing::VpsHtgBySys( AirLoopNum );
+
+			} else { // not vrp, zone sum, fill out values that still apply
+				// redo VpzClgSumBySys( AirLoopNum ) with latest values, for reporting
+				DataSizing::VpzClgSumBySys( AirLoopNum ) = 0.0;
+				for ( int zoneNumOnLoop = 1; zoneNumOnLoop <= DataAirLoop::AirLoopZoneInfo( AirLoopNum ).NumZones; ++zoneNumOnLoop  ) {
+					int CtrlZoneNum = DataAirLoop::AirLoopZoneInfo( AirLoopNum ).ActualZoneNumber( zoneNumOnLoop );
```
rraustad, line 1271: This and next for loop are looping through the same zones. Should have combined calculations into one loop.

src/EnergyPlus/SizingManager.cc
```diff
@@ -770,6 +782,778 @@ namespace SizingManager {
...
+						DataSizing::ZdzHtgByZone( CtrlZoneNum ) = min( 1.0, FinalZoneSizing( CtrlZoneNum ).VozHtgByZone / DataSizing::VdzMinHtgByZone( CtrlZoneNum ) );
+					} else { // would divid by zero, so set to max
+						DataSizing::ZdzHtgByZone( CtrlZoneNum )  = 1.0;
+					}
+				} // end loop over zones on air loop to calculate Zdz values
+
+				//Sum Voz values for System Vou, in E+ the Vbz value has now been corrected to remove population Diversity, so we add the term back in here directly to get Vou
+				DataSizing::VouBySys( AirLoopNum ) = 0.0;
+				for ( int zoneNumOnLoop = 1; zoneNumOnLoop <= DataAirLoop::AirLoopZoneInfo( AirLoopNum ).NumZones; ++zoneNumOnLoop  ) {
+					int CtrlZoneNum = DataAirLoop::AirLoopZoneInfo( AirLoopNum ).ActualZoneNumber( zoneNumOnLoop );
```
rraustad, line 1166: This and next for loop are looping through the same zones. Should have combined the calculations.

rraustad:
Completed first pass of code review. Nothing stands out except some places where for loops could be combined. Not a big deal for now. TBD: I guess it would be prudent to at least run one or more example file to review results.

mjwitte:
@rraustad Thanks. I'm concerned about the big coil size diffs noted [here](https://github.com/NREL/EnergyPlus/pull/6372#issuecomment-356430121) Do you agree with the change?

rraustad:
Here's one difference I see in LgOffVAV before and after. The annual energy use if roughly the same yet there is a difference in peak sensible heat gain (and cooling coil size but we already know that). Not sure this has anything to do with these changes other than a change in air flow rate. But if the same load is met (unmet hours are roughly the same) then I don't understand these numbers. LgOffVAV uses non-coincident system sizing, that might explain this difference in cooling coil sizing (because it's working now). But why didn't the zone heating coil sizes change? I guess because they are already using the peak zone heating air flow rate.

lgoffvavstd621

The maximum OA did not change (41.38 m3/s). Before these changes the cooling coil was sizing using a lower main duct flow rate (33.56 m3/s) than the OA...why? So these changes do make sense I just don't understand what was happening before. The coil size scaled up by the fraction 41.38/33.56.

    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Maximum Outdoor Air Flow Rate [m3/s], 41.38217
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Minimum Outdoor Air Flow Rate [m3/s], 2.18064
    Component Sizing Information, Fan:VariableVolume, SUPPLY FAN 1, Design Size Maximum Flow Rate [m3/s], 41.38217
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Coil Load [W], 642477.14079
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Water Flow Rate [m3/s], 2.29474E-002
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Air Flow Rate [m3/s], 33.56326
 
After:

    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Maximum Outdoor Air Flow Rate [m3/s], 41.38217
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Minimum Outdoor Air Flow Rate [m3/s], 2.18064
    Component Sizing Information, Fan:VariableVolume, SUPPLY FAN 1, Design Size Maximum Flow Rate [m3/s], 41.38217
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Coil Load [W], 792149.01424
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Water Flow Rate [m3/s], 2.82933E-002
    Component Sizing Information, Coil:Cooling:Water, SIMPLE COOLING COIL, Design Size Design Air Flow Rate [m3/s], 41.38217

I've seen things that make me ask questions but haven't seen anything that scares me off of this.

rraustad:
5ZoneAirCooled_VRPSizing. Zone heating coils the same as before. Big difference is OA and main heating coil sizing. Noticable thing here is that min OA is now same as max OA. Not sure if that's correct or not.

@EnergyArchmage , why is the new sum of air terminal max/min heating flow rates the same? Will they always be the same?

Before:

    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Design Supply Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Maximum Outdoor Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Minimum Outdoor Air Flow Rate [m3/s], 0.62078
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size Rated Capacity [W], 16011.44943
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size Maximum Water Flow Rate [m3/s], 3.47846E-004
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size U-Factor Times Area Value [W/K], 199.37265
    Water Heating Coil Capacity Information,Coil:Heating:Water,OA HEATING COIL 1,10602.20
    Component Sizing Information, Controller:WaterCoil, OA HC CONTROLLER 1, Maximum Actuated Flow [m3/s], 3.47846E-004
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Coil Load [W], 27220.44497
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 3.24240E-003
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Air Flow Rate [m3/s], 0.62078

After:

    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Sum of Air Terminal Maximum Heating Flow Rates [m3/s], 0.31450
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Sum of Air Terminal Minimum Heating Flow Rates [m3/s], 0.31450
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Sum of Air Terminal Maximum Flow Rates [m3/s], 1.04833
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Adjusted Heating Design Air Flow Rate [m3/s], 0.43073
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Adjusted Cooling Design Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Adjusted Main Design Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, User Heating Air Flow Ratio [], 0.30000
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Calculated Heating Air Flow Ratio [], 0.41087
    Component Sizing Information, AirLoopHVAC, VAV SYS 1, Design Supply Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Maximum Outdoor Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Controller:OutdoorAir, OA CONTROLLER 1, Minimum Outdoor Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size Rated Capacity [W], 27039.07967
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size Maximum Water Flow Rate [m3/s], 5.87419E-004
    Component Sizing Information, Coil:Heating:Water, OA HEATING COIL 1, Design Size U-Factor Times Area Value [W/K], 336.68738
    Water Heating Coil Capacity Information,Coil:Heating:Water,OA HEATING COIL 1,17904.29
    Component Sizing Information, Controller:WaterCoil, OA HC CONTROLLER 1, Maximum Actuated Flow [m3/s], 5.87419E-004
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Coil Load [W], 45968.09199
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 5.47556E-003
    Component Sizing Information, Coil:Cooling:Water, OA COOLING COIL 1, Design Size Design Air Flow Rate [m3/s], 1.04833

rraustad:
5ZoneAirCooled_VRPSizing main coil sizing. Main cooling coil size went way down but OA cooling coil (above) went up.

Before:

    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 5278.83016
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 3.14398E-004
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Air Flow Rate [m3/s], 0.91260
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Air Temperature [C], 15.50224
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Water Temperature [C], 7.00000
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Temperature [C], 12.80000
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Air Humidity Ratio, 8.56498E-003
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 8.46737E-003
    Water Cooling Coil Capacity Information,Coil:Cooling:Water,MAIN COOLING COIL 1,9162.19,7334.61,1827.58,0.80,632.95,6.42
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size Rated Capacity [W], 4539.58860
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size Maximum Water Flow Rate [m3/s], 9.86217E-005
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size U-Factor Times Area Value [W/K], 72.10500

After:

    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 898.56612
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 5.35170E-005
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Air Flow Rate [m3/s], 1.04833
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Air Temperature [C], 11.00000
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Water Temperature [C], 7.00000
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Temperature [C], 12.80000
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Air Humidity Ratio, 8.00000E-003
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 8.00000E-003
    Water Cooling Coil Capacity Information,Coil:Cooling:Water,MAIN COOLING COIL 1,0.00,0.00,0.00,0.00,3487.24,35.37
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size Rated Capacity [W], 4539.58860
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size Maximum Water Flow Rate [m3/s], 9.86217E-005
    Component Sizing Information, Coil:Heating:Water, MAIN HEATING COIL 1, Design Size U-Factor Times Area Value [W/K], 72.10500

rraustad:
@EnergyArchmage why is the design size inlet air temperature now 11C while before it was 15.5C? Shouldn't it have been 11C before also?

    Sizing:System,
      VAV Sys 1,               !- AirLoop Name
      sensible,                !- Type of Load to Size On
      autosize,                !- Design Outdoor Air Flow Rate {m3/s}
      0.3,                     !- Central Heating Maximum System Air Flow Ratio
      4.5,                     !- Preheat Design Temperature {C}
      0.008,                   !- Preheat Design Humidity Ratio {kgWater/kgDryAir}
      11.0,                    !- Precool Design Temperature {C}
      0.008,                   !- Precool Design Humidity Ratio {kgWater/kgDryAir}
      12.8,                    !- Central Cooling Design Supply Air Temperature {C}
      16.7,                    !- Central Heating Design Supply Air Temperature {C}
      noncoincident,           !- Type of Zone Sum to Use

This also looks like it was fixed.

mjwitte:
@rraustad This one I think is because previously, the system wasn't at 100% OA, so only some of the air was at 11C and it mixed with return air.  I still don't understand why this one sizes out to 100% OA. It also has coil sizing [warnings now](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2017-12/60878bd885be786b57485acb2d6a5ee4d9343311-5c49025e07ea494367bba1e3e6444e39a17af65d/5ZoneAirCooled_VRPSizing_MaxZd/x86_64-MacOS-10.9-clang/eplusout.err.diff.html). Oh wait, the main cooling coil is sizing to zero now.  Should that be so?

rraustad:
@EnergyArchmage why is cooling flow rate being adjusted upwards? Can't the cooling coil be sized on the cooling load flow rate?

     } else if ( FinalSysSizing( AirLoopNum ).sysSizeHeatingDominant ) { // use maximum heating flow sum from air terminals 
        FinalSysSizing( AirLoopNum ).DesHeatVolFlow = max ( airLoopHeatingMaximumFlowRateSum, FinalSysSizing( AirLoopNum ).DesHeatVolFlow );
        FinalSysSizing( AirLoopNum ).DesMainVolFlow = max( airLoopHeatingMaximumFlowRateSum, FinalSysSizing( AirLoopNum ).DesMainVolFlow );
        //make sure cooling is at least at minimum.
        FinalSysSizing( AirLoopNum ).DesCoolVolFlow =  max( airLoopHeatingMinimumFlowRateSum, FinalSysSizing( AirLoopNum ).DesCoolVolFlow );

rraustad:
Design size inlet air temp = 11C, design outlet = 12.8C. So in this case it looks like it should size to 0.

rraustad:
I imagine the reason that min OA increased is in these numbers. Zone Ventilation efficiency went down for cooling. 0.2731 / 0.2519 = 1.08 m3/s so it's capped at 1.0483?.

vrpcalcs

rraustad:
The biggest question here is how do we know the 62.1 calculations are correct? There is a lot going on here. One way would be let users have a crack at proving this wrong. I don't know enough about VRP calculations to validate these equations or results.

rraustad:
I think the only way we could know if this is correct now is to identify the limiting zone, the one that sets down zone ventilation efficiency, and manually calculate the OA amount, then see if that amount of OA makes it to the zone before and after this fix. I will attempt this but not sure when I can get this done.

rraustad:
Trying to figure out just how much OA is required for the 5ZoneAirCooled_VRPSizing simulation. I reported occupancy and then calculated OA based on occupancy and floor area. Each zone used 0.00236 m3/s per person and 0.000305 m3/s per m2 of floor area. I also calculated max OA based on max occupancy and floor area.

Before these changes, it appears that the required amount of OA was met (as long as VRP uses my calculation method). With these changes there is more OA.

1st figure - compare OA mixing box OA with max OA needed by zones
2nd figure - compare OA mixing box OA with sum of OA provided to zones (match, sum of difference is 2.4 m3/s over entire year)
3rd figure - compare max zone OA with sum of actual OA (never need more than max sum, obviously)

So what would be the reason for increasing min OA amount?
Should outdoor air system be limiting OA based on zone requirement for VRP method?
Is there some other way of looking at this to know what amount of OA should be brought in at mixing box?

@EnergyArchmage comments?

    OutdoorAir:Mixer,
      OA Mixing Box 1 Inlet Node,  !- Outdoor Air Stream Node Name

5zoneaircooled_vrpsizing

rraustad:
Difference of sum of actual zone OA reported by TU minus mixing box OA, basically a match for this discussion.

5zoneaircooled_vrpsizing_oacompare

rraustad:
There does appear to be an issue with Zone 5 where insufficient OA enters the zone. Will look at Std 62.1 results next. Negative number indicates insufficient OA (last figure is calculation of zone OA reported by TU - zone OA need)

5zoneaircooled_vrpsizing_zoneoaneed

rraustad:
This branch does not correct what appears to be low OA in zone 5.

Not sure what else to look for.

5zoneaircooled_vrpsizing_zoneoaneed_thisbranch

EnergyArchmage:
ah lets see, I also spent several hours investigating the file 5ZoneAirCooled_VRPSizing.  I did side calcs in this file to check.    
[misc calcs1.xlsx](https://github.com/NREL/EnergyPlus/files/1741154/misc.calcs1.xlsx)

I had a long comment on this file above, did you see that in number 4.  Basically the file should be fixed up by changing inputs, but I did convince myself that the results are correct for the poor input.

EnergyArchmage:
My view is that there are terminal sizing problems that are separate from the focus here.  The terminals get sized on zone loads and don't always size based on minimum OA per std. 62.1

mjwitte:
Some of that may correct itself with #6327 so we'll revisit these after this goes in which should be soon.

EnergyArchmage:
regarding question above:  why is cooling flow rate being adjusted upwards? Can't the cooling coil be sized on the cooling load flow rate?

FinalSysSizing( AirLoopNum ).DesCoolVolFlow =  max( airLoopHeatingMinimumFlowRateSum, FinalSysSizing( AirLoopNum ).DesCoolVolFlow );

The idea is that the cooling air flow rate should be at least as high as the minimum operating flow rate.  If the design cooling air flow rate comes out lower than that it needs to be adjusted (increased) because the air handler will never run below that lower limit.  We now know more about the air handler operating limits and can use that information to make adjustments. This is in the heating dominant code block so heating design flow rates are dominating the air handler design. This ensures that the cooling coil is able to make setpoint during operation at the minimum damper positions.

EnergyArchmage:
In 5zoneAirCooled_VRPSizing, the air handler becomes 100% OA because the inputs for zone 5 are bad, discussed above.

rraustad:
Single stepping this shows 0 OA mass flow rate during the TU calculations (i.e., Sys( SysNum ).NoOAFlowInputFromUser = true;). Further investigation shows that the TU's in 5ZoneAirCooled_VRPSizing do not include the name of the DesignSpecification:OutdoorAir so the TU thinks there is no OA requirement. This is probably why I saw low OA amounts last night. I wonder if/how the html tables would change if this name were included in each TU.

    A11; \field Design Specification Outdoor Air Object Name
       \type object-list
       \object-list DesignSpecificationOutdoorAirNames
       \note When the name of a DesignSpecification:OutdoorAir object is entered, the terminal
       \note unit will increase flow as needed to meet this outdoor air requirement.
       \note If Outdoor Air Flow per Person is non-zero, then the outdoor air requirement will
       \note be computed based on the current number of occupants in the zone.
       \note At no time will the supply air flow rate exceed the value for Maximum Air Flow Rate.
       \note If this field is blank, then the terminal unit will not be controlled for outdoor air flow.


    AirTerminal:SingleDuct:VAV:Reheat,
      SPACE1-1 VAV Reheat,     !- Name
      ReheatCoilAvailSched,    !- Availability Schedule Name
      SPACE1-1 Zone Coil Air In Node,  !- Damper Air Outlet Node Name
      SPACE1-1 ATU In Node,    !- Air Inlet Node Name
      autosize,                !- Maximum Air Flow Rate {m3/s}
      Constant,                !- Zone Minimum Air Flow Input Method
      0.3,                     !- Constant Minimum Air Flow Fraction
      ,                        !- Fixed Minimum Air Flow Rate {m3/s}
      ,                        !- Minimum Air Flow Fraction Schedule Name
      Coil:Heating:Water,      !- Reheat Coil Object Type
      SPACE1-1 Zone Coil,      !- Reheat Coil Name
      autosize,                !- Maximum Hot Water or Steam Flow Rate {m3/s}
      0.0,                     !- Minimum Hot Water or Steam Flow Rate {m3/s}
      SPACE1-1 In Node,        !- Air Outlet Node Name
      0.001,                   !- Convergence Tolerance
      NORMAL,                  !- Damper Heating Action
      AUTOCALCULATE,           !- Maximum Flow per Zone Floor Area During Reheat {m3/s-m2}
      AUTOCALCULATE;           !- Maximum Flow Fraction During Reheat
      ,
      ;         < --- DesignSpecification:OutdoorAir name goes right here

      sum,                     !- Outdoor Air Method
      0.00236,                 !- Outdoor Air Flow per Person {m3/s-person}
      0.000305,                !- Outdoor Air Flow per Zone Floor Area {m3/s-m2}

EnergyArchmage:
Yes, I found the same thing.  I think there should be a follow up issue/PR to fix up the VRP test files.

mjwitte:
@EnergyArchmage @rraustad Some of the TU vs OA stuff may get fixed in #6338. I've posted some followup needs in #6481. You can add the VRP idf to that list or make a separate issue.  The Std62 sizing changes also slowed down the performance tests by about 10%. Any idea where the extra time would be coming from here?

EnergyArchmage:
I would guess the performance hit is from the population diversity side calculation.

PR #6380, opened 2017-10-30
Fix bug in PCM code

Myoldmopar:
Pull request overview
---------------------
@nealrkuis found a typo-defect in the PCM code.  This was addressed here.  I will add a unit test.

The bug (in v8.8.0) causes a crash or invalid results if there is more than one material layer in the construction which holds the PCM material.
Ref: helpdesk ticket 12504

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nealkruis:
@Myoldmopar looks good. I think the Linux failure is a CI issue. Should we force a rebuild just to be safe?

Myoldmopar:
Nah, I'll push up a unit test sometime soon anyway, so we'll let it retry then.  Thanks!

nrel-bot-3:

Myoldmopar:
For reference, I checked out the diffs here and they are very small diffs, and is likely because in some of the changes we are now updating all the possible dynamic properties, not just specific heat.  Which we should've been doing all along.  This is OK.  I'm still thinking about how to best test that new little function.

Myoldmopar:
CI appears happy.  The SQLiteFixture failure is fully unrelated.  The one regression diff is expected and reasonable.  If CI continues to be happy I'll merge it in.

Myoldmopar:
OK, late diffs are due to HWBaseBoard PR, this branch is now 7 commits behind develop.  Not waiting for another go.  Merging.

Issue #6382, opened 2017-10-31
VAV airflow problems with Ideal Loads Zone Equipment


lgu1234:
Issue overview
--------------
#6173 provided warnings when exhaust air fields are not set up correctly for ZoneHVAC:OutdoorAirUnit objects. It does not address exhaust air flow rate of ZoneAirExhaust node for ZoneHVAC:IdealLoadsAirSystem objects. 
 
5952  ### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6386

Issue #5788, opened 2016-07-25
ZoneHVAC:IdeaLoadsAirSystem EMS mass flow rate and supply temperature actuators not working correctly


mjwitte:
## Issue overview

User noticed that the actual zone sensible cooling reported by ZoneHVAC:IdealLoadsAirSystem was not consistent with the EMS actuated values for supply temperature and mass flow rate.  Detailed reporting revealed that the ideal loads system was off (zero supply flow) at times when the EMS actuated mass flow rate was > 0.  It appears that ideal loads system can shut itself off when the thermostat requests zero heating or cooling and then does not check the EMS flow rate.  A quick check of source code confirms that EMS overrides are skipped when the unit if "off" due to schedule or zero load.  

It seems that zero load should not skip the EMS actuators, but then what should the supply temperature be if that is not also set by EMS?

And should they be skipped when the ideal loads system is scheduled off as it currently does?

And should the EMS temperature and mass flow overrides honor the user inputs for max heating/cooling flow rate and max/min supply temperatures?  Not sure what's done with other EMS actuators in this regard.

### Workaround
*Added 6/28/2017*
Adding ventilation flow that is on 24/7 can force the system to "operate" all the time and it will then execute the EMS actuators.  If desired, the OA flow can be set back to zero using EMS along with setting the supply flow rate, temperature and humidity ratio.

### Details

Some additional details for this issue (if relevant):
- Platform (Win64)
- Version of EnergyPlus v8.5.0
- Helpdesk ticket number 11494
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Also, it would be nice to add two more output variables
Zone Ideal Loads Supply Air Temperature
Zone Ideal Loads Supply Air Humidity Ratio

mjwitte:
Added another defect file that uses EMS to actuate both the ideal loads mass flow rate and supply temperature.  Symptoms include warnings about supply humidity ratio exceeding saturation (which should not happen) and times where the supply temp is colder than the mixed air temp but there is no ideal loads coil load shown.  For example, see 5am to 7am on July 3.

mjwitte:
The new variables here were addressed in #8203.

PR #6386, opened 2017-11-06
Fix ideal loads zone air flow balance

lgu1234:
Pull request overview
---------------------
Addresses #6382.

The ideal loads system was not always setting the correct mass flow rate on the exhaust node. When used together with an airloop serving the same zone this  caused the return air flow rate for the airloop to be incorrect. For the defect file which has a VAV system together with the ideal loads system, this resulted in the VAV system operating at nearly full flow any time the ideal loads system was operating.  
Always settting the mass flow rate of ideal loads exhaust node (or return node if there is no exahaust node) to match the ideal loads supply flow solves this issue.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


nrel-bot-3:

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2307,6 +2307,10 @@ namespace PurchasedAirManager {
 		if ( PurchAir( PurchAirNum ).PlenumExhaustAirNodeNum > 0 ) {
 			Node( PurchAir( PurchAirNum ).PlenumExhaustAirNodeNum ).MassFlowRate = SupplyMassFlowRate;
 			Node( RecircNodeNum ).MassFlowRate = SupplyMassFlowRate;
+		} else {
+			if ( PurchAir( PurchAirNum ).ZoneExhaustAirNodeNum > 0 && abs( Node( PurchAir( PurchAirNum ).ZoneExhaustAirNodeNum ).MassFlowRate - SupplyMassFlowRate) > rTinyValue ) {
+				Node( PurchAir( PurchAirNum ).ZoneExhaustAirNodeNum ).MassFlowRate = SupplyMassFlowRate;
+			}
```
mjwitte: @lgu1234 I am not sure that this solution is general. If there is no exhaust node, the airflow imbalance will still occur. 

An better solution is to *always* set the recirc node mass flow rate equal to the supply flow rate.  The current line 2309
```
Node( RecircNodeNum ).MassFlowRate = SupplyMassFlowRate;
```
should just come out of the above if block and always execute.

Also, [this block](https://github.com/NREL/EnergyPlus/blob/017868e84f0eb0de7711a942c72aebc273bafcc8/src/EnergyPlus/PurchasedAirManager.cc#L2514-L2517) in `CalcPurchAirMixedAir` should be deleted.
a) This function shouldn't be touching any node massflowrates in the first place (or any node conditions at all)
b) The recirc node should always be set to `SupplyMassFlowRate`. Setting it to `RecircMassFlowRate` was just plain wrong.

lgu1234: @mjwitte Here is my understanding. There are 3 nodes in the system, zone inlet node, zone exhaust node, and recirc node. 

If we set Node( RecircNodeNum ).MassFlowRate = SupplyMassFlowRate, and the ideal system is a DOAS system, do we have a relief node to handle outdoor air. If we don't have a relief node, here is my vision:
Node( RecircNodeNum ).MassFlowRate = 0.0;
Node( ZoneInletNodeNum ).MassFlowRate = SupplyMassFlowRate;
Node( ExhaustNodeNum ).MassFlowRate = SupplyMassFlowRate;

mjwitte:
The exhaust node is optional.  If present then RecircNodeNum = ExhaustNodeNum. If not, then RecircNodeNum = ReturnNodeNum. So only two of these are ever active. There is an implied relief flow but no relief node.


mjwitte:
@lgu1234 This looks good now, as far as fixing the flow rate issue.  There's still something wrong with EMS actuating ideal loads - the warnings about supply humidity exceeding saturation should not be happening.  There were there before and after this fix.  A version of the defect file without any EMS code does not have these warnings but still exhibits the flow rate problem and the fix.  Will add more to #5788 which already identifies one problem with EMS and ideal loads.

PR #6395, opened 2017-11-20
Simple Cooling Panel OO Mods, Corrections, and Documentation Fixes

RKStrand:
Pull request overview
---------------------
This pull request addresses the changes to the simple cooling panel model outlined in GitHub Issue #5823.  Various changes were made (and other work verified) relating to this model.  Changes included making the code more OO overall, correcting a reporting bug, and fixing documentation problems such as incorrect equations, poor wording and references to equation numbers that did not exist, etc.  This also included a new unit test to verify the correction of the reporting bug.

### Work Checklist
 - [x ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue #5823 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
@@ -890,13 +890,13 @@ The development of the method for estimating the panel UA value uses the same eq
 \begin{itemize}
 \tightlist
 \item
-  Rating Inlet Water Temperature \PB{T_{rwi}}
+  Rating Inlet Water Temperature $\PB{T_{rwi}}$
 \item
-  Rating Air/Surroundings Temperature \PB{T_{rz}}
+  Rating Air/Surroundings Temperature $\PB{T_{rz}}$
 \item
-  Rating Water Mass Flow Rate \PB{m_r}
+  Rating Water Mass Flow Rate $\PB{m_r}$
 \item
-  Rated Output Capacity \PB{Q_r}
+  Rated Output Capacity $\PB{Q_r}$
 \end{itemize}
 
 To find UA, first rearrange the equations above to make NTU a function of effectiveness as follows:
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
@@ -819,7 +819,7 @@ Many of the control and integration aspects of the high temperature radiant syst
 
 \subsubsection{Overview}\label{overview}
 
-The simple chilled ceiling model is similar to the high temperature radiant system and the hot water radiant/convective baseboard models and was developed to make the modeling of certain radiant systems easier.  The existing chilled beam model in EnergyPlus, while useful for modeling active chilled beam systems, is a convective heat transfer only model that is not suitable for all radiant panels or passive chilled beam systems.
+The simple chilled ceiling model (ZoneHVAC:CoolingPanel:RadiantConvective:Water) is similar to the high temperature radiant system and the hot water radiant/convective baseboard models and was developed to make the modeling of certain radiant systems easier.  The existing chilled beam model in EnergyPlus, while useful for modeling active chilled beam systems, is a convective heat transfer only model that is not suitable for all radiant panels or passive chilled beam systems.
 
 \subsubsection{Modeling Considerations/Assumptions}\label{modeling-considerations-assumptions}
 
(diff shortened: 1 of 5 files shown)


Myoldmopar:
@RKStrand I built it locally and was able to reproduce the LaTeX errors keeping the docs from building and thus showing a failed Linux build.  I fixed those and pushed up the commit.  I'll review the code now.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
```diff
@@ -1569,33 +1577,36 @@ namespace CoolingPanelSimple {
...
-			QCoolingPanelToPerson( ZoneNum ) += CoolingPanelSource( CoolingPanelNum ) * CoolingPanel( CoolingPanelNum ).FracDistribPerson;
+			QCoolingPanelToPerson( ZoneNum ) += CoolingPanelSource( CoolingPanelNum ) * ThisCP.FracDistribPerson;
 
-			for ( RadSurfNum = 1; RadSurfNum <= CoolingPanel( CoolingPanelNum ).TotSurfToDistrib; ++RadSurfNum ) {
-				SurfNum = CoolingPanel( CoolingPanelNum ).SurfacePtr( RadSurfNum );
-				if ( Surface( SurfNum ).Area > SmallestArea ) {
-					ThisSurfIntensity = ( CoolingPanelSource( CoolingPanelNum ) * CoolingPanel( CoolingPanelNum ).FracDistribToSurf( RadSurfNum ) / Surface( SurfNum ).Area );
+			for ( RadSurfNum = 1; RadSurfNum <= ThisCP.TotSurfToDistrib; ++RadSurfNum ) {
+				SurfNum = ThisCP.SurfacePtr( RadSurfNum );
+				auto & ThisSurf( Surface( SurfNum ) );
```
Myoldmopar, line 1591: These are all very good steps, just keep in mind that in the Utopian future, we'll have lower case variable names, so this would be `thisSurf`.  The reduction in array lookups here makes the code much more readable.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
```diff
@@ -855,89 +855,85 @@ \subsubsection{Modeling Considerations/Assumptions}\label{modeling-consideration
...
 
-The development of the method for estimating the panel UA value will use the same equations shown above but simply with the purpose of calculating UA rather than the performance of the device, i.e. determining the output of the panel.  The method for estimating the UA value of the panel begins with the panel rating data that includes the following values:
+The development of the method for estimating the panel UA value uses the same equations shown above but simply with the purpose of calculating UA rather than the performance of the device, i.e. determining the output of the panel.  The method for estimating the UA value of the panel begins with the panel rating data that includes the following values:
 
 \begin{itemize}
 \tightlist
 \item
-  Rating Inlet Water Temperature (Trwi)
+  Rating Inlet Water Temperature $\PB{T_{rwi}}$
```
Myoldmopar, line 893: Just need to capture the math inside a math mode environment to fix the build errors here.

Myoldmopar (approved):

Myoldmopar:
If this comes back clean, it should be ready to go.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-004/radiant-system-models.tex
RKStrand: @Myoldmopar Thanks, @myoldmopar, for cleaning this up.  I'll try to remember to keep all things like that inside math mode.

src/EnergyPlus/ChilledCeilingPanelSimple.cc
RKStrand: @Myoldmopar Ah, I will definitely keep this in mind in the future.  Still working in old EnergyPlus F90 mode there.  Good thing that is an easy fix and I will make sure I start all variables with lowercase in future work.  Thanks for getting to this review so quickly!

Myoldmopar:
@RKStrand Seeing diffs doesn't bother me given the changes you've made, but I'm just wondering if you've investigated them to ensure they are OK.  I can build develop and run things and compare results, but if you already have then I'm OK with it.

RKStrand:
@myoldmopar I did check the diffs when the commit with the reporting change went in. That commit basically multiplied the output by -1 so it made more sense.   Before that commit it was showing negative cooling which would indicate heating. But the numbers were negative because the panel is cooling. So it was just a mismatch between the output variable name and what was being output. So, when the diffs came up that was expected. I looked at the output for one case and it showed that the results were identical except the sign of the output of the cooling panel heat transfer. I looked at the percent diffs of the other files and saw the diffs were exactly 200% which I felt meant that the sign had flipped on just those variables. So I felt like the output was exactly what I expected since no other output like zone temps, etc changed. Hope that helps clarify what I saw.

Myoldmopar:
Thanks, I saw the 200% also, and that's what triggered my question to you about the diffs.  This is all good.  Merging.

PR #6397, opened 2017-11-21
Correction of Radiant/convective Baseboard Max Iteration Issues

RKStrand:
Pull request overview
---------------------
GitHub Issue #5884 noted a problem with solution convergence for the hot water radiant/convective hot water baseboard model.  In files in the test suite, the HW rad/conv BB unit showed a number of times during an annual run when the baseboard model exceeded max iterations (over 60 times).  After in-depth of tests/solution ideas and analysis, a solution that involved a new convergence criteria was implemented.  The problem with the radiant/convective baseboard units is that they have a radiant component which can potentially result in solutions that wander a bit because of the radiant absorbed/released by surfaces.  The solution involves capping iterations when the non-dimensional change in the flow rate drops below 0.00001.  Since the control of radiant/convective devices like these units is not as exact as air components, it was decided that this is a reasonable approach and it does in fact eliminate the max iteration errors.  This new algorithm is only applied to the radiant/convective baseboard units (steam and hot water).  The only changes are in the test suite output files are in input files that use these devices (as expected).  The changes are relatively small--smaller than anticipated considering that the iteration scheme cutoff has changed--and are not a major change in the output.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #5884

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -616,6 +631,50 @@ ControlCompOutput(
...
+	// on the actual system output is not as well behaved as for convective
+	// systems.  This additional check avoids excessive iterations and max
+	// iteration warnings and provides sufficiently converged results.  It is
+	// only called from ControlCompOutput.
+
+	// Return Value
+	bool BBConvergeCheck;
+	
+	// SUBROUTINE PARAMETER DEFINITIONS:
+	static Real64 const BBIterLimit( 0.00001 );
```
Myoldmopar, line 675: This criterion seems very reasonable.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -616,6 +631,50 @@ ControlCompOutput(
...
+
+	if ( SimCompNum != 5 && SimCompNum != 6 ) {
```
Myoldmopar: I don't really like these magic numbers.  Are these called out as actual named enums in the equipment data modules somewhere?  Otherwise I don't understand what these are.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -616,6 +631,50 @@ ControlCompOutput(
...
+		// For all zone equipment except radiant/convective baseboard (steam and water) units:
+		BBConvergeCheck = false;
+	} else {
+		// For steam and water radiant/convective baseboard units:
+		if ( ( MaxFlow - MinFlow ) > BBIterLimit ) {
```
Myoldmopar, line 682: There's no reason to assume max could ever be below min here, right?  At first I was thinking about putting an absolute value call around that difference, but really, that sort of condition would probably reveal a bigger problem, and we shouldn't put a band-aid over it for no reason.  Nothing to see here then...carry on...

src/EnergyPlus/GeneralRoutines.hh
```diff
@@ -81,6 +81,13 @@ ControlCompOutput(
 	Optional_int_const BranchIndex = _, // for plant components, plant branch index
 	Optional_int_const ControlledZoneIndex = _ // controlled zone index for the zone containing the component
 );
+	
+bool
+BBConvergeCheck(
+	int const SimCompNum,
+	Real64 const MaxFlow,
+	Real64 const MinFlow
+);
```
Myoldmopar, line 90: Great job extracting this into an easily testable function

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
```diff
@@ -0,0 +1,114 @@
...
+// IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY
+// AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+// EnergyPlus::SortAndStringUtilities Unit Tests
```
Myoldmopar: Stray comment left here in the header

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
```diff
@@ -0,0 +1,114 @@
...
+// EnergyPlus Headers
+#include <EnergyPlus/GeneralRoutines.hh>
+
+#include "Fixtures/EnergyPlusFixture.hh"
+
+namespace EnergyPlus {
+
+	TEST_F( EnergyPlusFixture, BBConvergeCheckTest )
+	{
+	// R.Strand - November 2017
```
Myoldmopar: Don't need this comment, we can always use Git Blame to find out who added what

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
```diff
@@ -0,0 +1,114 @@
...
+	Real64 MaxFlow;
+	Real64 MinFlow;
+	bool FunctionResult;
+
+	// Test 1: Not a radiant/convective baseboard unit
+	SimCompNum = 1;
+	MaxFlow = 1.0;
+	MinFlow = 0.99999999;
+	FunctionResult = BBConvergeCheck( SimCompNum, MaxFlow, MinFlow );
+	EXPECT_EQ( FunctionResult, false );
```
Myoldmopar: Use `EXPECT_FALSE( FunctionResult );` for even easier to read code.

Myoldmopar (changes requested):
There are some small changes that need to be addressed, but overall this looks like a good step.  I obviously haven't tested yet; I'll wait and see what CI initially reports.

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
RKStrand: @Myoldmopar Love that it is called Git Blame...

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
RKStrand: Deleted

tst/EnergyPlus/unit/GeneralRoutines.unit.cc
RKStrand: Fixed

src/EnergyPlus/GeneralRoutines.hh
RKStrand: @Myoldmopar Thanks...it seems to be my speciality...

src/EnergyPlus/GeneralRoutines.cc
RKStrand: Right...if Max is smaller than Min something major is wrong and it will probably not even get here.

src/EnergyPlus/GeneralRoutines.cc
RKStrand: @Myoldmopar Hmmm...so in ControlCompOutput, this is sort of hardwired into a local data array.  There aren't even any integer parameters.  I can tackle this two ways:
1. I can create parameters locally in my new subroutine that spell this out.
2. Maybe this can be done at the namespace level and then ControlCompOutput can use them as well.
What do you prefer?

RKStrand: @Myoldmopar went with option 2 as that was a bit cleaner.  Hopefully I didn't introduce any errors in the process and this is now ready to go.

RKStrand:
@Myoldmopar Ok, I think I took care of all of the review comments.  I am marking this in Pivotal Tracker as finished and delivered.  If you have any other issues that come up, please let me know.  Thanks!

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -101,6 +101,19 @@ namespace EnergyPlus {
...
+int const HeatingCoilWaterNum( 3 );
+int const BBWaterConvOnlyNum( 4 );
+int const BBSteamRadConvNum( 5 );
+int const BBWaterRadConvNum( 6 );
+int const FourPipeFanCoilNum( 7 );
+int const OutdoorAirUnitNum( 8 );
+int const UnitHeaterNum( 9 );
+int const UnitVentilatorNum( 10 );
+int const VentilatedSlabNum( 11 );
+
```
Myoldmopar: Thanks @RKStrand, I wasn't clear enough.  I was hoping for an enum class instead of stray constants.  Like:

```
enum class LocalEquipTypes {
  ParallelPIUReheat = 1;
  ...
};
```

Then using it like:

```
  case LocalEquipTypes::ParallelPIUReheat
```

I'll probably make this change before I do final testing/merge.  It will be nicer than just the list of constants.

Myoldmopar:
I pushed a commit changing the integers to a simple enum.  I tried to do an enum class but it was going to be more work than I wanted to invest.  I don't see anything else wrong with the code, I'll do some testing.

Myoldmopar:
And merged develop so that CI should be all clean.

Myoldmopar (approved):

Myoldmopar:
Small diffs; no major issues, merging.

Issue #6405, opened 2017-11-29
Unitary system as both zone equipment and airloop equipment in the same file fails in v8.8


mjwitte:
Issue overview
--------------
Multiple airloop changes in #6248 required some rearranging of HVACUnitarySystem::GetUnitarySystemInputData to separate the input processing for unitary system as zone equipment and air loop equipment.  This solved some issues with internal test files, but some files fail.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus 8.9.0 d8a479ad5f
 - e-mail on 2017-11-28 from @rraustad 

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Closed via #6406.

PR #6406, opened 2017-11-29
Fix get input problem when same file has unitary systems as zone equipment and on airloop

mjwitte:
Pull request overview
---------------------
Addresses #6405, get input errors for unitary systems used as both zone equipment and airloop equipment in the same file.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HVACUnitarySystem.cc
+++ b/src/EnergyPlus/HVACUnitarySystem.cc
@@ -2703,10 +2703,7 @@ namespace HVACUnitarySystem {
 		bool ErrorFlag( false ); // true if errors detected in GetUnitarySystemInputData
 
 		// Flow
-		if ( GetInputFlag ) { //First time subroutine has been entered
-			GetUnitarySystemInputData( ErrorFlag );
-			GetInputFlag = false;
-		}
+		GetUnitarySystemInputData( ErrorFlag );
 
 		if( ErrorFlag ) {
 			ShowFatalError( RoutineName + "Errors found in getting AirLoopHVAC:UnitarySystem input. Preceding condition(s) causes termination." );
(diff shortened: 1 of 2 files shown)


mjwitte:
@rraustad This fixes the defect file.  The diff on github is a mess.  There are two changes here:
1.  Remove the GetInputFlag checks in [GetUnitarySystemInput](https://github.com/NREL/EnergyPlus/blob/a969bdd4338121f9ae15b58fd2ebbe38475cbab2/src/EnergyPlus/HVACUnitarySystem.cc#L2706) that were added as part of the multiple airloop changes.
2.  Move [this entire block](https://github.com/NREL/EnergyPlus/blob/a969bdd4338121f9ae15b58fd2ebbe38475cbab2/src/EnergyPlus/HVACUnitarySystem.cc#L3244-L3504) up before the fan and coil input processing.

rraustad:
I made these changes to my local develop and then on my SZVAV branch. Compiled both exe's to compare a few test files, including this defect file. I also copied this defect file and changed to Load based control for the 2 UnitarySystem objects in this file. This appears to be working. The only SMALL diff's I see are in my SZVAV test file. The defect file using Load control has 0 diff's.

mjwitte:
Well, foo.  The failing tests ran locally.  Hmm.

rraustad:
These are hard to fix. If you look up in the virtual void Setup section where I try to set everything up that's needed, except for the parent and child objects for unitary systems. It's probably that something is missing in that setup and these unit tests hit an uninitialized variable, unallocated array, or something. The only way I've figured these out is to single step very carefully looking for something missing (from Setup). MAYBE you could find it in debug but I'm not sure.

mjwitte:
I had these tests in `ZoneUnitarySystemTest` instead of `EnergyPlusFixture`.  Hopefully that will fix it.

rraustad:
@Myoldmopar why can't I see the test failure for the SQLiteFixture tests? When I click the Test Badge button (which isn't there but the text is) I get a GitHub 404 error page that says File Not Found.
Would you agree that a failure in SQLiteFixture has nothing to do with these changes and I can accept this PR since everything else passes?

mjwitte:
@rraustad I deleted those results and it should rebuild - it was system errors - not real failures.  I figured it would be done by now.

rraustad (approved):
Reviewed and implemented in parallel branch (SZVAV) and works as expected. CI clean.

Issue #3469, opened 2014-01-14
Update mean radiant temperature ( MRT ) section in Engineering Reference. (CR #8696)


axelstudios:
## 
#### Description

MJW 22 Dec 2011
User asked . . .

---

We are getting very different PMV results from two different simulation programs and wish to judge which is modelling closest to the requirements of ISO7730 and then why the results are different. This alternate package cites CIBSE guide A (which in turn cites ISO7726) as its source for MRT calculation, so it would be helpful to know whether EnergyPlus cites the same document.
## The discussion on page 962 of the eng-ref seems to describe how the Mean radiant temperature is averaged, depending on which model the user selects, but I can't seem to find information on how the mean radiant temperature is arrived at to begin with (if that makes sense). Tr (mean radiant temperature) is an input to the formulae, not a result of the formulae.

Not sure what version of docs the users is looking at, in v7, this discussion starts on pdf p. 1033.  There are equations for Tr which is the resulting MRT used in the thermal comfort calcs, but I don't see an equation for Tr-avg which is the Zone averaged MRT.  Also, the existing equations and text seem to be a mix of real equations plus source code snippets using fortran variable names without any definitions.  
1.  So, this whole section could use a review and edit, 
2.  but most importantly add the equation for zone averaged MRT which is (from Rick's reply):

Here is the equation that is used for the "zone averaged" MRT in EnergyPlus.

MRT = Sum [(emissivity(i)_area(i)_temperature(i)] / Sum [(emissivity(i)*area(i)]

Sum[] is the summation over all surfaces (i=1 to the number of surfaces in a zone). So it is essentially the summation of the product of each surface emissivity times area times temperature divided by the summation of the product of each surface emissivity times area.
1.  Double-check the reference citations - Linda's comment:  "Note that ISO 7730 (I think it's a typo) is cited there in the references section. (I think it's a typo how it is cited)."
## 

External Ref: Ticket 5125
Last build tested: `11.12.19 V7.1.0.001`

mjwitte:
Also note that the I/O Reference describes MRT in words

> Zone Mean Radiant Temperature [C]
> The Mean Radiant Temperature (MRT) in degrees Celsius of a space is a measure of the combined effects of temperatures of surfaces within that space. Specifically it is the surface area × emissivity weighted average of the zone inside surface temperatures (ref. Surface Inside Temperature), where emissivity is the Thermal Absorptance of the inside material layer of each surface.

[...]

PR #6407, opened 2017-11-29
Thermal Comfort Mean Radiant Temperature Calculation Upgrades

RKStrand:
Pull request overview
---------------------
This pull request addresses issues that first came up in GitHub Issue #3469.  This started as a documentation fix and then became a more critical look at how we are calculating various flavors of MRT in the thermal comfort routine.  Two were identified as having something less than ideal.  The surface weighted option which averages one surface temperature and the zone MRT was basically double counting the surface that the person was closest to.  So, this was altered so that the zone MRT used for this calculation did not include the effect of the surface the person was closest to.  The angle factor option was using a simplified equation that did not account for surface emissivity and was also using a linear version of the equation.  Both of those deficiencies were addressed with this work.  Two unit tests and documentation mods were also completed.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #3469 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
@@ -1064,7 +1064,7 @@ T\(_{r-avg}\) & Zone averaged radiant temperature & °C & -  \tabularnewline
 T\(_{surf}\) & Surface temperature & °C & -  \tabularnewline
 F\(_{surf}\) & Angle factor between person and surface & - & 0-1  \tabularnewline
 A\(_{surf}\) & Area of the surface & m\(^{2}\) & - \tabularnewline
-\epsilon\(_{surf}\) & Emissivity of the surface & - & - \tabularnewline
+\(\epsilon_{surf}\) & Emissivity of the surface & - & - \tabularnewline
 \bottomrule
 \end{longtable}
 
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
@@ -1063,6 +1063,8 @@ T\(_{r}\) & Mean radiant temperature & °C & - \tabularnewline
 T\(_{r-avg}\) & Zone averaged radiant temperature & °C & -  \tabularnewline
 T\(_{surf}\) & Surface temperature & °C & -  \tabularnewline
 F\(_{surf}\) & Angle factor between person and surface & - & 0-1  \tabularnewline
+A\(_{surf}\) & Area of the surface & m\(^{2}\) & - \tabularnewline
+\epsilon\(_{surf}\) & Emissivity of the surface & - & - \tabularnewline
 \bottomrule
 \end{longtable}
 
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
@@ -1041,66 +1041,66 @@ Lee KH, Schiavon S. 2013. Influence of three dynamic predictive clothing insulat
 
 \subsection{Mean Radiant Temperature Calculation}\label{mean-radiant-temperature-calculation}
 
-There are three options to calculate mean radiant temperature in the thermal comfort models. One is the zone averaged MRT, another is the surface weighted MRT, and the other is angle factor MRT. The zone averaged MRT is calculated on the assumption that a person is in the center of a space, whereas the surface weighted MRT is calculated in consideration of the surface that a person is closest to, and the angle factor MRT is calculated based on angle factors between a person and the different surfaces in a space. Here, the surface weighted MRT is the average temperature of the selected surface and zone averaged MRT and is intended to represent conditions in the limit as a person gets closer and closer to a particular surface.~ In that limit, half of the person's radiant field will be dominated by that surface and the other half will be exposed to the rest of the zone.~ Note that the surface weighted MRT is only an approximation. The angle factor MRT is the mean temperature of the surrounding surface temperatures weighted according to the magnitude of the respective angle factors and allows the user to more accurately predict thermal comfort at a particular location within a space.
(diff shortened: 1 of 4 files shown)


src/EnergyPlus/ThermalComfort.cc
```diff
@@ -1881,47 +1708,51 @@ namespace ThermalComfort {
...
+			thisAngFacList.SurfacePtr.allocate( thisAngFacList.TotAngleFacSurfaces );
+			thisAngFacList.AngleFactor.allocate( thisAngFacList.TotAngleFacSurfaces );
+			thisAngFacList.SurfaceEmissivity.allocate( thisAngFacList.TotAngleFacSurfaces );
+
+			for ( SurfNum = 1; SurfNum <= thisAngFacList.TotAngleFacSurfaces; ++SurfNum ) {
+				thisAngFacList.SurfaceName( SurfNum ) = cAlphaArgs( SurfNum + 2 );
+				thisAngFacList.SurfacePtr( SurfNum ) = FindItemInList( cAlphaArgs( SurfNum + 2 ), Surface );
+				thisAngFacList.AngleFactor( SurfNum ) = rNumericArgs( SurfNum );
+				thisAngFacList.SurfaceEmissivity( SurfNum ) = Construct( Surface( thisAngFacList.SurfacePtr( SurfNum ) ).Construction ).InsideAbsorpThermal;
+				thisAngFacList.SumSurfaceEmissAngleFactor += thisAngFacList.SurfaceEmissivity( SurfNum ) * thisAngFacList.AngleFactor( SurfNum );
```
mjwitte: Nice efficiency idea to precalculate this, but `InsideAbsorpThermal` can be changed with EMS, so this needs to be recalculated every timestep just in case.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -1985,35 +1807,94 @@ namespace ThermalComfort {
...
 	}
 
+	Real64
+	CalcSurfaceWeightedMRT(
+		int const ZoneNum,
+		int const SurfNum
+	)
+	{
+		
+		// Purpose: Calculate a modified zone MRT that excludes the Surface( SurfNum ).
```
mjwitte, line 1843: This function does more than this, it gets the final answer including the weighted surface.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -1985,35 +1807,94 @@ namespace ThermalComfort {
...
+				SumAET += SurfaceAE( SurfNum2 ) * TH( 2, 1, SurfNum2 );
+			}
+		}
+		
+		// Now weight the MRT--half comes from the surface used for weighting (SurfNum) and the rest from the adjusted MRT that excludes this surface
+		if ( ZoneAESum( ZoneNum ) > 0.01 ) {
+			CalcSurfaceWeightedMRT = 0.5 * ( TH( 2, 1, SurfNum ) + ( SumAET / ZoneAESum( ZoneNum ) ) );
+		} else {
+			ShowWarningError( "Zone areas*inside surface emissivities are summing to zero, for Zone=\"" + Zone( ZoneNum ).Name + "\"" );
+			ShowContinueError( "As a result, MAT will be used for MRT when calculating a surface weighted MRT for this zone." );
```
mjwitte: Don't want these warnings to write every timestep.  Sufficient to write this once - and add "only the first occurrence is reported" or something to that effect.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -1985,35 +1807,94 @@ namespace ThermalComfort {
...
+		// Initialize ZoneAESum for all zones and SurfaceAE for all surfaces
+		if ( FirstTimeSurfaceWeightedFlag ) {
+			FirstTimeSurfaceWeightedFlag = false;
+			SurfaceAE.allocate( TotSurfaces );
+			ZoneAESum.allocate( NumOfZones );
+			SurfaceAE = 0.0;
+			ZoneAESum = 0.0;
+			for ( SurfNum2 = 1; SurfNum2 <= TotSurfaces; ++SurfNum2 ) {
+				if ( Surface( SurfNum2 ).HeatTransSurf ) {
+					SurfaceAE( SurfNum2 ) = Surface( SurfNum2 ).Area * Construct( Surface( SurfNum2 ).Construction ).InsideAbsorpThermal;
```
mjwitte, line 1879: Again, this can't be precalculated if EMS actuates the thermal absorptance.  Since this is acting on a single zone now, it should simple enough to just have one loop that calculates both ZoneAESum and SumAET, exlcuding the weighted surface.

mjwitte (commented):
Some initial comments.

mjwitte:
@RKStrand Also, I merged in latest develop on this branch, so be sure to pull before making changes.

src/EnergyPlus/ThermalComfort.cc
RKStrand: @mjwitte Ah, I didn't realize that--good catch.  So much for efficiency when you have EMS!  I will fix this.

src/EnergyPlus/ThermalComfort.cc
RKStrand: @mjwitte Good point.  I will correct this.

src/EnergyPlus/ThermalComfort.cc
RKStrand: @mjwitte Got it.  Will fix this also.

src/EnergyPlus/ThermalComfort.cc
RKStrand: @mjwitte Ok, I will change this as well.  It shouldn't ever happen, but I get that if it does that no one wants a 20Gb file with this message repeated over and over and over again.

Myoldmopar:
@RKStrand as soon as you get his comments addressed I'll make one final pass over this and if all is well I'll get it merged in.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-005/occupant-thermal-comfort.tex
```diff
@@ -1041,66 +1041,66 @@ \subsubsection{References}\label{references-033}
...
 \endhead
 
-T\(_{r}\) & Mean radiant temperature & °C & - & RadTemp \tabularnewline
-T\(_{r-avg}\) & Zone averaged radiant temperature & °C & - & ZoneRadTemp \tabularnewline
-T\(_{surf}\) & Surface temperature & °C & - & SurfaceTemp \tabularnewline
-F\(_{surf}\) & Angle factor between person and surface & - & 0-1 & AngleFactor \tabularnewline
+T\(_{r}\) & Mean radiant temperature & °C & - \tabularnewline
+T\(_{r-avg}\) & Zone averaged radiant temperature & °C & -  \tabularnewline
+T\(_{surf}\) & Surface temperature & °C & -  \tabularnewline
+F\(_{surf}\) & Angle factor between person and surface & - & 0-1  \tabularnewline
```
mjwitte, line 1065: Minor detail, the nomenclature table is missing A and epsilon.

RKStrand:
@mjwitte @Myoldmopar Ok, I think I have addressed all of Mike's comments and have pushed another commit up.  Once the CI cycle completes I will take one more look at the results but I think this should be ready to go.  Thanks again for your help!

Myoldmopar (approved):

Myoldmopar:
I'll let the debug builds run before merging, but this seems good now.

Myoldmopar:
Fixed both the build warning, and the Linux build error (which was due to failing documentation build).  Not waiting for another pass of CI, this should be good.  Merging.

PR #6409, opened 2017-11-30
Fix up PCM unit test problem

Myoldmopar:
@mjwitte Could you test this out on your setup where you were seeing the failing test?  I believe the main problem was that the PCM `clear_state` function was written but never actually added to the `EnergyPlusFixture` class.  I also added a deallocation to be thorough.


mjwitte:
@Myoldmopar That fixes the problem locally.  Will merge in the morning assuming CI comes back clean.

mjwitte (approved):

Issue #6413, opened 2017-12-02
Bad primary air system controller index in SolveWaterCoilController


mjwitte:
Issue overview
--------------
Stumbled across this while addressing #6358. In [SimAirServingZone::SolveWaterCoilcontroller](https://github.com/NREL/EnergyPlus/blame/682bda2e1b9b62fa41965ddd84b23eb5b168035f/src/EnergyPlus/SimAirServingZones.cc#L2944) this line trips an array bounds crash,:
```
			PrimaryAirSystem( AirLoopNum ).ControlConverged( ControllerIndex ) = ControllerConvergedFlag;
```
The problem is that `ControllerIndex` is for the entire array of controllers, but `ControlConverged` is only allocated for the number of controllers on this particular air loop.  This may not be the only place that's using the wrong controller index.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus 8.9.0-plus #6358  - defect file will fail early without the 6358 fix.  Then it fails here.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@rraustad This code was added as part of #6197

rraustad:
That's what I get for testing with only a single air loop. We should require any changes be tested using multiple "whatever we are changing".

rraustad:
See PR #6414.

mjwitte:
Closed via #6414

PR #6414, opened 2017-12-03
Corrects improper controller index for OA system coils.

rraustad:
Pull request overview
---------------------
Recent code change introduced an issue with tracking OA System water controllers. When multiple water coil controllers are used on multiple air loops and controllers are also used in those air loop's OA Systems, the simulation will terminate with a crash. This change corrects Issue #6413.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2941,7 +2947,9 @@ namespace SimAirServingZones {
 
 			ManageControllers( ControllerName, ControllerIndex, FirstHVACIteration, AirLoopNum, iControllerOpIterate, ControllerConvergedFlag, IsUpToDateFlag, BypassOAController );
 
-			PrimaryAirSystem( AirLoopNum ).ControlConverged( ControllerIndex ) = ControllerConvergedFlag;
+			if ( HVACControllers::ControllerProps( ControllerIndex ).AirLoopControllerIndex > 0 ) {
+				PrimaryAirSystem( AirLoopNum ).ControlConverged( HVACControllers::ControllerProps( ControllerIndex ).AirLoopControllerIndex ) = ControllerConvergedFlag;
```
rraustad: This line right here, 2951 in this branch, is where it failed. This is an array of convergence for each air loop. But each air loop has multiple controllers AND can have an OA system with multiple controllers. So what the converged flag needs to be is an aggregate of all controllers on this air loop. I don't think this flag represents what it should. Seems it should be an .AND. of air loop and OA sys controllers.

rraustad: The fix was to save the index to the controller for each air loop. But this is confusing and I'm not sure this works correctly.

rraustad: Also, this can probably be done somewhere just once to speed things up. Not sure yet where that might be.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1141,15 +1141,15 @@ namespace SimAirServingZones {
 								}
 							}
 						}
```
rraustad: I commented out the below since a controller in the OA system is now simulated with the coil. This was added just before my correction so I'm not sure it's still needed. If a coil is in the OA system, should it be locked out if economizer is active AND the user told me to lock it out? I think yes.

rraustad: Oh, and I commented these lines out because there was an array bounds exception because AirSysNum exceeded the allocation of OutsideAirSys. My previous argument is just a good reason to keep it out.

mjwitte: The fix for this incorrect index variable is what started all of this.  You can grab the pending changes in #6413 if you want.  The check on OA sys HW coils was added by @Nigusse to fix a problem where an HW preheat coil didn't behave the same as an electric.

nrel-bot-3:

rraustad:
The only thing needed here is to uncomment the lines setting CanBeLockedOutByEcono=false for OA Sys water heating coils. Also, the defect file won't run with these lines uncommented so PR #6358 is a bottleneck.

rraustad:
This is ready for review.

rraustad:
I've got 1 unused variable so I will clean that up after I get some feedback.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -3616,6 +3616,37 @@ Label100: ;
 
 	}
 
+	void
+	GetControllerIndex(
+		std::string const & ControllerName, // name of coil controller
+		int & ControllerIndex // input actuator node number
+	) {
```
mjwitte: Preferable for this to be an integer function that returns the index directly.

src/EnergyPlus/HVACControllers.cc
```diff
@@ -3616,6 +3616,37 @@ Label100: ;
...
+		// SUBROUTINE INFORMATION:
+		//       AUTHOR         Richard Raustad, FSEC
+		//       DATE WRITTEN   January 2018
+
+		// PURPOSE OF THIS FUNCTION:
+		// This subroutine finds the controllers actuator node number
+
+		// Using/Aliasing
+		using InputProcessor::FindItemInList;
+
```
mjwitte: Delete `using`

mjwitte (commented):
@rraustad This looks good for the most part. Some minor comments and the unused variable to address.  Tested before and after and this fixes the issue. Only diffs are due to triggering a change in getinputs order for UnitarySystem_WaterCoils_wMultiSpeedFan, no numeric diffs.

src/EnergyPlus/HVACControllers.cc
rraustad: OK, done. But I don't know why changing this from void to int changes anything.

src/EnergyPlus/HVACControllers.cc
rraustad: If I delete this using statement I get a compiler error on FindItemInList. So leaving this until I get more feedback.

rraustad: I can't delete this using or don't know how. I get a compiler error on FindItemInList.

mjwitte: Fixed the using and did some other "cleanup".

@Myoldmopar can elaborate on why int/real functions are preferred over void (at least that's the direction I think I've been given before, so just parroting it - squawk!).

mjwitte (approved):
Ready to merge after CI finishes on these minor changes.

rraustad:
OK I get it.

mjwitte:
@Myoldmopar linux integration debug tests seem to be stuck.  It's been over 24 hrs since we've heard from Gort.

Myoldmopar:
OK.  We took it down last week to update some dependencies, but it seemed like it was coming back online fine.  Could get dicey what with ASHRAE and people unavailable.  I'll check on it though.

Issue #6233, opened 2017-08-09
Remove a section with ZoneMassFlowRate <= 0.0 in moisture balance equation


lgu1234:
Issue overview
--------------
#6119 corrected items related to exhaust mass flow rate with ZoneMassFlowRate > 0.0 in moisture balance equation. The section with ZoneMassFlowRate <= 0.0 is not needed and should be removed in order to be consistent with energy balance equation.
 
### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6415

PR #6415, opened 2017-12-11
Consolidate code in CorrectZoneHumRat

lgu1234:
Pull request overview
---------------------
Addresses #6233. Remove a surplus section of code for ZoneMassFlowRate <= 0.0 in moisture balance equation. The coefficient calculations in moisture balance equation are consistent with energy balance equation.

Small differences are found for an example file of GSHP-GLHE-CalcGFunctions.idf in local computer runs.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


nrel-bot-3:

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -4780,25 +4781,16 @@ namespace ZoneTempPredictorCorrector {
...
-		} else if ( ZoneMassFlowRate <= 0.0 ) {
-			B = ( LatentGain / H2OHtOfVap ) + ( ( OAMFL( ZoneNum ) + VAMFL( ZoneNum ) + CTMFL( ZoneNum ) + ExhMassFlowRate ) * OutHumRat ) + EAMFLxHumRat( ZoneNum ) + SumHmARaW( ZoneNum ) + MixingMassFlowXHumRat( ZoneNum );
-			A = OAMFL( ZoneNum ) + VAMFL( ZoneNum ) + EAMFL( ZoneNum ) + CTMFL( ZoneNum ) + ExhMassFlowRate + SumHmARa( ZoneNum ) + MixingMassFlowZone( ZoneNum );
-			if ( SimulateAirflowNetwork == AirflowNetworkControlMultizone || SimulateAirflowNetwork == AirflowNetworkControlMultiADS || ( SimulateAirflowNetwork == AirflowNetworkControlSimpleADS && AirflowNetworkFanActivated ) ) {
-				// Multizone airflow calculated in AirflowNetwork
-				B = ( LatentGain / H2OHtOfVap ) + SumHmARaW( ZoneNum ) + AirflowNetworkExchangeData( ZoneNum ).SumMHrW + AirflowNetworkExchangeData( ZoneNum ).SumMMHrW;
-				A = AirflowNetworkExchangeData( ZoneNum ).SumMHr + AirflowNetworkExchangeData( ZoneNum ).SumMMHr + SumHmARa( ZoneNum );
-			}
-			C = RhoAir * Zone( ZoneNum ).Volume * Zone( ZoneNum ).ZoneVolCapMultpMoist / SysTimeStepInSeconds;
+		ZoneOABal = MDotOA( ZoneNum );
```
mjwitte: What is `MDotOA`?  I only see it used in the moisture prediction/correction, but I don't find it in the temperature calculations.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -4682,6 +4682,7 @@ namespace ZoneTempPredictorCorrector {
 		int ADUNum;
 		int ADUInNode;
 		int ADUOutNode;
+		Real64 ZoneOABal;
```
mjwitte: Looks like `ExhMassFlowRate` and `TotExitMassFlowRate` (a few lines up) are not used for anything.  Can these be deleted?

mjwitte (commented):
@lgu1234 Where is the equivalent section for the energy balance?  Do I understand correctly that this is simply a code cleanup and there should be no diffs?
I merged in develop to bring this branch up to date.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte If you agree with my changes, I will remove these two variables: ExhMassFlowRate and TotExitMassFlowRate. 

I also notice that this issue also happens in CorrectZoneContaminants of the ZoneContaminantPredictorCorrector module.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte MDotOA is defined as Airbalance MASS FLOW rate in DataHeatBalFanSys, The equivalent item of MDotOA in the temperature calculation is MDotCPOA defined as Airbalance MASS FLOW * AIR SPECIFIC HEAT.

In order to avoid large differences in an example file, the following two lines of correction are used:

		ZoneOABal = MDotOA( ZoneNum );
		if ( ZoneMassFlowRate <= 0.0 ) ZoneOABal = 0.0;
However, the temperature calculation uses MDotCPOA as is without modification. Therefore, in order to be consistent with energy calculation, the correct may be removed.

lgu1234:
@mjwitte  The equivalent section is located in CalcZoneSums. Theoretically, the code after cleanup should not show differences. However, some calculations entered the section with MassFlowRate <=0. This should not happen. Therefore, if we use consistent code in both moisture and temperature, we may see some differences, but the revision is correct.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
mjwitte: @lgu1234 I still don't understand what "Airbalance mass flow rate" is.  Where does it come from?

lgu1234: @mjwitte Let me dig deeper to find out "MDotOA".

lgu1234: @mjwitte "MDotOA" is calculated with Air Balance Method = Quadrature in the ZoneAirBalance:OutdoorAir. It should be independent of ZoneMassFlowRate and calculated every time step. Therefore, the code in temperature calculation is correct, while the following two lines should be removed in moisture calculation, even though large differences are observed in SingleFamilyHouse_TwoSpeed_ZoneAirBalance.idf:

        ZoneOABal = MDotOA( ZoneNum );
	if ( ZoneMassFlowRate <= 0.0 ) ZoneOABal = 0.0;

mjwitte: @lgu1234 Thanks.  I forgot about that feature.  I would be good to expand the comment for MDotOA where it's declared.  Please go ahead and make this correction and remove the unused variables than I'll give it one final review.

lgu1234:
@mjwitte I uploaded the branch a while ago. Here is what I have done:
1. Remove two variables: ExhMassFlowRate and TotExitMassFlowRate
2. Remove a section to define MDotOA to be consistent with temperature calculation

Test run shows differences in SingleFamilyHouse_TwoSpeed_ZoneAirBalance as expected.

It is ready for final review.

mjwitte (approved):

Issue #6418, opened 2017-12-18
Crash when running sim with DualDuct terminals in Debug mode


jmarrec:
Issue overview
--------------

I'm trying to fix an OpenStudio-resources test. I realized that for `AirTerminal:DualDuct:VAV` and `AirTerminal:DualDuct:VAV:OutdoorAir` I don't have the autosized value for "Maximum Air Flow Rate" while I was clearly expecting it given [DualDuct.cc::SizeDualDuct](https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/DualDuct.cc#L1086).

I started running my file with E+ 8.8.0 built in Debug mode with a debugger, and I get an assertion error when running it.

See attached idf (rename extension), V8-8-0 (original)

[autosize_hvac.txt](https://github.com/NREL/EnergyPlus/files/1568344/autosize_hvac.txt)

And this file is the same but I ran it through the Transition utility to port it to V-8-9-0, which exhibits exactly the same problem (the DualDuct.cc didn't change since september/V8-8-0)

[autosize_hvac_V890.txt](https://github.com/NREL/EnergyPlus/files/1568374/autosize_hvac_V890.txt)

It throws on [DualDuct.cc#L228](https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/DualDuct.cc#L228)

DamperNum is 2, Damper ( DamperNum )  is Ok, Damper(DamperNum).ADUNum = 0 which is the problem, since 0 doesn't exist.

I'm pretty sure the problem is in GetDualDuctInput: on this line for eg it should be `DamperNum` and not `DamperIndex`: https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/DualDuct.cc#L484

I will try to change it and build again see if it solves it.

### Details
Some additional details for this issue (if relevant):
 - Platform: Ubuntu 16.04 LTS
 - Version of EnergyPlus: current develop (017868e84f0eb0de7711a942c72aebc273bafcc8), and 8.8.0 too.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

jmarrec:
Fixed in #6419

jmarrec:
Closed in #6419

PR #6419, opened 2017-12-18
Fix DualDuct VAV and VAV:OA: Wrong int index used, fix for #6418

jmarrec:
Pull request overview
---------------------

Fix for #6418: Found a mistake in DualDuct::GetDualDuctInput where the wrong int index was used.
The rationale is that there are three types of DualDuct terminals in there: CV, VAV and VAV:OutdoorAir. 
For each type, you loop on the number of objects with an int `DamperIndex`, but you put all the results in a single `Damper` array (CV first, then VAV, then VAVOA), so you use a second int DamperNum. 
In the routine, it was wrongly using DamperIndex, which lead to:
* Overwrite of existing dampers of another type
* Uninitialized ADUNum for some dampers, leading to an assertion (contains) error.

The key is finding this bug is that you need to have at least two different damper types in your file, so the overwriting actually happens.

The file I submitted in #6418 will fail to run properly (and it's way too big to make it a simulation tests anyways) but I have checked that it no longer crashes, and using a debugger I can check that the Damper array (especially the ADUNums) are fine.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - [ ] ExpandObjects changes?


jmarrec:
I unfortunately have no free time to implement a unit test or a regression test for this, but as I have stated above: the key is finding this bug is that you need to have at least two different dual duct damper types in your file (eg: `AirTerminal:DualDuct:ConstantVolume` and `AirTerminal:DualDuct:VAV:OutdoorAir`), so the overwriting actually happens.

mjwitte:
@jmarrec At first glance, these changes look correct.  Because the entire block(s) that are misusing `DamperIndex` use it everywhere in the block, the overwriting is putting the correct `ADUNum` in place for the earlier damper types, but the problem is that the later damper types end up with a zero.  I can pick this one up and finish it out with unit test and a smaller defect file.

jmarrec:
@mjwitte I pushed the start of a unit test. It isn’t working as such (commit message indicates where), but may be a helpful as a starting point. I had no time to troubleshoot why the input isn’t working.

Myoldmopar:
I'm not sure what to do on this one.  No one had time to write a unit test, but this is a clear fix and should go in.  I might comment out the start of the unit test that is currently there so I can uncomment it out and fix it so it actually tests.  Ugh.

PR #6434, opened 2018-01-13
32bit test

xuanluo113:


Myoldmopar:
For the record, I would probably have started by just changing the typedef for what "Real64" actually means from a double to a float.  That would avoid tons of code changes.  I never understood why we used Real64 for that alias name instead of like "RealEp", which would adapt as we changed floating point precision, but oh well.

xuanluo113:
@Myoldmopar 
That's how I did this now:
`typedef  double        Real32; // Platform-specific: C++ has no defined precision floating point types
typedef  float         Real64; // Platform-specific: C++ has no defined precision floating point types`

nrel-bot-3:
@XuanLuoLBL @lgentile it has been 14 days since this pull request was last updated.

Myoldmopar:
Per email with @XuanLuoLBL, I'm closing this PR and deleting the branch for now.

Issue #6396, opened 2017-11-20
User file with PTHP has negative heating electricity


mjwitte:
Issue overview
--------------
User file with PTHP shows a negative value for heating electricity.  Further investigation shows that DX heating coil ratings show a negative value for low temperature heating capacity:
```
! <DX Heating Coil Standard Rating Information>, Component Type, Component Name, High Temperature Heating (net) Rating Capacity {W}, Low Temperature Heating (net) Rating Capacity {W}, HSPF {Btu/W-h}, Region Number
 DX Heating Coil Standard Rating Information, Coil:Heating:DX:SingleSpeed, THERMAL ZONE: SPACE 101_WEST PTHP DX HEATING COIL, 115912.6, -16638.0, 5.40, 4
```
This points to the Heating Capacity Function of Temperature Curve as the problem.
```
Curve:Cubic,
  Thermal Zone: Space 102_South PTHP Heating CAPFT, !- Name
  0.876825,                               !- Coefficient1 Constant
  -0.002955,                              !- Coefficient2 x
  5.8e-005,                               !- Coefficient3 x**2
  0.025335,                               !- Coefficient4 x**3
  -5,                                     !- Minimum Value of x {BasedOnField A2}
  25;                                     !- Maximum Value of x {BasedOnField A2}

```
The DX heating coil object (or any coil for that matter) should reject a negative heating capacity and/or negative heating electricity.  In the case of a negative heating capacity, one option is to reset to zero and throw a recurring warning.  Negative electric power should probably be a severe/fatal error.

*Update*
The heating CAPFT curve is part of the problem - it contributes to the negative low-temp rated capacity, but adding outputs for curve values shows that this curve is not going negative during the simulation.  The negative power problem is from the Energy Input Ratio Function of Temperature Curve, "Curve Cubic 16". This curve is negative above 4.5C, and at 25C, the curve value is -140.
```
  Curve:Cubic,
    Curve Cubic 16,          !- Name
    0.704658,                !- Coefficient1 Constant
    0.008767,                !- Coefficient2 x
    0.000625,                !- Coefficient3 x**2
    -0.009037,               !- Coefficient4 x**3
    -5,                      !- Minimum Value of x
    25;                      !- Maximum Value of x
```

### Workaround
Fix the curve or add fields for Minimum Curve Output (and Maximum Curve Output) to keep the curve result within reasonable limits.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v8.8.0
[...]

PR #6441, opened 2018-01-23
User file with PTHP has negative heating electricity 

Nigusse:
Issue #6396

Addresses PTHP with negative heating electricity due to EIR modifier curve as function of temperature curve value being less than zero. This is caused by bad curve. This issue also impacts standard rating heating capacity for DX heating coils. 

Pull request overview
---------------------
This is a defect caused by bad EIR modifier curves. If any of the EIR or capacity modifier curves as a function of temperature results in negative value, then the curve value is set to zero and recurring warning error message is generated. Also severe warning error message for standard ratings calculation.
Diffs are not expected for this change.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
@mjwitte I check for the value of heating capacity and EIR function of temperature curves. If any of the curves value is less than zero, I set HSPF value to zero and throw error message. Also if the capacity modifier curves value are less than zero, I set the standard ratings net heating capacity value to zero. In both cases the simulation continues. Now I see 28 example files showing diffs in EIO, err and eplustbl.htm files. These diffs are causes by new error messages and HSPF values being set to zero. I will selectively investigate the curves in some of those idf with diffs.

tst/EnergyPlus/unit/StandardRatings.unit.cc
```diff
@@ -0,0 +1,232 @@
...
+		// evaluate EIR curves
+		Real64 EIRTempModFacRated = CurveValue( Coil.EIRFTemp( 1 ), StandardRatings::HeatingOutdoorCoilInletAirDBTempRated );
+		Real64 EIRTempModFacH2Test = CurveValue( Coil.EIRFTemp( 1 ), StandardRatings::HeatingOutdoorCoilInletAirDBTempH2Test );
+		Real64 EIRTempModFacH3Test = CurveValue( Coil.EIRFTemp( 1 ), StandardRatings::HeatingOutdoorCoilInletAirDBTempH3Test );
+		// check EIR curve value
+		EXPECT_LT( EIRTempModFacRated, 0.0 );
+		EXPECT_GT( EIRTempModFacH2Test, 0.0 );
+		EXPECT_GT( EIRTempModFacH3Test, 0.0 );
+		// if one of the CAP or EIR curves value is less than zero, then HSPF is set to zero
+		EXPECT_DOUBLE_EQ( HSPF, 0.0 );
```
Myoldmopar, line 230: :+1:

tst/EnergyPlus/unit/StandardRatings.unit.cc
```diff
@@ -0,0 +1,232 @@
...
+		Coil.DefrostTime = 0.058333;
+		Coil.DefrostCapacity = 1000;
+		Coil.PLRImpact = false;
+		Coil.FuelType = FuelTypeElectricity;
+		Coil.RegionNum = 4;
+		Coil.OATempCompressorOn = -5.0;
+		Coil.OATempCompressorOnOffBlank = "true";
+		NumCurves = 5;
+		PerfCurve.allocate( NumCurves );
+		PerfomanceCurveData * pCurve;
```
Myoldmopar, line 120: This is OK here.  This pointer isn't being used to store anything, it's just a reference.  For "reference", you don't need to do this complexity.  Just create an actual reference:

```
auto & thisCurve = PerfCurve( nCapfT );
thisCurve.CurveType = Cubic;
```

Then you don't have the sometimes yellow/red flag of a pointer variable during review.  The only downside I can think of is that you have to create multiple variable names below, one for each curve.  Unless you put them inside individual scopes (braces).

Anyway, I see your intent here, and it's OK for here, but in the future, consider the alternative approach.

Myoldmopar (approved):

Myoldmopar:
Built locally, confirmed the energy use is now zero with the curve values being constrained to 0.  I also merged develop and resolved the tiny conflict.  I'm pushing it up as soon as it builds successfully here.  If CI is happy, this will merge.

tst/EnergyPlus/unit/StandardRatings.unit.cc
Nigusse: @Myoldmopar  Thanks. Comment noted.

Myoldmopar:
I deleted the Mac results there and we'll see if my regression fixes help get rid of those audit diffs.

Myoldmopar:
As much as I hoped seeing the audit diffs go away, they haven't yet.  I think it might not get a new regression repo clone until it finishes its current set of known builds.  We'll see later.

Issue #6443, opened 2018-01-25
Coil:Cooling:DX:VariableSpeed shows max iteration warnings when supply air is very close to saturation curve


rraustad:
Issue overview
--------------
When users specify a capacity, airflow, and/or SHR that results in an outlet condition close to the saturation curve, the VS coil model may limit SHR to 1 when the bypass factor is < 0.001. So if the rated CBF (a model variable not a user input) is close to 0.001, as the model iterates on PLR, the coil bypass factor changes just enough to oscillate around 0.001. This causes a discontinuity in the supply air conditions and the iterative solution cannot converge.

The resulting warnings are:

    ** Warning ** CoilSystem:Cooling:DX - Iteration limit exceeded calculating DX unit sensible part-load ratio for unit = GUID11_SYS0SCC-1SYSTEM
    **   ~~~   ** Estimated part-load ratio  = 0.757
    **   ~~~   ** Calculated part-load ratio = 1.000
    **   ~~~   ** The calculated part-load ratio will be used and the simulation continues. Occurrence info:
    **   ~~~   **  Environment=ANNUAL COOLING DESIGN CONDITION6, at Simulation time=07/21 08:00 - 08:20

And occur frequently:

    *************  ** Warning ** CoilSystem:Cooling:DX "GUID11_SYS0SCC-1SYSTEM" - Iteration limit exceeded calculating sensible part-load ratio error continues. Sensible PLR statistics follow.
    *************  **   ~~~   **   This error occurred 35277 total times;
    *************  **   ~~~   **   during Warmup 0 times;
    *************  **   ~~~   **   during Sizing 0 times.
    *************  **   ~~~   **   Max=1.  Min=1.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
See PR #6444.

Myoldmopar:
Closed via #6444

PR #6444, opened 2018-01-25
Correct variable speed DX cooling coil SHR calculation when coil bypass factor < 0.001

rraustad:
Pull request overview
---------------------
Variable speed DX coil shows max iteration warnings when supply air condition is very near the saturation curve. A model variable, in this case the coil bypass factor, would be near 0 or < 0.001. At these times the model incorrectly applies an SHR of 1. This change corrects issue #6443.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/HVACDXSystem.cc
```diff
@@ -1824,6 +1824,7 @@ namespace HVACDXSystem {
 
 									if ( ( TempSpeedOut - TempSpeedReqst ) < Acc ) {
 										SpeedNum = I;
+										FullOutput = TempSpeedOut; // save in case warning message is reported
```
rraustad, line 1800: This and next are incidental changes for warning messages seen during testing. Just correcting them here since they are minor.

src/EnergyPlus/VariableSpeedCoils.cc
```diff
@@ -6491,53 +6464,44 @@ namespace VariableSpeedCoils {
 
 			}
 
-			if ( CBF < 0.001 ) {
-				SHRCalc = 1.0;
-				LoopOn = false;
-			}
-			else {
```
rraustad, line 6597: The problem is here where if CBF is small, the SHR is set to 1. This is just wrong and SHR has nothing to do with coil bypass factor. This created a discontinuity that caused RootFinder to fail on max iterations.

nrel-bot-3:

Issue #6427, opened 2018-01-05
PIU doesn't set the ADU inlet node num


lymereJ:
Issue overview
--------------
The air distribution inlet node num is initialized [here](https://github.com/NREL/EnergyPlus/blob/973b46637653ab47301b7c9e8d29de06b8e406ad/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L323) but is never updated. This affects the calculations done by the mechanical ventilation controller. The supply mass flow rate passed to the mechanical ventilation controller is never incremented by the primary mass flow of the PIU units, hence an incorrect system supply mass flow rate is used for the calculations of the ventilation flow rates (for example, VRP).

@mjwitte and I exchanged a few messages via the helpdesk, his last message summarizes the issue pretty well:

>Looks like the PIU has *never* set the ADU inlet node num.
>
>Looking at the code prior to v8.8 changes, this line is there as a comment in PoweredInductionUnits::GetPIUs, was line 596 then
>
>// AirDistUnit(ADUNum)%InletNodeNum = PIU(PIUNum)%InletNodeNum
>
>https://github.com/NREL/EnergyPlus/blame/bb16db32963d6e56325dfd51297e42a75e240ff3/src/EnergyPlus/PoweredInductionUnits.cc#L596
>
>
>For v8.8, this block was moved (by me) and copied to a second place.
>
>https://github.com/NREL/EnergyPlus/commit/8d2cbf82be0358f0977d58148fe56c89e1e76a50#diff-a649c9900f7e0de68c516ad1c22b01c7R441
>https://github.com/NREL/EnergyPlus/commit/8d2cbf82be0358f0977d58148fe56c89e1e76a50#diff-a649c9900f7e0de68c516ad1c22b01c7R587
>
>The commented line was deleted in one place, still there in the other. Little did I know that this line was needed all along, except with PIU( PIUNum ).PriAirInNode.
>
>Looking back in the fortran version archive, this line was commented out from the very beginning. I can't explain why. 

### Details
 - Helpdesk ticket number: 12378

Nigusse:
I tried two things for this issue:

(1) First I attempted to set node number to the air distribution inlet node. This is not straightforward because, unlike other air terminal units, PIUs have two inlet nodes and the air distribution code structure is not setup to handle two inlet nodes. Any ways I managed to connect to the PIU's mixed mixed air outlet node but that did not make any difference on the problem reported. 

[...]

PR #6446, opened 2018-01-26
Incorrect primary air mass flow rate in PIU due to missing inlet node and airloop number

Nigusse:
Issue #6427  

This is a defect in PIU and ADU units. In series and parallel PIU air terminal unit the primary air let node number was initialized to zero but never set to a node number. As the result the Air Distribution Unit (ADU) inlet node was always zero. Furthermore, the air loop supply flow rate variable cannot be updated because the PIU did not set the airloop number to which the PIU is connected. This fix addresses these two issues. Example files with PIU air terminal unit may show diffs.
 
### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:


Nigusse:
The inlet node number of Air distribution Unit (ADU) and the AirLoop number to which the PIU is connected were not set, they were always zero. As the result the primary air flow rate could not be updated correctly. This fix sets the appropriate ADU inlet node and AirLoop numbers. Now the primary air flow rate fed to the PIU and ADU primary air inlet node can be updated correctly. Example idf files may show diffs.

Nigusse:
Example files with PIU air terminal units only showed diffs. Interestingly all of them had small diffs.

Myoldmopar:
Pulled develop in, I think @nrel-bot-2 will be happier with it now.  I'll review it again once it is happy.

Issue #6448, opened 2018-01-29
UnitarySystem with Coil:Cooling:DX:VariableSpeed does not cycle the fan with compressor when cycling fan mode is used


rraustad:
Issue overview
--------------
Test of the VS DX cooling coil shows the fan operating at max speed when the coil is cycling and fan operating mode is cycling fan. Comparison of Fan Runtime Fraction and Unitary System Part Load Ratio shows the issue. A plot of the Unitary System Part Load Ratio and Cooling Coil Part Load Ratio shows that the coil and unitary system have the same part load ratio.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (DevSupport 6448 w/ Chicago weather)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
The defect file shows the issue where the fan is operating at max speed instead of cycling with the coil/compressor. File set up shown as:

    Schedule:Constant,
      SYS0SZUHC-1FanOpSchedule_0,             !- Name
      AnyValue,                               !- Schedule Type Limits Name
      0.0000;                                 !- Hourly Value

    AirLoopHVAC:UnitarySystem,
      SYS0SZUHC-1FanOpSchedule_0,             !- Supply Air Fan Operating Mode Schedule Name


![6448-fanplrequals1withcyclingfan](https://user-images.githubusercontent.com/5778156/35514183-460ccfa6-04fd-11e8-8227-0cfe0db1024a.jpg)

rraustad:
See PR #6449.

Myoldmopar:
Fixed via #6449

PR #5519, opened 2016-03-03
Corrects air flow defect in AirloopHVAC:UnitarySystem when MS coils operate at Speed = 1

rraustad:
## Pull request overview

Addresses #5518  There are 2 changes associated with this issue.

The first is that the DX MS cooling coil has a discontinuity between 1st and 2nd stage. This is due to the calculation of capacity using hfg for speed 2.

The second change is in regards to the internal variable MSHPMassFlowRateLow is currently set to 0 at Speed = 1. This variable should instead be set proportional to PLR. This apparently only affects the TotCapFlowModFac calculation and causes a discontinuity between capacity at Speed=1 and Speed=2. Minor changes will be made to correct how this internal variable is set. Diff's are expected only in UnitarySystem files which use MS coils, and in MSHP files using the DX MS cooing coil.
- Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
### Review Checklist

This will not be exhaustively relevant to every PR.
- [x] Code style (parentheses padding, variable names)
- [x] Functional code review (it has to work!)
- [x] If defect, results of running current develop vs this branch should exhibit the fix
- [x] CI status: all green or justified
- [x] Performance: CI Linux results include performance check -- verify this
- [x] Unit Test(s)
- C++ checks:
  - [x] Argument types
  - [x] If any virtual classes, ensure virtual destructor included, other things
- IDD changes n/a

Issue #5518, opened 2016-03-03
Multispeed Coils used with AirloopHVAC:UnitarySystem do not control air flow properly at Speed = 1


rraustad:
## Issue overview

Please change this line to a description of the defect, with useful supporting information including how the issue was found, conditions that trigger this defect, workarounds for users of the current version, and a clear description of how to validate that a fix is complete for this defect.
### Details

Some additional details for this issue (if relevant):
- Platform (Operating system, version)
- Version of EnergyPlus (if using an intermediate build, include SHA)
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (list location of defect file here)
- [x] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #5519

PR #6449, opened 2018-01-29
Correct VS coil air flow and fan RTF calculations

rraustad:
Pull request overview
---------------------
The Variable Speed DX coil model uses an idle air flow rate to select air flow when the compressor is off and the fan operates continuously. When cycling fan is used and the compressor operates at low speed, the off air flow rate should = 0.  This allows the model to turn the compressor and fan on and off in unison. This change correctly uses 0 during cycling fan operation when low compressor speed is used. Cycling losses are also accounted for using the standard adjustment from part load ratio to run time fraction as RTF = PLR / PLF, where PLF is the adjustment factor to account for cycling losses. This corrects issue #6448.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
Fan now operates at correct RTF and accounts for global OnOffFanPartLoadFraction. Here fan RTF should be greater than coil PLR by RTF = PLR / PLF. The PLF term was not being passed to the fan by the VS coil model. Also, the wrong OFF air flow rate was being used by the Unitary System model.

6448-fanplrequals1withcyclingfan_fixed

Myoldmopar:
Built locally and compared to develop results.  Confirmed the better behaved fan.  I'll review the code here, then merge develop in and let CI have one more pass before merging.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -3776,6 +3776,445 @@ TEST_F( EnergyPlusFixture, UnitarySystem_VarSpeedCoils ) {
...
+	EXPECT_NEAR( ZoneSysEnergyDemand( ControlZoneNum ).RemainingOutputRequired, Qsens_sys, 1.0 ); // Watts
+	EXPECT_DOUBLE_EQ( Node( InletNode ).MassFlowRate, ( UnitarySystem( 1 ).CoolMassFlowRate( UnitarySystem( 1 ).CoolingSpeedNum ) * UnitarySystem( 1 ).PartLoadFrac ) );
+	EXPECT_DOUBLE_EQ( Node( InletNode ).MassFlowRate, Node( OutletNode ).MassFlowRate );
+
+	// compare fan RTF with fan PLR and global PLF
+	FanPLR = Node( InletNode ).MassFlowRate / Fans::Fan( 1 ).MaxAirMassFlowRate;
+	FanRTF = FanPLR / DataHVACGlobals::OnOffFanPartLoadFraction;
+	EXPECT_GT( FanRTF, FanPLR );
+	EXPECT_DOUBLE_EQ( FanRTF, Fans::Fan( 1 ).FanRuntimeFraction );
+	EXPECT_NEAR( DataHVACGlobals::OnOffFanPartLoadFraction, 0.9004014, 0.0000001 );
```
Myoldmopar: :+1:

src/EnergyPlus/VariableSpeedCoils.cc
```diff
@@ -4247,12 +4247,14 @@ namespace VariableSpeedCoils {
...
 
 		RuntimeFrac = 1.0;
+		OnOffFanPartLoadFraction = 1.0;
 		VarSpeedCoil( DXCoilNum ).RunFrac = 1.0;
 		if ( ( SpeedNum == 1 ) && ( PartLoadRatio < 1.0 ) ) {
 			PLF = CurveValue( VarSpeedCoil( DXCoilNum ).PLFFPLR, PartLoadRatio );
 			if ( PLF < 0.7 ) {
 				PLF = 0.7;
 			}
+			OnOffFanPartLoadFraction = PLF; // save PLF for fan model
```
Myoldmopar: :+1:

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -11300,18 +11292,10 @@ namespace HVACUnitarySystem {
...
-							CompOffMassFlow = 0.0; // #5518
-							CompOffFlowRatio = 0.0;
-						} else {
-							CompOffMassFlow = UnitarySystem( UnitarySysNum ).IdleMassFlowRate;
-							CompOffFlowRatio = UnitarySystem( UnitarySysNum ).IdleSpeedRatio;
-						}
+				} else { // cycling fan mode
+					if ( CoolSpeedNum <= 1 ) {
+						CompOffMassFlow = 0.0; // #5518
+						CompOffFlowRatio = 0.0;
```
Myoldmopar, line 10896: I guess you decided that speednum didn't need to be split out like this anymore? (https://github.com/NREL/EnergyPlus/pull/5519/files#r54980671)

Myoldmopar (approved):

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: Yes, the VS and MS coils should behave the same at speed = 1 for cycling fan mode. #5518 addressed a multi-speed coil issue and I didn't think much about the VS coil at that time until I saw this same problem with VS coils. I then concluded that VS and MS should be treated the same when speed = 1 and the fan cycles (i.e., should cycle between x m3/s and 0 if cycling fan, not x m3/s and idle air flow rate).

Myoldmopar: :+1:

Myoldmopar:
@rraustad I merged develop into this branch, and fixed a couple things in the new unit test keeping it from building (UnitType instead of UnitarySystemType member variable, etc.).

I haven't pushed my changes up yet, however, because the test is failing locally.  

All three of these last tests are failing:
```
EXPECT_GT( FanRTF, FanPLR );
EXPECT_DOUBLE_EQ( FanRTF, Fans::Fan( 1 ).FanRuntimeFraction );
EXPECT_NEAR( DataHVACGlobals::OnOffFanPartLoadFraction, 0.9004014, 0.0000001 );
```

The failures are:
```
Expected: (FanRTF) > (FanPLR), actual: 0.135147 vs 0.135147
```
- So basically the RTF is now equal to the PLR, whereas you expcted it to be higher

```
Value of: Fans::Fan( 1 ).FanRuntimeFraction
Actual: 0.15009620585213448
Expected: FanRTF
Which is: 0.1351468387101242
```
- Again, the run time frac is now .15, whereas you were expecting it to be .135

```
The difference between DataHVACGlobals::OnOffFanPartLoadFraction and 0.9004014 is 0.099598599999999982, which exceeds 0.0000001
```
- Where OnOffFanPartLoadFraction is 1.0 now, and you expected it to be 0.9.

I had thought I'd just figure this out before pushing my branch, but I think it'll be better now if I just push it up.  You should be able to Git Pull (assuming GitHub is working again) and build it and see the same results.

rraustad:
They should be different, I'll take a look.

rraustad:
Works here.

fanplrrtf

rraustad:
Now I am not sure which branch I have. I can't pull from Git and I don't see the new SZVAV model. So stand by.

Myoldmopar:
And this was after the merge from develop into this branch?
-- 
Edwin Lee, PhD
Engineer, Commercial Buildings Research Staff
National Renewable Energy Laboratory

Myoldmopar:
Ahh ok. Carry on and let me know.
-- 
Edwin Lee, PhD
Engineer, Commercial Buildings Research Staff
National Renewable Energy Laboratory

rraustad:
giterror

Myoldmopar:
Same thing gu and Mike saw today. It's nothing we've done. I'll try to find
a solution still though.
-- 
Edwin Lee, PhD
Engineer, Commercial Buildings Research Staff
National Renewable Energy Laboratory

rraustad:
I can't pull from the Trane Git repository either. Same error message.

shorowit:
https://stackoverflow.com/questions/48938019/git-pull-push-unable-to-access-https-ssl-routines-seem-to-be-down

mjwitte:
@rraustad @Myoldmopar @lgu1234 It's this
https://githubengineering.com/crypto-removal-notice/
which I found here
https://stackoverflow.com/questions/48950971/tortoisegit-ssl-tlsv1-alert-protocol-version-issue
Not sure yet if the suggested fix will do any good.

rraustad:
Update Git did not work, this did:
https://git-scm.com/download/win

rraustad:
@shorowit thank you, thank you, thank you!

rraustad:
What the heck is this?

    4>C:\MyProjects\EnergyPlus_CPlus4\src\EnergyPlus\ConfiguredFunctions.cc(54): error C2084: function 'std::string EnergyPlus::configured_source_directory(void)' already has a body

    std::string configured_source_directory()
    {
        return ("C:/Users/Richard/Documents/GitHub/EnergyPlusTeam");
    }

Myoldmopar:
That's the multiply defined problem that is something I caused. Sort of. I
changed a file around and cmake can't figure out what to do with it. Delete
your build directory, start a new one, run cmake, and you should be good to
go again.
-- 
Edwin Lee, PhD
Engineer, Commercial Buildings Research Staff
National Renewable Energy Laboratory

rraustad:
Ran CMake to a new build directory and everything worked fine.

src/EnergyPlus/VariableSpeedCoils.cc
```diff
@@ -4362,7 +4362,7 @@ namespace VariableSpeedCoils {
 			if ( PLF < 0.7 ) {
 				PLF = 0.7;
 			}
-			OnOffFanPartLoadFraction = PLF; // save PLF for fan model
+            OnOffFanPartLoadFraction = PLF; // save PLF for fan model
```
rraustad: WTF.. my editor is taking over my formatting.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -4422,10 +4422,11 @@ TEST_F( EnergyPlusFixture, UnitarySystem_VarSpeedCoils_CyclingFan ) {
...
 	FanPLR = Node( InletNode ).MassFlowRate / Fans::Fan( 1 ).MaxAirMassFlowRate;
-	FanRTF = FanPLR / DataHVACGlobals::OnOffFanPartLoadFraction;
-	EXPECT_GT( FanRTF, FanPLR );
-	EXPECT_DOUBLE_EQ( FanRTF, Fans::Fan( 1 ).FanRuntimeFraction );
-	EXPECT_NEAR( DataHVACGlobals::OnOffFanPartLoadFraction, 0.9004014, 0.0000001 );
+    // blow thru fan resets OnOffFanPartLoadFraction = 1 so other equipment not using PLF are not affected. OnOffFanPartLoadFraction = 1 here.
+    // Unitary System also sets OnOffFanPartLoadFraction = 1, maybe too agressive (see end of ReportUnitarySystem) so this variable will = 0
+    // issue with calling blowthru fan again needs to be resolved, seems excessive.
+	EXPECT_EQ( 1.0, DataHVACGlobals::OnOffFanPartLoadFraction );
+	EXPECT_GT( Fans::Fan( 1 ).FanRuntimeFraction, FanPLR );
```
rraustad: I'm not sure when the UnitarySystem started setting OnOffFanPartLoadFraction = 1 but it must have been after this (and I did it). If blow thru fan, then someone needs to reset that flag so that other models don't get impacted. See last line of ReportUnitarySystem (if it weren't reset here then the next OnOff fan would use this value). So that means that if you call UnitarySystem, then OnOffFanPartLoadFrac = 1, but the FAN should hold the correct RTF.

rraustad:
Differences are exactly what they should be. Comparing E+ V8.8 (left plot) with this branch (right plot)  shows the results are still the same with the only difference in fan power (and miniscule change in compressor speed in cooling mode, not shown). When PLR = 1 no difference, when PLR < 1 OnOffFanPartLoadFraction has an impact on fan power/energy and fan heat therefore compressor speed.
The difference is only seen in cooling mode since on the heating design day the compressor is off on low temp cutout.

heatpumpvsasfixed

rraustad:
@Myoldmopar is something really happening on the MacOS or is the CI not working correctly there? All other tests pass as expected.

Myoldmopar:
Hmm.  I'm not sure.  Marvin has been doing a great job since @kbenne got him going again.  I deleted the results and we'll see if he can give it another try.  I'll recheck in the morning.  If it is repeatable by CI, I'll try a build on my Mac and see what's up.

rraustad:
10:1 it passes...

Myoldmopar:
That was successful right?

rraustad:
I haven't seen any new CI results. Those above look like the older results (e.g., replied 23 or 16 hours ago for MacOS). The one 16 hours ago could be the one from the previous commit and it just took a while. I'm good with these changes but I am still curious as to the answer with the CI.

Myoldmopar:
Both of these messages above are from Mac CI and on the same commit:

image

On the top (first) one, Mac had an odd failure, so I deleted those CI results, and CI re-ran that same commit.  This time the anomaly was gone, and the bottom result shows that it actually ran, but with some diffs.  Those diffs look very similar to the Windows and Linux results also showing up for that commit.

rraustad:
OK then successful. I didn't realize that one was new.

Myoldmopar:
Cool.  So you are done and this is ready for final review now?

rraustad:
Yes, thank you.

rraustad:
Just reviewed again, these changes are good to go.

Myoldmopar:
OK cool, I've got the Pivotal ticket on my list, I'll do it soon.  Thanks!

PR #6451, opened 2018-01-30
PVWatts

nmerket:
Pull request overview
---------------------
This change adds PVWatts from NREL's SAM model to EnergyPlus. It makes it available as `Generator:PVWatts` and `ElectricLoadCenter:Inverter:PVWatts`. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


doc/engineering-reference/src/on-site-generation/photovoltaic-arrays.tex
```diff
@@ -1,10 +1,10 @@
 \section{Photovoltaic Arrays}\label{photovoltaic-arrays}
 
-The Photovoltaics.f90 module includes three different models referred to as ``Simple'', ``Equivalent One-Diode'' and ``Sandia'' and the choice will determine the mathematical models (and input data) used to determine the energy produced by solar/electric conversion panels. The EnergyPlus photovoltaic array models are called one at a time at the HVAC system timestep along with other electrical generation components such as gas turbines and diesel engines.
+The Photovoltaics module includes three different models referred to as ``Simple'', ``Equivalent One-Diode'' and ``Sandia'' and the choice will determine the mathematical models (and input data) used to determine the energy produced by solar/electric conversion panels. There is also a PVWatts generator which abstracts away specifying PV modules and array properties for a simpler input structure. The EnergyPlus photovoltaic array models are called one at a time at the HVAC system timestep along with other electrical generation components such as gas turbines and diesel engines.
```
Myoldmopar, line 3: Yay no more f90!

doc/input-output-reference/src/overview/group-electric-load-center-generator.tex
```diff
@@ -64,7 +64,7 @@ \subsubsection{Inputs}\label{inputs-014}
 
 \paragraph{Field: Full Load Temperature Rise}\label{field-full-load-temperature-rise}
 
-This numeric field defines the temperature rise of the windings above the ambient temperature, when the transformer is loaded at its nameplate rating. For liquid-immersed distribution transformers, the full load temperature rise usually takes the value of 65 °C. For dry-type transformers, the temperature rise usually takes three optional values: 80, 115 and 150 °C. This field accepts any value between the minimum (50 °C) and the maximum (180 °C). The default is 150 °C.
+This numeric field defines the temperature rise of the windings above the ambient temperature, when the transformer is loaded at its nameplate rating. For liquid-immersed distribution transformers, the full load temperature rise usually takes the value of 65 \textdegreeC. For dry-type transformers, the temperature rise usually takes three optional values: 80, 115 and 150 \textdegreeC. This field accepts any value between the minimum (50 \textdegreeC) and the maximum (180 \textdegreeC). The default is 150 \textdegreeC.
```
Myoldmopar, line 67: Hmm.  Where is textDegreeC declared?  Guess I'll see as I go along here.

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -1939,6 +1960,10 @@ namespace EnergyPlus {
 			generatorType = GeneratorType::pV;
 			compGenTypeOf_Num = DataGlobalConstants::iGeneratorPV;
 			compPlantTypeOf_Num = DataPlant::TypeOf_PVTSolarCollectorFlatPlate;
+				} else if ( InputProcessor::SameString( objectType, "Generator:PVWatts" ) ) {
```
Myoldmopar: Something went awry here on the indentation it appears.

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -2026,6 +2051,14 @@ namespace EnergyPlus {
 			thermalPowerOutput = thermProdRate;
 			break;
 		}
+		case GeneratorType::pvWatts: {
+			PVWatts::PVWattsGenerator &pvwattsGenerator(PVWatts::GetOrCreatePVWattsGenerator(name));
```
Myoldmopar, line 2055: This name isn't passed around every simulation call right?  Kinda looks like it is, and if so, we need to change that to have this object keep a reference to the correct generator and just call it each time.  We won't look up the object every single call.  

If this is a one-time thing and I'm just reading it wrong, then nothing to change.

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -2276,6 +2331,21 @@ namespace EnergyPlus {
 		qdotRadZone_             = 0.0;
 	}
 
+	void
+	DCtoACInverter::setPVWattsDCCapacity(const Real64 dcCapacity)
+	{
+		if ( modelType_ != InverterModelType::pvWatts ) {
+			ShowFatalError("Setting the DC Capacity for the inverter only works with PVWatts Inverters.");
+		}
+		ratedPower_ = dcCapacity / pvWattsDCtoACSizeRatio_;
```
Myoldmopar, line 2340: I don't think we have ever fully resolved our pattern for private object variable naming.  The trailing underscore isn't terrible.  I think we should let this go in as one possible example and then as another class or two come along, make a decision based on those examples.

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -2294,6 +2364,12 @@ namespace EnergyPlus {
 		return aCEnergyOut_;
 	}
 
+	DCtoACInverter::InverterModelType
+	DCtoACInverter::modelType() const
+	{
+		return modelType_;
+	}
```
Myoldmopar, line 2371: I do feel concerned about unnecessary use of getters and setters.  I love using getters and setters in library code where you are concealing the underlying simulation logic.  In EnergyPlus, where our goal should always be to maximize performance, I don't want arbitrary getters/setters to accidentally, and unnecessarily, take clock cycles.

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -2554,32 +2653,32 @@ namespace EnergyPlus {
...
 	{
 		return thermLossRate_;
 	}
 
-	Real64 
+	Real64
 	ACtoDCConverter::dCPowerOut() const
 	{
 		return dCPowerOut_;
 	}
```
Myoldmopar, line 2673: I know you didn't write these with your PR here.  But my comment about unnecessary getters holds.  More importantly, the casing of these is gross.  dC and aC?  Should be: getDCPowerOut and getACPowerOut, etc.

src/EnergyPlus/PVWatts.hh
```diff
@@ -0,0 +1,430 @@
...
+		Real64 sazi;
+		Real64 rot;
+		Real64 btd;
+		Real64 ibeam;
+		Real64 iskydiff;
+		Real64 ignddiff;
+		int sunup;
+	};
+
+	// From here down is from the SAM Simulation core library
```
Myoldmopar: So...if the SAM simulation core library has an update, how does that get patched here?  Is the SAM core library available as something we could've instead dropped into the `third_party/` folder so updates could be dropped right in? (Assuming their API is stable, of course.)

tst/EnergyPlus/unit/PVWatts.unit.cc
```diff
@@ -0,0 +1,283 @@
...
+	});
+	ASSERT_FALSE(process_idf(idfTxt));
+	auto eplc(ElectPowerLoadCenter(1));
+	ASSERT_TRUE(eplc.inverterPresent);
+	EXPECT_DOUBLE_EQ(eplc.inverterObj->pvWattsDCCapacity(), 4000.0);
+	DataHVACGlobals::TimeStepSys = 1.0;
+	eplc.inverterObj->simulate(884.018);
+	EXPECT_NEAR(eplc.inverterObj->aCPowerOut(), 842.527, 0.001);
+
+}
```
Myoldmopar, line 324: Overall the unit test approach looks great.  The calc test seems light, I'm curious how much of the code you are actually testing with it.  Looks like 47%.  You aren't covering the code for thin film, one axis backtracking, two axis, or surface geometry type.  Adding in some examples of those looks like it would push coverage of your new code up to 80%-ish.

Myoldmopar (changes requested):
Could you respond to some comments I made on here?  The three primary ones are the one about the object name being possibly passed around each simulation call, the SAM core library code, and the amount of unit testing.  Honestly this does look like it's in really good shape though.  There's at least one minor indentation issue to be fixed, so I will mark this as "Request Changes" and we can resolve it through discussion.

src/EnergyPlus/ElectricPowerServiceManager.cc
nmerket: No, we're passing the name of the generator and looking it up each time. In my defense, look at the other `Sim*Generator` calls on the preceding and following lines. They're all doing the same thing, except redundantly passing both the index and the name. I was trying to make my class fit into that pattern.

src/EnergyPlus/ElectricPowerServiceManager.cc
nmerket: Yeah, I went with the trailing underscore because that's what was used in the Electric Load Center Distribution class.

src/EnergyPlus/ElectricPowerServiceManager.cc
nmerket: Maybe I've spend too much time in OpenStudio development where prolific getters and setters are the reason for getting out of bed in the morning. They're not generally being used, except in the unit tests. It makes it so I don't have to make all the member variables public for inspection. If it's that important to you I can remove these and make the member variables public or protected with the test fixture being a friend class.

src/EnergyPlus/ElectricPowerServiceManager.cc
nmerket: Yeah, that is ugly. I'm wishing my IDE hadn't "cleaned up" the trailing whitespace and you would have never seen it.

src/EnergyPlus/ElectricPowerServiceManager.cc
Myoldmopar: Yeah, I get it, it's a bit of a balancing act, between continuing E+ patterns and breaking them to make the whole thing better.  I'm not going to make you change it for this.  In the future, keeping a reference to an object and only looking it up once then directly operating on the reference is my preferred path forward.

tst/EnergyPlus/unit/PVWatts.unit.cc
nmerket: Fair enough.

src/EnergyPlus/ElectricPowerServiceManager.cc
nmerket: Yep. Thought I caught all of those.

src/EnergyPlus/ElectricPowerServiceManager.cc
Myoldmopar: Ugly, but nothing you added.  If you want to use your IDE to rapidly rename them then that would be great.  But it's not directly in your scope, so your choice.

tst/EnergyPlus/unit/PVWatts.unit.cc
Myoldmopar: Frankly, if you can add a couple cases, to cover your variations, this thing will become a shining example of a unit tested new feature.  The bar will be set...

Myoldmopar:
Oh, and it looks like custom_check is failing due to license year issues.  Make sure the license year is 2018 in any new files you added here.

nmerket:
@Myoldmopar, I addressed the following:

- Moved the PVWatts code from SSC into separate files. It's much cleaner that way.
- Added some more unit tests for `PVWattsGenerator.calc`.
- Fixed the indentation in `ElectricPowerServiceManager.cc`
- Changed the year to 2018 in all my added files.

I think that should do it. Let me know if there's anything else.

Myoldmopar:
Thanks @nmerket! This looks great. I pulled the changes, built, and tested locally. Everything looks good. I'm excited to see the coverage of the PVWatts file now, it should be way better.  As soon as CI is also happy about it I'll merge.

PR #6452, opened 2018-01-30
Reduce external interface maximum variables

kbenne:
A recent increase in the maximum number of allowed variables caused in integer to overflow causing the program to crash. Ref #6339 
See here: https://github.com/NREL/EnergyPlus/blob/2a735636b95cbc73b3e133b6d3319a3db8cc582a/src/EnergyPlus/ExternalInterface.cc#L544

Pull request overview
---------------------

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ExternalInterface.cc
+++ b/src/EnergyPlus/ExternalInterface.cc
@@ -108,7 +108,7 @@ namespace ExternalInterface {
 	std::string FMURootWorkingFolder; // FMU root working folder
 
 	// MODULE PARAMETER DEFINITIONS:
-	int const maxVar( 10240000 ); // Maximum number of variables to be exchanged
+	int const maxVar( 100000 ); // Maximum number of variables to be exchanged
 	int const maxErrMsgLength( 10000 ); // Maximum error message length from xml schema validation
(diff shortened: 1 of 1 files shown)


kbenne:
@tsnouidui I reduced the size of maxVars to 100000 so that the ```int lenXmlStr``` is not initialized to a number that it is not capable of holding, but I think my fix is still poor quality and brittle. The max size of int is platform dependent so it might be better to use std::numeric_limits<int>::max() but this get tricky because maxVars is multiplied by MaxNameLength. 

In any case this change makes ExternalInterface work for me on Linux 64 bit. Prior to this change E+ was killed at the line I referenced in the commit message.

Myoldmopar:
@kbenne did you ever get any feedback from @tsnouidui about this?

tsnouidui:
@Myoldmopar Yeah I gave feedback to @kbenne and said that the fix is OK.

Myoldmopar:
Works for me.

PR #6454, opened 2018-01-31
New Coil Sizing Reports

mjwitte:
Pull request overview
---------------------
Add two new table output reports:
Coil Sizing Details Report - Detailed coil sizing data table contributed by Trane. Report key is "CoilSizingDetails"
Coil Sizing Summary subtable in the HVAC Sizing SummaryReport. This is a subset of the Coil Sizing Details Report.

For complete code and documentation changes, see also #6500,

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-3:

Myoldmopar:
Custom check is reporting that the license year is bad in the ReportCoilSelection.* files.

The signed integer warning is because vector.size() returns an "unsigned" integer value, but you declared your loop variable as an integer.  Although this won't cause any problems for this particular case, the compiler still warns about comparing a signed versus unsigned integer. You can either make your loop variable a size_t, instead of a plain integer, or you can cast the result of the size() call to a regular int.  Let me know if that is super muddy.

The reorder warning is because coilCapFTIdealPeak( 1.0 ) is defined before coilRatedTotCap in the class definition, but initialized after it in the list of constructor initializations.  This won't cause a problem for a simple data type like this, but the compiler still warns.  Just make sure the order in the constructor initialization list matches the order in the class and this will be good to go.

I'm building locally to find the problem with the integration tests.

mjwitte:
@Myoldmopar Thanks.  I think I've addressed all of those.  I'll wait to push until I hear if you've found anything on the watercoil test failures.

Myoldmopar:
As you expected, `DesiccantDehumidifierWithAirToAirCoil.idf` encounters a divide by zero on mass flow rate:

```
Program received signal SIGFPE, Arithmetic exception.
0x00007ffff4c2ed82 in EnergyPlus::DXCoils::CalcTotCapSHR (InletDryBulb=26.666699999999999, InletHumRat=0.011178906890772281, 
    InletEnthalpy=55307.703692498355, InletWetBulb=19.428230207360055, AirMassFlowRatio=1, AirMassFlow=0, TotCapNom=14286, CBF=0, 
    CCapFTemp=3, CCapFFlow=9, TotCap=@0x7fffffffa368: 4.635571359542523e-310, SHR=@0x7fffffffa370: 6.9533558066990486e-310, 
    CondInletTemp=35, Pressure=101217, TotCapModFac=@0x55555652d280: 0.99986245570042243)
    at /src/EnergyPlus/DXCoils.cc:10211
10211				hDelta = TotCapCalc / AirMassFlow;
```

Myoldmopar:
Tested two more of the failures, same divide by zero on that line.

Myoldmopar:
@mjwitte OK, with commit 41ab0aa, I believe I have all the unit test segfaults fixed and the integration failures addressed. But you'll want to look close at that commit to make sure my changes make sense.  
- In the cc file, there was a divide by zero that I mentioned earlier, and I tried to protect against that by leaving that function early.  However, I'm not sure if my TotCap and SHR set to zero fix is complete enough, or even correct.  So please verify that.  
- Second, the code was segfaulting because the `coilSelectionReportObj` variable was being used uninitialized.  I inspected the code and found it really just needed a call to the `createCoilSelectionReportObj()` function.  Adding that to each of the unit tests before each call to `SizeWaterCoil` made the unit tests pass successfully, so that seems OK.

Maybe that's all that's needed here.  CI will tell us.  And you can tell me if my changes make sense to you as well.

Myoldmopar:
OK, this is looking way better.  @mjwitte please confirm my changes as per the [comment above](https://github.com/NREL/EnergyPlus/pull/6454#issuecomment-366488750), and make any changes needed on there.  Once you then push up your changes to clean up the build warnings, this should be ready to go in.

Myoldmopar:
Thanks for pushing that @EnergyArchmage.  I was hesitant to do it because @mjwitte [mentioned](https://github.com/NREL/EnergyPlus/pull/6454#issuecomment-366382027) he had addressed them locally but just hadn't pushed yet, so I didn't want to stomp on his work.  But he ran out of time didn't he! :laughing:

tst/EnergyPlus/unit/WaterCoils.unit.cc
```diff
@@ -216,6 +217,7 @@ TEST_F( WaterCoilsTest, WaterCoolingCoilSizing )
 	DataWaterLoopNum = 1;
 	NumOfGlycols = 1;
 
+	createCoilSelectionReportObj();
```
mjwitte, line 220: @Myoldmopar Thanks.  I had added this to energyplusfixture but I forgot that these tests use their own fixture.

src/EnergyPlus/DXCoils.cc
```diff
@@ -10049,6 +10198,12 @@ Label50: ;
 		InletWetBulbCalc = InletWetBulb;
 		InletHumRatCalc = InletHumRat;
 
+		if ( AirMassFlow <= 0.00001 ) {
+			TotCap = 0.0;
+			SHR = 0.0;
+			return;
```
mjwitte, line 10204: @Myoldmopar Thanks. This looks ok. Wondering why we're getting here with zero airflow, but that can be a followup investigation.

mjwitte (commented):
Thanks @Myoldmopar  and @EnergyArchmage for the cleanups.  This still needs the IDD changes for the two new report names. Plus it needs the code to produce the smaller version of this report. Soon.

EnergyArchmage:
It could be that the cooling capacity changes are causing diffs in three files are because of some subtle bug fixes that got in here.  The calls to RequestSizing need to set the temporary working variable to AutoSize prior to the call -- but this isn't always being done properly.  If the value passed in is not autosize value then the central sizing methods don't execute as expected.  This set of code changes includes this subtle fix in WaterCoils.cc in the following locations. 

https://github.com/NREL/EnergyPlus/blob/CoilSizingReport/src/EnergyPlus/WaterCoils.cc#L1928
https://github.com/NREL/EnergyPlus/blob/CoilSizingReport/src/EnergyPlus/WaterCoils.cc#L1944
https://github.com/NREL/EnergyPlus/blob/CoilSizingReport/src/EnergyPlus/WaterCoils.cc#L2208
https://github.com/NREL/EnergyPlus/blob/CoilSizingReport/src/EnergyPlus/WaterCoils.cc#L2221

@rraustad might be able to confirm, but I would guess that the diffs in 3 regression files are because of real changes in cooling coil capacity that are caused by this correction.  (not otherwise document in a CR I think)

mjwitte:
@EnergyArchmage We should pull those changes out of here and make that a separate PR. I can do that along with the other changes I need to do.

mjwitte:
Currently this code adds the full detailed coil selection data as a "Coil Sizing Summary Report" subtable in "Report: HVAC Sizing Summary" (HVACSizingSummary). I think this is a good spot for an abbreviated coil sizing table with about a dozen columns.  Then I propose adding a new "CoilSizingDetails" report that produces the full coil selection table as a separate report. The detailed report would be included as part of "AllSummary" reports. 

@EnergyArchmage @rraustad @Myoldmopar Sound OK?

An example of  the current full report inside HVAC Sizing Summary is in the attached zip file.
[5ZoneAirCooledConvCoefTable.zip](https://github.com/NREL/EnergyPlus/files/1735288/5ZoneAirCooledConvCoefTable.zip)

EnergyArchmage:
@mjwitte yes that sounds good.  I originally intended to do as a separate report request but the extra effort seemed considerable compared to adding under HVACSizingSummary.

mjwitte:
@EnergyArchmage For standard format reports it's not too bad. `newPreDefReport` automates most of it. This branch now has the new CoilSizingDetails report (in full glory) and an abbreviated Coil Sizing Summary Report subtable in the HVAC Sizing Summary.  There's probably too many columns yet in the smaller version.

mjwitte:
Forgot to revert the water coil diff-causers. There now.

src/EnergyPlus/HVACDXSystem.cc
```diff
@@ -619,7 +623,11 @@ namespace HVACDXSystem {
 			// considered as as 100% DOAS DX cooling coil
 			if ( DXCoolingSystem( DXCoolSysNum ).ISHundredPercentDOASDXCoil ) {
 				// set the system DX Coil application type to the child DX coil
-				SetDXCoilTypeData( DXCoolingSystem( DXCoolSysNum ).CoolingCoilName );
+				if ( ! ( DXCoolingSystem( DXCoolSysNum ).CoolingCoilType_Num == Coil_CoolingAirToAirVariableSpeed ) ) {
```
rraustad, line 626: interesting syntax

src/EnergyPlus/WaterCoils.cc
```diff
@@ -1783,6 +1878,8 @@ namespace WaterCoils {
 				SizingString = WaterCoilNumericFields( CoilNum ).FieldNames( FieldNum ) + " [C]";
 				DataFlowUsedForSizing = DataAirFlowUsedForSizing; // used by air loop coils
 				TempSize = WaterCoil( CoilNum ).DesInletAirTemp;
+				TempSize = AutoSize;
```
rraustad, line 1881: This erases the user input for design inlet air temp. 2 files shows diff's in eio because of this.

5ZoneAirCooled_ZoneAirMassFlowBalance:

    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Air Temperature [C], 21.00753 
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, User-Specified Design Inlet Air Temperature [C], 21.60000 
    Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Inlet Water Temperature [C], 7.00000 
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, User-Specified Design Inlet Water Temperature [C], 7.00000

    Coil:Cooling:Water,
      Main Cooling Coil 1,     !- Name
      CoolingCoilAvailSched,   !- Availability Schedule Name
      autosize,                !- Design Water Flow Rate {m3/s}
      autosize,                !- Design Air Flow Rate {m3/s}
      7.0,                     !- Design Inlet Water Temperature {C}
      21.6,                    !- Design Inlet Air Temperature {C}

I suggest a followup issue to correct any last minute findings.

rraustad:
Wow! Lot's of new information now. There are some things we could probably improve, like identifying zone name for TU heating coils or finding the sensible capacity and UA for water cooling coils that don't currently have that available (Coil Sizing Details). This should merge and we can iron out the issues.

mjwitte:
@rraustad There's 3 files with big eso diffs - maybe you can track those down?

rraustad:
I think I just did. Sizing is ignoring the user inputs for the cooling water coil and resulting in a different size. All 3 have different cooling coil size. At least that is where I would start looking. 

How do you guys want to handle these last IDD change PR's. Split them up and work towards finalizing? Or just merge ones that are likely and follow up issue those that need some correcting after IDD freeze?

mjwitte:
@rraustad If you have a fix for this, please point me to it, and I'll include that along with a section in output rules and a merge up to current develop.  Trying to minimize pushes.

rraustad:
WaterCoils.cc line 1881. bPRINT = true here if coil type is water simple. So using TempSize = AutoSize will neglect the user input for design inlet water temp. Why was this line added? I assume to be able to report sizing at actual vs design conditions.

rraustad:
and line 1897. I see where you reverted these out but must have missed these. Why do this anyway?

EnergyArchmage:
I had added several of those TempSize = AutoSize statements to get more of the code to execute inside the sizing routine.  I don't recall the details.

mjwitte:
@Myoldmopar This is ready for merge (with followups noted in #6481) once CI reports back, if you agree.

mjwitte:
@Myoldmopar Note that we expect a few diffs, similar to a [few days ago](https://github.com/NREL/EnergyPlus/pull/6454#commitcomment-27645175).
619 tests had: AUD diffs.
3 tests had: EIO diffs.
3 tests had: ESO big diffs.
3 tests had: Table big diffs.
101 tests had: ERR diffs.
1 test had: Table small diffs.
1 test had: MTR big diffs.
Failures:

regression Test Summary

Passed: 616
Failed: 3

The err warnings are in a lot of files - sorry.

mjwitte:
The units warnings should be gone now.  Sorry to kick off another cycle.

Myoldmopar:
@mjwitte this is going in.  We have to get the JSON finalized, and this needs to go first.  If you still see a problem with these results, please add it to the follow-up.

Issue #5603, opened 2016-04-11
Documentation Issues - April-August 2016


mjwitte:
- [x] In the IDD, WeatherProperty:SkyTemperature, the Schedule Name field can accept either a day schedule or an annual schedule, but the object-list reference only points to day schedules.

```
  A3;  \field Schedule Name
       \required-field
       \note if name matches a SizingPeriod:DesignDay, put in a day schedule of this name
       \note if name is for a SizingPeriod:WeatherFileDays, SizingPeriod:WeatherFileConditionType or
       \note RunPeriod, put in a full year schedule that covers the appropriate days.
       \type object-list
       \object-list ScheduleAndDayScheduleNames
```

Suggest changing the above to two lines:

```
\object-list ScheduleNames
\object-list DayScheduleNames
```

Then delete the three occurences of \reference ScheduleAndDayScheduleNames

_Addressed in #5817_

mjwitte:
- [ ] Engineering Reference, section 6.8.15.2 Equivalent Layer Optical Model, in the nomenclature list following figure 6.62, the variables which should be greek letters are not.  These were wrong also in the md version of the docs.  Refer back to v8.2 docs for the correct letters, following Figure 114 in that version.
- ***addressed in #5603***

JasonGlazer:
- [x]  Add to Energy+.idd: AvailabilityManager:OptimumStart and AvailabilityManager:HybridVentilation as \key choices in the  AvailabilityManagerAssignmentList object

_Addressed in #5817. But AvailaiblityManager:HybridVentilation is not valid in the availability manager list, it hangs out there alone.  So only added keys for AvailiabilityManager:OptimumStart"._

mjwitte:
- [ ] Dataset typo, see #5619 

- ***addressed in issue #5603 (as suggested).***

mjwitte:
- [ ] Helpdesk ticket 11322, questions about modeling a shade layer in a four pane window.  The v8.5 I/O Ref, on p. 347 under WindowProperty:ShadingControl Shading Type implies that shading only works for up to three panes and that between-glass shades can only be placed between the inner two panes.  Is this still true?
- [ ] The equivalent layer window model supports many layers in any order, but the I/O Ref and Engineering Ref are silent about whether this can be used with WindowProperty:ShadingControl (at least I didn't find anything about that) and how it would be applied.

- ***addressed in issue #5603. (added notes in I/O and Eng. Ref documents).***

[...]

PR #6463, opened 2018-02-05
Addressed Documentation Issues - April-August 2016 

Nigusse:
Issue #5603

Pull request overview
---------------------
Documentation update and Idd changes. Below are the list of changes:
- [x] Equivalent Layer Optical Model, equation symbols update
- [x] corrected typo in the ASHRAE_2005_HOF_Materials.idf data set
- [x] added a statement in I/O and Eng. Ref that WindowProperty:ShadingControl is not supported
- [x] corrected typo in five example files
- [x] Added People object Fraction Radiant default of 0.3 in the I/O ref document
- [x] added Ground exterior boundary condition saturated air assumption in HMAT model Eng Ref docs
- [x] Coil:Cooling:Water default heat exchanger configuration in I/O Ref changed to match the idd
- [x] corrected equations for Curve:ExponentialDecay and Curve:ExponentialSkewNormal in I/O Ref 
- [x] corrected equation for Sky emissivity in the Auxiliary Programs doc
- [x] verified the equation for EIRfplr for Chiller:Electric:EIR in Eng. Ref
- [x] corrected the AirTerminal:SingleDuct:Mixer image275.png and image276.png in I/O Ref
- [x] corrected  the GroundHeatExchanger:Surface image205.png and image206.png in IORef 
- [x] corrected the default "Control Algorithm" for AvailabilityManager:OptimumStart in IORef
- [x] corrected reference for "Glass Optical Properties Conversion" in WindowMaterial:Glazing in I/O Ref
- [x] added a missing note for Number of Stepped Control Steps field in Daylighting:Controls in I/O ref
- [x] corrected  VRF Heat Pump Heating Electric Power (and Consumption) output variable description

Checklist

### Review Checklist
 - IDD changes:


Nigusse:
@Myoldmopar I am not able to trace the reason for Ubuntu build failure. Can you please check and let me know? Thanks.

Nigusse:
@Myoldmopar  I cannot see the build failure problem locally. Can you please try to kick off Linux build and let me know which one is the offending equation?

Myoldmopar:
@Nigusse I found the build problem.  When running LaTeX locally, I found the `LaTeX Error:` string in the output.  

```
! LaTeX Error: Bad math environment delimiter.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.2404 ...ange from $\theta = 0$ to $\theta = 90\(
                                                  ^{o}\) (ASHRAE 1311-RP). T...

[420] [421] [422]
! Extra }, or forgotten $.
l.2515 ...at normal incidence, A\(_{o}\) = {$\tau}
                                                  \(_{bb}\) ($\theta = 0$), ...
```

Look at l.2404 in that snippet.  In the theta = 90 part, you open up math mode using `$`, but then you try to open up math mode again using \(.  Fail.  :+1: 

I pushed up a fix for one of them, but I actually just noticed the second one there.  Looks to be the same problem.  I'll push that one up also.

Myoldmopar:
Alright, if those fix the Ubuntu error then this can merge in.

Myoldmopar (approved):

Nigusse:
@Myoldmopar Thank you for fixing the building problem.

Myoldmopar:
Oh come on it's still there.  Ugh...rebuilding locally.

Myoldmopar:
@Nigusse OK, **now** they should all be cleaned.  A full build of all doc files touched on this branch reveals no more instances of "LaTeX Error" in the compiler output.

One thing that I think may have caused this is some confusion about braces.  In LaTeX, escaping parentheses is actually invoking math mode, so you can easily write inline math:

```
something like this \(a_1\) so that will render like nice math
```

But if you are already inside math mode, it will cause a failure, so:

```
\begin{equation}
ThisWont \(a_1\) work
\end{equation}
```

To be clear, I like to use single dollar signs for entering inline math mode:
```
like $a_1$
```

Now, all that said, please take a look at my last [commit](https://github.com/NREL/EnergyPlus/pull/6463/commits/e65dda715369d9f00a7fcc6885cacc9e6dd38f5f) for two reasons.  
1. Can you verify the lines I added are still showing the right form of those equations?
2. When you are making changes to the docs to fix errors, please take a little time to clean up the LaTeX code.  As you can see from the lines I changed, the original equations were awful, filled with nonsense that dates back to the Word->Markdown transition.  In almost all our documentation, the math can be cleaned up in a major way.  Just keep that in mind as you make more changes.

Thanks though, these look good now and this will make another good set of cleanup on the docs and input files.  If CI is happy on Linux, I'll merge it.

Nigusse:
@Myoldmopar Just check documentation in a local build. The equation for the curve Curve:ExponentialSkewNormal, in output reference on page 2270 is incorrect. It should have been Z2 = [(exp(C3*x) * C4*x) - C1] / C2, instead of C3, C4 was entered is used.

Myoldmopar:
Gah, OK.  I will make an issue to resolve after IO freeze so it doesn't add to CI burden now.  Thanks for finding it.

Myoldmopar:
OK, new issue made https://github.com/NREL/EnergyPlus/issues/6476; carry on.

Myoldmopar:
Closed via #6494

Issue #5903, opened 2016-10-12
EMS Actuator for Window Shading Control can cause crash if status set to "blind on" but there are no blinds


mjwitte:
## Issue overview

User file crashes becaseu EMS actuator for Window Shading Control is setting the control status to 6 "Interior blind on" but there are no blinds in the file.  The crash occurs [here](https://github.com/NREL/EnergyPlus/blob/198c6a3cffe4df717545e75a97f01510aa6a1875/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L3198), because it's trying to pass Blind(0).IRBackTrans.  Need some protection here, and possibly a few lines above for shade materials.  Or, add some checks where the actuator is read?  Not sure if this should warn (recurring warning), fatal, or just silently ignore invalid status.
### Details

Some additional details for this issue (if relevant):
- Win64
- Version of EnergyPlus 8.6.0
- helpdesk ticket 11599
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles, run with any weather)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
And it can also cause NAN errors if the shading control is set up as exterior blind and the actuator tries to deploy an interior blind.  Seems like the shading control needs to validate any EMS signals to be sure they are compatible with the specified shading material/type/construction.

New defect file added from helpdesk ticket 11605.

Myoldmopar:
Closed via #6466

PR #6466, opened 2018-02-09
EMS Actuator for Window Shading Control can cause crash if status set to "blind on" but there are no blinds 

Nigusse:
Addressed issue #5903

This is a defect. If EMS actuator for Window Shading Control requires a shade or blind layers to set the control status, then the code checks for shade or blind material layers in the construction. And if there are no such layers the EMS actuator will not be set.  This is done by setting a flag variable to true for window surfaces with shading control and has no shade and blind material layer.  The same variable will be used as a control logic before calculating Blind(0).IRBackTrans variable. Now the defect file fatal with "Invalid Actuated Component Type" and "Invalid Actuated Component Control Type" error message. Also added new defect file. There are no diffs except one audit diff warning error message. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
Audit diffs in EMSAirflowNetworkOpeningControlByHumidity.idf example file is expected because previously EMS actuator Control Status is always created if there is Window Shading Control object associated with a surface regardless of shade type.  Now the fix allows creating EMS actuator Control Status only when there is a shade or a blind layer in the window construction.

nrel-bot-3:

PR #6472, opened 2018-02-13
Clean up all warnings

### LINKED ISSUE #6400: Eigen Warnings
Issue overview
--------------
Building this morning on Ubuntu 17.10 (GCC 7.2.0), I noticed lots of warnings.  A bunch are directly E+ concerns, but then there are a bunch coming from Eigen.  I found this [bug](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1402) on the Eigen defect tracker where they noted the issue and addressed it.  It shows the fixes will be available in 3.3.5, but the latest release I found was 3.3.4 in June.  Maybe we can just add CMake flags to mute the warnings from this third party.  Either way, with the recently cleaned up build log from the LaTeX interaction mode change, it would be great to get this cleaned up as well.

@nealkruis I don't care to just assign this to you, but if you have any thoughts on updating Eigen, I'd like to know before I go in and start muting warnings.

### Details
 - Platform: Ubuntu 17.10

Myoldmopar:
Pull request overview
---------------------
On Ubuntu 17.10, with GCC 7.2.0, there were a bunch of build warnings.  Most were E+, though one was from Eigen.  These warnings have been showing up for a while, but I've ignored them.  Now that I am building lots of branches and reviewing, they reached the tipping point and I had to address them.  Here are the things I did:

 - Add `[[fallthrough]];` declarations to switch blocks where we actually intend on letting the code fall through without a `break;`. The C++ team understands how rare it is for such a condition so they added that declaration to mute the compiler about it.
 - Modify the Eigen code.  This technically fixes #6400 that I previously posted.  So Eigen already knows about [the problem](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=1402), and they have fixed it in an updated version of Eigen.  But I didn't know how long it would be before we get around to increment up to a newer version.  Luckily, the Eigen folks pointed to the changes ([here](https://bitbucket.org/eigen/eigen/commits/b4f969795d1b0adbb43ebdd8c6bbcb42cb559687?at=3.3) and [here](https://bitbucket.org/eigen/eigen/commits/131da2cbc6958c6576d23872946aac2da5a8467c?at=3.3)) that can be cherry picked into the 3.3 codebase to get rid of the warnings.  And that's what I did.
 - Modify the ParametricProcessor code to avoid the GFortran warning about a superfluous comma on WRITE() calls.  
 - Modify the EnergyPlusFixture code so that the C++ compiler would shut up about unclear/unguarded statements after an IF without curly braces.

Zero functional changes to the codebase. Now the build is all green and beautiful and I can get back to work.  This should/will merge if there are no diffs.


Myoldmopar:
As soon as one Linux bot gets done with this I'm merging it so it doesn't take CI time from other branches.

Issue #6471, opened 2018-02-13
Heating Coil energy transfer report variables use different names and should be unified 


rraustad:
Issue overview
--------------
Setting up reporting is difficult for heating coils since the heating coils currently use different naming conventions. Previous work to unify report variable names missed the heating coils.

Water heating coils:
Heating Coil Heating Rate
Heating Coil Heating Energy

DX heating coils (includes VRF):
Heating Coil Total Heating Rate
Heating Coil Total Heating Energy

non-DX/water heating coils (i.e., elect, multistage, other fuel, gas, desuperheater):
Heating Coil Air Heating Rate
Heating Coil Air Heating Energy

Not only does this make it difficult for users to remember which report variable to use it also adds complexity when different coil types are used in the same simulation.

In addition, setting up a monthly report table for heating and/or cooling coils would require the user to include all report variable names to make this table generic across different models.

    Output:Table:Monthly,
      Unitary Coil Report Variable Test,    !- Name
      ,                                     !- Digits After Decimal
      Heating Coil Heating Energy,          !- Variable or Meter Name 1
      SumOrAverage,                         !- Aggregation Type for Variable or Meter 1
      Heating Coil Heating Rate,            !- Variable or Meter Name 2
      Maximum,                              !- Aggregation Type for Variable or Meter 2
      Heating Coil Total Heating Energy,    !- Variable or Meter Name 3
      SumOrAverage,                         !- Aggregation Type for Variable or Meter 3
      Heating Coil Total Heating Rate,      !- Variable or Meter Name 4
      Maximum,                              !- Aggregation Type for Variable or Meter 4
      Heating Coil Air Heating Energy,      !- Variable or Meter Name 5
      SumOrAverage,                         !- Aggregation Type for Variable or Meter 5
      Heating Coil Air Heating Rate,        !- Variable or Meter Name 6
      Maximum,                              !- Aggregation Type for Variable or Meter 6
      Cooling Coil Total Cooling Energy,    !- Variable or Meter Name 4
      SumOrAverage,                         !- Aggregation Type for Variable or Meter 4
      Cooling Coil Sensible Cooling Energy, !- Variable or Meter Name 5
      SumOrAverage,                         !- Aggregation Type for Variable or Meter 5
      Cooling Coil Sensible Cooling Rate,   !- Variable or Meter Name 5
      Maximum,                              !- Aggregation Type for Variable or Meter 5
[...]

PR #6473, opened 2018-02-13
Name changes for heating coil energy/rate report variables

rraustad:
Pull request overview
---------------------
Heating coils have different naming convention for reporting heating coil energy/rate. This makes it difficult to remember which variable name applies to which coil type. This change addresses #6471 and unifies the naming convention for heating coils across all heating coil types.

Existing naming convention:

<strong>Water heating coils:</strong>
Heating Coil Heating Rate
Heating Coil Heating Energy

<strong>DX heating coils</strong> (includes VRF):
Heating Coil Total Heating Rate
Heating Coil Total Heating Energy

<strong>non-DX/water heating coils</strong> (i.e., elect, multistage, other fuel, gas, desuperheater):
Heating Coil Air Heating Rate
Heating Coil Air Heating Energy

The report variable names were changed to:
Heating Coil Heating Rate
Heating Coil Heating Energy

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
All heating coil reports now show up in the same place, columns 1-3 for all coil types (see description [here](https://github.com/NREL/EnergyPlus/issues/6471)).

heatingcoilreporting3

rraustad:
Before I change 112 example files maybe someone can say yeah or nay on this change. This specifically addresses GUI needs.

EnergyArchmage:
I have reviewed these changes and agree they are a helpful improvement.  I created a new label for output change (so it doesn't get ignored for I/O freeze).

Myoldmopar:
It seemed a bit fuzzy whether this needed to drop before IO freeze or not.  There's no structural changes, but since the output variables names are changing, it probably does make sense.  The output change label is a good idea, we'll just need to be specific about what exactly it means.  Hmm...can you add a description to a label?

Myoldmopar (approved):

EnergyArchmage:
Regarding the label, I don't find anyplace to add a description to them.

Myoldmopar:
Yeah, I didn't either, kind of a surprise.  If both the IDDChange and Output Change labels are intended to mark a PR based on when it needs to be merged, maybe we can just combine them into an IOFreeze label or something so we don't have to keep track of both.

mjwitte:
@Myoldmopar Happened to notice some related I/O ref mistakes that were wrong before this PR. I can push them up here or send to @rraustad to go in with the example file changes (just to save a CI cycle).

Myoldmopar:
I'm indifferent about where the IO ref changes go as long as they don't trigger another CI build.

rraustad:
While transitioning example files, noticed that there were several files with both flavors of report request (i.e., Heating Coil Air Heating vs Heating Coil Total Heating) (~20ish). Only kept one. I also removed duplicate reports (existing) where appropriate. I am thinking this will cause big diff's, regardless if there was only one flavor in example file or both, because now all heating coils will report to the csv while before only specific heating coil types would report.

rraustad:
Looks like I missed one:

SingleFamilyHouse_TwoSpeed_CutoutTemperature

    +   ** Warning ** The following Report Variables were requested but not generated -- check.rdd file
    +   **   ~~~   ** Either the IDF did not contain these elements, the variable name is misspelled,
    +   **   ~~~   ** or the requested variable is an advanced output which requires Output : Diagnostics, DisplayAdvancedReportVariables;
    +   ************* Key=*, VarName=HEATING COIL TOTAL HEATING ENERGY, Frequency=TimeStep
    +   ************* Key=*, VarName=HEATING COIL AIR HEATING ENERGY, Frequency=TimeStep

I could fix that on a different commit?

Myoldmopar:
I say fix it in another commit and push, but then we'll merge before CI wastes time on it.

Myoldmopar:
(And by fix it in another commit, I mean push a commit to this branch, but we won't wait for a full CI pass just for the variable name fixup.)

Myoldmopar:
(And by all that above, I mean I'll just do it real quick and merge this, thanks @rraustad)

rraustad:
Thanks, I am in the middle of 6372 - Std 62.1 refactor.

Issue #5381, opened 2015-12-17
Zone Thermal Comfort ASHRAE 55 Adaptive Model Running Average Outdoor Air Temperature Error at beginning of year


mjwitte:
Helpdesk ticket 10614
From user (have not verified, user files in dev support defect files folder):
The People objects provide for the calculation of the ASHRAE 55 Adaptive comfort temp for use in output, and which can also be used as a window control option if using AFN. The Standard provides for calculation of a prevailing mean outdoor air temp, called by E+ the "Zone Thermal Comfort ASHRAE 55 Adaptive Model Running Average Outdoor Air Temperature" (which I'll call Tpma following the standard). This is then used to calculate the center of the comfort band, called by E+ "Zone Thermal Comfort ASHRAE 55 Adaptive Model Temperature."

The issue is that the Tpma calculated by E+ seems to have some error in the beginning of the year. I would imagine this is some type of a startup issue. It's pretty severe the first month, and takes about 3 months to be fully forgotten by the moving average. In the attached, I just ran the idf for two run periods, and you'll see the calculated Tpma is different for the two years. Since it's just based on outdoor air temp, it should be the same. In this case, the Tpma error not only changes the indoor comfort temp but actually leads to it giving an error the first year (since Tpma with the error appeared to be out of the bounds allowed by the Standard), but not the second year.

I ran this with the San Fran weather file for demo. (Tpme in the file name because of the Portuguese translation -- same thing as Tpma).

lgu1234:
I test file. "Zone Thermal Comfort ASHRAE 55 Adaptive Model Running Average Outdoor Air Temperature" is not calculated correctly. Here is my observation. 

There are several bugs in initialization in the CalcThermalComfortAdaptiveASH55 function of the ThermalComfort module.
1. Skip 9 lines when reading a weather file
Each epw file has 8 lines of headers. The code reads 9 lines. This needs to be corrected.
2. Use a wrong starting day
The Tpme is supposed to be an average value in the previous 30 days. When Jan. 1 is used in the first day for simulation, the starting day for average is Dec. 2. The code uses Dec. 1.
3. Wrong calculation of Tpma 
Here is a line of calculation:
	runningAverageASH = ( 29.0 * runningAverageASH + avgDryBulbASH ) / 30.0;
When the initial value of runningAverageASH is zero, the above calculation will give wrong output. The correct calculation is  
	runningAverageASH = runningAverageASH + avgDryBulbASH  / 30.0;

[...]

PR #6474, opened 2018-02-13
Fix the running average outdoor air temperature in the first day 

lgu1234:
Pull request overview
---------------------
This fix is to provide correct running average outdoor air temperature of the first day as Phase 1, addresses #5381  Here is what has been fixed in the CalcThermalComfortAdaptiveASH55 function of the ThermalComfort module.
1. Skip 9 lines when reading a weather file
 Each epw file has 8 lines of headers. The code reads 9 lines.
2. Use a wrong starting day
 The Tpme is supposed to be an average value in the previous 30 days. When Jan. 1 is used in the first day for simulation, the starting day for average is Dec. 2. The code uses Dec. 1.
3. Wrong calculation of Tpma
 Here is a line of calculation:
 runningAverageASH = ( 29.0 * runningAverageASH + avgDryBulbASH ) / 30.0;
 When the initial value of runningAverageASH is zero, the above calculation will give wrong output. The correct calculation is
 runningAverageASH = runningAverageASH + avgDryBulbASH / 30.0;

I test the test file using San Francisco weather file, the first day average OA temperature is the same as the calculated value from weather file.

Phase 2 fixed calculation in the rest of time.

It should be pointed out that when reading a weather file, hourly weather data are used to calculate daily average temperature. When starting a simulation, outdoor dry bulb temperature is obtained in the following lines:

		if ( BeginHourFlag ) {
			avgDryBulbASH += ( OutDryBulbTemp / 24.0 );
		}
 When BeginHourFlag is true, the current time is current hour + a time step, so that OutDryBulbTemp is not exactly same as one in the hourly weather file due to interpolation. Therefore, there are small differences in the first 30 days from program and weather file direct calculation.

Enclosed is a figure to show Running-Average-Outdoor-Air-Temperature in both years from a defect file runs. 

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - n/a IDD changes:


src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2515,13 +2517,13 @@ namespace ThermalComfort {
 							dryBulb = StrToReal( epwLine.substr( 0, pos ) );
 							avgDryBulbASH += ( dryBulb / 24.0 );
 						}
-						runningAverageASH = ( 29.0 * runningAverageASH + avgDryBulbASH ) / 30.0;
+						DailyAveOutTemp( i ) = avgDryBulbASH;
 					}
 				} else { // Do special things for wrapping the epw
 					calcEndDay = jStartDay;
 					calcStartDay += 365;
```
mjwitte: Does this need to know if the epw file has 366 days for a leap year?

mjwitte (changes requested):
@lgu1234 I ran the defect file with SF weather with develop ( 1c5ba897d1 ) and with this branch ( e53dcc785a ). I got the same results with both for the reported variables, and "Zone Thermal Comfort ASHRAE 55 Adaptive Model Temperature" is still -1C for all of January, in both years.

lgu1234:
@mjwitte It is true that "Zone Thermal Comfort ASHRAE 55 Adaptive Model Temperature" is still -1C for all of January, in both years". Here is code (line 2616):

			if ( runningAverageASH >= 10.0 && runningAverageASH <= 33.5 ) {
If the above conditions are true, the temperature is calculated. Otherwise, the temperatures are -1C. Since most average outdoor temperatures are below 10C in Jan, this is why the most Adaptive Model Temperature is -1C.

I am going to check-in to allow leap year.

mjwitte:
@lgu1234 I just realized that the reason I saw no change with this fix is because my runs were pointing to an epw file that had a .stat file present, so none of this code was active.  When I run your branch now with a lone epw (no stat file), it crashes with a debug build. I'm going to see if I can locate where.

mjwitte:
Hmm, that's odd, EnergyPlus isn't crashing, something else is failing.

lgu1234:
@mjwitte I am running debug version of EnergyPlus with current code. It is very slow. So far, it ran to 5/21 without crash.

mjwitte (approved):

mjwitte:
@lgu1234 Retested with latest code and *without* a stat file present. Now this all looks good.  Will merge once CI completes.

lgu1234:
@mjwitte Thanks.

mjwitte:
linux CI failed again, merging anyway.

PR #6489, opened 2018-02-21
add trap to not process population concurrence unless VRP present

EnergyArchmage:
follow up for performance issue for standard 62.1 changes from #6372 

Pull request overview
---------------------

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/SizingManager.cc
```diff
@@ -1431,6 +1431,17 @@ namespace SizingManager {
 
 	void
 	DetermineSystemPopulationDiversity(){
+
+		//first determine if any airloops use VRP, if not then quick return
+		bool anyVRPinModel( false );
+		for ( int AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum ) {
+			if ( FinalSysSizing( AirLoopNum ).SystemOAMethod == SOAM_VRP ) {
+				anyVRPinModel = true;
```
rraustad: could break here

Myoldmopar:
@EnergyArchmage did you expect to see diffs like this produced?  I checked the absdiff table results and it shows there is a difference of 1.0 on the population diversity value.  Please confirm.  Also, I agree with @rraustad that since this is a performance improvement, you might as well break early on that loop when you calculate `anyVRPinModel = true`.

EnergyArchmage:
Well there is tension between filling out std 62.1 tables as much as possible when VRP is not used and doing the calculations properly when VRP is used.  But I can rework this to still provide people sums and a diversity of 1.0 without marching thru the schedules.  I will push up changes shortly.

Myoldmopar:
@EnergyArchmage did you expect that last batch of changes to clean up the diffs "completely"?  I'm not saying you need to, I was just curious about your intent with that last set of commits after my diff comment.  Since there are still diffs.

mjwitte:
We expect diffs, right? Because we're not calculating diversity anymore unless VRP is active.

EnergyArchmage:
Yes, the Ps value and D value are expected to change because the D is not being accurately calculated.  These values are not really used for anything if not using VRP but they still have table entries.

PR #6491, opened 2018-02-21
Eliminate fallthrough warnings on new GCC

Myoldmopar:
Pull request overview
---------------------
Trying another approach to get rid of the switch block fallthrough warnings on new GCC.  This should be more robust on different versions and platforms because I protect it in a #define for GCC only.

### DIFF
--- a/src/EnergyPlus/EnergyPlus.hh
+++ b/src/EnergyPlus/EnergyPlus.hh
@@ -59,6 +59,13 @@
 // UNUSED( foo );
 #define EP_UNUSED( expr )
 
+// macro to suppress fallthrough warnings on switch statements
+#ifdef __GNUC__
+  #define ATTR_FALLTHROUGH __attribute__((fallthrough))
+#else
+  #define ATTR_FALLTHROUGH
+#endif
+
 // ObjexxFCL
 namespace ObjexxFCL {
 namespace fmt {
--- a/src/EnergyPlus/EnergyPlus.hh
+++ b/src/EnergyPlus/EnergyPlus.hh
@@ -59,13 +59,6 @@
 // UNUSED( foo );
 #define EP_UNUSED( expr )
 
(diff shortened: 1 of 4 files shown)


mbadams5 (changes requested):
Apparently`__attribute__((fallthrough))` is a GCC7 only attribute. However, you can apparently just add a comment in its own line that follows certain regexes and it'll ignore the fallthrough warning for you as well. See the following links: [link 1](https://stackoverflow.com/questions/45349079/how-to-use-attribute-fallthrough-correctly-in-gcc), [link 2](https://stackoverflow.com/questions/45129741/gcc-7-wimplicit-fallthrough-warnings-and-portable-way-to-clear-them), [link 3](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/), and [link 4](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html).

Myoldmopar:
I had tried some comments but the warnings persisted. I'll retry. Thanks
for the links!

-- 
Edwin Lee, PhD
Engineer, Commercial Buildings Research Staff
National Renewable Energy Laboratory

mbadams5:
Tested locally on GCC 7.3.0 that this addresses the build warnings.

mbadams5 (approved):

Myoldmopar:
So @mbadams5 I'm not sure if it's already documented in the JSON follow-up issue, but this PR revealed an oddity where the input file (found) line for external dependent files, in the audit file, is being reported with the full path.  This is causing diffs in every test file that uses an external file.  I'll have to check what it was doing before there to know a solution.  If it turns out we do want it to be like that, then we just need to teach CI to ignore those lines (like it already does for runtime reporting strings).

Myoldmopar:
This branch does make my build look nice and green.  Thanks for getting the comment-only approach working @mbadams5. Merging.

PR #6493, opened 2018-02-22
An Intermediate Fix for Constant Speed Branch Pumps and Refactor

Myoldmopar:
Pull request overview
---------------------
This pull request started as a fix to let constant speed branch pumps turn off when they aren't needed.  This was partially addressed by allowing a resimulation of the loop side with some branches disabled if they are turned off in the initial solution.  This allowed the loop flow request to drop, but the solution isn't complete, as flow distribution still puts some flow on the off branch.

During this development, many files were refactored and lots of unit tests were added to improve code coverage.  This led to gobs of file diffs, but only 1 test file had diffs in results.  I'm going to continue working on this issue in a followup but want to get this merged in now so the large mass of file changes doesn't interfere with the actual functional changes I'll provide in the follow-up.

Resimulation Logic
---------------------
The basic idea I am floating is to add a flag to the branch data structure to allow an override of the flow request evaluation for that branch, only for special constant-speed-branch-pump situations.  This flag will initially be set to false, allowing the pump to be turned on as normal, and a full flow/load dispatch solution is obtained.  Once obtained, however, if the loop topology indicates a constant speed branch pump situation, the load on each branch is checked.  If there is any parallel branch with no load, the branch is temporarily disabled and the flow/load simulation is repeated.  This time around, the flow request will not include the disabled branch, and so unless there is a bad flow restriction issue, the disabled branch shouldn't get any flow.  There definitely shouldn't be any load accidentally dispatched to the disabled branch anyway.

Lots of testing and design work to do, but at least I have something to work with.  Things to do:
 - [ ] 3522 defect file did not fully load up both chillers, so I need to verify this logic works with that
 - [ ] Only need to disable/rerun if there is both no load and some flow
 - [ ] Verify with and without common pipe, though I don't think it will matter
 - [ ] Try to capture as much functionality as possible into standalone worker functions so we can unit test the heck out of this
 - [ ] Unit testing the SetupFlowRequest function might be a lot of code/time, but really shouldn't be too hard: just read in the plant topology for each case we want to test, then update MassFlowRateRequest variables around and re-call the function.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
Well, I checked out the 1 file with ESO diffs (_BranchPumpsWithCommonPipe) and found that the supply side of the loop is now running at a lower flow rate but still meeting setpoint perfectly.

Myoldmopar:
Uhhhhh this sorta took off a bit more than I anticipated, but it's organization that's long overdue.  I started objectifying the loop solver and quickly hit a circular dependency in header files.  Instead of doing a quick hack, I broke the structs (most of them) out of DataPlant and into their own files.  It will help a lot of things.

I pulled develop into this branch, so even with all these changes, there shouldn't be any major diffs, except the one branch pump file that this actually fixes.

src/EnergyPlus/Plant/Operation/CallingOrder.hh
```diff
@@ -0,0 +1,72 @@
+// EnergyPlus, Copyright (c) 1996-2018, The Board of Trustees of the University of Illinois,
```
Myoldmopar: @mjwitte @EnergyArchmage One of the big organization enhancements I did on this PR was pull classes and structs out of DataPlant and into their own appropriate files.  Every time I tried to do some organization, I ended up with circular dependencies because the classes were so intermingled.  And also, every time I made a tiny change to one struct in DataPlant, it would trigger a near full rebuild of E+ because so many files depend on it.  

Now (almost) every struct is placed in its own file with license text header, and if a small change is needed, it won't automatically cause a full rebuild.  I pulled all the functions from DataPlant that shouldn't have been there in the first place, and dropped them into PlantUtilities.  I deleted at least one function that wasn't used, and I deleted a bunch of arrays and random variables in DataPlant that weren't used.

I know the changes here have become difficult to diff through, but I feel very confident that as long as the code compiles (and we aren't seeing unexpected diffs), these organizational changes were purely beneficial.  I don't intend on pulling all the components into directories, at least not here.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -93,18 +93,14 @@ namespace PlantLoopSolver {
...
-	// DERIVED TYPE DEFINITIONS
-
 	// MODULE VARIABLE DEFINITIONS
 	Real64 InitialDemandToLoopSetPoint;
 	Real64 CurrentAlterationsToDemand;
 	Real64 UpdatedDemandToLoopSetPoint;
 	Real64 LoadToLoopSetPointThatWasntMet; // Unmet Demand
 	Real64 InitialDemandToLoopSetPointSAVED;
-	int RefrigIndex( 0 ); // Index denoting refrigerant used (possibly steam)
+    bool CurrentLoopSideIsConstantSpeedBranchPumped;
```
Myoldmopar: Indentation went awry all over the place here.  I'll fix that.

This variable, that I will probably place in a better location, is a flag to indicate that the current loopside is constant speed branch pumped.  This flag signals to the solver that the entire load/flow solution may need to be repeated a second time for these loops.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -168,215 +158,248 @@ namespace PlantLoopSolver {
...
 		} else {
 
 			// Set the flag to true so that it is activated for the next time step
-			PlantLoop( LoopNum ).LoopSide( ThisSide ).OncePerTimeStepOperations = true;
+            thisLoopSide.OncePerTimeStepOperations = true;
 
 		}
 
 		// Do pressure system initialize if this is the demand side (therefore once per whole loop)
-		if ( ThisSide == DemandSide ) SimPressureDropSystem( LoopNum, FirstHVACIteration, PressureCall_Init );
```
Myoldmopar: These changes are hard to dig through, but it's just organizational.  Logic was pulled into tighter functions to make it easy to repeat the call to the solution.  I'll highlight the actual functional change below.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -168,215 +158,248 @@ namespace PlantLoopSolver {
...
 	}
 
-	//==================================================================!
-	//==================================================================!
-	//==================================================================!
-
-	//==================================================================!
-	//================= TOPOLOGY VALIDATION ROUTINE ====================!
-	//==================================================================!
+	void
```
Myoldmopar: This new helper function re-enables all branches on the loop in case any were disabled as a part of this branch pumping fix.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -168,215 +158,248 @@ namespace PlantLoopSolver {
...
+	{
+		for ( int branchNum = 2; branchNum <= loop_side.TotalBranches-1; ++branchNum ) {
+			auto & branch = loop_side.Branch( branchNum );
+			branch.QuickDisableForCSBranchPumping = false;
+		}
+	}
 
-	m_FlowControlValidator
-	ValidateFlowControlPaths(
+    void
```
Myoldmopar: This function is to be called after a full load/flow solution has been achieved on the loopside, but there are constant speed branch pumps present that could be causing branches to turn on when they aren't needed.  It loops over the branches and components, checks load, and if there isn't any load on the loop, it disables the branch to allow a second pass.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -168,215 +158,248 @@ namespace PlantLoopSolver {
...
-		//  sets the corrected flow on the inlet to each parallel branch
-		ResolveParallelFlows( LoopNum, ThisSide, ThisLoopSideFlow, FirstHVACIteration );
-		//  CALL PropagateResolvedFlow(LoopNum, ThisSide)
-
-		// Re-Initialize variables for this next pass
-		InitialDemandToLoopSetPointSAVED = InitialDemandToLoopSetPoint;
-		CurrentAlterationsToDemand = 0.0;
-		UpdatedDemandToLoopSetPoint = InitialDemandToLoopSetPoint;
+		// Do the actual simulation here every time
+		thisPlantLoop.loopSolver.DoFlowAndLoadSolutionPass(LoopNum, ThisSide, OtherSide, ThisSideInletNode, FirstHVACIteration);
```
Myoldmopar: The entire flow unlocked simulate, flow resolution, flow locked simulate process is captured inside this function now.  For all files which don't contain constant speed branch pumps, there is no change to the algorithm, and this is called once and everything is fine.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -168,215 +158,248 @@ namespace PlantLoopSolver {
...
 
-		// Now that flow rates have been resolved, we just need to set the flow lock status
-		//  flag, and resimulate.  During this simulation each component will still use the
-		//  SetFlowRequest routine, but this routine will also set the outlet flow rate
-		//  equal to the inlet flow rate, accoridng to flowlock logic.
-		PlantLoop( LoopNum ).LoopSide( ThisSide ).FlowLock = FlowLocked;
-		SimulateAllLoopSideBranches( LoopNum, ThisSide, ThisLoopSideFlow, FirstHVACIteration, LoopShutDownFlag );
+		// On constant speed branch pump loop sides we need to resimulate
+        if ( CurrentLoopSideIsConstantSpeedBranchPumped ) {
```
Myoldmopar: However, if there are constant speed branch pumps, we now check to see if there were any branches that were on but shouldn't have been.  If so, we need to disable them and re-run the entire solution again to allow the flow to be dispatched properly without those extraneous branches available.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -726,6 +719,11 @@ namespace PlantLoopSolver {
 				auto & branch( loop_side.Branch( BranchCounter ) );
 
 				if ( BranchCounter > 1 && BranchCounter < NumBranchesOnThisLoopSide ) ++ParallelBranchIndex;
+
+				if ( branch.QuickDisableForCSBranchPumping ) {
```
Myoldmopar: Here we check to see if the branch was disabled for this fix, and if so, we don't allocate any flow request for that branch.  It has the effect of turning off the branch.

src/EnergyPlus/Plant/Solution/PlantLoopSolver.cc
```diff
@@ -1005,18 +1003,14 @@ namespace PlantLoopSolver {
...
+        return LoopFlow;
 
-	//==================================================================!
-	//================== LOOPSIDE BRANCH SIMULATION ====================!
-	//==================================================================!
+	}
 
 	void
-	SimulateAllLoopSideBranches(
+	PlantLoopSolverClass::SimulateAllLoopSideBranches(
```
Myoldmopar: Oh, and I pulled all of these functions into methods on a struct.  I didn't really touch the guts of each method, so this is an incremental improvement.  Eventually, we could have each plant loop properly store its own solver instance, and reduce the number of arguments being passed around to each function and the amount of global data required to test.

src/EnergyPlus/Plant/Solution/PlantManager.cc
```diff
@@ -1518,47 +1482,6 @@ namespace PlantManager {
 					InletNodeNumbers.deallocate();
 				}
 
-				if ( NumOfPipesInLoop > 0 ) {
```
Myoldmopar: This was fully dead code as far as my investigation could tell.

tst/EnergyPlus/unit/PlantUtilities.unit.cc
```diff
@@ -84,3 +87,342 @@ TEST_F( EnergyPlusFixture, PlantUtilities_RegisterPlantCompDesignFlowTest1 )
 	EXPECT_EQ( TestFlowRate3, CompDesWaterFlow( 1 ).DesVolFlowRate );
 
 }
+
```
Myoldmopar, line 90: I added a lot of tests here to the PlantUtilities functions.  I didn't exhaust the entire file to 100% coverage, but it is much closer now.

Myoldmopar (commented):
@mjwitte @EnergyArchmage Could you provide some initial feedback on this fix?  I know it's a lot of changes, I got real excited as I started refactoring.  It's almost entirely organizational changes.  I highlighted the functional changes in the comments attached to this "review".  

In every test I've run so far, I've seen no diffs except in the 2 constant speed branch pump files I've tested with, and those are now showing better, expected, flow distribution.  It should also be a breakeven or net gain in performance because in files without const-speed-branch-pumps, there is a single IF check added and that's it.

I'll follow this "review" up with some plots or results summaries showing the improvement.

src/EnergyPlus/Plant/Operation/CallingOrder.hh
EnergyArchmage: I can't seem to find the new directories and little files in VS2015 solution explorer.  I reran CMAKE and successfully built the branch, but I can't find the new files in subfolders in the IDE.

mjwitte: @EnergyArchmage For whatever reason, all of the source files display as a flat list in VS2015. And most of the new files are headers. I did a build and then let VS reload all and I see the new files, `Branch.hh` is in the EnergyPlusLib\Header Files folder. This will be a navigation problem soon (if not already). I'm looking for an option setting to display the list using the subfolders but I haven't found it. Maybe it's controlled by cmake?

Myoldmopar: Thanks for starting to look at this guys.  On the Visual Studio stuff, I honestly don't have a clue.

A quick Google reveals this link shows something about changing the project files view, for a Java project, but I don't expect it to be helpful: https://msdn.microsoft.com/en-us/library/aa265177(v=vs.60).aspx

A little more searching reveals Visual Studio doesn't want to do anything with the hierarchy.  Though there was also this https://stackoverflow.com/a/28702616 which describes a CMake "source_group".  Might be worth trying on someone's machine locally to see if they show up in folders.  @kbenne any thoughts on this front?

Myoldmopar:
OK, someone will have to stop me I guess.  I keep looking over the test coverage, getting discouraged finding files that are 100% uncovered by unit tests, and then writing unit tests over them.  This PR has gone way out of scope, but the unit test coverage should be nice and shiny by the time I'm done.

src/EnergyPlus/Plant/Operation/CallingOrder.hh
mjwitte: Discovered that what looks like folders in the VS solution explorer are *filters*. \Build\src\EnergyPlus\energypluslib.vcxproj.filters is an xml file that looks like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <ClCompile Include="D:\MJW-Work2\EnergyPlus Git2\EnergyPlus-ReviewBranch4\src\EnergyPlus\AirflowNetworkBalanceManager.cc">
      <Filter>Source Files</Filter>
    </ClCompile>
    <ClCompile Include="D:\MJW-Work2\EnergyPlus Git2\EnergyPlus-ReviewBranch4\src\EnergyPlus\AirflowNetworkSolver.cc">
      <Filter>Source Files</Filter>
```
Haven't figured out what builds this file.

EnergyArchmage: I was looking for enums.cc, branch.cc, etc but now I realize the file breakout was just header files.

Myoldmopar: Ahh makes sense. Sorry for the confusion. Yeah it was pretty much all header stuff. If it ever broke into more implementation files they'd go in there too, but for now it's just headers.

Myoldmopar:
Note to self: I need to go back and add any necessary `clear_state`s.

src/EnergyPlus/Plant/Operation/CallingOrder.hh
EnergyArchmage: Without the plant subfolder in the IDE, it seems like the new header filess should be called PlantEnums.hh, PlantBranch.hh etc.    Not going to have all enums in there.

jasondegraw: @Myoldmopar @mjwitte Sorry to be a wet blanket, but I'm just really not a fan of this type of organization. I get that we don't want lots of files in one directory like certain other projects (cough cough), so one layer of organizing directory structure makes sense. But this to me is going too far, and is going to be an IDE nightmare.

Myoldmopar: What other projects?

image

One layer of organizing seems good.  I can pull them all into the single `Plant/` subdirectory and name the files with more descriptive names.  Those two fixes should satisfy all of you wet blankets and Visual Studio users :upside_down_face:.

mjwitte:
@Myoldmopar @EnergyArchmage Getting  warmer on the project organization in Visual Studio (and maybe other IDEs).  In VS, the energyplus_tests project *is* organized by folders. Looks like this [cmake macro](https://github.com/NREL/EnergyPlus/blame/8df72926681d9827a5ea69ab4303f30df39b1f43/cmake/ProjectMacros.cmake#L19-L30) is doing the work. Blame points to @kbenne. Maybe we can add a similar macro for energypluslib?  Maybe even add a cmake option to organize by Headers/Source or by folder or by both?

Myoldmopar:
That's a good find @mjwitte.  Although that doesn't show up in current OpenStudio code, it looks like it [was there](https://github.com/lefticus/OpenStudio-Core/blob/master/ProjectMacros.cmake#L23) at one point.  I also found a different person with their [own version](https://github.com/rttrorg/rttr/blob/621cd1df0c9b2d6133aabcf5d3cc129f17a87fac/CMake/utility.cmake#L37) of it.  It looks like you just call the create source groups once you have the files listed. 

Could you try this locally?  I think it is just a matter of calling `CREATE_SRC_GROUPS( "${SRC}" )` right about [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/CMakeLists.txt#L597).

Myoldmopar:
For the record I tried it on my Linux box, and CMake ran fine, and it is building fine.  I didn't expect any change in behavior for me, but I wanted to confirm it.  Hopefully it will fix it for Visual Studio users.

Myoldmopar:
I pulled up a VM and added that line and reran the ZERO_CHECK project.  The solution reloaded and looked like this:

image

Which is basically what we want.  I don't know exactly why the `C:` folder got in there.  The only item in there is buried deep inside the milo third_party folder.  Everything else seems OK at first glance.  

Please try it out locally and browse around more than I did.  If it looks good, I can just keep it in my branch and merge it in ... soon?

mitchute:
I'm not seeing any of the folder structure here:

image

mjwitte:
@mitchute @Myoldmopar I added this (and added .hh to a couple of the file names) and it's working for me (after I figured out I had to turn off "Show All Files" in the solution explorer). I've pushed this up.

mitchute:
Working over here as well (after I deleted the build folder)

image

The D: directory tree is just where the src and build folders live on my machine.

Myoldmopar:
Welp, I ended up finding a bug in the piping system while trying to get unit tests to cover that file.  If there was a non-piping domain in the file mixed with piping domains, it would hard crash.  The SimulateGroundDomains function would try to do initialization on all the domains it found in the file.  Unfortunately it wouldn't pass in a circuit index, and inside the initialization loop, if it found a pipe cell, it would still try to access the circuit, which would lead to an access violation.  My fix was just to ignore circuit-based domains during that initialization loop.  Circuit-based domains should still be initialized from the plant-side entry point of that module, so I'm expecting no diffs there.  Remaining steps here:

 - [x] Go back through these unit tests and actually add expectations/assertions
 - [x] Go back through modified files and make sure `clear_state`s are in there as needed
 - [x] Verify fix one more time and provide plots
 - [ ] Else?

Myoldmopar:
OK, so taking some plots from the 3522 test file that I attributed to @kbenne, these show the plant behaving better:

This figure shows the load being demanded onto the supply side.  
image
This branch and develop both line up, which makes sense, as the file does still meet demand in current develop.  The problem is that it does it by turning on branches unnecessarily.  There is a very small difference between the two profiles that can't be seen on the plot.  If the profiles were significantly different between the two branches, I would be concerned.  

This next figure shows the supply side outlet temperature.
image
Again, this shows the current approach allows the plant to meet the demand and remain controlled.

This final figure shows the real gain here, the improvement in supply side operating flow rate.
image
As seen, the supply side is now able to turn off the unused constant speed branch pump on the second parallel branch, so the loop-side flow rate is cut in half.  Which will obviously allow a reduced energy footprint for these configurations.

Myoldmopar:
Need to make a plant-only file with constant speed branch pumps on the supply side with a common pipe.  Follow this loading:
 - No load
 - Partial load for chiller 1
 - Full load for chiller 1
 - Full load for chiller 1 + Partial load for chiller 2
 - Full load for both chillers
 - Exactly the same process back down

I'll probably schedule it for at least a couple hours each so things can settle out with each change.

Then take out the common pipe and demand side pump and redo it all.  Should be just fine, but want to verify.

Open a ticket to properly test SetupFlowRequest with a plan for covering it really well.

Myoldmopar:
@mjwitte I'm going to have to merge this in and follow it up with a separate PR that is likely to include some diffs. This has helped the flow dispatch for branch pumps, but there are still some cases that need work (as is always the case with constant speed branch pumps and common pipes).

Are you available to review this?  There's really only one functional change, and that's a resimulation of the loopside with the constant speed branch pump branches disabled if they didn't receive any load during the first pass.  This resulted in the improved behavior shown [above](https://github.com/NREL/EnergyPlus/pull/6493#issuecomment-371639024).

mjwitte:
@Myoldmopar On it . . .

mjwitte:
@Myoldmopar Please update/clean up the PR description to reflect the final result.

mjwitte:
@Myoldmopar Looking at 3522-in-from-kbenne and, yes, the total plant supply flow is lower now as expected, but it's still being split evenly across the two chiller branches. Both pumps are operating at half the ChW supply flow rate.

Myoldmopar:
That's what I saw too in my smaller test file.  But I'm afraid a fix on that could cause some diffs, so that's why I want to get this incremental step with refactoring in place with essentially no diffs first.

Myoldmopar:
And I updated the PR body to describe what I did plus why it needs a followup.

src/EnergyPlus/Plant/LoopSidePumpInformation.hh
```diff
@@ -0,0 +1,81 @@
...
+					PumpFlowRate(0.0),
+					CurrentMinAvail(0.0),
+					CurrentMaxAvail(0.0),
+					PumpHeatToFluid(0.0) {}
+
+		};
+	}
+}
+
+#endif
```
mjwitte: Missing newline here.

src/EnergyPlus/Plant/PlantConvergencePoint.hh
```diff
@@ -0,0 +1,78 @@
...
+					Real64 const TemperatureValue
+			) :
+					MassFlowRateHistory(NumConvergenceHistoryTerms, MassFlowRateValue),
+					TemperatureHistory(NumConvergenceHistoryTerms, TemperatureValue) {}
+
+		};
+	}
+}
+
+#endif
```
mjwitte: Another missing newline.

src/EnergyPlus/Plant/PlantLoopSolver.cc
```diff
@@ -93,18 +93,14 @@ namespace PlantLoopSolver {
...
-
 	// MODULE VARIABLE DEFINITIONS
 	Real64 InitialDemandToLoopSetPoint;
 	Real64 CurrentAlterationsToDemand;
 	Real64 UpdatedDemandToLoopSetPoint;
 	Real64 LoadToLoopSetPointThatWasntMet; // Unmet Demand
 	Real64 InitialDemandToLoopSetPointSAVED;
-	int RefrigIndex( 0 ); // Index denoting refrigerant used (possibly steam)
+    bool CurrentLoopSideIsConstantSpeedBranchPumped;
+    int RefrigIndex( 0 ); // Index denoting refrigerant used (possibly steam)
```
mjwitte: Got some spaces here instead of tab, basically all throughout this file where changes were made. I know that'll change at some point, but looks messy here. Clean up now or let it ride?

src/EnergyPlus/PlantUtilities.cc
```diff
@@ -2467,68 +2295,521 @@ namespace PlantUtilities {
...
+		return true;
+
+	}
+
+
+	void
+	ScanPlantLoopsForObject(
+			std::string const & CompName,
+			int const CompType,
+			int & LoopNum,
```
mjwitte: Extra tabs here, some of those happening in some of the new header files, too. Not a big deal.

tst/EnergyPlus/unit/Pipes.unit.cc
```diff
@@ -0,0 +1,78 @@
...
+			" Pipe Outlet Node 2;    !- Outlet Node Name",
+		});
+
+		ASSERT_TRUE(process_idf(idf_objects));
+		Pipes::GetPipeInput();
+		EXPECT_EQ(2u, Pipes::LocalPipe.size());
+		EXPECT_EQ(DataPlant::TypeOf_Pipe, Pipes::LocalPipe(1).TypeOf);
+		EXPECT_EQ(DataPlant::TypeOf_PipeSteam, Pipes::LocalPipe(2).TypeOf);
+	}
+}
```
mjwitte: Missing newline.

tst/EnergyPlus/unit/PlantComponentTemperatureSources.unit.cc
```diff
@@ -0,0 +1,134 @@
...
+		PlantComponentTemperatureSources::SimWaterSource("FLUIDSOURCE", dummyInteger, compIndex, dummyRunFlag, firstHVACIteration, initLoopEquip, myLoad, maxLoad, minLoad, optLoad, getSizingFactor, sizingFactor);
+		EXPECT_NEAR(0.05, waterSource1.MassFlowRate, 0.001);
+
+		// Do this for scheduled temperature
+		// NumOfTimeStepInHour = 1; // must initialize this to get schedules initialized
+		// MinutesPerTimeStep = 60; // must initialize this to get schedules initialized
+		// ProcessScheduleInput(); // read schedules
+
+	}
+}
```
mjwitte: Newline.

tst/EnergyPlus/unit/PlantPipingSystemsManager.unit.cc
```diff
@@ -1623,3 +1625,173 @@ TEST_F( EnergyPlusFixture, SiteGroundDomainBasement_CheckInputs_BadTimestepSelec
...
+		"    2.05,                    !- X Position {m}",
+		"    1.25,                    !- Y Position {m}",
+		"    DecreasingZ;             !- Flow Direction"
+	});
+
+	ASSERT_TRUE( process_idf( idf_objects ) );
+
+	// Setup the plant itself manually
+	DataPlant::TotNumLoops = 1;
+	DataPlant::PlantLoop.allocate(1);
```
mjwitte, line 1755: More spaces vs tabs.

tst/EnergyPlus/unit/PlantUtilities.unit.cc
```diff
@@ -84,3 +87,342 @@ TEST_F( EnergyPlusFixture, PlantUtilities_RegisterPlantCompDesignFlowTest1 )
...
+	// We expect it to be false here since the temperature changed
+	EXPECT_FALSE(PlantUtilities::CheckPlantConvergence(1, 1, false));
+	// But if we run it 4 more times and let the value propagate, we expect it to be stable and pass
+	// Need to call it 5 times total to fully initialize the history
+	for (int i = 1; i < 5; ++i) {
+		PlantUtilities::LogPlantConvergencePoints(false);
+	}
+	EXPECT_TRUE(PlantUtilities::CheckPlantConvergence(1, 1, false));
+
+}
```
mjwitte: Newline.

mjwitte (approved):
@Myoldmopar Changes approved in general with expectation of followup to allocate the new flow rate to the correct branches. Beautiful refactor and lots of nice shiny new unit tests. There are some missing newlines noted here - please fix those and then I'll merge. Also some whitespace items noted - your call whether to clean up or not given the upcoming re-formatting.

mjwitte:
Spoke a little too soon - running the unit tests serially, `EnergyPlusFixture.PipingSystemFullSimulation` fails [here](https://github.com/NREL/EnergyPlus/blob/d4befa1c6d4c96eaa330958ede253e947d6c0a04/src/EnergyPlus/PlantPipingSystemsManager.cc#L530) with a fatal input error - not sure what the error is. Probably something missing in clear_state or?

Myoldmopar:
OK, I'll check it out.  Thanks.

And on the formatting, it looks like we are going to be doing that during the calm after the bug fix freeze, before the actual release.  So yeah I'm not super worried about formatting on any PRs this week.

mjwitte:
It's failing in `VerifyUniqueInterObjectName` when called from [here](https://github.com/NREL/EnergyPlus/blob/d4befa1c6d4c96eaa330958ede253e947d6c0a04/src/EnergyPlus/PlantPipingSystemsManager.cc#L1302) Error: Site:GroundDomain:Basement=CoupledBasement is a duplicate name - probably leftover from `EnergyPlusFixture.SiteGroundDomainBasement_CheckInputs_BadTimestepSelection`.

Myoldmopar:
I reproduced the failure locally, I'll debug in and fix it and this can be good to go, afaik.

Myoldmopar:
Oh OK, well that's helpful then! Thanks!

Myoldmopar:
It looks like a `clear()` call on the unique names map in the piping system `clear_state` helped it get past that unit test successfully when running them serially.  I'll push that change up as soon as the tests get all done.

Myoldmopar:
@mjwitte OK, I cleaned up the newlines and spaces/tabs stuff anyway.  The tests are still churning away on FluidPropertiesRefData, but they didn't fail on the one we could both produce, so I think it is good to go.

Myoldmopar:
Alright, I may have a working fix for this, but honestly, I'd still prefer to get it in by itself.  It may cause diffs in common pipe files.

mjwitte:
Just waiting on Marvin to finish up and then I'll merge. Will be helpful to separate out the functional fixes in a separate PR. I know this one has some initial functional fixes, too, but you can point back to those.

Myoldmopar:
Thanks @mjwitte.  Doing it now.  I've already got the branch pushed and I'm writing the PR description.  Fingers crossed...

Myoldmopar:
Build warning?  Dang.  It's just one; a 'sometimes-uninitialized' warning because the code doesn't know that the if-checks are actually fully inclusive.  

I will fix that real quick.  Your call on whether to wait on CI.

mjwitte:
OK, Marvin says the build warning is gone. Merging.

Issue #6485, opened 2018-02-21
VRF power in heat recovery using ThermostatOffsetPriority is 0 when coils are active


rraustad:
Issue overview
--------------
Testing of the VariableRefrigerantFlow_5Zone example file and changing the master thermostat priority control type to ThermostatOffsetPriority shows that power sometimes goes to 0 when the coils are actively cooling or heating. The figure below exemplifies the problem.

    AirConditioner:VariableRefrigerantFlow,
      ThermostatOffsetPriority,            !- Master Thermostat Priority Control Type

![vrfpower0usingtstatoffsetpriority](https://user-images.githubusercontent.com/5778156/36482234-831ac38e-170a-11e8-9aca-bb57bdd0a95d.jpg)
 

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [x] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6498

PR #6498, opened 2018-02-23
VRF power in heat recovery using thermostat offset priority is 0 when coils are active

rraustad:
Pull request overview
---------------------
The VRF power reported sometimes equals 0 when the Master Thermostat Priority Control Type = ThermostatOffsetPriority. This change corrects Issue #6485. Also corrected is the energy consumption during heat recovery and the start up multiplier used to derate performance at the beginning of heat recovery mode. A warning that notifies users when outdoor conditions exceed heat recovery mode operating temperatures was also improved to report only if heat recovery mode is requested.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -6689,39 +6490,14 @@ namespace HVACVariableRefrigerantFlow {
...
-		//    CondenserInletNode = VRF(VRFCond)%CondenserNodeNum
 		CondenserOutletNode = VRF( VRFCond ).CondenserOutletNodeNum;
 
 		Node( CondenserOutletNode ).Temp = VRF( VRFCond ).CondenserSideOutletTemp;
 
-		//    Node(CondenserInletNode)%MassFlowRate = CondenserWaterMassFlowRate
 		Node( CondenserOutletNode ).MassFlowRate = CondenserWaterMassFlowRate;
-
 		Node( CondenserOutletNode ).MassFlowRateMaxAvail = Node( CondenserOutletNode ).MassFlowRateMaxAvail;
 		Node( CondenserOutletNode ).MassFlowRateMinAvail = Node( CondenserOutletNode ).MassFlowRateMinAvail;
```
rraustad, line 6480: Well these 2 lines look suspect. Should be outlet = inlet.

rraustad: Well, that uneasy feeling of not knowing exactly how to use these Min/Max/MinAvail/MaxAvail variables is upon me.

1) UpdateVRFCondenser is only called for water-cooled equipment
2) Water based equipment should not set Min/MaxAvail since SetComponentFlowRate handles that. These lines don't hurt anything here since it writes outlet = outlet (but should be removed eventually).
3) Setting Node( CondenserOutletNode ).MassFlowRate = CondenserWaterMassFlowRate; here is OK but redundant since SetComponentMassFlowRate already did that. 
4) I noticed the VRF TU's are not passing MinAvail and MaxAvail through to the coil outlet node as do other air-side equipment. Probably works because TU's are zone equipment. 
5) Do no harm, no changes here at this point.

rraustad:
CI shows 1 file with big diff's, HVACTemplte_5ZoneVRF. Not sure where these diff's are coming from.

The eplusout.csv.absdiff file says there are diff's in the Cooling Tower Heat Transfer Rate (and other tiny diff's but this is the largest). When I plot V8.8, V8.9 develop and V8.9 this branch, this data are identical. Even on this scale you can see there is not a difference of 600 W when the unit is off at x-axis time 30 and 31 (6:00 AM and 7:00 AM)

So this looks good to me and the diff's are explained as not real.

hvactemplate_5zonevrfdiffs

rraustad:
Here's the diff's side-by-side. There is a 12 W diff right at startup (see This branch column at 9 AM) that I can't explain, but from there the results are identical. This data was used to create the graph above. 

vrfdiffs

Myoldmopar:
@rraustad I don't expect you to also address this one tonight, but if you look at the CI results [here](https://github.com/NREL/EnergyPlus/pull/6498#commitcomment-27877419) you'll see one example file (HVACTemplate-5ZoneVRF) is actually failing, not just showing diffs.  If you want to point me to a possible issue, I'm happy to try to find a few minutes to help track it down over the next day or so, but no guarantee.  Thanks for all the effort.

rraustad:
This is what worried me. First off that's HVAC template that failed, not the example file. So is it what I hacked up (or didn't) in expand objects or something else?

mjwitte:
@Myoldmopar The test failure was a divide by zero. Let's see if CI is happy now.

Myoldmopar:
Pulled latest develop in here; letting CI run overnight, anticipate merging in the morning.

Myoldmopar:
@nrel-bot-2 was taking a while doing the Linux builds.  I went ahead and merged in some other PRs, and they would start causing diffs here, so I pulled develop into this branch and I'll get a full set of CI results.  It is looking good though from the initial results so I anticipate it dropping in easily.

PR #6499, opened 2018-02-23
Move setpoint calculation from zone time step to system time step

lgu1234:
Pull request overview
---------------------
This is a followup issue for OnOff Thermostat. When the condition is satisfied, the setpoint is revised at every system time step, instead of every zone time step. 

Differences are expected in SingleFamilyHouse_TwoSpeed_CutoutTemperature.idf

There are other 5 files with differences due to report frequency change from "Zone" to "System": ZoneThermostatSetPointLo and ZoneThermostatSetPointHi.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


lgu1234:
@mjwitte I uploaded the code a while ago for review.

Myoldmopar:
@lgu1234 two test files are failing to run, not just showing diffs.  It is an array index problem, so be on the lookout.  Since we are right against the freeze date, I'm going to anticipate we won't get this addressed and re-reviewed, so I'm going to put this in the next milestone.  If we magically get done with it that's great too.

lgu1234:
@Myoldmopar I am going to take a look and hope I will fix today.

lgu1234:
@Myoldmopar I fixed failure and merge the code again. I need to get this new feature for the spring release. I am going to change milestone.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3340,7 +3331,7 @@ namespace ZoneTempPredictorCorrector {
 				LoadToHeatingSetPoint = AIRRAT( ZoneNum ) * ( TempZoneThermostatSetPoint( ZoneNum ) - ZoneT1( ZoneNum ) ) + TempDepZnLd( ZoneNum ) * ( TempZoneThermostatSetPoint( ZoneNum ) ) - TempIndZnLd( ZoneNum );
 			}
 			if ( RAFNFrac > 0.0 ) LoadToHeatingSetPoint = LoadToHeatingSetPoint / RAFNFrac;
-			if ( TempControlledZone( RelativeZoneNum ).HeatOffFlag ) {
+			if ( HeatOffFlag ) {
 				LoadToHeatingSetPoint = 0.0;
```
mjwitte: @lgu1234 I don't understand why `LoadToHeatingSetPoint` is set to 0.0 here. I think I follow the `HeatOffFlag` logic, but once the zone reaches the setpoint+cutoff and `HeatOffFlag=true`, then the thermostat setpoint should revert to the lower temperature and the LoadToHeatSetPoint should be based on that (the original setpoint).
@Myoldmopar This behaves far better than what's in develop, but I think there's one more improvement that could be made. I can experiment with that tomorrow or we merge this as-is and add the improvement post-release if it proves to be useful.

Myoldmopar: I will move this to the end of my list of tabs as they are ordered by when they should drop in.  If you have some time to experiment tomorrow that would be fine.  If it doesn't make it into this, that's fine too.

lgu1234: @mjwitte Thanks for your comments. I think of the same way by removing line LoadToHeatingSetPoint = 0.0. I will test tonight and let you know what happens.

lgu1234:
@mjwitte I removed reset of load prediction. The results are better. The indoor temperature is between 22 and 24 during heating and 24.6 and 26.6 during cooling. Please let me know your comments after your experiment.

Myoldmopar:
Thanks @lgu1234; I just merged in the latest develop branch so no unrelated diffs will pop up.  I'm letting CI run overnight on these branches before merging anything else, so the CI results will only be related to the changes on this branch.  If @mjwitte agrees with these changes and CI is happy it can go in in the morning.

lgu1234:
@Myoldmopar Thanks.

PR #6042, opened 2017-03-04
Fix static Fortran compilation

jasondegraw:
Pull request overview
---------------------
This fixes #5854, but doesn't yet fix a couple of related issues that I discovered while looking into the original issue. The issue arises when trying to find library dependencies of the Fortran executables, which are static for some platforms/compilers. It worked in CMake 2.8.X, which is what Ubuntu 14.04 has, but for newer versions of CMake (e.g. the 16.04 version) it does not. Older versions of CMake were not checking the return code of `ldd`, which is 1 (failure) for static exes. Newer versions are checking, and the build bombs out. The earlier CentOS fix only fixed things for CentOS/Red Hat. The fix is to not call CMake's GET_PREREQUISITES on static Fortran executables. 

The related issues are:
 - [x] There are some naming irregularities that could use straightening out, investigating that, too.
 - [x] The missing GITHUB_TOKEN message in the Linux package build is a little vague.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
 - [ ] Code style (parentheses padding, variable names)
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check -- verify this

PR #6504, opened 2018-02-27
Add default TeX interaction if not set

Myoldmopar:
Pull request overview
---------------------
This works around the odd cases where CI, and now my local Windows VM, can't build the documentation because the TeX interaction variable isn't being set by CMake.  Other devs seem to still be able to build package afaik.  Anyway, this provides the same effective behavior there was before, just in a different way.  And now CI should build package.  I'll kick off a tag to verify.  If all is well this should drop in right away.


jasondegraw:
@Myoldmopar I think this will clash with my static Fortran fix branch, where I've also made some (different) changes to get past this. Could we fix this over there and get that PR merged?

Myoldmopar:
Thanks @jasondegraw. It doesn't look like this will clash with yours actually.  You have added defines for TEX_INTERACTION on the Windows builds, which is great.  Frankly though, CI should have picked up the default value for TEX_INTERACTION even if it wasn't defined explicitly.  Somehow CI (and my local machine) was making it all the way to the XeLaTeX call with an empty value on that variable.  This change here simply protects that XeLaTeX call from being invalid if somehow TEX_INTERACTION doesn't make it to that point.  I think it's an OK thing to bring this in here in addition to your changes over there.  It will at worst simply be an extra check on that variable.  

I already kicked off a tag of this branch, so I'll at least keep it in place for a while to see if Windows CI is more happy with this change.  If so, we can merge it into your branch or continue discussion.

jasondegraw:
@Myoldmopar I'm fine either way, just want to make sure #6042 goes in for 8.9. I'm pretty sure I've had this issue on my local machine for some time due to my nonstandard TeX installation.

Myoldmopar:
So this branch and one other are seeing non-UTF8 issues in decent_ci in stdout, which fails in a call to split the string into lines.  

I started investigating by looking at how decent_ci is protecting against this.  It is currently trying to encode stdout with an argument `:invalid=>:replace`, which should replace any invalid characters it finds.  I made a [branch](https://github.com/lefticus/decent_ci/compare/ModifyEncodingArgs?expand=1) of decent_ci with some added arguments that will also replace any undefined characters.  I think once this gets tested, it should help protect the final call to split and make this more robust.  I hoped I could just point this branch here on E+ to the new branch over on decent_ci, but I don't see an option to do it.  I'd probably have to modify the decent_ci_runner config file.

I then took this branch and tried to sanitize all non-utf8 characters in the source, docs, and test files folders.  I pushed this change up now.  As far as I could tell they were comment only changes in the IDFs, but I haven't looked at the Files Changed tab here yet, so it may reveal meaningful changes.  For the error itself, I am betting it was the TeX file causing it, since that gets echoed during builds.  If the test files all still run and there aren't any diffs, then this will clean up the bad degree symbols.  If the test files fail, I'll revert that and push the TeX change only back up.

Myoldmopar:
@mjwitte @jasondegraw OK, this looks great now.  The Linux CI is now all happy, no more UTF-8 issues.  I think it was mostly the tex change, but overall there are now a lot less bad characters in the repo.  The one file that is showing ERR diffs is just because the order of the two lines in the ERR file, so one line has a `~~~~~~~~~` type of prefix and the other has a `~~~***~~~` type of prefix.  I'd like to merge this in right away, since it is fairly large but no impact.  Any objections?

PR #6505, opened 2018-02-28
Speed up DataSetFixture.FluidPropertiesRefData unit test

mjwitte:
Pull request overview
---------------------
Add a flag for `preserveIDFOrder` which disables functions `InputProcessor::getIDFObjNum and getJSONObjNum` when it is false. Added this to the DataSetFixture.FluidPropertiesRefData unit test. Cuts the test time from 1435k to 786k ns.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@mbadams5 @Myoldmopar I see why this unit test takes so long. `FluidProperties::GetFluidPropertyData` overuses `GetObjectItem` by looping through and getting all of the objects of each type multiple times to do validations. Instead of just getting them all first, it's looking for matching fluid names, etc., and then finally keeps them to put in the data structure. 

For example, here's the [first pass](https://github.com/NREL/EnergyPlus/blob/06606575801583060a3534d6af72b0f2d91fd13d/src/EnergyPlus/FluidProperties.cc#L1129) through all of the FluidProperties:Superheated objects. There are over 1500 of these in the dataset file.
Here's the [second pass](https://github.com/NREL/EnergyPlus/blob/06606575801583060a3534d6af72b0f2d91fd13d/src/EnergyPlus/FluidProperties.cc#L1179), the [third pass](https://github.com/NREL/EnergyPlus/blob/06606575801583060a3534d6af72b0f2d91fd13d/src/EnergyPlus/FluidProperties.cc#L1179), and the [fourth pass](https://github.com/NREL/EnergyPlus/blob/06606575801583060a3534d6af72b0f2d91fd13d/src/EnergyPlus/FluidProperties.cc#L1298).

Oh, and if there's more than one [fluid type](https://github.com/NREL/EnergyPlus/blob/06606575801583060a3534d6af72b0f2d91fd13d/src/EnergyPlus/FluidProperties.cc#L626-L629) we do it all again. 

There 12 fluid names in the dataset file. So we're getting all 1500 FluidProperties:Superheated obejcts 48 times instead of once.

Myoldmopar (approved):

Issue #5910, opened 2016-10-17
Chiller:Absorption hot water flow problems


mjwitte:
## Issue overview

User notes that Chiller:Absorption hot water source flow is always at max, even though the HW loop has variable flow pumps and the chiller flow mode is NotModulated.  This results in the source HW loop overheating late in the cooling season as the pump heat exceeds the heat requirement from the absorption chiller.  

I tried a test of with the chiller flow mode changed to LeavingSetpointModulated, but this failed with very cold temperatures leaving the chiller hot water outlet node, -90C.  The flow rates do vary with this flow mode, and the leaving hot water temperatures look reasonable on the summer design day, and again when the chiller comes on for a few days in April. But on the day that the simulation fails, the chiller loads and flow rates seem reasonable, but the leaving hot water temperature nosedives.  See ABSORBERHWOUTLETNODE:System Node Temperature.
### Details

Some additional details for this issue (if relevant):
- Platform Win64
- Version of EnergyPlus 8.6.0
- helpdesk ticket 11467
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added EnergyPlusDevSupport\DefectFiles, use ITA_BOLOGNA-BORGO PANIGALE_IGDG.epw
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Nigusse:
Chiller:Absorption with flow type specified as "NoModulated" results maximum flow through the generator hot water loop by design, and when there is a cooling load on the chiller, the generator hot water loop pump adds heat to the loop fluid depending on the water flow rate and pump pressure rise. There are two issues that I noted when investigating this bug issue:

(1) The generator hot water loop pump pressure rise specified is very high, 1247600.0 Pa. The loop pump max fluid heat gain rate is 19,193.8 W.  This is much higher than the load on the generator during low cooling load season. This is April in moderate climate location Bologna, Italy weather. Thus, there is a constant heat build up in the hot water loop. Reducing the pump pressure rise equal to that of the chilled water loop pump (244,435 Pa), reduced the heat build up and the example file to run to completion. Sometimes, pump fluid heat was enough to provide heat to the generator.

[...]

PR #6511, opened 2018-03-06
Chiller:Absorption hot water Flow Mode control problem 

Nigusse:
Issue #5910 

Pull request overview
---------------------
This is a defect for Chiller:Absorption that uses hot water as a heat source for generator and the hot water loop is connected to the generator hot water inlet and outlet nodes. This fix modified how the generator hot water flow rate is calculated when the Chiller Flow Mode is "LeavingSetpointModulated". Diffs are expected if the absorption chiller flow mode specified is "LeavingSetpointModulated" and there are generator hot water inlet and outlet nodes specified in the chiller object. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)


Nigusse:
Summary:
- I think the NotModulated flow mode problem in the original defect file is not directly related to the chiller model. I suspect in part it is a bad input problem. I tested a modified version of the original defect file by removing the two HEX connecting hot water loops and adding a boiler as heat source. Investigating further showed that the fatal out is related to Shading Surfaces and DayLighting calculations. After removing all the daylighting objects the modified defect file run to completion successfully.

- Also created a new defect file based on example file "AbsorptionChiller".  The new defect file did not fail for NotModulated chiller flow mode but it did for flow mode "LeavingSetpointModulated". This fix addresses the temperature run-down problem for flow mode "LeavingSetpointModulated".
   
- Added new and modified defect files

Nigusse:
@Myoldmopar Ready for review.

src/EnergyPlus/ChillerAbsorption.cc
```diff
@@ -1541,24 +1541,35 @@ namespace ChillerAbsorption {
...
+				if ( PlantLoop( GenLoopNum ).LoopSide( GenLoopSideNum ).FlowLock == 0 ) {
+					if ( (BLASTAbsorber( ChillNum ).FlowMode == ConstantFlow) || (BLASTAbsorber( ChillNum ).FlowMode == NotModulated) ) {
+						GenMassFlowRate = BLASTAbsorber( ChillNum ).GenMassFlowRateMax;
+					} else { // LeavingSetpointModulated
+						// since the .FlowMode applies to the chiller evaporator, the generater mass flow rate will be proportional to the evaporator mass flow rate
+						Real64 GenFlowRatio = EvapMassFlowRate / BLASTAbsorber( ChillNum ).EvapMassFlowRateMax;
+						GenMassFlowRate = min( BLASTAbsorber( ChillNum ).GenMassFlowRateMax, GenFlowRatio * BLASTAbsorber( ChillNum ).GenMassFlowRateMax );
+					}
+				} else { // If FlowLock is True
+					GenMassFlowRate = Node( GeneratorInletNode ).MassFlowRate;
```
Myoldmopar, line 1559: The FlowLock logic isn't needed in the component model.  The component model should just calculate a flow to request and call SetComponentFlowRate.  If FlowLock is false on this loop, that method will set this request on the loop structure, and return a suitable flow based on the original request and any flow limits.  If FlowLock is true on this loop then SetComponentFlowRate will just return the inlet node flow rate.  If there are any other changes I find on this review, then this will have to be remedied and will push this PR past the freeze.  If there isn't any other changes to be made on this branch, then it can go in.

tst/EnergyPlus/unit/ChillerAbsorption.unit.cc
```diff
@@ -0,0 +1,1826 @@
...
+		"    Hot Loop Operation,      !- Name",
+		"    PlantEquipmentOperation:HeatingLoad,  !- Control Scheme 1 Object Type",
+		"    Purchased Only,          !- Control Scheme 1 Name",
+		"    ON;                      !- Control Scheme 1 Schedule Name",
+
+		"  PlantEquipmentOperation:HeatingLoad,",
+		"    Purchased Only,          !- Name",
+		"    0,                       !- Load Range 1 Lower Limit {W}",
+		"    1000000,                 !- Load Range 1 Upper Limit {W}",
+
```
Myoldmopar, line 1824: 👍

Myoldmopar (approved):
OK, this is good.  Please take my comment about flow lock into consideration for all future code changes.  But this is good.  I'll merge.

Issue #6480, opened 2018-02-16
Followup issues for Enhancements to Data Center Model


mjwitte:
Issue overview
--------------
Followup items from #6154.

- [x] I/O Ref needs a definition for the new output variable "Zone ITE Adjusted Return Air Temperature".
- [x] While in the I/O Ref, it would be nice to break up the description of Air Flow Calculation Method into separate paragraphs for the two methods with the choices in bold.
- [ ] Need to finalize transition rules and add transition code.
- [x] ThermalComfort.cc line 2290 has a stray tab at the end.
- [x] Move check for terminal unit type from [here](https://github.com/NREL/EnergyPlus/blob/c0223f6ed62d7242a3a84cd568a402706033ba6a/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L227-L235) to `InitZoneAirLoopEquipment`.
- [x] Maybe more comments on the method - out of time to review further.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

xuanluo113:
#6514

Myoldmopar:
Closed via #6514

PR #6514, opened 2018-03-08
Fix Followup issues for Enhancements to Data Center Model

xuanluo113:
Pull request overview
--------------
Fix issues mentioned in #6480.
Followup items from #6154.

- [x] I/O Ref needs a definition for the new output variable "Zone ITE Adjusted Return Air Temperature".
- [x] While in the I/O Ref, it would be nice to break up the description of Air Flow Calculation Method into separate paragraphs for the two methods with the choices in bold.
- [x] Need to finalize transition rules and add transition code.
- [x] ThermalComfort.cc line 2290 has a stray tab at the end.
- [x] Move check for terminal unit type from [here](https://github.com/NREL/EnergyPlus/blob/c0223f6ed62d7242a3a84cd568a402706033ba6a/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L227-L235) to `InitZoneAirLoopEquipment`.
- [x] Maybe more comments on the method - out of time to review further.

### Checklist


xuanluo113:
@mjwitte 
Thanks for taking care of these. I did the checked items. Not sure how to deal with the transition rules. And please let me know if other specific comments.

doc/input-output-reference/src/overview/group-internal-gains-people-lights-other.tex
```diff
@@ -2493,6 +2497,8 @@ \subsubsection{Inputs}\label{inputs-9-009}
   Zone,Average,ITE Air Inlet Relative Humidity Difference Above Operating Range {[}\%{]}
 \item
   Zone,Average,ITE Air Inlet Relative Humidity Difference Below Operating Range {[}\%{]}
+\item
+  Zone,Average,Zone ITE Adjusted Return Air Temperature {[}C{]}
```
mjwitte, line 2501: New output variable has been added to this list - good. But it also needs to be added below with a heading and a definition.

src/EnergyPlus/ZoneAirLoopEquipmentManager.cc
```diff
@@ -530,6 +521,16 @@ namespace ZoneAirLoopEquipmentManager {
 			EachOnceFlag( AirDistUnitNum ) = false;
 		}
 
+		if ( AirDistUnit( AirDistUnitNum ).ZoneNum != 0 && FirstHVACIteration && DataHeatBalance::Zone( AirDistUnit( AirDistUnitNum ).ZoneNum ).HasAdjustedReturnTempByITE ) {
+			for ( int AirDistCompNum = 1; AirDistCompNum <= AirDistUnit( AirDistUnitNum ).NumComponents; ++AirDistCompNum ) {
+				if ( AirDistUnit( AirDistUnitNum ).EquipType_Num( AirDistCompNum ) != SingleDuctVAVReheat && AirDistUnit( AirDistUnitNum ).EquipType_Num( AirDistCompNum ) != SingleDuctVAVNoReheat ) {
+					ShowSevereError( "The FlowControlWithApproachTemperatures only applies to ITE zones with single duct VAV terminal unit." );
+					ShowContinueError( "The return air temperature of the ITE will not be overwritten." );
```
mjwitte: Since the next line is a fatal error, this line is not needed.
In the line above I would change the wording from "only applies to . . . " to "only works with  . . . ".

Placed here, this check will execute every time step, looping through all AirDistUnits. This should move up into the EachOnceFlag block above, and it doesn't need to loop through the AirDistUnits, Just check if the zone has HasAdjustedReturnTempByITE  and the terminal unit type is not single duct VAV (reheat or no reheat).
Also, don't need to pass in FirstHVACIteration.

Myoldmopar: @mjwitte if I'm reading this correctly, it's not looping through all the AirDistUnits.  It's looping through the components on the current AirDistUnit.  It could be moved into the one time init section above it, but at this point, I've already done the transition and added an only-minorly-helpful output variable description to the doc.  

I verified the transition by looking at the diffs after the transition ran, they looked perfect.  Running the transitioned file was successful as well.  I built the doc after I made my tiny change and it built fine.

I'm not going to spend more of my time on it.  If you want to include further polishing of this in yet another followup issue for @XuanLuoLBL that's fine with me, but I think I'll merge this once CI starts coming back green.

mjwitte: @Myoldmopar OK thanks. I misread the loop, but it's still unnecessary to even do this once every time step in every run with air distribution units.  I can fix that next after I get the coil sizing stuff up.

Myoldmopar: That's up to you.  This isn't **your** pull request to be fixing at the last minute...

Myoldmopar:
Diffs are clean, err file diffs are unrelated.

src/EnergyPlus/ZoneAirLoopEquipmentManager.cc
xuanluo113: Thanks @Myoldmopar for cleaning up the doc issue. I fixed this accordingly.

mjwitte (approved):
@XuanLuoLBL @Myoldmopar This is ok to merge pending CI results.

Myoldmopar:
After some of the prior merges, there was a conflict.  It looked like @mjwitte had touched the init functions in ZoneAirLoopEquipmentManager in his sizing work, and of course that was changed in this one as well.  @XuanLuoLBL @mjwitte Could both of you have a peek at the current changes and make sure they make sense?

I'm going to let CI run on this branch overnight.  If you are happy with the changes and CI is happy then this can go in in the morning.

Issue #5891, opened 2016-09-28
OA system controller list apparently requires the OA controller to come first followed by water coil controllers


mjwitte:
## Issue overview

This input

```
AirLoopHVAC:ControllerList,
    Central_AHU OA System Controllers,  !- Name
    Controller:WaterCoil,    !- Controller 1 Object Type
    Central_AHU Dehum Coil Controller,  !- Controller 1 Name
    Controller:OutdoorAir,   !- Controller 2 Object Type
    Central_AHU OA Controller;  !- Controller 2 Name
```

results in this error

```
   **  Fatal  ** SimOAController: Outside Air Controller not found=CENTRAL_AHU DEHUM COIL CONTROLLER
```

Reversing the order of the controller list eliminates the error.  This needs to be trapped with a better error message during input processing for AirLoopHVAC:ControllerList, or automatically fix the order.
### Details

Some additional details for this issue (if relevant):
- Version of EnergyPlus 8.6.0
- helpdesk ticket 11515
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [x] Defect file added (EnergyPlusDevSupport\DefectFiles, reproduced with modified 5ZoneAirCooled)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6518

PR #6518, opened 2018-03-11
OA system controller list apparently requires the OA controller to come first followed by water coil controllers

Nigusse:
#5891

Pull request overview
---------------------
This is defect. The order of the controller doesn't matter if we add the OAController name and index to the OutsideAirSysProps struct instead of rearranging the OA controller order. Added a member variables "OAControllerName" and "OAControllerIndex" to the "OutsideAirSys" data struct and use these variables to get to the OA controller during simulation. Previously the SimOAController routine was run with argument OAControllerIndex value always zero hence every run required searching for OA controller index using "FindItemInList( CtrlName, OAController )". This fix is expected to slightly improve performance. Diffs are not expected. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


src/EnergyPlus/MixedAir.cc
```diff
@@ -492,21 +491,21 @@ namespace MixedAir {
 		//    CtrlName = OutsideAirSys(OASysNum)%ControllerName(CtrlNum)
 		//    CALL SimOAController(CtrlName,FirstHVACIteration)
 		//  END DO
-		CtrlName = OutsideAirSys( OASysNum ).ControllerName( 1 );
 		CurOASysNum = OASysNum;
-		SimOAController( CtrlName, OutsideAirSys( OASysNum ).ControllerIndex( 1 ), FirstHVACIteration, AirLoopNum );
+		auto & CurrentOASystem( OutsideAirSys( OASysNum ) );
```
Myoldmopar, line 495: 👍

Myoldmopar (approved):

PR #6522, opened 2018-03-12
Continuation of Constant Speed Branch Pump Fixes

Myoldmopar:
Pull request overview
---------------------
Pull request #6493 ended up being a partial solution to branch pumping with minimal impact on solution results.  This pull request continues that work, but I was worried it may introduce diffs, so it is a smaller set of code changes to review.  This includes the prior work, so at the moment the pull request is big, but once #6493 gets merged in, this will get reduced to a very small number of changes.

When I was working on #6493 I ultimately noticed that something was going on with the energy balance in the plant loop with the common pipe.  The main issue is that the demand that the supply side needed to meet was not including the common pipe flow.  The demand side would operate at a given flow rate and impose a demand on the loop.  The supply side would come along and determine the flow rate that it wanted to run at (possibly leaving some flow through the common pipe), and calculate the loop demand based on that flow rate.  This ultimately left the common pipe flow behind, and the loops would at best struggle to meet demand, and at worst, there seemed to be an energy imbalance.

I added in a block of code to capture any common pipe flow, and include it in the demand calculations.  That way, when the flow finally got back to the demand side components, it would be the full flow again, and at the desired plant setpoint temperature.  This seemed to work.  Really well.  I set up an example file with the following attributes:
 - Demand side is a single series branch, with a constant speed secondary pump driving a load profile
 - The load profile gradually increases demand to a peak of 50 kW then reduces it back to zero
 - The load profile flow request is constant at ~2 kg/s, even when there is no load requested
 - The loop uses a common pipe to allow a different primary/secondary flow
 - The supply side is two parallel branches, each with a constant speed primary pump and a boiler
 - Each primary pump is sized to 1 kg/s so that when they are both on, there should be no common pipe flow
 - Each boiler is also sized to 1 kg/s, and a peak heating rate of 25 kW, so that at full load, both boilers should be running at near full capacity
The results are showing in these plots:

The first plot shows the heat extracted by the load profile object and the demand that the supply side calculates that it needs to meet.  Note the small difference between the two is the heat added to the loop by the pumps. :+1: 
image

This second plot shows the demand and supply side mass flow rates.  The demand side has a constant request, while the supply side is staged according to how many boilers are on. :+1: 
image

This third plot shows the heat being added to the loop by each boiler.  It is nicely showing the staging of boilers where boiler 2 does not come on until about timestep 25, when the demand goes above 25 kW, and the loop cannot otherwise meet demand.  There are also some transients in the system due to the loop capacitance.  It is not visible on this plot, but boiler 2 does actually come on, momentarily, around timestep 18, as the loop demand goes slightly above 25 kW.  Also, the second boiler never hits full capacity because of the pump heat also providing some heat to the loop.  :+1: 
image

This fourth plot shows first the brief turn on of the second boiler due to transients in the system.  After that, it shows the boiler flow rate staging in action, where boiler 1 and 2 both operate at their full design flow when needed, but shut down completely (including the dedicated constant speed supply pump) when there is no demand on them. :+1: 
image

This final plot is confirmation of the flow balance on the loop, and is simply the common pipe flow rate.  During low demand times, there is common pipe flow, but when both boilers turn on, there is no common pipe flow. :+1: 
image


Myoldmopar:
Also, I should note that the fix currently in code here will need to be made much more robust if this is to go in.  At this point I'm just curious whether this will lead to many diffs in the common pipe files or not.

Myoldmopar:
Github is having trouble picking up that all these changes were already merged into develop on the BranchPumps branch that this branch was based on.  I may try closing it and reopening it or something to see if that clears the cached results.

Myoldmopar:
Github is having trouble picking up that all these changes were already merged into develop on the BranchPumps branch that this branch was based on.  I may try closing it and reopening it or something to see if that clears the cached results.

PR #6541, opened 2018-03-18
Improve setpoint calculation

lgu1234:
Pull request overview
---------------------
The improvement makes following steps to calculate setpoint:
When system is off, the setpoint is original until system is on. When the system is on in first time step, setpoint keeps the original. The setpoint is changes based on cutout temperature difference in the next time step until the zone condition is met.
### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@mjwitte will you have a chance to check this out today?  I was hoping to lock us down during the weekend.  As far as I know we only have this one, #6523 which should be ready now, and an incoming fix from @mbadams5.

mjwitte:
@lgu1234 I've tested SingleFamilyHouse_TwoSpeed_CutoutTemperatureI with this branch and current develop. The temperature pattern with this branch is much better. The system see that this control strategy does not work well for a heat pump, because the supplemental heat is coming on (even in April) to push the zone to the higher heating setpoint. For this to work correctly with a heat pump, the heat pump will need to know more information so that the supp heat does not kick in for this. Also, there are quite a few times where the zone air temperature overshoots the upper heating setpoint one timestep after the heating shuts off - doesn't seem possible. For example
image


I also tested with a modified version of HVACTemplate-5ZoneFurnace with fan operating mode set to cycling fan. The daytime heating setpoint is 23C and the cooling setpoint is 24C.
For some reason, with this system type the cooling setpoint is tracking the heating setpoint which is causing problems (requesting cooling on the winter day, and it fails with a fatal thermostat error). I don't see the same setpoint behavior in SingleFamilyHouse_TwoSpeed_CutoutTemperature. 

Then I changed the cooling setpoint to 27C and the setpoints were indepednent. So, apparently if the cooling and heating setpoint are too close to each other they stomp on each other?
 If it's possiible to detect and protect against it or throw a meaningful error message, that would be good here. New input files attached.

Also,  excess hours of thermostat setpoint not met are being counted with this method. That's another item that will need improvement.

In summary;
1. Please try to add an error message or fix the problem when the heating and cooling setpoints are too close together such that the cutoff differential causes overlap and apparent linking of the two setpoints.

2. Add to the Input Output Reference comments that 
    a. The heating and cooling setpoints must be separated by at least 2x the cutoff differential (I think that's right) or there will be an error.
    b. The thermostat setpoint not met hours are incorrect with this option.
    c. This option will cause excess operation of heat pump system supplemental heating coils.

3. Please post a followup message to address these limitations and to look at the overshoot problem.

mjwitte:
@lgu1234 Forgot to attach the furnace test files. 
[6541-FurnaceTestFiles.zip](https://github.com/NREL/EnergyPlus/files/1826813/6541-FurnaceTestFiles.zip)

lgu1234:
@mjwitte Thanks for testing the changes thoroughly. Here is my plan right now. 

1. I am going to add an error message when heating and cooling setpoint to close together. 
2. Revise IO Reference based on Mike's comments
3. After check-in, I am going to post a followup issue.

@Myoldmopar I am going to check-in. Hope you will have enough time to make preparation. Thanks for your patience.

lgu1234:
@mjwitte I update the branch to address your comments. Please take a final look.

At the same time, I posted a new issue: Address limitations and overshooting problem  #6546.

Thanks.

doc/input-output-reference/src/overview/group-zone-controls-thermostats.tex
```diff
@@ -117,6 +117,15 @@ \subsubsection{Inputs}\label{inputs-054}
...
 \end{lstlisting}
 
+Note that this option should be used in caution. Following recommendations may be used to input the value of Temperature Difference Between Cutout And Setpoint:
+
+\begin{lstlisting}
+1. The heating and cooling setpoints must be separated by at least 2 times the Temperature Difference Between Cutout And Setpoint or there will be a fatal error.
+2. The thermostat setpoint not met hours are incorrect with this option.
+3. This option will cause excess operation of heat pump system supplemental heating coils.
+
+\end{lstlisting}
```
Myoldmopar, line 127: Next time instead of a code `lstlisting` environment, use an `enumerate` environment.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3488,21 +3488,23 @@ namespace ZoneTempPredictorCorrector {
...
 					HeatOffFlag = true;
 				}
+				// check setpoint for both and provde an error message
+				if ( ZoneThermostatSetPointLo( ZoneNum ) >= ZoneThermostatSetPointHi( ZoneNum ) ) {
+					ShowSevereError( "DualSetPointWithDeadBand: When Temperature Difference Between Cutout And Setpoint is applied, the heating setpoint is greater than the cooling setpoint. " );
+					ShowContinueErrorTimeStamp( "occurs in Zone=" + Zone( ZoneNum ).Name );
+					ShowContinueError( "Zone Heating ThermostatSetPoint=" + RoundSigDigits( ZoneThermostatSetPointLo( ZoneNum ), 2 ) );
+					ShowContinueError( "Zone Cooling ThermostatSetPoint=" + RoundSigDigits( ZoneThermostatSetPointHi( ZoneNum ), 2 ) );
+					ShowFatalError( "Program terminates due to above conditions." );
+				}
```
Myoldmopar, line 3507: :+1:

Myoldmopar (approved):

Issue #6521, opened 2018-03-12
8.9 Release Items


mjwitte:
Issue overview
--------------
Checklist of things to finish before 8.9 release.

- [x] V8-9-0-Energy+.idd is not included in the IDFVersionUpdater folder in the release package.
- [x] Untransitioned v8.8 file can cause crash with empty error file (at lest when run in debug version). e.g. Files attached to #6508.
- [x] Also make sure package includes Rules8-7-0-to-8-8-0.md and OutputChanges8-7-0-to-8-8-0.md (IDD freeze package had old ones)
- [x] Add input value to select error messages
- [x] character encoding on windows in climatic data summary table output - garbage and causes missing rows
- [x] In the readme under Platforms "Windows 7, 8 & 8"
- [x] In the readme under Platforms, should we delete "(32-bit and RHEL-compatible versions may be available via Helpdesk)"?
- [x] In the readme under Known Problems, the link to issues is getting modified to include the install path (similar to the way the change log linke is correctly modified):
`file:///C:/EnergyPlusV8-9-0-RC0/github.com/NREL/EnergyPlus/issues`
- [x] Turn back on the duplicate item error and fix unit tests that failed.
- [x] the link in readme to "this EnergyPlus website for third party interfaces" which points to https://energyplus.net/interfaces is shown as access denied
- [x] the link in readme to "this EnergyPlus website for third party IDF creation tools." which points to https://energyplus.net/input-files is shown as access denied
- [x] LBuildingAppGRotPar on Win10 . . .
![image](https://user-images.githubusercontent.com/4991075/37773002-66ef5f6a-2daa-11e8-8dda-3e728f0c7389.png)

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
@mjwitte I didn't know about the untransitioned file error.  Do you know whether this is JSON related or something I can fix?  @mbadams5 is still polishing up a branch with the json follow-up changes.

mjwitte:
[...]

PR #6564, opened 2018-03-22
Final approach on 8.9.0

Myoldmopar:
Tasks from https://github.com/NREL/EnergyPlus/issues/6521 plus anything else we find.


Myoldmopar:
Expecting a few unit tests (~10) that need to be remedied (removing duplicate names) to get this ready to go.

mjwitte:
@mbadams5 @Myoldmopar I can't find where to fix the encoding settings to fix degree symbols in the Climatic Data Summary output table on windows. Where do I look to fix that? [This line](https://github.com/NREL/EnergyPlus/blame/f75dd060e840227c6eb4411df72be1c679fc44c3/src/EnergyPlus/OutputReportTabular.cc#L5405) was touched by the original input refactor (along with others that have special characters). I've tried pasting this in from v8.8 source to get a good degree symbol in the output table but it's not working in VS on Windows. What's controlling this?

Myoldmopar:
I'll have to investigate a little further to find out.  I am currently trying to figure out why Linux builds keep stalling on the RC1 package build (it built the tar gz, but it won't build the sh installer...)

Myoldmopar:
So it looks like CI wasn't having trouble with the build of the package.  I kept checking the dashboard and it kept showing "pending", so I assumed CI had failed.  So I tried building it locally, and it built fine.  I decided to just upload my manually built version of the package to the release.  When I edited the release, I saw that CI had attempted to upload that package already, but Github had an error.  So I think the CI engine got stuck after Github failed to allow the upload.  I have restarted the CI box.  It will hopefully be OK.  I am uploading my manually built package.  Carrying on...

jasondegraw:
@mjwitte @Myoldmopar I've pushed a fix for the exe placement that I've tested locally.

mjwitte:
@Myoldmopar @JasonGlazer @mbadams5 I've changed epl-run.bat to always produce the in.epJSON from in.idf and save it as \<filename\>.epJSONout to avoid confusion with a real \<filename\>.epJSON intended as input. So, this means every run on Windows, using EP-Launch, and using the batch file will produce \<filename\>.epJSONout mostly for informational purposes. If anyone objects, then we can revert this and make the alternate batch file available on the helpdesk.

mjwitte:
Oh and @mbadams5 The documentation for command line arguments is automated using `energyplus --help` so the new `-c` is already listed. For reference, this is noted in the [Auxiliary Programs doc](https://github.com/NREL/EnergyPlus/blob/f75dd060e840227c6eb4411df72be1c679fc44c3/doc/auxiliary-programs/src/technical-details-of-running-energyplus/running-energyplus-by-hand.tex#L3).
```
[D:\EnergyPlusV8-9-0-RC1]energyplus.exe --help
EnergyPlus, Version 8.9.0-f75dd060e8
Usage: energyplus [options] [input-file]
Options:
  -a, --annual                 Force annual simulation
  -c, --convert                Output IDF->epJSON or epJSON->IDF, dependent on
                               input file type
  -d, --output-directory ARG   Output directory path (default: current
                               directory)
  -D, --design-day             Force design-day-only simulation
  -h, --help                   Display help information
  -i, --idd ARG                Input data dictionary path (default: Energy+.idd
                               in executable directory)
  -m, --epmacro                Run EPMacro prior to simulation
  -p, --output-prefix ARG      Prefix for output file names (default: eplus)
  -r, --readvars               Run ReadVarsESO after simulation
  -s, --output-suffix ARG      Suffix style for output file names (default: L)
                                  L: Legacy (e.g., eplustbl.csv)
                                  C: Capital (e.g., eplusTable.csv)
                                  D: Dash (e.g., eplus-table.csv)
  -v, --version                Display version information
  -w, --weather ARG            Weather file path (default: in.epw in current
                               directory)
  -x, --expandobjects          Run ExpandObjects prior to simulation
Example: energyplus -w weather.epw -r input.idf

[D:\EnergyPlusV8-9-0-RC1]
```

Myoldmopar:
We should be getting close, datasets will now be done, I think there are a couple others I need to do, but closer...

mjwitte:
@Myoldmopar Working on the failing unit tests

mjwitte:
Go Ramblers!

Myoldmopar:
> Go Ramblers!

Woo! Awesome!

Myoldmopar:
Great, thanks @mjwitte did that last commit fix those two final things?  I fixed up the readme now, so the only **known** issue is the bad character.  I don't have any immediate clues.  Did @mbadams5 have any clues for tracking it down?

Myoldmopar:
@mjwitte so the HeatBalanceManager tests were failing on my machine because of the extra line ending.  Did you **have** to add the `\r\n` to the end to get it to pass?  I would've thought it would work with just the `\n` on all platforms and let C++ work out the different platform meanings of that string.  I just pushed up a change where I only do the `\n`.  If it fails on CI, or for you locally, we can figure out the best path.

mjwitte:
@Myoldmopar No that didn't work here. Proper fix - I hope - pushing up now.

Myoldmopar:
You are welcome to just comment out that error stream comparison.  We can keep the input check and make sure it throws an error.  That would be sufficient for that unit test.  Then we don't have any problems with line endings here.

mjwitte:
Will do if this one fails

Myoldmopar:
That works on my machine!  Great fix!

mjwitte:
Monkey see monkey do  ee-ee-ee-aa-aa-aa

Myoldmopar:
Looks to be all clean.  If that encoding issue appears fixed, I'm testing, merging, tagging, and moving on.

mjwitte:
Checking now on Win 8.1

Myoldmopar:
Thanks @mjwitte I'm building packages on Mac and Linux and will test a few things out.

mjwitte:
Looks good - Climatic Data Summary is back to where it was in v8.8 and Energy+.schema.epJSON is present in the root install folder.

Myoldmopar:
Tested package on Linux, played with lots of binaries, everything seems OK.  Just opened an issue for the parametric preprocessor, but it's not a change for now.

Tested package on Mac, ran some simulations with ep-launch-lite successfully.  Install folder looks fine.

We are a wrap on 8.9.0.

Issue #6574, opened 2018-03-23
Problems with Airloop operation with Zone Exhaust Fan and Cooled Beam


EnergyArchmage:
Issue overview
--------------
The AirloopHVAC will not operate when a ZoneHVAC:ExhaustFan is added.  There seems to be serious problems with zone return air node flow calculations when a zone exhaust fan is added to a zone served by AirTerminal:SingleDuct:ConstantVolume:CooledBeam.  The central fan does not operate and the air system provides no ventilation air to any of zones.  

Defect file adds one zone exhaust fan to example file 5ZoneCooledBeam.idf.  Problem is that after adding the zone exhaust the airloop central fan does not operate. 

After much debugging, the problem appears to be with MassFlowRateZSup term here https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L4487

Unclear why zero return air flow is propagating past the OA system and shutting down the air handler.   

### Workaround ###
Make the zone exhaust fan "balanced".

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (\EnergyPlusDevSupport\DefectFiles\6000s\6574\6574_5ZoneCoolBeam.idf)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@lgu1234 There is an initial fix in #6575 but I haven't gone back to confirm and add a unit test. Are you planning to complete this?

eringold:
This might be related to https://github.com/NREL/EnergyPlus/issues/6492

Assuming "ZoneHVAC:ExhaustFan" is Fan:ZoneExhaust.

lgu1234:
@mjwitte I took a look at #6574 and have not started it formally. Since you have an initial fix, it is better for you to complete it. Thanks.

Issue #6492, opened 2018-02-22
Unexpected interaction between Fan:ZoneExhaust and Controller:OutdoorAir


eringold:
Found in v8.8 (made with OpenStudio v2.4.0)

I'm not entirely certain this is a bug, if not it's just really unexpected and confusing behavior.

I have a VAV with Reheat air loop serving multiple zones, where one zone also has a Fan:ZoneExhaust which is Decoupled from the System Availability Manager. I’ve set the Air Loop to follow an operation schedule, but cycle on at night to meet any heating/cooling loads. Everything is autosized, and the terminals set to control for outside air. I also want to shut down outside air at night, so I’ve set a schedule (same as the loop operation schedule) to the Controller:Outdoor Air ‘Maximum Fraction of Outside Air Schedule Name’.

Here’s the problem: If the Fan:ZoneExhaust attached to one of the zones does NOT have a ‘Flow Fraction Schedule Name’ schedule set, the entire air system will shut off during the outdoor air shutoff hours, even if the attached systems are showing a heating load:

![image](https://user-images.githubusercontent.com/10799524/36514612-7cc6912a-1732-11e8-8e9d-df32f1bf64f5.png)

If I DO set the exhaust fan ‘‘Flow Fraction Schedule Name’ to a schedule, which matches the system operation schedule, I get the correct air system operation: outdoor air off at night, but system cycles to meet heating loads:

![image](https://user-images.githubusercontent.com/10799524/36514628-904a0e7a-1732-11e8-8360-18ef172e9359.png)

Extra: if the exhaust fan operation schedule is different from the system operation/outdoor air schedule, say turns on at 4AM instead of 8, I get an air system which shuts down from 4AM to 8AM (the exhaust fan ON before system ON hours), then works normally the rest of the day:

![image](https://user-images.githubusercontent.com/10799524/36514673-c49c744c-1732-11e8-9aa7-51607688a7da.png)

@mjwitte I think this might have to do with the changes in https://github.com/NREL/EnergyPlus/pull/5794 (since the interaction seems to disappear when I don't define a schedule in Maximum Fraction of Outdoor Air Schedule of the Controller:OutdoorAir), but that's just a guess.

I can provide an example file if necessary.

mjwitte:
@eringold Please upload example(s) here or e-mail me.

eringold:
@mjwitte 
idf with no exhaust fan schedule:
[...]

PR #6575, opened 2018-03-23
Chilled beam airloop flow balance - maybe bigger than chilled beams

mjwitte:
Pull request overview
---------------------
Attempting to address #6574 .
This may also address #6492 (see [comment here](https://github.com/NREL/EnergyPlus/issues/6574#issuecomment-407796677))

@EnergyArchmage Noted that `MassFlowRateZSup` is zero [here](https://github.com/NREL/EnergyPlus/blob/40101eaafd1525ee63406264ae9535f25acf4805/src/EnergyPlus/ZoneEquipmentManager.cc#L4487) in `CalcZoneReturnFlows`.

`MassFlowRateZSup` gets set [here](https://github.com/NREL/EnergyPlus/blob/40101eaafd1525ee63406264ae9535f25acf4805/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L714-L742) in `SimZoneAirLoopEquipment`, but this code was never executing for this chilled beam (and for any other chilled beam for that matter), because `InNodeNum` = 0.

`InNodeNum` comes from `AirDistUnit( AirDistUnitNum ).InletNodeNum` which was never initialized for both flavors of chilled beams. That's been fixed here, but the system flow is still zero.

From what I see in the debugger at this point is that the airloop `MassFlowRateSetPoint` keeps ratcheting down the flow until the system ends up at zero. Wondering about the logic [here](https://github.com/NREL/EnergyPlus/blob/40101eaafd1525ee63406264ae9535f25acf4805/src/EnergyPlus/SimAirServingZones.cc#L2141-L2143) in `InitAirLoops` - but that code has been around a long time.

Making the exhaust fan "balanced" results in system flow, but that shouldn't be necessary. Need to investigate this more.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2:

nrel-bot-3:

nrel-bot-2:

nrel-bot-2:
@mjwitte @lgentile it has been 34 days since this pull request was last updated.

nrel-bot-3:

nrel-bot-2:

rraustad:
Fix discussed via email corrects #6492.

image

rraustad:
I added my code changes here so others could review. I did not address conflicting files at this time.

rraustad:
The problem with this change is that the mixed air mass flow rate is artificially increased to account for the missing exhaust air (and happens to be the flow rate that the TUs want to operate at). That missing air mass flow is magically added to the recirc air stream when it should actually come from OA. So my thought is to add the flow discrepancy on top of the OA amount to maintain neutral building pressure. As @lgu1234 says, what happens if there is no outdoor air system? Or what if there is an OA system and the user has made up the exhaust with infiltration or ventilation? Still need to think through all of this.

Myoldmopar:
@rraustad or @mjwitte please merge develop into this branch before you make any more pushes.  I am getting branches updated to use the new Ubuntu 18.04 CI specs, but I didn't take this one on because it has conflicts.

nrel-bot-2:

rraustad:
Very strange. I reverted all changes and kept just the one line of code that seemed most likely as the fix and results look reasonable.

fixzoneexhaust

Then I merged develop and got this.

fixzoneexhaust_mergeddevelop

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -6116,7 +6116,7 @@ namespace InternalHeatGains {
         Real64 OperSchedFrac;                             // Operating schedule fraction
         Real64 CPULoadSchedFrac;                          // CPU loading schedule fraction
         Real64 AirConnection;                             // Air connection type
-        Real64 TSupply;                                   // Supply air temperature [C]
+        Real64 TSupply(0.0);                              // Supply air temperature [C]
```
rraustad, line 6119: Part of hack for #7041

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -6313,7 +6313,7 @@ namespace InternalHeatGains {
                 TAirOut = TAirIn;
             }
 
-            if (abs(TAirOut - TSupply) < SmallTempDiff) {
+            if (SupplyNodeNum != 0 && abs(TAirOut - TSupply) < SmallTempDiff) {
```
rraustad, line 6316: Rest of hack for #7041

src/EnergyPlus/MixedAir.cc
```diff
@@ -3758,8 +3758,8 @@ namespace MixedAir {
 
         } // if (AirLoopNum > 0)
 
-        // Set the relief air flow rate (must be done last to account for changes in OAMassFlow
-        this->RelMassFlow = max(this->OAMassFlow - this->ExhMassFlow, 0.0);
+          // Set the relief air flow rate (must be done last to account for changes in OAMassFlow
+        this->RelMassFlow = this->OAMassFlow - this->ExhMassFlow;
```
rraustad, line 3762: Big question here is if relief air can be negative. Not sure where this limitation came from. Seems reasonable to expect no intake into relief node but this issue raises questions.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2281,7 +2281,7 @@ namespace SimAirServingZones {
                 // Need to make sure that flows are greater than zero
                 if (MassFlowSet >= 0.0) {
                     Node(NodeNumOut).MassFlowRateSetPoint = MassFlowSet;
-                } else if (MassFlowSet < 0.0) {
+                } else {
```
rraustad, line 2284: excess logic, if mass flow rate is not >= it must be <

rraustad:
@mjwitte passing this over to you. Enjoy.

nrel-bot-3:

mjwitte:
Replaced with #7059

PR #6583, opened 2018-03-24
Add curve unit tests stylesapplied

lefticus:
This is a test PR to make sure the different branches with styles applied are able to merge cleanly.

Re @Myoldmopar and @jasondegraw if you have any comments on this.


Myoldmopar:
I'm assuming you aren't actually looking for comments yet though, right?  I see the build failing, probably for the same reason your develop-with-style-applied branch is failing.  I'll wait until this is passing before making any comments unless you say otherwise.

lefticus:
@Myoldmopar comments would be fine right now, although I also just saw that the build failed - that was unexpected to me. The question is if the formatting is a go or if you have any concerns otherwise

lefticus:
FYI, see the build errors - it's due to header include order mattering (because some .cc file was including some other header that a subsequent header relied on) and clang-format is alphabetizing the headers

jasondegraw:
Oops, that's not great. We should fix that.

lefticus:
In process right now. I'll have it all sorted out and building soon. I have
the styles applied to all branches already, will just need to propagate
this fix.

lefticus:
@jasondegraw @Myoldmopar I'm almost 100% ready to go, except for one question:

I've limited all of this to the `src` directory, as specified, but do we want to do the `tst` folder as well?

Myoldmopar:
@lefticus so was this actually intended to be merged at some point?  Or just a test to verify the diffs look proper?

Myoldmopar:
Deleting the branch now.

Issue #6599, opened 2018-03-28
HVACAirloop:UnitarySystem does not honor continuous fan mode using multispeed heating coil


rraustad:
Issue overview
--------------
HelpDesk ticket 12594 shows incorrect fan air flow using constant fan operating mode. Fan should operate continuously at low speed during no load condition. Constant fan mode is scheduled between 6 AM and 10 PM, otherwise cycling fan.

![12594_multispeedheatcoilcontfan](https://user-images.githubusercontent.com/5778156/38001195-83e497b8-31f9-11e8-9ef7-9edf9362b7d3.jpg)

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Preliminary changes show what appears to be correct fan operation. Full flow when load exists, 25% of flow when no load exists.

![12594_multispeedheatcoilcontfan_fixed](https://user-images.githubusercontent.com/5778156/38003404-674823fc-3205-11e8-8956-d508521e5f6a.jpg)

rraustad:
See PR #6600.

mjwitte:
Closed via #6600

PR #6600, opened 2018-03-28
AirloopHVAC:UnitarySystem with constant fan and multispeed gas/elec heating coil does not honor no load flow rate.

rraustad:
Pull request overview
---------------------
1. The logic used to set the no load air flow rate did not account for when there was not a multi-speed cooling coil upstream of the heating coil. This corrects Issue #6599.

2. HVACTemplate:System:UnitarySystem for multi-speed coils, previously left the "No Load Supply Air Flow Rate Ratio" blank in `UnitarySystemPerformance:Multispeed`. 
Now it writes `1 / max( "Number of Speeds for Heating", "Number of Speeds for Cooling" )`

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


rraustad:
Interesting, this line in a unit test is failing on clang.

PackagedTerminalHP_VSCoils_Sizing:

    // the fan vol flow rate should equal the max of cooling and heating coil flow rates
    EXPECT_EQ(Fan(1).MaxAirFlowRate,
              max(VariableSpeedCoils::VarSpeedCoil(1).RatedAirVolFlowRate, VariableSpeedCoils::VarSpeedCoil(2).RatedAirVolFlowRate));

Fan(1).MaxAirFlowRate = 6.61564
VarSpeedCoil(1).RatedAirVolFlowRate = 6.61564
VarSpeedCoil(2).RatedAirVolFlowRate = 5.09533

The failure is that 6.61564 != 5.09533 ? I don't see anything wrong with this line in the unit test.

mjwitte:
I've found sometimes I need to add a line above the `EXPECT` to compute and store the comparison value and give a simple argument to `EXPECT`.

rraustad:
Jeeez, do I finally have if right? This issue of what fan flow rate to use for constant fan mode for multi-speed coils when no load existed was killin me. Not sure I have it exactly right but at least now everything passes.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2392,6 +2392,18 @@ namespace HVACUnitarySystem {
...
+        if (MultiOrVarSpeedCoolCoil(UnitarySysNum) || MultiOrVarSpeedHeatCoil(UnitarySysNum)) {
+            if (UnitarySystem(UnitarySysNum).DesignFanVolFlowRate > 0.0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).DesignFanVolFlowRate;
+            } else {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = max(UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow, UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow);
+            }
+            UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).IdleVolumeAirRate * StdRhoAir;
+            UnitarySystem(UnitarySysNum).IdleSpeedRatio = 1.0;
+        }
+
```
rraustad, line 2605: OK, so why do this? If there is no cooling coil then the idle flow rate won't be set and will likely be 0. If a multi-speed heating coil is present then the idle flow rate will be 0 when it is time to take the min of 0 and the heating flow rate. See heating coil idle flow rate set below.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2423,25 +2435,25 @@ namespace HVACUnitarySystem {
                                                      UnitarySystem(UnitarySysNum).CoolingCoilName, ErrFound);
             EqSizing.DesCoolingLoad = DXCoolCap;
 
-            for (Iter = 1; Iter <= UnitarySystem(UnitarySysNum).NumOfSpeedCooling; ++Iter) {
+            for (Iter = UnitarySystem(UnitarySysNum).NumOfSpeedCooling; Iter >= 1;
+                --Iter) { // use reverse order since we divide by CoolVolumeFlowRate(max)
```
rraustad: Gost to go backwards since it's a divide by UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(Iter)

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2423,25 +2435,25 @@ namespace HVACUnitarySystem {
...
-            UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1);
+            if (MSHPIndex > 0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).CoolMassFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+            } else if (allocated(UnitarySystem(UnitarySysNum).CoolVolumeFlowRate)) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(1);
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).CoolMassFlowRate(1);
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1);
+            }
```
rraustad: This one is really hard. If there is no DSMultiSpeed object then the no load flow rate is set by the lowest flow rate.  If there is a DSMS object then the no load flow rate is set by the user input.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2563,70 +2616,50 @@ namespace HVACUnitarySystem {
...
                     UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(1);
                     UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).HeatMassFlowRate(1);
                     UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(1);
                 }
-            } else {
+            } else if (MSHPIndex > 0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).HeatMassFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+            } else if (allocated(UnitarySystem(UnitarySysNum).HeatVolumeFlowRate)){
```
rraustad: OK, now it gets harder here. There might not be a cooling coil, or it might not be a multi-speed coil. If it's not a multi-speed cooling coil then idle flow rate should not be set (except that I set if initially above). So if a cooling coil exists and it's multispeed then set the idle flow rate based on the min of both. If a cooling coil exists and it's not a multispeed then set the idle flow rate based on the heating coil. If there is no cooling coil then set based on heating coil. And what if there isn't a DSMSHP object? So seeeee, it gets complicated.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2563,70 +2616,50 @@ namespace HVACUnitarySystem {
...
                 UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(1);
                 UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).HeatMassFlowRate(1);
                 UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(1);
             }
         }
 
-        if (UnitarySystem(UnitarySysNum).CoolingCoilType_Num == Coil_CoolingWater ||
-            UnitarySystem(UnitarySysNum).CoolingCoilType_Num == Coil_CoolingWaterDetailed) {
```
rraustad, line 2779: Why the heck was these cooling coils checked with the heating coils? I moved this up to where the cooling coils are tested. Do cooling first, then heating.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2652,9 +2685,26 @@ namespace HVACUnitarySystem {
                         UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(Iter) /
                         UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(DesignSpecMSHP(MSHPIndex).NumOfSpeedHeating);
                 }
-                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(1);
-                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).HeatMassFlowRate(1);
-                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(1);
+            }
+            if (UnitarySystem(UnitarySysNum).CoolCoilExists) {
```
rraustad, line 2953: This section can be critiqued. If a cooling coil exists, then do this. If it is multispeed then do this. If not multispeed then do this. Maybe it should be if MultiOrVarSpeedCoolCoil, that would make more sense.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -2388,6 +2582,10 @@ TEST_F(EnergyPlusFixture, UnitarySystemSizingTest_ConfirmUnitarySystemSizingTest
     DataEnvironment::StdRhoAir = 1000; // Prevent divide by zero in ReportSizingManager
 
     UnitarySystem.allocate(HVACUnitarySystem::NumUnitarySystem);
+    MultiOrVarSpeedCoolCoil.allocate(HVACUnitarySystem::NumUnitarySystem);
+    MultiOrVarSpeedCoolCoil = false;
+    MultiOrVarSpeedHeatCoil.allocate(HVACUnitarySystem::NumUnitarySystem);
+    MultiOrVarSpeedHeatCoil = false;
```
rraustad, line 2591: The code now only initializes the idle flow rate (based on these variables) if it is a multi-speed coil so this variable must be allocated in unit tests if GetUnitarySystemInput is not called.

tst/EnergyPlus/unit/PackagedTerminalHeatPump.unit.cc
```diff
@@ -524,8 +524,8 @@ TEST_F(EnergyPlusFixture, PackagedTerminalHP_VSCoils_Sizing)
 
     SizeFan(1);
     // the fan vol flow rate should equal the max of cooling and heating coil flow rates
-    EXPECT_EQ(Fan(1).MaxAirFlowRate,
-              max(VariableSpeedCoils::VarSpeedCoil(1).RatedAirVolFlowRate, VariableSpeedCoils::VarSpeedCoil(2).RatedAirVolFlowRate));
+    Real64 maxCoilAirFlow = max(VariableSpeedCoils::VarSpeedCoil( 1 ).RatedAirVolFlowRate, VariableSpeedCoils::VarSpeedCoil( 2 ).RatedAirVolFlowRate);
+    EXPECT_EQ(Fan(1).MaxAirFlowRate, maxCoilAirFlow);
```
rraustad, line 528: clang fix.

rraustad:
Comparing defect file fan operation to the issue description of #6599 shows expected result when idle air flow rate is non-zero.

northwestzoneheatingcoil

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2530,18 +2572,29 @@ namespace HVACUnitarySystem {
                         UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(Iter) /
                         UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(DesignSpecMSHP(MSHPIndex).NumOfSpeedHeating);
                 }
-                // these coil types do not have an idle speed air flow rate
-                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = 0.0;
-                UnitarySystem(UnitarySysNum).IdleMassFlowRate = 0.0;
-                UnitarySystem(UnitarySysNum).IdleSpeedRatio = 0.0;
```
rraustad, line 2704: These few lines previously set idle air flow rate to 0 for these coils types. Although these coil types do not have inputs for no load flow rate, the unitary system does, or the DSMSHP object does so this is wrong when constant fan operating mode is used.

rraustad:
This is complete and ready for review.

rraustad:
Where is the .clang_format file so that I can see what some of these warnings are referring to. I don't think I changed some of these lines.

rraustad:
Well I found the .clang_format file but it's fairly cryptic and didn't help much. What is wrong with lines 6731-6738 of HVACUnitarySystem.cc ? I didn't change those lines, I don't see any tabs, they aren't longer than 150 characters, so then what is the format issue? Indentation?

Myoldmopar:
@rraustad I just pushed up the code with the style fixed.  You can take a look at the diff and see what I had to change.  And by "I had to change" I mean I went: `$ clang-format -i HVACUnitarySystem.cc`

rraustad:
Thanks for that, however, there are many changes there that weren't there before. Even more than the warnings indicated. Shouldn't the format conversion commit already have made all those changes? Won't there be many more of these type of warnings in the future?

mjwitte:
@rraustad See https://github.com/NREL/EnergyPlus/wiki/Coding-Guidelines#formatting-style You have to add a plugin to VS to get it to apply the rules automatically while editing.

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

nrel-bot-2:

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2550,6 +2550,20 @@ namespace HVACUnitarySystem {
                 UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow / UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow;
         }
 
+        // initialize idle air flow rate variables in case these are needed in multi-speed heating coil sizing when multi-speed cooling coil does not
+        // exist
+        // the multi-speed coils will overwrite this data and these variablse are only used for multi-speed coils in function SetOnOffMassFlowRate
+        if (MultiOrVarSpeedCoolCoil(UnitarySysNum) || MultiOrVarSpeedHeatCoil(UnitarySysNum)) {
```
mjwitte: Minor - type here "variablse" - fix only if there are other changes needed.

src/EnergyPlus/HVACUnitarySystem.cc
mjwitte: Shouldn't this default section use the no load flow rate?

src/EnergyPlus/HVACUnitarySystem.cc
mjwitte: Confused. Both if blocks above use `CoolVolumeFlowRate(1)` , but only the else if checks if CoolVolumeFlowRate is allocated?  And again, why not use the no load air flow rate when there is no DSMS?

mjwitte (commented):
I've made a first pass through the code changes. It all seems much more complicated than it needs to be, but I don't have a good big picture understanding of this yet.

Running the defect file, I see how the flow was going to zero before, and now it doesn't.  But the flow when the coil is off is not what I expected.  Full flow is 1.29799, no-load flow is 0.0811.  I expected the no-load flow to be 25% of 1.29799, or 0.3245.

Note I merged in develop to bring this branch up to date.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: Good point. This is a carry over from the furnace. But your argument is a good one. Not sure I would want to change that here.

    if (Furnace(FurnaceNum).NumOfSpeedHeating > 0) {
        Furnace(FurnaceNum).IdleMassFlowRate = min(Furnace(FurnaceNum).HeatMassFlowRate(1), Furnace(FurnaceNum).CoolMassFlowRate(1));
        Furnace(FurnaceNum).IdleSpeedRatio = min(Furnace(FurnaceNum).MSHeatingSpeedRatio(1), Furnace(FurnaceNum).MSCoolingSpeedRatio(1));
        Furnace(FurnaceNum).IdleVolumeAirRate = min(Furnace(FurnaceNum).HeatVolumeFlowRate(1), Furnace(FurnaceNum).CoolVolumeFlowRate(1));
    } else {
        Furnace(FurnaceNum).IdleMassFlowRate = Furnace(FurnaceNum).CoolMassFlowRate(1);
        Furnace(FurnaceNum).IdleSpeedRatio = Furnace(FurnaceNum).MSCoolingSpeedRatio(1);
        Furnace(FurnaceNum).IdleVolumeAirRate = Furnace(FurnaceNum).CoolVolumeFlowRate(1);
    }

These are allocated up at 2571 inside this VS coil IF block so this is overprotective. It should be just if MSHPIndex > 0 ELSE.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I chose the higher flow rate because of the min(heatingflow, coolingflow) calcs below and I didn't want this default to collide with "current" logic (i.e., idle flow is not based on no load flow). Not saying I'm right, your wrong, just saying that's what I did. I think you have a good point. Only VS and MS coils use that variable, but when implemented we already had a no load flow rate variable. I can't stand using 2 different variables based on coil type. I recall messing around with idle flow before and got diff's. Maybe a new issue to just focus on idle flow rate usage?

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: If there are no hold-ups I'll fix this typo in another branch. Otherwise I'll fix if we agree something in this branch needs changing.

mjwitte: See above, the results look wrong here.

rraustad: Sorry, forgot to answer that one. I thought they were wrong too when I fixed this (because I used 0.25 * 1.298 and something happened, can't recall now what that was). Then I read the IO and found it's a fraction of the low speed flow rate. So this means the no load flow rate could never be greater than the lowest speed. Not sure anyone would ever want that but it does limit flexibility. So it's 0.25 * 0.3245 = 0.081125. So yes, the results look wrong, even after this explanation, but they are correct.

mjwitte: That's not how I read the I/O Ref:
> The no load air flow rate will be calculated as this fraction multiplied by the minimum of the cooling and heating supply air flow rate.

To me, that means it's the fraction of either Heating Supply Air Flow Rate or Cooling Supply Air Flow Rate from the AirloopHVAC:UnitarySystem object, whichever is smaller.

rraustad: OK, maybe I read that too quickly. But I had it as a fraction of the max (i.e., the min of heating and cooling max) before and something happened. I think it was that I tried autosize and it gave me a no load flow rate equal to the max. So the autosized multispeed coil operated at full flow when there was no load yet operated at lower flows when there was a load. That just seemed wrong since most files are autosized. Confirm you want me to change this before I go to that trouble (of making sure it's the same everywhere).

mjwitte: Well, the default fraction is 1.0 which is probably not the best choice, but it's been that way. 
```
UnitarySystemPerformance:Multispeed,
  N3 , \field No Load Supply Air Flow Rate Ratio
       \type real
       \minimum 0
       \maximum 1
       \default 1
       \note Used to define the no load operating air flow rate when the system fan
       \note is specified to operate continuously.
```
I'm not following your comments. It shouldn't matter if the coil is autosized or not, this is at the unitary system level. Now, if there are example files that have a blank in this field or a 1.0, then, perhaps those should be change, but I don't know if that's relevant here.

As it stands now, the defect file is autosizing the no load flow rate to 0.32450 (25% of full flow) as expected.  It's just not using that during operation.

rraustad: OK so I'll change it, and the typo. I agree with your assessment. That's the only change, right?

mjwitte: I think so. There is a strange test failure [here](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-e17266cba39a9a4fe68cfcb5ff574472db47fd3e-x86_64-Linux-Ubuntu-14.04-gcc-4.8-IntegrationCoverage-Debug.html) in integration.OutdoorAirUnitwithAirloopHVAC but hopefully that goes away when the missing line fix is merged in.  Waiting for CI to finish that and then I'll merge it.

rraustad: I can't think of a way to default this logically. I could make it the min of min speed flows or make it a required field (don't like that idea). I just don't like the way a blank or autosized field is handled.

mjwitte: I'm not following. Are you talking about No Load Supply Air Flow Rate Ratio? Just let it be 1.0. It's easy to set to something else.

rraustad:
I made the changes and think this is better. I tried to test 2 speed heating and got a crash in UpdateDataandReport. So on goes the fixing. I'll post a new issue for this one.

rraustad:
Created #6851 for 2 speed gas heating coil.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2698,18 +2771,43 @@ namespace HVACUnitarySystem {
...
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate =
+                    UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate =
+                    UnitarySystem(UnitarySysNum).HeatMassFlowRate(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio =
+                    UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+            } else {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirMassFlow;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow / UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds);
```
rraustad: This looks like it's a cut/paste error since we are in the heating coil section. Now I think what IdleSpeedRatio means. It's a scalar, not dependent on heating or cooling. So this should probably be nocoolheatflow to fan flow? to max flow? This is all very convoluted.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2694,32 +2709,38 @@ namespace HVACUnitarySystem {
...
                 UnitarySystem(UnitarySysNum).IdleMassFlowRate =
-                    UnitarySystem(UnitarySysNum).CoolMassFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                    UnitarySystem(UnitarySysNum).CoolMassFlowRate(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
                 UnitarySystem(UnitarySysNum).IdleSpeedRatio =
-                    UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                    UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+            } else {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirMassFlow;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow / UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds);
```
mjwitte: 2 things here.  `NumSpeeds` could be zero at this point (same at line 2682 above) - might be what's causing the segfault in the CI tests.  Second, while `UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds)` will typically be the full cooling flow (because we expect the top speed to have a flow fraction of 1.0) , it won't necessarily be that.  The denominator here should be whatever IdleSpeedRatio gets multiplied by later - is it the cooling or heating flow or the max or min of these?

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: In Init this is done:

    if (MyFanFlag(UnitarySysNum)) {
        if (UnitarySystem(UnitarySysNum).ActualFanVolFlowRate != AutoSize) {
            if (UnitarySystem(UnitarySysNum).ActualFanVolFlowRate > 0.0) {
                UnitarySystem(UnitarySysNum).HeatingFanSpeedRatio =
                        UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;
                UnitarySystem(UnitarySysNum).CoolingFanSpeedRatio =
                        UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;
                UnitarySystem(UnitarySysNum).NoHeatCoolSpeedRatio =
                        UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;

So why not initialize these speed ratio variables here?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2662,22 +2671,28 @@ namespace HVACUnitarySystem {
                         UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(DesignSpecMSHP(MSHPIndex).NumOfSpeedCooling);
                 }
                 UnitarySystem(UnitarySysNum).IdleVolumeAirRate =
-                    UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(1) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                    UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
```
mjwitte: This (and below) should not be using `UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds)`. The I/O Ref says the no-load fraction will be multiplied by the "the minimum of the cooling and heating supply air flow rate." so that would be `min(UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow,UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow)`

mjwitte: Or as you just suggested, for noload this could all be set earlier using MaxNoCoolHeatAirVolFlow which is already set correctly(?) in the sizing routine.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: The speed ratio is passed to the OnOff fan model. So it has to be the ratio of flow to max flow of this operating mode. Pretty sure I added speed ratio a long time ago. 

    //      adjust RTF to be in line with speed ratio (i.e., MaxAirMassFlowRate is not MAX when SpeedRatio /= 1)
    //      PLR = Mdot/MAXFlow => Mdot/(MAXFlow * SpeedRatio), RTF = PLR/PLF => PLR/SpeedRatio/PLF = RTF / SpeedRatio
    if (SpeedRatio > 0.0) Fan(FanNum).FanRuntimeFraction = min(1.0, Fan(FanNum).FanRuntimeFraction / SpeedRatio);
    SpeedRaisedToPower = CurveValue(Fan(FanNum).FanPowerRatAtSpeedRatCurveIndex, SpeedRatio);
    Fan(FanNum).FanPower *= SpeedRaisedToPower / EffRatioAtSpeedRatio;

FanSystemModel also uses this:

    if (present(flowFraction)) {
        localFlowFraction = flowFraction;
        localAirMassFlow[0] = localFlowFraction * m_maxAirMassFlowRate;
    } else {
        if (m_maxAirMassFlowRate > 0.0) { // protect div by 0
            localFlowFraction = m_inletAirMassFlowRate / m_maxAirMassFlowRate;
        } else {
            localFlowFraction = 1.0;
        }
        localAirMassFlow[0] = m_inletAirMassFlowRate;
    }

So it looks to me as though speed ratio is always with respect to the max fan flow. Not sure speed ratio is even needed outside the fan model. Just let the fan calculate that value? Or does a 2 speed fan need the ratio wrt to max flow at each speed (which would mean the parent needs to handle)?

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I think this could be simplified but won't do that here. The cooling coils are set first, so here it's just based on cooling in case there is no heating coil. Then the heating coils are set later, and there it's based on the min.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I'm adding this in InitLoadBasedControl to get this right and should overwrite what is done during sizing. The speed ratio should be relative to max fan flow but I may not know the fan flow during sizing. I'll have to think more about what to do with set point based control, since that is constant fan, but it seems to me these should be initialized the same way regardless of control type. Once I know the fan flow rate I'll initialized these variables:

    if (MultiOrVarSpeedHeatCoil(UnitarySysNum) || MultiOrVarSpeedCoolCoil(UnitarySysNum)) {
        if (MultiOrVarSpeedCoolCoil(UnitarySysNum)) {
            int NumSpeeds = UnitarySystem(UnitarySysNum).NumOfSpeedCooling;
            if (!allocated(UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio)) UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio.allocate(NumSpeeds);
            for (int Iter = 1; Iter <= NumSpeeds; ++Iter) {
                UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(Iter) = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(Iter) / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;
            }
        }
        if (MultiOrVarSpeedHeatCoil(UnitarySysNum)) {
            int NumSpeeds = UnitarySystem(UnitarySysNum).NumOfSpeedHeating;
            if (!allocated(UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio)) UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio.allocate(NumSpeeds);
            for (int Iter = 1; Iter <= NumSpeeds; ++Iter) {
                UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(Iter) = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(Iter) / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;
            }
        }
        UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).IdleVolumeAirRate / UnitarySystem(UnitarySysNum).ActualFanVolFlowRate;
    }

I wonder if I should use MaxNoCoolHeatAirVolFlow on the last line instead of IdleVolumeAirRate?

rraustad:
Hmm, there are 3 diff's with warmup convergence diff's. We don't report speed ratio but I suspect this has changed.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2586,40 +2619,50 @@ namespace HVACUnitarySystem {
...
                 cAllCoilTypes(UnitarySystem(UnitarySysNum).CoolingCoilType_Num), UnitarySystem(UnitarySysNum).CoolingCoilName, ErrFound);
             EqSizing.DesCoolingLoad = DXCoolCap;
 
-            for (Iter = 1; Iter <= UnitarySystem(UnitarySysNum).NumOfSpeedCooling; ++Iter) {
+            for (Iter = NumSpeeds; Iter >= 1;
+                 --Iter) { // use reverse order since we divide by CoolVolumeFlowRate(max)
                 UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(Iter) =
                     VarSpeedCoil(UnitarySystem(UnitarySysNum).CoolingCoilIndex).MSRatedAirVolFlowRate(Iter);
                 UnitarySystem(UnitarySysNum).CoolMassFlowRate(Iter) = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(Iter) * StdRhoAir;
-                // it seems the ratio should reference the actual flow rates, not the fan flow ???
```
rraustad: So here, we set speed ratio relative to the fan if we knew what the fan flow rate was. That seems correct to me. Otherwise we set = coil max flow which is incorrect if fan flow is different from coil flow.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2586,40 +2619,50 @@ namespace HVACUnitarySystem {
...
+                    UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(Iter) /
+                    UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds);
             }
 
-            UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(1);
-            UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).CoolMassFlowRate(1);
-            UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1);
+            if (MSHPIndex > 0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate =
+                    UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds) * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
```
mjwitte: The current code produces the expected result for the defect file.
Still waiting to see all occurrences of `UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(NumSpeeds)` replaced with `UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow` (likewise for heating).

rraustad:
@Myoldmopar @mjwitte Something is going on. The Mac is failing with files not affected by changes. The issue is that I also get failures here on Windows. So it seems that something got added to develop that causes these failures?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -1698,6 +1698,23 @@ namespace HVACUnitarySystem {
                             }
                         }
                     }
+                    if (MultiOrVarSpeedHeatCoil(UnitarySysNum) || MultiOrVarSpeedCoolCoil(UnitarySysNum)) {
+                        if (MultiOrVarSpeedCoolCoil(UnitarySysNum)) {
```
mjwitte: This is in `InitLoadBasedControl`. Don't these need to be set also for setpoint based control?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2550,17 +2567,39 @@ namespace HVACUnitarySystem {
...
+        UnitarySystem(UnitarySysNum).MaxCoolAirMassFlow = UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow * StdRhoAir;
+        UnitarySystem(UnitarySysNum).MaxHeatAirMassFlow = UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow * StdRhoAir;
+        UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirMassFlow = UnitarySystem(UnitarySysNum).MaxNoCoolHeatAirVolFlow * StdRhoAir;
+
+        // initialize idle air flow rate variables in case these are needed in multi-speed heating coil sizing when multi-speed cooling coil does not
+        // exist
+        // the multi-speed coils will overwrite this data and these variables are only used for multi-speed coils in function SetOnOffMassFlowRate
+        if (MultiOrVarSpeedCoolCoil(UnitarySysNum) || MultiOrVarSpeedHeatCoil(UnitarySysNum)) {
+            if (UnitarySystem(UnitarySysNum).DesignFanVolFlowRate > 0.0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).DesignFanVolFlowRate;
```
mjwitte: Here we're using `DesignFanVolFlowRate` but above in the init, we're using `ActualFanVolFlowRate` - what's the difference between these?  Should they be the same?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2586,40 +2625,47 @@ namespace HVACUnitarySystem {
...
             }
 
-            UnitarySystem(UnitarySysNum).IdleVolumeAirRate = UnitarySystem(UnitarySysNum).CoolVolumeFlowRate(1);
-            UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).CoolMassFlowRate(1);
-            UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).MSCoolingSpeedRatio(1);
+            if (MSHPIndex > 0) {
+                UnitarySystem(UnitarySysNum).IdleVolumeAirRate =
+                    UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow * DesignSpecMSHP(MSHPIndex).NoLoadAirFlowRateRatio;
+                UnitarySystem(UnitarySysNum).IdleMassFlowRate = UnitarySystem(UnitarySysNum).IdleVolumeAirRate * StdRhoAir;
+                UnitarySystem(UnitarySysNum).IdleSpeedRatio = UnitarySystem(UnitarySysNum).IdleVolumeAirRate / UnitarySystem(UnitarySysNum).MaxCoolAirVolFlow;
```
mjwitte: And remind me, we're using `MaxCoolAirVolFlow` here because we might not know `DesignFanVolFlowRate` yet at this point?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2696,20 +2784,48 @@ namespace HVACUnitarySystem {
                     UnitarySystem(UnitarySysNum).HeatMassFlowRate(Iter) = UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(Iter) * StdRhoAir;
                     UnitarySystem(UnitarySysNum).MSHeatingSpeedRatio(Iter) =
                         UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(Iter) /
-                        UnitarySystem(UnitarySysNum).HeatVolumeFlowRate(DesignSpecMSHP(MSHPIndex).NumOfSpeedHeating);
+                        UnitarySystem(UnitarySysNum).MaxHeatAirVolFlow;
+                }
+            }
+            if (UnitarySystem(UnitarySysNum).CoolCoilExists) {
```
mjwitte: This block here could really use a comment to explain why the idle flow rates in the heating coil section are being touched if there's a cooling coil. "\\ If both heating and cooling coils are present, idle flow rate is based on the minimum of the heating and cooling supply air flow rates".

In the unitary refactor work, all of these similar idle flow rate blocks could be consolidated into a single section before all of the coil-specific blocks, because it isn't coil-type specific, it's more dependent on whether MSHPIndex is active or not.

mjwitte (commented):
Looking much better. Concerned about the large diffs in UnitarySystem_WaterCoils_wMultiSpeedFan [42% increase in July HVAC electricity](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2018-07/27c02bc659f002ef986f3e16a4755166c3fc3c20-48202596cd38b37c2304ed87c37e5a6142992058/UnitarySystem_WaterCoils_wMultiSpeedFan/x86_64-MacOS-10.9-clang/eplusmtr.csv.diffsummary.csv.html) and the new max iteration warnings in HVACTemplate-5ZoneUnitarySystem.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I moved this whole section (starting at "if (MyFanFlag(UnitarySysNum)) {") to InitUnitarySystems. This now gets done regardless of control type.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: Well, yes, they should be the same. This is in sizing so the fan hasn't sized yet. I don't know if the fan is autosized. Or I guess I do since they are both initialized to fan flow in GetInput so they could equate to Autosized. So at this point DesignFanVolFlowRate equals the autosized value of fan flow. But the fan could be hardsized. To complicate things, there is also an EMS override on this variable. So to me, the speed ratio initialization should not be done here and should wait until we know the actual fan flow. I think I have addressed that with the new calculations in Init.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I guess this should be DesignFanVolFlowRate given how speed ratio is used in the fan model. I think I'll change that since I am making other changes.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I'm not sure anymore that it's the min. It should be just IdleFlow / FanFlow.

rraustad:
The last thing I need to do is look at those large diff's to see if something was fixed or broken.

rraustad:
UnitarySystem_WaterCoils_wMultiSpeedFan has a blank input for no load flow rate. The default = 1.

    UnitarySystemPerformance:Multispeed,
      UnitarySystemFanFlowRatio,  !- Name
      4,                       !- Number of Speeds for Heating
      4,                       !- Number of Speeds for Cooling
      No,                      !- Single Mode Operation
      ,                        !- No Load Supply Air Flow Rate Ratio

    UnitarySystemPerformance:Multispeed,
      N3 , \field No Load Supply Air Flow Rate Ratio
           \type real
           \minimum 0
           \maximum 1
           \default 1
           \note Used to define the no load operating air flow rate when the system fan
           \note is specified to operate continuously.

This file uses a continuous fan operating mode schedule = 1 at all times. So it looks like it's working correctly now. The fan goes to full flow when there is no load. This is the strange results I described before and you commented about ill formed example files.

unitarysyswatercoilresults

mjwitte:
I see.  If you want to modify that example file to put a more reasonable no-load fraction there, that would be fine.  But if there are not other changes to make, then I'm ok with it as-is.

rraustad:
I added the water coil rate variable so you could see when the coils were off. This is just the sum of water cooling and heating rate variables in each case. The coils are off when the fan goes to full flow.

unitarysyswatercoilresults2

rraustad:
I could easily modify that example file. If I made the no load input 0 then the results should match. It would be as if the unit shuts down when there is no load.

mjwitte:
It's your call.  Not sure zero is a good choice, but I see your point of matching the old result.  Maybe use the lowest speed fraction?

mjwitte:
There's a build warning now, and UnitarySystem_DXCoilSystemAuto fails with a fatal error.
```
   ** Warning ** SizeUnitarySystem: AirLoopHVAC:UnitarySystem = DX COOLING COIL
   **  Fatal  ** Unable to determine fan air flow rate.
```
So, yeah, what happens with airflows when the fan is loose on the branch?

mjwitte:
@rraustad Looks very good.  Now there's just this nagging question of why HVACTemplate-5ZoneUnitarySystem with design days has new max iteration warnings. The problem is with system 2 which has multi-speed DX with an electric heating coil.
```
   **   ~~~   ** Air System Named = SYS 2 FURNACE DX COOL MULTISPD did not converge for mass flow rate
   **   ~~~   ** Check values should be zero. Most Recent values listed first.
   **   ~~~   ** Demand-to-Supply interface mass flow rate check value iteration history trace: 0.181256,0.181256,0.181256,0.181256,0.181256,0.000000,0.181256,0.181256,0.181256,0.181256,
   **   ~~~   ** Supply-to-demand interface deck 1 mass flow rate check value iteration history trace: 0.181256,0.181256,0.181256,0.181256,0.181256,0.181256,0.181256,0.181256,0.181256,0.181256,
   **   ~~~   ** Air System Named = SYS 2 FURNACE DX COOL MULTISPD did not converge for energy
   **   ~~~   ** Check values should be zero. Most Recent values listed first.
   **   ~~~   ** Demand-to-Supply interface energy check value iteration history trace: 4395.206591,4395.206591,4395.206591,4395.206591,4395.206591,0.000000,4395.206591,4395.206591,4395.206591,4395.206591,
   **   ~~~   ** Supply-to-demand interface deck 1 energy check value iteration history trace: 4339.106553,-4339.106553,4339.106553,-4339.106553,4349.014691,4339.106553,-4339.106553,4339.106553,-4339.106553,4339.106553,
   **   ~~~   ** Node named SPACE2-1 ZONE EQUIP INLET shows oscillating humidity ratio across iterations with a repeated value of 1.447481E-002
   **   ~~~   ** Node named SPACE2-1 ZONE EQUIP INLET humidity ratio [kg-water/kg-dryair] iteration history trace (most recent first): 1.447481E-002,1.430833E-002,1.447481E-002,1.430833E-002,1.447481E-002,1.430833E-002,1.447481E-002,1.430833E-002,1.447481E-002,1.430833E-002,
   **   ~~~   ** Node named SPACE2-1 ZONE EQUIP INLET shows oscillating mass flow rate across iterations with a repeated value of 0.271883
   **   ~~~   ** Node named SPACE2-1 ZONE EQUIP INLET mass flow rate [kg/s] iteration history trace (most recent first): 0.271883,9.062778E-002,0.271883,9.062778E-002,0.271883,9.062778E-002,0.271883,9.062778E-002,0.271883,9.062778E-002,
```
Probably the most telling info is the zone inlet oscillating mass flow rate - changing between 0.27 and 0.009 (which happens to be 33% which would be the speed 1 mass flow rate).

rraustad:
I would have thought [this](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HVACUnitarySystem.cc#L1955-L1976) would have caught that. I recall you didn't like that very much so maybe it's time to revisit to see if there is something better to do here. A new issue? or fix this here?

mjwitte:
Well, in this case, it's happening because expandobjects writes this to the expidf, which results in no load frac = 1.0. Changing that to something smaller (like 0.33 or 0.25) eliminates the max iteration warnings.
```
UnitarySystemPerformance:Multispeed,
  Sys 2 Furnace DX Cool MultiSpd Unitary System MultiSpeed Performance,  !- Name
  1,                                                       !- Number of Speeds for Heating
  3,                                                       !- Number of Speeds for Cooling
  No,                                                      !- Single Mode Operation
  ,                                                        !- No Load Supply Air Flow Rate Ratio
  autosize,                                                !- Heating Speed 1 Supply Air Flow Ratio
  autosize,                                                !- Cooling Speed 1 Supply Air Flow Ratio
  autosize,                                                !- Heating Speed 2 Supply Air Flow Ratio
  autosize,                                                !- Cooling Speed 2 Supply Air Flow Ratio
  autosize,                                                !- Heating Speed 3 Supply Air Flow Ratio
  autosize,                                                !- Cooling Speed 3 Supply Air Flow Ratio
```
Soooo, given the change in how blank no load frac is getting applied now, perhaps we should change expandobjects to write 1/(max(NumHeatingSpeeds,NumCoolingSpeeds)?
<sigh>

rraustad:
I agree. I was trying to think of a way to default the IDD to just that value.

rraustad:
OK, so my little oscillating check only looks at operating mode, cooling or heating, not air flow. If I had been looking at air flow this check could have locked in the speed?

rraustad:
Added ExpandObjects code to use min speed flow:

    UnitarySystemPerformance:Multispeed,
      Sys 2 Furnace DX Cool MultiSpd Unitary System MultiSpeed Performance,  !- Name
      1,                                                       !- Number of Speeds for Heating
      3,                                                       !- Number of Speeds for Cooling
      No,                                                      !- Single Mode Operation
      0.3333333,                                               !- No Load Supply Air Flow Rate Ratio

mjwitte:
@rraustad It's better to use the `StringToReal` function that's in expandobjects instead of `Read`.  I'll change this and bring in develop again. After CI is done with that this should be ready to merge.

mjwitte (approved):
Ready to merge this. @Myoldmopar I'll wait for WindowsCalc tests to finish so this doesn't muddy those results.

PR #6608, opened 2018-03-28
Output Variable Parser Fix

Myoldmopar:
Pull request overview
---------------------
With the style change after 8.9.0, the script that parses the SetupOutputVariable calls no longer functioned well.  I modified it to match by regex instead of line-by-line parsing.  It works really well now.  I didn't take the time to have the regex ignore commented SetupOutputVariable calls, so instead, I first check for any commented ones and warn the developer about it, then I go find all of them.  I am open to improving that if someone has the burning desire to make the regex match more robust.

To test this fix, just make a tag of it and see if CI is successful at packaging.  Currently it [fails](http://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-v8.9.0-PlusStyle-i386-Windows-7-VisualStudio-14.html#package).

### DIFF
--- a/doc/tools/parse_output_variables.py
+++ b/doc/tools/parse_output_variables.py
@@ -1,21 +1,23 @@
+#!/usr/bin/env python
 # python 2/3 compatibility imports
 from __future__ import absolute_import
-from __future__ import unicode_literals
 from __future__ import print_function
+from __future__ import unicode_literals
+
+import csv
+import glob
+import io
+import sys
+from os import path
+import re
 
 # generates a csv summary of the output variables in E+
-# assumption: function calls are single line -- **NOT** wrapped
+# assumption: doesn't check for commented lines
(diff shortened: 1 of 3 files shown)


nrel-bot-3:

nrel-bot-3:

nrel-bot-2:

Myoldmopar:
I just re-tested this.  I applied the latest clang-format style to AirflowNextworkBalanceManager, confirmed I got the new argument style, and then ran this updated script.  It successfully produced the output variable calls.  I think this is good.  It's definitely way better than the current one, since the current one is completely broken and definitely fails.  I'm going to kick off a tag from this branch and we'll confirm it makes package.  If so I'll merge this in.

Myoldmopar:
Everything looks good here, release packages now build with this updated script.  It's good to go.  Merging.

PR #6741, opened 2018-05-08
Test diff branch

Myoldmopar:
Pull request overview
---------------------
OK, so now develop has moved again with #6740.  This afternoon, CI should start picking up this PR and if all is well, it will be green and this can drop in and we can all get back to normal operation.  If not, I'll keep digging.


Myoldmopar:
Bingo

rraustad:
Why are these popping up in the AUD file?

    +found (input file)=C:\ci\EnergyPlus-6733386040-Win64-VS14\datasets\Window5DataFile.dat

Myoldmopar:
They started popping up with the JSON work.  Nothing important, but I would like to get rid of them sometime.  I had thought my modifications to the regression scripts would fix it but they didn't seem to.  I'll try to fix that up sometime.

PR #6745, opened 2018-05-14
Set multiple CO2 controllers correctly

lgu1234:
Pull request overview
---------------------
The fix set multiple CO2 controllers correctly and predict required OA rates based on its own setpoint.  
### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
+++ b/src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
@@ -1366,8 +1366,8 @@ namespace ZoneContaminantPredictorCorrector {
         if (allocated(ZoneEquipConfig) && MyConfigOneTimeFlag) {
             for (ContZoneNum = 1; ContZoneNum <= NumContControlledZones; ++ContZoneNum) {
                 ZoneNum = ContaminantControlledZone(ContZoneNum).ActualZoneNum;
-                for (int zoneInNode = 1; zoneInNode <= ZoneEquipConfig(ContZoneNum).NumInletNodes; ++zoneInNode) {
-                    int AirLoopNum = ZoneEquipConfig(ContZoneNum).InletNodeAirLoopNum(zoneInNode);
+                for (int zoneInNode = 1; zoneInNode <= ZoneEquipConfig(ZoneNum).NumInletNodes; ++zoneInNode) {
+                    int AirLoopNum = ZoneEquipConfig(ZoneNum).InletNodeAirLoopNum(zoneInNode);
                     ContaminantControlledZone(ContZoneNum).NumOfZones = 0;
                     for (Loop = 1; Loop <= NumOfZones; ++Loop) {
                         if (!ZoneEquipConfig(Loop).IsControlled) continue;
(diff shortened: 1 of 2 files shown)


Myoldmopar:
@lgu1234 I built this on my Linux box and my Mac, and the unit test passes when run by itself in both locations.  I'm kicking off a full run of all the unit tests and we'll see what happens.

Myoldmopar:
The full run of all the unit tests on my Mac ran fine.  It was a debug build, however, so I'll kick off one more run in release mode and see if that possibly shows the issue.  I deleted those CI results so CI should pick that up and rerun it as a confirmation step that it wasn't a fluke.

lgu1234:
@Myoldmopar The Mac runs show green. Thanks.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
```diff
@@ -1366,8 +1366,8 @@ namespace ZoneContaminantPredictorCorrector {
         if (allocated(ZoneEquipConfig) && MyConfigOneTimeFlag) {
             for (ContZoneNum = 1; ContZoneNum <= NumContControlledZones; ++ContZoneNum) {
                 ZoneNum = ContaminantControlledZone(ContZoneNum).ActualZoneNum;
-                for (int zoneInNode = 1; zoneInNode <= ZoneEquipConfig(ContZoneNum).NumInletNodes; ++zoneInNode) {
-                    int AirLoopNum = ZoneEquipConfig(ContZoneNum).InletNodeAirLoopNum(zoneInNode);
+                for (int zoneInNode = 1; zoneInNode <= ZoneEquipConfig(ZoneNum).NumInletNodes; ++zoneInNode) {
+                    int AirLoopNum = ZoneEquipConfig(ZoneNum).InletNodeAirLoopNum(zoneInNode);
```
Myoldmopar, line 1372: This makes sense.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
```diff
@@ -1665,27 +1665,41 @@ namespace ZoneContaminantPredictorCorrector {
                 ControlledCO2ZoneFlag = false;
                 // Check all the controlled zones to see if it matches the zone simulated
                 for (ContControlledZoneNum = 1; ContControlledZoneNum <= NumContControlledZones; ++ContControlledZoneNum) {
-                    if (GetCurrentScheduleValue(ContaminantControlledZone(ContControlledZoneNum).AvaiSchedPtr) > 0.0) {
-                        ZoneAirCO2SetPoint = ZoneCO2SetPoint(ContaminantControlledZone(ContControlledZoneNum).ActualZoneNum);
-                        if (ContaminantControlledZone(ContControlledZoneNum).EMSOverrideCO2SetPointOn) {
-                            ZoneAirCO2SetPoint = ContaminantControlledZone(ContControlledZoneNum).EMSOverrideCO2SetPointValue;
+                    if (ContaminantControlledZone(ContControlledZoneNum).ActualZoneNum == ZoneNum) {
```
Myoldmopar, line 1670: So this loops over all contaminant controllers, looking for one that matches this zone.  This sort of backward referencing could require a lot more looping than what is actually needed.  It would be nice if there was some sort of forward referencing where we just loop over each existing controller and avoid all the empty steps.  But that's out of scope for here.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
```diff
@@ -1665,27 +1665,41 @@ namespace ZoneContaminantPredictorCorrector {
...
+                        if (GetCurrentScheduleValue(ContaminantControlledZone(ContControlledZoneNum).AvaiSchedPtr) > 0.0) {
+                            ZoneAirCO2SetPoint = ZoneCO2SetPoint(ContaminantControlledZone(ContControlledZoneNum).ActualZoneNum);
+                            if (ContaminantControlledZone(ContControlledZoneNum).EMSOverrideCO2SetPointOn) {
+                                ZoneAirCO2SetPoint = ContaminantControlledZone(ContControlledZoneNum).EMSOverrideCO2SetPointValue;
+                            }
+                            ControlledCO2ZoneFlag = true;
+                            break;
```
Myoldmopar, line 1677: At first this `break` seemed odd to me.  I thought you'd want to keep looping to get all of them that apply, but that's not the case.  You only want to get the first one for a particular zone, but you do want to make sure all of the controllers (overall) are modeled.  This makes sense, but please confirm that is the intent.

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
```diff
@@ -390,3 +390,220 @@ TEST_F(EnergyPlusFixture, ZoneContaminantPredictorCorrector_CorrectZoneContamina
...
+    MixingMassFlowZone.allocate(3);
+
+    DataGlobals::NumOfZones = 3;
+    DataContaminantBalance::Contaminant.CO2Simulation = true;
+ 
+    DataContaminantBalance::AZ.allocate(3);
+    DataContaminantBalance::BZ.allocate(3);
+    DataContaminantBalance::CZ.allocate(3);
+
+    DataContaminantBalance::CO2ZoneTimeMinus1Temp.allocate(3);
```
Myoldmopar, line 411: It would be very nice if these were just member variables of one struct, and you only had to allocate it once, instead of a bunch of dangling arrays that all have to magically be allocated to the same amount... nothing for here, just ranting :laughing:

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
```diff
@@ -390,3 +390,220 @@ TEST_F(EnergyPlusFixture, ZoneContaminantPredictorCorrector_CorrectZoneContamina
...
+
+    OAMFL.allocate(3);
+    VAMFL.allocate(3);
+    EAMFL.allocate(3);
+    CTMFL.allocate(3);
+    MDotOA.allocate(3);
+    MDotOA = 0.001;
+    ScheduleManager::Schedule.allocate(1);
+
+    ScheduleManager::Schedule(1).CurrentValue = 1.0;
```
Myoldmopar, line 543: It's nice that this is so easy, I had assumed overriding the schedule would be a bit more effort.

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
```diff
@@ -390,3 +390,220 @@ TEST_F(EnergyPlusFixture, ZoneContaminantPredictorCorrector_CorrectZoneContamina
...
+    DataContaminantBalance::ContaminantControlledZone(1).ActualZoneNum = 1;
+    DataContaminantBalance::ContaminantControlledZone(1).NumOfZones = 1;
+    DataContaminantBalance::ContaminantControlledZone(2).AvaiSchedPtr = 1;
+    DataContaminantBalance::ContaminantControlledZone(2).ActualZoneNum = 2;
+    DataContaminantBalance::ContaminantControlledZone(2).NumOfZones = 1;
+    DataContaminantBalance::ContaminantControlledZone(3).AvaiSchedPtr = 1;
+    DataContaminantBalance::ContaminantControlledZone(3).ActualZoneNum = 3;
+    DataContaminantBalance::ContaminantControlledZone(3).NumOfZones = 1;
+
+    PredictZoneContaminants(ShortenTimeStepSys, UseZoneTimeStepHistory, PriorTimeStep);
```
Myoldmopar, line 602: :+1:

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
```diff
@@ -390,3 +390,220 @@ TEST_F(EnergyPlusFixture, ZoneContaminantPredictorCorrector_CorrectZoneContamina
...
+    EXPECT_NEAR(1.0416921806, DataContaminantBalance::CO2PredictedRate(1), 0.00001);
+    EXPECT_NEAR(1.0434496257, DataContaminantBalance::CO2PredictedRate(2), 0.00001);
+    EXPECT_NEAR(1.0399406399, DataContaminantBalance::CO2PredictedRate(3), 0.00001);
+
+    DataContaminantBalance::Contaminant.CO2Simulation = false;
+    DataContaminantBalance::Contaminant.GenericContamSimulation = false;
```
Myoldmopar: These types of re-initializations should be done in the clear_state function in `DataContaminantBalance.cc`.  That way other unit tests don't have to "remember" to clean up like this.

Myoldmopar (changes requested):
I'm requesting a very minor change to the unit test (very last comment), and I want to confirm the intended behavior.  Please verify.  If the only change is the unit test change, I can make that code change and push back up.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
lgu1234: Yes.

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
lgu1234: I agree with you. I will make a GitHub issue as a place holder for future improvement.

tst/EnergyPlus/unit/ZoneContaminantPredictorCorrector.unit.cc
lgu1234: @Myoldmopar I am going to add these two statements in the clear_state function.

Myoldmopar (approved):

Issue #6546, opened 2018-03-19
Address limitations and overshooting problem


lgu1234:
Issue overview
--------------
This is an issue to address limitations and overshooting problem for OnOffThermostat. Here are detailed limitations based on Mike's comments:

- [ ] I've tested SingleFamilyHouse_TwoSpeed_CutoutTemperatureI with this branch and current develop. The temperature pattern with this branch is much better. The system see that this control strategy does not work well for a heat pump, because the supplemental heat is coming on (even in April) to push the zone to the higher heating setpoint. For this to work correctly with a heat pump, the heat pump will need to know more information so that the supp heat does not kick in for this. Also, there are quite a few times where the zone air temperature overshoots the upper heating setpoint one timestep after the heating shuts off - doesn't seem possible. For example
image

- [x] I also tested with a modified version of HVACTemplate-5ZoneFurnace with fan operating mode set to cycling fan. The daytime heating setpoint is 23C and the cooling setpoint is 24C.
 For some reason, with this system type the cooling setpoint is tracking the heating setpoint which is causing problems (requesting cooling on the winter day, and it fails with a fatal thermostat error). I don't see the same setpoint behavior in SingleFamilyHouse_TwoSpeed_CutoutTemperature.
Then I changed the cooling setpoint to 27C and the setpoints were indepednent. So, apparently if the cooling and heating setpoint are too close to each other they stomp on each other?
 If it's possiible to detect and protect against it or throw a meaningful error message, that would be good here. New input files attached.
*This was addressed prior to v8.9.0 release.  Probably in #6541.*

- [x] User correct zone temperature variable to limit overshoot problems.
*Addressed in #6754*

- [ ] Also, excess hours of thermostat setpoint not met are being counted with this method. That's another item that will need improvement.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

lgu1234:
@mjwitte 
Overshooting issue:

[...]

PR #6754, opened 2018-05-18
Address overshooting using correct zone air temperature at previous timestep

lgu1234:
When ThirdOrderBackwardDifference is used as a choice of ZoneAirHeatBalanceAlgorithm, ZTM1 is used to represent the zone temperature at previous time step. The variable is used in zone energy balance equation to predict system load and solve zone temperature at the current time step. Unfortunately, ZTM1 does not represent the value in the OnOff thermostat. The correct value is used as zone temperature at the previous time step in OnOff thermostat. Partially addresses #6546.

Pull request overview
---------------------

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes n/a


src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3608,7 +3608,10 @@ namespace ZoneTempPredictorCorrector {
                 if (TempCtrlFound && TempControlledZone(RelativeZoneNum).DeltaTCutSet > 0.0) {
                     TempZoneThermostatSetPoint(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
                     ZoneThermostatSetPointLo(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
-                    if (ZoneAirSolutionAlgo == Use3rdOrder) Tprev = ZTM1(ZoneNum);
+                    if (ZoneAirSolutionAlgo == Use3rdOrder) {
+                        Tprev = MAT(ZoneNum);
+                        if (ShortenTimeStepSys) Tprev = XMPT(ZoneNum);
+                    }
```
rraustad: Instead of testing for shortened time step could ZTM1 be updated correctly so this test isn't needed here?

lgu1234: @rraustad You raised a good question. Here are several reasons I am reluctant to make changes of ZTM1 correctly in the original place.

1. ZTM1 is assigned by XMAT in line 3190 in ZoneTempPredictorCorrector
                ZTM1(ZoneNum) = XMAT(ZoneNum);
XMAT is assigned by ZTAV in line 4911 in ZoneTempPredictorCorrector
            XMAT(ZoneNum) = ZTAV(ZoneNum); // using average for whole zone time step.

2 When DownInterpolate4HistoryValues is called, the zone air temperature at previous time steps used in 3order equations are interpolated with lengths of previous time steps.

Therefore, ZTM1 is not exactly a value at previous time step.

The best way to fix is to remove downshifting, so that we don't have any issues for history terms. I proposed this idea before. 

My fix is to allow zone air temperature at previous time step is used correctly for this new feature only. If I change it in the originally place, it will cause many differences, so that it will not be a small fix.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3608,7 +3608,10 @@ namespace ZoneTempPredictorCorrector {
                 if (TempCtrlFound && TempControlledZone(RelativeZoneNum).DeltaTCutSet > 0.0) {
```
mjwitte: @lgu1234 When this if is false, `Tprev` is uninitialized below at [line 3630](https://github.com/NREL/EnergyPlus/pull/6754/files#diff-26a720ee34299122c810df4c89b63101R3630).  Looks like the closing } in 3629 should be move down.  Similar problem appears in the SingleCoolingSetPoint block, but DualSetPoint looks OK.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3608,7 +3608,10 @@ namespace ZoneTempPredictorCorrector {
...
                     TempZoneThermostatSetPoint(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
                     ZoneThermostatSetPointLo(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
-                    if (ZoneAirSolutionAlgo == Use3rdOrder) Tprev = ZTM1(ZoneNum);
+                    if (ZoneAirSolutionAlgo == Use3rdOrder) {
+                        Tprev = MAT(ZoneNum);
+                        if (ShortenTimeStepSys) Tprev = XMPT(ZoneNum);
+                    }
                     if (ZoneAirSolutionAlgo == UseAnalyticalSolution || ZoneAirSolutionAlgo == UseEulerMethod) Tprev = ZoneT1(ZoneNum);
```
mjwitte: This should be an elseif or simply else. Same comment in other similar blocks.

mjwitte: Also, all of this stuff related to onoff control should be skipped if there aren't any onoff thermostats in the idf.  As it stands now, this [loop to find RelativeZoneNumber](https://github.com/NREL/EnergyPlus/pull/6754/files#diff-26a720ee34299122c810df4c89b63101L3581) happens on every call, no matter what type of thermostat control these is or isn't. This whole RelativeZoneNum thing is unnecessary, but that's for another day. For now, mimic the staged thermostats and add `NumOnOffCtrZone` flag similar to `NumStageCtrZone ` that skips all of this if there aren't any.

Also, it seems that all of this (which was move here in #6499) would be better placed [here](https://github.com/NREL/EnergyPlus/blob/b60abaa44542b4495671647f18044141b63b1ac2/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L3060) in `PredictSystemLoads` rather than `CalcPredictedSystemLoads`. This would make it similar to the staged thermostat and would put this before some of the zone air temp updates which might be a better fix than this one.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte You are right. It is a very good catch. Changes are done. Thanks.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte I like it and am going to make changes.

lgu1234: @mjwitte The section was moved into PredictSystemLoads.

mjwitte:
@lgu1234 I do not see much difference in zone temperature behavior before and after this fix. Can you point me to which file/which times show improvement? Here are csv outputs for before, current, and the original fix on this branch.
Note that I merged in develop to bring this branch up to date.
[6546ToGu.zip](https://github.com/NREL/EnergyPlus/files/2117140/6546ToGu.zip)

lgu1234:
@mjwitte Thanks for bringing the branch up to date. The previous upload addressed comments only:

1.	Initialize Tprev
2.	Use an elseif to cover two scenarios
3.	Move a section of code from CalcPredictedSystemLoad to PredictSystemLoads

Therefore, results are not changed. The fix was from the previous upload on May 18, 2018 to use correct zone air temperature at previous time step, instead of using ZTM1 when ZoneAirSolutionAlgo = Use3rdOrder.

mjwitte:
@lgu1234 Which test file shows that this branch makes a change in results compared to current develop?

lgu1234:
@mjwitte You may try a test file in https://github.com/NREL/EnergyPlusDevSupport/blob/master/DefectFiles/6000s/6546/6546Detailed.idf.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3416,6 +3421,147 @@ namespace ZoneTempPredictorCorrector {
             }
         }
 
+        // Setpoint revision for onoff thermostat
+        Real64 TempTole = 0.02;
+        Real64 Tprev;
```
mjwitte: These two lines can move inside the main if block below.  No need for these to be created if there aren't any NumOnOffCtrZone.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -3416,6 +3421,147 @@ namespace ZoneTempPredictorCorrector {
...
+                    if (SELECT_CASE_var == SingleHeatingSetPoint) {
+                        TempZoneThermostatSetPoint(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
+                        ZoneThermostatSetPointLo(ZoneNum) = TempControlledZone(RelativeZoneNum).ZoneThermostatSetPointLo;
+                        if (ZoneAirSolutionAlgo == Use3rdOrder) {
+                            Tprev = MAT(ZoneNum);
+                            if (ShortenTimeStepSys) Tprev = XMPT(ZoneNum);
+                        }
+                        else {
+                            Tprev = ZoneT1(ZoneNum);
+                        }
```
mjwitte: This section that sets `Tprev` looks identical for all three thermostat types, so it could appear just once before line 3442 unless I'm missing something.

mjwitte (approved):
@lgu1234 This is almost ready to go in. This gives a nice little performance boost by avoiding some code when there aren't any on/off tstats. Just a couple more minor changes and one more round of update to current develop should finish this.

See more comments over in #6546.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
mjwitte: This comment refers to lines 3445:3451

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte Yes. I move these two line inside the if block.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
lgu1234: @mjwitte I agree. I am going to make changes.

Issue #6750, opened 2018-05-17
Confusing errors when no control zone name is specified for Load control in AirloopHVAC:UnitarySystem


rraustad:
Issue overview
--------------
Initial checks for system connections fail when Load control is specified and no Control Zone Name is entered. Review of code also indicates that the AirTerminal:SingleDuct:Mixer will not work correctly when used with AirloopHVAC:UnitarySystem if Setpoint control is specified.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [x] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
See PR #6757.

mjwitte:
Closed via #6757

PR #6757, opened 2018-05-21
AirloopHVAC:UnitarySystem: Correct AT Mixer operation for set point control and Improve error message when Control Type = Load and no control zone name is specified

rraustad:
Pull request overview
---------------------
Addresses #6750. The control zone name input field must be specified when Load control is used otherwise the simulation fails with confusing error messages. Also corrects AirTerminal:SingleDuct:Mixer operation when Setpoint control is specified.

Workaround: Specify control zone name when Load control is specified. If an AirTerminal:SingleDuct:Mixer is used, specify Load control and enter control zone name.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes n/a


rraustad:
New warning message:

    ** Warning ** AirLoopHVAC:UnitarySystem: GUID11_SYS0SZUHC-1
    **   ~~~   ** When Control Type = LOAD
    **   ~~~   ** Controlling Zone or Thermostat Location must be a valid zone name, zone name = 
    **  Fatal  ** GetUnitarySystemInput: Errors found in getting AirLoopHVAC:UnitarySystem input. Preceding condition(s) causes termination.

rraustad:
I was expecting no diff's and mostly get that. The Mac results look suspect. I am deleting the results and running again.

rraustad:
Now these results look as expected (no diff's). Although RefrigeratedWarehouse shows this in the err file, which should not show up in the err file. This looks like a list of objects reported by JSON?

         **   ~~~   ** Object=Output:Table:Monthly=Zone Electric Summary
         **   ~~~   ** Object=Output:Table:Monthly=Space Gains
         **   ~~~   ** Object=Output:Table:Monthly=Zone Heating Summary
    +   **   ~~~   ** Object=Output:Table:Monthly=Space Gain Components at Cooling Peak
          **   ~~~   ** Object=Output:Table:Monthly=Peak Space Gains
    -   **   ~~~   ** Object=Output:Table:Monthly=Space Gain Components at Cooling Peak
         **   ~~~   ** Object=Output:Table:Monthly=Energy Consumption - Electricity Generated & Propane
         **   ~~~   ** Object=Output:Table:Monthly=Energy Consumption - Diesel & Fuel Oil
    +   **   ~~~   ** Object=Output:Table:Monthly=Energy Consumption  - Purchased Heating & Cooling
         **   ~~~   ** Object=Output:Table:Monthly=Energy Consumption - Coal & Gasoline
    -   **   ~~~   ** Object=Output:Table:Monthly=Energy Consumption  - Purchased Heating & Cooling
    +   **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Electricity
    +   **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Fuel Oil
         **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Natural Gas
    -   **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Electricity
         **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Diesel
    -   **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Fuel Oil
         **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Coal
         **   ~~~   ** Object=Output:Table:Monthly=End-Use Energy Consumption - Propane

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -9327,7 +9345,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_GetInputATMixerInlet)
         "AirLoopHVAC:UnitarySystem,",
         "  East Zone Unitary System,       !- Name",
         "  Setpoint,                       !- Control Type",
-        "  East Zone,                      !- Controlling Zone or Thermostat Location",
+        "  ,                               !- Controlling Zone or Thermostat Location",
```
rraustad, line 9550: Removed the control zone name here and 2 places below since control zone name is not needed for Control Type = Setpoint.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -9665,3 +9683,116 @@ TEST_F(EnergyPlusFixture, UnitarySystem_GetInputZoneEquipment)
     EXPECT_FALSE(UnitarySystem(1).AirLoopEquipment);
     EXPECT_EQ(1, UnitarySystem(1).ControlZoneNum);
 }
+
```
rraustad, line 9888: New unit test.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -510,8 +510,8 @@ namespace HVACUnitarySystem {
         // Uses the status flags to trigger initializations.
```
rraustad: Massive diff's here due to clang.format. Will add comments at the few places changes were actually made.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -3466,6 +3694,16 @@ namespace HVACUnitarySystem {
             // Get the Controlling Zone or Location of the Thermostat
             UnitarySystem(UnitarySysNum).ControlZoneNum = UtilityRoutines::FindItemInList(Alphas(iControlZoneAlphaNum), Zone);
 
+            // check that control zone name is valid for load based control
```
rraustad, line 3867: Added warning that control zone name was missing when Control Type = Load or SZVAV.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -3487,6 +3725,7 @@ namespace HVACUnitarySystem {
                     }
                     if (UnitarySystem(UnitarySysNum).Humidistat && UnitarySystem(UnitarySysNum).ControlType == LoadBased) {
                         for (HStatZoneNum = 1; HStatZoneNum <= NumHumidityControlZones; ++HStatZoneNum) {
+                            if (UnitarySystem(UnitarySysNum).ControlZoneNum == 0) continue;
```
rraustad: Cycle here on ControlZoneNum == 0 since a zone can be found when ActualZoneNum = 0 and ControlZoneNum = 0.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -3757,6 +4010,7 @@ namespace HVACUnitarySystem {
                                 AirLoopNumber = AirLoopNum;
                                 AirLoopFound = true;
                                 for (ControlledZoneNum = 1; ControlledZoneNum <= NumOfZones; ++ControlledZoneNum) {
+                                    if (UnitarySystem(UnitarySysNum).ControlZoneNum == 0) continue;
```
rraustad: Cycle here on ControlZoneNum == 0 since a zone can be found when ActualZoneNum = 0 and ControlZoneNum = 0.

This concludes changes actually made for this issue.

src/EnergyPlus/HVACUnitarySystem.cc
mjwitte: This works, but if `ControlZoneNum=0` this whole loop should be skipped.

src/EnergyPlus/HVACUnitarySystem.cc
mjwitte: I don't think this new line is needed. But why is line 4017 setting `.ControlZoneNum` below? That line shouldn't be there, should it?

mjwitte (commented):
I've tested this, and the new error message shows now. A couple of comments below.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I agree. I put this line here because line 4017 was overwriting UnitarySystem.ControlZoneNum. ControlZoneNum which shouldn't ever be changed after it is read in. I can't recall why this (line 4017) is here. Do no harm. Let me think through this more.

rraustad: I see. This was added? or linked in with? the AT mixer error checking. Return or Supply mixer. I guess this test is only valid for load based control since only that control type requires a control zone name. This condition (load based and control zone = 0) causes a fatal so I guess I should skip all this?

rraustad:
I am now wondering if the AT mixers even work with set point control since I don't see use of .ATMixerExists in the set point code.

mjwitte:
OK, maybe. But ATMixer only applies if the unitary system is zone equipment. Can it do setpoint based control as zone equipment?

rraustad:
The only limitation should be load control in the OA system. Everything else should work.

Converted example file DOAToUnitarySystem to Setpoint control as:

    AirLoopHVAC:UnitarySystem,
      SPACE1-1 PTAC,           !- Name

    SetpointManager:SingleZone:Heating,
      Heating Supply Air Temp Manager,  !- Name
      Temperature,             !- Control Variable
      -99.,                    !- Minimum Supply Air Temperature {C}
      99.,                     !- Maximum Supply Air Temperature {C}
      SPACE1-1,                !- Control Zone Name
      SPACE1-1 Zone Air Node,  !- Zone Node Name
      SPACE1-1 Supply Inlet,   !- Zone Inlet Node Name
      SPACE1-1 Supply Inlet;   !- Setpoint Node or NodeList Name

    SetpointManager:SingleZone:Cooling,
      Cooling Supply Air Temp Manager,  !- Name
      Temperature,             !- Control Variable
      -99.,                    !- Minimum Supply Air Temperature {C}
      99.,                     !- Maximum Supply Air Temperature {C}
      SPACE1-1,                !- Control Zone Name
      SPACE1-1 Zone Air Node,  !- Zone Node Name
      SPACE1-1 Supply Inlet,   !- Zone Inlet Node Name
      SPACE1-1 Cooling Coil Outlet;  !- Setpoint Node or NodeList Name


ATMixer results for load and SP control before the last fix:

atmixer_unitarysystem_failed


And ATMixer results using SP control after the last fix:

atmixer_unitarysystem_fixed

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -929,33 +1012,69 @@ namespace HVACUnitarySystem {
             if (FirstHVACIteration) {
                 UnitarySystem(UnitarySysNum).iterationCounter = 0;
                 UnitarySystem(UnitarySysNum).iterationMode = 0;
```
rraustad, line 1094: I have been seeing issues with SP control where the system flow rate was always 0. This is because SP control relies on the TU to set the flow. If someone uses the uncontrolled TU then the unitary system would need to set the flow. So on first HVAC I want to set full flow and then let the TU modulate as needed on subsequent iterations.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -7127,12 +7681,23 @@ namespace HVACUnitarySystem {
         OnOffAirFlowRatio = 1.0;
         CoilCoolHeatRat = 1.0;
```
rraustad, line 7816: Simple fix for set point based control using ATMixer was to add these calls before and after the coils/fan are simulated in the set point calc routine.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -7127,12 +7681,23 @@ namespace HVACUnitarySystem {
...
+                SingleDuct::SimATMixer(UnitarySystem(UnitarySysNum).ATMixerName, FirstHVACIteration, UnitarySystem(UnitarySysNum).ATMixerIndex);
+            }
+        }
+
         // CALL the series of components that simulate a Unitary System
         if (UnitarySystem(UnitarySysNum).FanExists && UnitarySystem(UnitarySysNum).FanPlace == BlowThru) {
             if (UnitarySystem(UnitarySysNum).FanType_Num == DataHVACGlobals::FanType_SystemModelObject) {
-
-                HVACFan::fanObjs[UnitarySystem(UnitarySysNum).FanIndex]->simulate(_, _, _, _, m_massFlow1, m_runTimeFraction1, m_massFlow2,
-                                                                                  m_runTimeFraction2, _);
```
rraustad: The set point control code does not call SetOnOffMassFlowRate where these variables are set. So adjusting the system fan call so that the fan does not shut down flow (which it would if these variables are 0).

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -7141,71 +7706,109 @@ namespace HVACUnitarySystem {
...
+                                           FirstHVACIteration,
                                            OAUCoilOutTemp);
                 ControlCoolingSystemToSP(UnitarySysNum, AirLoopNum, FirstHVACIteration, HXUnitOn, CompOn);
                 PartLoadRatio = UnitarySystem(UnitarySysNum).CoolingPartLoadFrac;
                 CompOn = 0;
-                if (PartLoadRatio > 0.0) CompOn = 1;
-                CalcUnitaryCoolingSystem(UnitarySysNum, AirLoopNum, FirstHVACIteration, PartLoadRatio, CompOn, OnOffAirFlowRatio, CoilCoolHeatRat,
-                                         HXUnitOn);
+                if (PartLoadRatio > 0.0) {
+                    CompOn = 1;
```
rraustad: I'm saving the operating mode when SP based coils are called so that I have a way to set flow in Init for SP based control.

rraustad:
@mjwitte great review and comments. I think I have this resolved now.

nrel-bot-3:

Myoldmopar:
@rraustad I tried resolving the conflicts hoping they were only style related.  There is some change to some function calls, so I abandoned.  If you can get the conflicts I'll get this  in my queue to review and get in.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -7660,12 +7667,24 @@ namespace HVACUnitarySystem {
...
+            }
+        }
+
         // CALL the series of components that simulate a Unitary System
         if (UnitarySystem(UnitarySysNum).FanExists && UnitarySystem(UnitarySysNum).FanPlace == BlowThru) {
             if (UnitarySystem(UnitarySysNum).FanType_Num == DataHVACGlobals::FanType_SystemModelObject) {
-
-                HVACFan::fanObjs[UnitarySystem(UnitarySysNum).FanIndex]->simulate(
-                    _, _, _, _, m_massFlow1, m_runTimeFraction1, m_massFlow2, m_runTimeFraction2, _);
+                HVACFan::fanObjs[UnitarySystem(UnitarySysNum).FanIndex]->simulate(_, _, _, _);
```
rraustad, line 7834: For set point based control, the SetOnOffMassFlowRate function is not called and is where m_massFlow1, m_runTimeFraction1, etc. get set. Since these are missing (and are 0) they should not be included in the fan call.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -7684,7 +7703,10 @@ namespace HVACUnitarySystem {
                 ControlCoolingSystemToSP(UnitarySysNum, AirLoopNum, FirstHVACIteration, HXUnitOn, CompOn);
                 PartLoadRatio = UnitarySystem(UnitarySysNum).CoolingPartLoadFrac;
                 CompOn = 0;
-                if (PartLoadRatio > 0.0) CompOn = 1;
+                if (PartLoadRatio > 0.0) {
+                    CompOn = 1;
+                    UnitarySystem(UnitarySysNum).LastMode = CoolingMode;
+                }
```
rraustad, line 7856: Keeping track of last mode for set point control mode.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -9805,10 +9842,10 @@ namespace HVACUnitarySystem {
         Real64 CoilPLR;            // variable speed coils run at PLR = 1 when SpeedNum > 1
         Real64 OutdoorPressure;    // Outdoor barometric pressure at condenser (Pa)
         bool errFlag;              // returned flag from called routine
-        bool HeatingActive; // dummy variable for UserDefined coil which are passed back indicating if coil is on or off. Not needed here since coil
```
rraustad, line 9951: only style changes after this point.

nrel-bot-3:

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -3946,7 +3964,6 @@ namespace HVACUnitarySystem {
                                         ZoneEquipList(ZoneEquipConfig(ControlledZoneNum).EquipListIndex).HeatingPriority(EquipNum);
                                 }
                             }
-                            UnitarySystem(UnitarySysNum).ControlZoneNum = ControlledZoneNum;
```
mjwitte, line 4092: Why delete this line?

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -4000,14 +4010,11 @@ namespace HVACUnitarySystem {
                                 AirLoopFound = true;
                                 for (ControlledZoneNum = 1; ControlledZoneNum <= NumOfZones; ++ControlledZoneNum) {
                                     if (ZoneEquipConfig(ControlledZoneNum).ActualZoneNum != UnitarySystem(UnitarySysNum).ControlZoneNum) continue;
-                                    //             Find the controlled zone number for the specified thermostat location
+                                    // Find the controlled zone number for the specified thermostat location
                                     UnitarySystem(UnitarySysNum).NodeNumOfControlledZone = ZoneEquipConfig(ControlledZoneNum).ZoneNode;
-                                    UnitarySystem(UnitarySysNum).ControlZoneNum = ControlledZoneNum;
```
mjwitte, line 4148: Again, why delete?

mjwitte (commented):
I'm seeing unexpected diffs in the eio output for the SP defect file.

Before (current Develop Version 9.0.0-2d029a9105)
```
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Maximum Heating Flow Rates [m3/s], 0.14933
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Minimum Heating Flow Rates [m3/s], 0.14933
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Maximum Flow Rates [m3/s], 0.14933
```

After
```
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Maximum Heating Flow Rates [m3/s], 0.20658
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Minimum Heating Flow Rates [m3/s], 0.20658
 Component Sizing Information, AirLoopHVAC, DOAS, Sum of Air Terminal Maximum Flow Rates [m3/s], 0.20658
```

Still trying to understand this defect file.  It has unitary system zone equipment fed by a DOAS via AT Mixers, so not understanding why the unitary system would have any impact on the air flow rates in the DOAS system.

src/EnergyPlus/HVACUnitarySystem.cc
rraustad: I'm just getting back to this. ControlZoneNum was set in GetInput if the zone name was specified. If load control or SZVAV is used, the control zone name is required. For set point control the control zone number isn't needed. So 1) these 2 lines aren't needed anymore  and 2) I don't want this loop to be picking up any old zone number. The user either gives me the correct zone name or I issue a fatal. I see something strange up at line 3753 of this branch so I'll be looking at that again and correcting as needed. I want to check for the humidistat, but only issue the warning if zone num > 0.

rraustad:
Tested this change with every combination of control type, thermostat location and humidistat location I could think of. Control type blank for Load control gives fatal, control type blank for set point runs. Added dehumidification control type CoolReheat and got Hstat warning. Tested with missing Tstat and did not get same warning so added check for valid Tstat. Tested with dehum control CoolReheat using load control without Hstat and got missing humidity ratio set point. This seems to give warnings when expected now, at least as far as I could think to test.

rraustad:
No real diff's here. One is warmup convergence, another is that epfilter in develop changed [here](https://github.com/NREL/EnergyPlus/pull/6600/files) where HVAC template multispeed coils now use the lowest flow rate possible for no load (flow proportional to 1/num speeds) and the last one is the example file UnitarySystem_WaterCoils_wMultiSpeedFan.idf changed [here](https://github.com/NREL/EnergyPlus/pull/6600/files) to use 0.25 as no load flow ratio instead of leaving the field blank.

rraustad:
I just ran the defect SP file in develop and this branch and got the same numbers you did. The reason is that in develop, SysATMixer(1).AirLoopNum = 0 (all others are 1) and the loop skips that zone/ATMixer in the summation of airLoopHeatingMaximumFlowRateSum. In my branch, all the SysATMixers air loop value = 1 so all are included in the summation. So that answers why the results are different between develop and this branch. Now I just have to figure out why AirLoopNum = 0 for the first mixer in develop and not this branch.

atmixersumforoa

rraustad:
Oh, and the first ATMixer has a flow rate of 0.057248 m3/s, so 0.14933 + 0.05725 = 0.20658. The answer is correct now, not sure why yet.

rraustad:
@mjwitte such an easy answer and it took me 2 days to find it. In this branch I added the call to SimATMixer for set point based control. That's working now. But also working is the inclusion of that TU's flow in the summation of the DOAS total air loop flow.

    AirLoopHVAC:UnitarySystem,
      SPACE1-1 PTAC,           !- Name

Since the SP code was not calling the single duct model, the SingleDuct::SysATMixer(1).AirLoopNum = 0 never got updated

SingleDuct line 5722 (i.e., this never got called for ATMixer 1):

    // Keep trying until we find it, the airloopnum, that is
    if (this->OneTimeInitFlag2) {
        this->AirLoopNum = DataZoneEquipment::ZoneEquipConfig(DataDefineEquip::AirDistUnit(this->ADUNum).ZoneEqNum)
                               .InletNodeAirLoopNum(this->CtrlZoneInNodeIndex);
        DataDefineEquip::AirDistUnit(this->ADUNum).AirLoopNum = this->AirLoopNum;
        if (this->AirLoopNum > 0) {
            this->OneTimeInitFlag2 = false;
        }
    }


and that flow rate was skipped in the figure from yesterday at line 1211 for the total  flow. This branch corrects that and now " if (AirLoopNum=1  ==  SingleDuct::SysATMixer(1).AirLoopNum=1) is true ". Before it was " if (1 == 0) and it got skipped.

This looks good to go now.

mjwitte (commented):
Merged in develop and tested defect files. For a bad control zone name, no longer crashes, but throws error:
```
   ** Severe  ** AirLoopHVAC:UnitarySystem: DXAC HEAT PUMP 1
   **   ~~~   ** When Control Type = LOAD
   **   ~~~   ** Controlling Zone or Thermostat Location must be a valid zone name, zone name = XYZ
   **  Fatal  ** GetUnitarySystemInput: Errors found in getting AirLoopHVAC:UnitarySystem input. Preceding condition(s) causes termination.
```
For SP control with ATMixer, with this branch, Space1-1 controls well, but if you report the System Node Setpoint Temperature for SPACE1-1 Supply Inlet, it matches the SPACE1-1 Supply Inlet temperature for the 01/21 heating day, but it's all over the place on the cooling days. e.g. 07/21 12:00, SPACE1-1 Supply Inlet temperature is 11.9, but the setpoint temperature is -65.

src/EnergyPlus/HVACUnitarySystem.cc
mjwitte: @rraustad One more question - why do we need to track last mode?

rraustad:
SPACE1-1 Supply Inlet is the heating coil outlet node so in summer the set point goes negative. I had to look too. From this plot the control doesn't look ideal but it is controlling as expected. Space1-1 is set point control and Space2-1 is load control and they look the same so the AT Mixer must have something to do with zone temp control. Column 1, row 2 and 3 are the individual coil control temp and SP. They look correct. Column 2, row 2 is the zone predicted load (green line) with cooling (orange) a mirror opposite due to the sign. Heating (yellow) is right on top of the green line except at x-axis time 250-ish where the heating capacity is limited. So merge away!

unitarysys_spcontrol

mjwitte (commented):
Duh - right - separate coil setpoints, not a single leaving unitary setpoint. OK, so the control looks good.

Myoldmopar:
This pull request reads like a good book I can't put down.  The suspense of whether each commit would finally address it, and Detective Raustad on the case tracking down the culprits.  This looks great.  I'm happy with it.. @mjwitte merge when you are ready.

mjwitte:
@rraustad @Myoldmopar Does #6799 make this moot? Are these same fixes present in #6799 or do we need to retest the defect files with the refactored code?

rraustad:
LastMode is used to track the last operating mode when the system goes into no load operation. For the VS/MS coils the lowest speed was used (heating or cooling) to determine the air flow rate for no load. Maybe that tracking variable isn't needed anymore since the idleMassFlowRate is becoming more of the driver for flow during no load, but that's where it came from. As far as these issue being moot, no, these are issues that need to go into the refactored code. And that refactored code wasn't available yet to make those changes. So these still need to go in and then I need to move them over to #6799.

mjwitte:
@Myoldmopar  . . . suspenseful music . . . to merge or not to merge, that is the question . . . what will the HVAC monster do?!!!
image

Issue #6634, opened 2018-03-29
Moisture sorption curve slope equation is incorrrect


Nigusse:
Issue overview
--------------
Please change this line to a description of the defect, with useful supporting information including how the issue was found, conditions that trigger this defect, workarounds for users of the current version, and a clear description of how to validate that a fix is complete for this defect.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus V8.8 

While working on another project I came across the following issues in the EMPD model:
- [du_dRH](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MoistureBalanceEMPDManager.cc#L560-L561) calculation equation has duplicate "material.MoistCCoeff" term.
- Suspicious how the vapor density at the surface and deep layers are calculated during init [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MoistureBalanceEMPDManager.cc#L394-L396). shouldn't we use surface temperature instead zone air temperature?  

Documentation issues:
- Equations 3.50 and 3.51 in engineering reference (V8.8) should have temperature (T) in the denominator
- Equation 3.48 the moisture penetration depth should have subscript 1
- In Engineering Reference section 3.4.3 Nomenclature universal gas constant for water is defined as 452.61, instead of 461.52. Correctly defined earlier section.

shorowit:
FYI @jdwoods21 @nmerket

Myoldmopar:
Closed via #6759

PR #6759, opened 2018-05-24
EMPD model moisture sorption curve slope equation is incorrrect

Nigusse:
Issue #6634 

Pull request overview
---------------------
EMPD model du_dRH calculation equation has duplicate "material.MoistCCoeff" term. This is defect. Removed the duplicate coefficient and also corrected equations in engineering reference. There are two EMPD model example files (1ZoneUncontrolledFourAlgorithms.idf and EMPD5ZoneWaterCooled_HighRHControl.idf) that affected by this fix.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)

### DIFF
--- a/doc/engineering-reference/src/surface-heat-balance-manager-processes/effective-moisture-penetration-depth-empd.tex
+++ b/doc/engineering-reference/src/surface-heat-balance-manager-processes/effective-moisture-penetration-depth-empd.tex
@@ -97,7 +97,7 @@ R_{MT,2} = \frac {d_{EMPD,1} + d_{EMPD,2}} {2 \bar R_{water} T \delta_{perm} }
 These moisture fluxes are used to update the moisture content (or equivalent RH, \(\phi\) of each layer:
 
 \begin{equation}
-\phi_{1}(t) = \phi_{1}(t-\delta t) - \frac { (j_{zone} + j_{2}) \cdot TimeStep} {\rho_{matl} d_{EMPD,2} \frac {du}{d\phi}}
+\phi_{1}(t) = \phi_{1}(t-\delta t) - \frac { (j_{zone} + j_{2}) \cdot TimeStep} {\rho_{matl} d_{EMPD,1} \frac {du}{d\phi}}
 \end{equation}
 
 \begin{equation}
@@ -115,11 +115,11 @@ $\frac {du}{d\phi}$ = slope of moisture curve (change in moisture content / chan
 The equivalent RH and equivalent vapor density are related with:
 
 \begin{equation}
-\rho_{v,1} = \frac {\phi_{1} P_{sat}(T)} {\bar R_{water}}
+\rho_{v,1} = \frac {\phi_{1} P_{sat}(T)} {\bar R_{water} T}
 \end{equation}
 
 \begin{equation}
-\rho_{v,2} = \frac {\phi_{2} P_{sat}(T)} {\bar R_{water}}
+\rho_{v,2} = \frac {\phi_{2} P_{sat}(T)} {\bar R_{water} T}
 \end{equation}
 
 The zone flux is used, along with the area of the particular surface, to update the zone humidity ratio, as discussed in \ref{moisture-predictor-corrector}.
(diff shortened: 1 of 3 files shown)


Nigusse:
Diffs are expected in example files (EMPD5ZoneWaterCooled_HighRHControl) with EMPD model. The documentation change local build is successful. Failures in MacOS-10.9-clang CI machine is not expected.

tst/EnergyPlus/unit/MoistureBalanceEMPD.unit.cc
```diff
@@ -282,3 +284,117 @@ TEST_F(EnergyPlusFixture, EMPDRcoating)
...
+    if (std::abs(mass_flux_surf_deep_max) < std::abs(mass_flux_surf_deep_result)) {
+        mass_flux_surf_deep_result = mass_flux_surf_deep_max;
+    }
+
+    // Calculate and verify it against the results determined above
+    MoistureBalanceEMPDManager::CalcMoistureBalanceEMPD(1, Taver, Taver, Tsat);
+    auto const &report_vars = MoistureBalanceEMPDManager::EMPDReportVars(surfNum);
+    EXPECT_DOUBLE_EQ(mass_flux_surf_deep_result, report_vars.mass_flux_deep);
+
+}
```
Myoldmopar, line 400: This is a nice unit test!

doc/engineering-reference/src/surface-heat-balance-manager-processes/effective-moisture-penetration-depth-empd.tex
```diff
@@ -139,7 +139,7 @@ \subsection{EMPD Nomenclature}\label{empd-nomenclature}
   A  &= \text{surface area} \RB{m^{2}} \\
   d &= \text{penetration depth} \RB{m} \\
   j &= \text{water mass flux} \RB{\frac{kg}{m^{2} \cdot s}} \\
-  \bar R_{water} &= \text{universal gas constant for water,} \; 452.61 \RB{\frac{J}{kg \cdot K}}   \\
+  \bar R_{water} &= \text{universal gas constant for water,} \; 461.52 \RB{\frac{J}{kg \cdot K}}   \\
```
Myoldmopar, line 142: Any idea where this 452 value came from?  I believe 461 is correct, I'm just wanting to make sure I'm not missing something where the value is scaled or something else.

Myoldmopar (commented):
I don't have any problem with the code as it is, but I don't want to mark it approved just yet, so I'll leave this review as a generic comment.

I want @nmerket to chime in on the actual fix, and then I had just a couple other comments we can discuss as needed.

src/EnergyPlus/MoistureBalanceEMPDManager.cc
```diff
@@ -558,7 +558,7 @@ namespace MoistureBalanceEMPDManager {
 
         // Calculate slope of moisture sorption curve at current RH. [kg/kg-RH]
         dU_dRH = material.MoistACoeff * material.MoistBCoeff * pow(RHaver, material.MoistBCoeff - 1) +
-                 material.MoistCCoeff * material.MoistCCoeff * material.MoistDCoeff * pow(RHaver, material.MoistDCoeff - 1);
```
Myoldmopar, line 561: I had to dig for the commit where this was added: https://github.com/NREL/EnergyPlus/commit/de255ee9b8dfa1050c7e99e702c56b82bd199ce3#diff-691a3b35c3d65b682c8cf7acda7ffac9R485

It was a commit by @nmerket.  I tag him here, not to code shame him (maybe a little...we gotta keep each other humble), but instead to make him aware of the change for two reasons:
 - Could you verify the change is correct, that the extra C coefficient **should** be removed?
 - If it is the correct fix, then the original commit message grabs my attention: "porting @jdwoods21 empd code from 7.1".  You might want to make sure the original code this was adapted from is also corrected with this change (if it's still being used at all).

nmerket: It looks correct to me. I'm going to have @jdwoods21 look at it as well. I think he should be able to make the final say.

nmerket:
@jdwoods21 Can you look at the changes in this pull request with some fixes to the EMPD model in EnergyPlus? It looks correct to me, but I'd like your stamp of approval.

src/EnergyPlus/MoistureBalanceEMPDManager.cc
jdwoods21: Yes, there should only be one material.MoistCCoeff term in that equation, not two. One should be deleted.

The original code was in Matlab, and doesn't have this issue. But I transported that to EnergyPlus Fortran a while back...it may have been introduced then.

nmerket: Thanks, we're good here then. Sorry about the bug. I deserve the code shame. I don't think anybody is using the old modified E+ 7.1 code anymore, so no need to backport.

nrel-bot-3:

nrel-bot-3:

Myoldmopar:
@nmerket @jdwoods21 thanks for looking this over.  It's good to go in.  Merging.

PR #6762, opened 2018-05-29
Correct terminal unit heating coil inlet air conditions used for sizing and/or reporting.

rraustad:
Pull request overview
---------------------
Coil Sizing Summary table shows incorrect coil inlet air conditions for VAV boxes using non-water heating coils.

 - [x] VAV coil entering air conditions incorrect for non-water coils.
 - [x] PIU calculates entering air temperature differently than used for coil capacity sizing
 - [x] Inch Pound conversion converts -999 and -99999 to large negative numbers which can be confusing to users
 - [x] Add zone name to Coil Sizing Details table (Yahoo groups user was complaining about not knowing which zone was associated with zone coils)

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
@Myoldmopar @mjwitte am I doing something wrong with the clang.format issue. Every file I touch seems to reformat such that it is hard to see what changed. For example in my last [commit](https://github.com/NREL/EnergyPlus/pull/6762/commits/cb5a71fda529f4a9bd670d665c5e8a0b8619b2f2) I changed 1 line (the last line). Just checking before I move forward.

rraustad:
First step here was to retain the 999's in the Coil Sizing Summary table. This should affect all files (5 or 6) using InchPound conversion.

6761_step1_retain999incoilsummaryreporttable

rraustad:
The second step was to include the zone name for zone equipment. Didn't work as expected. MAC didn't like the implementation for some reason.
Added these 4 lines just after the else if for curZoneEqNum > 0 in ReportCoilSelection::setCoilCoolingCapacity and again in ReportCoilSelection::setCoilHeatingCapacity.
Each of these lines seems harmless.

    } else if (curZoneEqNum > 0 && allocated(DataSizing::FinalZoneSizing)) {
        c->zoneNum.resize(1);
        c->zoneName.resize(1);
        c->zoneNum[0] = curZoneEqNum;
        c->zoneName[0] = DataSizing::FinalZoneSizing(curZoneEqNum).ZoneName;

Third step was to correct zone coil inlet conditions for non-water coils. This figure shows correct zone coil inlet conditions for the single duct TU's. Also shown is the zone name in the Coil Sizing Details table. I would like to get this working if possible.

6761_step3_correctzonecoilinletconditions_showzonename

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -15738,7 +15973,7 @@ namespace OutputReportTabular {
 
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
         //    na
-        if (unitConvIndex == 0) {
+        if (unitConvIndex == 0 || SIvalue == -999.0 || SIvalue == -99999.0) { // don't convert unknown data to IP
```
rraustad, line 15976: Here is the only code change in this module. It reports back the SIValue if equal to -999 so that the table reports look the same (the 999's that is) regardless of units.

src/EnergyPlus/ReportCoilSelection.cc
```diff
@@ -1233,24 +1249,28 @@ void ReportCoilSelection::setCoilCoolingCapacity(
...
             if (DataAirSystems::PrimaryAirSystem(curSysNum).NumOACoolCoils > 0) { // there is precooling of the OA stream
                 c->oaPretreated = true;
             }
         }
 
     } else if (curZoneEqNum > 0 && allocated(DataSizing::FinalZoneSizing)) {
+        c->zoneNum.resize(1);
+        c->zoneName.resize(1);
+        if (allocated(DataZoneEquipment::ZoneEquipConfig)) c->zoneNum[0] = DataZoneEquipment::ZoneEquipConfig(curZoneEqNum).ActualZoneNum;
+        if (allocated(DataZoneEquipment::ZoneEquipConfig)) c->zoneName[0] = DataZoneEquipment::ZoneEquipConfig(curZoneEqNum).ZoneName;
```
rraustad, line 1273: First change added to allow zone name to be reported in Coil Sizing Details table.

src/EnergyPlus/ReportCoilSelection.cc
```diff
@@ -1422,32 +1442,36 @@ void ReportCoilSelection::setCoilHeatingCapacity(
...
             if (DataAirSystems::PrimaryAirSystem(curSysNum).NumOAHeatCoils > 0) { // there is preHeating of the OA stream
                 c->oaPretreated = true;
             }
         }
 
     } else if (curZoneEqNum > 0 && allocated(DataSizing::FinalZoneSizing)) {
+        c->zoneNum.resize(1);
+        c->zoneName.resize(1);
+        if (allocated(DataZoneEquipment::ZoneEquipConfig)) c->zoneNum[0] = DataZoneEquipment::ZoneEquipConfig(curZoneEqNum).ActualZoneNum;
+        if (allocated(DataZoneEquipment::ZoneEquipConfig)) c->zoneName[0] = DataZoneEquipment::ZoneEquipConfig(curZoneEqNum).ZoneName;
```
rraustad, line 1474: Second change added to allow zone name to be reported in Coil Sizing Details table.

src/EnergyPlus/ReportCoilSelection.cc
```diff
@@ -1478,14 +1502,45 @@ void ReportCoilSelection::setCoilHeatingCapacity(
         }
         // coil entering conditions depend on the type of zone equipment involved
         // set typeof_Coil integer
-        if (DataSizing::TermUnitIU) { // an unpowered induction terminal unit
-            // should be picked up by HeatingWaterDesAirInletTempSizing and HeatingWaterDesAirInletHumRatSizing
-
```
rraustad, line 1499: This next section is new to update coil inlet temps if they were not set elsewhere. The electric or gas heating coil in a terminal unit is an example.

Myoldmopar:
BTW, we restarted Marvin. Hopefully Mac builds will come back clean again.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1347,8 +1363,13 @@ namespace ReportSizingManager {
                 } else if (SizingType == HeatingWaterDesAirInletTempSizing) {
                     if (TermUnitPIU && (CurTermUnitSizingNum > 0)) {
                         MinFlowFrac = TermUnitSizing(CurTermUnitSizingNum).MinFlowFrac;
-                        AutosizeDes = TermUnitFinalZoneSizing(CurTermUnitSizingNum).DesHeatCoilInTempTU * MinFlowFrac +
-                                      FinalZoneSizing(CurZoneEqNum).ZoneTempAtHeatPeak * (1.0 - MinFlowFrac);
```
rraustad, line 1368: Expanded this section to account for differences in PIU unit if induced air is present.

tst/EnergyPlus/unit/ReportCoilSelection.unit.cc
```diff
@@ -426,4 +451,76 @@ TEST_F(EnergyPlusFixture, ReportCoilSelection_ZoneEqCoil)
 
     // Exercise report writing again
     coilSelectionReportObj->finishCoilSummaryReportTable();
+
```
rraustad, line 454: Test these changes in unit test.

rraustad:
Reviewed the files with diff's and results are as expected given these changes. TU coil inlet conditions are now correct, 999's are retained in tables were data is unavailable (applies to files using InchPound), and zone name now appears in Coil Sizing Details table.

rraustad:
The Mac regression is stuck and I can't delete and force rebuild...so just click merge.

rraustad:
The Mac finally responded with diff's similar to other operating systems. This is ready for review.

nrel-bot-3:

src/EnergyPlus/OutputReportTabular.cc
Myoldmopar: I hesitate here.  For a couple reasons.
- I realize it is really unlikely that an SIValue will come in with a value of -999 or -99999 that we actually want converted.  I know we use those values as flags in numerous places, this isn't something new you've added.  So I am OK with it...
- But don't you really need to check with a tolerance?  When the memory is initialized, will some values get a value of -999.000000001?  Checking floating point equality seems risky.

With those in mind, is there _any_ other way to identify that the value doesn't need to be converted besides checking the numeric value itself?  Have the caller pass a flag?  If we can't figure out a different option, then so be it, but I think it's at least worth discussing.

Myoldmopar:
I pulled develop in and resolved the minor conflicts, but I haven't pushed.  I ran the tests locally and everything is good.  So if we decide no more changes are necessary here then I'll just push my changes up and merge the PR without wasting a CI cycle.

src/EnergyPlus/OutputReportTabular.cc
rraustad: The only think I can this of is abs(SIvalue+999.0) >= 0 and < floatingPointTol

rraustad: Don't need the >= 0 part.

Myoldmopar:
OK, I've thought it over, and I think this can go in as-is.  It's obviously working, so we don't need to rigor on it.  I'll push up my conflict resolved branch and merge it in.  Thanks @rraustad.

Issue #6539, opened 2018-03-16
FisherPedersenCeilingDiffuser* convection coefficients are based on zone outlet temp but use zone supply temp as the reference temperature


mjwitte:
Issue overview
--------------
Per the Engineering Reference, the CeilingDiffuser convection algorithm uses the zone outlet temperature.

> The ceiling diffuser algorithm is based on empirical correlations developed by Fisher and Pedersen (1997). The correlation was reformulated to use the room outlet temperature as the reference temperature.

**The zone-level CeilingDiffuser algorithm is correct - it uses the zone air temperature as the reference temperature.**

*The problem is with the surface-level FisherPedersenCeilingDiffuser\* algorithms. These are not used by default by the AdaptiveConvection option. They are only used if the user inputs specifies them at the surface-level or as an override to the defaults in AdaptiveConvection.*

The [current code](https://github.com/NREL/EnergyPlus/blob/501af898146625a9612782389de781aaf94b4139/src/EnergyPlus/ConvectionCoefficients.cc#L5534-L5579), sets the reference temperature to the zone supply air temperature.
```
	EvaluateIntHcModels(
		} else if ( SELECT_CASE_var == HcInt_FisherPedersenCeilDiffuserFloor ) {
				Surface( SurfNum ).TAirRef = ZoneSupplyAirTemp;
```

Looking back in the archive, there were two functions, one with the [original coefficients](https://github.com/NREL/EnergyPlusArchive/blame/8fd5ad8428e585860a8bd3ec5b18346dfd643f48/SourceCode/HeatBalanceConvectionCoeffs.f90#L2312) for supply air temperature, and one with the [modified coefficients](https://github.com/NREL/EnergyPlusArchive/blame/8fd5ad8428e585860a8bd3ec5b18346dfd643f48/SourceCode/HeatBalanceConvectionCoeffs.f90#L2219) for zone outlet temperature.
```
SUBROUTINE CalcCeilingDiffuserIntConvCoeff(ZoneNum)
          ! The correlations shown below differ from (and are less accurate than) those shown
          ! in the reference above (Fisher 1997).  They have been reformulated with an outlet
          ! temperature reference in order to accomodate the structure of the EnergyPlus code.

! CalcCeilingDiffuserInletCorr should replace CalcCeilingDiffuser (above), if ZoneTempPredictorCorrector can
! ever be made to work correctly with the inlet air temperature.
[...]

PR #6764, opened 2018-05-29
Correction of Reference Temperature for Fisher-Pedersen Convection Models when Specified for Surfaces

RKStrand:
Pull request overview
---------------------
Corrected the reference temperatures for the Fisher-Pedersen convection models when specified at the surface level.  The code was using the supply air temperature as the reference temperature for these models.  However, the code for the calculation of the coefficients were the equations based on the zone air temperature.  This was corrected by using the zone mean air temperature as the reference temperature to match the equations.

*The problem is with the surface-level FisherPedersenCeilingDiffuser\* algorithms. These are not used by default by the AdaptiveConvection option. They are only used if the user inputs specifies them at the surface-level or as an override to the defaults in AdaptiveConvection.*

*The zone-level CeilingDiffuser algorithm was correct - it has always used the zone air temperature as the reference temperature.*

### Work Checklist
 - [ x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue #6539 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes: n/a


mjwitte:
@RKStrand Functionally, the code changes look fine, but there are [formatting errors](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-138f843027936f6dc65d2004779fb69740056ffb-x86_64-Linux-Ubuntu-14.04-custom_check.html).

RKStrand:
@mjwitte Did the format changes I made trigger a new run of the test suites?  I committed the changes but I'm not seeing that it actually ran again.

mjwitte:
@RKStrand There a no new commits in github on this branch since May 29. Maybe you must need to push the commit?

RKStrand:
@mjwitte That's weird.  My desktop GitHub app usually does that when I "sync" things.  Not sure why it didn't do it this time.  But I just did a push, and then updated from develop, and then synced again.  If that doesn't trigger something, I don't know what will...

nrel-bot-3:
@RKStrand @lgentile it has been 14 days since this pull request was last updated.

RKStrand:
Mike:

Did this finally get pushed properly?

Rick

mjwitte:
@RKStrand I think it's all here.  It's a little confusing because the commits are dated when you commit them locally not when they get pushed into the branch on github.  CI is all green now, so I think it's good.

mjwitte:
@RKStrand Added some defect files in EnergyPlusDevSupport\DefectFiles based on example file Convection.idf.  One uses the zone-level ceilingdiffuser model, and the other uses the surface-level.  Before this fix, the adjacent air temps and surface convection heat transfer rate are very different.  With this fix, they are very close, but not identical. Can you explain why they wouldn't be exactly the same?

mjwitte:
And I merged in current develop to this branch for another round of CI.

RKStrand:
@mjwitte Well, I looked into this and found something very interesting in the code.  In subroutine CalcCeilingDiffuserIntConvCoeff for the Zone Level calculations, when AirDensity is calculated (ACH is then based on this density), it assumes a temperature of 0.0.  When AirDensity is calculated in EvaluateIntHcModels for the Surface Level calculations (same equation for ACH), it is based on the node temperature.  That's a bug and it explains why the two are slightly off--different air densities will result in slightly different ACH values and thus slightly different results.  Want to log that and point it to me?

mjwitte:
@RKStrand Can we just fix that here in this pull request?

RKStrand:
@mjwitte Ok, I'll put in the fix in this pull request and will push it up when it's ready.

RKStrand:
@mjwitte Ok, I need to take back what I wrote last night.  The zero temperature assumption is only at the beginning of the environment since it doesn't have a zone temperature yet.  Once BeginEnvrnFlag is false, it uses the same temperature from the same place.  So, what I wrote last night is not correct.
Which gets us back to your question--I wonder if the issue is the updating of the zone node temperature, when that happens relative to when these calculations are being made.  It is very possible that one happens after an additional update of the zone temperature that this would explain the difference.  So, I have a feeling that these will be close but not exactly the same as a result of that.

mjwitte:
That's a possible explanation, however, I would expect that all convection coefficients get calculated at the same point in the simulation regardless of whether it was set by zone or by surface.  If that's not the case, then maybe one of them should move, but that's a bigger issue.  Could we reproduce the same test using a different convection model?  Say, set to TARP in both?  But since that's the default it may not show anything interesting. Maybe I missed a class of surfaces in the one that uses SurfaceProperty:ConvectionCoefficients:MultipleSurface.

mjwitte (approved):
@RKStrand My bad - the zone-level test file still had a different outside convection algorithm for zone 2, so that was driving slightly different surface temps.  With that fixed, the two files produce very nearly identical results.  There's still some differences out after 7 or 8 figures, but that's ok. I've merged in develop for one last round of CI tests, then this will be ready to drop in.

RKStrand:
@mjwitte Awesome--glad this resolved itself.  Hopefully this is good to go now.  Thanks!

Issue #5675, opened 2016-05-20
Site:WaterMainsTemperature Default


matthew-larson:
## Issue overview

Based on the Engineering Reference, 'If there is no Site:WaterMainsTemperature object in the input file, a default constant value of 10 C is assumed.'  I have found that the actual calculated profile for the water mains temperature is significantly higher for the majority of climates, resulting in an overestimate of domestic hot water use if no Site:WaterMainsTemperature object is in the input file.  The graph below shows the water mains temperature profile for the Chicago-O’Hare TMY2 weather file, which many of the Reference Buildings and Example Files are based on.  I recommend adjusting the default value to somewhere closer to 12 C, or the average water mains temperature based on the profile for the Chicago-O’Hare TMY2 weather file, when no Site:WaterMainsTemperature object is in the input file to better reflect the majority of climates.

![water_mains](https://cloud.githubusercontent.com/assets/17144687/15436142/9d2e7478-1e7d-11e6-93a6-43747ab7a1a6.png)

shorowit:
EnergyPlus implements a good correlation method that provides varying mains temperatures throughout the year. Unfortunately, it requires the user to do some calculations based on the weather file in order to provide the appropriate inputs to EnergyPlus. It'd be nice if there was an "enhanced correlation" method that didn't require any inputs and operated based on the supplied weather file. That would be a much better default.

On a related note, if there is no Site:WaterMainsTemperature object, no warning is provided to the user. (As noted, a constant temperature is assumed, which is a poor assumption.) It'd be nice if a warning is provided as is done for other similar situations like this one:

> Warning: GetHTSurfaceData: Surfaces with interface to Ground found but no "Ground Temperatures" were input.
>  Found first in surface=SURFACE 1
>  Defaults, constant throughout the year of (18.0) will be used.

mjwitte:
It should be possible to make the default automatically get the values from the weather file.  Site:GroundTemperature:FCfactorMethod already does that.

Nigusse:
There are two separate issues here:
(1) Add autosizing option for the two numeric input fields by processing weather file:
-   N1 , \field Annual Average Outdoor Air Temperature
-   N2 ; \field Maximum Difference In Monthly Average Outdoor Air Temperatures
- the Calculation Method input field should be set to 'Correlation' when these input fields are set to autosize

(2) When 'If there is no Site:WaterMainsTemperature object in the input file, a default constant value of 10 C is assumed. There are two option how to address this issue.
[...]

PR #6771, opened 2018-06-06
Added New Water Mains Temperature Calculation Method  

Nigusse:
(#5675)

Pull request overview
---------------------
This is a new feature (enhancement). Adds new Water Mains Temperature calculation method "CorrelationFromWeatherFile" as choice. The annual average OA temperature and the OA monthly-daily average OA temperature maximum difference are calculated internally by energyplus from the weather or stat file. This choices reduces the effort from users in pre-processing weather files to get the values for these two input fields, in Site:WaterMainsTemperature object. Otherwise, uses the existing correlation to estimate the Water Mains Temperature.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:


Nigusse:
Sample results calculated for Chicago Weather using the new method added.

Nigusse:
@mjwitte revised EIO output per your suggestion:
! '<'Site Water Mains Temperature Object Parameters'>', Calculation Method{}, Annual Average Outdoor Air Temperature{C}, Maximum Difference In Monthly Average Outdoor Air Temperatures{deltaC}, Water Mains Temperature Schedule Name
Site Water Mains Temperature Object Parameters, CorrelationFromWeatherFile,9.98799,28.7813,

Nigusse:
@mjwitte If there are no Site Water Mains Temperature object in the idf, I will disable reporting Site Water Mains Temperature Information. This will preclude reporting Site Water Mains Temperature Information header to eio file and removes the EIO diffs observed in.

mjwitte:
@Nigusse I understand the diffs issue, so this pass is fine to document fewer diffs, but ultimately we want the eio output with or without the object.  Just like many other eio entries like heat balance, convection, etc.

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
```diff
@@ -1698,23 +1698,23 @@ \subsubsection{Inputs}\label{inputs-21-003}
 
 \paragraph{Field: Calculation Method}\label{field-calculation-method}
 
-This field selects the calculation method and must have the keyword Schedule or Correlation.
+This field selects the calculation method and must have the keyword Schedule, Correlation or CorrelationFromWeatherFile. If calculation method is CorrelationFromWeatherFile, the two numeric input fields below are ignored. Instead, Instead, EnergyPlus calculates them from weather file.
```
Myoldmopar: Typo - instead repeated

doc/output-details-and-examples/src/output-files/eplusout-eio.tex
```diff
@@ -1,6 +1,6 @@
...
+This file contains some standard and some optional ``reports''. It is intended to be a somewhat intelligent report of input conditions when they don't fit well in other places or when they aren't substantial enough for their own ``file''. (e.g. \textbf{eplusout.bnd})~ Contents of the file are somewhat documented in various places in the \href{file:///E:/Docs4PDFs/InputOutputReference.pdf}{Input Output Reference document} -- as results of objects. This file or portions of it can be easily imported into spreadsheet programs and more analysis done there. Contents of this file include construction details, location information, site water mains temperature information, ``environment'' information, number of ``warmup'' days required in each environment.
```
Myoldmopar: Can you clean out this weird link that was left in here?  I know you didn't add it, but it would be great to clean that out.

tst/EnergyPlus/unit/WeatherManager.unit.cc
```diff
@@ -342,3 +343,149 @@ TEST_F(EnergyPlusFixture, UnderwaterBoundaryConditionConvectionCoefficients)
...
+        "   CorrelationFromWeatherFile,  !- Calculation Method",
+        "   ,                            !- Temperature Schedule Name",
+        "   9.99,                        !- Annual Average Outdoor Air Temperature {C}",
+        "  28.78;                        !- Maximum Difference In Monthly Average Outdoor Air Temperatures {deltaC}",
+    });
+
+    ASSERT_TRUE(process_idf(idf_objects));
+
+    bool foundErrors(false);
+    WeatherManager::GetWaterMainsTemperatures(foundErrors);
```
Myoldmopar, line 365: So does this unit test actually go try to read an epw file?

src/EnergyPlus/WeatherManager.cc
```diff
@@ -174,6 +178,8 @@ namespace WeatherManager {
 
     bool Debugout(false);
 
+    // Real64 CalcWaterMainsTempCurrent( 0 );
```
Myoldmopar, line 181: This should be removed

src/EnergyPlus/WeatherManager.cc
```diff
@@ -9779,6 +10323,254 @@ namespace WeatherManager {
...
+                    gio::open(epwFile, DataStringGlobals::inputWeatherFileName, flags);
+                    readStat = flags.ios();
+                }
+                if (readStat != 0) {
+                    ShowSevereError("CalcAnnualAndMonthlyDryBulbTemp: Could not open file " + DataStringGlobals::inputWeatherFileName +
+                                    " for input (read).");
+                    ShowContinueError("Water Mains Temperature will be set to a fixed deafult value of 10 C.");
+                    return;
+                }
+                for (i = 1; i <= 8; ++i) { // Headers
```
Myoldmopar: This is kinda not what I wanted to see, another epw parsing function that we have to maintain.  There are at least two other ones in E+ already.  Ugh.  I'm not sure about this right now, I'll try to take a look at some other spots in the code where this is done and maybe we can collapse them down.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -9779,6 +10323,254 @@ namespace WeatherManager {
...
+                    AnnualDailyAverageDryBulbTempSum += MonthlyAverageDryBulbTemp(i) * EndDayOfMonth(i);
+                    MonthlyDailyDryBulbMin = min(MonthlyDailyDryBulbMin, MonthlyAverageDryBulbTemp(i));
+                    MonthlyDailyDryBulbMax = max(MonthlyDailyDryBulbMax, MonthlyAverageDryBulbTemp(i));
+                    AnnualNumberOfDays += EndDayOfMonth(i);
+                }
+                OADryBulbAverage.AnnualAvgOADryBulbTemp = AnnualDailyAverageDryBulbTempSum / AnnualNumberOfDays;
+                OADryBulbAverage.MonthlyAvgOADryBulbTempMaxDiff = MonthlyDailyDryBulbMax - MonthlyDailyDryBulbMin;
+                OADryBulbAverage.MonthlyDailyAverageDryBulbTemp = MonthlyAverageDryBulbTemp;
+                OADryBulbAverage.OADryBulbWeatherDataProcessed = true;
+            } else if (epwFileExists) {
```
Myoldmopar, line 10437: How well does this play with the SimulationControl object?  What about users who try to do this method for runs with "Run Weather File Simulations" turned off.  I see error messages below but they are commented out?

Myoldmopar (changes requested):
I've identified a few minor things to clean up, plus one big thing, and that's that we really don't need another epw parser in our code.  There should be a better way to do it reusing some of the other epw code somewhere.  I'm going to have to noodle on it for a bit.  Feel free to update with my other comments in the meantime

nrel-bot-3:

src/EnergyPlus/WeatherManager.cc
Nigusse: This method will not work unless there is a stat or weather file. We can provide an error message that tells the user that this method cannot be used without one of these files. I will update the error message.

Nigusse: @Myoldmopar If there is no weather or stat file, then severe error message will be thrown and the simulation continues with the default value.

Nigusse:
@Myoldmopar Merged in develop, resolved conflicts, added error message when there is no stat or weather file, and build and run the unit tests locally and all passed. I will push the branch shortly.

Nigusse:
Site Water Mains Temperature Information header is reported to eio file even when there is no Site Water Mains Temperature object in the idf. This caused the EIO file diffs. Disabled header reporting when there is no Site:WaterMainsTemperature object in the idf.

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
Nigusse: Removed duplicate.

doc/output-details-and-examples/src/output-files/eplusout-eio.tex
Nigusse: Removed the link.

tst/EnergyPlus/unit/WeatherManager.unit.cc
Nigusse: No, it does not. It skips weather file processing part.

src/EnergyPlus/WeatherManager.cc
Nigusse: Removed.

mjwitte:
I realize it causes diffs, but we want that to be reported even if there is no water mains object in the idf. That way the default assumptions are exposed/documented for the user. If you want to disable this if there are no water heaters or water use objects (thus making mains temperature irrelevant), I might be open to that.

Nigusse:
@mjwitte If there is no water mains object in the idf, then there is nothing to report except the header file and the header file contains the parameters of water mains object only. The default value that we have is a fixed water mains temperature (10C), which is not related to the parameters in the header.

mjwitte:
This isn't quite what I expected. The whole point here was to use a better default that 10.0C when there is no Site:WaterMainsTemperature object. Let's discuss this and agree before making any more code changes. @Myoldmopar @Nigusse 

Here's what I was expecting from this:

1. If Site:WaterMainsTemperature is present, then use it and report to eio.

2. If Site:WaterMainsTemperature is not present, then use the default, which is CorrelationFromWeatherFile. This is similar behavior to other top-level inputs like HeatBalanceAlgorithm, SurfaceConvectionAlgorithm:Inside, etc.

3. However, this comes with the added dependency on an epw ( or stat?) file being present, so there has to be a fallback if it's not. I guess that's where the 10C still has a place in life, but it shouldn't happen without throwing a warning so the user is aware of it (it looks like there is a warning but it only happens if the user requests CorrelationFromWeatherFile, so if CorrelationFromWeatherFile is the default even without an object then this should work).  

4. When the fixed 10C is used, the eio output should document it somehow.  Perhaps Method=Fixed, and an extra field at the end for a constant value?  But the eio outputs should be consistent with the input choices, but there aren't inputs available to make that happen . . . . so not sure what to suggest here.

5. If the method is CorrelationFromWeatherFile, then the values for "Annual Average Outdoor Air Temperature"  and "Maximum Difference In Monthly Average Outdoor Air Temperatures" are obtained from scanning the weather file. These would override any user inputs for these fields and would also be reported to the eio (looks like current code will do that).  But the schedule name should be written as "NA" if the method is not "Scheduled".

6. The eio output should be written anytime that water mains temperature is relevant (or simply always).

7. Also, just noticed that other similar methods, e.g. Site:GroundTemperature:Undisturbed:KusudaAchenbach has an input for "Phase Shift of Minimum Surface Temperature". Looks like this water mains method uses the same sine wave regardless of the timing of the weather with a check for northern/southern latitude (not mentioned in the engineering reference) to make sure it's not completely backwards for southern hemisphere.  Probably good enough as a default.

Comments - fire away.

mjwitte:
So, nobody has commented on my essay [above](https://github.com/NREL/EnergyPlus/pull/6771#issuecomment-408123867). The really big question is #2 - If there is no Site:WaterMainTemperature object in the idf, do we default to CorrelationFromWeatherFile?  And if so, if no weather file is present, then use 10C with a warning?  That's my vote.  @shorowit @Myoldmopar @matthew-larson

shorowit:
I had no idea you could run EnergyPlus without a weather file. What is that used for?

I'm good with @mjwitte's recommendation, though it will be a change for any users that don't have the Site:WaterMainTemperature object in their current IDFs. Should version translation maintain the same results?

mjwitte:
If you're running just sizing period design days and have a location object, then the weather file does not have to be present.
Good question about imposing (likely large) changes on current models. To force this with transition would mean using the Schedule option (with a constant schedule of 10). Do we want that? Do users realize they're getting a model based on 10C? 

If we think the change in results is too disruptive, then we can leave things alone and let the default without a Site:WaterMainTemperature remain as 10C but add a warning.  The new CorrelationFromWeatherFile option would require adding a Site:WaterMainTemperature object to the idf.

Nigusse:
I agree, if we force to default to CorrelationFromWeatherFile when there is no Site:WaterMainTemperature object would cause large diffs. I support to keep 10C as default when there is no waterMain object just to give users the option to duplicate their past results. And users will have the option to use the new method if they want to.

Nigusse:
@mjwitte Response:

1. If Site:WaterMainsTemperature is present, then use it and report to eio.

This branch already does this.
 
2. If Site:WaterMainsTemperature is not present, then use the default, which is CorrelationFromWeatherFile. This is similar behavior to other top-level inputs like HeatBalanceAlgorithm, SurfaceConvectionAlgorithm:Inside, etc.

Defaults to 10C. Users will be able to reproduce results. 

3. However, this comes with the added dependency on an epw ( or stat?) file being present, so there has to be a fallback if it's not. I guess that's where the 10C still has a place in life, but it shouldn't happen without throwing a warning so the user is aware of it (it looks like there is a warning but it only happens if the user requests CorrelationFromWeatherFile, so if CorrelationFromWeatherFile is the default even without an object then this should work).  

Defaults to 10C. This branch already produces warning message if an object exists if not simply report the fixed default to eio file. See the reporting format in item #4. 

4. When the fixed 10C is used, the eio output should document it somehow.  Perhaps Method=Fixed, and an extra field at the end for a constant value?  But the eio outputs should be consistent with the input choices, but there aren't inputs available to make that happen . . . . so not sure what to suggest here.

sample output to eio when it defaults to 10C
Site Water Mains Temperature Information ,Calculation Method{}, Water Mains Temperature Schedule Name{}, Annual Average Outdoor Air Temperature{C}, Maximum Difference In Monthly Average Outdoor Air Temperatures{deltaC}, Fixed Default Water Mains Temperature{C}
 "Site Water Mains Temperature Information,FixedDefault,NA,NA,NA,10.0" 

5. If the method is CorrelationFromWeatherFile, then the values for "Annual Average Outdoor Air Temperature"  and "Maximum Difference In Monthly Average Outdoor Air Temperatures" are obtained from scanning the weather file. These would override any user inputs for these fields and would also be reported to the eio (looks like current code will do that).  But the schedule name should be written as "NA" if the method is not "Scheduled".

Will do.

6. The eio output should be written anytime that water mains temperature is relevant (or simply always).

Will do. Will report to eio always.

7. Also, just noticed that other similar methods, e.g. Site:GroundTemperature:Undisturbed:KusudaAchenbach has an input for "Phase Shift of Minimum Surface Temperature". Looks like this water mains method uses the same sine wave regardless of the timing of the weather with a check for northern/southern latitude (not mentioned in the engineering reference) to make sure it's not completely backwards for southern hemisphere.  Probably good enough as a default.

This is contrary to allowing the user to have the option to use the fixed to reproduce past results. If we want sine-wave output (for future) I prefer to use default values for "Annual Average Outdoor Air Temperature" and "Maximum Difference In Monthly Average Outdoor Air Temperatures" and use existing function.

Nigusse:
eio sample output when Site:WaterMainsTemperature object exists and the method is CorrelationFromWeatherFile:
! <Site Water Mains Temperature Information>,Calculation Method{},Water Mains Temperature Schedule Name{},Annual Average Outdoor Air Temperature{C},Maximum Difference In Monthly Average Outdoor Air Temperatures{deltaC},Fixed Default Water Mains Temperature{C}
Site Water Mains Temperature Information,CorrelationFromWeatherFile,NA,9.99,28.78,NA

eio sample output when Site:WaterMainsTemperature object exists and the method is Correlation:

! <Site Water Mains Temperature Information>,Calculation Method{},Water Mains Temperature Schedule Name{},Annual Average Outdoor Air Temperature{C},Maximum Difference In Monthly Average Outdoor Air Temperatures{deltaC},Fixed Default Water Mains Temperature{C}
Site Water Mains Temperature Information,Correlation,NA,9.69,28.10,NA

eio sample output when Site:WaterMainsTemperature object does not exists:
! <Site Water Mains Temperature Information>,Calculation Method{},Water Mains Temperature Schedule Name{},Annual Average Outdoor Air Temperature{C},Maximum Difference In Monthly Average Outdoor Air Temperatures{deltaC},Fixed Default Water Mains Temperature{C}
Site Water Mains Temperature Information,FixedDefault,NA,NA,NA,10.0

Nigusse:
@mjwitte In addition to reporting a fixed default value of 10C to eio file,  the following warning error message is reported when Site:WaterMainsTemperature object does not exists in the idf. 

  ** Warning ** CalcWaterMainsTemp: Site:WaterMainsTemperature object does not exit.
   **   ~~~   ** Water Mains Monthly Temperature cannot be calculated.
   **   ~~~   ** Instead a fixed default value of 10 C will be used.

Nigusse:
@mjwitte @Myoldmopar The EIO and ERR diffs are expected. The EIO diffs is because of <Site Water Mains Temperature Information'> is reported always and the ERR diffs are warning messages for those idfs without Site:WaterMainsTemperature object. This is ready for final review.

Myoldmopar:
1. There's a typo in the error file message (does not exit)
2. We can't add warning messages to 549 files.  Do I need to write a script to add this object to every idf we have?  Does this need to be added through transition rules to user idfs?

mjwitte:
Funny thing, if you add this object to all the idfs, then you can't reproduce the existing results, unless you use the Scheduled option and create a constant 10C schedule, which I suppose could be done, but seems silly.

mjwitte:
So, the original issues suggested a new default. At this point, we haven't done that. Instead we added a warning and some eio output (which will probably not be noticed) to make users aware. We *could* go ahead and make CorrelationFromWeatherFile the new default and just accept that there will be different results for anything that's using mains water without an object present. @shorowit

Nigusse:
@Myoldmopar  I will fix the typo and there is also a comment section that needs to be removed.

Nigusse:
@mjwitte @Myoldmopar Instead of creating a constant 10C schedule, we can add the object and trick the correlation equation to produce constant water main temperature using 6.6666667 C value for annual average air temp and 0.0 DeltaC value for maximum difference of monthly average air temp. The other option is to not report the warning error message when there is no water main object but report the 10 C to eio file.

Myoldmopar:
I don't know where my feelings sit at the moment, but I am generally very queasy about having **all** of our example files get warnings dropped into them.  If I had to choose at the moment, I'd probably say just don't throw the warning.

Nigusse:
Removed the warning error message when Water Mains Temp object does not exist. Now we report the fixed default value of 10 C report to the eio file. This should be sufficient information, which was not reported previously, for now.

Myoldmopar (approved):

Myoldmopar:
All clean, this will go in.  If we need to have any further discussion about the warning message or anything else, we'll have that discussion offline and do a follow-up issue.

Issue #6422, opened 2017-12-28
Economizer faults do not work because fault objects are processed too late - other fault types may have a similar issue


mjwitte:
Issue overview
--------------
User complained that FaultModel:TemperatureSensorOffset:OutdoorAir has no impact on results.  As explained below, economizer faults stopped working in v8.7.

Tracing this in the debugger reveals that `NumFaultyEconomizer` is zero [here](https://github.com/NREL/EnergyPlus/blob/017868e84f0eb0de7711a942c72aebc273bafcc8/src/EnergyPlus/MixedAir.cc#L1442-L1449) in `MixedAir::GetOAControllerInputs` because `CheckAndReadFaults` is called too late [here](https://github.com/NREL/EnergyPlus/blob/017868e84f0eb0de7711a942c72aebc273bafcc8/src/EnergyPlus/SimulationManager.cc#L386) in `SimulationManager::ManageSimulation`.

Not sure why, but `CheckAndReadFaults` was moved to be after `SetupSimulation` as part of [this commit](https://github.com/NREL/EnergyPlus/commit/a9ceacfac709cc668d380727577c15d32de11138#diff-5b4ea1e38e96bd2ee33e0d0cf51d12b0L384).

As part of this fix, it seems that it would be more efficient to add the fault model index(es) to the OA controller data structure rather than the current method of adding the controller ID to the faults data structure (which is currently broken) and then searching through all of the economizer faults every time step for every OA controller [here](https://github.com/NREL/EnergyPlus/blob/017868e84f0eb0de7711a942c72aebc273bafcc8/src/EnergyPlus/MixedAir.cc#L3182-L3183) in `MixedAir::InitOAController`.

Just to confirm, ran example file AirEconomizerFaults_RefBldgLargeOfficeNew2004_Chicago.idf with v8.6, v8.7, and v8.8.  Cooling Electricity changed from 1486 GJ in v8.6 to 1696 GJ in v8.7 and v8.8.  (Interesting that the fault saves energy, but that's a secondary concern.)   Just to confirm, made a copy of this with faults deleted.  v8.6 results with faults are different from no faults.  v8.7 and v8.8 are identical with or without the faults.

### Workaround
There is no workaround other than going back to v8.6.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v8.8.0 (the problem was introduced in v8.7.0), and current develop v8.9 017868e84f
 - Helpdesk ticket 12395

### Checklist
[...]

PR #6778, opened 2018-06-12
Economizer faults do not work because fault objects are processed too late

Nigusse:
Issue #6422 

Pull request overview
---------------------
This is a defect. Economizer faults do not work because fault objects are processed too late. Now the OA controller get input requires processing Economizer faults. And pointers to economizer faults and number of of economizer faults are filled in OAController object data structure. Made similar fix to boiler faults processing. Now instead of searching through all of the economizer faults every time step for every OA controller in MixedAir::InitOAController economizer faults are identified and indexed once during get input. Diffs are expected in the example file with economizer fault. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
This fix addresses that the economizer and boiler faults are correctly set after the economizer and boiler objects input are read.

Nigusse:
@mjwitte there is a secondary issue as you pointed out. Below is my observation:
Run the example file (AirEconomizerFaults_RefBldgLargeOfficeNew2004_Chicago.idf) with and without the economizer faults using version 8.9, found no diffs in the results. Annual cooling electric energy use
761.68 GJ (V8.90 without fault) and 1761.68 GJ (V8.90 with fault). This confirms your observation.

Again run the example file (AirEconomizerFaults_RefBldgLargeOfficeNew2004_Chicago) with and without the economizer faults using this branch, found diffs in the annual cooling electric energy use:
761.68 GJ (this branch without fault), 1736.76 GJ (V8.90 with fault). The annual cooling electric energy use increased with fault contrary to my expectation. However, the annual natural gas energy use increased from 5330.59 GJ (this branch without fault) to 5333.49GJ (this branch with fault). This input file has 
five faults. Further testing shows similar trend even with only one fault on OA temperature. It needs further investigation, may be as a separate issue.

Nigusse:
Further investigation shows that there two issues that could lead to an increase in cooling electricity use with faulty economizer model:
(1) The faulty econoomizer model does not use updated properties (temperatures, humidity ratio and enthalpy) values in OAControllerProps::CalcOAController, InitOAMixer, CalcOAMixer) functions. 
The only function where updated faulty temperatures, humidity ratio and enthalpy currently are used is in OAControllerProps::CalcOAEconomizer function. 

(2) If the economizer has fault and either its inlet temperature or humidity ratio is offset, then shouldn't the enthalpy be updated immediately. But I do not see any where in the code that updates air enthalpy associated with faulty temperature or humidity ratio offset.

Nigusse:
This fix addresses the initial fault and OA controller setup but the economizer interaction with fault offset will be investigated in a separate issue. This is ready for review.

nrel-bot-3:

nrel-bot-3:

Myoldmopar:
@Nigusse I pulled develop into this branch.  I'll let CI verify the merge went properly after the conflict resolution.  If all goes fine, this can go in.

Myoldmopar:
@Nigusse with these changes, things are almost fully clean, but in addition to the Unitary System Water Coils Multi Speed file that's showing diffs, the AirEconomizerFaults_RefBldgLargeOfficeNew2004_Chicago file is also showing new warnings about cooling tower temperature.  Could you make one final pass on this and make sure the diffs are expected and clean up that error file before I do my final runs before merging?  Thanks!

Nigusse:
@Myoldmopar Sure will check the results and error file.

Nigusse:
@Myoldmopar I run the file AirEconomizerFaults_RefBldgLargeOfficeNew2004_Chicago.idf locally I get the same warning error. The economizer fault model is not fully functional. The fault values are not fully propagated into the economizer model to say the list, hence I do not expect the economizer fault model to work properly. We need post a separate issues that would look to this error as well.

src/EnergyPlus/FaultsManager.cc
```diff
@@ -147,6 +147,8 @@ namespace FaultsManager {
     int const iFault_Fouling_Chiller(115);
     int const iFault_Fouling_EvapCooler(116);
 
+    bool GetFaultsInputFlag(true); // Flag set to make sure you get input once
```
Myoldmopar: Does this need to be added to a clear_state function for unit testing?

tst/EnergyPlus/unit/FaultsManager.unit.cc
```diff
@@ -312,4 +315,142 @@ TEST_F(EnergyPlusFixture, FaultsManager_CalFaultOffsetAct)
...
+    EXPECT_EQ(MixedAir::OAController(1).NumFaultyEconomizer, 3);
+    EXPECT_EQ(MixedAir::OAController(1).EconmizerFaultNum(1), 1);
+    EXPECT_EQ(MixedAir::OAController(1).EconmizerFaultNum(2), 2);
+    EXPECT_EQ(MixedAir::OAController(1).EconmizerFaultNum(3), 3);
+
+    // there are two economizer faults in the 2nd OA controller
+    EXPECT_EQ(MixedAir::OAController(2).NumFaultyEconomizer, 2);
+    EXPECT_EQ(MixedAir::OAController(2).EconmizerFaultNum(1), 4);
+    EXPECT_EQ(MixedAir::OAController(2).EconmizerFaultNum(2), 5);
+}
```
Myoldmopar, line 453: :+1:

src/EnergyPlus/MixedAir.cc
```diff
@@ -1338,6 +1338,7 @@ namespace MixedAir {
         static int ADUNum(0);
         int jZone;
         int i;
+        int j;
```
Myoldmopar: No big deal but `j` could be defined down in the loop below where it is used instead of up here.

Myoldmopar (commented):
The only thing that really concerns me about this is whether the GetInput flag(s) need to be added to a proper clear_state function to ensure our unit testing is good to go.

Myoldmopar:
@Nigusse 
Thanks for the clarification on the faults issue.  I think we need to get a quick look from faults folks before faulting folks for fault failures.  

@XuanLuoLBL @hongtz68 
@Nigusse is seeing some issues with the faults.  This pull request fixes an issue related to indexing and get input, but with this fix in place, a test file is now showing a bunch of warnings.  I am intending on merging this in and opening up a new issue to address the warnings, but I would like to get someone's eyes on it from the west coast before proceeding.

src/EnergyPlus/FaultsManager.cc
Nigusse: It should have been included. Will you add it or do you want me to do it?

Myoldmopar: I could but it would be tomorrow before I can get to it.  If you can squeeze it in and also open up a new issue to track the remaining fault warning/behavior then this could merge in.

Nigusse: Sure. Will do.

Nigusse: @Myoldmopar It is already added in clear_state. May be I do not need to add this variable in the first place because there is one already but it cannot be used outside this module. I am testing if I can use the existing one.

tst/EnergyPlus/unit/FaultsManager.unit.cc
Nigusse: @Myoldmopar Removed the get input flag that I added. It was unnecessary.

Nigusse:
@Myoldmopar Posted new issue for economizer fault model investigation #6873.

Myoldmopar:
Thanks for this @Nigusse; merging.

Myoldmopar:
@Nigusse A unit test shows failing on Mac CI after this PR.  When I ran them locally they passed.  And they are still passing for me locally.  Do you have any thoughts on whether you'd expect this to fail?

http://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-bda9c90366c3805e19cd3f11174471a5af4df7b8-x86_64-MacOS-10.9-clang.html#test

Issue #6470, opened 2018-02-12
ThermalStorage:ChilledWater:Stratified is reporting the same value for Source Side Inlet and Outlet Temperatures


mjwitte:
Issue overview
--------------
User noted that ThermalStorage:ChilledWater:Stratified always reports the same value for both the Source Side Inlet Temperature and the Source Side Outlet Temperature. Comparing against the System Node Temperature outputs, it appears that the Souce Side Inlet Temperature output is incorrect.

### Workaround
Use the System Node Temperature for the corresponding node.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.8 and  v8.9.0-8346d7a61c (Feb 12)
 - Helpdesk ticket number 12507

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6782

PR #6782, opened 2018-06-16
Addressed Incorrect Source Side Inlet Temperature in Stratified Tank

Nigusse:
Issue #6470 

Pull request overview
---------------------
This is a defect. Source Side Inlet temperature of stratified tank is incorrectly re-set to the source side outlet temperature in CalcWaterThermalTankStratified() function at the end of tank heat balance calculation. Removed the source side inlet temperature reset and a different from the source side outlet temperature is expected. Example files with stratified storage tanks show diffs because of this fix. 

### Work Checklist
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the 
### Review Checklist
 - [ ] Unit Test(s)


Nigusse:
Since the chilled water storage tank is the demand side of the primary loop, its source side inlet temperature should match the primary loop demand side inlet node temperature. See the pictures below for these temperatures before and after the fix. After the fix these two temperatures are identical. These temperatures are based on the defect idf. The only example file showing ESO and EIO output diffs is "5ZoneVAV-ChilledWaterStorage-Stratified.idf"

issue 6470

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8378,8 +8378,6 @@ namespace WaterThermalTanks {
             DataLoopNode::NodeData const &HPWHCondWaterOutletNode = DataLoopNode::Node(HeatPump.CondWaterOutletNode);
             Real64 const HPWHCondenserDeltaT = HPWHCondWaterOutletNode.Temp - HPWHCondWaterInletNode.Temp;
             Tank.SourceInletTemp = Tank.SourceOutletTemp + HPWHCondenserDeltaT;
-        } else {
-            Tank.SourceInletTemp = Tank.SourceOutletTemp;
```
Myoldmopar, line 8376: @nmerket Can you please review this change?  At first it seemed like a straightforward change, but now I am a bit lost on why any of the logic in here would ever be overwriting Tank.SourceInletTemp.  That should be fixed once at the beginning of the simulation, with a temperature provided by the plant, and then never touched.  Only outlets.  Thoughts?

nrel-bot-3:

Myoldmopar:
@shorowit do you know if @nmerket is available to look at this?  I only ask because he was most recently in the tank code and it looks fishy to me.  Probably OK, but I would just like someone else to take a peek.   @Nigusse if not, I'll dig in, but it'll take a little time so I can understand the logic here.

nmerket:
@Myoldmopar I can look at this. Can it wait until the week of August 13th? I'm getting ready to head out of town.

Myoldmopar:
Yep, just a bug fix, not an IO change.  You've got time.  Thanks for confirming!

src/EnergyPlus/WaterThermalTanks.cc
nmerket: This is okay. The intention was that for wrapped condenser HPWHs, there is no source side heat transfer because it's handled separately. This was to ensure that `Qsource` was indeed zero for that case, but unintentionally applies it to other stratified tank configurations as well.

nmerket (approved):
This looks good to me. Thanks for finding and fixing this bug @Nigusse.

Myoldmopar:
Pulled develop, resolved the merge conflict, fixed the merge conflict, re-ran the test suite, everything looks good.  Thanks @nmerket for also looking at this.  Merging.

PR #6783, opened 2018-06-17
Apply clang format to all E+ source files

Myoldmopar:
Pull request overview
---------------------
This applies the clang format to the entirety of the src/EnergyPlus .cc and .hh files (and subdirectories).  I didn't apply it to the tst/EnergyPlus/unit directory.  I am hoping to minimize the impact on existing branches since the number is relatively high again.  Let's deal with the src directory for today.

I spot checked a few files, and it looks like it is just doing the expected cleanup: wrapped argument lists, alphabetical imports, misc. spacing.  It builds successfully locally.  If it builds on CI and there aren't any diffs, this should drop right in so we can get started on merging it into other branches.


Myoldmopar:
FYI @rraustad @mjwitte

rraustad (approved):
These changes look pretty much like I see when I edit any file. So yes I agree with the result.

rraustad:
decent_CI won't allow me to create a branch so I am showing what we are doing here so you can see.

The logic is if the CI is going to clean up the folders after a build (which it does now) then just test for the folder and delete if it's still there. p.clean_up and p.clean_up_regression both have tests for if !@test_run && !@keep_build_folder so this should work just fine when deleting left over folders. We are not actually doing this yet (we are calling try_hard_to_remove_dir and I don't like that since it doesn't have tests for the keep_build_folder flag), but I will implement this next week to test.  What is happening is that CMAKE is writing a log file after these files/folders are deleted which recreates the folder tree and breaks the existing logic.

    if !File.directory?(regression_base.get_build_dir(compiler))
         $logger.info "Beginning regression baseline (#{regression_base.descriptive_string}) build for #{compiler} #{p.descriptive_string}"
         regression_base.do_build compiler, nil
         regression_base.do_test compiler, nil
     else
         p.clean_up compiler
         p.clean_up_regressions compiler
         if !File.directory?(regression_base.get_build_dir(compiler))
             $logger.info "Beginning regression baseline (#{regression_base.descriptive_string}) build for #{compiler} #{p.descriptive_string}"
             regression_base.do_build compiler, nil
             regression_base.do_test compiler, ni
         else
             $logger.info "Skipping already completed regression baseline (#{regression_base.descriptive_string}) build for #{compiler} #{p.descriptive_string}"
         end
     end

mjwitte (commented):
@Myoldmopar Touched a random file `BoilerSteam.cc` in VisualStudio (2017, latest), copy/paste a block and apply clang format to entire file.  In develop this made changes, in this branch, no diffs.  Since the `x86_64-Linux-Ubuntu-14.04-gcc-4.8` tests failed, those results were deleted and it's re-running now.  If that comes back green, will merge this.

Myoldmopar:
Green :+1: :+1: :+1:

mjwitte (approved):

Issue #6712, opened 2018-04-04
Zone Coil Autosizing does not include impact of AirTerminal:SingleDuct:Mixer


EnergyArchmage:
Issue overview
--------------
From code inspection and studying coil sizing summary table output it was found that the coil inlet conditions are the zone return air conditions and not the mixed air from AirTerminal:SingleDuct:Mixer and the zone return air.  When a DOAS unit is connected to ZoneHVAC equipment there will often be a benefit that can reduce zone coil size.  For example if the DOAS is handling latent load and providing low wet bulb primary air, the actual mixed air condition will have a lower wet bulb than the zone return air leading to a smaller cooling coil size result.  Current code does not take this account. 

Suggest that the ZoneEqSizingData structure be expanded to include terms for air terminal mixer volume flow and primary air drybulb and humidity ratio for cooling and heating.  These values can be obtained from the central values in Sizing:System.  The coil parent objects are modified to fill in the sizing information needed in a way analogous to how OAVolFlow is used to account for direct outdoor air. Then all the coil sizing routines would be changed to use the new air terminal mixer information passed via ZoneEqSizing whenever the coil has an AirTerminal:SingleDuct:Mixer attached to the parent. 

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
See PR #6788.

mjwitte:
Closed via #6788

Issue #6172, opened 2017-06-12
Sizing Consolidation - Next Steps


rraustad:
Sizing issues highlight several facts on coil sizing:

 - [ ] HW and CW coils use different initialization temperatures for Cp and Rho. Use HWInitConvTemp and CWInitConvTemp consistently for each coil type.
 - [ ] sizing for rated air flow and capacity need to be synchronized.
 - [ ] Recent issues highlight a flaw in sizing where FinalZoneSizing.DesHeatCoilInTemp is not always correct. ReportSizingManager should always calculate coil entering air temp based on OA fraction. Likely same issue for cooling mode.
 - [ ] The use of SystemAirFlow, HeatingAirFlow and CoolingAirFlow flags set by parent objects has gotten a little out of hand. The SystemAirFlow flag was originally intended for fan sizing, at max of cooling and heating flow rates. The Heating and Cooling AirFlow flags were meant for the respective coils where each could be sized using a different air flow rate (if the SystemAirFlow check is present in coil cases then the coils will size the same as the fan instead of independently - if coils should size the same is the parent responsibility). In ReportSizingManager, the only occurrence of SystemAirFlow should be for fans. Also, for coil sizing the use of HeatingAirFlow for a cooling coil should not be present (e.g., CoolingCapacitySizing should not include the HeatingAirFlow flag or if they do remain they have corresponding priority?).
 - [ ] All zone equipment need to set ZoneEqSizing data at the end of the sizing routine.
      - [x] fan coil
      - [x] unit ventilator
      - [x] unit heater
      - [ ] PT Unit
      - [ ] Window AC
      - [ ] Others ?

 - [ ] All zone equipment need to set HeatingAirFlow and/or CoolingAirFlow flags and associated HeatingAirVolFlow and CoolingAirVolFlow magnitude data so coils size correctly. The same is true for parent OA flow rate using ZoneEqSizing( CurZoneEqNum ).OAVolFlow.
 - [ ] Some ReportSizingManager cases don't make sense, for example HeatingCapacitySizing for zone equipment has:

    if ( ZoneEqSizing(CurZoneEqNum).HeatingCapacity ) {, then 
    } else if ( DataCoolCoilCap > 0.0 && DataFlowUsedForSizing > 0.0 ) {, then
    } else if ( FinalZoneSizing( CurZoneEqNum ).DesHeatMassFlow >= SmallMassFlow ) {, and finally
    } else {
            NominalCapacityDes = 0.0;

[...]

PR #6788, opened 2018-06-19
Include impact of AirTerminal:SingleDuct:Mixer in ZoneHVAC coil sizing

rraustad:
Pull request overview
---------------------
Zone equipment does not include impacts of DOAS system (i.e., when zone equipment uses AirTerminal:SingleDuct:Mixer) on coil inlet conditions during sizing. This corrects Issue #6712. Cooling coil sizing now includes the impact of the DOAS air stream, either conditioned or unconditioned, on the mixed inlet condition of the zone equipment's cooling coil. This change will typically allow smaller cooling coil sizes.

Zone models included in this update:
 - [x] ZoneHVAC:FourPipeFanCoil
 - [x] ZoneHVAC:WaterToAirHeatPump
 - [x] ZoneHVAC:PackagedTerminalAirConditioner
 - [x] ZoneHVAC:PackagedTerminalHeatPump
 - [x] ZoneHVAC:TerminalUnit:VariableRefrigerantFlow
 - [x] ZoneHVAC:UnitVentilator
 - [x] AirLoopHVAC:UnitarySystem

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types


src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -3577,6 +3609,102 @@ namespace ReportSizingManager {
         }
     }
 
+    void setHeatingOAFracAndInletTempForZoneEq(Real64 &OutAirFrac, Real64 &CoilInTemp, Real64 &DesMassFlow)
```
rraustad: Anyone want to comment on these new functions? Too busy? Break out OutAirFrac, CoilInTemp and CoilInHumRat to individual functions? If change to individual functions change to Real64 instead of void ? etc.

rraustad:
Interesting. Won't build on the CI but builds at work and home on 64 bit Windows 7 and 8.

tst/EnergyPlus/unit/WaterCoils.unit.cc
```diff
@@ -376,10 +376,10 @@ TEST_F(WaterCoilsTest, WaterCoolingCoilSizing)
...
 
     SizeWaterCoil(CoilNum);
-    EXPECT_NEAR(WaterCoil(CoilNum).InletAirTemp, 16.0, 0.0001); // a mixture of zone air (20 C) and 10% OA (-20 C) = 16 C
-    EXPECT_NEAR(WaterCoil(CoilNum).DesTotWaterCoilLoad, 1709.8638, 0.0001);
-    EXPECT_NEAR(WaterCoil(CoilNum).UACoil, 51.3278, 0.0001);
-    EXPECT_NEAR(WaterCoil(CoilNum).OutletAirTemp, 30.1302, 0.0001); // reasonable delta T above inlet air temp
+    EXPECT_NEAR(WaterCoil(CoilNum).InletAirTemp, -1.99999, 0.0001); // a mixture of zone air (20 C) and 10% OA (-20 C) = 16 C
+    EXPECT_NEAR(WaterCoil(CoilNum).DesTotWaterCoilLoad, 3908.2603, 0.0001);
+    EXPECT_NEAR(WaterCoil(CoilNum).UACoil, 94.9905, 0.0001);
+    EXPECT_NEAR(WaterCoil(CoilNum).OutletAirTemp, 30.2984, 0.0001); // reasonable delta T above inlet air temp
```
rraustad: Made this change just to get it to pass. Have already corrected these so that the results are identical to before.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -1671,6 +1674,10 @@ namespace FanCoilUnits {
                         }
                     }
                 }
+                ZoneEqSizing(CurZoneEqNum).OAVolFlow = FanCoil(FanCoilNum).OutAirVolFlow;
+                if (FanCoil(FanCoilNum).ATMixerExists) {
+                    SingleDuct::setATMixerSizingProperties(FanCoil(FanCoilNum).ATMixerIndex, ControlledZoneNum, CurZoneEqNum);
+                }
```
rraustad: This function will be used in each zone equipment type to set up ATMixer inlet conditions.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1335,13 +1334,15 @@ namespace ReportSizingManager {
...
-                                     OutAirFrac * FinalZoneSizing(CurZoneEqNum).OutTempAtHeatPeak;
+                        //                        if (DesMassFlow > 0.0) {
+                        //                            OutAirFrac = min(StdRhoAir * ZoneEqSizing(CurZoneEqNum).OAVolFlow / DesMassFlow, 1.0);
+                        //                        } else {
+                        //                            OutAirFrac = 0.0;
+                        //                        }
+                        //                        CoilInTemp = (1.0 - OutAirFrac) * FinalZoneSizing(CurZoneEqNum).ZoneTempAtHeatPeak +
+                        //                                      OutAirFrac * FinalZoneSizing(CurZoneEqNum).OutTempAtHeatPeak;
+                        // this will change above CoilInTemp equation to use ZoneRetTempAtHeatPeak which accounts for lights to return fraction
+                        CoilInTemp = setHeatCoilInletTempForZoneEqSizing(setOAFracForZoneEqSizing(DesMassFlow));
```
rraustad: The coil inlet condition is now a function call which replaces many duplicate code blocks.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,26 @@ namespace SingleDuct {
...
+        if (SizingDesRunThisAirSys) {
+            // This needs work.
+            // If air loop has coils use SA conditions, else if OA sys has coils then precool conditions, else OA conditions
+            ZoneEqSizing(curZoneEqNum).ATMixerCoolPriDryBulb = SysSizInput(sysSizIndex).CoolSupTemp;
+            ZoneEqSizing(curZoneEqNum).ATMixerCoolPriHumRat = SysSizInput(sysSizIndex).CoolSupHumRat;
+            ZoneEqSizing(curZoneEqNum).ATMixerHeatPriDryBulb = SysSizInput(sysSizIndex).HeatSupTemp;
+            ZoneEqSizing(curZoneEqNum).ATMixerHeatPriHumRat = SysSizInput(sysSizIndex).HeatSupHumRat;
+        }
+    }
+
```
rraustad: And here is the ATMixer function that sets up inlet conditions the same for all equipment.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -3577,6 +3617,88 @@ namespace ReportSizingManager {
...
+
+    Real64 setCoolCoilInletTempForZoneEqSizing(Real64 const &outAirFrac)
+    {
+        Real64 coilInTemp = 0.0;
+        if (DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum).ATMixerVolFlow > 0.0) {
+            // adjust for central DOAS AT mixer mixed inlet temp
+            coilInTemp = (1.0 - outAirFrac) * DataSizing::FinalZoneSizing(DataSizing::CurZoneEqNum).ZoneRetTempAtCoolPeak +
+                         outAirFrac * DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum).ATMixerCoolPriDryBulb;
+        } else if (DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum).OAVolFlow > 0.0) { // adjust for raw OA mixed inlet temp
+            coilInTemp = (1.0 - outAirFrac) * DataSizing::FinalZoneSizing(DataSizing::CurZoneEqNum).ZoneTempAtCoolPeak +
```
rraustad: This should probably use ZoneRetTempAtCoolPeak as setHeatCoilInletTempForZoneEqSizing does.

rraustad:
@mjwitte @Myoldmopar @EnergyArchmage this is the method used to update fan coil units to include ATMixer impact on coil sizing. This method will be ported to all other zone equipment types so if there is a comment on chosen method now is the time to comment.

rraustad:
Here's some results for the DOAToFanCoilInlet example file. Sizing apparently thinks the OA is zero for the fan coil. Sizing also doesn't know there is an ATMixer attached.

    Sizing:Zone,

      flow/person,             !- Outdoor Air Method
      0.00944,                 !- Outdoor Air Flow per Person {m3/s-person}
      0.0,                     !- Outdoor Air Flow per Zone Floor Area {m3/s-m2}

    ZoneHVAC:FourPipeFanCoil,
      SPACE1-1 Fan Coil,       !- Name
      FanAvailSched,           !- Availability Schedule Name
      ConstantFanVariableFlow, !- Capacity Control Method
      autosize,                !- Maximum Supply Air Flow Rate {m3/s}
      ,                        !- Low Speed Supply Air Flow Ratio
      ,                        !- Medium Speed Supply Air Flow Ratio
      0,                       !- Maximum Outdoor Air Flow Rate {m3/s}
      ,                        !- Outdoor Air Schedule Name
      SPACE1-1 Fan Coil Inlet, !- Air Inlet Node Name
      SPACE1-1 Supply Inlet,   !- Air Outlet Node Name
      ,                        !- Outdoor Air Mixer Object Type
      ,                        !- Outdoor Air Mixer Name

    AirTerminal:SingleDuct:Mixer,
      SPACE1-1 DOAS Air Terminal,  !- Name
      ZoneHVAC:FourPipeFanCoil,!- ZoneHVAC Unit Object Type
      SPACE1-1 Fan Coil,       !- ZoneHVAC Unit Object Name
      SPACE1-1 Fan Coil Inlet, !- Mixer Outlet Node Name
      SPACE1-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
      SPACE1-1 Air Terminal Mixer Secondary Inlet,  !- Mixer Secondary Air Inlet Node Name
      InletSide;               !- Mixer Connection Type

So when the ATMixer impact on mixed air conditions is considered, the inlet conditions to the coil are actually lower than is shown in the current develop branch. So coil sizing is now impacted for inlet ATMixers. Basically lower cooling and higher heating coil sizes.

doatofancoilinlet

rraustad:
There are diff's showing in UnitVentilator when I only added the new ATMixer function call to fan coils. This is because I changed the return temp from ZoneTempAtHeatPeak to ZoneRetTempAtHeatPeak to account for lighting heat to return. Not sure if that makes sense now but will think more about it. Reverting this for now to get to minimal, and explainable, diff's before moving on.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,26 @@ namespace SingleDuct {
         }
     }
 
+    void setATMixerSizingProperties(int const &inletATMixerIndex, int const &controlledZoneNum, int const &curZoneEqNum)
+    {
+        ZoneEqSizing(curZoneEqNum).ATMixerVolFlow = SingleDuct::SysATMixer(inletATMixerIndex).DesignPrimaryAirVolRate;
```
rraustad: Maybe I should move this air vol flow down into if (SizingDesRunThisAirSys) so that all are set or none are set?

rraustad:
Perfect, now only the 7 fan coil example files have diff's. I'll worry about that subtle change to the coil inlet temp calc later. I can now move on and make the simple change to all other zone equipment. Should be easy sailing now. This code snippet should be all that is needed in other zone equipment sizing routines (before any child equipment are called).

    if (FanCoil(FanCoilNum).ATMixerExists) {
        SingleDuct::setATMixerSizingProperties(FanCoil(FanCoilNum).ATMixerIndex, ControlledZoneNum, CurZoneEqNum);
    }

rraustad:
Just figuring out which zone temp to use for zone equipment. The only difference I see is:

    if (ReturnNode > 0) {
        RetTemp = Node(ReturnNode).Temp;
    } else {
        RetTemp = Node(ZoneNode).Temp;
    }

    CalcZoneSizing(CurOverallSimDay, ControlledZoneNum).HeatZoneRetTemp = RetTemp;
    CalcZoneSizing(CurOverallSimDay, ControlledZoneNum).CoolZoneRetTemp = RetTemp;

So why would zone equipment ever use return temp to size the coils? Zone equipment draws air from the zone and any heat gain to return should be met by the air loop. And when there is no air loop these are the same value. I am now thinking this was just a mistake during equipment model development and it propagated through to some/all zone equipment models. There is going to be diffs during sizing consolidation, probably for zone equipment that serve the same zone as an air loop, but I wonder if I want to tackle that here or in a different issue. Leaning towards saving this for the existing sizing consolidation issue #6172.

rraustad:
@Myoldmopar @mjwitte this should now be complete. I will review CI diffs and try to post more comparisons.

rraustad:
Seeing non-DOAS diffs with fan coils I am switching back to ZoneRetTemp on the else to see if that fixes non-DOAS files. Also flipped priority on CoolingCapacitySizing to honor ATMixer over anything else.

src/EnergyPlus/HVACUnitarySystem.cc
```diff
@@ -2166,6 +2166,8 @@ namespace HVACUnitarySystem {
                 ZoneEqSizing(CurZoneEqNum).OAVolFlow = 0.0;   // Equipment OA flow should always be 0 when ATMixer is used
                 SingleDuct::setATMixerSizingProperties(
                     UnitarySystem(UnitarySysNum).ATMixerIndex, UnitarySystem(UnitarySysNum).ControlZoneNum, CurZoneEqNum);
+            } else {
+
```
rraustad: Was going to set OA flow here but found that UnitarySystem has no OA flow then forgot to remove this else.

src/EnergyPlus/DataSizing.hh
```diff
@@ -639,24 +639,29 @@ namespace DataSizing {
...
-        bool Capacity;               // TRUE if AirloopHVAC system capacity is calculated
-        bool CoolingCapacity;        // TRUE if AirloopHVAC system cooling capacity is calculated
-        bool HeatingCapacity;        // TRUE if AirloopHVAC system heating capacity is calculated
-        bool SystemCapacity;         // TRUE if AirloopHVAC system heating capacity is calculated
-        bool DesignSizeFromParent;   // TRUE if design size is set by parent object - normally false, set to true for special cases e.g. ERV
+        Real64 AirVolFlow;            // design air vol flow rate for zone equipment unit [m3/s]
+        Real64 MaxHWVolFlow;          // design Hot Water vol flow for zone equipment unit [m3/s]
+        Real64 MaxCWVolFlow;          // design Cold Water vol flow for zone equipment unit [m3/s]
+        Real64 OAVolFlow;             // design outside air flow for zone equipment unit [m3/s]
+        Real64 ATMixerVolFlow;        // design ventilation air flow rate from air terminal mixer (central DOAS) [m3/s] //TRANE AT Mixer sizing
```
Myoldmopar: We don't really need comments tagging TRANE in here.  If there's anything else to change in this PR, comments like this should go.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -3577,6 +3546,90 @@ namespace ReportSizingManager {
         }
     }
 
+    Real64 setOAFracForZoneEqSizing(Real64 const &desMassFlow)
+    {
+        Real64 outAirFrac = 0.0;
+        if (desMassFlow <= 0.0) return outAirFrac;
+
+        if (DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum).ATMixerVolFlow > 0.0) {
```
Myoldmopar: At some point, we need to break this habit of using indexes assigned external to functions.  We could easily pass in `DataSizing::CurZoneEqNum` to this function, making this function significantly less dependent on the global state.  Or even better, a reference to the entire `DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum)` object could've been passed in, nearly eliminating the dependence on global state completely (and making it much easier to test).  Imagine:

```
Real64 setOAFracForZoneEqSizing(Real64 const &desMassFlow, ZoneEqSizingData const & sizingData)
 {
     Real64 outAirFrac = 0.0;
     if (desMassFlow <= 0.0) return outAirFrac;
     if (sizingData.ATMixerVolFlow > 0.0) {
         // set central DOAS AT mixer OA fraction
         outAirFrac = min(DataEnvironment::StdRhoAir * sizingData.ATMixerVolFlow / desMassFlow, 1.0);
     } else if (sizingData.OAVolFlow > 0.0) { // set zone equipment OA fraction
         outAirFrac = min(DataEnvironment::StdRhoAir * sizingData.OAVolFlow / desMassFlow, 1.0);
     }
     return outAirFrac;
}
```

Then for testing, you would just create an instance of a ```ZoneEqSizingData``` yourself in the test, assign a value for ```DataEnvironment::StdRhoAir``` (though you could make that an argument too), and then call the function however many times you needed to test it thoroughly.

I'm not necessarily saying this needs to be done for this PR, but it's definitely a direction we want to go.  I don't see any red flags on the changes in this PR, so there probably won't need to be another commit, but if there is work to be done, this should be considered.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,27 @@ namespace SingleDuct {
...
+            DataZoneEquipment::ZoneEquipConfig(controlledZoneNum).InletNodeAirLoopNum(SingleDuct::SysATMixer(inletATMixerIndex).CtrlZoneInNodeIndex);
+        bool SizingDesRunThisAirSys = false;                               // Sizing:System object found flag
+        CheckThisAirSystemForSizing(airLoopIndex, SizingDesRunThisAirSys); // check for Sizing:System object
+        if (SizingDesRunThisAirSys) {
+            int sysSizIndex =
+                UtilityRoutines::FindItemInList(FinalSysSizing(airLoopIndex).AirPriLoopName, SysSizInput, &SystemSizingInputData::AirPriLoopName);
+            if (sysSizIndex == 0) sysSizIndex = 1; // use first when none applicable
+            // set ATMixer outlet air conditions in ZoneEqSizing array
+            ZoneEqSizing(curZoneEqNum).ATMixerVolFlow = SingleDuct::SysATMixer(inletATMixerIndex).DesignPrimaryAirVolRate;
+            // This needs work.
```
Myoldmopar: Does this still need work?

tst/EnergyPlus/unit/ReportSizingManager.unit.cc
```diff
@@ -568,8 +568,237 @@ TEST_F(SQLiteFixture, ReportSizingManager_SQLiteRecordReportSizingOutputTest)
...
+    DataSizing::ZoneEqSizing(DataSizing::CurZoneEqNum).OAVolFlow = 1.0;
+    coilInletCond = setHeatCoilInletTempForZoneEqSizing(oaFrac);
+    EXPECT_EQ(coilInletCond, 0.0);
+    coilInletCond = setHeatCoilInletHumRatForZoneEqSizing(oaFrac);
+    EXPECT_EQ(coilInletCond, 0.0);
+    coilInletCond = setCoolCoilInletTempForZoneEqSizing(oaFrac);
+    EXPECT_EQ(coilInletCond, 0.0);
+    coilInletCond = setCoolCoilInletHumRatForZoneEqSizing(oaFrac);
+    EXPECT_EQ(coilInletCond, 0.0);
+}
```
Myoldmopar: Thorough! 👍

Myoldmopar (commented):
@rraustad Can you comment on your "This needs work" statement?  Otherwise I made a few comments we can discuss but I think it's ready to go.

src/EnergyPlus/DataSizing.hh
rraustad: I missed that one.

src/EnergyPlus/SingleDuct.cc
rraustad: Using the Cool/HeatSupTemp as the ATMixer supply temp assumes there is a coil in the main air loop, which is probably not the case for DOAS. So it could be the PreCool/HeatTemp that is used here. Or maybe the SAT is the correct number as long as the user knows that. So "This needs work" simply means this needs more thought to catch all corner cases.

Myoldmopar:
Pulled develop locally, resolved merge conflict, ran tests, and they all pass.  @rraustad please provide your thoughts on the comment you left in there saying it still needed work, and this can go in.  I'm going to go ahead and remove the TRANE comment.

rraustad:
My thought is that there are 2 conditions in Sizing:System. One is the precool/heat condition and the other the supply condition. The ATMxer is at the supply outlet of the air loop. When I wrote this code to use SA conditions I wondered if there was a better method to handle all possible cases instead of just using SA conditions. I didn't have a good solution so I wrote that comment that this still needs work since I didn't have a good solution at the time. One thing that might happen is that I could check for main loop coils, if they exist use SA condition, otherwise use precool/heat condition.

src/EnergyPlus/ReportSizingManager.cc
rraustad: I do like that and am just getting up to speed with these possibilities with the refactoring of UnitarySystem. Is there an impact on speed with this method? Or is it just cleaner? If speed, then I would like to address.

src/EnergyPlus/SingleDuct.cc
Myoldmopar: I guess that's still my question then.  Do you want try to catch more corner cases with this before you call it done?

rraustad: I'll check to see what variable is used to size reheat coils.

rraustad: The single duct TU's use TermUnitFinalZoneSizing(CurTermUnitSizingNum).DesHeatCoilInTempTU to size the coils re heating capacity. There is a corresponding DesCoolCoilInTempTU. I still don't think sizing accounts for when coils are not present in either the air loop or OA system. What I would feel very comfortable with is checking for cooling/heating air loop coils, then use SAT, and if not present cooling/heating OA coils, then use pre-cool/heat temps, and if not present to use weather for respective cooling or heating sizing. I know there is a function to test for OA coils, not sure about a function to test for coils in air loop. This can be a real mess when air loop coils are present since I would have to mine equipment models (e.g., Furnace) to see which coils are present. So my original thought was to drop this in but I could work on it more if needed.

mjwitte: Interesting. On initial search, it looks like DesCoolCoilInTemp simply grabs the Sizing:System cooling supply temp (
But the for heating, DesHeatCoilInTemp calls a function [GetHeatingSATempForSizing](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SimAirServingZones.cc#L7237) which does check for the presence of system coils.

mjwitte: Also, I see that I made some stupid lines back in [this commit](https://github.com/NREL/EnergyPlus/commit/ad8586e37217a3448b3ea7d11ff7f2721b57d9e2#diff-f51329ac8013440dabda3acd6c2d2a9aR6143) for the air terminal sizing stuff. There are now several lines that simply assign a variable to itself, [here](https://github.com/NREL/EnergyPlus/blob/73281ffba303cbd55bfa8f4e7e5646843f39866d/src/EnergyPlus/SimAirServingZones.cc#L6372-L6373), [here](https://github.com/NREL/EnergyPlus/blob/73281ffba303cbd55bfa8f4e7e5646843f39866d/src/EnergyPlus/SimAirServingZones.cc#L6376-L6377), and [here](https://github.com/NREL/EnergyPlus/blob/73281ffba303cbd55bfa8f4e7e5646843f39866d/src/EnergyPlus/SimAirServingZones.cc#L6439-L6440). You may as well delete those lines while you're in here.

rraustad: Great, now the reheat coil inlet temp is probably wrong if there is only an OAHX. And in GetHeatingSATempForSizing why calculate w and H to get T instead of a simple mixing calc on T ? You get the same answer either way don't you? It's a straight line between the state points. Unless the psyc routines protect against exceeding the saturation curve? The RH function does, I don't think the others do. For my purposes I'd like to know if there is any coil in the air loop, not just water, elec and gas hanging on a branch.

    // Check whether there are Central Heating Coils in the Primary Air System
    for (AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
        FoundCentralHeatCoil = false;
        for (BranchNum = 1; !FoundCentralHeatCoil && BranchNum <= PrimaryAirSystem(AirLoopNum).NumBranches; ++BranchNum) {
            for (CompNum = 1; !FoundCentralHeatCoil && CompNum <= PrimaryAirSystem(AirLoopNum).Branch(BranchNum).TotalComponents; ++CompNum) {
                CompTypeNum = PrimaryAirSystem(AirLoopNum).Branch(BranchNum).Comp(CompNum).CompType_Num;
                if (CompTypeNum == WaterCoil_SimpleHeat || CompTypeNum == Coil_ElectricHeat || CompTypeNum == Coil_GasHeat) {
                    FoundCentralHeatCoil = true;
                }
            } // end of component loop
        }     // end of Branch loop
        PrimaryAirSystem(AirLoopNum).CentralHeatCoilExists = FoundCentralHeatCoil;
    } // end of AirLoop loop

mjwitte: Well, it's not always a straight line between state points, as you suggest, it's allowing for hitting the saturation line.
OAHX isn't accounted for in *any* of the sizing calcs yet, so I wouldn't inject that here. That should be a separate mod.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,48 @@ namespace SingleDuct {
...
+            if (sysSizIndex == 0) sysSizIndex = 1; // use first when none applicable
+            // set ATMixer outlet air conditions in ZoneEqSizing array
+            ZoneEqSizing(curZoneEqNum).ATMixerVolFlow = SingleDuct::SysATMixer(inletATMixerIndex).DesignPrimaryAirVolRate;
+            // If air loop has coils use SA conditions (below), else if OA sys has coils then precool conditions, else OA conditions
+            if (DataAirSystems::PrimaryAirSystem(airLoopIndex).CentralHeatCoilExists) {
+                ZoneEqSizing(curZoneEqNum).ATMixerHeatPriDryBulb = SysSizInput(sysSizIndex).HeatSupTemp;
+                ZoneEqSizing(curZoneEqNum).ATMixerHeatPriHumRat = SysSizInput(sysSizIndex).HeatSupHumRat;
+            } else if (DataAirSystems::PrimaryAirSystem(airLoopIndex).NumOAHeatCoils > 0) {
+                ZoneEqSizing(curZoneEqNum).ATMixerHeatPriDryBulb = SysSizInput(sysSizIndex).PreheatTemp;
+                ZoneEqSizing(curZoneEqNum).ATMixerHeatPriHumRat = SysSizInput(sysSizIndex).PreheatHumRat;
```
rraustad: I think these are wrong. The Precool/heatTemp is at the outdoor air outlet prior to mixing so there is a mixing problem here just like in those other functions that sets heating temp/humrat. I'll wait for comments before making that change.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -3587,6 +3557,90 @@ namespace ReportSizingManager {
...
+        if (desMassFlow <= 0.0) return outAirFrac;
+
+        if (zoneEqSizing.ATMixerVolFlow > 0.0) {
+            // set central DOAS AT mixer OA fraction
+            outAirFrac = min(DataEnvironment::StdRhoAir * zoneEqSizing.ATMixerVolFlow / desMassFlow, 1.0);
+        } else if (zoneEqSizing.OAVolFlow > 0.0) { // set zone equipment OA fraction
+            outAirFrac = min(DataEnvironment::StdRhoAir * zoneEqSizing.OAVolFlow / desMassFlow, 1.0);
+        }
+        return outAirFrac;
+    }
```
rraustad, line 3759: @Myoldmopar I changed this one just to get familiar. It works as expected. I didn't change the other 4 below since I ran out of time. I would need 2 additional arguments for those below and then the function calls in RequestSizing would get messier.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,95 @@ namespace SingleDuct {
...
+        if (controlledZoneNum == 0) return;
+        if (curZoneEqNum == 0) return;
+        if (SingleDuct::SysATMixer(inletATMixerIndex).MixerType != DataHVACGlobals::ATMixer_InletSide) return;
+
+        // proceed to set ATMixer properties used for sizing coils
+        int airLoopIndex = // find air loop associated with ATMixer
+            DataZoneEquipment::ZoneEquipConfig(controlledZoneNum).InletNodeAirLoopNum(SingleDuct::SysATMixer(inletATMixerIndex).CtrlZoneInNodeIndex);
+        bool SizingDesRunThisAirSys = false;                               // Sizing:System object found flag
+        CheckThisAirSystemForSizing(airLoopIndex, SizingDesRunThisAirSys); // check for Sizing:System object
+        if (SizingDesRunThisAirSys) {
```
rraustad: @mjwitte would you review the math here to make sure I didn't make a mistake. I see one already. OA Frac should be based on air loop OA vs main loop flow. So this one is wrong since a single air loop can serve multiple zone equipment.

    Real64 OutAirFrac = ZoneEqSizing(curZoneEqNum).ATMixerVolFlow / FinalSysSizing(airLoopIndex).DesMainVolFlow;

rraustad: @Myoldmopar it looks like I'll be fixing other things so feel free to add to the list.

rraustad: Geeez, this is just woeful. I don't know if it's the way clang is formatting the code or if I was just getting tired. But this is bad in many ways. I'm fixing it now. Are there any other comments other than here?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -5555,21 +5573,30 @@ namespace SimAirServingZones {
                                                                  ZoneSizing(CurOverallSimDay, CtrlZoneNum).HeatFlowSeqNoOA(TimeStepInDay));
                             SysSizing(CurOverallSimDay, AirLoopNum).HeatFlowSeq(TimeStepInDay) +=
                                 adjHeatFlowSeq / (1.0 + TermUnitSizing(TermUnitSizingIndex).InducRat);
+                            // sum up the zone heating load to be met by this system for this time step
+                            SysSizing(CurOverallSimDay, AirLoopNum).SumZoneHeatLoadSeq(TimeStepInDay) +=
+                                ZoneSizing(CurOverallSimDay, CtrlZoneNum).HeatLoadSeq(TimeStepInDay) /
+                                (1.0 + TermUnitSizing(TermUnitSizingIndex).InducRat);
```
rraustad, line 5613: Line 5576 of SimAirServingZones:

@EnergyArchmage has bee struggling for a while now to figure out why some sizing results for heating were always 0. He found that these few calculations were missing. Lines 5576 - 6000. I have swept those changes in here to correct this problem. I can move these corrections to another branch if anyone has a problem with fixing this here. These in itself will probably cause diff's, as will my changes on this branch.

tst/EnergyPlus/unit/FanCoilUnits.unit.cc
```diff
@@ -2108,7 +2108,7 @@ TEST_F(EnergyPlusFixture, FanCoil_ASHRAE90VariableFan)
     StdRhoAir = 1.2;
 
     BeginEnvrnFlag = true;
-    InitFanCoilUnits(FanCoilNum, ZoneNum);
+    InitFanCoilUnits(FanCoilNum, ZoneNum, ZoneNum);
```
rraustad, line 2110: What the heck is this? Oh, it's a unit test. InitFanCoilUnits now has arguments for FanCoilNum, ZoneNum, and ControlledZoneNum (to pass to sizing). So this change simply used ZoneNum as both arguments.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -502,6 +502,18 @@ namespace ReportSizingManager {
                     }
                 }
             } else {
+
+                int tempZoneNum = CurZoneEqNum;
+                bool FinalZoneSizingNotAllocated = false;
+                if (!allocated(DataSizing::FinalZoneSizing)) {
+                    DataSizing::FinalZoneSizing.allocate(1);
+                    DataSizing::ZoneSizingData &finalZoneSizing = DataSizing::FinalZoneSizing(1);
```
rraustad: The finalZoneSizing reference initialization here at line 510 is not needed and redundant since this is done at line 515. This line could be deleted but doesn't hurt anything.

rraustad:
@EnergyArchmage @mjwitte @Myoldmopar  I was reviewing the diff's and looked at DOAToPTHP. The capacity has changed as it should have yet the entering air temperature did not. I put a breakpoint at where this value gets set and see a logic flaw that should be corrected. In the figure you see where this data was set. It's in the OAVolFlow == 0 section and grabs ZoneRetTempAtCoolPeak at 23C (I think this is wrong for zone equipment but that's another issue). DesCoolCoilInTemp = 27C since it includes the impact of OA as specified in the design spec object in the Sizing:Zone object for that zone. But this PTHP has an air terminal mixer and note that the ATMixerCoolPriDryBulb value = 32C. The PTHP is being sized here so ATMixer properties have been set. 

So I think this ReportCoilSelection logic needs to be revised either in this branch or a new one to be:

if (ATMixerVolFlow > 0)
else if (OAVolFlow > 0)
else

atmixerproperties

rraustad:
I also think the type of equipment that uses this logic needs to be expanded to any equipment that can accept the ATMixer.

        } else if (DataSizing::ZoneEqDXCoil || DataSizing::ZoneEqFanCoil || DataSizing::ZoneEqUnitarySys) {

nrel-bot-3:
@rraustad @lgentile it has been 16 days since this pull request was last updated.

src/EnergyPlus/ReportSizingManager.cc
rraustad: Obviously I circled back and fixed the remaining functions below.

mjwitte (changes requested):
Ran all of the DOATo* test files and compared results before and after. Inlet side mixers change sizing results, supply side mixer don't (I guess that's OK, but inspection of the inputs show that the ones with supply side mixers should use the DOAS options in Sizing:Zone to allow for the change in load).

- [ ] DOAToPTAC - Cooling coils are larger, entering condition is warmer than before, expected the opposite, oh, there's no coils in this DOAS, so cooling results are ok.
But gas heating coils are exactly the same size, even though the entering coil conditions changes.
And this file has Sizing:Zone DOAS options that aren't consistent with unconditioned DOAS.

- [ ] DOAToPTHP has similar issues (in this case DX heating coil changes, but it's linked to the cooling coil, the gas supp heat coil shows different inlet but no change in capacity).

- [ ] DOAToUnitarySystem - cooling coil inlet conditions change, but no change in capacity; heating coil inlet conditions are cold before and after, so OK I guess, but can't really tell for sure what it's using. This one also has unconditioned DOAS with active DOAS inputs in Sizing:Zone.

- [ ] DOAToUnitVentilator has unconditioned DOAS, cooling coil results don't change (looks like OA was included before?), HW heating coil inlet temp colder, coil larger, OK.

- [ ] DOAToVRF - looks ok, heating and cooling coils report more extreme entering conditions and both are larger (of course, their capacities are somewhat linked). But this has the same inconsistent Sizing:Zone DOAS inputs with an unconditioned DOAS.

- [ ] DOAToWaterToAirHPInlet - no change in heat pump coil sizes? Gas supp heat coil shows different inlet conditions, but same capacity. Uncoditioned DOAS, no DOAS input in Sizing:Zone.

- [ ] DOAToWaterToAirHPSupply - no diffs, as expected.

I'll post a new issue to clean up these DOATo* inputs, don't want to cloud this PR with that.

And from CI, all fan coils, PTACs, and PTHPs etc. have diffs?

Time to read the codes changes.

src/EnergyPlus/DataSizing.hh
mjwitte: This still needs cleanup.

src/EnergyPlus/DataSizing.hh
```diff
@@ -639,24 +640,29 @@ namespace DataSizing {
...
+        Real64 ATMixerHeatPriHumRat;  // design ventilation humidity ratio from air terminal mixer during heating (central DOAS) [kgH20/kgDryAir]
+        Real64 DesCoolingLoad;        // design cooling load used for zone equipment [W]
+        Real64 DesHeatingLoad;        // design heating load used for zone equipment [W]
+        Real64 CoolingAirVolFlow;     // design cooling air vol flow rate for equipment[m3/s]
+        Real64 HeatingAirVolFlow;     // design heating air vol flow rate for equipment[m3/s]
+        Real64 SystemAirVolFlow;      // design heating air vol flow rate for equipment[m3/s]
+        bool AirFlow;                 // TRUE if AirloopHVAC system air flow rate is calcualted
+        bool CoolingAirFlow;          // TRUE if AirloopHVAC system cooling air flow rate is calcualted
+        bool HeatingAirFlow;          // TRUE if AirloopHVAC system heating air flow rate is calcualted
+        bool SystemAirFlow;           // TRUE if AirloopHVAC system heating air flow rate is calcualted
```
mjwitte: While you're in here, can fix "calcualted".

src/EnergyPlus/DataSizing.hh
```diff
@@ -665,6 +671,11 @@ namespace DataSizing {
         // Default Constructor
         ZoneEqSizingData()
             : AirVolFlow(0.0), MaxHWVolFlow(0.0), MaxCWVolFlow(0.0), OAVolFlow(0.0),
+              ATMixerVolFlow(0.0),        // design ventilation air flow rate from air terminal mixer (central DOAS) [m3/s]
+              ATMixerCoolPriDryBulb(0.0), // design air terminal mixer cooling outlet temperture [C]
+              ATMixerCoolPriHumRat(0.0),  // design air terminal mixer cooling outlet humidity ratio [kgH20/kgDryAir]
+              ATMixerHeatPriDryBulb(0.0), // design air terminal mixer heating outlet temperture [C]
```
mjwitte: And "temperture".

src/EnergyPlus/DataSizing.hh
```diff
@@ -665,6 +671,11 @@ namespace DataSizing {
...
+              ATMixerHeatPriHumRat(0.0),  // design air terminal mixer cheating outlet humidity ratio [kgH20/kgDryAir]
```
mjwitte: "cheating" --> "heating"?

src/EnergyPlus/SingleDuct.cc
```diff
@@ -6015,6 +6015,162 @@ namespace SingleDuct {
         }
     }
 
+    void setATMixerSizingProperties(int const &inletATMixerIndex, // index to ATMixer at inlet of zone equipment
+                                    int const &controlledZoneNum, // controlled zone number
+                                    int const &curZoneEqNum       // current zone equipment being simulated
```
mjwitte, line 6023: This might all work if curZoneEqNum was unique for every piece of zone equipment. But it's just the current controlledZoneNum, set [here in ZoneEquipmentManager::SimZoneEquipment](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L3475). So, unless I'm missing something, this only works if there's only one piece of zoneHVAC equipment in the zone. Seems like ZoneEqSizing in general could be messed up if there's more than one piece of zone equipment. Or does ZoneEqSizing get completely rewritten by each piece of zone equipment, getting re-used safely? I need to come up with a test file that tries to prove this.

If the other problems noted in the DOATo* files are fixed, I suppose this can go in, but I think we need a separate PR to change curZoneEqNum to be unique for each piece of zone equipment similar to CurTermUnitSizingNum.

rraustad: We'll have to think more about this. Currently, zone equipment (which don't use TU's) use the zone index to access FinalZoneSizing. So if there were 2 PTAC's in the same zone they would size the same unless one had an ATMixer and the other didn't. What may be needed is some proportional sizing method as is done with TU's (e.g., CurTermUnitSizingNum can include percentage of airflow or some such thing).

rraustad: Regarding the same heating capacity, these PTAC's use 0 flow rate for OA.

    ZoneHVAC:PackagedTerminalAirConditioner,
      SPACE1-1 PTAC,           !- Name
      0,                       !- Cooling Outdoor Air Flow Rate {m3/s}
      0,                       !- Heating Outdoor Air Flow Rate {m3/s}
      0,                       !- No Load Outdoor Air Flow Rate {m3/s}

and therefore use FinalZoneSizing.MinOA to size the heating coil capacity. In develop, MinOA = 0.05 ish (if I recall correctly, or maybe 0.005) and in this branch ATMixer flow = 0.05. And the heating coil inlet temp is the same in both cases, hence the same capacity. So this branch fixes the reporting of the heating coil inlet conditions.

rraustad: I am also concerned about the use of these inputs to adjust equipment flow rate.

    Sizing:Zone,
      Yes,                     !- Account for Dedicated Outdoor Air System
      ColdSupplyAir,           !- Dedicated Outdoor Air System Control Strategy
      13.8,                    !- Dedicated Outdoor Air Low Setpoint Temperature for Design {C}
      15.6;                    !- Dedicated Outdoor Air High Setpoint Temperature for Design {C}

The IO ref says:

``
Field: Account for Dedicated Outdoor Air System  -
This is a choice field with choices Yes or No. The default is No. Choosing Yes means that the zone sizing calculation will use the subsequent inputs to calculate the heat gain or loss (heat gains are positive, heat loss is negative) imposed on the zone by a Dedicated Outdoor Air System (DOAS). This heat gain is then added to the zone design heat gain for the zone and the zone design air flow rate is adjusted to meet the DOAS heat gain plus the zone design heat gain.
``

This shouldn't be needed if the coil inlet conditions are correctly calculated. With both air flow (I haven't found where this is done yet) and inlet conditions being adjusted, this could be double counting the impact/load due to DOAS.

rraustad: I traced this down to function SizeZoneEquipment in ZoneEquipmentManager where the zone load is offset by the DOAS. This would adjust the zone mass flow rate. But this is for the DOAS system not the ATMixer so this is probably OK.

mjwitte: As I understand this, you need this adjustment if it's a supply side mixer, because that type of mixer does not change the zone equipment coil inlet conditions.  The ATMixed primary flow goes directly to the zone from the DOAS and impacts the zone load, so the zone equipment should meet a different load. If it's an inlet side mixer, then you don't want this, because the coil inlet conditions will already acount for the DOAS air from the mixer, so this would double-count that.

rraustad: I think the ATMixer was added after DOAS. So how would we know during sizing where the DOAS air is directed, to the zone or to the inlet of the HVAC system?

rraustad: @mjwitte I could probably check right here (inside setATMixerSizingProperties) to see if Sizing:Zone input for Account for Dedicated Outdoor Air System is set to Yes and issue a warning and fatal. Other than that not sure what else I could do to avoid double counting DOAS.

mjwitte: We don't. As it stands now, the user has to understand that and know when to use this and when not to. Prior to this change, one should use the DOAS sizing inputs for any DOAS, whether it's direct or through an ATMixer (supply-side and inlet-side). Now, this change automatically makes the adjustment for inlet-side ATMixers, but not the other two. Some options to consider:
1. If there's a supply-side ATMixer, then adjust the base load/zone supply flow rate that this equipment is sizing against.
2. When there's an ATMixer, check the zone sizing DOAS inputs and throw warnings. If active for an inlet-side mixer, warn that it's double-counting. If not active for a supply-side mixer, warn that it's not being accounted for.
3. Ultimately, we'd like to move toward eliminating the zone sizing DOAS inputs and always figure it out automatically, but it's too late for that here.

mjwitte: I didn't see you last post while composing this.  At this point the warning is probably the best thing to do and then consider further changes for next release. But I don't think it should be fatal.

mjwitte:
Regarding the new warnings, 7 files now have warnings. Good to exercise the warnings here, but now we need to correct those test files so the warnings go away.

In DOAToUnitarySystem, the warnings are happening twice per zone.

In OutdoorAirUnit and OutdoorAirUnitwithAirloopHVAC (and maybe others), the desiccant regen coil capacity has changed - it shouldn't.

Minor - the warnings seem a bit verbose. 
```
   ** Warning ** AirTerminal:SingleDuct:Mixer: SPACE1-1 DOAS AIR TERMINAL
   **   ~~~   **  Sizing:Zone object sets input Account for Dedicated Outdoor Air System = Yes.
   **   ~~~   **  This inlet side Air Terminal Mixer will also adjust coil sizing and may result in inappropriately sized coils.
   **   ~~~   **  Set Account for Dedicated Outdoor Air System = No in Sizing:Zone object for zone = SPACE1-1

   ** Warning ** AirTerminal:SingleDuct:Mixer: SPACE4-1 DOAS AIR TERMINAL
   **   ~~~   **  Sizing:Zone object sets input Account for Dedicated Outdoor Air System = No.
   **   ~~~   **  This supply side Air Terminal Mixer will not adjust coil sizing and may result in inappropriately sized coils.
   **   ~~~   **  Set Account for Dedicated Outdoor Air System = Yes in Sizing:Zone object for zone = SPACE4-1
```
Suggest
```
   ** Warning ** AirTerminal:SingleDuct:Mixer: SPACE1-1 DOAS AIR TERMINAL
   **   ~~~   **  Inlet side Air Terminal Mixer automatically adjusts zone equipment coil sizing.
   **   ~~~   **  Set Account for Dedicated Outdoor Air System = No in Sizing:Zone object for zone = SPACE1-1.

   ** Warning ** AirTerminal:SingleDuct:Mixer: SPACE4-1 DOAS AIR TERMINAL
   **   ~~~   **  Supply side Air Terminal Mixer does not adjust zone equipment coil sizing and may result in inappropriately sized coils.
   **   ~~~   **  Set Account for Dedicated Outdoor Air System = Yes in Sizing:Zone object for zone = SPACE4-1.
```

mjwitte:
DOAToPTAC - Current results show the heating coil sensible capacity is now the same as total (that was broken before), but the coil inlet temp is the OA temp? And cooling coil inlet conditions are nearly the same as outdoor also? In the component sizing summary design cooling and heating flow rates for the PTAC 1-1 are both 0.22, and the ATMixer flow rate is 0.0572, so it seems like the coil entering conditions should be closer to zone conditions than outdoor. But the coil sizing details show the heating coil design flow rate at 0.572? Similar questions for other systems.

rraustad:
Looking at OutdoorAirUnit at the moment. The heating coil is sizing differently because I added the new function call in heating coils thinking it would reduce code. What is happening is that the desiccant heating coil previously sized using DesHeatCoilInTemp and now it sizes using ZoneTempAtHeatPeak, which is of course wrong. The OAUnit sizing function doesn't set ZoneEqSizing::OAVolFlow (it does now). The problem I see here is that zone sizing does not know what type of equipment there is or how many. In this case there is only the OutdoorAirUnit. But what if there were the OAUnit and a PTAC with no OA, then the PTAC would be using -17C to size the heating coil? I've seen this before with the OA unit. My change corrects the heating coil size but there is still this other lurking issue. Now on to other things you noticed.

    ZoneHVAC:EquipmentConnections,
      SPACE1-1,                !- Zone Name
      Zone1Equipment,          !- Zone Conditioning Equipment List Name
      Zone1Inlets,             !- Zone Air Inlet Node or NodeList Name
      Zone1Exhausts,           !- Zone Air Exhaust Node or NodeList Name
      Zone 1 Node,             !- Zone Air Node Name
      Zone 1 Outlet Node;      !- Zone Return Air Node or NodeList Name

    ZoneHVAC:EquipmentList,
      Zone1Equipment,          !- Name
      SequentialLoad,          !- Load Distribution Scheme
      ZoneHVAC:OutdoorAirUnit, !- Zone Equipment 1 Object Type
      Zone1OutAir,             !- Zone Equipment 1 Name
      1,                       !- Zone Equipment 1 Cooling Sequence
      1;                       !- Zone Equipment 1 Heating or No-Load Sequence

    Sizing:Zone,

      Flow/Zone,               !- Outdoor Air Method
      0,                       !- Outdoor Air Flow per Person {m3/s-person}
      0.0,                     !- Outdoor Air Flow per Zone Floor Area {m3/s-m2}
      0.42;                    !- Outdoor Air Flow per Zone {m3/s}


desiccantheatingcoilsizing

rraustad:
For DOAToPTAC, when capacity is sized, where CoilInTemp is also set and used to calculate capacity, the capacity gets reported but the coil conditions do not. So at the bottom of request sizing where capacity is reported, when it jumps over the setCoilCoolingCapacity, that function sets coil inlet temp as the ATMixer primary air temp. Since there are no coils in the ATMixer, the supply air temp = OA temp. I have fixed this over in #5763 (New Feature - Include fan heat in coil sizing). I've got a lot of kettles on the fire. Maybe a followup issue to make sure these are corrected after all PR's are merged? Or merge in those changes here to get these results correct?

coilselectionsizingissue

rraustad:
I've fixed the duplicate warnings and desiccant heating coil sizing issues. Just need to push those changes.

mjwitte:
Would it be better to review and merge the fan heat PR before this to set a new baseline for comparison?

rraustad:
Good idea. That would fix this last issue when merged into this branch.

mjwitte:
@rraustad Please resolve the conflict here and merge up today if you can.

rraustad:
Need to verify this comment when results are available:

DOAToPTAC - Current results show the heating coil sensible capacity is now the same as total (that was broken before), but the coil inlet temp is the OA temp? And cooling coil inlet conditions are nearly the same as outdoor also? In the component sizing summary design cooling and heating flow rates for the PTAC 1-1 are both 0.22, and the ATMixer flow rate is 0.0572, so it seems like the coil entering conditions should be closer to zone conditions than outdoor. But the coil sizing details show the heating coil design flow rate at 0.572? Similar questions for other systems.

mjwitte:
Just to be clear, the "0.572" should be "0.0572" - equal to the OA flow.

rraustad:
So if PTAC flow = OA flow then coil inlet temp = OA temp.

mjwitte:
Yes, but the question here is that the reported design cooling and heating flow rates for the PTAC 1-1 are both 0.22, so why is it sizing the heating coil for only 0.0572?

rraustad:
Looking at this now, DOAToPTAC. I see that SPACE1-1 PTAC air flow is sized larger than before. This is a little surprising since I only expect coil entering/exiting conditions (and capacity) to change. So comment here if you wish as I look at what is happening.

     ! <Component Sizing Information>, Component Type, Component Name, Input Field Description, Value
    - Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Cooling Supply Air Flow Rate [m3/s], 0.22088
    - Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Heating Supply Air Flow Rate [m3/s], 0.22088
    + Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Cooling Supply Air Flow Rate [m3/s], 0.26879
    + Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Heating Supply Air Flow Rate [m3/s], 0.26879
      Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, User-Specified No Load Supply Air Flow Rate [m3/s], 0.00000
    - Component Sizing Information, Fan:OnOff, SPACE1-1 SUPPLY FAN, Design Size Maximum Flow Rate [m3/s], 0.22088
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Rated Air Flow Rate [m3/s], 0.22088
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Gross Rated Total Cooling Capacity [W], 5469.07698
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Gross Rated Sensible Heat Ratio, 0.69880
    + Component Sizing Information, Fan:OnOff, SPACE1-1 SUPPLY FAN, Design Size Maximum Flow Rate [m3/s], 0.26879
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Rated Air Flow Rate [m3/s], 0.26879
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Gross Rated Total Cooling Capacity [W], 5858.00167
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, Design Size Gross Rated Sensible Heat Ratio, 0.74025
     ! <DX Cooling Coil Standard Rating Information>, Component Type, Component Name, Standard Rating (Net) Cooling Capacity {W}, Standard Rated Net COP {W/W}, EER {Btu/W-h}, SEER {Btu/W-h}, IEER {Btu/W-h}
    - DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, 5297.4, 2.66, 9.07, 9.63, 9.51
    + DX Cooling Coil Standard Rating Information, Coil:Cooling:DX:SingleSpeed, SPACE1-1 PTAC CCOIL, 5649.2, 2.61, 8.92, 9.47, 9.28
      Component Sizing Information, Coil:Heating:Fuel, SPACE1-1 HEATING COIL, Design Size Nominal Capacity [W], 2899.74027

rraustad:
So it's sizing the heating coil based on the zone heating air flow rate (DesHeatMassFlow). This should probably size using the PTAC flow rate? But it's a fuel heating coil, you've told me many times that this type of coil shouldn't size the same as the DX. Shouldn't fix this here.

    ZoneHVAC:PackagedTerminalAirConditioner,
      SPACE1-1 PTAC,           !- Name
      Coil:Heating:Fuel,       !- Heating Coil Object Type
      SPACE1-1 Heating Coil,   !- Heating Coil Name

heatcoilsizing

rraustad:
Bear in mind I could set any of these, HeatingCapacity, DataCoolCap or DataFlowUsedForSizing, But we have decided to size the heating coil based on the zone load when that coil wasn't DX. 

                } else if (SizingType == HeatingCapacitySizing) {
                    if (ZoneEqSizing(CurZoneEqNum).HeatingCapacity) {
                        NominalCapacityDes = ZoneEqSizing(CurZoneEqNum).DesHeatingLoad;
                        if (DataFlowUsedForSizing > 0.0) {
                            DesVolFlow = DataFlowUsedForSizing;
                        }
                    } else if (DataCoolCoilCap > 0.0 && DataFlowUsedForSizing > 0.0) {
                        NominalCapacityDes = DataCoolCoilCap;
                        DesVolFlow = DataFlowUsedForSizing;
                    } else if (FinalZoneSizing(CurZoneEqNum).DesHeatMassFlow >= SmallMassFlow) {
                        if (DataFlowUsedForSizing > 0.0) {
                            DesVolFlow = DataFlowUsedForSizing;
                        }

mjwitte:
Regarding the PTAC heating coil sizing, I recall there have been some fixes to this over time.  I'm going to move that file to Chicago and see if it can hold setpoint for a year. That checks out OK, very similar loads not met before and after this. However, for zones 4 and 5, the heating coils are signficantly smaller now than current develop. Running a design day zone 5 is ok, but zone 4 is maxing out the heating coil.  So, PTAC may still need some work.

Here's some zips with results.  I haven't looked at the other system types this closely.  
[6712-ATMixerSizing-After.zip](https://github.com/NREL/EnergyPlus/files/2386959/6712-ATMixerSizing-After.zip)
[6712-ATMixerSizing-Before.zip](https://github.com/NREL/EnergyPlus/files/2386960/6712-ATMixerSizing-Before.zip)


Reviewing the err diffs.
PackagedTerminalHeatPumpVSAS and PackagedTerminalAirConditionerVSAS have new warnings:
```
+   ** Warning ** ZoneHVAC:PackagedTerminalHeatPump - air flow rate = 0.2148646 in fan object is less than the MSHP system air flow rate when cooling is required (0.2725907).
+   **   ~~~   **  The MSHP system flow rate when cooling is required is reset to the fan flow rate and the simulation continues.
+   **   ~~~   **  Occurs in ZoneHVAC:PackagedTerminalHeatPump = ZONE2PTAC
```
DOAToFanCoilSupply and DOAToWaterToAirHPSupply
```
+   ** Warning ** AirTerminal:SingleDuct:Mixer: SPACE5-1 DOAS AIR TERMINAL
+   **   ~~~   **  Supply side Air Terminal Mixer does not adjust zone equipment coil sizing and may result in inappropriately sized coils.
+   **   ~~~   **  Set Account for Dedicated Outdoor Air System = Yes in Sizing:Zone object for zone = SPACE5-1
```
These should get cleaned up, but I'm not going to hold this up for that. You can throw in a separate little PR for that if there aren't any other changes to do here.

mjwitte:
Probably should have some doc changes, too for this to explain the sizing difference for inletside vs supplyside, in the mixer object and in the Sizing:Zone DOAS fields.

mjwitte:
But zone 4 is a supply side mixer, it maxes out the heating coil on the winter design day before this changes as well.  Oh, I see, the DOAS inputs in Sizing:Zone don't work for Chicago, because it's an unconditioned DOAS.  So, these example files could use a little more tweaking.  Merging this.

mjwitte (approved):

mjwitte:
But it would still be good to add some notes in the I/O Ref about when an what sizing does with ATMixers now.

PR #6799, opened 2018-06-27
Refactor AirloopHVAC:UnitarySystem

rraustad:
Pull request overview
---------------------
Refactor existing AirLoopHVAC:UnitarySystem. Keep existing model for comparison until comfortable with this new model as AirLoopHVAC:UnitarySystem:Legacy.

 - [x] Initial development of class and functions
 - [x] Convert getInput to JSON
 - [x] Review and comments on design by others before proceeding
    - [ ] move *compPtr to generic location
    - [ ] how to add multiple cooling/heating/supheating coils in best way
    - [x] clean up PrimaryAirSystem loop in SimAirServingZones
    - [ ] add new getInput functions for JSON (e.g., getNumericInputData, getNumericFieldName, etc.)
 - [x] Pull necessary code from existing model into new streamline version
 - [x] Need for CompIndex or can get by with CompPtr?
 - [x] Unit tests along the way
 - [x] Other issues arising from comments
 - [ ] resolve use of "this->" as appropriate
 - [x] add Enums as needed
 - [x] incorporate changes and comments in #6600 
 - [x] include changes in #6757

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
This runs clean (because it doesn't do anything) with example file and unit test runs. Need help with design before proceeding.

src/EnergyPlus/DataAirSystems.hh
```diff
@@ -90,6 +92,7 @@ namespace DataAirSystems {
         std::string Name;        // Component name
         int CompType_Num;        // Numeric designator for CompType (TypeOf)
         int CompIndex;           // Component Index in whatever is using this component
+        UnitarySys *CompVecIndex;       // Component vector pointer to component class
```
rraustad: How do I move this to say DataHVACGlobals so all air loop equipment can use it?

src/EnergyPlus/MixedAir.cc
```diff
@@ -203,7 +203,7 @@ namespace MixedAir {
     int const Fan_ComponentModel(18); // cpw22Aug2010 (new)
     int const DXHeatPumpSystem(19);
     int const Coil_UserDefined(20);
-    int const UnitarySystem(21);
+    int const UnitarySystemHVAC(21);
```
rraustad: Changed existing AirloopHVAC:UnitarySystem type name to avoid collision with new class name

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -1286,8 +1291,13 @@ namespace SimAirServingZones {
                             PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).CompType_Num = DXHeatPumpSystem;
                         } else if (componentType == "COIL:USERDEFINED") {
                             PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).CompType_Num = CoilUserDefined;
+                        } else if (componentType == "UNITARYSYSTEM") {
+                            PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).CompType_Num = UnitarySystemModel;
+                            UnitarySys thisSys;
+                            PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).CompVecIndex =
+                                thisSys.factory(UnitarySystemModel, PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).Name);
```
rraustad: Factory call in GetAirPathData

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3465,7 +3475,22 @@ namespace SimAirServingZones {
...
 
-            } else if (SELECT_CASE_var == UnitarySystem) { // 'AirLoopHVAC:UnitarySystem'
+            } else if (SELECT_CASE_var == UnitarySystemModel) { // 'UnitarySystem'
+                // set up a reference for this component
+                bool foundComp = false;
+                for (int branchNum = 1; branchNum <= PrimaryAirSystem(AirLoopNum).NumBranches; ++branchNum) {
+                    for (int compNum = 1; compNum <= PrimaryAirSystem(AirLoopNum).Branch(branchNum).TotalComponents; ++compNum) {
+                        if (CompName == PrimaryAirSystem(AirLoopNum).Branch(branchNum).Comp(compNum).Name) {
+                            auto &sim_component(PrimaryAirSystem(AirLoopNum).Branch(branchNum).Comp(compNum));
+                            sim_component.CompVecIndex->simulate(CompName, FirstHVACIteration);
```
rraustad: ->simulate call in SimAirLoopComponent

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,137 @@
...
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR
+// OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+// POSSIBILITY OF SUCH DAMAGE.
+
+#include <UnitarySystem.hh>
+#include <DataIPShortCuts.hh>
+#include <InputProcessing/InputProcessor.hh>
+#include <SimAirServingZones.hh>
+#include <UtilityRoutines.hh>
+
```
rraustad: And then lots of help in here to get this right and tight

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,87 @@
...
+
+namespace UnitarySystems {
+
+    class UnitarySys {
+
+        //bool myOneTimeFlag;
+        //bool getInputOnceFlag;
+
+    public:
+        UnitarySys *compPtr;
```
rraustad: I'd like to move this compPtr to somewhere where all HVAC equipment could access it. DataHVACGlobals?

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,137 @@
...
+namespace EnergyPlus {
+namespace UnitarySystems {
+
+    bool myOneTimeFlag(true);
+    bool getInputOnceFlag(true);
+    std::vector<UnitarySys> unitarySys;
+
+    void UnitarySys::simulate(std::string const & objectName, bool const firstHVACIteration)
+    {
+        UnitarySys::init(firstHVACIteration);
```
rraustad: I'm not comfortable with the "this" thing. Is this where this->init() can be used? And then all the member variables are this->InletNodeNum, etc.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    UnitarySys * UnitarySys::factory(int object_type_of_num, std::string const objectName)
+    {
+        if (getInputOnceFlag) {
+            UnitarySys mySys;
+            mySys.getInput();
+            getInputOnceFlag = false;
+        }
```
rraustad: I don't understand why I had to do this when the EIR WWHP did it this way:

    if (getInputsWWHP) {
         EIRWaterToWaterHeatPump::processInputForEIRWWHP();
         getInputsWWHP = false;
    }

Figured this out, make the getInput function static.

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,87 @@
...
+#define ENERGYPLUS_UNITARYSYSTEM_HH
+
+#include <string>
+#include <vector>
+
+namespace EnergyPlus {
+
+namespace UnitarySystems {
+
+    class UnitarySys {
```
rraustad: Chose UnitarySys to avoid collision with AirloopHVAC:UnitarySystem struct array named UnitarySystem

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4202 @@
...
+                    //}
+                    // if (fields.find("cooling_speed_" + std::to_string(speedNum+1) + "_supply_air_flow_ratio") != fields.end()) { // not required
+                    // field
+                    //    thisDesignSpec.MSCoolingSpeedRatio[speedNum] =
+                    //        fields.at("cooling_speed_" + std::to_string(speedNum+1) + "_supply_air_flow_ratio");
+                    //}
+                    // this find does work and finds "flow_ratios"
+                    if (fields.find("flow_ratios") != fields.end()) {
+                        // but how to get the data for heating and cooling speeds ratios?
+                        // thisDesignSpec.MSHeatingSpeedRatio[speedNum] = fields.at.array[speedNum]("heating_speed_supply_air_flow_ratio");
```
rraustad: @Myoldmopar how do I read in the cooling/heating speed data from these extensible fields? These data show up under "flow_ratios" in the field reference. The strings e.g. "Heating Speed 1 Supply Air Flow Ratio" are not there but there is an array of e.g. "Heating Speed Supply Air Flow Ratio" but I don't know how to access the data.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4202 @@
...
+
+                    Real64 loc_coolingSAFMethod_SAFlow(-999.0);
+                    if (fields.find("cooling_supply_air_flow_rate") != fields.end()) { // not required field, autosizable
+                        std::string tempFieldVal = UtilityRoutines::MakeUPPERCase(fields.at("cooling_supply_air_flow_rate"));
+                        if (UtilityRoutines::SameString(tempFieldVal, "AutoSize")) {
+                            loc_coolingSAFMethod_SAFlow = DataSizing::AutoSize;
+                        } else {
+                            loc_coolingSAFMethod_SAFlow = fields.at("cooling_supply_air_flow_rate");
+                        }
+                    }
```
rraustad: @Myoldmopar this seems like a lot of overhead in getInput functions. Wouldn't it be better to write one or more functions in InputProcessor that handle all the subtleties of getting JSON input data?

One function:
getJSONInput(enum fieldType (e.g., alpha or numeric), std::string fieldString);

    thisSys.coolSAFlowRate = getJSONInput(InputProcessor::numericFieldTypeEnum, "supply_air_flow_rate");

Multiple functions:
getJSONAlphaInput(std::string fieldString);
getJSONNumericInput(std:string fieldString);

Also need a function to return the name of the field:
getJSONFieldName(std::string fieldString);

AirLoopHVAC:UnitarySystem,
  A2,  \field Control Type

    ShowWarningMessage("When "+ getJSONFieldName("control_type") + " = " + getJSONAlphaInput("control_type") + " blah blah ";

Not sure how this will work with the new warning message effort.

src/EnergyPlus/UnitarySystem.cc
rraustad: @mbadams5 can you school me on how to JSON read the Heating/Cooling Speed x Air Flow Ratio data in the UnitarySystemPerformance:Multispeed object? This branch compiles so you can test right here if you like. I've tried and am just spinning my wheels.

rraustad: @mbadams5 this is in UnitarySystem.cc at line 228.

rraustad: @mbadams5 never mind @Myoldmopar found a method.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4296 @@
...
+            //    CheckEquipName(UnitarySysNum) = false;
+            //}
+        }
+
+        if (!this->myGetInputSuccessfulFlag) {
+            // Need to do inputs again for this and any others
+            UnitarySystems::getInputFlag = true;
+            this->getUnitarySystemInput();
+            UnitarySystems::getInputFlag = false;
+        }
```
rraustad: @EnergyArchmage I have a feeling this has something to do with calling during sizing. Line 177-181 of UnitarySystem.cc on this branch (or 397 - 402 of HVACUnitarySystem.cc). Why would a system not be gotten when called the first time? Or maybe it's zone name, index, etc. that needs to be gotten again? So are the HVAC systems being called during sizing and if yes, then why?

    if (!AirLoopFound && !ZoneEquipmentFound && !OASysFound && !DataHVACGlobals::GetAirPathDataDone) {
       // Unsucessful attempt
        continue;
    } else {
        MyGetInputSuccessfulFlag(UnitarySysNum) = true;
    }

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4296 @@
...
+                        }
+                        auto heatingSpeedRatioObject = flowRatio.at("heating_speed_supply_air_flow_ratio");
+                        if (heatingSpeedRatioObject == "Autosize") {
+                            thisDesignSpec.MSHeatingSpeedRatio[speedNum] = -99999;
+                        } else {
+                            thisDesignSpec.MSHeatingSpeedRatio[speedNum] = heatingSpeedRatioObject;
+                        }
+                    }
+                }
+                designSpecMSHP.push_back(thisDesignSpec);
```
rraustad: @Myoldmopar here's where I ended up. This seems good.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3465,7 +3474,25 @@ namespace SimAirServingZones {
...
+                            bool ZoneEquipFlag = false;
+                            sim_component.compPointer->simulate(CompName, FirstHVACIteration, AirLoopNum, CompIndex, HeatingActive, CoolingActive, OAUnitNum, OAUCoilOutTemp, ZoneEquipFlag);
+                            foundComp = true;
+                            break;
+                        }
+                    }
+                    if (foundComp) break;
+                }
+
+            } else if (SELECT_CASE_var == UnitarySystemHVAC) { // 'AirLoopHVAC:UnitarySystem'
```
rraustad: Don't need this loop. Just call the object.

src/EnergyPlus/UnitarySystem.cc
mjwitte: @rraustad This is not because of sizing, this is because unitary system can be both zone equipment and airloop equipment.  This was a hack to get it working when I added support for multiple airloops service a zone (I think that's when it became a problem.  What needs to happen here is for the input data processing to be separated from the init-type checks that go looking for matching nodes to figure out if this is zone equipment or airloop equipment and then it does all sorts of other checks.

rraustad: Got it. It may be necessary to call a getInput multiple times to get all the data.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4299 @@
...
+        if (instances == inputProcessor->epJSON.end()) {
+            ShowSevereError("getUnitarySystemInputData: did not find UnitarySystem object in input file. Check inputs");
+            errorsFound = true;
+        } else {
+            auto &instancesValue = instances.value();
+            for (auto instance = instancesValue.begin(); instance != instancesValue.end(); ++instance) {
+                
+                auto const &thisObjectName = instance.key();
+                // only get the current data
+                if (!UtilityRoutines::SameString(objectName, thisObjectName) || getUnitarySystemIndex(objectName) > -1) continue;
```
rraustad: this allows multiple calls, if not thisOjbectName or it's already gotten then cycle. So get only 1 data set. Still need to resolve UnitarySystems::getInputOnceFlag and will probably make it a local to UnitarySys.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,4296 @@
...
+                thisDesignSpec.MSCoolingSpeedRatio.resize(maxSpeeds);
+
+                auto speedFlowRatios = fields.find("flow_ratios");
+                if (speedFlowRatios != fields.end()) {
+                    auto flowRatioArray = speedFlowRatios.value();
+                    int speedNum = -1;
+                    for (auto flowRatio : flowRatioArray) {
+                        speedNum += 1;
+                        auto coolingSpeedRatioObject = flowRatio.at("cooling_speed_supply_air_flow_ratio");
+                        if (coolingSpeedRatioObject == "Autosize") {
```
Myoldmopar: @mbadams5 What's the best way to check if a field is autosized?  I mean this works for this unit test, and it could be made case insensitive if needed, but is there a better way?

src/EnergyPlus/UnitarySystem.cc
mjwitte: I think it would be better to call getinput only once, and fill all of the input data, but separate out the various checks that go looking at node connections into a separate init function which only tries to initialize the currently called piece of equipment.  That way it knows if CurZoneEquip or AirloopNum is active and it should have the information it needs at that point.

src/EnergyPlus/UnitarySystem.cc
mbadams5: You shouldn't need to make it case insensitive. If the field is Autosizable then it should always be "Autosize" since the idf->epJSON does this automatically and the schema validation will throw an error if it isn't "Autosize".

This way is the appropriate method for right now, however, there might be a better way to check if a field is autosized/autocalculated. I need to spend a little bit of time thinking through a better approach (likely a function within InputProcessor).

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,12162 @@
...
+                    1.0,
+                    initUnitarySystemsQActual); // QCoilReq, simulate any load > 0 to get max capacity of steam coil
+                this->designHeatingCapacity = SteamCoils::GetCoilCapacity(
+                    DataHVACGlobals::cAllCoilTypes(this->heatingCoilType_Num), this->heatingCoilName, initUnitarySystemsErrorsFound);
+
+            } // from IF(UnitarySystem(UnitarySysNum)%HeatingCoilType_Num == Coil_HeatingSteam) THEN
+            if (this->suppHeatCoilType_Num == DataHVACGlobals::Coil_HeatingWater) {
+
+                //     set steam-side mass flow rates
+
```
rraustad: @Myoldmopar I need your help here. Put a break point at 9150 of UnitarySystems.cc and show me how to call SolveRoot. Par is now a std::vector and the compiler is telling me to use a reference to the function. I don't understand. Uncomment 9151 to test.

rraustad: @mbadams5 how do I access the field names of objects for sizing and warning messages? I want e.g.,  "Cooling Supply Air Flow Rate Per Floor Area". Thank you in advance.

    "field_info": {
        "cooling_supply_air_flow_rate_per_floor_area": {
            "field_type": "n", 
            "field_name": "Cooling Supply Air Flow Rate Per Floor Area"
     },

rraustad: I hope it was obvious that running the unit test gets you here.

rraustad:
This is weird. I sent 2 comments at different parts of code and this conversation page shows both at line 1029 of UnitarySystem.cc. Not even close.

src/EnergyPlus/UnitarySystem.cc
rraustad: @mbadams5 here's an example of where I need to access field names. Line 1401 of UnitarySystem.cc on this branch. This field is cooling supply air flow rate so I need std:string thisFieldName = N3fieldString.

    N3 , \field Cooling Supply Air Flow Rate
       \type real
       \units m3/s
       \minimum 0.0
       \autosizable
       \note Enter the magnitude of the supply air volume flow rate during cooling operation.
       \note Required field when Cooling Supply Air Flow Rate Method is SupplyAirFlowRate.
       \note This field may be blank if a cooling coil is not included in the unitary system.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -183,7 +184,7 @@ namespace SimAirServingZones {
     int const Desiccant(16);
     int const Unglazed_SolarCollector(17);
     int const EvapCooler(18);
-    int const UnitarySystem(19);
+    int const UnitarySystemHVAC(19);
```
jasondegraw: Any chance we can change this into an `enum class` while we're here? Maybe it's out of scope, but...

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3465,7 +3474,25 @@ namespace SimAirServingZones {
             } else if (SELECT_CASE_var == CoilUserDefined) { // Coil:UserDefined
                 SimCoilUserDefined(CompName, CompIndex, AirLoopNum, HeatingActive, CoolingActive);
 
-            } else if (SELECT_CASE_var == UnitarySystem) { // 'AirLoopHVAC:UnitarySystem'
+            } else if (SELECT_CASE_var == UnitarySystemModel) { // 'UnitarySystem'
```
jasondegraw: If `UnitarySystemModel` was in an `enum`, then this could be `switch` rather than a big `if` forest.

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,684 @@
...
+    extern int const On;  // normal compressor operation
+    extern int const Off; // signal DXCoil that compressor shouldn't run
+
+    // Coil type for SimWater and SimSteamCoil
+    extern int const CoolingCoil;
+    extern int const HeatingCoil;
+    extern int const SuppHeatCoil;
+
+    // Supply Air Sizing Option
+    extern int const None;
```
jasondegraw, line 84: Would be great if all these `extern int`s could be `enum`s, e.g. 

`
enum class SupplyAirSizing { None,  SupplyAirFlowRate, FlowPerFloorArea, FractionOfAutoSizedCoolingValue, FractionOfAutoSizedHeatingValue, FlowPerCoolingCapacity, FlowPerHeatingCapacity};
`

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,684 @@
...
+        DesignSpecMSHP(); // constructor
+        ~DesignSpecMSHP() // destructor
+        {
+        }
+
+        std::string name;
+        static DesignSpecMSHP *factory(int object_type_of_num, std::string const objectName);
+
+    private:
+        int designSpecMSHPType_Num;
```
jasondegraw: Prefer that private member variables be prefixed with `m_`, e.g. `m_designSpecMSHPType_Num`

rraustad:
I don't understand what is happening. I put a comment at a line and this conversation representation has no relation to where I placed the comment. I don't care if Git ... Git's it, but I do need help from you guys. I f you want to help just let me know.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,12162 @@
...
+          HXAssistedLatPLRIterIndex(0), HXAssistedLatPLRFail(0), HXAssistedLatPLRFailIndex(0), HXAssistedCRLatPLRIter(0),
+          HXAssistedCRLatPLRIterIndex(0), HXAssistedCRLatPLRFail(0), HXAssistedCRLatPLRFailIndex(0), HXAssistedCRLatPLRFail2(0),
+          HXAssistedCRLatPLRFailIndex2(0), SensPLRIter(0), SensPLRIterIndex(0), SensPLRFail(0), SensPLRFailIndex(0), LatPLRIter(0),
+          LatPLRIterIndex(0), LatPLRFail(0), LatPLRFailIndex(0), HeatCoilSensPLRIter(0), HeatCoilSensPLRIterIndex(0), HeatCoilSensPLRFail(0),
+          HeatCoilSensPLRFailIndex(0), SuppHeatCoilSensPLRIter(0), SuppHeatCoilSensPLRIterIndex(0), SuppHeatCoilSensPLRFail(0),
+          SuppHeatCoilSensPLRFailIndex(0), DXCoilSensPLRIter(0), DXCoilSensPLRIterIndex(0), DXCoilSensPLRFail(0), DXCoilSensPLRFailIndex(0),
+          MSpdSensPLRIter(0), MSpdSensPLRIterIndex(0), MSpdCycSensPLRIter(0), MSpdCycSensPLRIterIndex(0), MSpdLatPLRIter(0), MSpdLatPLRIterIndex(0),
+          MSpdCycLatPLRIter(0), MSpdCycLatPLRIterIndex(0), LatMaxIterIndex(0), LatRegulaFalsIFailedIndex(0), lowSpeedCoolFanRatio(0.0),
+          lowSpeedHeatFanRatio(0.0)
```
jasondegraw: This is one long initialization  list. Can any of these be packaged up in structs? Looks likes a lot of these could be grouped into structs.

src/EnergyPlus/UnitarySystem.cc
jasondegraw: There's an awful lot of `this->` going on in here. Do we really need those?

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,684 @@
...
+    class UnitarySys
+    {
+
+        // bool myOneTimeFlag;
+        // bool getInputOnceFlag;
+        int fanSpeedRatio;
+
+        enum controlTypeEnum : int
+        {
+            controlTypeNone,
```
jasondegraw: First, I like the use of `enum`s here, but use an `enum class` instead and drop the `controlType` prefix from the members and `enum` from the name. Also, capitalize (`ControlType`, `None`, `Load`, etc.).

```
enum class ControlType : int
{
    None,
    TypeLoad,
    Setpoint,
    CCMASHRAE
};
```

src/EnergyPlus/UnitarySystem.hh
```diff
@@ -0,0 +1,684 @@
...
+        );
+
+        void calcUnitarySystemToLoad(int const AirLoopNum,              // index to air loop
+                                     bool const FirstHVACIteration,     // True when first HVAC iteration
+                                     Real64 const CoolPLR,              // operating cooling part-load ratio []
+                                     Real64 const HeatPLR,              // operating cooling part-load ratio []
+                                     Real64 &OnOffAirFlowRatio,         // ratio of heating PLR to cooling PLR (is this correct?)
+                                     Real64 &SensOutput,                // sensible capacity (W)
+                                     Real64 &LatOutput,                 // latent capacity (W)
+                                     Optional_bool HXUnitOn = _,        // Flag to control HX for HXAssisted Cooling Coil
```
jasondegraw: I'd personally prefer that we start moving new code away from these ObjexxFCL `Optional`s.

jasondegraw (changes requested):
Generally looks good to me. Would like to see more `enum` related clean-up, further collection of info into structs, and less use of `this->`. We should only use `this->` when absolutely necessary.

rraustad:
Thank you @jasondegraw for the review. I don't understand everything you mentioned but will address things I do get. What I want here is to get this to compile and run, then address issues such as you describe. So right now it's the conversion to object oriented, then we can address the issues you see here. I'm happy to do that, just want to get this darn thing to run.

src/EnergyPlus/UnitarySystem.cc
rraustad: This is one of the things I don't get. I don't know what it means to package into structs. I'll work on this later and will need your guidance to get this right.

src/EnergyPlus/UnitarySystem.cc
rraustad: I think "this" is something we can discuss on a call. I thought this is what was expected and maybe I got it wrong. From my perspective I don't care either way, I'm just happy it's running and working( well not now since I added in other functions that don't work - e.g., calcSZVAVModel).

src/EnergyPlus/UnitarySystem.hh
rraustad: Yes I agree. As long as the int representation of these is the same as other models expect. So maybe it's None=0 (I think that's default so I maybe overthinking this), SupplyAirFlowRate=1,etc. I just want to make sure this new code (i.e., these enum's) is compatible with existing.

src/EnergyPlus/SimAirServingZones.cc
rraustad: Nothing is out of scope with this change. The problem is that I don't understand your point. Are you saying that this list should be an Enum? I'm OK with that but have a sheet load of work here to get this working. I am certainly open to this.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3465,7 +3474,25 @@ namespace SimAirServingZones {
...
+                // set up a reference for this component
+                bool foundComp = false;
+                for (int branchNum = 1; branchNum <= PrimaryAirSystem(AirLoopNum).NumBranches; ++branchNum) {
+                    for (int compNum = 1; compNum <= PrimaryAirSystem(AirLoopNum).Branch(branchNum).TotalComponents; ++compNum) {
+                        if (CompName == PrimaryAirSystem(AirLoopNum).Branch(branchNum).Comp(compNum).Name) {
+                            auto &sim_component(PrimaryAirSystem(AirLoopNum).Branch(branchNum).Comp(compNum));
+                            int OAUnitNum = 0;
+                            Real64 OAUCoilOutTemp = 0.0;
+                            bool ZoneEquipFlag = false;
+                            sim_component.compPointer->simulate(CompName, FirstHVACIteration, AirLoopNum, CompIndex, HeatingActive, CoolingActive, OAUnitNum, OAUCoilOutTemp, ZoneEquipFlag);
```
rraustad: @Myoldmopar I am not doing this right. When I get to the simulate routine "this->" isn't pointing to the correct system. It's as if it is pointing to a new system. In fact getInput is called again and now unitarySys.size() = 3 when there are only 2 systems in the input file. This is at line 3487 of SimAirServingZones. The factory is called on line 1297.

I tried this method in the factory with the same result. Any suggestions?

    int sysNum = -1;
    for (auto &sys : unitarySys) {
        ++sysNum;
        if (UtilityRoutines::SameString(sys.name, objectName) && sys.unitarySystemType_Num == object_type_of_num) {
            if (unitarySys[sysNum].myGetInputSuccessfulFlag) unitarySys[sysNum].thisSysInputShouldBeGotten = false;
            unitarySys[sysNum].unitarySysNum = sysNum;
            return &sys;
        }
    }

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -0,0 +1,13534 @@
...
+                    }
+                }
+                designSpecMSHP.push_back(thisDesignSpec);
+            }
+        }
+    } // namespace UnitarySystems
+
+    UnitarySys *UnitarySys::factory(int object_type_of_num, std::string const objectName)
+    {
+        if (UnitarySystems::getInputOnceFlag) {
```
rraustad: @Myoldmopar if you do try to run this you will have to comment out line 479 (getInputOnceFlag) since getInput now only gets 1 unitary system. Maybe that has something to do with my pointer issue?

src/EnergyPlus/SimAirServingZones.cc
jasondegraw: @rraustad Yes, that's what I'm saying. It can be a huge pain doing the changeover, but it is worth it once the change is made.

src/EnergyPlus/UnitarySystem.cc
jasondegraw: @rraustad I'm seeing a lot of args with the same prefix (e.g. `MSpdSensPLRIter`), and for me those are things that are candidates for being put into a struct together rather than just floating around loose with the name prefix connecting them together. Does that make sense?

src/EnergyPlus/SimAirServingZones.cc
rraustad: This could be a switch with either int or enum, correct? I still get the point on enum's and am trying to use them more.

src/EnergyPlus/UnitarySystem.hh
rraustad: I'll be addressing this soon. Still working to get the code to run all the example files.

src/EnergyPlus/UnitarySystem.hh
rraustad: OK, this one should be easy to change. I don't know the difference between enum and enum class.

src/EnergyPlus/UnitarySystem.hh
rraustad: That was the last thing I did (3 or 4 days ago) and am now trying to correct the code to run with all the arguments.

src/EnergyPlus/UnitarySystem.cc
rraustad: Yes it does. I have been thinking about what to do here. All these ints (the *iter and *index vars) are only used for recurring warning messages, I'll be that makes you stop and think. So I could move all these to a warning struct and inherit? that in UnitarySys. Not sure how this is done but will give it a try when I get farther along.

src/EnergyPlus/DataAirLoop.hh
```diff
@@ -265,6 +266,7 @@ namespace DataAirLoop {
         Array1D_int ComponentType_Num; // Parameterized (see above) Component Types this
         // module can address
         Array1D_int ComponentIndex; // Which one in list -- updated by routines called from here
+        std::vector <UnitarySystems::UnitarySys *> compPointer;
```
rraustad, line 262: @Myoldmopar I have set up these pointers (1 each for zone, air loop, and OA system) to save the factory return. Is there a more generic way to do this so that other models can use this same vector to store their pointer? std::shared, std::unique, etc. See line 269 DataAirLoop.hh, line 397 DataZoneEquipment.hh, line 128 OutdoorAirUnit.hh

src/EnergyPlus/OutdoorAirUnit.hh
```diff
@@ -124,6 +125,7 @@ namespace OutdoorAirUnit {
         std::string ComponentType;
         int ComponentType_Num; // Parameterized Component Types this module can address
         int ComponentIndex;    // Which one in list -- updated by routines called from here
+        UnitarySystems::UnitarySys *compPointer;
```
rraustad, line 128: @Myoldmopar what is wrong with this declaration? line 128 OutdoorAirUnit.hh. There is no #include <UnitarySystem.hh> in the header file and Windows MSVS seems OK with that.

rraustad: @Myoldmopar here's how I am using that pointer:
OutdoorAirUnit.cc line 858:

    } else if (SELECT_CASE_var == "AIRLOOPHVAC:UNITARYSYSTEM") {
        OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentType_Num = UnitarySystemModel;
        UnitarySystems::UnitarySys thisSys;
        OutAirUnit(OAUnitNum).OAEquip(CompNum).compPointer = thisSys.factory(SimAirServingZones::UnitarySystemModel, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName, false);
        OutAirUnit(OAUnitNum).OAEquip(CompNum).compPointer->checkUnitarySysCoilInOASysExists(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName);
    } else if (SELECT_CASE_var == "AIRLOOPHVAC:UNITARYSYSTEM:LEGACY") {
        OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentType_Num = UnitarySystemHVAC;
        CheckUnitarySysCoilInOASysExists(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName);

rraustad: @jasondegraw do you have time to help me understand the nuance of pointer declaration? Warnings point to line 128 of OutdoorAirUnit.hh

If I add this to the OAEquipList struct do I need to add #include <UnitarySystem.hh> to the OutdoorAirUnit header file? I included that in OutdoorAirUnit.cc but don't know if I need it in both places. This compiles fine on my machines.

UnitarySystems::UnitarySys *compPointer;

jasondegraw: @rraustad I can try. I'm on my phone right now so it's a bit hard to see what's going on, but once I get back to my desk I can take a look. Are you seeing local warnings or CI warnings?

rraustad: The CI pointed to this line. It seems to run fine and I was expecting a lot of green on the CI. Instead I get a lot of red. I'd like to GoTo meet if possible.

rraustad:
@Myoldmopar these last few warnings on the Mac are throwing me. The constructor order looks right to me. Maybe I should move the vector and pointer declarations to the end of the class? The warning for function not used is confusing because I use getDesignSpecMSHPIndex in GetInput just like I do getUnitarySystemIndex (bottom of hh file). Maybe it's the static declaration on getDesignSpecMSHPIndex? I'll remove static and see if that fixes this warning. I am expecting 0 diff's now and would like to wrap up these last few items. Could you compile this on the Mac? You won't have my changes from today but should see these warnings?

I've got something else I want to test while I wait for your response.

rraustad:
Finally got the SZVAV results to match. I ran into the same thing @mjwitte ran into with timing of GetInput. This file has a ZoneHVAC and AirLoopHVAC UnitarySystem both configured as SingleZoneVAV. To make matters worse, the factory is called very early in the simulation. The air loop model has to call GetInput multiple times to fill out the zone data. First thing that happened was sizing was messed up after a call to GetInput, volume flows got reset to -99999. So sizing has to also be called again. And then there must be a delay in reporting or else the eio gets multiple lines of the same data. I'm sure there is a better way to do this than what I have done, but it's running.

refactoredszvav

mjwitte:
@rraustad I haven't looked at your code, but if the factory and the inits all just act on one unit at a time, it shouldn't be necessary to worry about timing. If it's called from zoneequipment that information should be setup in time. And then if another one is called from airloop, then that info should be setup at that point.

rraustad:
I tried just processing one unit at a time. What happened was the vector unitarySys[] got adjusted/reallocated/something where the pointer returned by the factory was not pointing to the right place in memory AFTER the next unit was processed (i.e., "this" pointed to the correct data perfectly the first time though but that data got scrambled after the next push_back). So to counter that, the first time through I read a few inputs from all units, namely the name, and push-back the entire vector. And THEN I pass back the first pointer. The next time through only processes the current unit (the current unitarySys[]). Works like a charm, yet still it takes a few passes to get all the data for the air loop units, and I assume the OA units. As you know, this was really tricky but I'm sure it will work itself out with more thought. If I recall correctly ZoneInletNodeNum was coming back as 0. Maybe wait for a flag like ZoneEquipmentGotten or something like that. If not true, just pass inlet to outlet and return out without any calculations. A few passes, then get all the data. Sizing waits for flags to be set before writing to the eio, but I don't want to put IF statements throughout the code, would rather just return back and wait.

mjwitte:
One option then would be to let the factory instantiate all of the objects by filling a data array with the raw input fields, but do nothing more at that point. Then actually process the input one-by-one.

mjwitte:
Didn't really mean a "data array" exactly just simply object variables that hold the input values. Similar to what's in the new dx coil work.

rraustad:
Interesting, there are no diff's now in the results. But this annoying error file warning is showing up... x unused objects, Use Output:Diagnostics,DisplayUnusedObjects; to see them. I thought I wasn't up to date with develop. Then I ran one of those files and yup, it's this branch.

    ** Warning ** The following lines are "Unused Objects".  These objects are in the input
    **   ~~~   **  file but are never obtained by the simulation and therefore are NOT used.
    **   ~~~   **  Only the first unused named object of an object class is shown.  Use Output:Diagnostics,DisplayAllWarnings; to see all.
    **   ~~~   **  See InputOutputReference document for more details.
    ************* Object=AirLoopHVAC:UnitarySystem=HeatPump DX Coil 1

So how could this object be unused when there are no results diff's ?

@mbadams5 the `UnorderedUnusedObjectMap unusedInputs` array is handled in InputProcessor::getObjectItem which I and @Myoldmopar  are not using in the refactored code. Our getInput is all Json based. So how do I (we) get around this? I could just call getObjectItem for now and not use those results until this is worked out.

rraustad:
This branch now has no diff's with develop. The one file that shows diff's is because of a shift in rows in the html. That is obvious since the eio has no diff's but the html does. I am showing here this branch results for Sys 1 Cool Coil  with develop (the "parallel" branch in the URL) to show they are the same. I think the CI script is comparing rows without checking for coil name/type.

hvactemplateresults

So now this branch has no diff's given the unit tests and example files available.. Now there can be a discussion of what else needs to be done here. @jasondegraw @Myoldmopar @mjwitte what else needs to happen here? I am still concerned that this branch behaves EXACTLY as the original model since optional function arguments were eliminated (and I need to replicate exactly what happened with and without those arguments) but these results look good and we have to start somewhere.

There are 2 things left in my mind. One is organization of header file, which in my mind is low priority since I don't think it matters that much other than readability (and my knowledge of c++), and two is the application of "this->'. I would like to get this checked in and then on a new branch eliminate all occurrences of  "this->" and see if there is a speed impact. If there is no speed impact we can debate the use of "this->", if there is a speed impact, and it's positive, well then, there is no debate. I don't expect a speed impact.

Regardless of what is needed here I would like to get this into develop so I can compare any changes to the design against this implementation with new commits.

rraustad:
The other diff I can't explain is HVACStandAloneERV_Economizer where this example file does not have an AirLoopHVAC:UnitarySytsem object. I have reviewed the diff's and see that there is an issue with conversion to vector based with HVACStandAloneERV (i.e., 0 based vs 1 based, and actually it's HeatRecovery that causes the extra eio write) but I can't find where this branch causes those diff's. I could be wrong here but that's where I stand now.

Here the supply fan index = 0 since it is a Fan:SytemModel and the supply fan does not get called here. What I don't understand is why when this fan is eventuyally called it doesn't write to the eio. Maybe it never gets called? But then wouldn't the eso results be different? There is also a duplicate eio line of User Defined flow, which I see happening but don't know why this branch causes it.

    if (StandAloneERV(StandAloneERVNum).SupplyAirFanIndex > 0) {
        if (!(StandAloneERV(StandAloneERVNum).SupplyAirFanType_Num == DataHVACGlobals::FanType_SystemModelObject)) {
            SimulateFanComponents(StandAloneERV(StandAloneERVNum).SupplyAirFanName, true, StandAloneERV(StandAloneERVNum).SupplyAirFanIndex);
        } else {
            HVACFan::fanObjs[StandAloneERV(StandAloneERVNum).SupplyAirFanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
        }
    }
    if (StandAloneERV(StandAloneERVNum).ExhaustAirFanIndex > 0) {
        if (!(StandAloneERV(StandAloneERVNum).ExhaustAirFanType_Num == DataHVACGlobals::FanType_SystemModelObject)) {
            SimulateFanComponents(StandAloneERV(StandAloneERVNum).ExhaustAirFanName, true, StandAloneERV(StandAloneERVNum).ExhaustAirFanIndex);
        } else {
            HVACFan::fanObjs[StandAloneERV(StandAloneERVNum).ExhaustAirFanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
        }
    }

Myoldmopar:
@rraustad, this looks great.  It's awesome to see the results lining up so well.  

- We should have a chat to discuss whether to keep the original unitary code in place.  I know you have said you want to keep it in there to allow easy comparisons later on, but we should figure out what that means.  Like would you still maintain that code at all?  And if we do keep it in there, when do we get rid of it?
- As for the diffs, I don't really have much to add for the one difference you are seeing in the standalone ERV code.  If/when we do have our chat, we should discuss strategies for addressing that one remaining issue.
- As for the code, I wanted to discuss the `this->` usage.  I am fully in favor of using it.  For your reference it **does not** add a performance hit to use it.  In fact, it is always defined, as a hidden variable available in the local scope of all non-static member functions of a class.  And if you don't put the word `this->` to access a member variable or method, the compiler essentially puts it there for you when dereferencing the current class instance.  I love the intentionality of using it, because it makes it fully clear that you are accessing something on the class, and not something that is local to the function.  I think that is great, especially as we start really optimizing the EnergyPlus codebase over the next couple years, where we will want to compartmentalize the code as much as possible.  After the C++ translation, we ended up with many, many `using` statements, that made the code difficult to read because you couldn't tell the difference between `used` variables and local variables without investigating further.  We pushed developers to eliminate that and just be explicit when accessing a variable from outside the local scope.  The code is a bit longer, but it is very clear, and very intentional, and it has been great.  I don't think we should push people to avoid `this->` just because of style or preference, when it does provide a benefit to mentally processing and understanding the code as you read it.  And finally, having it in there allows us to ultimately get rid of it very easily: `find . -type f -name '*.cc' | xargs sed -i 's/this->//g'`.  If we get rid of them first, then decide they would be nice to have, it's only like 1000 times as hard to add them back in.  So that's my list of pros, I don't have any cons except that you see the word `this->` a bunch of times.

rraustad:
Thanks for the review. So it sounds as if the legacy code and 1 errant example file are the remaining issues. My take is that the UnitarySystem is one of the more important parent models in E+. There has been a lot of time and effort to improve the performance wrt specific coil types. I am just hesitant to punt on that old code in case I have made a mistake somewhere during refactoring. The biggest issue is I eliminated the optionals. I found it very easy to fix issues by running the same file through the refactored code and the legacy code at the same time, in 2 environments, step by step until I saw a diff. Then boom, the diff's were gone with a simple code change. A simple work around is to add ":Legacy" to the object type and your right back to the original model. And yes I understand the issue with this hanging around. We're not immediately getting rid of all the other cooling coil models when the new coil model goes in are we? Other than that I guess I could let go but fear I'd be right back to single stepping for hours to figure something out that was a simple error on my part during refactoring. So I would think a release or 2 with this hanging around couldn't hurt anything. I guess it's your call, but I still need to get any changes to HVACUnitarySystem.cc over to the new UnitarySystem.cc. Also, I'm out on vacation next week.

Myoldmopar:
OK, let's make sure to wrap this up by this week so it doesn't linger.  If you can make sure the code is basically ready to drop in by Thursday (try to figure out the remaining issue, bring any other recent updates from "legacy"), we should have a chat (you, me, @mjwitte, and @EnergyArchmage @jasondegraw if interested) to lay out a final plan.  Then you could take any final action items there and implement them Thursday/Friday, and we could merge Friday.

mjwitte:
@rraustad Tested the defect files from #6750 (PR #6757). You may already know this . . .
6750-DOAToUnitarySystem_SPControl zone control looks good, but there are small diffs compared to the 6757 results.
6750-UnitarySystem_HeatPumpAuto_LoadControl crashes

rraustad:
6750-UnitarySystem_HeatPumpAuto_LoadControl ran fine just now. I must have fix what caused the crash today.

   ** Severe  ** Input errors for AirLoopHVAC:UnitarySystem:DXAC HEAT PUMP 1
   **   ~~~   ** Controlling Zone or Thermostat Location cannot be blank when Control Type = Load or SingleZoneVAV

rraustad:
I do see slight diff's in the results for the #6750 SP control. I also used Chicago weather, I think. Here is what I see now, it actually looks a little better.

spcontrol_chicago

mjwitte:
@rraustad The original test file with a blank zone name runs OK on this branch, but I had another one with a non-existent zone name, and it crashes.
[6750-UnitarySystem_HeatPumpAuto_LoadControl-AfterRefactor-MJW.idf.txt](https://github.com/NREL/EnergyPlus/files/2334342/6750-UnitarySystem_HeatPumpAuto_LoadControl-AfterRefactor-MJW.idf.txt)

mjwitte:
And getting identical results now vs develop with the SP control test file, in Miami.

rraustad:
For a reason I don't know yet, the fan is not being found.

crash

rraustad:
Wow, then I single step it and it does find the fan .......

notcrash

Myoldmopar:
Funny how things can start becoming ominous when developing in the middle of the night.  I've got a real head scratcher myself but I'm throwing up the white flag for a few hours until morning.  Hope you get your fan issue squared up.

rraustad:
I am getting much more confident with this refactor with each unit test pulled over. They just work and give the same answer. So I am ready to chat along with @mjwitte, @EnergyArchmage and @jasondegraw, and anyone else who is watching this, if interested. I assume we can use the same GoTo meeting as last time.

jasondegraw:
@Myoldmopar It's not 100% true that the `this->` can always be removed. Granted, it's highly unlikely that would happen here, but that's where my preference for using it only when necessary comes from.

rraustad:
@Myoldmopar I just realized that outstanding PRs affecting HVACUnitarySystem need to be merged before I eliminate the legacy model. PR #6753 and #6788 are blockers (more?). There will be a serious merge conflict if I delete the old model.

Myoldmopar:
Since this and both of those PRs are yours, I will give you the liberty of choosing a path forward:
1. You can wrap this one up today as is (there are build warnings and a failing unit test still) and merge it in.  This will leave those PRs basically orphaned because you'll have to reimplement those fixes in the new codebase which will pretty much be a "start-over" situation for those.
2. You could merge those PRs in ahead of this big one.  This will result in this PR being put on hold for a bit, and essentially those same merge conflicts will arise when you try to then pull develop into this branch.
3. You could just bring in those fixes into this branch and close those other PRs.  That may introduce some diffs in this branch from those fixes, but they would be acceptable as long as this branch was clean before that.  And you'd essentially be doing the same thing in steps 1 or 2 anyway when dealing with merge conflicts.
4. We can just sorta take a few days to breath and evaluate and handle this when you get back so we don't rush something in that could cause problems.  (I doubt anyone else will be significantly touching unitary system for the next few days.)

Let me know which approach you'd like to take.

rraustad:
#6753 touched 40 files and #6788 touched 24. I'm already on overtime here so I choose to breath a little. If these blockers get in while I'm away, the effort to merge just the HVACUnitarySystem changes into this branch is probably a few hours work. I've fixed the warnings failing unit test, pulled in the remaining unit tests, and addressed comments from yesterday's cc. Hopefully the next commit will be clean and ready for the next step(s).

rraustad:
I spent hours looking for a reason for the unit test failure. These unit tests run in debug mode and release mode individually. Also runs in debug mode when all unit tests are run together. But UnitarySystemModel_ASHRAEModel_WaterCoils fails in release mode when all unit tests are run together. The other unit test, UnitarySystemModel_MultispeedDXCoilSizing, only fails on i386 Windows. I'm at a loss for why these are failing.

rraustad:
Persistence does pay off. I inadvertently pulled this unit test twice. I could easily just delete the second, but then I wouldn't know why the second one failed. I added print statementsto help narrow this down and that's how I realized there were 2 of the same unit test. I don't have an answer yet but an sure I can figure this one out.

szvavunittesting

rraustad:
One down one to go. Although I am out of time at the moment and will pick this up when I return.

nrel-bot-3:

rraustad:
@Myoldmopar this branch now has the old HVACUnitarySystem removed and is looking good regrading CI results.

rraustad:
Interesting warning I think I'll leave as is for now:

    src/EnergyPlus/DataSizing.hh:130:25    unused variable 'EnergyPlus::DataSizing::AutoSize' [-Wunused-variable]

mjwitte:
That warning indicates a `using` statement that is no longer needed.

rraustad:
I found several `using DataSizing:AutoSize` statements that weren't used. I don't understand why this warning pops up in only this branch.

Myoldmopar:
Build is happy on my Linux machine, which is where it appears the unused variable warnings have been coming from.  Windows is fully happy now on both 32 and 64 bit builds as well.  I'm OK pulling the trigger on this at any time.  Final thoughts @mjwitte ?

mjwitte:
No objections here.

mjwitte:
Except just realized there's some odd diffs [here](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2018-09/40434b6ebb5889085a6f575cf999018c209379e8-b2f13f2505699d5ff388c5416b7ebdb4bbc9b757/HVACStandAloneERV_Economizer/x86_64-MacOS-10.9-clang/eplusout.eio.diff.html). @rraustad Please check on this.

rraustad:
I see that and see what could be causing this in HVACStandAloneERV. The fans don't respond correctly during sizing to the Fan:SystemModel object. 

    if (StandAloneERV(StandAloneERVNum).SupplyAirFanIndex > 0) {

The Fan:SystemModel index is 0 based so the first fan doesn't get called correctly.

I can't explain what I see in the diff's since this example file has window AC and baseboards, but it doesn't appear to be related to UnitarySystem except that it's happening in this branch.

Issue #6792, opened 2018-06-20
Throw error for excess Controller:WaterCoil objects


mjwitte:
Issue overview
--------------
User file has (many) AirloopHVAC:UnitarySystem with water coils. Plus there is a Controller:WaterCoil for every coil, which shouldn't be used when the water coil is inside another system. This file results in many  FindRootSimpleController errors.

In order to make a simpler test file, tried to start with example file UnitarySystem_WaterCoils_wMultiSpeedFan which has one AirLoopHVAC:UnitarySystem with lwater coils, but discovered that this has a Controller:WaterCoil present for the CW coil - which it should not.  Made a version of this with the water coil deleted and the results change.

1. Need to fix UnitarySystem_WaterCoils_wMultiSpeedFan and remove the excess controller object (and controller list, and excess setpoint managers)
2. Need to throw a fatal error when a Controller:WaterCoil object is present for a water coil that is not directly on a branch or outdoor air system, or ??? This may be difficult to determine. Maybe the water coil init can figure this out.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v8.9.0
 - Private e-mail 2018-06-20

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #6804

PR #6804, opened 2018-07-01
Throw error for excess Controller:WaterCoil objects

Nigusse:
Addresses Issue #6792

Pull request overview
---------------------
This is defect. The fix now checks that if Coil:*:Water or CoilSystem:Cooling:Water:HeatExchangerAssisted object is neither on primary air loop branch nor outdoor air system and has a water controller specified, then throws error message and fatal out. Diffs are expected in example file UnitarySystem_WaterCoils_wMultiSpeedFan.idf because of removal of redundant Controller:CoilWater object. 

### Work Checklist


Nigusse:
There are eight example files that fatal out due to this fix. Current fix checks for water coils but not for water coils with heat exchanger assisted. The check is done in InitWaterCoil() function and the waterCoils objects deosn't know if the water coil is part of CoilSystem (water coils with heat exchanger assisted). I need a way to find out if the watercoil is part of CoilSystem object placed on primary airloop or OA system branch.

Nigusse:
@mjwitte @Myoldmopar I see audit diffs like this "+found (input file)=/Users/kbenne/ci_runs/EnergyPlus-13d69af0a4-x86_64-clang/datasets/Window5DataFile.dat). What does it mean?

Myoldmopar:
Completely unrelated, the diff scripts aren't ignoring the lines that show different paths to the dat file between builds.  I've tried ignoring it, but they still show up.  Anyway, nothing to worry about.

Nigusse:
Diffs in example file "UnitarySystem_WaterCoils_wMultiSpeedFan.idf" is expected since the redundant Controller:CoilWater object is removed. Water Coils in UnitarySystem does not need controller. This is also explained in the issue #6792 description.

Nigusse:
PackagedTerminalHP_VSCoils_Sizing unit test failure is not expected hence needs investigation.

Nigusse:
I see the unit test PackagedTerminalHP_VSCoils_Sizing is failing in CI machine but I could not duplicate the unit test failure in my Win64 machine. I have run the entire unit tests and this unit test in release mode and all passes. @Myoldmopar any clue? Otherwise, this issue is complete and ready for review.

nrel-bot-3:

Myoldmopar:
CI is much happier these days, so I pulled develop in once more and it should be all green and ready to go in.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -7391,6 +7391,249 @@ namespace SimAirServingZones {
...
 
+    void CheckWaterCoilIsOnAirLoop(int const CompTypeNum, std::string const CompType, std::string const CompName, bool &WaterCoilOnAirLoop)
+    {
+
+        // PURPOSE OF THIS FUNCTION:
+        // This function returns true if a water coil that has water controller is either on
+        // primary air or outdoor air system branch. Searches for water coil name and type
+        // that match components list in primary air and outside air systems.
+
+        // METHODOLOGY EMPLOYED:
```
Myoldmopar: I know this is what we used to do with these functions but we don't really need all these empty comment blocks.  If everything else is OK here I won't let that hold it up, but in the future clean them out.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -7391,6 +7391,249 @@ namespace SimAirServingZones {
...
+
+        CheckWaterCoilIsOnPrimaryAirLoopBranch = false;
+        if (DataHVACGlobals::NumPrimaryAirSys > 0) {
+            for (int AirSysNum = 1; AirSysNum <= DataHVACGlobals::NumPrimaryAirSys; ++AirSysNum) {
+                for (int BranchNum = 1; BranchNum <= PrimaryAirSystem(AirSysNum).NumBranches; ++BranchNum) {
+                    for (int CompNum = 1; CompNum <= PrimaryAirSystem(AirSysNum).Branch(BranchNum).TotalComponents; ++CompNum) {
+                        if ((CompTypeNum == PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).CompType_Num) &&
+                            UtilityRoutines::SameString(CompName, PrimaryAirSystem(AirSysNum).Branch(BranchNum).Comp(CompNum).Name)) {
+                            CheckWaterCoilIsOnPrimaryAirLoopBranch = true;
+                            goto PrimaryAirSystemLoop_exit;
```
Myoldmopar: Annnd nope.  Just put a return statement in here.  Get rid of the local variable and **no goto**.

```
if (Sys > 0) {
  for each air system {
    for each branch {
      for each comp {
        if comp is type and same string name {
          return true;
        }
      }
    }
  }
}
return false;
```

src/EnergyPlus/WaterCoils.hh
```diff
@@ -117,6 +117,7 @@ namespace WaterCoils {
     extern Array1D_int PartWetCoolCoilErrs;   // error counting for detailed coils
     extern bool GetWaterCoilsInputFlag;       // Flag set to make sure you get input once
     extern Array1D_bool CheckEquipName;
+    // extern bool WaterCoilControllerCheckOneTimeFlag; // flg used to check water coil controller
```
Myoldmopar: This should just be removed if it's not used anymore unless there is a good reason to keep it.

Myoldmopar (changes requested):
Great unit test, and the fix seems OK.  However, the code needs some (very quick) changes.  The GoTo logic needs to be removed.  The change is really just to simplify the functions that use GoTo, and instead delete the local variable and just use a return statement.  Will be less lines of code, and more importantly, no GoTo.  While you are in there please go ahead and remove some of the bulky comment blocks at the tops of the functions.  Keep a description, you can keep the author if you want, but not all the empty sections.

src/EnergyPlus/SimAirServingZones.cc
Nigusse: Removed empty blocks as suggested.

src/EnergyPlus/SimAirServingZones.cc
Nigusse: Implemented as suggested.

src/EnergyPlus/WaterCoils.hh
Nigusse: Forgot to clean up. Removed it.

Nigusse:
@Myoldmopar Addressed your review comments, tested the unit tests locally and all passed. This is ready for merging.

Nigusse:
@Myoldmopar Diffs in example file UnitarySystem_WaterCoils_wMultiSpeedFan is expected due to redundant Controller:WaterCoil object, which is removed from this branch.

Myoldmopar (approved):

Issue #6598, opened 2018-03-27
Fan:SystemModel crashes using Continuous speed control method.


rraustad:
Issue overview
--------------
A result of HelpDesk ticket 12594. Changed Fan:SystemModel speed control to Continuous just to test if constant fan operating mode would work correctly for UnitarySystem. Crash where m_flowFractionAtSpeed is not yet allocated.

        if (m_numSpeeds > 1) { // set up values at speeds
            m_massFlowAtSpeed.resize(m_numSpeeds, 0.0);
            m_totEfficAtSpeed.resize(m_numSpeeds, 0.0);
            for (auto loop = 0; loop < m_numSpeeds; ++loop) {
                m_massFlowAtSpeed[loop] = m_maxAirMassFlowRate * m_flowFractionAtSpeed[loop];

Chicago weather but shouldn't matter.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (EnergyPlusDevSupport defect files folder)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Nigusse:
Any one working on this bug issue?

rraustad:
I already have a fix for this one.

rraustad:
See PR #6805.

Myoldmopar:
Closed via #6805

PR #6805, opened 2018-07-02
Correct Fan:SystemModel crash when Continuous mode is used without power as a function of flow fraction curve

rraustad:
Pull request overview
---------------------
The Fan:SystemModel has two operating modes, Discrete and Continuous. The Continuous operating mode requires that a power as a function of flow fraction curve is provided so that power can be calculated. If this curve is not provided the simulation will crash. This corrects #6598.

Work around: Provide the required performance curve.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/HVACFan.cc
+++ b/src/EnergyPlus/HVACFan.cc
@@ -278,7 +278,7 @@ namespace HVACFan {
         // calculate total fan system efficiency at design
         m_fanTotalEff = designAirVolFlowRate * deltaPress / designElecPower;
 
-        if (m_numSpeeds > 1) { // set up values at speeds
+        if (speedControl == SpeedControlMethod::Discrete && m_numSpeeds > 1) { // set up values at speeds
             m_massFlowAtSpeed.resize(m_numSpeeds, 0.0);
             m_totEfficAtSpeed.resize(m_numSpeeds, 0.0);
             for (auto loop = 0; loop < m_numSpeeds; ++loop) {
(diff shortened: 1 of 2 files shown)


rraustad:
Tested all cases where Power as a function of Flow Fraction could be used.
- Discrete, no curve
- Discrete w/curve
- Continuous w/curve
- added fatal when no curve is used with Continuous

Performance curve in question is:

    Fan:SystemModel,
      CombinedPowerAndFanEff,  !- Electric Power Function of Flow Fraction Curve Name

6598-fanpowercurveresults

The performance curve is not used for Discrete operating mode so it doesn't matter if it's present or not. Fan power for both discrete test cases are the same.

6598-fanpowercurveresults_samefanpower

Myoldmopar:
So these changes are fine, but I might be missing something so I just want to check.  I took the UnitarySystem_MultiSpeedCoils_SingleMode.idf file and removed the Electric Power curve value.  In **develop**, the file ran until a fatal error saying the field is required.  With this branch I get the same error.  I'm just wanting to recreate the error.  I could use the defect file if there is one, but I also want to understand it just a bit.

rraustad:
I had a file that showed an issue with fan operation and I fixed that. Then I wanted to make sure it worked when I switched the fan control type. When I switched the fan control type from discrete to continuous, the program crashed (defect file = 6598-Office_Chicago_MultiStageGasHeatingCoil_Continuous) ... because there was no power function of flow curve. It's because the code checked for discrete in Init but didn't check for discrete in calc. This fixes that.

Myoldmopar:
OK, I took the defect file and modified our example file to be similar for the Fan:System model object.  I was able to see the hard crash.  I did the same thing on this branch and saw the fatal error.  This is good, CI is happy, I'm re-running the unit tests locally to make sure nothing is introduced when this merges with develop, but I anticipate it will be fine and I'll merge this in a few minutes.  Thanks @rraustad.

Myoldmopar:
Tests pass locally, good to go.

PR #6836, opened 2018-07-13
Prep for 9.0

Myoldmopar:
Pull request overview
---------------------
Addressed the usual things for releases:
 - Change version in test files, unit tests, and datasets
 - Change version in a few spots in CMake
 - Change the SHA from the previous release in CMake
 - Add the configured IDD from the prior release, including install rules
 - Add a new transition Fortran file with CMake build rules
 - Add a new blank report variables file with CMake build rules updated
 - Add Input/Output rules MD files
 - Update install.cmake with updated version numbers

Then, I'll watch for error file diffs that are clues that I forgot to update an example file, along with any other suspicious stuff from CI.  (Table diffs and EIO diffs are expected.)


Myoldmopar:
OK, expected to see EIO diffs and table diffs, because they will warn about the changed version.  Audit diffs are the usual CI noise.  The ERR files are my clue that something was still wrong, because if I update the version in CMake, but I forget to update the version in an example file, it will show up as a new error in the error file.  Which it did here for two files.  For the files that are updated properly, there will be no error.  I have updated the two failing files and pushed it and I'll go ahead and merge this in so that we can start writing transition rules off of this stuff.

Issue #6243, opened 2017-08-15
Suspect zone hvac heating coil sizing with DOAS mixer


mjwitte:
Issue overview
--------------
While testing changes for #6223, I tried changing the OA spec to be 10% of the original and then compared the eio outputs expecting the DOAS components to all be 10% of the original  That much works as expected.  I also expected no change in the zone HVAC system sizing.

1. For the fan coil test file, the fan coil components were all the same size in zones 1-4, as expected.  But in zone 5, the fan coil heating coil got smaller.

2. For the PTHP test file, all five of the supp heating coils got smaller.

In both cases, I made sure that the Sizing:Zone "account for DOAS" was set to no.  

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus v8.7.0 https://github.com/NREL/EnergyPlus/commit/fdb3cf27a3f041101f9cbbceab78a7bcfc0c89e3

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
I'm not sure i agree that the coil sizes should not change when an ATMixer is connected to zone equipment and the OA flow rate through the DOAS, and each unit, changes. What is hard to comprehend is how the coil sizes should change. The only equipment in the DOAS is a HX so I would expect the cooling AND heating coil size to decrease, which is what I see. I reviewed the discussion on [6838](https://github.com/NatLabRockies/EnergyPlus/pull/6838) which was closed without any changes, and think this is likely not an issue. I did run both PTHP files and see the same zone air temperature, which means the coil sizes are adequate to maintain zone conditions. For that reason I am closing this issue. The old defect files are added here and did not transition very will so use these in the future.

<img width="751" height="459" alt="Image" src="https://github.com/user-attachments/assets/5be64e11-2760-4a44-abe0-ec7698954d2d" />

[6243-DOAToPTHP_new-10PercentOA.zip](https://github.com/user-attachments/files/30605329/6243-DOAToPTHP_new-10PercentOA.zip)

PR #6838, opened 2018-07-13
Addressed zone hvac heating coil sizing with DOAS mixer

Nigusse:
Addressed Issue #6243 

This is a defect in packaged terminal unit (PTHP, PTAC, and PWSHP). PTUnits coils sizing always uses mixed air temperature regardless of the OA amount specified. When there is no OA air, the coil entering temperature for sizing should have been zone return air temperature. Diffs are expected in PTUnts used with DOAS mixer.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)


Nigusse:
@mjwitte I found two issue during investigation that could lead to difference when changing the OA amount in design specification object:
(1) Change in OA amount can result in different design air flow rate 
(2) Incorrect use of coil entering air temperature for supplemental heating coil sizing when there is no OA amount specified in the parent object

The first issue is by design hence the diffs is expected if the situation allows. The diffs observed in FanCoil unit 5 is expected. The minimum OA flow rate is higher than the sizing heating flow rate, hence the design heating air flow rate was set to OA min flow rate. But when the OA Design specification value is reduced by a factor of 10, then tthe sizing flow rate became higher than the OA min flow rate, hence design heating flow rate was set to sizing flow rate. Thus, test case FanCoil Unit 5 and all the five PTHP test cases were impacted when the design OA specification amount reduced by a factor of 10. 

The second issue impacts Packaged Terminal Units (PTUnit) only. PTUnits use mixed air temperature for supplemental  heating coil regardless of the OA amount specified in OA Design Specification object. FanCoil units uses return air temperature for coils sizing if OA amount specified in parent object is zero (for example when OA to the FanCoil is provided by DOAS). This is not the case for PTUnit supplemental heating coils sizing hence addressed in this fix. As the result diffs were observed in 9 example files: DOAToPTAC.idf, DOAToPTHP.idf, DOAToWaterToAirHPInlet.idf
DOAToWaterToAirHPSupply.idf, HVACTemplate-5ZonePTAC-DOAS.idf, HospitalLowEnergy.idf, ZoneSysAvailManager.idf, ZoneVSWSHP_wDOAS.idf, and ZoneWSHP_wDOAS.idf. These diffs are caused by decrease in supplemental heating coil Design Size Nominal Capacity.

Nigusse:
Due to this fix nine example files showed reduced supplemental heating coils nominal capacity. Sample diffs extracted from DOAToPTAC.idf and HospitalLowEnergy.idf example files are shown below. These diffs are caused by the use of return air temperature instead of mixed air temperature for supplemental heating coils sizing in PTUnits when there is no OA specified in the PTUnits (e.g. when OA is provided by DOAS).


6243_doatoptac_eio_diffs

6243_hospitallowenergy_eio_diffs

mjwitte:
@Nigusse @rraustad If I'm following correctly, this fixes the case where this is no OA from any source, but when there is an ATMixer present, if the mixer is an inlet side mixer, then the coil sizing should be using a mixed air temperature based on the ATMixer airloop's design heating/cooing supply temp.

Nigusse:
@mjwitte I addressed one of the issues that I observed and the second issue that I was thinking to resolve was what happens if there is OA from TAMixer. Looking at the code I see (I think it is Fred's work) that was added [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L651-L700) that accounts for DOAS load to the ZoneHVAC:* equipment. But the estimated DOAS load never propagated to the zone equipment sizing calculation. Now @rraustad also added a separate fix for this problem in PR #6838. 

I will update my branch when this PR is merged and see if there is any problem not addressed by Rich's fix.

nrel-bot-3:

nrel-bot-3:

nrel-bot-3:

mjwitte:
@Nigusse Now that #6788 is merged, "Include impact of AirTerminal:SingleDuct:Mixer in ZoneHVAC coil sizing", please verify if this is still needed.  If so, please merge in develop and resolve conflicts.

Myoldmopar:
@Nigusse please confirm that you saw [this](https://github.com/NREL/EnergyPlus/pull/6838#issuecomment-423228298).  I know @mjwitte only commented a couple hours ago, but given the time crunch I just need to make sure you saw his message in the flurry of activity today.  Thanks!

rraustad:
I just checked and he is out on leave until Monday 9/24.

mjwitte:
@rraustad Would *you* be able to check to see if this is moot now after the large ATMixer sizing work?

rraustad:
I can try.

rraustad:
Interesting enough, using current develop, the heating coil sizes in DOAToPTAC shown above in chart from July 15 show up now as the same as Value Before Fix for Space 1-1, 2-1 and 3-1. And same as Value After Fix for Space 4-1 and 5-1. So something is missing in develop to date (i.e., my fix earlier).

    Component Sizing Information, Coil:Heating:Fuel, SPACE1-1 HEATING COIL, Design Size Nominal Capacity [W], 2899.74027
    Component Sizing Information, Coil:Heating:Fuel, SPACE2-1 HEATING COIL, Design Size Nominal Capacity [W], 1282.54041
    Component Sizing Information, Coil:Heating:Fuel, SPACE3-1 HEATING COIL, Design Size Nominal Capacity [W], 2859.01586
    Component Sizing Information, Coil:Heating:Fuel, SPACE4-1 HEATING COIL, Design Size Nominal Capacity [W], 838.46468
    Component Sizing Information, Coil:Heating:Fuel, SPACE5-1 HEATING COIL, Design Size Nominal Capacity [W], 3468.59894

I'll pull this branch and merge.

mjwitte:
@rraustad That may be correct if some are inletside and some are supplyside - and make sure your using a consistent idf with DOAS sizing as appropriate.

rraustad:
OK, in the DOAToPTAC file Space1-1 to 3-1 are inlet side mixers and Space 4-1 and 5-1 are supply side mixers. There are no coils in the DOAS air loop. So for inlet side mixers, the heating coil doesn't change size even though the cold air is mixed with PTAC return air. For supply side mixers, why would the heating coil size decrease if cold air is mixed at the PTAC outlet? Wouldn't that mean a larger heating coil is needed? I adjusted the inlet conditions to the coil so I would have thought heating coils for Space 1-1 to 3-1 would have increased, and from the chart above they didn't change size.


After I merged develop the coils sizes line up with what @Nigusse had before, right or wrong.

    Component Sizing Information, Coil:Heating:Fuel, SPACE1-1 HEATING COIL, Design Size Nominal Capacity [W], 1895.71398
    Component Sizing Information, Coil:Heating:Fuel, SPACE2-1 HEATING COIL, Design Size Nominal Capacity [W], 838.46468
    Component Sizing Information, Coil:Heating:Fuel, SPACE3-1 HEATING COIL, Design Size Nominal Capacity [W], 1869.09027
    Component Sizing Information, Coil:Heating:Fuel, SPACE4-1 HEATING COIL, Design Size Nominal Capacity [W], 838.46468
    Component Sizing Information, Coil:Heating:Fuel, SPACE5-1 HEATING COIL, Design Size Nominal Capacity [W], 3468.59894



    AirTerminal:SingleDuct:Mixer,
    SPACE1-1 DOAS Air Terminal,  !- Name
    ZoneHVAC:PackagedTerminalAirConditioner,  !- ZoneHVAC Unit Object Type
    SPACE1-1 PTAC,           !- ZoneHVAC Unit Object Name
    SPACE1-1 PTAC Inlet,     !- Mixer Outlet Node Name
    SPACE1-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
    SPACE1-1 Air Terminal Mixer Secondary Inlet,  !- Mixer Secondary Air Inlet Node Name
    InletSide,               !- Mixer Connection Type
    CurrentOccupancy;        !- Per Person Ventilation Rate Mode

    AirTerminal:SingleDuct:Mixer,
    SPACE2-1 DOAS Air Terminal,  !- Name
    ZoneHVAC:PackagedTerminalAirConditioner,  !- ZoneHVAC Unit Object Type
    SPACE2-1 PTAC,           !- ZoneHVAC Unit Object Name
    SPACE2-1 PTAC Inlet,     !- Mixer Outlet Node Name
    SPACE2-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
    SPACE2-1 Air Terminal Mixer Secondary Inlet,  !- Mixer Secondary Air Inlet Node Name
    InletSide,               !- Mixer Connection Type
    SZ DSOA SPACE2-1,        !- Design Specification Outdoor Air Object Name
    CurrentOccupancy;        !- Per Person Ventilation Rate Mode

    AirTerminal:SingleDuct:Mixer,
    SPACE3-1 DOAS Air Terminal,  !- Name
    ZoneHVAC:PackagedTerminalAirConditioner,  !- ZoneHVAC Unit Object Type
    SPACE3-1 PTAC,           !- ZoneHVAC Unit Object Name
    SPACE3-1 PTAC Inlet,     !- Mixer Outlet Node Name
    SPACE3-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
    SPACE3-1 Air Terminal Mixer Secondary Inlet,  !- Mixer Secondary Air Inlet Node Name
    InletSide,               !- Mixer Connection Type
    SZ DSOA SPACE3-1,        !- Design Specification Outdoor Air Object Name
    CurrentOccupancy;        !- Per Person Ventilation Rate Mode

    AirTerminal:SingleDuct:Mixer,
    SPACE4-1 DOAS Air Terminal,  !- Name
    ZoneHVAC:PackagedTerminalAirConditioner,  !- ZoneHVAC Unit Object Type
    SPACE4-1 PTAC,           !- ZoneHVAC Unit Object Name
    SPACE4-1 Supply Inlet,   !- Mixer Outlet Node Name
    SPACE4-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
    SPACE4-1 PTAC Outlet,    !- Mixer Secondary Air Inlet Node Name
    SupplySide,              !- Mixer Connection Type
    SZ DSOA SPACE4-1,        !- Design Specification Outdoor Air Object Name
    CurrentOccupancy;        !- Per Person Ventilation Rate Mode

    AirTerminal:SingleDuct:Mixer,
    SPACE5-1 DOAS Air Terminal,  !- Name
    ZoneHVAC:PackagedTerminalAirConditioner,  !- ZoneHVAC Unit Object Type
    SPACE5-1 PTAC,           !- ZoneHVAC Unit Object Name
    SPACE5-1 Supply Inlet,   !- Mixer Outlet Node Name
    SPACE5-1 Air Terminal Mixer Primary Inlet,  !- Mixer Primary Air Inlet Node Name
    SPACE5-1 PTAC Outlet,    !- Mixer Secondary Air Inlet Node Name
    SupplySide,              !- Mixer Connection Type
    SZ DSOA SPACE5-1,        !- Design Specification Outdoor Air Object Name
    CurrentOccupancy;        !- Per Person Ventilation Rate Mode

mjwitte:
Remember we went round about on PTAC in your other PR.

rraustad:
I don't think these results make sense. And I hate to do this to you guys, but my Aunt is turning 100 this Sunday and my brother and I fly out tomorrow to visit.

mjwitte:
Which results don't make sense?  Develop or this branch?  And Happy Birthday to Auntie!

Myoldmopar:
OK, we're going to have to punt on this one.  We still have issues that cause crashes that haven't been wrapped up, and there's not going to be anyone to work on this before final freeze.  Someone could prove me wrong of course, but for now I'm changing the milestone and moving on.

rraustad:
Well, if I did the math right, both branches are calculating the correct coil capacity. I ran develop and this branch and pulled in the zone load (same in both) and zone air flow (same in both) and coil ent/lvg temps from html. Then calculated MdotCpDt (math right?, VrhoCpDt = V * 1000 * 1.2 * Dt?) and compared both. The inlet mixer should cause lower inlet temps, which in develop it does (much lower), this branch it doesn't. And supply mixer should also require a larger coil than the zone load, which it does in both branches. Honestly, I can't tell which is right without more investigation. What I don't understand is why the inlet mixer causes a much larger increase in heat coil capacity when the delivered load by the mixer is the same regardless of inlet or supply mixer configuration. So if feels like something is missing. For example, calculating OA frac based on mixer flow vs PTAC flow or mixer flow vs zone heating air flow gives a different fraction and therefore different inlet temp during sizing. So is that playing a role here? Unless I actually calculate OA load and add to zone load will I know which is actually correct. They both show increased heating capacity over the zone load. My gut says that the inlet mixer coil inlet temp is too low in develop. I remember thinking about that. The flow rate of the PTAC in heating mode won't be the zone heating air flow rate, it'll be higher I believe and therefore coil inlet temp won't be so low.

    Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Cooling Supply Air Flow Rate [m3/s], 0.26879
    Component Sizing Information, ZoneHVAC:PackagedTerminalAirConditioner, SPACE1-1 PTAC, Design Size Heating Supply Air Flow Rate [m3/s], 0.26879

compareheatcoilsizing

rraustad:
Trying to resolve this I calculate total DOAS load and compare to total zone load (assuming OA T = 8.7 and zone T = 23). Develop is a little higher than total load and this branch is a little lower.

compareheatcoilsizing

Nigusse:
@mjwitte @Myoldmopar @rraustad Reviewed the code changes and sizing results diffs in the eio file between this branch and develop.  This branch was meant to address PTUnit sizing when the OA flow in the  PTUnit is zero and the PTUnit is not part of the AirTerminal Mixer. But this sizing scenario is already accommodated in @rraustad fix [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ReportSizingManager.cc#L1768-L1778)  and [here](https://github.com/NREL/EnergyPlus/blob/159013173_Issue%236243/src/EnergyPlus/ReportSizingManager.cc#L3809-L3815). The fix in this branch is not necessary.

Issue #6506, opened 2018-02-28
Outdoor air node wet-bulb EMS override only work with with OutdoorAir:Node not OutdoorAir:NodeList and other problems


mjwitte:
Issue overview
--------------
1. In `OutAirNodeManager::GetOutAirNodeInputs` the [input processing](https://github.com/NREL/EnergyPlus/blob/f84f30a62482f0b7881f33516b972faf077c3cfe/src/EnergyPlus/OutAirNodeManager.cc#L329-L332) for `OutdoorAir:Node` sets `AnyLocalEnvironmentsInModel = true` if there is more than one alpha field.  This is wrong - it will cause a performance hit elsewhere in the code if local environments are not really being used. This should test if `AnyLocalEnvironmentsInModel == true` in deciding whether to set the `IsLocalNode` flag.

2. In `OutAirNodeManager::InitOutDoorAirNodes` the [logic for setting humrat](https://github.com/NREL/EnergyPlus/blob/f84f30a62482f0b7881f33516b972faf077c3cfe/src/EnergyPlus/OutAirNodeManager.cc#L461-L467) needs to be much more general. 
    a. It should not rely on `IsLocalNode` because EMS or schedules could be used to override OA node data even without using the local environments model.
    b. It needs a flag that gets set if any of the overrides for EMS dry bulb, EMS wet bulb, drybulb schedule, or wet bulb schedule are active and then recalculate humrat.
    c. But what if only dry bulb is overridden - then humrat should use the OA condition and wetbulb should be recalculated?
    d. And shouldn't the OA condition be height-based if applicable?
    e. Other cases I haven't thought of here?

3. There is a similar block of code in [CheckAndAddAirNodeNumber](https://github.com/NREL/EnergyPlus/blob/f84f30a62482f0b7881f33516b972faf077c3cfe/src/EnergyPlus/OutAirNodeManager.cc#L603-L626). These should be consolidated into a single function that sets a single node.

### Workaround
Use `OutdoorAir:Node` with the last field non-blank for any OA node that will have an EMS override for wetbulb temperature.
```
OutdoorAir:Node,
    Node 12,                 !- Name
    ,                        !- Height Above Ground {m}
    ,                        !- Drybulb Temperature Schedule Name
    ,                        !- Wetbulb Schedule Name
    ,                        !- Wind Speed Schedule Name
    ,                        !- Wind Direction Schedule Name
    ,                        !- Wind Pressure Coefficient Curve Name
[...]

PR #6841, opened 2018-07-16
Allow OA node EMS override for both OutdoorAir:Node and OutdoorAir:NodeList

lgu1234:
Pull request overview
---------------------
Addresses #6506.

Upload an initial fix to allow outdoor air node wet-bulb EMS override for both OutdoorAir:Node and OutdoorAir:NodeList objects. The fix addressed Item 1. 

A section of code was added to check if actuator types are "Outdoor Air System Node" or not. If yes, set the Node.IsLocalNode = true. An additional test file was uploaded to have a full inputs of OutdoorAir:Node. Simulation results are the same from 3 test files.

Differences were found as expected for an example file of RetailPackagedTESCoil.idf.  

I added more actions in addition to IsLocalNode. When outdoor drybulb is overridden only, the outdoor humrat is used to calculate outdoor wetbulb. This causes differences in an example file of AirflowNetwork_MultiZone_LocalNode.idf. Some other issues in Item 2 were done earlier.
   
A new function was created as SetOANodeValues to combine a section of code in both InitOutAirNodes and CheckAndAddAirNodeNumber.

A bug was found and fixed in the CheckAndAddAirNodeNumber function:

Node(NodeNumber).HumRat = PsyHFnTdbW(Node(NodeNumber).OutAirDryBulb, Node(NodeNumber).OutAirWetBulb);

PsyWFnTdbTwbPb should be used instead of PsyHFnTdbW.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1294,6 +1294,7 @@ namespace ReportSizingManager {
                         CoilInHumRat = FinalZoneSizing(CurZoneEqNum).DesCoolCoilInHumRat;
                         DesCoilLoad = FinalZoneSizing(CurZoneEqNum).DesCoolMassFlow *
                                       (PsyHFnTdbW(CoilInTemp, CoilInHumRat) - PsyHFnTdbW(CoilOutTemp, CoilOutHumRat));
+                        CoilDesWaterDeltaT = DataWaterCoilSizCoolDeltaT; // Temporary fix by Gu on 7/19/18
```
Myoldmopar: Temporary fix by Gu?

Myoldmopar (changes requested):
I will try to read through the PR description a bit more to understand all the nuances here.  From a pure code review standpoint, we need to address the "Temporary fix" line.

src/EnergyPlus/ReportSizingManager.cc
mjwitte: @Myoldmopar @lgu1234 This line can be removed now, then merge in develop.  It was fixed in #6850.

lgu1234: @mjwitte @Myoldmopar I am going to remove it.

Myoldmopar (approved):

nrel-bot-3:

src/EnergyPlus/EMSManager.cc
```diff
@@ -1537,6 +1537,7 @@ namespace EMSManager {
         } // NumOfNodes > 0
 
         if (NumOutsideAirNodes > 0) {
+            DataGlobals::AnyLocalEnvironmentsInModel = true;
```
Myoldmopar: @lgu1234 is this right?  Anytime you have outside air nodes you need to set this flag to true?  I thought this flag was only true if the simulation had any outside air nodes with specific heights, so that you needed to update the local environment nodes appropriately.  I feel like setting this true here could cause the simulation to attempt to update local environment nodes when it doesn't need to.  It'll probably be a minuscule performance effect, but more importantly it just doesn't seem right.

Could you clarify why you needed it here?

mjwitte: AnyLocalEnvironmentsInModel is related to SurfaceProperty:LocalEnvironment and ZoneProperty:LocalEnvironment.

mjwitte: I think it could be significant performance hit for places that look for surface properties to apply.  I don't remember for sure, but should search the code on  `AnyLocalEnvironmentsInModel` and see how it's used.

Myoldmopar: Interesting.

Over in the [OA Node Manager](https://github.com/NREL/EnergyPlus/blob/eb68ccb2bfb3d01807412e33399299da8cdb4b11/src/EnergyPlus/OutAirNodeManager.cc), during the [GetInput for OutdoorAir:Node](https://github.com/NREL/EnergyPlus/blob/eb68ccb2bfb3d01807412e33399299da8cdb4b11/src/EnergyPlus/OutAirNodeManager.cc#L305), the flag is [set to true](https://github.com/NREL/EnergyPlus/blob/eb68ccb2bfb3d01807412e33399299da8cdb4b11/src/EnergyPlus/OutAirNodeManager.cc#L364-L367) for any node who has number of alphas greater than 1, which would imply that the field has scheduled conditions.  So the trigger there is not the height, but instead scheduled.  This makes more sense given the EMS connection here.

Outdoor air node actuators are always set up, and so with this change, every model that has outdoor air nodes will have a true value on this flag.  Or I'm still missing something.

@lgu1234 I'm going to push this toward the back burner for tonight, chime in in the morning and let's see if we can get this resolved.

lgu1234: @Myoldmopar Your understanding is correct. The purpose to set AnyLocalEnvironmentsInModel=true is to allow all nodes defined under OutdoorAir:NodeList to be overridden by EMS.

lgu1234: @Myoldmopar In addition, the AnyLocalEnvironmentsInModel is set to true in GetOutAIrNode Input by reading the "OutdoorAir:Node" object in OutAirNodeManager at line 365

                if (NumAlphas > 1) {
                    AnyLocalEnvironmentsInModel = true;
                }
My intention is to allow OutdoorAir:NodeList to have the same setting as OutdoorAir:Node.

mjwitte: Since you have the check for `NumAlphas>1`, then why set it when if `NumOutsideAirNodes > 0`? It seems like we're using `AnyLocalEnvironmentsInModel` for dual purposes here, for local environment *and* for EMS.

lgu1234: @mjwitte The check for NumAlphas>1 is the original code. The change I made is to mimic OutdorAir:Node for OutdoorAir:NodeList. I agree with you the AnyLocalEnvironmentsInModel is for dual purpose. We may need a new issue to cleanup later.

mjwitte: If I'm following all of this, the only thing that the outdoor air node manager cares about is `IsLocalNode`. This can be true if there's a direct EMS actuator on the node or if one of the extended fields are used, correct?  Beyond that, I don't see why the outdoor air node manager or EMSmanager should set AnyLocalEnvironments at all. Only the LocalEnvironment objects should set that.

mjwitte: @lgu1234 @Myoldmopar Just to emphasize why this is important here, we don't want [this loop](https://github.com/NREL/EnergyPlus/blob/9e6c9e481451903c5860edfd7a3446a74f5499ce/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L447-L494) and others like it to fire for any run that has an outdoor air node (essentially every simulation).

mjwitte: Sorry for the piecemeal comments. So, this line in EMSManager should be deleted.

src/EnergyPlus/EMSManager.cc
```diff
@@ -1563,6 +1564,12 @@ namespace EMSManager {
                                  "[degree]",
                                  Node(NodeNum).EMSOverrideOutAirWindDir,
                                  Node(NodeNum).EMSValueForOutAirWindDir);
+                for (int ActuatorUsedLoop = 1; ActuatorUsedLoop <= numActuatorsUsed; ActuatorUsedLoop++) {
+                    if (UtilityRoutines::SameString(EMSActuatorUsed(ActuatorUsedLoop).ComponentTypeName, "Outdoor Air System Node") && UtilityRoutines::SameString(EMSActuatorUsed(ActuatorUsedLoop).UniqueIDName,NodeID(NodeNum))) {
+                        Node(NodeNum).IsLocalNode = true;
```
mjwitte, line 1568: This is good here - important to set IsLocalNode.

src/EnergyPlus/OutAirNodeManager.cc
```diff
@@ -363,7 +363,6 @@ namespace OutAirNodeManager {
 
                 if (NumAlphas > 1) {
                     AnyLocalEnvironmentsInModel = true;
-                    Node(NodeNums(1)).IsLocalNode = true;
```
mjwitte, line 366: Delete both of these lines here (IsLocalNode already deleted), but delete AnyLocalEnvironments line.

src/EnergyPlus/OutAirNodeManager.cc
```diff
@@ -408,6 +407,9 @@ namespace OutAirNodeManager {
                     ErrorsFound = true;
                     continue;
                 }
+                if (Node(NodeNums(1)).OutAirDryBulbSchedNum > 0 || Node(NodeNums(1)).OutAirWetBulbSchedNum > 0) {
+                    Node(NodeNums(1)).IsLocalNode = true;
+                }
```
mjwitte, line 412: This is good here.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -1526,3 +1526,45 @@ TEST_F(EnergyPlusFixture, EMSManager_TestFuntionCall)
...
+    EMSActuatorUsed(1).ComponentTypeName = "Outdoor Air System Node";
+    EMSActuatorUsed(2).ComponentTypeName = "Outdoor Air System Node";
+    EMSActuatorUsed(3).ComponentTypeName = "Outdoor Air System Node";
+    EMSActuatorUsed(1).UniqueIDName = NodeID(1);
+    EMSActuatorUsed(2).UniqueIDName = NodeID(2);
+    EMSActuatorUsed(3).UniqueIDName = NodeID(3);
+
+    SetupNodeSetPointsAsActuators();
+
+    EXPECT_TRUE(DataGlobals::AnyLocalEnvironmentsInModel);
```
mjwitte: And delete this one.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -1526,3 +1526,45 @@ TEST_F(EnergyPlusFixture, EMSManager_TestFuntionCall)
...
+    EXPECT_TRUE(Node(1).IsLocalNode);
+    EXPECT_TRUE(Node(2).IsLocalNode);
+    EXPECT_TRUE(Node(3).IsLocalNode);
```
mjwitte: To make this unit test more interesting, I would delete one of the EMSActuators above and expect one of these to be FALSE.

mjwitte (commented):
Sorry for all of the piecemeal comments.

Myoldmopar:
@mjwitte thank you for being persistent on this. I think there's a clear path here now.  Maybe.  If I understand it. :rofl: 

@lgu1234 please delete the assignments of `AnyLocalEnvironmentsInModel` as they appear to be unnecessary and will cause a performance penalty.  Adjust the unit tests according to @mjwitte's suggestions.  Merge in develop.  And if everything is happy, we'll be good to merge.

src/EnergyPlus/EMSManager.cc
lgu1234: @Myoldmopar @mjwitte I am going to delete this line

src/EnergyPlus/OutAirNodeManager.cc
lgu1234: @mjwitte This is the original code. Let me upload the fix in EMSManager first. If everything is OK, I will delete this line.

tst/EnergyPlus/unit/EMSManager.unit.cc
lgu1234: @mjwitte Done.

tst/EnergyPlus/unit/EMSManager.unit.cc
lgu1234: @mjwitte Done. I remove Node 2. It is false.

Myoldmopar (approved):

Myoldmopar:
Looks like the review comments have been addressed.  CI is starting off clean with the latest commit, still 2 diff files, but they are the same as the last set of CI results.  Continued CI results will likely start showing some unrelated diffs due to merges to develop.  I'm inclined to go ahead and merge soon.  I'll pull develop into this branch locally and run the full unit test suite first, but if that passes I'll merge.  That OK with you @mjwitte ?

src/EnergyPlus/OutAirNodeManager.cc
mjwitte: @Myoldmopar Remember I'd like this line deleted also.

Myoldmopar: It looked like it was...sorry...I'll delete that line locally and run the test suite.  If it passes I'll push up my updated branch and then merge.

Myoldmopar:
OK, I deleted that line and built locally with develop merged in.  The build was fine, the unit test suite passes.  I ran SurfaceZonePropTest_LocalEnv.idf with develop and this branch and get no diffs in the eso output.  I think this is ready to go.

lgu1234:
@Myoldmopar Thanks.

Issue #6843, opened 2018-07-17
OutdoorAirUnit crashes when used with steam heating coil


Nigusse:
Found two issues when using ZoneHVAC:OutdoorAirUnit with steam heating coil: 

- [ ] Parent object steam flow rate autosizing problem. Steam flow rate  Steam mass flow rate  is calculated using liquid water density in the OutdoorAirUnit parent. Steam heating coils object volume flow rates are sized using saturated vapor. Conversion of vapor volume flow rate to mass flow rate should use vapor density instead of liquid density.  InitOutdoorAirUnit is using incorrect density [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/OutdoorAirUnit.cc#L1281-L1288).  
- [ ] ZoneHVAC:OutdoorAirUnit crashes when used with steam heating coil due to missing coil outlet node 
initialization.
- [ ] Also found coil cooling water flow rate sizing issue due to uninitialized variable CoilDesWaterDeltaT when the water cooling coil is part of ZoneHVAC equipment (e.g. ZoneHVAC:OutdoorAirUnit)

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Nigusse:
When auto-sizing water cooling coil flow rate and the coil is part of ZoneHVAC:* equipment such as ZoneHVAC:OutdoorAirUnit, the variable CoilDesWaterDeltaT is undefined. As the results the water coil flow rate cannot be autosized.

Nigusse:
Further testing showed that ZoneHVAC:OutdoorAirUnit crashes when steam heating coil is used instead of hot water heating coil due to missing to init steam coil outlet node.

Myoldmopar:
No defect file?

Myoldmopar:
Closed via #6848

PR #6848, opened 2018-07-20
Addresses OutdoorAirUnit crash when used with steam heating coil

Nigusse:
Issue #6843

This is a defect. OutdoorAirUnit does not work with steam heating coils and the steam flow rate autosizing did not work either. This fix resolves three issues: steam coil outlet node was not initialized, steam flow rate autosizing problem, and cooling coils had WaterCoilDeltaT initialization missing for coil water flow sizing calculation. The third one affects any ZoneHVAc equipment that has water cooling coil such as FANCOILs. 

### Work Checklist

### DIFF
--- a/src/EnergyPlus/OutdoorAirUnit.cc
+++ b/src/EnergyPlus/OutdoorAirUnit.cc
@@ -1101,6 +1101,7 @@ namespace OutdoorAirUnit {
         using DataSizing::AutoSize;
         using HVACHXAssistedCoolingCoil::SimHXAssistedCoolingCoil;
         using WaterCoils::SimulateWaterCoilComponents;
+        auto &GetSteamCoilMaxFlowRate(SteamCoils::GetCoilMaxWaterFlowRate);
 
         // Locals
         // SUBROUTINE ARGUMENT DEFINITIONS:
@@ -1280,10 +1281,11 @@ namespace OutdoorAirUnit {
                     }
                     if (OutAirUnit(OAUnitNum).OAEquip(compLoop).CoilPlantTypeOfNum == TypeOf_CoilSteamAirHeating) {
                         // DSU deal with steam mass flow rate , currenlty just like hot water  DSU?
-                        rho = GetDensityGlycol(PlantLoop(OutAirUnit(OAUnitNum).OAEquip(compLoop).LoopNum).FluidName,
-                                               60.0,
-                                               PlantLoop(OutAirUnit(OAUnitNum).OAEquip(compLoop).LoopNum).FluidIndex,
-                                               RoutineName);
+                        OutAirUnit(OAUnitNum).OAEquip(compLoop).MaxVolWaterFlow = GetSteamCoilMaxFlowRate(
+                            OutAirUnit(OAUnitNum).OAEquip(compLoop).ComponentType, OutAirUnit(OAUnitNum).OAEquip(compLoop).ComponentName, errFlag);
+                        Real64 TempSteamIn = 100.0;
+                        int SteamIndex = 0;
+                        Real64 rho = GetSatDensityRefrig(fluidNameSteam, TempSteamIn, 1.0, SteamIndex, RoutineName);
                         OutAirUnit(OAUnitNum).OAEquip(compLoop).MaxWaterMassFlow = rho * OutAirUnit(OAUnitNum).OAEquip(compLoop).MaxVolWaterFlow;
                         OutAirUnit(OAUnitNum).OAEquip(compLoop).MinWaterMassFlow = rho * OutAirUnit(OAUnitNum).OAEquip(compLoop).MinVolWaterFlow;
                         InitComponentNodes(OutAirUnit(OAUnitNum).OAEquip(compLoop).MinWaterMassFlow,
--- a/src/EnergyPlus/OutdoorAirUnit.cc
+++ b/src/EnergyPlus/OutdoorAirUnit.cc
@@ -129,6 +129,7 @@ namespace OutdoorAirUnit {
     using DataHVACGlobals::SmallAirVolFlow;
     using DataHVACGlobals::SmallLoad;
     using DataHVACGlobals::SmallMassFlow;
+    using DataSizing::ZoneEqOutdoorAirUnit;
     using namespace ScheduleManager;
     using namespace Psychrometrics;
     using namespace FluidProperties;
@@ -273,6 +274,8 @@ namespace OutdoorAirUnit {
             }
         }
 
+        ZoneEqOutdoorAirUnit = true;
+
         if (ZoneSizingCalc || SysSizingCalc) return;
 
         InitOutdoorAirUnit(OAUnitNum, ZoneNum, FirstHVACIteration);
@@ -282,6 +285,8 @@ namespace OutdoorAirUnit {
         // CALL UpdateOutdoorAirUnit(OAUnitNum, FirstHVACIteration)
 
         ReportOutdoorAirUnit(OAUnitNum);
+
+        ZoneEqOutdoorAirUnit = false;
     }
 
     void GetOutdoorAirUnitInputs()
@@ -318,7 +323,9 @@ namespace OutdoorAirUnit {
         using ScheduleManager::GetScheduleIndex;
         using SteamCoils::GetCoilAirInletNode;
         using SteamCoils::GetCoilAirOutletNode;
+        using SteamCoils::GetCoilMaxSteamFlowRate;
         using SteamCoils::GetCoilSteamInletNode;
+        using SteamCoils::GetCoilSteamOutletNode;
         using SteamCoils::GetSteamCoilIndex;
         using WaterCoils::CheckWaterCoilSchedule;
         using namespace DataLoopNode;
@@ -742,7 +749,12 @@ namespace OutdoorAirUnit {
                                     GetCoilSteamInletNode(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentIndex,
                                                           OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName,
                                                           ErrorsFound);
-
+                                OutAirUnit(OAUnitNum).OAEquip(CompNum).CoilWaterOutletNode =
+                                    GetCoilSteamOutletNode(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentType,
+                                                           OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName,
+                                                           ErrorsFound);
+                                OutAirUnit(OAUnitNum).OAEquip(CompNum).MaxVolWaterFlow =
+                                    GetCoilMaxSteamFlowRate(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentIndex, ErrorsFound);
                                 OutAirUnit(OAUnitNum).OAEquip(CompNum).MinVolWaterFlow = 0.0;
                                 // below: no extra error needed if steam properties not in input
                                 // file because getting the steam coil will have done that.
@@ -1100,8 +1112,8 @@ namespace OutdoorAirUnit {
         auto &GetWaterCoilMaxFlowRate(WaterCoils::GetCoilMaxWaterFlowRate);
         using DataSizing::AutoSize;
         using HVACHXAssistedCoolingCoil::SimHXAssistedCoolingCoil;
+        using SteamCoils::GetCoilMaxSteamFlowRate;
         using WaterCoils::SimulateWaterCoilComponents;
-        auto &GetSteamCoilMaxFlowRate(SteamCoils::GetCoilMaxWaterFlowRate);
 
         // Locals
         // SUBROUTINE ARGUMENT DEFINITIONS:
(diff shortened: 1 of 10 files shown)


src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1279,6 +1279,11 @@ namespace ReportSizingManager {
                         }
                     }
                 } else if (SizingType == CoolingWaterflowSizing) {
+                    if (ZoneEqOutdoorAirUnit > 0) {
+                        CoilDesWaterDeltaT = 0.5 * DataWaterCoilSizCoolDeltaT;
```
mjwitte: @Nigusse Why the 0.5 for outdoor air unit coils?  @rraustad How could this have been missing for so long? Wouldn't ChW fancoil sizing, for example, have failed?

Nigusse: @mjwitte That is what we do when cooling coils are on 100%OA stream such as for coils on OA system. OutdoorAirUnt is for 100% OA system, and I am assuming the same logic here as well.

mjwitte: @rraustad Why the 0.5? Seems arbitrary, will result in oversized plant water flow rate.

rraustad: This variable IS initialized in my Refactor Unitary System branch and I see it in other branches from a month ago so my guess is that someone mistakenly deleted this line lately which would make sense. The implementation for air loop equipment is:

    if (CurOASysNum > 0) {
        CoilDesWaterDeltaT = 0.5 * DataWaterCoilSizCoolDeltaT;
    } else {
        CoilDesWaterDeltaT = DataWaterCoilSizCoolDeltaT;
    }

This isn't what "we" do, it is what the developer did. This is wrong and should be corrected as a separate issue.

mjwitte: It happened [here](https://github.com/NREL/EnergyPlus/commit/fb7543800d9a4f53560fdb3c9e4d9209c4fe1a1f#diff-de781ad7ec80e690e175811f4da40a5aL1282) in a merge conflict resolution. We need to get that back in place sooner than waiting for the OA unit PR to get reviewed and merged.

rraustad: Heating coils in the OA systems are sized directly using plant delta T

    } else if (SizingType == HeatingWaterflowSizing) {
        ..snip..
        AutosizeDes = DesCoilLoad / (DataWaterCoilSizHeatDeltaT * Cp * rho);

rraustad: @Nigusse change this back the way is was. Which is just:

    CoilDesWaterDeltaT = DataWaterCoilSizCoolDeltaT;

rraustad: @mjwitte do you want me to push a quick change or are you doing that now?

mjwitte: @rraustad If you would, please.  Thanks.  Also maybe diff ReportSizingManager with one of your branches to make sure all the changes in it are legit.

rraustad: Created new branch and diff'ed with my Refactor UnitarySystem branch. Only missing code is this line. There is new code which looks like something I did recently. Will push this one line code change soon.

This is new:

    if (TermUnitSizing(CurTermUnitSizingNum).InducesPlenumAir) {
        AutosizeDes = (TermUnitFinalZoneSizing(CurTermUnitSizingNum).DesHeatCoilInTempTU * MinFlowFrac) +
                                          (TermUnitFinalZoneSizing(CurTermUnitSizingNum).ZoneRetTempAtHeatPeak * (1.0 - MinFlowFrac));
    } else {
        AutosizeDes = TermUnitFinalZoneSizing(CurTermUnitSizingNum).DesHeatCoilInTempTU * MinFlowFrac +
                                          FinalZoneSizing(CurZoneEqNum).ZoneTempAtHeatPeak * (1.0 - MinFlowFrac);
    }

Nigusse:
@Myoldmopar The new unit tests were added to OutdoorAirUnit.unit.cc. This file was excluded from build when this branch was created from develop. Now this file is included in this branch.

src/EnergyPlus/SteamCoils.cc
```diff
@@ -160,6 +160,12 @@ namespace SteamCoils {
     // MODULE SUBROUTINES:
 
     // Functions
+    void clear_state()
+    {
+        NumSteamCoils = 0;
+        GetSteamCoilsInputFlag = true;
+        SteamCoil.deallocate();
+    }
```
rraustad: The entire struct should be reset on clear_state? @Myoldmopar @mjwitte is this correct?

        // Default Constructor
        SteamCoilEquipConditions()
            : SteamCoilType(0), SteamCoilModel(0), SteamCoilType_Num(0), SchedPtr(0), InletAirMassFlowRate(0.0), OutletAirMassFlowRate(0.0),
              InletAirTemp(0.0), OutletAirTemp(0.0), InletAirHumRat(0.0), OutletAirHumRat(0.0), InletAirEnthalpy(0.0), OutletAirEnthalpy(0.0),
              TotSteamCoilLoad(0.0), SenSteamCoilLoad(0.0), TotSteamHeatingCoilEnergy(0.0), TotSteamCoolingCoilEnergy(0.0),
              SenSteamCoolingCoilEnergy(0.0), TotSteamHeatingCoilRate(0.0), LoopLoss(0.0), TotSteamCoolingCoilRate(0.0), SenSteamCoolingCoilRate(0.0),
              LeavingRelHum(0.0), DesiredOutletTemp(0.0), DesiredOutletHumRat(0.0), InletSteamTemp(0.0), OutletSteamTemp(0.0),
              InletSteamMassFlowRate(0.0), OutletSteamMassFlowRate(0.0), MaxSteamVolFlowRate(0.0), MaxSteamMassFlowRate(0.0), InletSteamEnthalpy(0.0),
              OutletWaterEnthalpy(0.0), InletSteamPress(0.0), InletSteamQuality(0.0), OutletSteamQuality(0.0), DegOfSubcooling(0.0),
              LoopSubcoolReturn(0.0), AirInletNodeNum(0), AirOutletNodeNum(0), SteamInletNodeNum(0), SteamOutletNodeNum(0), TempSetPointNodeNum(0),
              TypeOfCoil(0), FluidIndex(0), LoopNum(0), LoopSide(0), BranchNum(0), CompNum(0), Coil_PlantTypeNum(0), OperatingCapacity(0.0),
              DesiccantRegenerationCoil(false), DesiccantDehumNum(0), FaultyCoilSATFlag(false), FaultyCoilSATIndex(0), FaultyCoilSATOffset(0.0),
              reportCoilFinalSizes(true), DesCoilCapacity(0.0), DesAirVolFlow(0.0)

Nigusse: I think SteamCoil.deallocate() should clear all members.

Nigusse:
The CI machine is all green. This is ready for review and merge.

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -742,7 +749,13 @@ namespace OutdoorAirUnit {
                                     GetCoilSteamInletNode(OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentIndex,
                                                           OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName,
                                                           ErrorsFound);
+                                OutAirUnit(OAUnitNum).OAEquip(CompNum).CoilWaterOutletNode =
```
Myoldmopar, line 752: :+1: So this addresses the outlet node initialization crash.

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -1279,11 +1293,13 @@ namespace OutdoorAirUnit {
...
-                        // DSU deal with steam mass flow rate , currenlty just like hot water  DSU?
-                        rho = GetDensityGlycol(PlantLoop(OutAirUnit(OAUnitNum).OAEquip(compLoop).LoopNum).FluidName,
-                                               60.0,
-                                               PlantLoop(OutAirUnit(OAUnitNum).OAEquip(compLoop).LoopNum).FluidIndex,
-                                               RoutineName);
+                        OutAirUnit(OAUnitNum).OAEquip(compLoop).MaxVolWaterFlow =
+                            GetCoilMaxSteamFlowRate(OutAirUnit(OAUnitNum).OAEquip(compLoop).ComponentIndex, errFlag);
+                        Real64 rho = GetSatDensityRefrig(PlantLoop(OutAirUnit(OAUnitNum).OAEquip(compLoop).LoopNum).FluidName,
+                                                         DataGlobals::SteamInitConvTemp,
+                                                         1.0,
```
Myoldmopar, line 1300: :+1: So this gets the proper density

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -1594,7 +1610,7 @@ namespace OutdoorAirUnit {
             if (OutAirUnit(OAUnitNum).OAEquip(CompNum).CoilPlantTypeOfNum == TypeOf_CoilSteamAirHeating) {
                 if (OutAirUnit(OAUnitNum).OAEquip(CompNum).MaxVolWaterFlow == AutoSize) {
                     SimulateSteamCoilComponents(
-                        OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName, true, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentIndex);
+                        OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName, true, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentIndex, 1.0, 0.0);
```
Myoldmopar: I'm a little hesitant about what these new magic 1 and 0 are for...

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -108,7 +108,7 @@ set( test_src
   MixerComponent.unit.cc
   NodeInputManager.unit.cc
   OASystemHWPreheatCoil.unit.cc
-#  OutdoorAirUnit.unit.cc
+  OutdoorAirUnit.unit.cc
```
Myoldmopar, line 111: Re-enabling this is great, thanks for adding the clear state to SteamCoils.

Myoldmopar (approved):
I got a little confused about the stuff around the `CoilDesWaterDeltaT` variable with all the merge/revert, then saying it was fixed with this PR or something.  I'm assuming it's all resolved.  I'll pull develop into this branch and run all the unit tests to confirm, and if it passes I'll merge this.

src/EnergyPlus/OutdoorAirUnit.cc
Nigusse: @Myoldmopar I could not call an existing function without this magic arguments. Adopted from other similar calls from some where else, I think from UnitarySystem.

Nigusse:
@Myoldmopar CoilDesWaterDeltaT was accidentally deleted during merging a few weeks. I run into this problem, and Gu also run to similar problem. Finally Rich and Mike discussed to make quick fix and merge into develop instead of waiting for my fix in this branch.  No issue to worry at this point. It is this accidental delete that cause the OutdoorAirUnit unit test to fail.

src/EnergyPlus/OutdoorAirUnit.cc
Myoldmopar: But you didn't change the interface to SimulateSteamCoilComponents, right?  So why would you have to now pass additional variables -- that don't seem to make a lot of sense.  Those magic numbers 1.0 and 0.0 are going to be used as `QCoilReq` and `QCoilActual`.  I mean I'm also confused about why we are passing `true` in for `FirstHVACIteration`, but I know you didn't change that here, so I won't _sweat_ that.  (steam pun intended)

```
186     void SimulateSteamCoilComponents(std::string const &CompName,$
187                                      bool const FirstHVACIteration,$
188                                      int &CompIndex,$
189                                      Optional<Real64 const> QCoilReq = _, // coil load to be met$
190                                      Optional<Real64> QCoilActual = _,    // coil load actually delivered returned to calling componen    t$
191                                      Optional_int_const FanOpMode = _,$
192                                      Optional<Real64 const> PartLoadRatio = _);$
```

Maybe since we are forcing it to FirstHVAC, it doesn't make a difference, but I'm just not sure.  It feels like either good numbers should be going in, or none at all.

Myoldmopar:
Thanks for clarifying the CoildDesWaterDeltaT issue, I'd like a little more resolution about the Simulate call arguments and then this should be ready.  I pulled develop into the branch and ran the test suite locally.  All good.

src/EnergyPlus/OutdoorAirUnit.cc
rraustad: So this was never tested, and a unit test would have failed.

rraustad: How can an example file ever pass with this? Because we don't have one.

src/EnergyPlus/OutdoorAirUnit.cc
rraustad: I'll give you a 1.0 and 0.0 for trying, but delete those and figure something else out.

tst/EnergyPlus/unit/OutdoorAirUnit.unit.cc
```diff
@@ -358,4 +371,618 @@ TEST_F(EnergyPlusFixture, OutdoorAirUnit_AutoSize)
...
+    Real64 EnthSteamIn = GetSatEnthalpyRefrig("STEAM", DataGlobals::SteamInitConvTemp, 1.0, SteamCoil(1).FluidIndex, "");
+    Real64 EnthSteamOut = GetSatEnthalpyRefrig("STEAM", DataGlobals::SteamInitConvTemp, 0.0, SteamCoil(1).FluidIndex, "");
+    Real64 SteamDensity = GetSatDensityRefrig("STEAM", DataGlobals::SteamInitConvTemp, 1.0, SteamCoil(1).FluidIndex, "");
+    Real64 CpOfCondensate = GetSatSpecificHeatRefrig("STEAM", DataGlobals::SteamInitConvTemp, 0.0, SteamCoil(1).FluidIndex, "");
+    Real64 LatentHeatChange = EnthSteamIn - EnthSteamOut;
+    Real64 DesMaxSteamVolFlowRate = DesSteamCoilLoad / (SteamDensity * (LatentHeatChange + SteamCoil(1).DegOfSubcooling * CpOfCondensate));
+
+    // check water coil water flow rate calc
+    EXPECT_EQ(DesSteamCoilLoad, SteamCoil(1).DesCoilCapacity);
+    EXPECT_EQ(DesMaxSteamVolFlowRate, SteamCoil(1).MaxSteamVolFlowRate);
```
rraustad, line 986: I won't say no to this ( I might tomorrow), but 600 lines of code to prove capacity = load, we have to get better at this,

src/EnergyPlus/OutdoorAirUnit.cc
Nigusse: I have spent enough time to figure it out but no success. Finally I tried tried mimic what is done else where.

src/EnergyPlus/OutdoorAirUnit.cc
rraustad: How did this pass CI before? We need more test cases.

tst/EnergyPlus/unit/OutdoorAirUnit.unit.cc
Nigusse: I thought the unit test were intended to demonstrate the fix. This unit test demonstrates just that. Before the fix, steam coil cannot be used in Outdoor Unit and I am demonstrating that and also demonstrating how the steam density is used to calculate steam flow rate.

src/EnergyPlus/OutdoorAirUnit.cc
rraustad: @Nigusse point me to an example file that has this configuration and tests this defect? I know it doesn't exist, I just want to start the conversation.

Nigusse: Just added a defect file that created. [Here](https://github.com/NREL/EnergyPlusDevSupport/tree/master/DefectFiles/6000s/6843).

tst/EnergyPlus/unit/OutdoorAirUnit.unit.cc
Myoldmopar: @Nigusse I think @rraustad is just lamenting that it took 600 lines of test code to demonstrate it.  If I'm reading it properly, you did everything just fine.  As we continue to objectify the code, the amount of code required to create nice unit tests **will** reduce, and it will become an easier lift to write more and more tests.

src/EnergyPlus/OutdoorAirUnit.cc
Nigusse: @Myoldmopar @rraustad The problem is that we did not have a check for optional variable QCoilReq [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SteamCoils.cc#L233-L236) in order to avoid the magic numbers 1.0 and 0.0. I added the following check for QCoilReq and replaced it with local variable and resolved the problem. 
        if (present(QCoilReq)) {
            QCoilReqLocal = QCoilReq;
        } else {
            QCoilReqLocal = 0.0;
        }
I just pushed an updated branch.

Issue #6857, opened 2018-07-25
AFN wind angles in the EIO are incorrect for advanced single-sided natural ventilation


jasondegraw:
Issue overview
--------------
The wind angles that are being output for advanced single-sided natural ventilation are the default angles (in increments of 30 degrees) rather than the higher-frequency angles for the feature (in increments of 10 degrees).

### Details
Some additional details for this issue (if relevant):
 - Version 8.9

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Ticket added to Pivotal for defect (development team task)
 - [x] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6858

PR #6858, opened 2018-07-25
Correct EIO Wind Angle Output

jasondegraw:
Pull request overview
---------------------
The angles being written to the EIO when the advanced single-sided natural ventilation feature is used are not correct. Reordering the logic a bit fixes the output so that the correct values are used. Addresses #6857.

### Work Checklist

### Review Checklist
 - C++ checks:
   - [ ] Argument types


jasondegraw:
Oops, accidentally submitted this off of my own fork. That was not intentional, I can move the branch into this repo if need be.

nrel-bot-3:
@jasondegraw @lgentile it has been 14 days since this pull request was last updated.

mjwitte:
@jasondegraw This is blocked by #6912. But in the meantime, please add a description of the output changes to the [output rules doc](https://github.com/NREL/EnergyPlus/blob/develop/src/Transition/OutputRulesFiles/OutputChanges8-9-0-to-9-0-0.md).

mjwitte:
This code works as intended. However, the Initialization Summary table output doesn't display the full tables in the html (not sure about sql), likely because the row headers aren't set up with all of the columns.
image
A for loop to add fill out the column headings of "Wind Direction #2 {deg}" etc. is needed [here](https://github.com/NREL/EnergyPlus/blob/f2e822120ce2620da9ab77c203634f1feb6e4eb8/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L2571).

And for the wind coefficients table, those column headings should also be "Wind Direction #1 {deg}", "Wind Direction #2 {deg}", etc. That code is [here](https://github.com/NREL/EnergyPlus/blob/f2e822120ce2620da9ab77c203634f1feb6e4eb8/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L2594).

mjwitte:
@jasondegraw Any chance we can get this fixed today?

jasondegraw:
@mjwitte I doubt it, and since this has been this way at least 6 months with no reports, I'm inclined to think it's OK to push it to the next release.

mjwitte:
@jasondegraw OK. I'll drop this so that the eio angles are all there, and then post a new issue to fix the tables.

mjwitte:
drop this = drop this in = merge . . . but after RunPeriod is merged.

Issue #6796, opened 2018-06-23
Heat exchanger bypass control problems


mjwitte:
Issue overview
--------------
User file has heat exchanger with BypassWhenOAFlowGreaterThanMinimum, with economizer set to NoLockout.  HX should be running in cold weather, but it's not. There is a Controller:MechanicalVentilation with VentilationRateProcedure control, not sure if this is contributing or not.  

Detailed output for 01/02, midday, shows that VAV_BOT WITH REHEAT:Air System Outdoor Air Minimum Flow Fraction [](TimeStep) is running at 0.16, but
VAV_BOT   WITH REHEAT:Air System Outdoor Air Flow Fraction [](TimeStep) is running around 0.39.  These should be equal, because the economizer is calling for minimum OA because it's cold out.

Actual MinOAFrac > SysOAMinFlowFrac is confusing the BypassWhenOAFlowGreaterThanMinimum control, so it thinks the economizer is calling for extra OA, thus it bypasses the HX.

Which is wrong?  Not sure.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v8.9.0
 - Direct e-mail 2018-06-20

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6861

PR #6861, opened 2018-07-26
Addressed Heat exchanger bypass control problems 

Nigusse:
Pull request overview
---------------------
This is a defect. The Minimum Outdoor Air Fraction and actual Outdoor Air Fraction are calculated using a different basis. Outdoor air fraction calculated using a different basis is invalid for comparison and leads to incorrect HX bypass status. Diffs are expected.
Addresses #6796.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
The Minimum Outdoor Air Fraction based on design supply air mass flow rate is calculated [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3580) and [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3613), where as the actual Outdoor Air Fraction is first estimated based on mixed air setpoint [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4413), and based on based on current mixed air mass flow rate using rootsolve  [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4413) and finally [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3754). As the result the same amount of OA mass flow rate can have OASignal > OutAirMinFrac can lead to incorrect determination of HX bypass status [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4655-L4658).

Nigusse:
Sample results before the fix.

6796_outairfrcmin



Sample result after the fix:

6796_outairfrcminafterfix

mjwitte:
Sure seems at first glance that the current definition isn't very useful (min OA divided by design supply). Have you checked to see if `OutAirMinFrac` is used in any other way that might break if the basis was changed?

Nigusse:
Yes, there are a couple of places in the code were that the OutAirMinFrac is used incorrectly.
(1) OutAirMinFrac is passed to CalcOAEconomizer function [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3645) and subsequently used to calculate "OA Node" mass flow rate [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4563-L4566)  by multiplying it with the "Mixed Air Node" mass flow rate.
(2) OutAirMinFrac is used to determine HX bypass status by comparing it with actual OA fraction (OASignal) [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4654-L4659). And the actual OA fraction is calculated using the Mixed Air Flow rate as a basis.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3639,6 +3639,12 @@ namespace MixedAir {
                 MixedAirTempAtMinOAFlow = Node(this->RetNode).Temp;
             }
             this->MixedAirTempAtMinOAFlow = MixedAirTempAtMinOAFlow;
+
+            if (this->MixMassFlow > 0) {
+                // calc outdoor air minimum fraction based on actual mixed air flow
+                Real64 OutAirMinFracTemp = curAirLoopFlow.MinOutAir / this->MixMassFlow;
+                OutAirMinFrac = OutAirMinFracTemp;
+            }
```
mjwitte: Rather than make this change here, why not change [line 3580](https://github.com/NREL/EnergyPlus/pull/6861/files#diff-19711df337cdc7fe75d3e87b4c0cd502R3580) to use `MixMassFlow` - or `curAirLoopFlow.SupFlow`. And then the calculation for `MechVentOutsideAirMinFrac` [here](https://github.com/NREL/EnergyPlus/pull/6861/files#diff-19711df337cdc7fe75d3e87b4c0cd502R3613) would also be changed to use `.SupFlow` or simply `SysSA`. Hopefully that doesn't break anything other than changing the values reported for "Air System Outdoor Air Minimum Flow Fraction" and fixing HX bypass problems.

Nigusse: @mjwitte Thanks. I will try your suggestion and test it locally and see how it goes.

Nigusse: @mjwitte Just for your information. Since we are now passing a new OutAirMinFrac value calculated with different basis to CalcOAEconomizer function, then the new OutAirMinFrac value used [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4563) may cause diffs.

mjwitte: Yeah, this may need further work, because there's the issue of cycling fans and FixedMin which doesn't get applied until later (but seems it should be done before this).

Nigusse: I think some of the checks should come after CalcOAEconomizer function call to cap the OA flow based on the various restrictions that may apply and the OA flow determined by the economizer.

Nigusse:
@mjwitte @rraustad This fix addresses the Air HX bypass control without changing the minimum OA flow fraction calculation. Now compares the actual OA mass flow to minimum OA mass flow. A few of the checks for *.HeatRecoveryBypass and *.HeatRecoveryBypassStatus in the unit test MixedAir_HXBypassOptionTest failed due to this fix were corrected.

rraustad:
I like that much more than changing the basis of the calculations. The OutAirMinFrac is useful and identifies the minimum OA mass flow rate when used correctly. This looks good to me now.

Nigusse:
I still need to post new issue for some of the cases that does look incorrect.

Nigusse:
Pulled in develop to this branch to make sure some of the unexpected diffs (Design Size No Load Supply Air Flow Rate in UnitarySystem example files) are caused by lag from develop.

nrel-bot-3:

Myoldmopar:
CI got stuck on this a few weeks back, plus it got pushed to low priority with the IO freeze.  I pulled develop again.  CI is generally behaving nicely these days (except Mac), so let's see this turn green and push it along.

src/EnergyPlus/MixedAir.cc
```diff
@@ -4655,7 +4655,9 @@ namespace MixedAir {
                     AirLoopControlInfo(AirLoopNum).HeatRecoveryBypass = true;
                     this->HeatRecoveryBypassStatus = 1;
                 } else if (this->HeatRecoveryBypassControlType == BypassWhenOAFlowGreaterThanMinimum) {
-                    if (OASignal > OutAirMinFrac) {
+                    Real64 OAMassFlowMin = OutAirMinFrac * AirLoopFlow(AirLoopNum).DesSupply;
+                    Real64 OAMassFlowActual = OASignal * this->MixMassFlow;
+                    if (OAMassFlowActual > OAMassFlowMin) {
                         AirLoopControlInfo(AirLoopNum).HeatRecoveryBypass = true;
```
mjwitte, line 4661: Such a simple, straightforward change, yet it introduces max iteration warnings . . . ugh. Why? Investigating.

mjwitte (commented):
Also, the original diagnosis was somewhat incorrect. The mixed air setpoint is outdoor air reset with a upper limit of 15.6C. This frequently causes the economizer to bring in extra outdoor air in cold weather during unoccupied hours when the minOA=0, and then this trips the HX bypass making matters even worse. Because this system has reheat coils and no central heating coil, the lockout with heating never kicks in. Looking at this now, it's not clear to me there was a real problem before. Will keep looking.

src/EnergyPlus/MixedAir.cc
mjwitte: Looking at detailed output with more output variables. The bypass flag is working as expected now, it's only 1 now when the OA mass flow is > the min OA (which can be reported using Air System Outdoor Air Mechanical Ventilation Requested Mass Flow Rate for this file anyway). So that much is ok.  But there are times when it is bouncing around from one HVAC timestep to the next, e.g. 01/02 08:00-10:00. During this time, the system flips betwen two states - HX active at min OA flow with a mixed air temp of 17C or so, or HX bypassed with OA flow > min OA and mixed air temps of 10C or less.  The times where it's bypassed, the OA flow must have been calculated with the HX active resulting in a warmer OA inlet node temp so the economizer opened up to reach the mixed air setpoint but then that caused the HX bypass to kick in and the next iteration comes around and the air is cold now.

*IF* this file had LockoutWithHeating instead of NoLockout, then the bouncy problem goes away. There's logic [here](https://github.com/NREL/EnergyPlus/blob/159313706_Issue%236796/src/EnergyPlus/MixedAir.cc#L4385-L4394) to avoid the no lockout nonsense we're seeing here. But this only executes for economizer LockoutWithHeating. Maybe this should execute also if BypassWhenOAFlowGreaterThanMinimum?

But even with LockoutWithHeating when the minimum OA is zero, then the MixedAirTempAtMinOAFlow = ReturnTemp which is 23C so the economizer thinks it should open up to reach 15.6C mixed air and then that triggers bypassing the HX which then results in a mixed air temp well below 15.6C. That's bad.

Is the [calculation of MixedAirTempAtMinOAFlow](https://github.com/NREL/EnergyPlus/blob/159313706_Issue%236796/src/EnergyPlus/MixedAir.cc#L3637-L3645) correct if there's an HX that could possibly be bypassed? Looking at the output for "Air System Outdoor Air Heat Recovery Bypass Minimum Outdoor Air Mixed Air Temperature" is seems like it's getting a cold value, but only when the min OA is > 0.

Thinking out loud here and documenting things for further review.

And this code is bad:
`this->MixedAirTempAtMinOAFlow = MixedAirTempAtMinOAFlow` - there shouldn't be a local MixedAirTempAtMinOAFlow declared earlier, delete that and this line.

Will revisit this tomorrow to propose a fix. Any suggestions from fresh eyes are welcome. @Myoldmopar @Nigusse

Nigusse: @mjwitte Looking at the code MixedAirTempAtMinOAFlow is calculated using Outdoor Air Node (OANode) temperature [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3639) where as the actual OA fraction (OutAirSignal) is calculated using mixer inlet node (InletNode, which is the heat recovery OA outlet node) temperature [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4416). These temperature can be different depending on the status of the heat recovery HX. Could this be an cause for HX bypassed status flip?

mjwitte: Yes, I think that's part of it. The thorny part is that the OA controller doesn't know what kind of equipment sits between OANode and InletNode, and it doesn't know if that equipment will be active or not. In the case of the HX, even that we don't know for sure that the HX Economizer Lockout = Yes. So it's not clear that `MixedAirTempAtMinOAFlow` could ever happen in some models. But for the original test file, this section doesn't execute because LockoutWithHeating is not set in the idf.

But for the test file *which does have LockoutWithHeating` the max iteration warnings are gone. It still doesn't do the right thing when minOA=0, but first we should get rid of the max iteration problem. 

The problem is likely that `InletTemp` will be different depending on whether the HXbypass flag was true on the prior iteration. There's some flag hocus pocus going on [here](https://github.com/NREL/EnergyPlus/blob/d7683729b22e5d468703c6d90789aa609a5f674e/src/EnergyPlus/MixedAir.cc#L3725-L3748) that's probably part of this.

mjwitte: @Nigusse I tried moving the HX bypass check up before [this block](https://github.com/NREL/EnergyPlus/blob/159313706_Issue%236796/src/EnergyPlus/MixedAir.cc#L4541-L4613) which includes a re-simulation of the OA equipment. This got rid of the max iterations, but for some reason, the HX *never* goes into bypass with that. It may be necessary to call `SimOASysComponents` twice at this point (once with `HeatRecoveryBypass = true` and again with it `false`) and save the two InletTemp values and make a decision that way?  I'm fishing here for a robust solution.

Nigusse: @mjwitte I did further investigation using a new example file with single zone and found out that the minimum and initial actual OA amount comparison based on fractions [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4544-L4545) leads to wrong implication. Because of the different basis used to calculate the fractions, OutAirSignal > OutAirMinFrac can be true even if the minimum and actual (initial estimate) amounts are the same. In the example file I investigated this happens for days. I will modify the comparison to be based on absolute amounts instead of fraction, just like I did for HeatRecoveryBypassStatus check.

Nigusse: @mjwitte Comparison based on OA amount rather than fan fraction did not change the results for the test but sometimes it definitely avoids unnecessary SolveRoot call. The reason that the results did not change is that the HXBypass status is already captured by the original fix, I think. 

But I confirmed that MixedAirTempAtMinOAFlow calculated with and without HXBypass can be significantly different. For my test case I have seen (MixedAirTempAtMinOAFlow - MinedAirTemp) difference of as low as -2.5K deltaT when the HX is on and as high as 6.0K deltaT when the HX is bypassed.

mjwitte: @Nigusse I've reverted my change, so things are back to the way you originally proposed. I will post a new issue regarding the max iterations. In many respects this bypass control doesn't make sense without also having LockoutWithHeating. That can be the workaround and can be explored in a reorg of this complicated controller which has many special parts all to fix various special problems.

PR #6872, opened 2018-07-31
Improve Kiva longwave radiation methodology and performance

nealkruis:
Pull request overview
---------------------
Updates EnergyPlus to use Kiva 0.5.0 (soon to be released!) which contains an enhanced interior radiation model and performance improvements. The new radiation model allows the foundation surfaces to exchange longwave radiation on a cell-by-cell basis (rather than on a surface-by-surface basis--which could cause some unrealistic oscillations at the cell level).

It also fixes an issue where the longwave radiation from foundation walls was being double counted in the heat balance. These changes will impact results, generally decreasing the magnitude of heat loss through the foundation walls.

Work completed for Modification No. 1 to Task Order KAGN-4-42502-06 under Task Ordering Agreement No. KAGN-4-42502-00: "EnergyPlus Modeling and Simulation Software Development and Support".

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


nealkruis:
@mjwitte @Myoldmopar @shorowit I think this is ready for someone to start looking at it and getting some feedback. I've got some more minor changes coming as I finalize the corresponding Kiva release. Let me know if you have any questions.

nrel-bot-3:

Myoldmopar:
@nealkruis this is showing some odd cppcheck [warnings](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-48e55c3b3caf127560b371edd3d6c4e3f218b165-x86_64-Linux-Ubuntu-14.04-cppcheck-1.61.html).  I can't really make out why it's giving the warnings it's giving, but it seems like we're not the only ones [confused](https://stackoverflow.com/a/4872612/2600110).  As far back as 2011, cppcheck was apparently reporting this wrongly.  According to Ubuntu's package [system](https://packages.ubuntu.com/search?keywords=cppcheck), the CI machine is likely using cppcheck version 1.61, which according to cppcheck's [release log](https://sourceforge.net/p/cppcheck/news/?page=1), was released in August 2013.  Since then there have been a number of releases, and when we upgrade our CI boxes to 18.04, we should be up to 1.82, which was released in January 2018.  Hopefully in that 4.5 year period, they fixed that issue, but it's hard to tell because the full release notes aren't shown on that release log page.

Anyway, I think we'll have to add some cppcheck muting to get this to go away before then.  Could you add the muting to the CI's [yaml file](https://github.com/NREL/EnergyPlus/blob/8f99a9e9dfa40792902ba4409858d34fa3b494fd/.decent_ci-Linux.yaml#L24)?  I couldn't find an immediate answer to exactly what string to add to the suppression from cppcheck's [manual](http://cppcheck.sourceforge.net/manual.pdf), but I'm sure you can find something.

mbadams5:
@nealkruis @Myoldmopar After a quick look at the warning and the code in question, it may be worth a quick refactor. For example, [with this one](https://github.com/NREL/EnergyPlus/blob/48e55c3b3caf127560b371edd3d6c4e3f218b165/third_party/kiva-ep/src/libkiva/Cell.cpp#L153) I would instead try the following if you are trying to initialize the array to zero:
```cpp
double C[3][2] = {0};
```

[For this one](https://github.com/NREL/EnergyPlus/blob/48e55c3b3caf127560b371edd3d6c4e3f218b165/third_party/kiva-ep/src/libkiva/Ground.cpp#L157), I think it is just getting hung up on the brace initialization for some reason with the older version of cppcheck. You could convert to `std::size_t index(0);`.

nealkruis:
Ok. So options are:
1. Mute the warnings (unmute them after we update CI)
2. Refactor code to not use list initializations (we're not exactly using them consistently in kiva anyway, though they are [considered(?) best practice](https://stackoverflow.com/questions/18222926/why-is-list-initialization-using-curly-braces-better-than-the-alternatives))
3. Do nothing knowing they are false positives and update cppcheck after release (protecting us in the unlikely event that legitimate warnings are accidentally muted).

I don't have a strong preference, though I'd prefer 1 or 3 unless @mbadams5 can make a convincing argument that I shouldn't be using list initializations.

Myoldmopar:
I will need some more info before I can support item 2.  I would argue that item 3 does **not** protect us, as the dashboard would always be yellow/red, and other issues could sneak under the radar.  I think item 1 will be the best path forward, and only mute warnings specifically for this error, and specifically in this file.  You should see patterns for that in the existing decent_ci yaml file for cppcheck.

nealkruis:
@Myoldmopar cppcheck warnings are gone. Appears to be a problem with integration tests on the Mac CI (occurs on other branches as well).

PR #6885, opened 2018-08-03
Add API function RunEnergyPlus that uses return codes instead of std::exit()

michael-okeefe:
Pull request overview
---------------------
**Summary**: This pull request adds an additional library interface (int RunEnergyPlus(...)) to EnergyPlus that passes return codes for success/failure back to the caller and also modifies streams and how streams are handled such that stdin and stdout are not accidentally closed or put in an errored state. The previous interface (void EnergyPlusPgm(...)) that calls std::exit() is retained for backwards compatibility. Note: this pull request contains no changes to the ObjexxFCL library. The behavior that was seen where stdin/stdout/stderr (cin/cout/cerr) was reported separately and a fix is being considered for the next revision of ObjexxFCL that will prevent stdin/stdout/stderr from going into an errored state. This PR contains a workaround that checks for whether cin/cout/cerr are good and, if not, clears their error bits which effectively resolves the issue.

-----------------

A new feature proposal for this effort is available: ["Add C-API for EnergyPlus DLL and API Method that uses Return Codes instead of std::exit()"](https://github.com/NREL/EnergyPlus/blob/capi_pull_request/design/FY2018/NFP-capi-and-return-codes.md)

-----------------

In recent work with using EnergyPlus as a library, EnergyPlus's default behavior to call std::exit has proved difficult to work with. As a clear example of what is not working correctly, please see file `src/EnergyPlus/test_ep_as_library.cc`.  Add the following line to the existing version of the code after the call to EnergyPlusPgm(...):

`std::cout << "THIS LINE SHOULD DISPLAY." << std::endl;`

If you run the modified `test_ep_as_library.cc` against the current EnergyPlus, the line is not printed out for two reasons:

1. The call to EnergyPlusPgm(...) (indirectly) calls std::exit(...) which exits the program (and anything calling EnergyPlus in the same process) and thus your code never returns from EnergyPlusPgm to do any follow-on tasks
2. Even if you modify EnergyPlusPgm(...) to not call std::exit(...), you will be able to return from the simulation but you will find that std::cout is not ready to write to (the stream is in an errored state); this seems to be due to a combination of an explicit call to close these streams and also calls that attempt to ask the "size" of streams such as stdout (cout). 

This pull request:

- adds a new interface called RunEnergyPlus(...) which preserves the previous interface's name (EnergyPlusPgm) and behavior (to exit via std::exit). This allows calling an energyplus simulation as a library without exiting the caller's process.
- proposes one possible way to handle stdin/stdout/stderr so as to resolve any error states upon return from the simulation call
- modifies `src/EnergyPlus/test_ep_as_library.cc` to demonstrate this newly working functionality

NOTE: The pull request ["Provide a C-API for the EnergyPlus DLL for ease of calling from other programming languages (6893)"](https://github.com/NREL/EnergyPlus/pull/6893) is a superset of this pull request. It includes all the changes suggested here and additionally adds a C-API for ease of calling the EnergyPlus DLL from other programming languages.

**Work completed under the Big Ladder Contribution License Agreement.**

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/EnergyPlusPgm.cc
```diff
@@ -224,6 +224,11 @@
 #endif
 
 void EnergyPlusPgm(std::string const &filepath)
+{
+    std::exit(EnergyPlusPgmReturnCodes(filepath));
+}
+
+int EnergyPlusPgmReturnCodes(std::string const & filepath)
```
Myoldmopar: I really thought from the PR description that this function was going to give me the possible return codes from E+.  We'll figure out a better name.

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -597,7 +597,7 @@ void AbortEnergyPlus()
     // Close the socket used by ExternalInterface. This call also sends the flag "-1" to the ExternalInterface,
     // indicating that E+ terminated with an error.
     if (NumExternalInterfaces > 0) CloseSocket(-1);
-    std::exit(EXIT_FAILURE);
+    return EXIT_FAILURE;
```
Myoldmopar, line 600: I was a bit nervous about this, because of course a call to ShowFatal needs to result in the program stopping instantly, not returning and accidentally continuing.  After some browsing in the code, however, I see that this is only called in one spot, EnergyPlusPgm, and only if an exception is thrown during the simulation, and is immediately followed by the EndEnergyPlus call.  Since ShowFatal raises an exception now (instead of calling AbortEnergyPlus), this is a reasonable approach.  :+1:

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -719,12 +724,18 @@ void CloseOutOpenFiles()
             exists = flags.exists();
             opened = flags.open();
             ios = flags.ios();
+            name = flags.name();
+        }
+        if (exists && opened && ios == 0) {
```
Myoldmopar, line 729: A little hacky, but makes sense.

src/EnergyPlus/EnergyPlusPgm.cc
jasondegraw: Seems like `RunEnergyPlus` is the consensus on a better name.

src/EnergyPlus/test_ep_as_library.cc
```diff
@@ -64,10 +64,17 @@ int main(int argc, char *argv[])
...
-        return 1;
+        return EXIT_FAILURE;
     } else {
-        EnergyPlusPgm(argv[1]);
+        status = EnergyPlusPgmReturnCodes(argv[1]);
     }
+    if (!std::cin.good()) std::cin.clear();
+    if (!std::cerr.good()) std::cerr.clear();
+    if (!std::cout.good()) std::cout.clear();
+    std::cerr << "THIS LINE SHOULD DISPLAY." << std::endl;
```
jasondegraw: Do we still need these lines here (this one and the next one)? If we do, I suggest changing them to print out "Standard error is still available for use" and "Standard output is still available for use". The context for the addition of these lines are here will fade over time, so it's better to change them to something more self explanatory.

jasondegraw (changes requested):
After going through the modifications, I'd like to see a couple changes: (1) change the function name (RunEnergyPlus seems to be the consensus) and (2) edit the messages at the end of the library test program. Other than those two minor things, it looks good to me.

src/EnergyPlus/test_ep_as_library.cc
michael-okeefe: I like the suggested changes to the lines. @DeadParrot gave me a preview of the new ObjexxFCL and that did indeed fix the stdin/stdout issue but it may be nice to have code in this example use file that explicitly demonstrates that there is activity after the RunEnergyPlus call (showing that we're not just exiting anymore). That said, happy to go either way on whether we keep the lines or not.

jasondegraw (approved):

nrel-bot-3:
@michael-okeefe @lgentile it has been 14 days since this pull request was last updated.

Issue #6881, opened 2018-08-02
Incorrect minimum OA amount when the Minimum Limit Type is FixedMinimum 


Nigusse:
Issue overview
--------------
Incorrect minimum OA mass flow rate calculation when the OA controller Minimum Limit Type is  FixedMinimum. It uses the calculation method applicable for ProportionalMinimum. This was found while working on another bug issue #6796. 

### Details
 - V8.9 (or even older versions)

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6886

Issue #6819, opened 2018-07-10
Controller:OutdoorAir Minimum Limit Type not behaving as expected


rmcclung:
In version 8.9 the controller:outdoor air seems to be affecting ventilation rates and mixed air flow rates in strange ways, that I suspect are tied to the minimum limit type operating as proportional minimum despite setting it to fixed minimum.

In this file, I would like all the CAHU air handlers to provide the minimum ventilation rates during the day, and only operate to provide heating or cooling if required at night.  

In the AEM file, the air handlers do operate at night, but the mixed air flow rate seems suspiciously fixed, and higher than necessary.  Operation during the day is as expected.

In the AEM 2 file, the air handlers do not operate at night, despite the fact that CAHU-17 should be as there are zones below the setpoint.  The only change was adding the minimum outdoor air schedule to the DSOA objects in addition to specifying them in the controller OA objects.  This seems like the system is operating with a proportional minimum OA, but because the min OA rate is set to 0, there is no mixed flow either.

For another project, I also had problems with a 100% OA unit that was not supplying 100% OA because I had hard sized the fans and terminals, but the controller OA was providing the proportional minimum OA according to the autosized value for the system, which was larger than the hard sized values, instead of the actual fixed minimum flow rate.

[NTC PROP AEM 2.zip](https://github.com/NREL/EnergyPlus/files/2181585/NTC.PROP.AEM.2.zip)

PR #6886, opened 2018-08-03
Incorrect minimum OA amount when the Minimum Limit Type is FixedMinimum

Nigusse:
Addresses Issue #6881 

Pull request overview
---------------------
This fix changes how the fixed minimum OA amount is calculated by OA controller. Now it honors the minimum OA amount depending the minimum value specified. The Fixed Minimum OA amount depends on how the minimum amount specified or autosized value, applicable minimum schedule value multipliers, or Mechanical Ventilation Control if specified. Diffs are expected if the OA controller Minimum Limit Type input is FixedMinimum.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


Nigusse:
The OA controller in CalcOAEconomizer() function [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4574) uses the same steps (method) to calculate the minimum OA amount and set the value to an OutdoorAirNode. What this code does is that, if the mixed air flow is zero, the minimum OA flow amount is always set to zero. This step work only for ProportionalMinimum Limit Type but it has been used for FixedMinimum Limit Type as well. This is incorrect and addressed in this fix. The following four diagram shows results before and after the fix based on example file 5ZoneSupRetPlenRAB.idf.  The green line plots show results after the fix.

Whenever the mixed air flow rate is zero (see the 4th diagram), the minimum OA amount is always zero before the fix (V8.9 and develop version results) and the results after the fix show non-zero value as it should be for non-zero fixed minimum value specified. The system actual OA flow fraction is non-zero after the fix whereas it is zero before the fix (see date 01/14, 0:00 -7:00 am hours, the 1st diagram). The 2nd diagram also shows zero minimum OA flow fraction for the same reporting period. For FixedMinimum Limit Type the minimum OA flow fraction should have been non-zero if there is a minimum amount specified as is the case in this defect file. This was not honored before the fix.

6881_oafractionflow

Nigusse:
Sample results of effect of this fix on the main heating and cooling coils heat/cool rate and annual cooling and heating energy end uses are shown below. The non-zero and higher Minimum OA Flow Rates may increase the annual cooling and heating energy end uses as shown below for this defect file. The green line plots and the green values in the table below show results after the fix. 
 
6881_oafractionflowheatingcoils

mjwitte:
@Nigusse Not sure I'm following this completely. This could be a useful change as a new feature, but the current behavior is not a defect. The original design was that mixed air flow rate is set by other criteria and the OA flow rate is never > mixed air flow.  So, a mixed air flow of zero should result in zero OA flow. Some users have requested the ability to override and cause the system to operate to meet OA flow requirements even if this means increasing mixed air flow rate (but I can't find a documented issue with this request), so this could be transformed into a new feature but will require a new input or new input option to trigger it.  Also, I don't know how overriding the mixed air flow rate here will play with the unitary systems that want to control the supply air flow rate based on load and/or multi-speed coils.

mjwitte:
@Nigusse Also, is this possibly related to #6819?

Nigusse:
@mjwitte Thank you for your feedback. I tried to interpret what a fixed minimum is expected to behave. My understanding of fixed minimum is to provide a fixed minimum OA amount depending user specification including setting it to zero based on scheduled multipliers. Prorating the OA flow just like ProportionalMinimum for FixedMinimum as well seemed incorrect to me. In this fix, I allowed how the minimum OA flow rate is applied depending on the Minimum Limit Type; otherwise, I do not think I changed/override the mixed air flow rate. Any ways, I am open to modify it if this is not the intended definition of FixedMinimum.

Nigusse:
@mjwitte The fix in this branch addresses one of the problems described in issue #6819 although I have not run the defect files attached in the issue.

mjwitte:
My comment about the mixed air flow changing, was based on the bottom graph in  [your comment above](https://github.com/NREL/EnergyPlus/pull/6886#issuecomment-410331955) which shows a large change in mixed air mass flow rate. If you modify this fix to limit OA flow to be no greater than the current mixed air flow, then I think it will work as intended.

Nigusse:
If I modify this fix to limit OA flow to be no greater than the current mixed air flow and the current mixed air flow rate is zero, then the minimum OA amount will also be zero. My concern is that we are not allowing the OA controller to have the option to provide a minimum OA amount regardless of the mixed air flow rate. Probably we need a new feature to add this capability. What do you think @mjwitte?

Nigusse:
Modified the fix to limit the minimum OA mass flow rate not exceed mixed air node mass flow rate.

mjwitte:
Yes, a new feature request would be appropriate, some type of system override to enforce a minimum supply air flow rate. Just to clarify here, Controller:OutdoorAir is intended to control the outdoor air damper position, so it has no authority over the rest of the system operation, including the mixed air flow rate. This would need to be some other higher-level control.

nrel-bot-3:

src/EnergyPlus/MixedAir.cc
```diff
@@ -3714,7 +3714,11 @@ namespace MixedAir {
             auto &curAirLoopFlow(AirLoopFlow(AirLoopNum));
 
             curAirLoopFlow.OAMinFrac = OutAirMinFrac;
-            curAirLoopFlow.MinOutAir = OutAirMinFrac * this->MixMassFlow;
+            if (this->FixedMin) {
+                curAirLoopFlow.MinOutAir = min(OutAirMinFrac * curAirLoopFlow.DesSupply, this->MixMassFlow);
+            } else {
+                curAirLoopFlow.MinOutAir = OutAirMinFrac * this->MixMassFlow;
+            }
```
Myoldmopar, line 3721: OK, so this section just constrains the minimum outside air to the current mixed air flow rate.

src/EnergyPlus/MixedAir.cc
```diff
@@ -4568,10 +4572,18 @@ namespace MixedAir {
...
+                            min(max(this->ExhMassFlow, OutAirMinFrac * AirLoopFlow(AirLoopNum).DesSupply), Node(this->MixNode).MassFlowRate);
+                        Node(this->RelNode).MassFlowRate = max(Node(this->OANode).MassFlowRate - this->ExhMassFlow, 0.0);
+                        // save actual OA flow frac for use as min value for RegulaFalsi call
+                        minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / AirLoopFlow(AirLoopNum).DesSupply);
+                    } else {
+                        Node(this->OANode).MassFlowRate = max(this->ExhMassFlow, OutAirMinFrac * Node(this->MixNode).MassFlowRate);
+                        Node(this->RelNode).MassFlowRate = max(Node(this->OANode).MassFlowRate - this->ExhMassFlow, 0.0);
+                        // save actual OA flow frac for use as min value for RegulaFalsi call
+                        minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / this->MixMassFlow);
+                    }
```
Myoldmopar, line 4586: Then this section calculates the outdoor air using the greater of the exhaust air and current OA, constrained to the total mixed mass flow.  I believe this makes sense but I'm going to think about it for a bit before testing/merging anything.

Your unit test helps build confidence though, so thanks for that.

Myoldmopar (commented):
Withholding judgement for right now while I think things through.

Myoldmopar:
OK, I've run some tests locally.  I was concerned about the diffs in `5ZoneSupRetPlenRAB`, but they are reproducing what @Nigusse was intending as described in the comment [above](https://github.com/NREL/EnergyPlus/pull/6886#issuecomment-410335633).

Here's my results from today on the system's OA node mass flow rate: 
image

I don't have any additional strong feelings about this one.  I get the feeling that @mjwitte was wanting to have some more design discussion back on [August 6](https://github.com/NREL/EnergyPlus/pull/6886#issuecomment-410703705) but I don't think anything really came out of that.  I'm OK holding this one out if it needs more consideration.  It looks like it overall is only causing significant diffs in that one test file, however, so it's not going to cause widespread panic.

mjwitte:
Does the system supply flow change with this fix?  OA shouldn't impact supply flows.

Myoldmopar:
image

src/EnergyPlus/MixedAir.cc
```diff
@@ -4568,10 +4572,18 @@ namespace MixedAir {
...
-                    Node(this->OANode).MassFlowRate = max(this->ExhMassFlow, OutAirMinFrac * Node(this->MixNode).MassFlowRate);
-                    Node(this->RelNode).MassFlowRate = max(Node(this->OANode).MassFlowRate - this->ExhMassFlow, 0.0);
-                    // save actual OA flow frac for use as min value for RegulaFalsi call
-                    minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / this->MixMassFlow);
+                    if (this->FixedMin) {
+                        Node(this->OANode).MassFlowRate =
+                            min(max(this->ExhMassFlow, OutAirMinFrac * AirLoopFlow(AirLoopNum).DesSupply), Node(this->MixNode).MassFlowRate);
+                        Node(this->RelNode).MassFlowRate = max(Node(this->OANode).MassFlowRate - this->ExhMassFlow, 0.0);
+                        // save actual OA flow frac for use as min value for RegulaFalsi call
+                        minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / AirLoopFlow(AirLoopNum).DesSupply);
```
mjwitte: Why is this dividing by DesSupply? If we're shooting for the actual OAFrac here, I understand the line above using DesSupply to apply the fixedmin flow, but then the actual OAfrac will be higher if MixMassFlow < DesSupply.

mjwitte:
And regarding supply flows, sure they could change if the supply air temperature ends up changing.

Myoldmopar:
Unless this gets resolved magically by tomorrow this isn't going to make it.  I'm pushing this to the next milestone.  I'm still open to discussion, but as of right now this won't make it.

src/EnergyPlus/MixedAir.cc
Nigusse: OutAirMinFrac is defined based on "Design Supply" flow as a basis. Can we compare it with a fraction calculated based on a different basis? However, this min OA faction is used as lower bound in the rootsolver and the rootsolver in MultiCompControlTempResidual function uses MixMassFlow. I think it makes sense to change to:

minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / (this-> MixMassFlow);

Nigusse:
I think it makes sense to change this code [here](https://github.com/NREL/EnergyPlus/blob/159513269_Issue%236881/src/EnergyPlus/MixedAir.cc#L4585) to this:


minOAFrac = max(OutAirMinFrac, Node(this->OANode).MassFlowRate / (this-> MixMassFlow);

Nigusse:
The MinOA fraction calculation for rootsolver lower bound did not make difference on the VAV system and OA mass flow rate compared to @Myoldmopar earlier post. Revising the OA fraction lower bound is not expected to make differences, I think. Revised results are shown below:

6881_vav_outletmassflowrate

Nigusse:
This push takes care of the @mjwitte comment about the minOAfrac calculation [here](https://github.com/NREL/EnergyPlus/pull/6886/files/63a1de05307ba1e2776f33caf503c317fa8c3f35#diff-19711df337cdc7fe75d3e87b4c0cd502R4585).

mjwitte (approved):
A little more code review, a little more testing, I'm ok with this.

mjwitte:
@Myoldmopar Merge now, or hold?

Myoldmopar:
> @Myoldmopar Merge now, or hold?

Merge

Issue #6770, opened 2018-06-05
Daylighting:Controls with SplitFlux method only uses the first two reference points


mjwitte:
Issue overview
--------------
If more than two reference points are specified in `Daylighting:Controls` with the SplitFlux method, only the first two reference points have valid illuminance values. The remaining reference points are all zero.

Also, for the test file, many of the Daylighting Window Reference Point *n* View Luminance output variables are reporting values with very large or very small exponents, e.g. e-324, e+73 (see columns DT:EC for example in the variables output).

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.9.0
 - Helpdesk ticket number 12843

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6889

PR #6889, opened 2018-08-06
Remove some dangling limitations to the number of daylighting reference points

JasonGlazer:
While working on a window shading effort, noticed that the number of reference points was still being limited in a few places. These should have all been fixed in #5725.
Addresses #6770.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@JasonGlazer See #6770. Also, I recall seeing some hard-coded variables in the daylighting code for ref points 1 and 2, [here perhaps](https://github.com/NREL/EnergyPlus/blob/2fe2d4ba98d1c41bbb6f65e3d173fd7424e1e1a8/src/EnergyPlus/DaylightingManager.cc#L6331-L6334).

mjwitte:
And here is another place in [daylighting glare control](https://github.com/NREL/EnergyPlus/blob/388f60e37b098ffeb933af1a4f2ae94fb448697f/src/EnergyPlus/DaylightingManager.cc#L6901-L6920) that assumes two reference points.

nrel-bot-3:
@JasonGlazer @lgentile it has been 15 days since this pull request was last updated.

nrel-bot-3:

mjwitte:
@JasonGlazer Ran this with the defect file from #6770 and it fails with an array bounds error [here](https://github.com/NREL/EnergyPlus/blob/40434b6ebb5889085a6f575cf999018c209379e8/src/EnergyPlus/DaylightingManager.cc#L2184). This gets allocated [here](https://github.com/NREL/EnergyPlus/blob/40434b6ebb5889085a6f575cf999018c209379e8/src/EnergyPlus/DaylightingManager.cc#L4479)to `MaxRefPoints` (along with a bunch of others) which is initialized to 2 [here](https://github.com/NREL/EnergyPlus/blob/40434b6ebb5889085a6f575cf999018c209379e8/src/EnergyPlus/DataDaylighting.cc#L84).

JasonGlazer:
@mjwitte I have found quite a few places in the code so far that were hard sizing to two reference points.  I expect some more.

JasonGlazer:
@mjwitte I have implemented all the fixes that I think are necessary except for one that changes output and I posted a new issue for that. I hope the CI comes back clean. Last time I think it was showing issues just because I was not up to date on develop.

Myoldmopar:
This needs to be updated to develop again, it's 9 commits behind.  I'll do it.  Then it should be clean (unless some other commits add diffs, but we'll be able to tell that)

Myoldmopar:
Like right then when I had to pull develop yet again :laughing:

Myoldmopar:
So far so good...Clang reported 100% clean on e7f2bac.  I'll let a few more go in here and this should be ready to go in tonight or tomorrow morning.

mjwitte:
Does the 6770 test file run without crashing?  It's in EnergyPlusDevSupport. That's step 1. Does it produce results for ref point 3?

JasonGlazer:
@mjwitte I reran the 6770 test file this morning which includes all my final code fixes and pulling in develop and it does not crash. One of the two of you should probably replicate that.

mjwitte:
So far, so good. No crash, illuminance values for all reference points, trying some different combinations of lighting fraction.

mjwitte:
But there may be a clear_state problem - when running all of the unit tests in series, getting a crash in `EnergyPlusFixture.HeatBalanceManager_EMSConstructionTest`.

mjwitte:
Daylighting lighting power multiplier looks good with different combinations of reference point fractions.
The unit test crash happens [here](https://github.com/NREL/EnergyPlus/blob/e7f2bac3657c63307e02e196b1a2ec2f880ec49b/src/EnergyPlus/DaylightingManager.cc#L1129-L1130), error is
```
Assertion failed: slice.empty() || range.contains( slice.l(), slice.last() ), file c:\mjw git\energyplus-mjwreview2\third_party\objexxfcl\src\objexxfcl\dimensionslice.hh, line 99
```

mjwitte:
@JasonGlazer If it helps give you a clue, when it gets there `ZoneDaylight(1).DaylIllFacSky` is sized at 1:24, 1:20, 1:4, 1:1, 1:1. Should this section be using `ZoneDaylight().TotalDaylRefPoints` instead of `MaxRefPoints`?  Odd that it fails here, but works OK with the defect file which has zones with more and less reference points.

JasonGlazer:
@mjwitte I was tempted to remove all references to MaxRefPoints but they are used in the illuminance map generation in a way I have not studied so I'm leaving those alone.

mjwitte:
OK, perhaps try adding maps to every zone in the defect test file and see if it runs OK.

mjwitte (approved):
Confirmed that test files still produce good results (same as before the last commit to fix the unit test failure).

Confirmed that all unit tests run serially now.

KhanhEner:
@JasonGlazer Unfortunately, after removing the limitation, daylight calculation in case where there are tubular daylighting devices (DaylightingDevice:Tubular) became incorrect. Perhaps DaylightingDevice were not considered during the tests?
image

JasonGlazer:
I'm sorry it is not working for DaylightDevice:Tubular. You probably want to post this as a new issue so it gets some attention. Go to:

https://github.com/NREL/EnergyPlus/issues

And click on "new issue"

Please reference this pull request when you post the issue.

Issue #6864, opened 2018-07-27
ASHRAE Revised Clear Sky "tau" Model


mjroth:
Issue overview
--------------
The ASHRAE "tau model" uses the original but obsolete coefficients for the air mass exponents from HOF 2009.  The last two updates (HOF 2013 and HOF 2017) have published improved coefficients.

The pertinent lines in WeatherManager.cc for ASHRAETauModel:
```
AB = 1.219 - 0.043 * TauB - 0.151 * TauD - 0.204 * TauB * TauD;
AD = 0.202 + 0.852 * TauB - 0.007 * TauD - 0.357 * TauB * TauD;
```

Should read:
```			
AB = 1.454 - 0.406 * TauB - 0.268 * TauD + 0.021 * TauB * TauD;
AD = 0.507 + 0.205 * TauB - 0.080 * TauD - 0.190 * TauB * TauD;
```

### Details
Reference:  ASHRAE Handbook - Fundamentals, 2017, Chapter 14, pg. 14.10, eqns (19) & (20)

mjwitte:
True, but the coefficients for the air mass exponents are not the only thing that changed.  All of the location-specific data for TauB and TauD also changed. For example, from handbook Ch14, Table 1 for Atlanta, GA, here are taub, taud, and the results of an example calulation for Atlanta:
2009 HOF July taub=0.556, taud=1.779, Eb=751 W/m2, Ed=220 (at solar noon) 737/215 at 3pm EDT
2013 HOF July taub=0.440, taud=2.202, Eb=846, Ed=144 (at solar noon) 834/141 at 3pm EDT
2017 HOF July taub=0.515, taud=2.066, Eb=784, Ed=166 (at solar noon) 770/162 at 3pm EDT

The EnergyPlus ddy files which are available on [energyplus.net](https://energyplus.net/weather) use the 2009 values for taub and taud, at least that's true for [Atlanta TMY3](https://energyplus.net/weather-location/north_and_central_america_wmo_region_4/USA/GA/USA_GA_Atlanta-Hartsfield-Jackson.Intl.AP.722190_TMY3).
```
 ! Atlanta Hartsfield Intl Ap Annual Cooling Design Conditions Wind Speed=4m/s Wind Dir=300
 ! Hottest Month=JUL
 ! Atlanta Hartsfield Intl Ap_GA_USA Annual Cooling (DB=>MWB) .4%, MaxDB=34.4°C MWB=23.5°C
[...]

PR #6891, opened 2018-08-07
Updated ASHRAE Revised Clear Sky "tau" Model with new 2017 option

Nigusse:
Addresses Issue #6864

Pull request overview
---------------------
This fix added new ASHRAE's Revised Clear Sky "TauModel" solar model indicator key choice based on updated coefficients from ASHRAE 2017 Handbook of Fundamentals. User can run ASHRAE's Revised Clear Sky "TauModel" solar model using  taub and taud data from the 2009, 2013, or 2017 ASHRAE HOF. Diffs are not expected for this change. 

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


Nigusse:
19 test files shown diffs in regression tests are expected since those test files have "ASHRAETAU" as their solar indicator model.

lgu1234:
@Nigusse I would like to make a suggestion to give users a choice to select different sets of coefficients. In other words, the old coefficients may be kept in some way. For example, we may add a new object to store coefficients. The suggestion may not be for this fix. Instead, it may be a new feature development

Nigusse:
The unit test "PackagedTerminalHP_VSCoils_Sizing" fails but the cause for this unit test failure is not related to this fix. However, the 19 test files that show diffs are caused by the ASHRAETau Model coefficients update and is expected.

Nigusse:
@mjwitte @Myoldmopar the ASHRAETau Model coefficients are hard coded. Reproducing previous results is not possible when we change the hard coded coefficients. Is this is a concern? @lgu1234 is suggesting to consider adding the a objects that EnergyPlus can read the coefficients from an object instead of hard coding.

mjwitte:
@lgu1234 @Nigusse @Myoldmopar Not only have the coefficients changed from 2009 to 2013/2017, the user inputs for taub and taud have changed from 2009 to 2013 and again from 2013 to 2017. Given that, users will need the option to select ASHRAETau2009 or ASHRAETau20132017 or something like that. But since ASHRAETau is embedded in many (all?) of the ddy files, we may want to go with ASHRAETau as the 2009 correlation, and ASHRAETau201317 as the new one?  Marking this as IDD change.

mjwitte:
Or maybe just ASHRAETau2017 and forget that 2013 ever happened.

Nigusse:
@mjwitte I support to add ASHRAETau2017 only since the coefficients ab and ad remain the same for 2013 and 2017 HOF versions. But needs to add explanation in the I/O ref documentation that ASHRAETau2017 can be used for 2013 HOF with matching tauB and tauD coefficients.

Nigusse:
I will start working on this branch to accommodate these changes. So, I will add new key choice ASHRAETau2017.

Nigusse:
Now the DesignDay Sizing object model allows to select solar model indicator ASHRAETau or ASHRAETau2017 keys. ASHRAETau is used with the 2009 ASHRAE HOF and ASHRAETau can be used with either the 2013 or 2017 ASHRAE HOF data. No diffs are expected.

Nigusse:
Also updated I/O and Eng reference documents and Idd as needed.

mjwitte (approved):
Code, idd, and doc changes all look OK.  Made a set of test files using 2009, 2013, and 2017 options for Atlanta, July 21 to reproduce the example results in the handbooks. Solar noon and 3pm EDT all agree.

Issue #6897, opened 2018-08-08
EnergyPlus fails clang's "undefined" sanitizer


kbenne:
If you compile EnergyPlus using clang and enable the undefined behavior sanitizer, the program will fail with the following information.

The is true on Mac OS X, but likely on other systems.

```
EnergyPlus, Version 9.0.0-388f60e37b, YMD=2018.08.08 10:31
Process 55870 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = Pointer overflow
    frame #0: 0x000000012365de00 libclang_rt.ubsan_osx_dynamic.dylib`__ubsan_on_report
libclang_rt.ubsan_osx_dynamic.dylib`__ubsan_on_report:
->  0x12365de00 <+0>: pushq  %rbp
    0x12365de01 <+1>: movq   %rsp, %rbp
    0x12365de04 <+4>: popq   %rbp
    0x12365de05 <+5>: retq
Target 0: (energyplus) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = Pointer overflow
  * frame #0: 0x000000012365de00 libclang_rt.ubsan_osx_dynamic.dylib`__ubsan_on_report
    frame #1: 0x000000012365875d libclang_rt.ubsan_osx_dynamic.dylib`__ubsan::Diag::~Diag() + 157
    frame #2: 0x000000012365d8af libclang_rt.ubsan_osx_dynamic.dylib`handlePointerOverflowImpl(__ubsan::PointerOverflowData*, unsigned long, unsigned long, __ubsan::ReportOptions) + 1199
    frame #3: 0x000000012365d3f4 libclang_rt.ubsan_osx_dynamic.dylib`__ubsan_handle_pointer_overflow + 68
    frame #4: 0x000000010390767f libenergyplusapi.9.0.0.dylib`ObjexxFCL::Array<double>::shift_set(this=0x00007fff5fbe63a8, shift=1) at Array.hh:2559
    frame #5: 0x000000010667fa2f libenergyplusapi.9.0.0.dylib`ObjexxFCL::Array1<double>::clear_move(this=0x00007fff5fbe63a8) at Array1.hh:2808
    frame #6: 0x000000010667f281 libenergyplusapi.9.0.0.dylib`ObjexxFCL::Array1<double>::Array1(this=0x0000000127a5b790, a=0x00007fff5fbe63a8) at Array1.hh:122
[...]

PR #6902, opened 2018-08-12
ObjexxFCL 4.3 Upgrade

DeadParrot:
This updates ObjexxFCL to version 4.3, which has some minor refinements:
* Visual C++ /W4 warning level compatibility
* Undefined sanitizer compatibility (#6897)
* Multithreading support _via_ defining OBJEXXFCL_THREADS (@kbenne)
* Stream wrapper protected cin/cout/cerr handling (#6885)

No EnergyPlus code changes were made and no solution changes are expected.

**Update** This appears to incur a surprising performance penalty on the GCC 4.8 Ubuntu platform. My testing at the loop and application levels did not show this. Investigating should:
* See if slowdown is reproduced on other platforms with newer (better optimizing) compilers
* See if profiling can identify some key loops where linear indexing can recover the performance
* If a penalty remains team must decide if they prefer speed or making the undefined sanitizer happy

I cannot do this work for a while so let's put this PR on hold.


mjwitte:
Re-running the failed linux tests, may show some diffs due to recent change in develop.

mjwitte:
@DeadParrot This is showing a signficant [performance](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-3282e0eab840b406eb3d7da2434682a98ff4c6b6-x86_64-Linux-Ubuntu-14.04-gcc-4.8.html) decline.

DeadParrot:
@mjwitte A 6% slowdown is very unlikely for the Array change but I don't run an ancient compiler like GCC 4.8 (why are we doing performance runs on such an old compiler?). And if we insist on being undefined sanitizer clean this change is needed.

I'll do a merge+submit tonight and we'll see what the failures and performance look like after that. This may be more related to the many new commits than anything in ObjexxFCL.

mjwitte:
@DeadParrot Still showing a 6.6% slowdown. Use any of the [active performance test files](https://github.com/NREL/EnergyPlus/blob/develop/performance_tests/CMakeLists.txt) to confirm locally. 15zonePSZ.idf should work.
@Myoldmopar Would you be ok with moving the inactive performance test idfs to an Inactive subfolder ?

DeadParrot:
@mjwitte Please see update that was added to the summary.

Myoldmopar:
> @Myoldmopar Would you be ok with moving the inactive performance test idfs to an Inactive subfolder ?

Fine with me, there should be updates to the performance test suite in the coming months.

mjwitte:
With this on hold, does that block #6885?

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

Myoldmopar:
@DeadParrot after we released 9.0, we overhauled our Linux CI box and got Ubuntu 18.04 with GCC 7.3 set up and running.  We now have that setup in our develop branch and we are starting to pull that into the other open branches.  I did that to your ObjexxFCL 4.3 branch here, hoping to see a performance gain, but it is still showing a performance degradation in this branch compared to develop.  Now that we are seeing this on a more up-to-date compiler toolchain, could you provide some commentary or investigate?  It would be great to get this merged, but we'll need to understand/address the performance implications of doing so.

Myoldmopar:
(https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-282e421b7a41c08281df1ebb290ce0fe70f23947-x86_64-Linux-Ubuntu-18.04-gcc-7.3.html)

DeadParrot:
@Myoldmopar thanks for testing this. The performance hit is surprising -- I got ~1% in an array-heavy application. I don't see a way to get the pointer offset speedup without using technically undefined behavior that triggers undefined sanitizer warnings. My preference (and I think that of @kbenne) would be to eliminate undefined behavior and then use linear indexing in hot spots to recover the performance. I don't really want to further fork the ObjexxFCL into "fast" and "proper" versions or to only enable the offset pointer hack in non-debug builds.

The way forward is your call. Some profiling to see where the performance is going would be a good idea. And I can review linear indexing transformations with whoever does the tuning.

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@DeadParrot @lgentile it has been 14 days since this pull request was last updated.

Myoldmopar:
Since this PR is on hold, I'm going to close it but we'll leave the branch in place so we can open it back up later.

DeadParrot:
@Myoldmopar I have a new ObjexxFCL 4.3 that preserves the original faster lookups and a compile-time macro to enable the (slower) undefined sanitizer clean code. This update also has some C++11 modernizations. I can provide it upon request.

Issue #6913, opened 2018-08-17
No example files currently use MeetDaylightIlluminanceSetpoint window shading control type


mjwitte:
Issue overview
--------------
In order to confirm the changes in #6421 the combination of SwitchableGlazing with MeetDaylightIlluminanceSetpoint is important to have in the test files, because this is supposed to adjust the window darkness as a group instead of sequentially. All of the existing test files that currently have SwitchableGlazing use OnIfHighGlare and only apply this to a single window. Suggest adding a new zone to WindowTests which already tests a variety of shading types applied to multiple windows in a zone.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6914

PR #6914, opened 2018-08-17
Extend WindowTests example to include switchable glazing with illuminance control, and other minor test changes

mjwitte:
Pull request overview
---------------------
1. Address #6913 by adding another zone to the WindowTests example file which has shading type SwitchableGlazing with shading control type MeetDaylightIlluminanceSetpoint. Added daylighting controls to existing zone DPW (double pane no shading) and new zone DPSG (double pane switchable glazing) plus additional output variables to confirm operation.

2. Moved all of the inactive performance test inputs to an "Inactive" subfolder.

3. Added to annual simulations to the integration/regression tests.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@Myoldmopar This is ready for review.

mjwitte:
@Myoldmopar Would you be open to adding one or two annual simulations to the testfiles? e.g. 5ZoneAirCooled. And why are some of the 5ZoneAirCooled files being [run with CO weather](https://github.com/NREL/EnergyPlus/blob/d79a0f2f2cd90a2d57a44b3b3ceefbc05d87a0f3/testfiles/CMakeLists.txt#L29-L31) instead of Chicago?

Myoldmopar (approved):
Changes look reasonable.  Built locally and ran the modified test file.  In design-day-only, the only warning was the tabular outputs for a non-annual run.  In annual run, the only warning was that I used Chicago weather instead of Denver.  This is good to go.  Merging.

Myoldmopar:
Let me rephrase myself a bit.  This is all good to go, but feel free to add a couple annual run cases to CI here before I merge.

Myoldmopar:
I'd suggest renaming those files.  The annual version is probably going to overwrite the first one.  I'd just copy the file with _annual appended to the name -- even though there's no actual difference in the idf.  Then change the CMakeLists accordingly and there won't be any concern about overwritten results.

mjwitte:
Should I do those with _ prefix so then they don't get duplicated in the install?

Myoldmopar:
Excellent idea!

Issue #6912, opened 2018-08-17
AirflowNetworkAdvanced_SingleSided_NV crashes after AFN multi-airloop changes


mjwitte:
Issue overview
--------------
Crashing [here](https://github.com/NREL/EnergyPlus/blob/d79a0f2f2cd90a2d57a44b3b3ceefbc05d87a0f3/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L476) due to uninitialized `AFNSupplyFanType`. Looks like some of this code needs to be inside the airloop for loop?

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus 9.0.0-d79a0f2f2c
 - Ran into this while testing #6858 

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #6915

PR #6915, opened 2018-08-17
Avoid crash of AirflowNetworkAdvanced_SingleSided_NV

lgu1234:
Pull request overview
---------------------
A local variable of AFNSupplyFanType will be checked with one more condition in ManageAirflowNetworkBalance. Addresses #6912.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -469,7 +469,7 @@ namespace AirflowNetworkBalanceManager {
         if (allocated(ZoneEquipConfig) && NumHybridVentSysAvailMgrs > 0 && allocated(PrimaryAirSystem)) HybridVentilationControl();
         if (VentilationCtrl == 1 && NumHybridVentSysAvailMgrs > 0) AirflowNetworkFanActivated = false;
 
-        if (present(Iter) && present(ResimulateAirZone)) {
+        if (present(Iter) && present(ResimulateAirZone) && SimulateAirflowNetwork >= AirflowNetworkControlSimpleADS) {
```
mjwitte, line 472: @lgu1234 This avoids the crash, but this section of code still seems wrong for multiple airloops. It seems that AFNSupplyFanType will always be zero unless FirstHVACIteration is true, and even then it will be the type for the last airloop only.

lgu1234: @mjwitte When FirstHVACIteration is true, it is the current time step. The function is called in SimSelectedEquipment of HVACManager. In addition, AFNSupplyFanType is set by FanTypeNum and we don't change fan type in simulations. We only allow fan mode change for onoff fans. This is taken care of the LoopFanOperationMode. The main purpose is to catch if there is a FanType_SimpleOnOff or not in multiple air loops.

mjwitte: OK, I'll trust you on this.

PR #6918, opened 2018-08-20
Fix infinite parsing loop caused by stray semicolon

### LINKED ISSUE #6916: Stray semi-colon cause infinite loop in IdfParser
Issue overview
--------------
User file has a stray semi-colon by itself between complete objects. This results in an infinite loop [here](https://github.com/NREL/EnergyPlus/blob/d79a0f2f2cd90a2d57a44b3b3ceefbc05d87a0f3/src/EnergyPlus/InputProcessing/IdfParser.cc#L202-L220) in IdfParse that just keeps chugging away eating more and more memory as it goes. Reproduced with a smaller example file.

[1ZoneUncontrolled-StraySemiColon.idf.txt](https://github.com/NREL/EnergyPlus/files/2300686/1ZoneUncontrolled-StraySemiColon.idf.txt)

### Details
 - Win64
 - Version of EnergyPlus Version 9.0.0-d79a0f2f2c
 - Sort of related to helpdesk ticket 11958

### Checklist

mbadams5:
Pull request overview
---------------------
Fixes #6916

This fixes an infinite parsing loop caused by a stray semicolon in the input file. It will also now catch stray comma's as well, throwing an error. The stray semicolon does not throw an error and instead just continues parsing.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


.gitignore
```diff
@@ -35,7 +35,7 @@ Build
 Builds
 
 # by default CLion uses cmake-build-debug as the build folder, so add that as well
-cmake-build-debug
+cmake-build-*
```
Myoldmopar, line 38: :+1:

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -261,8 +261,7 @@ TEST_F(InputProcessorFixture, decode_encode_2)
                                           "  Counterclockwise,",
                                           "  Relative,",
                                           "  Relative,",
-                                          "  Relative;"
-                                          "",
+                                          "  Relative;",
```
Myoldmopar, line 264: This unit test doesn't seem to cover the proposed fix here.  I'll modify a test file to verify the fix, but should we also create a unit test for it here?

mbadams5: It doesn't. I just cleaned up this unit test.

I tried to think of a way to unit test this but wasn't successful in my fairly quick attempt to fix this. The problem is that google test doesn't seem to have timeout functionality so how do you test for something that causes infinite looping?

There is a defect file for this issue though.

Myoldmopar:
Ahh, that makes good sense @mbadams5.  I went ahead and tested locally.

#### Develop Branch
 - An unmodified 1ZoneUncontrolled.idf file of course runs fine.  
 - I modified it with a stray comma, and it gives `** Severe  ** Line: 92 Index: 0 - "" is not a valid Object Type.`
 - I modified it with a stray semicolon instead, and it runs for 65 seconds on my machine, essentially stuck in an infinite loop, until finally failing.  I'm guessing it's hitting a stack overflow if it is a recursive problem.  Anyway, it eventually aborted without a helpful message.

#### This Branch
 - Unmodified file passes fine
 - Stray comma file fails with ` ** Severe  ** Line: 91 Index: 13 - Extraneous comma found.`
 - Stray semicolon file passes without issue.

This is good to go.  Merging.

PR #6948, opened 2018-09-07
Fix VRF curve dimension check and clarify xy variables for all DX heat pump defrost EIR curves

mjwitte:
Pull request overview
---------------------
This fixes the curve dimension check for AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl:HR "Defrost Energy Input Ratio Modifier Function of Temperature Curve Name". For some reason, the v8.9 IDD specified UnivariateCurves for this curve in both of the AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl* objects. The source code correctly validates the non-HR curve for 2 dimensions, but the other was checking for 1 and causes a fatal error.

Two occurrences of \object-list UnivariateFunction in the idd need fixing (deferred to #6951 for after v9.0 release).

All defrost EIR function of temperature curves are evaluated with coil entering wet bulb as the x variable, and outdoor condenser entering dry bulb as the y variable.  Search the source code for `DefrostEIR` and every `CurveValue` function call is similar to this:
```
DefrostEIRTempModFac = CurveValue(DXCoil(DXCoilNum).DefrostEIRFT, max(15.555, InletAirWetBulbC), max(15.555, OutdoorDryBulb));
```

The equations in the Engineering Reference are consistent with this, but the curve descriptions in the I/O ref all said something like this curve is function of outdoor dry bulb and entering wet bulb which is confusing.  All of these curve descriptions have been changed to clarify the order of the variables.

This was discovered when moving the defect file for #6950 to v9.0. Use that defect file to test this.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar (changes requested):
I think we're too far past the IO freeze to let the IDD change slip in.  Please take those changes back out (and drop them in a separate PR whenever you like for after the release), and this fix can go in.

Myoldmopar (approved):

PR #6965, opened 2018-09-14
Update EPLaunch workflow files to Beta2 API

Myoldmopar:
Pull request overview
---------------------
The API for EPLaunch workflows changed slightly, this updates the workflows to the new version.


Myoldmopar:
I won't be waiting on CI to finish on all the platforms.  There is a bit of CMake configuration going on, so I do want it to test it at least once before I merge down to develop, but no reason to waste cycles on all the others.  I'll merge when Clang is done here.

Myoldmopar:
I've tested this multiple times, and all I'm doing is updating the EPLaunch workflow files, nothing that could cause diffs.  I'm going to merge this.

PR #6968, opened 2018-09-17
Fix bug in input parser

mbadams5:
See https://unmethours.com/question/34222/error-parsing-schedulefile-error-in-following-line-in-e-89/

Strings that had \r, \n, and other special characters were being parsed as true carriage return, newline, etc. This was causing issues with file paths on windows where they could legitimately have \r or \n in the file path.

Pull request overview
---------------------

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/InputProcessing/IdfParser.cc
+++ b/src/EnergyPlus/InputProcessing/IdfParser.cc
@@ -572,30 +572,6 @@ std::string IdfParser::parse_string(std::string const &idf, size_t &index, bool 
             complete = true;
             decrement_both_index(index, index_into_cur_line);
             break;
-        } else if (c == '\\') {
-            if (index == idf.size()) break;
-            char next_c = idf[index];
-            increment_both_index(index, index_into_cur_line);
-            if (next_c == '"') {
-                s += '"';
-            } else if (next_c == '\\') {
-                s += '\\';
-            } else if (next_c == '/') {
-                s += '/';
-            } else if (next_c == 'b') {
-                s += '\b';
-            } else if (next_c == 't') {
-                s += '\t';
-            } else if (next_c == 'n') {
-                complete = false;
-                break;
-            } else if (next_c == 'r') {
-                complete = false;
-                break;
-            } else {
-                s += c;
-                s += next_c;
-            }
         } else {
             s += c;
         }
--- a/tst/EnergyPlus/unit/InputProcessor.unit.cc
+++ b/tst/EnergyPlus/unit/InputProcessor.unit.cc
@@ -284,6 +284,57 @@ TEST_F(InputProcessorFixture, decode_encode_2)
     EXPECT_EQ(expected, encoded);
 }
 
+TEST_F(InputProcessorFixture, decode_encode_3)
+{
+    auto const idf(delimited_string({
+      "Schedule:File,",
+      "  Test Schedule File,      !- Name",
+      "  Any Number,              !- Schedule Type Limits Name",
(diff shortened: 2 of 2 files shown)


tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -1407,16 +1458,16 @@ TEST_F(InputProcessorFixture, parse_string)
     index = 0;
     success = true;
     output_string = parse_string(R"(\b\t/\\\";)", index, success);
-    EXPECT_EQ("\b\t/\\\"", output_string);
+    EXPECT_EQ(R"(\b\t/\\\")", output_string);
```
Myoldmopar, line 1461: > `R"(\b\t/\\\")"`

image

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -572,30 +572,6 @@ std::string IdfParser::parse_string(std::string const &idf, size_t &index, bool
             complete = true;
             decrement_both_index(index, index_into_cur_line);
             break;
-        } else if (c == '\\') {
```
Myoldmopar, line 575: So really it's just removing this code to fix it?  That's a nice thing...

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -284,6 +284,57 @@ TEST_F(InputProcessorFixture, decode_encode_2)
...
+      "  8760.0,",
+      "  Comma,",
+      "  ,",
+      "  10.0;",
+      ""
+    }));
+
+    ASSERT_TRUE(process_idf(idf));
+    std::string encoded = encodeIDF();
+    EXPECT_EQ(expected, encoded);
```
Myoldmopar, line 335: Ahhh this was confusing for a minute.  I forgot the test fixture is going to add the building and global geometry objects automatically.  Ok, ok.

Myoldmopar (approved):

Myoldmopar:
CI tests look OK, code is interesting.  @mbadams5 could you describe why removing that block of code fixes it?  I mean I can imagine the parser simply moves into a different block and leaves `\r` alone, but can you imagine any consequences?  I feel like you added that code on purpose when you wrote the parser originally.

mbadams5:
@Myoldmopar 

So what was happening is that with that section of code, when a string was parsed and it finds a `\` then it looks to see if it is an escaped special character (`\"`, `\r`, `\n`, etc). If it finds one of these special characters it then turns it into that special character, meaning instead of 2 `char` (`\` and `t`) it becomes `\t`, the tab character. This kind of functionality is important when parsing formats such as JSON where you can have a string with `\n` in it and want the string when parsed to have a newline but NOT the JSON file itself. That said, I don't think this is appropriate when parsing IDF and is in fact causing this bug since IDF doesn't have explicit string delimited fields, `""`. This portion of the code was adapted from some prior work and shouldn't have been added in the first place for IDF parsing.

Hopefully this explanation was clear as mud.

Myoldmopar:
That's a perfect explanation.  I am good with the code changes here.  I'll try to play around with it to verify for a few minutes this morning and if it's all good I'll go ahead and merge.

Myoldmopar:
Tested this in develop by putting in some escape characters in a test file.  It fails with an explanatory error message.  Running this in this branch allows it to run just fine.  Merging.  Thanks @mbadams5.

Issue #6793, opened 2018-06-20
Use of Schedule:File suppresses progress messages to the console window


mjwitte:
Issue overview
--------------
Any idf that uses Schedule:File runs almost silently with no progress messages to the console window. Happened to notice this with the defect file for #6792 but didn't understand why. until sharing the file with @Myoldmopar but forgot to send the schedule csv file.

Example file 5ZoneTDV shows the problem.  Here's a screen shot on Win64 using v8.9.0.

![image](https://user-images.githubusercontent.com/4991075/41674376-6cad5994-7485-11e8-8919-a61e4188c22d.png)

And here's a slightly different result using HybridZoneModel:

![image](https://user-images.githubusercontent.com/4991075/41674560-03069798-7486-11e8-894c-26a4f3cf02f8.png)

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v8.9.0

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file - use 5ZoneTDV and HybridZoneModel example files.
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
And it did work OK for me on Linux (I see the normal standard output.)

I wonder if some IO setting is being set when the program attempts to read the csv file on Windows and it muffles all standard output from that point on.

Myoldmopar:
Closed via #6975

PR #6975, opened 2018-09-20
Fix console messages on Windows with Schedule:File or other extra files

mjwitte:
Pull request overview
---------------------
Addresses #6793. Added a unit test, but the test doesn't fail when run without this fix. So, not sure if we want to keep it there. Run 5ZoneTDV as a test on Windows to confirm that progress messages now appear as normal.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


mjwitte:
@Myoldmopar This should be ready for review. Not sure why [this failure](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-673f073e931acb3bb91544e9ba58c7ada01f6d6d-x86_64-MacOS-10.9-clang.html) happened on Mac. I've merged in develop locally and this test runs OK.

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -1007,6 +1007,9 @@ int FindUnitNumber(std::string const &FileName) // File name to be searched.
         std::string::size_type TestFileLength;
         std::string::size_type Pos; // Position pointer
         for (UnitNumber = 1; UnitNumber <= MaxUnitNumber; ++UnitNumber) {
+            // Skip preassigned units - gio::inquire breaks std::cout on Windows - see GlobalStreams::
+            if ((UnitNumber == 0) || (UnitNumber == 5) || (UnitNumber == 6) || (UnitNumber == 100) || (UnitNumber == 101) || (UnitNumber == 102))
+                continue;
```
jasondegraw: Could we enclose this line in braces so it is clear that the `continue` goes with the `if` and not the block below?

jasondegraw (changes requested):
Had one small nitpick, but this fixes the issue for me on Windows.

src/EnergyPlus/UtilityRoutines.cc
mjwitte: OK - added braces and updated comment which was incomplete before. Ready again.

Myoldmopar:
@mjwitte that test failure happens intermittently.  It's something missing in a clear_state, but I haven't had a chance to dig through and see what isn't being cleaned up.  

I'm going to go ahead and merge this now, I don't want to waste CI cycles on it.  

@jasondegraw thanks for reviewing this!

PR #6980, opened 2018-09-21
Fix doc issues

Myoldmopar:
Some documentation issues slipped in.  CI was able to get passed them, but if you built the engineering reference in, say, TeXMaker, you would see lots of errors.  This addresses that.

### DIFF
--- a/doc/engineering-reference/src/alternative-modeling-processes/room-air-models.tex
+++ b/doc/engineering-reference/src/alternative-modeling-processes/room-air-models.tex
@@ -117,7 +117,7 @@ The values for \({Z_i}\) are constrained to be between 0.01 and 0.99 because the
 
 The user-defined profile is treated as a look up table or piecewise linear model.~ The values for \(\Delta {T_{ai}}\) are determined by searching the \({\rm Z}\) values in the user-defined profile and performing linear interpolation on the associated \(\Delta {T_a}\) values.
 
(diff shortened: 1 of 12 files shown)


mjwitte:
Patience is @Myoldmopar 's middle name, as patient as his mopar engine.

jasondegraw:
@Myoldmopar Which bot are we waiting for?

Myoldmopar:
I would love to see Gort report back that he built the docs successfully.  But we don't have to.  Let me do a full fresh build of this branch and all docs and package.  If it builds OK, we can merge this and get it in the other branches.

jasondegraw:
@Myoldmopar I highly doubt that will happen, it's currently hanging on the doc build. I'm building develop on the VM to see if that also hangs, and there is a long running xelatex process, so we'll see.

Myoldmopar:
I just built a full package on Linux.  Everything seems super clean.  I'm going to merge this down and it should be safe for other branches to pick it up as needed.  It looks like CI is churning now, however, so it may not be important for other branches to integrate these fixes anymore.

Myoldmopar:
Sorry, I missed your comment there @jasondegraw.  I had been able to reproduce a delay in building on CI, it seemed stuck.  I just pulled a fresh develop on that machine, configured, and made the engineering reference in about 40 seconds including all 3 passes of LaTeX.  We may have a couple issues here and there where CI takes a while, but I think once branches get this in place, things should improve.

I'll continue to monitor my CI box and if I see any very long running XeLaTeX processes I'll kill them.  The builds will have errors but at least they can move forward to better commits.

Myoldmopar:
So actually, on that note @jasondegraw, go ahead and kill the latex process you are seeing.  The CI box will be able to move to the next commit that way!

jasondegraw:
@Myoldmopar Actually, I'm not sure that it will based on what I was seeing. If I have to reboot, I'll let you know.

Issue #6995, opened 2018-10-04
Mismatch between Sides and size of Vertex - fatal error in v9.0


mjwitte:
Issue overview
--------------
User file which runs in v8.9, fails in v9.0 with errors like this:
```
   ** Warning ** GetVertices: Distance between two vertices < .01, possibly coincident. for Surface=SURFACE 8536, in Zone=THERMAL ZONE 142
   **   ~~~   ** Vertex [12]=(-1.94,64.72,21.40)
   **   ~~~   ** Vertex [1]=(-1.94,64.72,21.40)
   **   ~~~   ** Dropping Vertex [12].
   ** Severe  ** BuildingSurface:Detailed="SURFACE 8536", After CheckConvexity, mismatch between Sides (11) and size of Vertex (12).
```
Looks like the vertex size is not getting reduced when a coincident vertex is dropped. Likely related to #6853.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus 9.0.0-2ef880da82
 - Helpdesk ticket number 13067
### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added to EnergyPlusDevSupport\DefectFiles - *however, the user files are enormous (approx 600 zones), so you can't even see the v8.9 file finish successfully.  But the 9.0 file fails quickly. Working on a smaller file to reproduce the same issue.*
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
The plot thickens. 
1. Made a smaller version of the original file with just one zone that includes one of the failing surfaces. Ran fine in both v8.9 and v9.0.
2. Noted that the failing surface was originally an interzone surface, so added a second zone and the paired surface - now it runs in v8.9 and fails in v9.0.
Smaller defect file attached here.


[6995-SmallerDefectFile.idf.txt](https://github.com/NREL/EnergyPlus/files/2448410/6995-SmallerDefectFile.idf.txt)
[6995-SmallerDefectFile_V890.idf.txt](https://github.com/NREL/EnergyPlus/files/2448411/6995-SmallerDefectFile_V890.idf.txt)
[6995-SmallerDefectFile-SingleZone.idf.txt](https://github.com/NREL/EnergyPlus/files/2448430/6995-SmallerDefectFile-SingleZone.idf.txt)

[...]

PR #6853, opened 2018-07-23
Cleanup surface vertex points and reset vertex size 

lgu1234:
Pull request overview
---------------------
Address #6834.
1. Remove duplicated points of surface vertexes
2. Check colinearity
3. If colinearity, remove extra points
4. Resize vertex vectors
5. Provide a warning to show vertex reprocessing

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Code style (parentheses padding, variable names)
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check -- verify this
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add rules to spreadsheet
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Documentation changes in place
 - [ ] Changed docs build successfully
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #6996, opened 2018-10-05
Fix Vertex array size when a coincident vertex is dropped

mjwitte:
Pull request overview
---------------------
Address #6995. It's not clear to me that this problem was triggered by the changes in #6853, because the defect file does not seem to hit that section of the code.

The defect file is large and takes about 2hrs to run on my machine, so I've put zips of outputs in EnergyPlusDevSupport\DefectFiles.
Note that the 6995-SmallerDefectFile.idf runs quickly and exhibits the symptom, but it did not cover all the bases. An initial fix allowed 6995-SmallerDefectFile to run, but the user file failed in other places. So, the fix had to be expanded to two other places to get the user file to run to completion.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/SurfaceGeometry.cc
+++ b/src/EnergyPlus/SurfaceGeometry.cc
@@ -7270,6 +7270,7 @@ namespace SurfaceGeometry {
                         ShowContinueError("Dropping Vertex [" + RoundSigDigits(SurfaceTmp(SurfNum).Sides) + "].");
                     }
                     --SurfaceTmp(SurfNum).Sides;
+                    SurfaceTmp(SurfNum).Vertex.redimension(SurfaceTmp(SurfNum).Sides);
                 } else {
                     if (DisplayExtraWarnings) {
                         ShowContinueError("Cannot Drop Vertex [" + RoundSigDigits(SurfaceTmp(SurfNum).Sides) +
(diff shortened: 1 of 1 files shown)


mjwitte:
@lgu1234 @Myoldmopar This is ready for initial review. See comments at top regarding defect files. I *had* a full output for the user file and then overwrote it, so it's running again now. I"ll post the zip of outputs when it's done (takes about 2 hrs).

mjwitte:
I'll think about a unit test for this.

Myoldmopar:
Thanks for the quick fix on this @mjwitte; I'll watch for the outputs.  It looks like we might have a couple of issues that we should fix and create a 9.0.1 release.  Oh, and Gort has really been struggling.  Don't sweat his unhandled exceptions.  I'm able to build successfully on my CI instance.  And I am in the process of upgrading the CI system anyway.

lgu1234:
@mjwitte I tested 6995-SmallerDefectFile.idf with debug on my local computer and worked as expected.

I have a concern for future. We use 0.01 as a distance check between two points. It may not be user's intent. We may allow an input of distance limit and use 0.01 as default.

lgu1234:
@mjwitte I also tested 6995-SmallerDefectFile-SingleZone.idf. It worked as expected.

mjwitte:
@Myoldmopar @lgu1234 Just added a zip of outputs for the user file with v9.0 after this fix.  So, you can compare v8.9 vs v9.0 before fix (fatal) vs v9.0 after fix.

lgu1234:
@mjwitte @Myoldmopar I added one more line in Revised_V890.idf: Output:Diagnostics, DisplayExtraWarnings; and ran it again. Uploaded is an err file as Revised_V890 WithDisplayExtra.err. Then I compare both files of Revised_V890 WithDisplayExtra.err and Revised-FixSetpoinsAfterFix.err. They are very compatible.

Myoldmopar:
This looks great.  Error file from 8.9 to after fix looks similar.  Compared some outputs in the csv data and they match up well.  CI is generally happy except for the Ubuntu issue (which I have a fix for, w00t!).  I'm inclined to merge this in so we can try to re-release today.  Any issues with that @mjwitte ?

PR #7033, opened 2018-10-25
Fix #6442 Equipment Summary has incorrect conversions from SI to IP

### LINKED ISSUE #6442: Equipment Summary has incorrect conversions from SI to IP
Issue overview
--------------

On the Equipment Summary report, when the standard HTML output report is reported in IP units the values on DX Cooling Coils and VAV DX Cooling Standard Rating Details are incorrect. The header suggests tons, but the magnitude is too large for it to be tons - the Cooling Coils section reports as btuh so I'm guessing it should be that.

I'm guessing the DX Cooling Coils and VAV DX Cooling Standard Rating Details are reporting as W because the difference between them and Cooling Coils is just over 3.412 (W to btuh conversion). Being a little over makes sense as DX Cooling Coils and VAV DX Cooling Standard Rating Details I believe include fan heat.

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): Energyplus 9.0.1

### Checklist
 - [x] Defect file added : `testfiles/1ZoneDataCenterCRAC_wApproachTemp.idf`

jmarrec:
Pull request overview
---------------------
Fix #6442 Equipment Summary has incorrect conversions from SI to IP

The problem was this function call:
```
PreDefTableEntry(pdchDXCoolCoilNetCapSIA, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 1), 1));
```

By doing this, you do not call the function prototype you need for the conversion to happen, which is this one: https://github.com/NREL/EnergyPlus/blob/c1a8cbbf47edf480e3c9e680a32d310591a11b3c/src/EnergyPlus/OutputReportPredefined.cc#L2600

```
void PreDefTableEntry(int const columnIndex, std::string const &objName, Real64 const tableEntryReal, Optional_int_const numSigDigits)
```

This prototype is the only one that will set `origEntryIsReal = true`

So inside WritePredefinedTables, it's not a real on this line, so no conversion happens;

https://github.com/NREL/EnergyPlus/blob/c1a8cbbf47edf480e3c9e680a32d310591a11b3c/src/EnergyPlus/OutputReportTabular.cc#L10894

I extended one unit test to check that the correct conversion factor is chosen based on the column heading, and added a new one that will write this DX Coil table to the SQL with IP conversion, and check that the resulting unit and converted values are ok.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


nrel-bot-3:
@jmarrec @lgentile it has been 14 days since this pull request was last updated.

src/EnergyPlus/StandardRatings.cc
```diff
@@ -2576,14 +2576,21 @@ namespace StandardRatings {
...
+                    // Note JM 2018-10-25: If you call RoundSigDigits(NetCoolingCapRated(Num + 1), 1),
+                    // Then it's not the OutputReportPredefined::PreDefTableEntry prototype with Real64 that is called.
+                    // As a result, the entry isn't marked as being Real (origEntryIsReal) and unit conversion does not occur
+                    // Bad: PreDefTableEntry(pdchDXCoolCoilNetCapSIA, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 1), 1));
+                    PreDefTableEntry(pdchDXCoolCoilNetCapSIA, CompNameNew, NetCoolingCapRated(Num + 1), 1);
+                    PreDefTableEntry(pdchDXCoolCoilNetCapSIB, CompNameNew, NetCoolingCapRated(Num + 2), 1);
+                    PreDefTableEntry(pdchDXCoolCoilNetCapSIC, CompNameNew, NetCoolingCapRated(Num + 3), 1);
+                    PreDefTableEntry(pdchDXCoolCoilNetCapSID, CompNameNew, NetCoolingCapRated(Num + 4), 1);
+
+                    // These will stay in W, so it doesn't matter as much, so I'll leave it that way
```
mjwitte: I would also change these in case a future units option is added that would convert these. It seems that *all* PreDEfTableEntry calls that have RoundSigDigits in the arguments should be changed. Are there any occurrences of this where no conversion is ever wanted? Possible case of this is the Central Plant subtable where there are columns for IPLV in both SI and IP units. See new [comment](https://github.com/NREL/EnergyPlus/issues/6442#issuecomment-445363719)showing what this is currently doing.

mjwitte (commented):
Tested this with the original defect file, and that problem is fixed. Unit tests all run fine.

src/EnergyPlus/StandardRatings.cc
```diff
@@ -2576,14 +2580,21 @@ namespace StandardRatings {
...
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerB, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 2), 1));
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSIC, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 3), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerC, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 3), 1));
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSID, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 4), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerD, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 4), 1));
+                    // Note JM 2018-10-25: If you call RoundSigDigits(NetCoolingCapRated(Num + 1), 1),
+                    // Then it's not the OutputReportPredefined::PreDefTableEntry prototype with Real64 that is called.
+                    // As a result, the entry isn't marked as being Real (origEntryIsReal) and unit conversion does not occur
+                    // Bad: PreDefTableEntry(pdchDXCoolCoilNetCapSIA, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 1), 1));
```
rraustad, line 2586: Is there a way to check inside this function if the string that is passed converts to a Real so that this doesn't happen again?

    void PreDefTableEntry(int const columnIndex, std::string const &objName, std::string const &tableEntryChar);

src/EnergyPlus/StandardRatings.cc
```diff
@@ -2473,8 +2475,9 @@ namespace StandardRatings {
                 PreDefTableEntry(pdchDXHeatCoilType, CompName, CompType);
                 PreDefTableEntry(pdchDXHeatCoilHighCap, CompName, HighHeatingCapVal, 1);
                 PreDefTableEntry(pdchDXHeatCoilLowCap, CompName, LowHeatingCapVal, 1);
-                PreDefTableEntry(pdchDXHeatCoilHSPFIP, CompName, RoundSigDigits(HSPFValueIP, 2));
-                PreDefTableEntry(pdchDXHeatCoilRegionNum, CompName, RoundSigDigits(RegionNum));
+                // Btu/W-h will convert to itself
+                PreDefTableEntry(pdchDXHeatCoilHSPFIP, CompName, HSPFValueIP, 2);
+                PreDefTableEntry(pdchDXHeatCoilRegionNum, CompName, RegionNum, 0);
```
mjwitte: The region number is now showing with a decimal point "4.". Since `RegionNum` is an integer, drop the last argument here and it will use the integer form of the function.

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -1782,6 +1782,8 @@ namespace OutputReportPredefined {
         pdchMechType = newPreDefColumn(pdstMech, "Type");
         pdchMechNomCap = newPreDefColumn(pdstMech, "Nominal Capacity [W]");
         pdchMechNomEff = newPreDefColumn(pdstMech, "Nominal Efficiency [W/W]");
+        // Note JM 2018-12-11: We don't want any of these to convert.
```
mjwitte: Current style is to let git track blame and dates, so delete ~JM 2018-12-11~.

src/EnergyPlus/StandardRatings.cc
```diff
@@ -2576,14 +2580,21 @@ namespace StandardRatings {
...
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSIA, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 1), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerA, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 1), 1));
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSIB, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 2), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerB, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 2), 1));
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSIC, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 3), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerC, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 3), 1));
-                    PreDefTableEntry(pdchDXCoolCoilNetCapSID, CompNameNew, RoundSigDigits(NetCoolingCapRated(Num + 4), 1));
-                    PreDefTableEntry(pdchDXCoolCoilElecPowerD, CompNameNew, RoundSigDigits(TotElectricPowerRated(Num + 4), 1));
+                    // Note JM 2018-10-25: If you call RoundSigDigits(NetCoolingCapRated(Num + 1), 1),
```
mjwitte: Same here, delete ~JM 2018-10-25~.

mjwitte (commented):
Tested before and after this fix, with both SI and IP output:
Coil:Cooling:DX:SingleSpeed - 1ZoneDataCenterCRAC_wApproachTemp.idf
Coil:Cooling:DX:MultiSpeed and Coil:Heating:DX:MultiSpeed - MultispeedHeatPump.idf
Coil:Cooling:DX:TwoSpeed - 5ZoneAutoDXVAV.idf (change coil to AirCooled condenser)
Coil:Cooling:DX:VariableSpeed and Coil:Heating:DX:VariableSpeed - ChangeoverBypassVAV_AirToAirHeatPump.idf
Chiller:ConstantCOP and Chiller:Electric:EIR - ElectricEIRChiller.idf

All of the units in the headings and the values in the tables change as expected for IP vs SI units.  Just a few small things to clean up and then this can drop in.

And there are some unit test failures.
The build warnings appear to be unrelated to this PR, probably caught in the middle of some CI changes.

src/EnergyPlus/StandardRatings.cc
jmarrec: Yeah sorry that's a snippet I have (`note<TAB>` creates a comment with the date). I'll keep that in mind going forward

jmarrec:
@mjwitte  noted your comments and will address as soon as I can which could be only after the holiday or tomorrow

src/EnergyPlus/StandardRatings.cc
jmarrec: @rraustad sorry I missed this initially. 

I feel like it's the responsibility of the dev to ensure he selects the right overloaded prototype, and it would create extra operations to check if it's a real.

Perhaps a compromise might be to use macro code to only do this when E+ is built in `Debug` mode and throw if that's the case?

src/EnergyPlus/StandardRatings.cc
jmarrec: Fixed in 876290d

jmarrec:
@mjwitte I addressed your comment and fixed the unit test.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -15870,15 +15872,16 @@ namespace OutputReportTabular {
...
-        UnitConv(114).mult = 0.737562149277;
-        UnitConv(115).mult = 1.0;
-        UnitConv(116).mult = 0.09290304;
+        UnitConv(114).mult = 1.0;
+        UnitConv(115).mult = 0.737562149277;
+        UnitConv(116).mult = 1.0;
+        UnitConv(117).mult = 0.09290304;
 
         UnitConv(2).offset = 32.0;
         UnitConv(11).offset = 32.0;
```
mjwitte, line 15887: Noticed unexpected diffs in table warmup convergence information [here](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2019-01/8d7fcb46b41e083a30a872f2b9fc23ca65912c2e-45ac1bb721b1bef6a081ccd753f84c3c6543291a/5ZoneAutoDXVAV/Win64-Windows-7-VisualStudio-14/). The offsets did not get shifted, see line 15889 (lame github won't display the entire file, so a link won't work):
`        UnitConv(80).offset = 7.6736; // 80 is KJ/KG -- should this be multiplied by 1000?
`
Because this didn't get moved to (81), it's being applied to deltaC/deltaF. It looks like all the other members of UnitConv have been shifted appropriately.

jmarrec: Good catch! Corrected in latest commit

Not the first time I've fond the UnitConv section to be hard to play with. I'd like to revamp that whole section at some point, use some sort of std::vector<std::pair> to avoid being burned again, make the process of adding a unit easier, and make it easier to read as well.

mjwitte: Agreed, this could be improved. A couple of years ago, the units strings for `SetupOutputVariable` were standardized and converted to use [`OutputProcessor::Unit` enum](https://github.com/NREL/EnergyPlus/blob/b0d6191edfecd4a35971336365ef893651e55199/src/EnergyPlus/OutputProcessor.hh#L188-L239). At some point, we would like to do the same to standardize table units and have a common units conversion function that can be used for both tables and output variables (so the eso could be written directly in IP units).

mjwitte (approved):
The small table diff in HVACTemplate-5ZonePackagedVAV-After is in the DX rating details table, old  "6.6447E-002" is now "0.0664" - which is good. Merging.

mjwitte:
@Myoldmopar Hmmm. After I pressed "confirm merge" this page circled for a while and then came back asking again to "confirm merge". Suspicious of that, I didn't press it again (maybe I should have). Now the PR thinks it's conflicted, but develop shows that this was merged.

PR #7036, opened 2018-10-26
Fix #6720 - Objects with different names will trigger a "Duplicate Name" Error in JSON

### LINKED ISSUE #6720: JSON validation duplicate name problem
Issue overview
--------------
User file with two material objects named "Standard insulation_0.1" and "Standard insulation_0.01" results in
```
   ** Severe  ** Duplicate name found. name: "Standard insulation_0.01". Overwriting existing object.
```
Changing one of them to "Standard insulation_0.01a" fixes the problem.

### Details
 - Platform Win 10
 - Version of EnergyPlus 8.9.0
 - Helpdesk ticket number 12634

### Checklist

jmarrec:
Pull request overview
---------------------

Fix #6720 duplicate name json

Issue was a problem in doj/alphanum.hpp when you have leading zeros. Basically it computes a number for the left side, and another for right side, and compares that. But that means "01" and "001" end up being the same, so it reports them as being equal. Adding a clause where when diff is zero, you also check  the length of the left and right sides, and potentially return that.

Added unit tests, that fail without the fix (see https://github.com/jmarrec/EnergyPlus/tree/Do_Not_Publish/6720_UnitTestOnly) and work with it.

### Rationale for leading zeroes

I had to made a choice for the case where you have leading zeroes as I couldn't really find any information online: I assume that if you have two strings, `"n_001"` and `"n_1"`, I assume you want `"n_001" < "n_1" ` (behavior is easy to change if you disagree)

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@jmarrec could you pull develop into this branch and see if the test failures were due to a CI thing or if this branch actually breaks everything?

jmarrec:
@Myoldmopar done. 

Even though I push to my remote, any maintainer has write access to it so feel free to do that kind of stuff.

If that's easier I could push to NREL/EnergyPlus in the future instead if you prefer (but that increases the attention needed to delete merge branches though in order to not clutter everything).

Myoldmopar:
@jmarrec so this is causing lots of issues.  Our CI bots are failing (panicking on Linux...) due to bad allocations as they run out of memory.  Something has gone very awry with this.  I need to close this PR so that our CI machines don't continue to try to run.

@mjwitte this is the cause of the memory issue I saw which caused me to ramp down the number of processors on the Linux machine.  I'll be ramping it back up now...

Myoldmopar:
Oh, I kinda left that open-ended.  @jmarrec please continue to investigate this, but again, I'm closing this PR so our CI machines won't continue trying to build it.

jmarrec:
@Myoldmopar I caught a tiny mistake in the alphanum stuff (my bad, copy paste error).
I built on Ubuntu and didn't get any problems running the test (but then again, neither did I before). I also built on Windows just now and was able to run the tests as well.

Could we try the linux CI bots again?

Myoldmopar:
CI is _slightly_ happier with this branch now :upside_down_face: :+1: 

I'll take a look at this one today.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -3563,6 +3563,93 @@ TEST_F(InputProcessorFixture, getObjectItem_curve_biquadratic2)
...
+    std::string obj_name = "Material";
+    std::string name1 = "Standard insulation_01";
+    json mat1 = { {"name", name1}, {"Roughness", "MediumRough"}};
+
+    EXPECT_TRUE(mat1.is_object());
+
+    // Add the first material to it
+    root[obj_name][name1] = mat1;
+
+    auto test = [=](std::string search_name) {
```
Myoldmopar, line 3610: @jmarrec it seems odd to put a lambda in here as opposed to making another function, but it's OK with me.

third_party/doj/alphanum.hpp
```diff
@@ -2,38 +2,41 @@
...
 
-    3. This notice may not be removed or altered from any source
-       distribution. */
+   3. This notice may not be removed or altered from any source
+   distribution. */
 
 /* $Header: /code/doj/alphanum.hpp,v 1.3 2008/01/28 23:06:47 doj Exp $ */
 
+
+// MODIFIED: Julien Marrec, EffiBEM, on 2018-10-26 to avoids "n_001" being recognized as the same as "n_01"
```
mbadams5: Typically we haven't been adding attribution/bug issue comments like this in new E+ code and especially in third party libraries.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -3563,6 +3563,93 @@ TEST_F(InputProcessorFixture, getObjectItem_curve_biquadratic2)
...
+        if (it != root[obj_name].end()) {
+            EXPECT_TRUE(false) << it.key();
+        }
+
+        // This works...
+        EXPECT_TRUE(std::find(root[obj_name].begin(), root[obj_name].end(), search_name) == root[obj_name].end());
+    };
+
+    // This all works just fine
+    test("Standard insulation");
```
mbadams5, line 3624: Interesting use of lambda's in the unit test. Seems like it might be overkill for this unit test, but I can think of other unit tests where this approach may come in handy for sure.

mbadams5 (approved):
The changes are appropriate in the alphanum library and there are good unit tests for this bug.

third_party/doj/alphanum.hpp
Myoldmopar: @jmarrec it's important to track our changes to third_party libraries, but I don't like comments as the mechanism.  These can easily get overridden when we bring in a new version of the library.  I'm going to transform this comment into the start of a new file at `third_party/PATCHES.md` where we'll keep track of all the changes we make to third_party libraries.

Myoldmopar (approved):
@jmarrec I'm going to push a comment-only commit and merge this in.  Thanks!

jmarrec:
I’m still learning the E+ coding style so please continue to flag unusual stuff like this (and please excuse said unusual stuff, I don’t do it on purpose!).

As for the use of lambda here, it’s a short code block that wouldn’t be used by any other test (but I hate repeating myself) so I thought a lambda made sense here. (And I guess I could have cleaned it up a bit though)

PR #7040, opened 2018-10-30
Heat Emissions from Buildings to Ambient

xuanluo113:
Pull request overview
---------------------
Heat emissions from buildings are a crucial component influencing climate change and urban microclimate, e.g., urban heat island effect. We propose to calculate heat emissions from buildings in EnergyPlus with a bottom-up approach and report them using new output variables as well as new tables in the current tabular summary reports. The heat emissions from a building include heat releases from three levels of building systems: building envelope, zone/space, and HVAC systems. The feature would allow users to quantify the dynamic behavior of heat emissions, and analyze their influencing factors including building use type, vintage, and climate. This feature directly supports DOE’s multi-year effort of enhancing EnergyPlus to support urban scale building modeling.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes:
   n/a If transition, add to input rules file for interfaces
 n/a ExpandObjects changes?


design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+
+We propose a bottom-up approach to calculate heat emissions from buildings with any building energy modeling programs. The scope is restricted to report the buildings’ heat directly discharged to their ambient air. As illustrated in Figure 1, we propose to include heat releases from three levels of building components: (1) building envelope - exterior surface convection and radiative heat transfer to the ambient air, (2) zone - exhaust air or exfiltration to ambient, (3) HVAC system - relief/exhaust air from AHUs, and (4) HVAC system - heat rejected by air-cooled condensers or central plants, including cooling towers of water-cooled chillers, gas-fired boilers, furnaces and water heaters. It should be noted that any under-ground zone or envelope is excluded, since heat from those spaces is rejected to ground instead of to ambient air. 
+
```
lgu1234, line 18: Do you include exhaust air from heat exchangers in Item 3?

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+**1.	Heat emission from building envelope**
+
+Exterior surfaces emit heat to ambient air in the form of convection and long-wave radiation, which could be absorbed by particles and gas molecules in air. Figure 2 illustrates the heat balance of an exterior surface. The internal sources modify the outside surface temperature through envelope conduction, along with incident solar radiation, which triggers the heat exchange with the ambient air. _Surface Outside Face Convection Heat Gain Rate_ and _Surface Outside Face Thermal Radiation to Air Heat Transfer Coefficient_ have already been reported. We would calculate the _Surface Outside Face Thermal Radiation to Air Heat Transfer Rate_ as:
+
+Q<sub>lwr, surf</sub> = Hr * A * (T<sub>surf</sub> - T<sub>out</sub>) &nbsp;&nbsp; Eq. (1),
+
+where **Hr** is the surface outside face thermal radiation to air heat transfer coefficient, **A** is the surface area, **T<sub>surf</sub>** and **T<sub>out</sub>** are the surface and outdoor air temperature.
+
+![Diagram](bldgenvelopheatbalance.png)
+
```
lgu1234, line 36: I am not sure if solar reflection is a part of heat emission from building envelope. However, solar reflection is caused by a building.

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+Detailed calculation of HVAC rejected heat varies with different HVAC object groups and component types, as summarized in Table 1.
+
+**Table 1. Heat rejection calculation for different HVAC component types**
+
+Object Group | Component | Calculation Methods
+-------------  | -------------  | -------------
+Coils | Chilled water cooling coil | Emit heat in condensing unit defined in Condenser Equipment
+ |  | Hot water heating coil          | Emit heat in plant source defined in Plant Heating and Cooling Equipment
+ |  | DX cooling coil (air-cooled)    | Cooling Coil Source Side Heat Transfer Rate + Cooling Coil Electric Power
```
lgu1234: Coil:Cooling:DX:MultiSpeed allows multiple fuel types, in addition to electricity. If the fuel type is not electricity, it will have waste heat. Do you include?

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+
+Q<sub>exf, zone, sensible</sub> = m<sub>exf</sub> * c * (T<sub>zone</sub> - T<sub>out</sub>) &nbsp;&nbsp; Eq. (4),
+
+Q<sub>exh, zone, sensible</sub> = m<sub>exh</sub> * c * (T<sub>zone</sub> - T<sub>out</sub>) &nbsp;&nbsp; Eq. (5),
+
+where **c** is the specific heat of air, **T<sub>zone</sub>** and **T<sub>out</sub>** are the zone and outdoor air temperature. And the latent parts are determined by taking the difference between the total and the sensible rate.
+
+In EnergyPlus, for models without airflow network objects, when a _ZoneAirBalance:OutdoorAir_ object is assigned, the sensible and latent heat loss and gain due to OA air balance for each zone are reported and we will count these as part of the zone heat emission. When no _ZoneAirBalance:OutdoorAir_ object is assigned, we will count the reported sensible and latent heat loss and gain for both infiltration and ventilation at each zone for this part. For airflow network models, the infiltration and ventilation heat loss and gain are also reported for each zone.
```
mjwitte: Also need to consider the settings for ZoneAirMassFlowConservation if present. Exfiltration mass flow will be messy to calculate. In some cases, infiltration = exfiltration, but exfiltration could be less than or greater than infiltration if the HVAC system and exhaust fans are unbalanced or this there is zone-to-zone mixing flow. So, it may be most reliable to do a mass flow balance on the zone including infiltration, zone-to-zone mixing, and all of the zone inlet, exhaust, and return nodes to determine if there is any flow left over to exfiltrate.

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+
+The following new report variables will be added:
+
+- Surface, Average, Surface Outside Face Thermal Radiation to Air Heat Transfer Rate
+- Zone, Average, Zone Exfiltration Heat Transfer Rate
+- Zone, Average, Zone Exfiltration Sensible Heat Transfer Rate
+- Zone, Average, Zone Exfiltration Latent Heat Transfer Rate
+- Zone, Average, Zone Relief Air Heat Transfer Rate
+- Zone, Average, Zone Relief Air Sensible Heat Transfer Rate
+- Zone, Average, Zone Relief Air Latent Heat Transfer Rate
```
mjwitte: Should these three be Zone *Exhaust* Air . . . ?

design/FY2019/Heat Emissions from Buildings to Ambient.md
mjwitte: DX coils can also be evaporatively cooled.

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+ |  | VRF air-to-air heat pump condensing unit (evaporative-cooled) | VRF Heat Pump Total Cooling Rate + VRF Heat Pump Total Electric Power + VRF Heat Pump Evaporative Condenser Pump Electric Power (cooling mode) <br>VRF Heat Pump Total Heating Electric Power / Fuel Rate + VRF Heat Pump Evaporative Condenser Pump Electric Power - VRF Heat Pump Total Heating Rate (heating mode)
+ |  | VRF air-to-air heat pump condensing unit (water-cooled) | VRF Heat Pump Condenser Heat Transfer Rate
+Plant Heating and Cooling Equipment | Hot water/Steam boiler | Boiler <Fuel Type> Rate - Boiler Heating Rate
+ |  | Chiller water-cooled | Emit heat in Condenser Equipment
+ |  | Chiller air-cooled   | Chiller Condenser Heat Transfer Rate
+Condenser Equipment | Cooling Tower | Cooling Tower Heat Transfer Rate + Cooling Tower Fan Electric Power
+Water Heaters and Thermal Storage | Water heater | Water Heater <Fuel Type> Rate - Water Heater Heating Rate
```
mjwitte, line 105: 1. Also need to look at water heater location (Ambient Temperature Indicator), the tank losses could be to the outdoors. 
2. Same for ThermalStorage:ChilledWater:*
3. ThermalStorage:Ice:* do not indicate where losses go, so I guess you neglect this.
4. Heat pump water heaters can draw heat from outdoor air.
5. Refrigeration:* equipment also can reject outdoors.
6. ElectricEquipment, GasEquipment, etc. all have a "Fraction Lost" field. This is lost from the zone, could be to outdoors or could be to a process load. Not sure if we should count this or not.

design/FY2019/Heat Emissions from Buildings to Ambient.md
```diff
@@ -0,0 +1,216 @@
...
+- Zone, Average, Zone Relief Air Heat Transfer Rate
+- Zone, Average, Zone Relief Air Sensible Heat Transfer Rate
+- Zone, Average, Zone Relief Air Latent Heat Transfer Rate
+- HVAC, Average, HVAC Relief Air Heat Transfer Rate
+- HVAC, Average, HVAC Relief Air Sensible Heat Transfer Rate
+- HVAC, Average, HVAC Relief Air Latent Heat Transfer Rate
+- HVAC, Average, HVAC Heat Rejection Rate
+- HVAC, Average, HVAC Heat Rejection Sensible Rate
+- HVAC, Average, HVAC Heat Rejection Latent Rate
+- HVAC, Sum, Environmental Impact Heat Emissions
```
mjwitte, line 195: What do you see as the key values for the HVAC outputs? For relief air, it could be the OA mixer object name, but currently most (all?) air loop outputs use the air loop name. 
There is an existing meter enduse category called "HeatRejection" which is used for electricity (e.g. cooling tower electricity), and EnergyTransfer ( e.g. chiller heat transfer, etc.)  It may be better to use "HVAC Heat Emission . . ." for these output variable names. We could add new EnergyTransfer meter end-uses for "HeatEmissionSensible" and "HeatEmissionLatent" so that the components feeding into these meters can be reported in the mtd output. Something to discuss.

rraustad:
More broadly, consider a plot of land where there is no building. This land area absorbs solar energy during the day and would radiate energy to the night sky. Then put a building in this location. The solar gain will now be included as heat rejected to atmosphere. How do you suggest to account for solar energy that would naturally be in this location when the building is not present?

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: Yes. We included this part.

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: The intention is to include the thermal impact of buildings to their ambient so the net thermal radiation to the air of surfaces is included. @lgu1234

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: @lgu1234 @mjwitte We added the cases for other fuel types and the case for evaporatively cooled

nrel-bot-3:
@XuanLuoLBL @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@XuanLuoLBL @lgentile it has been 14 days since this pull request was last updated.

nrel-bot-3:
@XuanLuoLBL @lgentile it has been 14 days since this pull request was last updated.

src/EnergyPlus/HeatBalanceAirManager.cc
```diff
@@ -515,6 +515,41 @@ namespace HeatBalanceAirManager {
                 SetupOutputVariable(
                     "Zone Air Heat Balance Deviation Rate", OutputProcessor::Unit::W, ZnAirRpt(Loop).imBalance, "System", "Average", Zone(Loop).Name);
             }
+
+            if (!DisplayAdvancedReportVariables) {
```
rraustad: Did you intend to turn off these report variables if DisplayAdvancedReportVariables is true?

xuanluo113: @rraustad Yes. Actually this is a design decision I haven't thought through and was intended to discuss with reviewers. Currently we have a flag for the heat emission - output:table:tabular, but there's no global flag for heat emission - turn on additional report variables. 

Intermediate vars (such as surface level heat emission, zone level exfiltration and exhaust) are calculated and reported if DisplayAdvanced, regardless of whether user declare this output as a key in Output:Table:Tabular or not.

So I was thinking if it's preferable to add a heat emission calc and report flag as well, for example, as a simulation parameter.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -11273,4 +11297,4 @@ namespace AirflowNetworkBalanceManager {
 
 } // namespace AirflowNetworkBalanceManager
 
-} // namespace EnergyPlus
+} // namespace EnergyPlus
```
mjwitte: Missing return at the end of this file.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -2070,6 +2091,10 @@ namespace OutputReportTabular {
                     displayEconomicResultSummary = true;
                     WriteTabularFiles = true;
                     nameFound = true;
+                } else if (UtilityRoutines::SameString(AlphArray(iReport), "HeatEmissionsSummary")) {
+                    displayHeatEmissionsSummary = true;
```
mjwitte, line 2096: Also add this line to the lists under AllSummary, AllSummaryAndMonthly, etc. [here](https://github.com/NREL/EnergyPlus/blob/bd2dd030748bb8199260801bc665f893a7d069e9/src/EnergyPlus/OutputReportTabular.cc#L2109) and following. Actually, that section could be consolidated with ||s to have just one list for summary, one list for monthly, etc. But that's not your job here. The monthly reports are done with a loop, so that should already be covered.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10700,6 +11024,82 @@ namespace OutputReportTabular {
         }
     }
 
+    void WriteHeatEmissionTable()
+    {
+
+        // SUBROUTINE INFORMATION:
+        //       AUTHOR         Xuan Luo
+        //       DATE WRITTEN   November 2018
+        //       MODIFIED       na
```
mjwitte: Remove all boilerplate comments, just simple comments to describe what the function does.
@JasonGlazer Would this have been easier to set up using the predefined report methods? Is that the preferred method for new reports? Not saying this needs to be changed now, just wondering.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4766,6 +4797,298 @@ namespace OutputReportTabular {
         }
     }
 
+    void GatherHeatEmissionReport(int const IndexTypeKey)
+    {
+        // SUBROUTINE INFORMATION:
+        //       AUTHOR         Jason Glazer
+        //       DATE WRITTEN   August 2011
+        //       MODIFIED       na
+        //       RE-ENGINEERED  na
```
mjwitte: Remove old comments.

src/EnergyPlus/HeatBalanceAirManager.cc
mjwitte: @XuanLuoLBL It's fine to make the new report variables active only with DisplayAdvancedReportVariables. 

I would only add a new simulation parameter if you think the extra calculations will drag down performance.   If you do, then just one flag should be adequate - if true then do the calcs and set up the report variables. And if you do add it, then don't require DisplayAdvancedReportVariables for these.

mjwitte (commented):
@XuanLuoLBL I jumped here because of @rraustad 's question about the report variables. I noticed a few other things, but haven't done a complete review here.

src/EnergyPlus/OutputReportTabular.cc
xuanluo113: @mjwitte Thanks Mike! For functions in this namespace I'm doing a slight change of design for better aggregating the monthly report variables. Haven't push it yet as unit tests need to be modified, and my clang formatting config in Visual Studio is somehow messed up at this time point... Will do soon and let reviewers know.

But I have been switching back and forth between predefined or written tables, so it would be great if I can get some general advice at this point.

src/EnergyPlus/DXCoils.cc
```diff
@@ -12339,7 +12339,9 @@ namespace DXCoils {
                 HeatReclaimDXCoil(DXCoilNum).AvailCapacity = DXCoil(DXCoilNum).TotalCoolingEnergyRate + DXCoil(DXCoilNum).ElecCoolingPower;
 
                 // Waste heat calculation
-                if (DXCoil(DXCoilNum).MSHPHeatRecActive) {
+                // TODO: waste heat not considered even if defined in Cooling:DX:MultiSpeed, N16, \field Speed 1 Rated Waste Heat Fraction of Power
+                // Input
+                if (DXCoil(DXCoilNum).FuelType != FuelTypeElectricity) {
```
xuanluo113, line 12344: @mjwitte I was hoping you can take a look at this piece. I checked you previous comments on considering waste heat for Cooling:DX:MultiSpeed in non- FuelTypeElectricity mode. But I found this part is currently not calculated even when user defines, as the calculation is only triggered when MSHPHeatRecActive.

Would you mind taking a look at whether my understanding and modifications are in the right direction? Thanks very much.

src/EnergyPlus/HeatBalanceAirManager.cc
xuanluo113: @mjwitte Thanks Mike. At design stage I thought extra calculation is comparably minimal so I implement it this way. But I'll double check this.

mjwitte: OK, but this line should be `if (DisplayAdvancedReportVariables)` ,without !.

xuanluo113:
@mjwitte @lgu1234 @rraustad I finished coding, addressing the NFP comments, along with some unit tests and testfiles (while still debugging the failed regression test as a standalone issue). Wondering if you can take a look when you have time. @JasonGlazer It would be great to have your comments as well as this relates to output report.
I'd appreciate your help.

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: @mjwitte Taking the advice, we explicitly calculate 

> the mass flow balance on the zone including infiltration, zone-to-zone mixing, and all of the zone inlet, exhaust, and return nodes to determine if there is any flow left over to exfiltrate.

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: 1 - 2: location checked.
4 - 6: considered in the calculation

3 is neglected and updated in the documents.

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: @mjwitte modified the name to Exhaust .

design/FY2019/Heat Emissions from Buildings to Ambient.md
xuanluo113: For relief air I'm using OA mixer controller names as the keys. I'm not sure about the answer to heat rejection. I still want to report it as an output variable to be consistent. Current I apply no key to it, just as the variable "HVAC System Solver Iteration Count".

src/EnergyPlus/OutputReportTabular.cc
xuanluo113: Added for `"AllSummary"`, `"AllSummaryAndSizingPeriod"`, `"AllSummaryAndMonthly"`, `"AllSummaryMonthlyAndSizingPeriod"`.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -8488,6 +8491,7 @@ namespace AirflowNetworkBalanceManager {
                             AirflowNetworkReportData(ZN1).MultiZoneMixSenLossJ +=
                                 (AirflowNetworkLinkReport1(i).FLOW2OFF * CpAir * (MAT(ZN1) - MAT(ZN2))) * ReportingConstant * ReportingFraction;
                         }
+                        // TODO: H2O Heat of Vap?
```
mjwitte: Is this still a ToDo or can this comment be deleted?

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -8517,6 +8521,7 @@ namespace AirflowNetworkBalanceManager {
             Tamb = Zone(i).OutDryBulbTemp;
             CpAir = PsyCpAirFnWTdb(ZoneAirHumRatAvg(i), MAT(i));
             AirDensity = PsyRhoAirFnPbTdbW(OutBaroPress, MAT(i), ZoneAirHumRatAvg(i));
+            // TODO: InfilHeatLoss is watt?
```
mjwitte: ToDo?

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -8536,6 +8541,30 @@ namespace AirflowNetworkBalanceManager {
                     ZonePreDefRep(i).AFNInfilVolMin = AirflowNetworkZnRpt(i).InfilVolume * Zone(i).Multiplier * Zone(i).ListMultiplier;
                 }
             }
+
+            // TODO: add report var?
```
mjwitte: ToDo?

src/EnergyPlus/AirflowNetworkBalanceManager.hh
```diff
@@ -146,12 +146,18 @@ namespace AirflowNetworkBalanceManager {
         Real64 VentilAirTemp;      // Air Temp {C} of ventilation
         Real64 MixVolume;          // Mixing volume of Air {m3}
         Real64 MixMass;            // Mixing mass of air {kg}
+        Real64 ExfilSensiLoss;     // Sensible heat Loss rate {W} due to exfiltration
+        Real64 ExfilLatentLoss;    // Latent heat Loss rate {W} due to exfiltration
+        Real64 ExfilTotalLoss;     // Total heat Loss rate {W} due to exfiltration
+        Real64 ExfilMass;          // Mass of Air {kg} due to exfiltration
+        Real64 InletMass;          // Mass of Air {kg} due to exfiltration
+        Real64 OutletMass;         // Mass of Air {kg} due to exfiltration
```
mjwitte: Minor quibble - From the zone's perspective, exfiltration is not really a loss. I would perhaps name these `ExfilSensiHeatEmission` etc. But I see "Loss" has been used for all the new variable names for heat emissions. Then leave it as-is.

The comments on `InletMass` and `OutletMass` are pasteos from `ExfilMass`.

src/EnergyPlus/DXCoils.cc
```diff
@@ -12352,7 +12354,11 @@ namespace DXCoils {
                         WasteHeatHS = CurveValue(DXCoil(DXCoilNum).MSWasteHeat(SpeedNumHS), OutdoorDryBulb, InletAirDryBulbTemp) *
                                       DXCoil(DXCoilNum).MSWasteHeatFrac(SpeedNumHS);
                     }
-                    MSHPWasteHeat = (SpeedRatio * WasteHeatHS + (1.0 - SpeedRatio) * WasteHeatLS) * DXCoil(DXCoilNum).ElecCoolingPower;
+                    DXCoil(DXCoilNum).MSFuelWasteHeat =
+                        (SpeedRatio * WasteHeatHS + (1.0 - SpeedRatio) * WasteHeatLS) * DXCoil(DXCoilNum).ElecCoolingPower;
```
mjwitte, line 12358: It's confusing here (and below) to see `MSFuelWasteHeat` calculated using `ElecCoolingPower` (or `ElecHeatingPower` below).

src/EnergyPlus/DXCoils.cc
```diff
@@ -12569,10 +12575,13 @@ namespace DXCoils {
                 // Waste heat
                 if (DXCoil(DXCoilNum).MSHPHeatRecActive) {
                     if (DXCoil(DXCoilNum).MSWasteHeat(SpeedNum) == 0) {
-                        MSHPWasteHeat = DXCoil(DXCoilNum).MSWasteHeatFrac(SpeedNum) * DXCoil(DXCoilNum).ElecCoolingPower;
+                        DXCoil(DXCoilNum).MSFuelWasteHeat = DXCoil(DXCoilNum).MSWasteHeatFrac(SpeedNum) * DXCoil(DXCoilNum).ElecCoolingPower;
                     } else {
-                        MSHPWasteHeat = CurveValue(DXCoil(DXCoilNum).MSWasteHeat(SpeedNum), OutdoorDryBulb, InletAirDryBulbTemp) *
-                                        DXCoil(DXCoilNum).MSWasteHeatFrac(SpeedNum) * DXCoil(DXCoilNum).ElecCoolingPower;
+                        DXCoil(DXCoilNum).MSFuelWasteHeat = CurveValue(DXCoil(DXCoilNum).MSWasteHeat(SpeedNum), OutdoorDryBulb, InletAirDryBulbTemp) *
+                                                            DXCoil(DXCoilNum).MSWasteHeatFrac(SpeedNum) * DXCoil(DXCoilNum).ElecCoolingPower;
```
mjwitte, line 12581: If waste heat is not active, then this curve may not be defined. I don't see any code in the PR to check this and throw a warning.
And I don't see any code to make sure `MSFuelWasteHeat` gets set to zero if the unit is not running.
It's hard to follow these changes here in the github diff. I'll have to look at this in more depth.

src/EnergyPlus/DataHeatBalSurface.hh
```diff
@@ -129,6 +129,9 @@ namespace DataHeatBalSurface {
     extern Array1D<Real64> QRadOutReport;        // Surface thermal radiation heat gain at Outside face [J]
     extern Array1D<Real64> QdotRadOutRep;        // Surface thermal radiation heat transfer outside face surface [W]
     extern Array1D<Real64> QdotRadOutRepPerArea; // [W/m2]Surface thermal radiation heat transfer rate per m2 at
+
+    extern Array1D<Real64> QAirExtReport;  // Surface Outside Face Thermal Radiation to Air Heat Transfer Rate [W]
```
mjwitte, line 133: Seems that Surface Outside Face Thermal Radiation to *Ground* Heat Transfer Rate should also be tracked and added to `QHeatEmiReport`.
In the later calculations of this, I was confused by variable `HAirExtSurf` and noticed that [these old comments are wrong](https://github.com/NREL/EnergyPlus/blob/e9cf73860342b899068a32749a1cff5ff50aa9cb/src/EnergyPlus/DataHeatBalSurface.cc#L98-L100). Please fix them.

*End MJW review here for today.*

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: @mjwitte @lgu1234 

This is more of a question than to do. I noticed that in the previous code the latent heat is calculated as
```
AirflowNetworkReportData(ZN1).MultiZoneMixLatGainW += (AirflowNetworkLinkReport1(i).FLOW2OFF * (ZoneAirHumRat(ZN2) - ZoneAirHumRat(ZN1))) * ReportingFraction;
AirflowNetworkReportData(ZN1).MultiZoneMixLatGainJ += (AirflowNetworkLinkReport1(i).FLOW2OFF * (ZoneAirHumRat(ZN2) - ZoneAirHumRat(ZN1))) * ReportingConstant * ReportingFraction;
```

I'm not sure why it does not need to be multiplied by H20 latent heat of evaporation here. I calculated the exfiltration latent heat loss as
```
AirflowNetworkZnRpt(i).ExfilLatentLoss =
                AirflowNetworkZnRpt(i).ExfilMass / ReportingConstant * (ZoneAirHumRat(i) - OutHumRat) * H2OHtOfVap;
```
and I'm not sure if the unit is right here.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
xuanluo113: @mjwitte @lgu1234 
Based on the implementation it seems the unit of InfilHeatGain should be [J] instead of [W] (as multiplied by ReportingConstant? But the comments in the header indicates it should be watt values. And I wanted to report exfiltration in watts as well.
```
    struct AirflowNetworkReportVars
    {
        // Members
...
        Real64 InfilHeatGain;      // Heat Gain {W} due to infiltration
        Real64 InfilHeatLoss;      // Heat Loss {W} due to infiltration
...
        Real64 ExfilSensiLoss;     // Sensible heat Loss rate {W} due to exfiltration
        Real64 ExfilLatentLoss;    // Latent heat Loss rate {W} due to exfiltration
```

src/EnergyPlus/DXCoils.cc
xuanluo113: @mjwitte I didn't realize that CoilDX_MultiSpeedCooling can only be referenced by a AirLoopHVAC:UnitaryHeatPump:AirToAir:Multispeed object. I'll revert this piece back.

mjwitte: It can also be used by `AirLoopHVAC:UnitarySystem`. I'm not following why you would revert this.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
mjwitte: I agree, this does look wrong.  @lgu1234 new issue posted https://github.com/NREL/EnergyPlus/issues/7162

src/EnergyPlus/AirflowNetworkBalanceManager.cc
mjwitte: @XuanLuoLBL I agree these look incorrect. Searching further, it appears that these are never used. See https://github.com/NREL/EnergyPlus/issues/7162#issuecomment-461261073

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4804,263 @@ namespace OutputReportTabular {
...
+        }
+
+        // Refrigerated Case - Condenser
+        for (iRef = 1; iRef <= NumRefrigCondensers; ++iRef) {
+            SysTotalHVACRejectHeatLoss += RefrigRack(iCoil).CondEnergy;
+        }
+
+        // Evaporative coolers
+        for (iCooler = 1; iCooler <= NumEvapCool; ++iCooler) {
+            SysTotalHVACRejectHeatLoss += EvapCond(iCooler).EvapWaterConsump * RhoWater * H2OHtOfVap_HVAC + EvapCond(iCooler).EvapCoolerEnergy;
```
mjwitte, line 5061: @XuanLuoLBL When all unit tests are run serially, this line fails with an array bounds error. There is no `clear_state` function for EvaporativeCoolers to reset NumEvapCool to zero and reset or deallocate other variables. If you need help adding this, let me know.

src/EnergyPlus/DataHeatBalSurface.hh
xuanluo113: @mjwitte I'm not sure about this. As in the NFP this feature is to calculate the heat emission to ambient **air**, so Radiation to Ground or Sky may not be considered.

src/EnergyPlus/DXCoils.cc
xuanluo113: @mjwitte Originally the DataHVACGlobals::MSHPWasteHeat is calculated as:
```
MSHPWasteHeat = (SpeedRatio * WasteHeatHS + (1.0 - SpeedRatio) * WasteHeatLS) * DXCoil(DXCoilNum).ElecCoolingPower;                  
```
I was trying to use this and assign it as a DXCoil variable.

I now realize this global waste heat is calculated with each DXCoilNum, overwriting one another instead of aggregating, which is weird to me.

src/EnergyPlus/DXCoils.cc
xuanluo113: Originally the waste heat is only calculated when `DXCoil(DXCoilNum).MSHPHeatRecActive`. I thought it only applies to HP cases, so I modified it as - calculating the waste heat when 
`DXCoil(DXCoilNum).FuelType != FuelTypeElectricity` and waste heat is defined.

But since `DXCoil(DXCoilNum).MSHPHeatRecActive` is triggered in **AirLoopHVAC:UnitaryHeatPump:AirToAir:Multispeed** and **AirLoopHVAC:UnitarySystem**, and those are the only cases apply, using the original logic should be fine.

src/EnergyPlus/DataHeatBalSurface.hh
mjwitte: I agree that radiation to sky would not be counted, but some portion of radiation to ground will ultimately end up in the air. I haven't reviewed the docs yet for this, but it should be made clear that radiant exchange to the sky and ground are not counted.

mjwitte:
@XuanLuoLBL @Myoldmopar Something did not go well with the merge from develop here.

xuanluo113:
@mjwitte @Myoldmopar redid the merge

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4804,263 @@ namespace OutputReportTabular {
...
+        static Real64 RhoWater = Psychrometrics::RhoH2O(DataEnvironment::OutDryBulbTemp);
+
+        SysTotalHVACReliefHeatLoss = 0;
+        SysTotalHVACRejectHeatLoss = 0;
+
+        if (!DoWeathSim) return;
+
+        if (!displayHeatEmissionsSummary) return; // don't gather data if report isn't requested
+
+        if (IndexTypeKey != stepTypeZone) return; // only add values over the zone timestep basis
```
mjwitte: No, this is incorrect. Nearly all of the variables gathered below are at the system timestep, so this needs to accumulate at every HVAC timestep. It think only `SumSurfaceHeatEmission` is at the zone timestep, everything else is HVAC or a mix - see more comments below. So, instead of returning for `!stepTypeZone` this should skip only `BuildingPreDefRep.emiEnvelopConv`.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -10828,8 +11147,84 @@ namespace OutputReportTabular {
...
+        }
+    }
+
+    void WriteHeatEmissionTable()
+    {
+
+        // SUBROUTINE INFORMATION:
+        //       AUTHOR         Xuan Luo
+        //       DATE WRITTEN   November 2018
+        //       MODIFIED       na
```
mjwitte: delete all of these old boilerplate comments

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4804,263 @@ namespace OutputReportTabular {
...
+        static int iTank(0);
+        static int iRef(0);
+
+        static Real64 H2OHtOfVap_HVAC = Psychrometrics::PsyHgAirFnWTdb(DataEnvironment::OutHumRat, DataEnvironment::OutDryBulbTemp);
+        static Real64 RhoWater = Psychrometrics::RhoH2O(DataEnvironment::OutDryBulbTemp);
+
+        SysTotalHVACReliefHeatLoss = 0;
+        SysTotalHVACRejectHeatLoss = 0;
+
+        if (!DoWeathSim) return;
```
mjwitte: Redundant. The call to `GatherHeatEmissionReport` is already inside an `if DoWeathSim`.

src/EnergyPlus/HeatBalanceAirManager.cc
```diff
@@ -515,8 +515,48 @@ namespace HeatBalanceAirManager {
                 SetupOutputVariable(
                     "Zone Air Heat Balance Deviation Rate", OutputProcessor::Unit::W, ZnAirRpt(Loop).imBalance, "System", "Average", Zone(Loop).Name);
             }
+
+            if (DisplayAdvancedReportVariables) {
+                SetupOutputVariable("Zone Exfiltration Heat Transfer Rate",
+                                    OutputProcessor::Unit::W,
+                                    ZnAirRpt(Loop).ExfilTotalLoss,
+                                    "Zone",
```
mjwitte: These all need to be at the System timestep.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -5602,6 +5602,26 @@ namespace AirflowNetworkBalanceManager {
                                 Zone(i).Name);
             SetupOutputVariable("AFN Zone Mixing Volume", OutputProcessor::Unit::m3, AirflowNetworkZnRpt(i).MixVolume, "System", "Sum", Zone(i).Name);
             SetupOutputVariable("AFN Zone Mixing Mass", OutputProcessor::Unit::kg, AirflowNetworkZnRpt(i).MixMass, "System", "Sum", Zone(i).Name);
+            if (DataGlobals::DisplayAdvancedReportVariables) {
+                SetupOutputVariable("AFN Zone Exfiltration Heat Transfer Rate",
+                                    OutputProcessor::Unit::W,
+                                    AirflowNetworkZnRpt(i).ExfilTotalLoss,
+                                    "Zone",
```
mjwitte: Again, these need to be at the System timestep.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -1750,6 +1753,18 @@ namespace HeatBalanceSurfaceManager {
                                     "Zone",
                                     "State",
                                     Surface(loop).Name);
+                SetupOutputVariable("Surface Outside Face Thermal Radiation to Air Heat Transfer Rate",
+                                    OutputProcessor::Unit::W,
+                                    QAirExtReport(loop),
+                                    "Zone",
```
mjwitte, line 1759: This is correct. The three new surface outputs are the only ones that accumulate at the zone timestep.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4804,263 @@ namespace OutputReportTabular {
...
+
+        BuildingPreDefRep.emiEnvelopConv += SumSurfaceHeatEmission;
+
+        BuildingPreDefRep.emiZoneExfiltration += ZoneTotalExfiltrationHeatLoss;
+
+        BuildingPreDefRep.emiZoneExhaust += ZoneTotalExhaustHeatLoss;
+
+        // HVAC relief air
+        for (iOACtrl = 1; iOACtrl <= NumOAControllers; ++iOACtrl) {
+            SysTotalHVACReliefHeatLoss += OAController(iOACtrl).RelTotalLossRate * TimeStepZoneSec * convertJtoGJ;
```
mjwitte: From here on to the end, `TimeStepZoneSec` should be `TimeStepSysSec` (which doesn't exist, but can be a local here as `TimeStepSysSec * SecInHour`.

mjwitte (changes requested):
There are zone vs system timestep problems.

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -6898,39 +7053,17 @@ TEST_F(EnergyPlusFixture, OutputReportTabularMonthlyPredefined_FindNeededOutputV
...
-                                "Sum",
-                                ZoneNames[i]);
-        SetupOutputVariable("Zone Cooling Setpoint Not Met While Occupied Time",
-                                OutputProcessor::Unit::hr,
-                                timeNotMet,
-                                "Zone",
-                                "Sum",
-                                ZoneNames[i]);
+        SetupOutputVariable("Zone Heating Setpoint Not Met Time", OutputProcessor::Unit::hr, timeNotMet, "Zone", "Sum", ZoneNames[i]);
+        SetupOutputVariable("Zone Heating Setpoint Not Met While Occupied Time", OutputProcessor::Unit::hr, timeNotMet, "Zone", "Sum", ZoneNames[i]);
```
mjwitte, line 7059: @Myoldmopar Quite a few changes in this file and elsewhere in this PR are reformatting changes. Which form is correct, the new changes here or what we had before? This seems to happen with VS mostly. Either VS is interpreting the clang formatting rules incorrectly, or maybe there's some additional settings we need to declare that default differently in VS than in clangformat? Fishing for a solution - tired of wading through superfluous diffs.

xuanluo113: @mjwitte Seems like my own VS fault. I reinstalled clang and it revert the clang changes. Will clean this up today.

mjwitte: Interesting. Reinstalled clang? I thought it was built-in for VS?

xuanluo113: Huh. Why not for mine... I installed it as a plugin/extension.

mjwitte: clang support was added in VS2017

xuanluo113:
@mjwitte I guess I've addressed most of your comments in the PR except:

- [x] For formatting, I reapplied the Clang and most of the changes don't go way. I took another look and it seems the changes show in this PR are supposed to be made, taking https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/tst/EnergyPlus/unit/OutputReportTabular.unit.cc#L7145 as an example. 

- [x] I don't get the linux build fail massage, and the two time-out integration tests, which I tested to run successfully locally.

- [ ] I added all the clean-ups I can think of in the `EvaporativeCoolers` and `EvaporativeFluidCoolers`, but I'm not sure if I should call it (`::clear_state()`) somewhere in other unit tests. Are they still causing problems?

- [ ] The issues in DXCoil.cc https://github.com/NREL/EnergyPlus/pull/7040/files#r255281521 here remain.

- [x] Left some minor changes in documentation, such as change some "zone"s to "system"s, but I'd hope to do it altogether after the review of docs.

- [ ] Use Joule instead of GJ.

src/EnergyPlus/HeatBalanceAirManager.cc
```diff
@@ -520,42 +520,46 @@ namespace HeatBalanceAirManager {
...
+                                    "System",
                                     "Average",
                                     Zone(Loop).Name);
                 SetupOutputVariable("Zone Exfiltration Sensible Heat Transfer Rate",
                                     OutputProcessor::Unit::W,
                                     ZnAirRpt(Loop).ExfilSensiLoss,
-                                    "Zone",
-                                    "Average",
+                                    "System",
+                                    "System",
```
mjwitte: This is causing error diffs. should be "Average".

mjwitte:
@Myoldmopar Can you help track down where this [build error](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-452ad34e1c1dd56c687a209ec7fd73d266bb116e-x86_64-Linux-Ubuntu-18.04-gcc-7.3.html) is from?

mjwitte:
"bad math environment" sounds like a latex error?

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+\centering
+\includegraphics[width=0.9\textwidth, height=0.9\textheight, keepaspectratio=true]{media/heatemission.png}
+\caption{Building released heat compositions \protect \label{fig:building-released-heat-compositions}}
+\end{figure}
+
+The heat emission we calculate and report measures the heat transfer from buildings to their ambient environment, and can have both positive and negative values. The positive values indicate the building injects heat to the environment, while the negative values indicate the building extracts heat from the environment.
+
+For heat transfer via air flow, we calculate and report both the sensible and latent heat emissions.
+
+\subsection{Heat emission from building envelope}\label{emission-from-envelope}
```
mjwitte: Minor, these headings should be Title Case

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
+\section{Heat Emissions}\label{heat-emissions}
+
+Heat emissions from buildings are a crucial component influencing climate change and urban microclimate, e.g., urban heat island effect. The heat emissions from a building include heat releases from three levels of building systems: building envelope, zone/space, and HVAC systems. Quantifying heat emissions from buildings and its spatiotemporal distribution have significant implications on urban environment and climate studies. Knowing how much heat is released into atmosphere can help quantify the magnitude of its impact on urban microclimate, e.g., urban heat island effect and climate change. Spatial and temporal distribution of heat emissions can serve to locate hot spots in space and time, therefore useful for prioritizing recourse to mitigate heat emissions from buildings.
```
mjwitte: ... into *the* atmosphere ...

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+This routine calculates heat emissions from buildings and the scope is restricted to report the buildings’ heat directly discharged to their ambient air. As shown in the figure, the emissions include heat releases from three levels of building components: (1) building envelope - exterior surface convection and radiative heat transfer to the ambient air, (2) zone - exhaust air or exfiltration to ambient, (3) HVAC system - relief/exhaust air from AHUs, and (4) HVAC system - heat rejected by air-cooled condensers or central plants, including cooling towers of water-cooled chillers, gas-fired boilers, furnaces and water heaters. It should be noted that any under-ground zone or envelope is excluded, since heat from those spaces is rejected to ground instead of to ambient air.
```
mjwitte: discharged to ~their~ *the* ambient air

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\begin{figure}[hbtp] 
+\centering
+\includegraphics[width=0.9\textwidth, height=0.9\textheight, keepaspectratio=true]{media/heatemission.png}
+\caption{Building released heat compositions \protect \label{fig:building-released-heat-compositions}}
+\end{figure}
+
+The heat emission we calculate and report measures the heat transfer from buildings to their ambient environment, and can have both positive and negative values. The positive values indicate the building injects heat to the environment, while the negative values indicate the building extracts heat from the environment.
+
+For heat transfer via air flow, we calculate and report both the sensible and latent heat emissions.
```
mjwitte: flow, ~we calculate and report~ both the sensible and latent heat emissions are *calculated and reported*.

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\subsection{Heat emission from building envelope}\label{emission-from-envelope}
+
+Exterior surfaces emit heat to ambient air in the form of convection and long-wave radiation, which could be absorbed by particles and gas molecules in air. The internal sources modify the outside surface temperature through envelope conduction, along with incident solar radiation, which triggers the heat exchange with the ambient air. We calculate the surface outside face heat emission rate  as:
```
mjwitte: ~We calculate~ The surface outside face heat emission rate *is calculated* as:

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\begin{figure}[hbtp] 
+\centering
+\includegraphics[width=0.9\textwidth, height=0.9\textheight, keepaspectratio=true]{media/heatemission.png}
+\caption{Building released heat compositions \protect \label{fig:building-released-heat-compositions}}
+\end{figure}
+
+The heat emission we calculate and report measures the heat transfer from buildings to their ambient environment, and can have both positive and negative values. The positive values indicate the building injects heat to the environment, while the negative values indicate the building extracts heat from the environment.
```
mjwitte: The *calculated* heat emissions ~calculate and report~ measure~s~ the . . .

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\begin{equation}  \label{eq:he-1}
+Q_{emission, surf} = Q_{conv, surf} +  Hr * A * (T_{surf} - T_{out})
+\end{equation}
+
+where \(Q_{conv, surf}\) is the surface outside face convection heat gain rate, \(Hr\) is the surface outside face thermal radiation to air heat transfer coefficient, \(A\) is the surface area, \(T_{surf}\) and \(T_{out}\) are the surface and outdoor air temperature.
+
+\subsection{Heat emission from zones}\label{emission-from-zones}
+
+Discharge air carries heat to ambient at the zone level from exfiltration and exhaust fan. Exfiltration is unintended airflow to outside through envelope cracks, as opposite to infiltration. Zones such as restroom, kitchen and laundry room may be equipped with exhaust fans for active venting. These items can be quantified by:
```
mjwitte, line 29: airflow to *the* outside through envelop cracks, as ~opposite~ *opposed* to infiltration.

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\begin{equation}  \label{eq:he-2}
+Q_{exf, zone} = m_{exf}(h_{zone} - h_{out})
+\end{equation}
+
+\begin{equation}  \label{eq:he-3}
+Q_{exh, zone} = m_{exh}(h_{zone} - h_{out})
+\end{equation}
+
+where \(m_{exf}\) is the exfiltration air flow rate, \(m_{exh}\) is the zone exhaust air flow rate from the zone exhaust node, \(h_{zone}\) and \(h_{out}\) are the zone and outdoor air enthalpy. Among these, the sensible parts are:
```
mjwitte: exfiltration air *mass* flow rate
exhaust air *mass* flow rate

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\begin{equation}  \label{eq:he-4}
+Q_{exf, zone, sensible} = m_{exf} * c * (T_{zone} - T_{out})
+\end{equation}
+
+\begin{equation}  \label{eq:he-5}
+Q_{exh, zone, sensible} = m_{exh} * c * (T_{zone} - T_{out})
+\end{equation}
+
+where \(c)\ is the specific heat of air, \(T_{zone}\) and \(T_{out}\) are the zone and outdoor air temperature. And the latent parts are determined by taking the difference between the total and the sensible rate. For each zone, we calculate its exfiltration rate by solving a mass flow balance on the zone including infiltration, ventilation, outdoor air mixing, zone-to-zone mixing, and all of the zone inlet, exhaust, and return nodes.
```
mjwitte: This may be where the math environment build error is coming from:
\\(c)\\ should be \\(c\\)

For each zone, ~we calculate~ its exfiltration rate *is calculate* by solving . . .

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+This part of zone exhaust only includes the exhaust heat at the zone level defined by \textbf{Fan:ZoneExhaust} or \textbf{AirflowNetwork:MultiZone:Component:ZoneExhaustFan}. For each zone, we aggregated this part by searching its exhaust fans. The exhaust heat from Zone HVAC system nodes is included in the next item. Besides, other part of air exchange with outdoors at system node level, balanced along with AirLoopHVAC, is considered in the next item.
```
mjwitte: I would just keep the first sentence here and delete the others.

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\subsection{Heat emission from HVAC systems through relief air}\label{emission-from-HVAC-relief}
+
+HVAC systems relieve heat from system outdoor air relief nodes, and the total and sensible emission rate are calculated as:
+
+\begin{equation}  \label{eq:he-6}
+Q_{exh, sys} = m_{exh}(h_{zone} - h_{out})
```
mjwitte: h_zone should be h_node

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+\end{equation}
+
+\begin{equation}  \label{eq:he-7}
+Q_{exh, sys, sensible} = m_{exh} * c * (T_{zone} - T_{out})
```
mjwitte: T_zone should be T_node

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+\end{equation}
+
+And the latent parts are determined by taking the difference between the total and sensible rate.
+
+The air exchange with outdoor is represented with an \textbf{OutdoorAir:Mixer} object in the \textbf{AirLoopHVAC:OutdoorAirSystem} object. We aggregate and report this part by each \textbf{Controller:OutdoorAir} object.
```
mjwitte: object ~. We aggregate and report this part~ *, aggregated and reported* by each . . .

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+\subsection{Heat rejection from HVAC system equipment}\label{emission-from-HVAC-reject}
+
+HVAC system equipment also directly exchanges heat with ambient air via refrigeration cycle and its energy usage. For example, for gas-fired boilers/heaters, waste heat is exhausted to outside together with combustion byproduct. For space cooling, heat is removed from indoors and dumped to outdoor at the condenser side or corresponding cooling equipment, such as cooling tower. In addition, energy consumed by outdoor equipment will directly dissipate to the ambient air, such as compressors and fans in packaged DX systems. In general, we assume the heat rejection is due to gas-powered combustion and condensers of any type, and we calculated these items based on the system type:
```
mjwitte: Replace late sentence with:
In general, the heat rejection from gas-power combustion and condensers is calculated based on the system type:

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+
+a.	Air-cooled: cooling rate + electric power of condenser fan and compressor
+
+b.	Water-cooled: total heat transfer rate with outdoor air
+
+Detailed calculation of HVAC rejected heat varies with different HVAC object groups and component types, as summarized in Table~\ref{table:emission-from-hvac-components}.
+
+% table 94
+\begin{longtable}[c]{p{4.5in}p{1.5in}}
+\caption{Heat rejection calculation for different HVAC component types \label{table:emission-from-hvac-components}} \tabularnewline
```
mjwitte: Title Case

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+\toprule 
+Object Group & Component & Calculation Methods \tabularnewline
+\midrule
+\endhead
+Coils      & Chilled water cooling coil   & Emit heat in condensing unit defined in Condenser Equipment \tabularnewline
+           & Heating coil (fuel heated)   & Fuel Consumed + Heating Coil Defrost Electric Consumption + Heating Coil Crankcase Heater Electric Consumption – Heating Coil Total Heating Energy \tabularnewline
+           & Hot water/steam heating coil & Fmit heat in plant source defined in Plant Heating and Cooling Equipment \tabularnewline
```
mjwitte: Fmit --> Emit

doc/engineering-reference/src/special-modules-reporting/heat-emissions.tex
```diff
@@ -0,0 +1,124 @@
...
+Water Heaters and Thermal Storage   & Water heater & Water Heater <Fuel Type> Rate - Water Heater Heating Rate
+Zone HVAC Forced Air Units          & Window air conditioner / Packaged terminal air conditioner (PTAC) / Packaged terminal heat pump (PTHP); Energy recovery ventilator (ERV); Unit ventilator/heater; Zone outdoor air unit & <unit type> Total Cooling Rate + <unit type> Electric Energy (cooling mode)<br><unit type> Heating Electric Power / Fuel Rate + <unit type> Electric Power - <unit type> Total Heating Rate (heating mode)
+                                    & Water-to-air heat pump & Emit heat in condensing unit defined in Condenser Equipment
+                                    & Zone evaporative cooler & Emit heat in condensing unit defined in Evaporative Coolers
+                                    & Hybrid Unitary HVAC & System dependent; Calculated according to the equipment specification as above
+\bottomrule
+\end{longtable}
+
+There are several parts of heat rejection that the system are not clear defined, and we do not consider the following part into calculation:
```
mjwitte: There are several types of components for which the heat rejection is not clearly defined. The following are not included in the heat emissions report:

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -291,6 +301,30 @@ \subsubsection{Zone Total Internal Total Heating Energy {[}J{]}}\label{zone-tota
...
 
+\subsubsection{Zone Exfiltration Heat Transfer Rate {[}W{]}}\label{zone-exfiltration-heat-transfer-rate-w}
+
+\subsubsection{Zone Exfiltration Sensible Heat Transfer Rate {[}W{]}}\label{zone-exfiltration-sensible-heat-transfer-rate-w}
+
+\subsubsection{Zone Exfiltration Latent Heat Transfer Rate {[}W{]}}\label{zone-exfiltration-latent-heat-transfer-rate-w}
+
+These output variables represent the heat emitted to ambient from exfiltration in watts. The exfiltration rate is calculated by solving a mass flow balance on the zone including infiltration, ventilation, outdoor air mixing, zone-to-zone mixing, and all of the zone inlet, exhaust, and return nodes. The latent parts are determined by taking the difference between the total and the sensible rate. The positive values indicate the zone injects heat to the environment, while the negative values indicate the building extracts heat from the environment. (This output variable is only generated if the user has set a computer system environment variable DisplayAdvancedReportVariables equal to "yes".)
```
mjwitte: Watts should be capitalized here an below.
~The~ Positive values . . .
while ~the~ negative values . . . <similar changes in other sections below>

mjwitte: While it's true the DisplayAdvancedReportVariables can be set as an environment variable, it can also be set in the idf with Output:Diagnostics, DisplayAdvancedReportVariables; In some parts of the I/O Ref Advanced Report Variables are grouped at the end of the list with a statement saying the following are available when advance report variables are requested (ref. Output:Diagnostics, DisplayAdvancedReportVariables):

This raises a question - does the table report work correctly if DisplayAdvancedReportVariables is *not* active? Looking at this now, I would suggest dropping the DisplayAdvancedReportVariables option for heat emissions and just always set them up.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -291,6 +301,30 @@ \subsubsection{Zone Total Internal Total Heating Energy {[}J{]}}\label{zone-tota
...
+
+\subsubsection{Zone Exhaust Air Sensible Heat Transfer Rate {[}W{]}}\label{zone-exhaust-sensible-heat-transfer-rate-w}
+
+\subsubsection{Zone Exhaust Air Latent Heat Transfer Rate {[}W{]}}\label{zone-exhaust-latent-heat-transfer-rate-w}
+
+These output variables represent the heat emitted to ambient from exhaust air in watts. The zone exhaust air flow rate is aggregated from the zone exhaust nodes. The positive values indicate the building injects heat to the environment, while the negative values indicate the building extracts heat from the environment. (This output variable is only generated if the user has set a computer system environment variable DisplayAdvancedReportVariables equal to "yes".)
+
+\subsubsection{Site Total Zone Exfiltration Heat Loss {[}W{]}}\label{site-total-exfiltration-heat-transfer-rate-w}
+
+These output variables represent the total amount of heat emitted to ambient from all zones by exfiltration in watts. The positive values indicate the building injects heat to the environment, while the negative values indicate the building extracts heat from the environment.
```
mjwitte: For this one and the next, state if this includes sensible and latent

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -2759,6 +2795,14 @@ \subsubsection{Surface Outside Face Thermal Radiation to Ground Heat Transfer Co
 
 This is the coefficient that describes thermal radiation heat transfer between the outside face and the ground surrounding the surface. It is the value of ``Hr'' in the classic linearized model for thermal radiation Q = Hr * A * (T\_surf -- T\_odb) when applied to the ground. Where T\_surf = Surface Outside Face Temperature, and T\_odb = Site Outdoor Air Drybulb Temperature (used as an approximation for the ground surface temperature). Former Name: Prior to version 7.1, this output was called ``Surface Ext Rad to Ground Coeff.''
 
+\subsubsection{Surface Outside Face Thermal Radiation to Air Heat Transfer Rate {[}W{]}}\label{surface-outside-face-thermal-radiation-to-air-heat-transfer-rate-w}
+
+This is thermal radiation heat transfer rate between the outside face and the air mass surrounding the surface, calculated by the coeffient.
```
mjwitte: I would just delete ~calculated by the coeffient~.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -2759,6 +2795,14 @@ \subsubsection{Surface Outside Face Thermal Radiation to Ground Heat Transfer Co
...
+
+\subsubsection{Surface Outside Face Heat Emission to Air Rate {[}W{]}}\label{surface-outside-face-heat-emission-to-air-heat-transfer-rate-w}
+
+This is total heat transfer rate between the outside face and the air mass surrounding the surface by convection and thermal radiation.
```
mjwitte, line 2804: Is this variable equal to "Surface Outside Face Thermal Radiation to Air Heat" minus "Surface Outside Face Convection Heat Gain Rate"? If so, state that here.

doc/input-output-reference/src/standard-output-reports/output-table-summaryreports.tex
```diff
@@ -425,6 +425,10 @@ \subsubsection{Predefined Annual Summary Reports}\label{predefined-annual-summar
 
 The Life Cycle Cost Report provides the a summary of the information from the Life Cycle Cost calculations.  This report shows the costs and the timing of costs, often called “cash flows,” along with the present value in several different tables. The tabular results show the present value of all current and future costs. The key used to obtain this report is LifeCycleCostReport.
 
+\paragraph{Heat Emissions Report}\label{HeatEmissionsReport}
+
+The Heat Emissions Report provides the a summary of the building heat emission to the ambient calculations. This report shows the heat emissions from buildings by surface heat transfer, zone exhaust, zone relief, HVAC relief, HVAC heat rejection, as well as in total. The key used to obtain this report is HeatEmissionsSummary.
```
mjwitte: provides ~the~ a summary (and please fix this in the life cycle cost report above, thanks)
. . . to the ambient ~calculations~ *air*.
. . . heat rejection~, as well as in total~.

mjwitte (commented):
Doc review complete.

mjwitte:
@XuanLuoLBL The timeout failures can be ignored, they are happening on other branches as well.

testfiles/ReportHeatEmission_RefFSRestaurant.idf
```diff
@@ -0,0 +1,3444 @@
+! RefBldgFullServiceRestaurantNew2004_Chicago.idf
```
mjwitte: ReportHeatEmission_RefFSRestaurant.idf

testfiles/ReportHeatEmission_RefFSRestaurant.idf
```diff
@@ -0,0 +1,3444 @@
...
+!
+! This example file is based on Version 1.3_5.0 and was transitioned to Version 6.0 using the transition utility
+!
+! DOE Commercial Reference Building
+! Full Service Restaurant, new construction 90.1-2004
+! Version 1.3_5.0
+! EnergyPlus Version 6.0
+! ASHRAE Standards 90.1-2004 and 62-1999
+!
+! Description:  Single story, Two zone building.
```
mjwitte, line 13: Mention here that this example file shows the use of the heat emissions report and is based on RefBldgFullServiceRestaurantNew2004_Chicago.idf

testfiles/ReportHeatEmission_RefFSRestaurant.idf
```diff
@@ -0,0 +1,3444 @@
...
+! Deru, M.; Field, K.; Studer, D.; Benne, K.; Griffith, B.; Torcellini, P;
+! Halverson, M.; Winiarski, D.; Liu, B.; Rosenberg, M.; Huang, J.;
+! Yazdanian, M.; Crawley, D. (2010).
+! U.S. Department of Energy Commercial Reference Building Models of the National Building Stock.
+! Washington, DC: U.S. Department of Energy, Energy Efficiency and
+! Renewable Energy, Office of Building Technologies.
+! ***GENERAL SIMULATION PARAMETERS***
+! Number of Zones: 3
+
+  Version,9.0;
```
mjwitte: 9.1;

src/EnergyPlus/SystemReports.cc
```diff
@@ -2284,7 +2284,18 @@ namespace SystemReports {
                                     PrimaryAirSystem(SysIndex).Name);
             }
         }
-
+        SetupOutputVariable("Air System Relief Air Total Heat Loss Energy",
+                            OutputProcessor::Unit::GJ,
```
mjwitte: Just realized that a new unit has been introduced for output variables. J is the standard unit everywhere else, not sure we want to add GJ here. @Myoldmopar ???

mjwitte (commented):
"Annual Heat Emissions Summary" report does not show in the html table of contents. Needs a block of code like [this](https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/src/EnergyPlus/OutputReportTabular.cc#L3770) and [this](https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/src/EnergyPlus/OutputReportTabular.cc#L3832-L3834).

In the example file outputs, the Surface Outside Face Heat Emission to Air Rate is always zero for the window surfaces. If need be, this particular problem can be fixed in a separate PR after the I/O Freeze.

PSZ-AC_2:2:Air System Relief Air Heat Transfer Rate [W](Hourly) is also always zero - is that expected?

src/EnergyPlus/SystemReports.cc
Myoldmopar: I don't have any reason why we need to add another unit here.  This should be joules.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
Myoldmopar: I agree with not using advanced report variables here, just always have it on.  It will make the code and docs simpler.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
xuanluo113: Yes, it is.

src/EnergyPlus/DataHeatBalSurface.hh
xuanluo113: Added clarifications in the docs.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
xuanluo113: I'm removing this but what about https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L1671

should the surface outputs stay when only `DisplayAdvancedReportVariables`, as the other Outside Face vars?

mjwitte: Leave those as-is, only touch variables related to heat emissions

xuanluo113:
> "Annual Heat Emissions Summary" report does not show in the html table of contents. Needs a block of code like [this](https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/src/EnergyPlus/OutputReportTabular.cc#L3770) and [this](https://github.com/NREL/EnergyPlus/blob/452ad34e1c1dd56c687a209ec7fd73d266bb116e/src/EnergyPlus/OutputReportTabular.cc#L3832-L3834).

done.

> In the example file outputs, the Surface Outside Face Heat Emission to Air Rate is always zero for the window surfaces. If need be, this particular problem can be fixed in a separate PR after the I/O Freeze.

will check this later today or tmr.

> PSZ-AC_2:2:Air System Relief Air Heat Transfer Rate [W](Hourly) is also always zero - is that expected?

I checked this and it seems the relief mass flow is mostly zero for PSZ-AC_2 air controller, but not necessarily every step in the iteration (sometimes a very small number). But when reported as sys time step it's always zero, and I have to double check on why.

mjwitte (approved):

mjwitte:
@XuanLuoLBL Looks good now. This will merge after CI finishes with the HybridModel - don't want to move develop before that's done. 
Just one small issue with the table in the Engineering Ref noted in https://github.com/NREL/EnergyPlus/issues/7160#issuecomment-464913822.

mjwitte:
Hybrid Model needs another round, so merging this now.

xuanluo113:
@mjwitte Thanks Mike. 

Related to [here](https://github.com/NREL/EnergyPlus/pull/7040#issuecomment-464266176), I took another look these days and I think the window issue does need another PR. I'll fix the table along with it.

The relief air seems right. I reported the hourly relief air mass flow rate for PSZ-AC_2 air controller and it is also always zero.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4810,263 @@ namespace OutputReportTabular {
...
+
+        // Refrigerated Rack
+        for (iRef = 1; iRef <= NumRefrigeratedRacks; ++iRef) {
+            if (RefrigRack(iRef).CondenserType == AirCooled) {
+                SysTotalHVACRejectHeatLoss += RefrigRack(iRef).RackElecConsumption + RefrigRack(iRef).RackCoolingEnergy;
+            } else if (RefrigRack(iRef).CondenserType == EvapCooled) {
+                SysTotalHVACRejectHeatLoss += RefrigRack(iRef).EvapPumpConsumption + RefrigRack(iRef).BasinHeaterConsumption +
+                                              RefrigRack(iRef).EvapWaterConsumption * RhoWater * H2OHtOfVap_HVAC;
+            } else if (RefrigRack(iRef).CondenserType == WaterCooled) {
+                SysTotalHVACRejectHeatLoss += RefrigRack(iCoil).CondEnergy;
```
jmarrec, line 5050: Should be iRef, not iCoil

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4767,6 +4810,263 @@ namespace OutputReportTabular {
...
+            }
+        }
+
+        // Refrigerated Case - Condenser
+        for (iRef = 1; iRef <= NumRefrigCondensers; ++iRef) {
+            SysTotalHVACRejectHeatLoss += RefrigRack(iCoil).CondEnergy;
```
jmarrec, line 5056: Should be iRef, not iCoil

jmarrec: And should problably be `RefrigeratedCase::Condenser(iRef)` and not RefrigRack.

jmarrec (commented):
Proposed fixes moved to #7192

PR #7045, opened 2018-11-02
Add Resource Properties to Windows Binaries and Cleanup CMake Rules

Myoldmopar:
Pull request overview
---------------------
This turned out to be a chunk of randomness.

- Clean out unused CMake stuff
- Make a bunch of CMake options advanced so that for a typical build, the number of entries shown is minimal
- Add details to binaries so they show up in the right-click-properties-details tab on Windows

Still to do:
- I think there was an issue about clarifying the interaction between build-fortran and build-testing.  I'll need to look into it, but this would be a solid place to fix it.
- Finish up the binary details, adding manifests and rc files along the way.
- Probably add binary details to the EPMacro binary in this PR since we don't build that binary...


Myoldmopar:
@axelstudios OK, I got it working with some combination of the null terminators and a manifest file and some trial and error.  I need to do some generalization of the file (it's got hardwired version numbers because I was getting them blank when I tried to build after the configure step), but it at least does build and show properties.  I also need to fill out the copyright, etc., properly, and then add this to the other tools.

Could I get a favor from you?  Could you take the existing `bin/EPMacro/Windows/EPMacro.exe` file in this branch, apply the details to it using `rcedit`, and then commit the changes back to this branch?  That way the details are in there but good, and we don't have to worry about trying to do them later.

Myoldmopar:
@axelstudios I've got a whole bunch of random cleanups in this branch that I'd like to get it.  I believe you were going to push forward with the rest of the property-details on the other binaries, so I wanted to ask whether you wanted me to wait and get all of them in from here, or if you just wanted to branch off and continue working at your leisure.  I don't care either way as long as this branch drops in soon.

Thanks!

Myoldmopar:
@axelstudios I know you were thinking about adding more binary details to this branch, but it has some cleanup that I want to get in.  I may end up merging it tomorrow morning, but obviously go ahead and push your changes when you have them done and we'll get them in as well.

PR #7055, opened 2018-11-08
Address bugs affecting UTF-8 input files

### LINKED ISSUE #7042: EPjson input file format does not support diacritics
Issue overview
--------------
Energyplus fails silently when running an epJSON file with diacritics ([example file failing with 1 'é', working if replaced by 'e'](https://github.com/NREL/EnergyPlus/files/2534475/N4.epJSON.txt)). It might not be a bug but adding a meaningful error message would be nice, especially considering that diacritics works well in IDF files.

### Details
 - Platform : windows 10
 - Version of EnergyPlus : 8.9.0-eba93e8e1b

### Checklist

### LINKED ISSUE #7043: EnergyPlus Not Supporting UTF-8
Issue overview
--------------
EnergyPlus no longer seems to support UTF-8 encoding as it did before breaking it's use for a lot of international applications. I tested using various UTF-8 characters, such as 模型, and found that from EnergyPlus 8.9 onwards the files don't run. EnergyPlus 8.8 worked, but 8.9 and 9.0.1 did not work.

 - Windows 10 + Mac OS Sierra
 - 8.9 and 9.0.1

mbadams5:
Pull request overview
---------------------
Fixes #7043 
Fixes #7042

This pull request addresses a few bugs related to UTF-8 input files. EnergyPlus now has equivalent behavior as far as I can tell compared to EP8.8 and lower. In fact, it now has stricter requirements than previous EnergyPlus versions since it requires UTF-8 encoding. Previously, EnergyPlus was blissfully unaware of input file encoding and only cared about extended ASCII for upper/lower casing strings.

If an IDF has a different encoding than UTF-8, the current default behavior will replace non-conforming UTF8 bytes with U+FFFD. The three JSON library options (strict, replace, ignore) for dumping a JSON file to disk are as follows:
* [json.exception.type_error.316] invalid UTF-8 byte at index 2: 0xA9 string with replaced invalid characters: "ä�ü" string with ignored invalid characters: "äü"
* string with replaced invalid characters: "ä�ü"
* string with ignored invalid characters: "äü"

This pull request also fixes a bug that now allows the HTML report to show unicode strings instead of the literal characters.
Additionally, this PR provides error messaging where there was none previously (usually during parsing errors).

I created unit tests and added an integration test that have UTF-8 characters in them.

The UTF-8 characters are visible in ESO, EIO, HTML, and SQL files as shown in screenshots below.
ESO
EIO
HTML
SQL

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -597,9 +598,9 @@ namespace GroundHeatExchangers {
 
         // save data for later
         if (!DataSystemVariables::DisableGLHECaching) {
-            myCacheData["Response Factors"]["time"] = myRespFactors->time;
-            myCacheData["Response Factors"]["LNTTS"] = myRespFactors->LNTTS;
-            myCacheData["Response Factors"]["GFNC"] = myRespFactors->GFNC;
+            myCacheData["Response Factors"]["time"] = std::vector<Real64>(myRespFactors->time.begin(), myRespFactors->time.end());
+            myCacheData["Response Factors"]["LNTTS"] = std::vector<Real64>(myRespFactors->LNTTS.begin(), myRespFactors->LNTTS.end());
+            myCacheData["Response Factors"]["GFNC"] = std::vector<Real64>(myRespFactors->GFNC.begin(), myRespFactors->GFNC.end());
```
Myoldmopar, line 603: Random fix here, but good catch.  FYI @mitchute.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -659,14 +659,11 @@ IdfParser::Token IdfParser::next_token(std::string const &idf, size_t &index)
...
     default:
-        static std::string const search_chars("-:.#/\\[]{}_@$%^&*()|+=<>?'\"~");
         static std::string const numeric(".-+0123456789");
         if (numeric.find_first_of(c) != std::string::npos) {
             return Token::NUMBER;
-        } else if (isalnum(c) || (std::string::npos != search_chars.find_first_of(c))) {
-            return Token::STRING;
         }
-        break;
+        return Token::STRING;
```
Myoldmopar, line 666: Nice.  Simpler.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -3577,6 +3577,7 @@ namespace OutputReportTabular {
                                << '\n';
                     tbl_stream << " - EnergyPlus</title>\n";
                     tbl_stream << "</head>\n";
+                    tbl_stream << "<meta http-equiv=\"Content-Type\" content=\"text/html; charset=utf-8\">\n";
```
Myoldmopar, line 3580: :+1:

testfiles/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ ADD_SIMULATION_TEST(IDF_FILE 1ZoneUncontrolledAnnualOutputs.idf EPW_FILE USA_CO_
 ADD_SIMULATION_TEST(IDF_FILE 1ZoneUncontrolledCondFDWithVariableKat24C.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
 ADD_SIMULATION_TEST(IDF_FILE 1ZoneUncontrolledFourAlgorithms.idf EPW_FILE 94810-1956.epw)
 ADD_SIMULATION_TEST(IDF_FILE 1ZoneUncontrolledResLayers.idf EPW_FILE 94810-1956.epw)
+ADD_SIMULATION_TEST(IDF_FILE 1ZoneUncontrolledUTF8.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
```
Myoldmopar, line 19: :+1 :+1:

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -374,6 +374,186 @@ TEST_F(InputProcessorFixture, parse_empty_fields)
...
+      "}}"
+    );
+
+    ASSERT_TRUE(process_idf(idf));
+    json &epJSON = getEpJSON();
+
+    auto const input_file = epJSON.dump(-1, ' ', false, json::error_handler_t::replace);
+
+    EXPECT_EQ( expected, input_file );
+}
```
Myoldmopar, line 555: Awesome.

Myoldmopar (approved):
Super thorough.  It's awesome to see the small fixes here and there that enable this to get back to the way it was prior to 8.9.  Great new unit tests and new example file.  I'm glad to see the SQLite changes were super minimal.

src/EnergyPlus/GroundHeatExchangers.cc
mbadams5: Not entirely random. This gave a build error with a newer version of the JSON library.

Myoldmopar:
Is [this change](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2018-11/d4e72c89b91b70bf6662bed2764514ad6410e3eb-49113e23dd58403413812cf28f1dddfb934bbde5/PlantApplicationsGuide_Example2/x86_64-MacOS-10.9-clang/eplusout.err.diff.html) in the error file output on Mac is expected?

mbadams5:
No, but I don't know if it is related to these changes or not. I'm seeing non-determinism when running that file regarding that line in the error file. I got 6 Veterans Day, 2 Columbus Day, 1 Presidents Day, and 0 Memorial Day.

mbadams5:
@Myoldmopar I see why there is the non-determinism. I am using a [json pointer as the key in a map](https://github.com/NREL/EnergyPlus/blob/92dc6f0b3ccea8d1dbe6fd6f8013c39edcc7aad8/src/EnergyPlus/InputProcessing/InputProcessor.hh#L238), where it is getting added [here](https://github.com/NREL/EnergyPlus/blob/92dc6f0b3ccea8d1dbe6fd6f8013c39edcc7aad8/src/EnergyPlus/InputProcessing/InputProcessor.cc#L208-L209).

I am surprised none of the other platforms have this non-determinism and that it hasn't been seen previously either. There must be a change in the JSON library regarding the JSON pointers.

If we want to address this then I'll need to think about how to fix this since I used a pointer on purpose (for space and speed considerations).

Myoldmopar:
So we have seen sporadic non-determinism on things like this before, but I don't recall if it was every platform or just Mac.  Anyway, I think we _do_ want to address it, so go ahead and take some time to think about it and let me know if you want me to do anything.

Myoldmopar:
EIO diffs are unrelated due to changes in develop.  This is great!  Thanks for the rapid fix, @mbadams5, I'm merging now.

Issue #6301, opened 2017-09-08
VRF Terminal unit flow balance problems


mjwitte:
Issue overview
--------------
1. Found this while investigating error diffs from #6272 VariableRefrigerantFlow_FluidTCtrl_HR_5Zone because it tripped new zone unbalanced flow warnings. Reporting the node flow rates shows that TUn Outlet Node flow is always significantly less than TUn Inlet Node flow.  Also made a version of VariableRefrigerantFlow_5Zone with cycling fan and the inlet outlet node flow rates are all over the place (sometimes more, sometimes less).

For both the VRF:HR and VRF without HR, using VariableTemp control results in flow imablances.  Use ConstantTemp control is ok (this is with cycling fan).  So, it may be only certain combinations of cycling fan and VRF AC control type that are broken.  Ultimately will need to test all combinations.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus Version 8.8.0-ddaa5689de
 - Helpdesk ticket 12306
### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Apparent adjustment to outlet air flow after inlet node air flow is set.

![vrf_fluidtctrl_flowimbalance](https://user-images.githubusercontent.com/5778156/30701173-9c6499d0-9eb7-11e7-9b4c-cdd625a0945f.png)

rraustad:
It looks like it happens here where the next Te/Tc is calculated for the next time step. This routine sets
the air flow rate again for some reason. I don't know enough about the model to make an educated guess. @rongpengzhang can you take a look at this?

![vrf_fluidtctrl_flowimbalancecode](https://user-images.githubusercontent.com/5778156/30709038-db70e022-9ece-11e7-86f1-084ed7853e56.png)

rraustad:
@rongpengzhang do you only need to calculate Te/Tc on the first pass through for each time step? Would calling CalcVRFIUTeTc_FluidTCtrl only when FirstHVACIteration = true be OK? The next time through the TUs would respond differently and cause Te/Tc to change again, but I see nothing in code that forces model iterations until Te/Tc converges? So I don't know what to do here.

rraustad:
[...]

PR #7059, opened 2018-11-14
Fix zone exhaust air flow balance when system OA is limited, also fix chilled beam airflow

### LINKED ISSUE #6492: Unexpected interaction between Fan:ZoneExhaust and Controller:OutdoorAir

### LINKED ISSUE #6574: Problems with Airloop operation with Zone Exhaust Fan and Cooled Beam

mjwitte:
Pull request overview
---------------------
Fixes #6492 - Unexpected interaction between Fan:ZoneExhaust and Controller:OutdoorAir
Fixes #6574 - Problems with Airloop operation with Zone Exhaust Fan and Cooled Beam
This pull request replaces #6575.

### Summary of Issues Fixed
The main problem in both of these issues is that the airloop flow resolution assumed that the Maximum Outdoor Air Flow Rate would always be available as makeup air to replace air removed by exhaust fans. This generally works as long as the outdoor air controller is allowed to increase OA flow up to the maximum. But the use of `Controller:OutdoorAir` "Maximum Fraction of Outdoor Air Schedule Name" can limit the available OA. Both of the defect files set this schedule to zero during unoccupied hours, effectively shutting off any possible outdoor air. The air loop return flow resolution now accounts for the *currently available* maximum outdoor air for each air loop.

Note that OA controller does [increase OA flow to match zone exhaust](https://github.com/NREL/EnergyPlus/blob/a7c9cc14ce05cf09dca6623de97faa8546e2e1bb/src/EnergyPlus/MixedAir.cc#L3664-L3667) if other constraints allow just as it has before. At some point, this could be made an option, but not now.

In addition, missing air distribution unit indices for cooled beam units caused the zone air flow balance to see zero inlet flow and drive the return flow to zero.

### Defect files fixed
**Reviewers - note that updated defect files for both of these are in EnergyPlusDevSupport\DefectFiles\**

**Issue6492_noschedule.idf** (Denver epw): Before this fix, at the far right of the csv output, system "VAV WITH REHEAT" has zero flow on 01/01 from 00:00 to 08:00, and the zone air temperatures are well below the heating setpoint of 15.6C. After this fix, there is flow when the night cycle manager kicks in and the zone air temperatures reach setpoint before the system shuts off.

**Alt4MZVAVwoOAwhenOccupied-AddITESupplyNode** (St. Paul - Minneapolis epw):
See columns R:X in the csv output, on 01/01. Before the fix, there is no flow and all of the dampers are at zero. After the fix, there is flow. And columns H:Q show that the zones are cold before this fix, and at or near setpoint after.

### Diffs review:
**5ZoneAirCooledWithDOASAirLoop**: This has a recirc-only airloop and a DOAS airloop which has not return. By definition the return flow to the recirc-only airloop must equal the supply flow from that loop. In 9.0.1 that's not the case (add more output and compare columns BN (SPACE1-1 IN NODE flow) and GD (SPACE1-1 OUT NODE flow). These should always be equal. With this branch, they are. This file did reveal a false error for DOAS airloops with no OA system. That was fixed.

**PIUAuto**: Has small diffs.

**RefBldgLargeHotelNew2004_Chicago, RefBldgPrimarySchoolNew2004_Chicago, RefBldgSecondarySchoolNew2004_Chicago**: 
These all *had* new zone air balance warnings which was fixed by changing the test files to autosize the min flow fraction in the bathrooms (schools) and the dining room (hotel). This causes diffs in results.

**VariableRefrigerantFlow_FluidTCtrl_HR_5Zone**: Has the same zone unbalanced air flow warnings as before, but the warning content has changed.
*This is a real flow imbalance - there's not even an airloop - see #6301.*

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte:
@rraustad Thanks! for catching the missing line and the typo. Looks like this is working. Will see what CI thinks of it.

mjwitte:
@rraustad Some unexpected diffs to resolve.

mjwitte:
### Diffs review:
*others moved to top comment*

~~DOAToFanCoilInlet~~ No more diffs.

rraustad:
DOAToFanCoilInlet: so there are 2 competing fans here with their own availability schedule? Seems it was more correct before if the fan coil fans didn't start until 8AM. Which should prevail, zone equip or DOAS?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -4013,7 +4013,7 @@ namespace SimAirServingZones {
                                    ScalableSM + "Supply Air Flow Rate [m3/s]",
                                    PrimaryAirSystem(AirLoopNum).DesignVolFlowRate);
                 // Initialize MaxOutAir for DOAS loops with no actual OASys, systems with an OA controller will overwrite this is CalcOAController
-                if (PrimaryAirSystem(AirLoopNum).OASysExists)
+                if (PrimaryAirSystem(AirLoopNum).isAllOA)
```
rraustad, line 4014: Not sure what the difference here is. If it's all OA then isn't OASysExists = true? And if not all OA then OASysExists could be true or false. If false then MaxOutAir = 0 (or there is a DOAS?). So not sure this change has any affect.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4994,11 +4996,11 @@ namespace ZoneEquipmentManager {
...
 
-                    airLoopReturnFrac = DataAirLoop::AirLoopFlow(airLoop).DesReturnFrac;
-                    if (DataAirSystems::PrimaryAirSystem(airLoop).OASysExists) {
-                        // Set return flow as fraction of matching inlet node flow if there is an OA system
+                    airLoopReturnFrac = thisAirLoopFlow.DesReturnFrac;
+                    if (DataAirSystems::PrimaryAirSystem(airLoop).OASysExists && (thisAirLoopFlow.OAFlow > 0.0)) {
+                        // Set return flow as fraction of matching inlet node flow if there is an OA system and OA flow > 0.0
                         returnNodeMassFlow = airLoopReturnFrac * inletMassFlow;
-                        thisZoneEquip.TotAvailAirLoopOA = DataAirLoop::AirLoopFlow(airLoop).MaxOutAir;
+                        thisZoneEquip.TotAvailAirLoopOA += thisAirLoopFlow.OAFlow;
```
rraustad: So if there are 2 zone equipment does each thisZoneEquip.TotAvailAirLoopOA = air loop OA? Wouldn't that somehow double count OA?

rraustad: How is "int airLoop = thisZoneEquip.ReturnNodeAirLoopNum(returnNum);" associated with a specific zone equipment?

src/EnergyPlus/SimAirServingZones.cc
mjwitte: No, isAllOA means the branch inlet node is an OA node, no OA sys, no OA controller, no return. See 5ZoneAirCooledWithDOASAirLoop.idf.

mjwitte:
Re: DOAToFanCoilInlet, the fan coil fans are running at 7am in both before and after.

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: Say there are two airloops serving a zone and both airloops have OA available, and there is an exhuast fan. The intent here is to reduce the return flows proportionally to the amount of OA each system is bringing in. Not sure as I'm writing this if TotAVailAirLoopOA should even be connected to a zone equip, it may just need to be a local value since this function goes zone-by-zone.  Need to review that.

mjwitte:
@Myoldmopar @rraustad This is ready for review. See [top comment](https://github.com/NREL/EnergyPlus/pull/7059#issue-230909397) for discussion of changes, defect files and diffs.

rraustad:
If OA is set to 0 at night and exhaust fan is left on, will this change bring in OA?

mjwitte:
Good question. It should behave the same as before if using Min OA Schedule but should be zero OA if using Max Fraction OA Schedule.

mjwitte:
By same as before I mean if the system is on it would open OA up to min of supply flow or max oa or the exhaust requirement. Probably need a test file to prove this.

src/EnergyPlus/HVACManager.cc
```diff
@@ -3005,6 +3009,30 @@ namespace HVACManager {
         }
     }
 
+    void CheckAirLoopFlowBalance()
+    {
+        // Check for unbalanced airloop
+        if (!isPulseZoneSizing && !DataHeatBalance::ZoneAirMassFlow.EnforceZoneMassBalance && !DataGlobals::WarmupFlag && AirLoopsSimOnce) {
```
rraustad: This is out of your scope but why don't we care about a mass imbalance if we are doing pulse sizing or enforcing mass balance?

src/EnergyPlus/HVACManager.cc
```diff
@@ -3005,6 +3009,30 @@ namespace HVACManager {
...
+            for (int AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
+                auto &thisAirLoopFlow(AirLoopFlow(AirLoopNum));
+                if (!thisAirLoopFlow.FlowError) {
+                    Real64 unbalancedExhaustDelta = thisAirLoopFlow.SupFlow - thisAirLoopFlow.OAFlow - thisAirLoopFlow.SysRetFlow;
```
rraustad, line 3092: Also out of your scope, but this has been bothering me for some time. If a user adds mixing, cross mixing, or ventilation to this zone then exhaust might be balanced and we still tell them there is a problem.

src/EnergyPlus/HVACManager.cc
rraustad: One other comment, put an outside IF block using IF(thisAirLoopFlow.FlowError) so these multiple checks don't have to be evaluated each call.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3683,11 +3684,13 @@ namespace MixedAir {
...
 
         // Apply Maximum Fraction of Outdoor Air Schedule
+        Real64 currentMaxOAMassFlowRate = this->MaxOAMassFlowRate;
         if (this->MaxOAflowSchPtr > 0) {
             Real64 MaxOAflowfracVal = GetCurrentScheduleValue(this->MaxOAflowSchPtr);
             MaxOAflowfracVal = min(max(MaxOAflowfracVal, 0.0), 1.0);
+            currentMaxOAMassFlowRate = min(this->MaxOAMassFlowRate, this->MixMassFlow * MaxOAflowfracVal);
             OutAirMinFrac = min(MaxOAflowfracVal, OutAirMinFrac);
-            this->OAMassFlow = min(this->OAMassFlow, this->MixMassFlow * MaxOAflowfracVal);
+            this->OAMassFlow = min(this->OAMassFlow, currentMaxOAMassFlowRate);
```
rraustad, line 3713: Just talking out loud here. So if OA is limited by the max schedule then it will go to 0 if the user so chooses. This is good, but if the exhaust fan is still running then what happens?

src/EnergyPlus/ZoneEquipmentManager.hh
```diff
@@ -129,7 +129,7 @@ namespace ZoneEquipmentManager {
                                     Optional_int_const EquipPriorityNum = _ // index in PrioritySimOrder for this update
     );
 
-    void CalcZoneMassBalance();
+    void CalcZoneMassBalance(bool const FirstHVACIteration);
```
rraustad, line 132: Where did you use this new bool?

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4994,11 +4977,11 @@ namespace ZoneEquipmentManager {
...
-                        // Set return flow as fraction of matching inlet node flow if there is an OA system
+                    airLoopReturnFrac = thisAirLoopFlow.DesReturnFrac;
+                    if (DataAirSystems::PrimaryAirSystem(airLoop).OASysExists && (thisAirLoopFlow.MaxOutAir > 0.0)) {
+                        // Set return flow as fraction of matching inlet node flow if there is an OA system and available OA flow > 0.0
                         returnNodeMassFlow = airLoopReturnFrac * inletMassFlow;
-                        thisZoneEquip.TotAvailAirLoopOA = DataAirLoop::AirLoopFlow(airLoop).MaxOutAir;
+                        thisZoneEquip.TotAvailAirLoopOA += thisAirLoopFlow.MaxOutAir;
                     } else {
                         // Set return flow to matching inlet node flow
                         returnNodeMassFlow = inletMassFlow;
```
rraustad, line 5043: What happens to unbalanced exhaust flow here?

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4994,11 +4977,11 @@ namespace ZoneEquipmentManager {
                         retNodeData.MassFlowRateMaxAvail = zoneNodeData.MassFlowRateMaxAvail;
                     }
 
-                    airLoopReturnFrac = DataAirLoop::AirLoopFlow(airLoop).DesReturnFrac;
-                    if (DataAirSystems::PrimaryAirSystem(airLoop).OASysExists) {
-                        // Set return flow as fraction of matching inlet node flow if there is an OA system
+                    airLoopReturnFrac = thisAirLoopFlow.DesReturnFrac;
+                    if (DataAirSystems::PrimaryAirSystem(airLoop).OASysExists && (thisAirLoopFlow.MaxOutAir > 0.0)) {
+                        // Set return flow as fraction of matching inlet node flow if there is an OA system and available OA flow > 0.0
                         returnNodeMassFlow = airLoopReturnFrac * inletMassFlow;
```
rraustad, line 5039: I'm not feeling this. The airloopReturnFrac is a constant, If exhaust is present then how does this account for the mass imbalance? This is meant to adjust returns for multiple air loops per zone so how do we account for unbalanced exhaust flow here?

src/EnergyPlus/HVACManager.cc
mjwitte: Good question. This airloop balance check was originally in `ZoneEquipmentManager:::CalcZoneMassBalance` and this if came along for the ride. The only one here that should matter is AirLoopsSimOnce.  If `EnforceZoneMassBalance` is active, then the loop should still be balanced.

src/EnergyPlus/HVACManager.cc
mjwitte: No, this check here is strictly inside the ductwork. It's basically a mass balance on the mixing box after all the pieces parts have finished. Is supply > return + OA?  The other check at the zone level accounts for the other flows.

src/EnergyPlus/MixedAir.cc
mjwitte: Then the return is forced to match the supply minus available OA (if OA is zero, then return=supply). The zone has to make up the exhaust flow with infiltration/mixing/zoneventilation.  If there's not enough of that, then the zone balance warning should trip.

src/EnergyPlus/ZoneEquipmentManager.hh
mjwitte: Here in the [zone balance check](https://github.com/NREL/EnergyPlus/pull/7059/files#diff-e954944c50c5b52c35f6c9f2d1ad3bf5R4845).

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: Nothing yet. The return flows get adjusted in step 2 (or 10 or whatever) [below](https://github.com/NREL/EnergyPlus/pull/7059/files#diff-e954944c50c5b52c35f6c9f2d1ad3bf5L5046).

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: We don't yet. At this point, we're trying to apportion the total return to each airloop.

mjwitte:
@rraustad Removed some of the flags in the `CheckAirBalance` if [here](https://github.com/NREL/EnergyPlus/pull/7059/commits/e9e8901097358304c5c0a08cbff966486654fd0a).

nrel-bot-2:

rraustad:
@mjwitte where do you feel this stands at the moment? I and @EnergyArchmage are interested in these changes.

src/EnergyPlus/HVACManager.cc
```diff
@@ -3005,6 +3009,30 @@ namespace HVACManager {
...
+        if (!DataGlobals::WarmupFlag && AirLoopsSimOnce) {
+            for (int AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
+                auto &thisAirLoopFlow(AirLoopFlow(AirLoopNum));
+                if (!thisAirLoopFlow.FlowError) {
+                    Real64 unbalancedExhaustDelta = thisAirLoopFlow.SupFlow - thisAirLoopFlow.OAFlow - thisAirLoopFlow.SysRetFlow;
+                    if (unbalancedExhaustDelta > SmallMassFlow) {
+                        ShowSevereError("CheckAirLoopFlowBalance: AirLoopHVAC " + DataAirSystems::PrimaryAirSystem(AirLoopNum).Name +
+                            " is unbalanced. Supply is > return plus outdoor air.");
+                        ShowContinueErrorTimeStamp("");
+                        ShowContinueError("  Flows [m3/s at standard density]: Supply=" + General::RoundSigDigits(thisAirLoopFlow.SupFlow * DataEnvironment::StdRhoAir, 6) +
```
rraustad: The warning here says "Flows [m3/s at standard density]", and then does SupFlow *StdRhoAir. But SupFlow (and others) is already in kg/s isn't it?

From hh file:
        Real64 SupFlow;               // supply air flow rate (includes LeakFlow) [kg/s]

mjwitte: @rraustad Good catch - these are all in kg/s, so need to *divide* by `StdRhoAir` to output m3/s.

mjwitte:
OK, that's better, now diffs are back to where they were in December.

rraustad:
Starting review with Alt4MZVAVwoOAwhenOccupied-AddITESupplyNode.idf.
I have questions. 

There is mixing (equal to exhaust fan flow) and no infiltration in the restrooms. The 2 restrooms are modeling as a single zone, GUID16_restrooms.

```
ZoneMixing,
  AirflowTransfer_GUID16_restrooms_2,     !- Name
  GUID16_restrooms,                       !- Zone Name
  Room Exhaust and Room Air Transfer - Available (100%), !- Schedule Name
  Flow/Zone,                              !- Design Flow Rate Calculation Method
  0.0330363210752851,                     !- Design Flow Rate {m3/s}
  ,                                       !- Flow Rate per Zone Floor Area {m3/s-m2}
  ,                                       !- Flow Rate per Person {m3/s-person}
  ,                                       !- Air Changes per Hour {1/hr}
  GUID12_Lobby;                           !- Source Zone Name

ZoneMixing,
  AirflowTransfer_GUID16_restrooms_1,     !- Name
  GUID16_restrooms,                       !- Zone Name
  Room Exhaust and Room Air Transfer - Available (100%), !- Schedule Name
  Flow/Zone,                              !- Design Flow Rate Calculation Method
  0.0330363210752851,                     !- Design Flow Rate {m3/s}
  ,                                       !- Flow Rate per Zone Floor Area {m3/s-m2}
  ,                                       !- Flow Rate per Person {m3/s-person}
  ,                                       !- Air Changes per Hour {1/hr}
  GUID12_Lobby;                           !- Source Zone Name

ZoneInfiltration:DesignFlowRate,
  Infiltration_GUID16_restrooms_Women 112_2, !- Name
  GUID16_restrooms,                       !- Zone or ZoneList Name
  Infiltration - Off (0%) All Hours,      !- Schedule Name
  Flow/Zone,                              !- Design Flow Rate Calculation Method
  0,                                      !- Design Flow Rate {m3/s}
  ,                                       !- Flow per Zone Floor Area {m3/s-m2}
  ,                                       !- Flow per Exterior Surface Area {m3/s-m2}
  ,                                       !- Air Changes per Hour {1/hr}
  1,                                      !- Constant Term Coefficient
  0,                                      !- Temperature Term Coefficient
  0,                                      !- Velocity Term Coefficient
  0;                                      !- Velocity Squared Term Coefficient

ZoneInfiltration:DesignFlowRate,
  Infiltration_GUID16_restrooms_Men 113_1, !- Name
  GUID16_restrooms,                       !- Zone or ZoneList Name
  Infiltration - Off (0%) All Hours,      !- Schedule Name
  Flow/Zone,                              !- Design Flow Rate Calculation Method
  0,                                      !- Design Flow Rate {m3/s}
  ,                                       !- Flow per Zone Floor Area {m3/s-m2}
  ,                                       !- Flow per Exterior Surface Area {m3/s-m2}
  ,                                       !- Air Changes per Hour {1/hr}
  1,                                      !- Constant Term Coefficient
  0,                                      !- Temperature Term Coefficient
  0,                                      !- Velocity Term Coefficient
  0;                                      !- Velocity Squared Term Coefficient

```

OK, so there is mixing all the time that is equal to the zone exhaust fan air flow.

Oddly, the warning for unbalanced flow is reported for the restroom first on 1/14 05:30.

    ** Warning ** In zone GUID16_RESTROOMS there is unbalanced air flow. Load due to induced outdoor air is neglected.
    **   ~~~   **  Environment=RUNPERIOD 1, at Simulation time=01/14 05:15 - 05:30
    **   ~~~   **   Flows [kg/s]: Inlets: 7.865924E-003  Unbalanced exhausts: 7.720463E-002  Returns: 7.865924E-003
    **   ~~~   **   Infiltration: 0.000000  Zone Ventilation: 0.000000  Mixing (incoming): 7.617664E-002
    **   ~~~   **   Imbalance (excess outflow): 1.027984E-003  Total system OA flow (for all airloops serving this zone): 0.000000
    **   ~~~   **   This error will only be reported once per zone.

The zone flow rates for the restroom are shown from the beginning of the simulation at 1/1 1:00.

When there is no outdoor air (grey) the return flow equals the supply flow. Good, the zone flows are balanced. 

1) When OA turns on at 1/2 7:00 the return flow goes to 0. This was the intent but if there is mixing making up for the zone exhaust, then shouldn't the return flow equal supply?
2) If the condition at 1/2 7:00 represents unbalanced zone flow then why did the unbalanced warning print for the first time on 1/14 5:30 ?

image

rraustad:
Plot for 1/14 does not show a reason for the unbalanced exhaust warning?

image

rraustad:
Is the zone inlet node mass flow rate based on current density and mixing based on standard density? Or the fact that the conditions to calculate mass using current density are different for these 2 airstreams which would also wind up with different mass flows for these 2 streams when the input intent is for them to be equal. Either of these could trip the warning.

rraustad:
Looking at the lobby, where the mixing for the restrooms originates, I see a difference between supply and return but am not sure why. There is no infiltration in the lobby. The difference in flow is 0.021 m3/s which differs from the mixing flow to the restrooms. Difference should be 0.066 m3/s?

image

Myoldmopar:
@mjwitte CI results make sense with your description.  I built with develop and this branch and confirmed your comments in the description as well.  I would like to wait until you resolve the comments from @rraustad before I review the code in case there is any other changes required.

mjwitte:
@rraustad You pegged it - ZoneMixing shows a constant value for Zone Mixing Current Density Volume Flow Rate and the mixing mass flow rate varies with conditions of the source (or receiving or both) zone.  While the flow rate at the Fan:ZoneExhaust inlet node shows a constant mass flow rate and System Node Standard Density Volume Flow Rate. On a mass flow basis there *is* an imbalance for some hours, and not for others.

Do we change ZoneMixing (and crossmixing) to use standard density to calculate the mass flow rate? And cause massive diffs.
Do we change Fan:ZoneExhaust to use current density to calculate mass flow rate? And cause massive diffs.
Do we add some slop to this balanced flow check to allow for density impacts (even though technically the mass flow rates are indeed unbalanced)?
Other??

rraustad:
I recall a problem with DX coils where the cfm/ton warnings were tripping on the same issue. Coil mass flow was calculated using std density and the volume used for the warning check was calculated using current density. The fix was to use std density to back calculate volume for warning purposes only. Now here the problem is similar. Per your suggestions, can't do 1) since that was a fix for an issue, can't recall which. I do like 2) since that's the way it should be but that probably won't correct this warning issue since the fan uses this zone's density and mixing uses the "from"? zone density. And for 3) a little slop should be added but not enough to correct this problem.

It would be nice to know what basis the mass flow is calculated for a particular model. Any of those that use std density can simply use that to back calculate volume. If a model uses current density then that density needs to be used to back calculate volume. So 4) Other: How hard would it be to add a density variable to the mixing object so that volume could be back calculated when needed (Volume = mixing.mass / mixing.density)? This, of course, just hides the fact that E+ uses different basis for mass calc's but it sounds easy enough to get past this warning.

mjwitte:
The mixing density may not be available directly, but the mixing current density volume flow rate is. I'll tinker with this and see if the warning can be gone for now without adding too many extra calculations.

rraustad:
The mixing density would be available if you saved it in a structure variable when mass was calculated, mixing.density. I don't particularly like this either but it meets the intent.

mjwitte:
@rraustad Adding a volume flow check gets rid of the unbalanced flow warnings for defect file Alt4MZVAVwoOAwhenOccupied-AddITESupplyNode.idf but doesn't appear to have changed the new warnings in the test suite files. I'll look at those in more detail to see if they are justified. Remember when evaluating this, that none of these files have `ZoneAirMassFlowConservation` active, so return flows will not be adjusted based on infiltration or mixing flows, but these flow are included in the checks that generate the warning.

rraustad:
I never did like that new object (which was created to avoid diff's when correcting something that was broken). I vote to internally turn on ZoneAirMassFlowConservation so we don't thumb our noses at the laws of conservation of mass/energy and eventually deprecate that object.

mjwitte:
RefBldgPrimarySchoolNew2004_Chicago - Bath_ZN_1_FLR_1 has an exhaust fan at 0.2832 m3/s and a VAV box which sizes to a minimum flow of 0.15 and a maximum flow of 0.50. But there is no ZoneMixing object. So, anytime that the VAV box goes below 0.2832, there will be an imbalancein this zone. Could add a ZoneMixing object, or increase the min flow fraction for this zone.

RefBldgLargeHotelNew2004_Chicago - DINING_FLR_6 has an exhaust fan at 1.84352 m3/s (which matches a ZoneMixing object from Dining to the Kitchen), but the supply to dining is VAV box with min flow of 1.3 and max flow of 4.36. So, again the min flow doesn't match the exhaust flow. In this case, the dining supply is supposed to be providing makeup air to the kitchen, so the dining OA requirement should really be the greater of the mixing flow to the kitchen or the dining zone requirement. But Sizing:Zone for Dining has this:
`1.84779, !- Cooling Minimum Air Flow [m3/s]` which is *supposed to* put a limit on the min flow rate - so why is that not working? Because the VAV terminal unit min flow fraction is fixed at 0.3 - it should be autosize for the min flow to get applied. Will fix that - it gets rid of the unbalanced flow warnings. And, of course will cause diffs.

mjwitte:
VariableRefrigerantFlow_FluidTCtrl_HR_5Zone should not have any unbalanced flow warnings, but that's nothing to do with this PR. VRF terminal units don't set the same inlet/outlet flows, see #6301.

mjwitte:
@rraustad @Myoldmopar Final answer (hopefully). See comments at top regarding files with diffs.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4919,28 +4919,41 @@ namespace ZoneEquipmentManager {
                     !DataGlobals::WarmupFlag && !DataGlobals::DoingSizing && !FirstHVACIteration) {
                     if (!thisZoneEquip.FlowError) {
```
rraustad, line 4919: Just a side note here. !DataGlobals::ZoneSizingCalc && !DataGlobals::SysSizingCalc are both set inside ManageSizing() while !DataGlobals::DoingSizing is set outside that function. Could remove ZoneSizingCalc and SysSizingCalc from this IF test.

        DoingSizing = true;
        ManageSizing();

        BeginFullSimFlag = true;
        SimsDone = false;
        if (DoDesDaySim || DoWeathSim || DoHVACSizingSimulation) {
            DoOutputReporting = true;
        }
        DoingSizing = false;

mjwitte: @rraustad Thanks. I will clean that up and check elsewhere.

rraustad: I didn't check but I wonder if isPulseZoneSizing is complete before BeginFullSimFlag = true is set? If so, could reduce this even futher.

rraustad: extern bool isPulseZoneSizing;                 // true during the set of zone sizing calcs that include the "pulse" for the load component report

rraustad: Forget BeginFullSimFlag, it's set and reset multiple times.

tst/EnergyPlus/unit/ZoneEquipmentManager.unit.cc
```diff
@@ -660,21 +671,196 @@ TEST_F(EnergyPlusFixture, ZoneEquipmentManager_CalcZoneMassBalanceTest3)
...
+
+    CalcZoneReturnFlows(ZoneNum, StdTotalReturnMassFlow, FinalTotalReturnMassFlow);
+    EXPECT_EQ(FinalTotalReturnMassFlow, 0.0);
+    EXPECT_EQ(Node(returnNode1).MassFlowRate, 0.0);
+    EXPECT_EQ(Node(returnNode2).MassFlowRate, 0.0);
+    EXPECT_EQ(Node(returnNode3).MassFlowRate, 0.0);
+
+    // Case 2 - send sum of inlet flow back, except system 2 back at 0.9 
+    StdTotalReturnMassFlow = 0.0;
+    FinalTotalReturnMassFlow = 0.0;
```
rraustad: @mjwitte do this only if your making changes. These 2 lines not needed. Same with 2 at line 844.

rraustad:
Confirmed defect file now operates at night when OA flow = 0 (root cause of defect).

image

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: Likewise, `isPulseZoneSizing` is only true inside ManageSizing, so `DoingSizing` covers that.

tst/EnergyPlus/unit/ZoneEquipmentManager.unit.cc
mjwitte: Removed redundant lines in several places in the new test and in an older test (which is where this pattern got copied from . . . )

rraustad:
Added zone inlet (blue), exhaust (green) and return (orange dotted) nodes to defect file and looked at zone volume balance. Return node is now offset from supply by amount of exhaust during the daytime hours (when OA flow exceeds exhaust). When OA node flow = 0 during the night (see above), the inlet and return node flows match. All good! Unit tests also keep this zone balance coordinated in case someone wants to change something in this part of the code. Very detailed warning to highlight the issue. Nice job @mjwitte. However, warning is first reported on 1/2 at 6 AM when imbalance is obvious at 1/1 00:06 AM. That one is a little confusing.

    ** Warning ** In zone THERMAL ZONE: STORY 1 CORE SPACE there is unbalanced air flow. Load due to induced outdoor air is neglected.
    **   ~~~   **  Environment=RUN PERIOD 1, at Simulation time=01/02 06:00 - 06:10
    **   ~~~   **   Flows [m3/s]: Inlets: 0.170361  Unbalanced exhausts: 4.719474E-002  Returns: 0.170361
    **   ~~~   **   Infiltration: 2.051160E-002  Zone Ventilation: 0.000000  Mixing (incoming): 0.000000
    **   ~~~   **   Imbalance (excess outflow): 2.668314E-002  Total system OA flow (for all airloops serving this zone): 0.000000
    **   ~~~   **   This error will only be reported once per zone.

image

mjwitte:
Remember that infiltration also counts and mass flow balance is checked first. The infiltration in this file has this schedule which explains why there's no imbalance until 1/2 at 6am when the infiltration falls to 25%.
```
  Schedule:Day:Interval,
    Office Infil Quarter On Default Schedule,  !- Name
    No,                      !- Interpolate to Timestep
    06:00,                   !- Time 1
    1,                       !- Value Until Time 1
    22:00,                   !- Time 2
    0.25,                    !- Value Until Time 2
    24:00,                   !- Time 3
    1;                       !- Value Until Time 3
```

rraustad:
Sure enough, right at around 6 AM.

image

rraustad (approved):
Reviewed code and unit tests. Ran defect file with comments answered. Results provide expected results.

mjwitte:
@rraustad Thanks.
@Myoldmopar Expecting this last commit to produce identical CI results as before. If not, there's a problem.

rraustad:
@Myoldmopar I've noticed that the i386 CI isn't running the tests. Always comes back with no diff's. Did that comment I had a while ago ever get implemented? Where the folder gets deleted before the tests are ran because CMake has that folder locked? until the log files are written (or it recreates that folder)? That method seems to work fine here and we get identical results for i386 and Win64. I can't recall now where that was in the script.

rraustad:
CI results are identical to previous result as expected. Considering this work complete.

PR #7065, opened 2018-11-16
Add Resource Properties to Windows Binaries (continued)

axelstudios:
Pull request overview
---------------------
Add details to [energyplus.exe, energyplusapi.dll, ExpandObjects.exe] so they show up in the right-click-properties-details tab on Windows:

image
image
image


nrel-bot-3:
@axelstudios @lgentile it has been 28 days since this pull request was last updated.

Myoldmopar:
This is all good.  The test failure is the same UTF8 failure as everywhere, and the new Mac 10.13 compiler warnings are the same as everywhere.  Merging this in.  FYI @axelstudios

PR #7074, opened 2018-11-23
CMake bug fixes including shipping AppGPostProcess

### LINKED ISSUE #4814: Determine whether to remove LTO
# Summary

A placeholder issue for discussing whether or not to remove LTO with @kbenne and @lefticus (and @mbadams5 indirectly since it kinda involves his sqlite branch).

So the point is that @mbadams5 created an updated sqlite branch, but then it went quiet because there was some issue with LTO I think.  I'd like to either fix LTO or remove it so that the sqlite update goes in.  Either way is fine.  

It seems right now LTO is only enabled on Linux and Mac, so this won't affect Windows at all.  I don't have a Mac to test performance on, so if that is of interest based on initial (Linux) results, we'll have to tag another package for Mac with disabled LTO.
# Performance Testing

The main point of LTO is performance, so I'm doing some testing for performance with and without.  I am comparing the following conditions:
- Base: 8.2.10 iteration build
- Mod: 8.3.0 (beta) "Without LTO" tag build

The only core difference between these builds is that I disabled LTO (for Linux) in the mod case.  I then ran a sample of 20 random files, in both design-day/hourly output and annual/monthly output to check the difference in performance.  The random file set includes short running files and large models.

This is for Linux.  And I ran a single thread, without anything else of significance running on the machine while the simulations ran.
## Design Day Results

image
OK, so this might be a little difficult to discern.  Basically, the columns are absolute runtimes in both the 8.2.10-iteration and the 8.3.0-WithoutLTO builds.  For the most part these are both very small.  So I also plotted the spark line with percent worseness when I removed LTO.  Basically, we expect that when I remove LTO, the runtimes will increase, and this line will be positive on the right hand axis.  For a few files, it goes up, but for a few it also goes down.  Negative here means that the simulation was actually faster when I removed LTO.  When the worseness is at its peak of 25%, the runtime actually went from 0.4 to 0.5 seconds, which is right on the edge of E+ runtime reporting.  So it could've been 0.49 to 0.50.  Basically, let's not worry too much about this here.  Let's look at annual runs.
## Annual Results

image
For the annual results, the simulations are almost entirely running faster without LTO.  There are still a couple files that run a little longer, but these are the very short cases that could be more affected by startup IO.  In the longer running models, the runtime decreases.  
# Summary

So, I think we should probably pull LTO back out.  It would make the project quite a bit cleaner, and I'm assuming it would let the sqlite updates get in (which I think have some performance gains of their own).  

Thoughts?

### LINKED ISSUE #6128: AppGPostProcess not installed in PostProcess directory
Issue overview
--------------
The AppGPostProcess utility was not part of the 8.6.0 or 8.7.0 installation. In 8.5.0 and before it was located in the PostProcess directory. On Windows.

### Checklist

### LINKED ISSUE #6833: CMake should not allow integration tests to run without BUILD_FORTRAN activated
Issue overview
--------------
If a new developer does not check the BUILD_FORTRAN configuration option on a new EnergyPlus build environment on windows but does select BUILD_TESTING, the integration tests will attempt to run without indication that the required FORTRAN packages are missing. This results in a confusing situation where the majority of integration tests fail with no indication as to the root cause.

### Details
 - Windows
 - EnergyPlus v8.9

jmarrec:
Pull request overview
---------------------

* Fix #6128: Ship AppGPostProcess as part of installation package (on windows notably)
* Fix #6833: Avoid having integration test all fail when BUILD_FORTRAN is off. The problem was that flag "-r" (=Call ReadVarEso) was always passed. Instead I have implemented the following logic:
      *  Add "-r" flag when DO_REGRESSION_TESTING=ON, regression testing includes parsing the ESO for diffs (so must have ReadVarEso). 
       * Also add "-r" flag for PERFORMANCE_TESTS, in order to keep things level with past tests.
       * For both, check if "BUILD_FORTRAN=ON" before activating, otherwise log an error
* Fix #4814: Investigate LTO status: it's already gone (an example from a time where it existed can be found [here](https://github.com/NREL/EnergyPlus/commit/530bea25a2faf1405227ef2164cc3175af2ec47b#diff-0174a8f69ec7eae20f93a98bd6d9333dR66)), just found an outdated flag (`ENABLE_LTO`) in decent-ci-MacOS that I removed.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@jmarrec the fix for AppGPostProcess and the LTO clean out are great.  Disabling the regression tests if Fortran is not built is OK too, but there's a use case missing.  I frequently build with Fortran ON, but not doing performance or regression testing, and I run example files to inspect output.  With these changes, now I will no longer automatically have the csv output and I'll have to run read vars manually.  I think for the `-r` flag, you should simplify things back, and just: if build_fortran is on, then add the -r flag.  Otherwise don't.  Does that make sense?

Myoldmopar:
If you can address that change, and pull develop in, this should go back to all green and be able to drop in.

jmarrec:
@Myoldmopar addressed in https://github.com/NREL/EnergyPlus/pull/7074/commits/5f5d16ded1d2a50c83d34e999783bca6ba8a5f0f

cmake/ProjectMacros.cmake
```diff
@@ -82,20 +82,40 @@ function( ADD_SIMULATION_TEST )
...
-   set( ENERGYPLUS_FLAGS "${ADD_SIM_TEST_ENERGYPLUS_FLAGS} -a -r" )
+   set( ENERGYPLUS_FLAGS "${ADD_SIM_TEST_ENERGYPLUS_FLAGS} -a" )
   else()
-   set( ENERGYPLUS_FLAGS "${ADD_SIM_TEST_ENERGYPLUS_FLAGS} -D -r" )
+   set( ENERGYPLUS_FLAGS "${ADD_SIM_TEST_ENERGYPLUS_FLAGS} -D" )
+  endif()
+
+  # Add -r flag if BUILD_FORTRAN is on, regardless of whether we run regression/performance tests
+  # So that it'll produce the CSV output automatically for convenience
+  if (BUILD_FORTRAN)
```
Myoldmopar, line 106: :+1:

Myoldmopar (approved):

Myoldmopar:
Thanks for adding that in there @jmarrec; this is good to go.

PR #7069, opened 2018-11-20
More than 10 components in PlantEquipmentList and CondenserEquipmentList and Fix UniformLoad and Optimal distribution schemes

mjwitte:
Pull request overview
---------------------
Fixes #6871 - Allow more than 10 pieces of equipment in a PlantEquipmentList or CondenserEquipmentList. The code supported this all along, but the idd was limiting these objects to 10 items.

Fixes #7070 - PlantLoop/CondenserLooop "UniformLoad" and "Optimal" load distribution schemes now count only available equipment for initial uniform load distribution or second-pass load for Optimal. Clarify documentation.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add to input rules file for interfaces
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Required documentation updates?
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #7108, opened 2018-12-19
Fix Free Cooling Chiller Failure - Optimal plant loading with zero available equipment

Myoldmopar:
Pull request overview
---------------------
Fixes divide-by-zero problem introduced in #7069 in Optimal plant load distribution when zero pieces of equipment are available.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte (approved):
Fix looks good. Reproduced divide by zero locally on Windows (with all the special gymnastics required) using the idf file. Added a unit test section with numavail=0 and cannot get unit test to fail locally in Windows (using develop+just the unit test). Running in debugger shows that `DivideLoad` becomes inf, but it's not trapped as a failure.

In general, why would numavail=0? Several reasons, including this file where the free cooling HX has disabled the chiller, or could be disabled via EMS, and possibly other reasons. This fix is good, because part 1 has to loop through to figure out that we have zero available, and then we skip parts 2 and 3 in that case.

Will push up the additional unit test lines and let CI fly through this.

cmake/RunSimulation.cmake
```diff
@@ -165,7 +165,7 @@ else()
   set(VALGRIND_COMMAND "")
 endif()
 
-
+# message(${VALGRIND_COMMAND} "${ENERGYPLUS_EXE}" -w "${EPW_PATH}" -d "${OUTPUT_DIR_PATH}" ${ENERGYPLUS_FLAGS_LIST} "${IDF_PATH}")
```
mjwitte, line 168: @Myoldmopar Is this change supposed to be in this PR?

Myoldmopar: I left it in there as a quick diagnostic line.  It's commented out so it doesn't do anything right now.

tst/EnergyPlus/unit/PlantCondLoopOperation.unit.cc
```diff
@@ -521,6 +521,16 @@ TEST_F(DistributePlantLoadTest, DistributePlantLoad_Optimal)
...
+    // 200W - no equipment available
+    DistributePlantLoadTest::ResetLoads();
+    remainingLoopDemand = 0.0;
+    loopDemand = 200.0;
+    thisBranch.Comp(1).Available = false;
+    thisBranch.Comp(2).Available = false;
+    PlantCondLoopOperation::DistributePlantLoad(1, 1, 1, 1, loopDemand, remainingLoopDemand);
+    EXPECT_EQ(thisBranch.Comp(1).MyLoad, 0.0);
+    EXPECT_EQ(thisBranch.Comp(2).MyLoad, 0.0);
+    EXPECT_EQ(remainingLoopDemand, 200.0);
```
Myoldmopar, line 533: :+1:

PR #7123, opened 2019-01-09
AirflowNetwork Diff-Free Refactor

jasondegraw:
Pull request overview
---------------------
Refactor and reorganize AirflowNetwork to make performance improvements easier to implement. The main files are now in a separate subdirectory and the code is partially disentangled from EnergyPlus proper. Additional disentanglement will likely trigger diffs. This pull request is largely diff-free (the only diffs are due to some reordering of elements in memory that are a side effect of using the new input processor).

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


jasondegraw:
@Myoldmopar @lgu1234 At this point, there are still some reorganization-type things to do, but I keep triggering diffs. So I'm going to stop here for now, diffs can wait until the next release. I'm tagging you both as reviewers, but feel free to disagree.

scripts/dev/license-check.py
```diff
@@ -30,7 +30,4 @@
 
 # Check files
 for base in dirs:
```
Myoldmopar, line 32: So are we just not checking on the subdirectories any more?  Maybe it's just a little hard to tell the intent from this small diff.

src/EnergyPlus/AirflowNetwork/CMakeLists.txt
```diff
@@ -0,0 +1,85 @@
...
+
+if (UNIX)
+  ADD_DEFINITIONS("-fPIC")
+endif()
+
+# Create a static library
+# this will be linked statically to create the DLL and also the unit tests
+add_library( airflownetworklib STATIC ${SRCS} ${HDRS} )
+
+target_include_directories( airflownetworklib PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include )
```
Myoldmopar, line 23: Once all the comments are cleaned out of here, this looks good to go.  I like that it is a little standalone static lib now.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -362,7 +358,11 @@ namespace DataAirflowNetwork {
         }
     };
 
-    struct MultizoneCompDetOpeningProp // Large detailed opening component
+    struct AirflowElement
```
Myoldmopar, line 361: At some point, this base class will presumably grow to include (at least) a pure virtual `calculate()` method, right?

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
```diff
@@ -0,0 +1,2994 @@
...
+        return x * x;
+    }
+
+    int Duct::calculate(bool const LFLAG,           // Initialization flag.If = 1, use laminar relationship
+                        Real64 const PDROP,         // Total pressure drop across a component (P1 - P2) [Pa]
+                        int const EP_UNUSED(i),     // Linkage number
+                        const AirProperties &propN, // Node 1 properties
+                        const AirProperties &propM, // Node 2 properties
+                        std::array<Real64, 2> &F,   // Airflow through the component [kg/s]
+                        std::array<Real64, 2> &DF   // Partial derivative:  DF/DP
```
Myoldmopar, line 242: Would it be beneficial for these two arrays to be const refs?

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
```diff
@@ -0,0 +1,2994 @@
...
+        // INTERFACE BLOCK SPECIFICATIONS
+        // na
+
+        // DERIVED TYPE DEFINITIONS
+        // na
+
+        // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
+        //     RE      - Reynolds number.
+        //     FL      - friction factor for laminar flow.
+        //     FT      - friction factor for turbulent flow.
```
Myoldmopar: Feel free to clean out these comments as a part of your diff-free cleanup.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -645,11 +641,14 @@ add_custom_target( GenerateEmbeddedEpJSONSchema
                    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
                    DEPENDS generate_embeddedable_epJSON_schema )
 
+add_subdirectory(AirflowNetwork)
+#message( $<TARGET_PROPERTY:airflownetworklib,INTERFACE_INCLUDE_DIRECTORIES> )
+#include_directories($<TARGET_PROPERTY:airflownetworklib,INTERFACE_INCLUDE_DIRECTORIES>)
```
Myoldmopar, line 646: Maybe a couple more comments to trim before merging, unless you want to leave them in as clues for later refactoring.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -213,6 +213,14 @@ void InputProcessor::initializeMaps()
     }
 }
 
+void InputProcessor::markObjectAsUsed(const std::string &objectType, const std::string &objectName)
+{
+    auto const find_unused = unusedInputs.find({objectType, objectName});
+    if (find_unused != unusedInputs.end()) {
+        unusedInputs.erase(find_unused);
+    }
+}
```
Myoldmopar, line 222: :+1: :+1: :+1: This will be very handy for other work that relies on raw JSON input processing.  @mbadams5 any thoughts or concerns here?

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -1905,9 +1913,11 @@ void InputProcessor::addRecordToOutputVariableStructure(std::string const &KeyVa
 
     auto const found = DataOutputs::OutputVariablesForSimulation.find(VarName);
     if (found == DataOutputs::OutputVariablesForSimulation.end()) {
-        std::unordered_map<std::string, DataOutputs::OutputReportingVariables,
+        std::unordered_map<std::string,
+                           DataOutputs::OutputReportingVariables,
                            UtilityRoutines::case_insensitive_hasher,
-                           UtilityRoutines::case_insensitive_comparator> data;
+                           UtilityRoutines::case_insensitive_comparator>
+            data;
```
Myoldmopar: Bleck, something got funny with the indentation here.  I don't like the data word just hanging out down there.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4727,7 +4725,8 @@ namespace ZoneEquipmentManager {
 
                 for (NodeNum = 1; NodeNum <= ZoneEquipConfig(ZoneNum).NumExhaustNodes; ++NodeNum) {
 
-                    if (AirflowNetworkNumOfExhFan == 0) TotExhaustAirMassFlowRate += Node(ZoneEquipConfig(ZoneNum).ExhaustNode(NodeNum)).MassFlowRate;
+                    if (AirflowNetwork::AirflowNetworkNumOfExhFan == 0)
+                        TotExhaustAirMassFlowRate += Node(ZoneEquipConfig(ZoneNum).ExhaustNode(NodeNum)).MassFlowRate;
```
Myoldmopar, line 4732: I'm inclined to ask if you'd wrap this in braces while you're making the rest of your changes.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -5537,7 +5532,9 @@ namespace ZoneEquipmentManager {
 
         if (AirFlowFlag != UseSimpleAirFlow) return;
         // AirflowNetwork Multizone field /= SIMPLE
-        if (!(SimulateAirflowNetwork == AirflowNetworkControlSimple || SimulateAirflowNetwork == AirflowNetworkControlSimpleADS)) return;
+        if (!(AirflowNetwork::SimulateAirflowNetwork == AirflowNetwork::AirflowNetworkControlSimple ||
+              AirflowNetwork::SimulateAirflowNetwork == AirflowNetwork::AirflowNetworkControlSimpleADS))
+            return;
```
Myoldmopar, line 5548: Same here about braces.

tst/EnergyPlus/unit/RoomAirflowNetwork.unit.cc
```diff
@@ -207,38 +206,38 @@ TEST_F(RoomAirflowNetworkTest, RAFNTest)
...
+    AirflowNetwork::AirflowNetworkLinkageData(1).NodeNums[0] = 1;
+    AirflowNetwork::AirflowNetworkLinkageData(2).NodeNums[0] = 1;
+    AirflowNetwork::AirflowNetworkLinkageData(3).NodeNums[0] = 1;
+    AirflowNetwork::AirflowNetworkLinkageData(1).NodeNums[1] = 3;
+    AirflowNetwork::AirflowNetworkLinkageData(2).NodeNums[1] = 4;
+    AirflowNetwork::AirflowNetworkLinkageData(3).NodeNums[1] = 2;
+    AirflowNetwork::AirflowNetworkLinkageData(4).NodeNums[0] = 2;
+    AirflowNetwork::AirflowNetworkLinkageData(5).NodeNums[0] = 2;
+    AirflowNetwork::AirflowNetworkLinkageData(4).NodeNums[1] = 5;
+    AirflowNetwork::AirflowNetworkLinkageData(5).NodeNums[1] = 6;
```
Myoldmopar, line 218: Why do these square braces make me so happy?

Myoldmopar (approved):
See my comments for just a few things.  Technically I'm requesting a few changes, but they are almost all just cleaning up commented stuff.  So I'm marking it approved.  This is a really great prefactoring step with no diffs.  The two CI warnings are super acceptable.  I will plan on building this locally once more right before merge to ensure everything is good when merging in develop, but otherwise this should be able to drop in pretty easy.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -1131,38 +1312,38 @@ namespace DataAirflowNetwork {
...
     extern Array1D<IntraZoneNodeProp> IntraZoneNodeData;       // Intra zone data set
     extern Array1D<IntraZoneLinkageProp> IntraZoneLinkageData; // Intra zone linkage adat set
     extern Array1D<DisSysNodeProp> DisSysNodeData;
-    extern Array1D<DisSysCompLeakProp> DisSysCompLeakData;
-    extern Array1D<DisSysCompELRProp> DisSysCompELRData;
-    extern Array1D<DisSysCompDuctProp> DisSysCompDuctData;
-    extern Array1D<DisSysCompDamperProp> DisSysCompDamperData;
-    extern Array1D<DisSysCompCVFProp> DisSysCompCVFData;
-    extern Array1D<DisSysCompDetFanProp> DisSysCompDetFanData;
+    extern Array1D<DuctLeak> DisSysCompLeakData;
```
lgu1234, line 1328: Currently, we use prefix, Multizone and DisSys, to represent which component is used in multizone airflow or distribution airflow calculations. By removing prefix, it is not easy to recognize component usage. Is it possible to use some flag to show component usage?

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Properties.hpp
```diff
@@ -0,0 +1,101 @@
...
+    Real64 airPrandtl(Real64 T, // Temperature in Celsius
+                      Real64 W, // Humidity ratio
+                      Real64 P  // Barometric pressure
+    );
+
+} // namespace AirflowNetwork
+
+} // namespace EnergyPlus
+
+#endif
```
lgu1234, line 101: Although these air properties are used in the AirflowNetwork right now, the properties may be used in other modules in the future. Is it possible to make a global data?

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
```diff
@@ -0,0 +1,2994 @@
...
+                DF[0] = CDM;
+            } else {
+                F[0] = FT;
+                DF[0] = 0.5 * FT / PDROP;
+            }
+        }
+        return 1;
+    }
+
+    int SurfaceCrack::calculate(bool const LFLAG,           // Initialization flag.If = 1, use laminar relationship
```
lgu1234, line 387: Is it essential to have a return value for each calculate function?

src/EnergyPlus/AirflowNetwork/src/Properties.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        Real64 const LowerLimit = -20;
+        Real64 const UpperLimit = 70;
+
+        Real64 const a = 0.02364;
+        Real64 const b = 0.0000754772569209165;
+        Real64 const c = -2.40977632412045e-8;
+
+        if (T < LowerLimit) {
+            ShowWarningMessage("Air temperature out of limits for conductivity calculation");
```
lgu1234: I know this is not your change. I prefer to revise a warning message slightly to add either lower or Upper limit with number.

ShowWarningMessage("Air temperature out of lower limit of -20C for conductivity calculation");

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
```diff
@@ -277,16 +266,20 @@ namespace AirflowNetworkSolver {
...
 
         for (i = 1; i <= NetworkNumOfNodes; ++i) {
-            TZ(i) = AirflowNetworkNodeSimu(i).TZ;
-            WZ(i) = AirflowNetworkNodeSimu(i).WZ;
+            // TZ(i) = AirflowNetworkNodeSimu(i).TZ;
+            // WZ(i) = AirflowNetworkNodeSimu(i).WZ;
             PZ(i) = AirflowNetworkNodeSimu(i).PZ;
+            properties[i].temperature = AirflowNetworkNodeSimu(i).TZ;
+            properties[i].humidityRatio = AirflowNetworkNodeSimu(i).WZ;
+            // properties[i].pressure = AirflowNetworkNodeSimu(i).PZ;
```
lgu1234, line 270: DO you plan to remove PZ(i) = AirflowNetworkNodeSimu(i).PZ; and use properties[i].pressure = AirflowNetworkNodeSimu(i).PZ;?

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
```diff
@@ -321,12 +314,13 @@ namespace AirflowNetworkSolver {
             }
             gio::write(Unit11, Format_900) << 0;
             for (i = 1; i <= NetworkNumOfLinks; ++i) {
-                gio::write(Unit11, Format_910) << i << AirflowNetworkLinkageData(i).NodeNums(1) << AirflowNetworkLinkageData(i).NodeHeights(1)
-                                               << AirflowNetworkLinkageData(i).NodeNums(2) << AirflowNetworkLinkageData(i).NodeHeights(2)
+                gio::write(Unit11, Format_910) << i << AirflowNetworkLinkageData(i).NodeNums[0] << AirflowNetworkLinkageData(i).NodeHeights[0]
```
lgu1234, line 313: Do we have a plan to make array index starting from 0, instead of 1?

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
```diff
@@ -381,23 +375,24 @@ namespace AirflowNetworkSolver {
         // na
 
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
-        int i;
-        int n;
 
         // FLOW:
-        for (n = 1; n <= NetworkNumOfNodes; ++n) {
-            ID(n) = n;
+        for (int i = 1; i <= NetworkNumOfNodes; ++i) {
```
lgu1234, line 376: Do we have any benefit to redefine "int i=1..." for every for loop?

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
```diff
@@ -564,21 +562,21 @@ namespace AirflowNetworkSolver {
         }
         if (LIST >= 1) gio::write(Unit21, Format_900);
         for (i = 1; i <= NetworkNumOfLinks; ++i) {
-            n = AirflowNetworkLinkageData(i).NodeNums(1);
-            M = AirflowNetworkLinkageData(i).NodeNums(2);
+            n = AirflowNetworkLinkageData(i).NodeNums[0];
+            m = AirflowNetworkLinkageData(i).NodeNums[1];
             if (LIST >= 1) {
-                gio::write(Unit21, Format_901) << "Flow: " << i << n << M << AirflowNetworkLinkSimu(i).DP << AFLOW(i) << AFLOW2(i);
+                gio::write(Unit21, Format_901) << "Flow: " << i << n << m << AirflowNetworkLinkSimu(i).DP << AFLOW(i) << AFLOW2(i);
```
lgu1234: These line may be deleted, because never used.

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
```diff
@@ -5460,15 +3336,15 @@ namespace AirflowNetworkSolver {
...
                         T += (Htop - Hbot) * BetaT;
                         X += (Htop - Hbot) * BetaXfct * X0;
-                        Rho1 = PsyRhoAirFnPbTdbW(Pbz + P, T, X);
+                        Rho1 = AIRDENSITY(Pbz + P, T, X);
                         BetaRho = (Rho1 - Rho0) / (Htop - Hbot);
                         Dp += psz(Pbz + P, Rho0, BetaRho, Hbot, Htop, G);
                     }
-                    RhoDr = PsyRhoAirFnPbTdbW(Pbz + PZ + Dp, T, X);
-                    Rho = PsyRhoAirFnPbTdbW(Pbz + PZ + Dp, T, X);
+                    RhoDr = AIRDENSITY(Pbz + PZ + Dp, T, X);
```
lgu1234, line 3342: We have an air density function in Psy module. Do we need an additional AIRDENSITY function?

scripts/dev/license-check.py
jasondegraw: No, I've changed to using Python's glob module differently so it should now hit all subdirectories now.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
jasondegraw: Yes, it didn't quite make into this PR.

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
jasondegraw: No, those are two of the three returns out of these functions. I thought about wrapping all three into a struct, but I'm not quite sold on that yet.

src/EnergyPlus/InputProcessing/InputProcessor.cc
jasondegraw: Blame the tyranny of clang-format for that.

src/EnergyPlus/ZoneEquipmentManager.cc
jasondegraw: I can do that.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
jasondegraw: I'm not a huge fan of labeling like that in the naming (the object names get way too long), but I'll try to think of something that'll help without bloating the names.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Properties.hpp
jasondegraw: Yes, that can be done.

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
jasondegraw: Yes, the intent here is to move to using polymorphism, so they all need to have the same signature.

src/EnergyPlus/AirflowNetwork/src/Properties.cpp
jasondegraw: I can make that change.

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
jasondegraw: Yes, at some point. It's a change that goes with other changes that end up causing diffs, so that'll come in a later PR.

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
jasondegraw: Yes, removal of all the Objexx arrays is part of the refactoring plan.

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
jasondegraw: Generally speaking I'd expect the compiler to do the best thing, so there may not be much performance benefit but I wouldn't expect a performance hit either. However, this way the scope where `i` is valid is just in the loop, which protects against unintentional use and clashes.

src/EnergyPlus/AirflowNetwork/src/Solver.cpp
jasondegraw: That's actually a `#define` to `PsyRhoAirFnPbTdbW`, so it is the same thing.

lgu1234 (approved):

src/EnergyPlus/AirflowNetwork/src/Elements.cpp
Myoldmopar: :+1:

Myoldmopar:
@jasondegraw the license-check script is failing.  Running locally with Python 2, which is what the CI machines have, I get:

```
TypeError: glob() got an unexpected keyword argument 'recursive'
```

Looks like the `recursive` keyword was added to Glob in Python 3.5.  
I would suggest changing the code to grab all the files recursively in a Py2/3 way like:

```
files = []
for root, _, filenames in os.walk(source_dir):
    for filename in fnmatch.filter(filenames, '*.cc'):
        files.append(os.path.join(root, filename))
```

jasondegraw:
@Myoldmopar Sigh. One more year.

Myoldmopar:
https://pythonclock.org/

Myoldmopar:
@jasondegraw I fixed the recursive call and verified it works by botching a source file deep in a source subdirectory.  It properly warned.  I'm merging this.  Thanks for this refactor!

Issue #6947, opened 2018-09-07
No section for HybridModel:Zone model in Eng. and Inputoutput Reference documents


Nigusse:
There is no section for HybridModel:Zone in the engineering and inputoutput reference documentation. Originally posted in issue #5997.
- "[ ] I/O Ref section for HybridModel:Zone is not in the pdf.  The source files were added in #5882 but there's no line in overview.tex to input the new file into the document?"

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

jmarrec:
There's one possibly incomplete section in I/O there: https://github.com/jmarrec/EnergyPlus/blob/develop/doc/input-output-reference/src/overview/group-hybrid-model.tex

It doesn't have the Output variables. A cursory look at the code base indicates that one variable should be there: https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/HeatBalanceManager.cc#L5142
"Zone Infiltration Hybrid Model Air Change Rate"

Interestingly enough, if I look at the the example file HybridZoneModel.idf, and add the request for Output:VariableDictionary, it's not in the RDD, while the Flag should be initialized [here](https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/HybridModel.cc#L163) since there are objects that have `Calculate Zone Internal Thermal Mass` or `Calculate Zone Air Infiltration Rate` enabled.


(Also it seems quite weird to me that this object would bear begin/end day and month. It should read that from the RunPeriod)

xuanluo113:
@mjwitte I checked the documentation and it seems the original documents are merged.

https://bigladdersoftware.com/epx/docs/9-0/engineering-reference/hybrid-model.html
https://bigladdersoftware.com/epx/docs/9-0/input-output-reference/group-hybrid-model.html

source code at

https://github.com/NREL/EnergyPlus/blob/develop/doc/input-output-reference/src/overview/group-hybrid-model.tex
[...]

PR #7130, opened 2019-01-18
Fix documentation for HybridModel:Zone model

xuanluo113:
- [x] I/O Ref section for HybridModel:Zone is not in the pdf. The source files were added in #5882 but there's no line in overview.tex 
- [x] I/O Ref doesn't have the output variables for "Zone Infiltration Hybrid Model Air Change Rate"
- [x] For example file HybridZoneModel.idf with the request for Output:VariableDictionary, "Zone Infiltration Hybrid Model Air Change Rate" is not in the RDD, while the Flag should be initialized here since there are objects that have Calculate Zone Internal Thermal Mass or Calculate Zone Air Infiltration Rate enabled.
- [x] This object could read begin/end day and month from the RunPeriod.

### DIFF
--- a/doc/input-output-reference/src/overview.tex
+++ b/doc/input-output-reference/src/overview.tex
@@ -119,3 +119,5 @@ The following descriptions are ``grouped'' by the elements in the IDD (ref: Gett
 \input{src/overview/group-performance-tables}
 
 \input{src/overview/group-fluid-properties}
+
+\input{src/overview/group-hybrid-model}
--- a/doc/input-output-reference/src/overview/group-hybrid-model.tex
+++ b/doc/input-output-reference/src/overview/group-hybrid-model.tex
@@ -1,4 +1,4 @@
-\section{Group - Hybrid Model}\label{group---hybrid-model}
+\section{Group -- Hybrid Model}\label{group-hybrid-model}
 
 This object is to provide a hybrid model feature to enhance the accuracy of the building energy simulation for existing buildings. The hybrid model feature replaces internal thermal mass and infiltration air flow rates with zone air temperature data. The hybrid model feature calculates a temperature capacitance multiplier, equivalent to the zone internal thermal mass and infiltration air change per hour (ACH) based on the inverse modeling of the heat balance algorithms. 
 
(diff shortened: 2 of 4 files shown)


xuanluo113:
#6947

jasondegraw:
@XuanLuoLBL Similar to my question elsewhere, how should this object behave in a multiyear simulation?

xuanluo113:
> @XuanLuoLBL Similar to my question elsewhere, how should this object behave in a multiyear simulation?

@jasondegraw I checked with the original developers. The start/end month are for the period of measured data. The feature only supports the measured period and does not support multiyear simulation. We'll update this in the document.

mjwitte (changes requested):
There is a build warning now:
unused variable ‘EnergyPlus::HybridModel::FlagHybridModel’

The engineering reference section from #5822 is still missing.

The Hybrid Zone Model group now shows in the Input Output Reference, but it is incomplete. It needs the following:
1. Section heading for the object `HybridModel:Zone`.
2. The overview description of the feature is not clear, it could use some additional text. e.g. "for existing buildings *by supplying measured temperature data*."
3. Sample idf object.
4. Outputs section.
5. Most of the field descriptions need work. The Yes/No fields don't mention "No" as an option, and they don't indicate what the default is.
6. The Year/Month field descriptions are all wrong. "... for the annual run period desired." should be "... for the period of the measured data during which the hybrid model will be active." (or something similar)
7. The description of the temperature schedule says to reference a Schedule:File object, but does the code restrict this to Schedule:File or can it be any schedule object? The description here needs to make that clear. And it also needs to mention that all schedules must include data for a full year, so especially when using schedule:file, the dates outside the range of the actual measured data can be zeros, but all of the rows must be there or the data will be read for the wrong dates. Unless the hybrid code is treating this schedule in a special way.

xuanluo113:
@mjwitte 

Thanks for the comments. As the new PR overlaps with this a lot I was thinking merging them together. I integrated the changes in this PR into #7094 in the latest commit and updated the PR review there including the tasks and your comments here.

For the new comments, the only question is that I didn't get how #5822 is related. The others are addressed in the other PR.

If you think it's Okay to do it I'll just close this one.

mjwitte:
Oh - sorry - I meant #5882

xuanluo113:
@mjwitte We completed the engineering reference of #5882 altogether with #7094 at [here](doc/engineering-reference/src/alternative-modeling-processes/hybrid-model.tex). The other listed documentation issues are also addressed at that PR. Would you mind taking another look if I can close this one?
Thanks very much.

mjwitte:
@XuanLuoLBL Yes, this has been covered in other PRs. Closing.

PR #7141, opened 2019-01-29
Cleanup and Refactoring of Plant Structures

Myoldmopar:
Pull request overview
---------------------
This cleans up a few things in the code:
 - Eliminate a repeated allocation made within the iteration loop
 - Initial cleanup of some plant utility functions
 - Change the plant data structure to use scalars for splitter and mixer -- we only allow a max of one each anyway, this should make a tiny dent in the runtime by not having to do lookups in the arrays (which are always allocated to 1...)


Myoldmopar:
Ahhh, another nice no-diff cleanup.  Merging.

Issue #7035, opened 2018-10-26
Heat exchanger effectiveness applied twice to stratified tanks


nmerket:
Issue overview
--------------
For both the source and use side, the heat exchanger effectiveness is [applied to the flow rates](https://github.com/NREL/EnergyPlus/blob/c1a8cbbf47edf480e3c9e680a32d310591a11b3c/src/EnergyPlus/WaterThermalTanks.cc#L8464-L8465) in `CalcNodeMassFlows` and then [saved onto the tank node for the inlets](https://github.com/NREL/EnergyPlus/blob/c1a8cbbf47edf480e3c9e680a32d310591a11b3c/src/EnergyPlus/WaterThermalTanks.cc#L8519-L8520). Back in `CalcWaterThermalTankStratified` they are [applied again](https://github.com/NREL/EnergyPlus/blob/c1a8cbbf47edf480e3c9e680a32d310591a11b3c/src/EnergyPlus/WaterThermalTanks.cc#L8197-L8198) which shouldn't be done.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus: `develop` branch, sha `c1a8cbbf47edf480e3c9e680a32d310591a11b3c`

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

@jmaguire1

nmerket:
Also, the on- and off- cycle parasitic fraction to tank is being applied twice.

- Off cycle fraction [here](https://github.com/NREL/EnergyPlus/blob/48cebab627f272efb7d370b4e8ab747858b5b503/src/EnergyPlus/WaterThermalTanks.cc#L6032) and [here](https://github.com/NREL/EnergyPlus/blob/48cebab627f272efb7d370b4e8ab747858b5b503/src/EnergyPlus/WaterThermalTanks.cc#L8259)
- On cycle fraction [here](https://github.com/NREL/EnergyPlus/blob/48cebab627f272efb7d370b4e8ab747858b5b503/src/EnergyPlus/WaterThermalTanks.cc#L6036) and [here](https://github.com/NREL/EnergyPlus/blob/48cebab627f272efb7d370b4e8ab747858b5b503/src/EnergyPlus/WaterThermalTanks.cc#L8246)

[...]

PR #7146, opened 2019-01-30
Stratified tank refactor for performance

### LINKED ISSUE #6993: Stratified tank model is slow
Issue overview
--------------
This is more or less a continuation of #5039. To recap, the stratified tank model solves the heat balance in the tank using the Forward Euler method to solve the initial value problem and calculate the temperature of each node at the end of the system time step. To maintain stability and some accuracy the internal time step used is 1 second, which causes the model to be quite slow. This is especially pronounced when used in conjunction with a heat pump water heater, which calls the stratified tank multiple times each system time step.

Replacing the Forward Euler method of solution with something more efficient has the potential to speed up simulation times utilizing this model significantly. 

### Checklist

nmerket:
Pull request overview
---------------------
This PR addresses a performance issue with the stratified tank model. The model was reformulated to use a partial analytical solution for the tank system of ODEs which permits taking larger timesteps. It is loosely based on TRNSYS's type 534 stratified tank model and was developed in coordination with [TESS](http://www.tess-inc.com/). It is currently yielding a 5-6x speedup. 

Fixes #6993 #7035

### Work Checklist
- [x] Change `Water Heater Unmet Demand Heat Transfer Rate` output variable to make sense. This will cause diffs.
- [x] Venting
- [x] Decide on a reasonably performant `maxDt`
- [x] Remove old stratified tank function and the environment variable hack
- [x] Engineering Reference documentation.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8410,6 +8429,487 @@ namespace WaterThermalTanks {
...
+                    return Tank.SourceEffectiveness * Tank.SourceMassFlowRate * Cp * (Tank.SourceInletTemp - Tavg[Tank.SourceOutletStratNode - 1]);
+                }
+            }();
+            Esource += Qsource * dt;
+            if (Tank.HeaterOn1) Runtime1 += dt;
+            if (Tank.HeaterOn2) Runtime2 += dt;
+            if (Tank.HeaterOn1 || Tank.HeaterOn2) Runtime += dt;
+            Eheater1 += Qheater1 * dt;
+            Eheater2 += Qheater2 * dt;
+            Eunmet += max(Tank.UseEffectiveness * Tank.UseMassFlowRate * (UseOutletExpectedTemp - Tavg[Tank.UseOutletStratNode - 1]), 0.0) * dt;
```
nmerket: @mjwitte @EnergyArchmage This is being calculated differently that before. That's intentional. This is measuring amount of additional heat needed in the outlet stream to meet setpoint. 

The way it was being calculated before didn't make sense. (For example, it was accounting for the heaters in every node, not just the ones with the heaters). There wasn't anything in the documentation about this output and what its physical definition was. This seemed more defensible.

nmerket:
@mjwitte @EnergyArchmage I'd be greatly indebted to you for your reviews on this PR. Obviously there are some housekeeping items to take care of, but your feedback would be useful at this point. 

I'm going to be digging into whatever diffs once the CI results come back and I'll post here if I see anything concerning. So far I'm seeing that we're getting the same answers--just faster now.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8410,6 +8429,487 @@ namespace WaterThermalTanks {
...
+            e.Temp = e.SavedTemp;
+
+        // Condenser configuration of heat pump water heater
+        const int HPWHCondenserConfig = Tank.HeatPumpNum > 0 ? HPWaterHeater(Tank.HeatPumpNum).TypeNum : 0;
+
+        // Heat rate from the heat pump (W)
+        const Real64 Qheatpump = [&Tank]{
+            if (!(Tank.HeatPumpNum > 0)) {
+                return 0.0;
+            }
```
rraustad: Why isn't this just    if(Tank.HeatPumpNum == 0) return 0.0;   ??

Also, doesn't this above (line 8464) have some sort of protection where there must be a WaterThermalTank instance? Meaning you wouldn't even need this test at 8499 (i.e., how did we get here if there is no tank)?

    WaterThermalTankData &Tank = WaterThermalTank(WaterThermalTankNum);

src/EnergyPlus/WaterThermalTanks.cc
rraustad: Hmm, the load on a WH is not only equal to the local load at the heater node but the load in the rest of the tank (thought, where does mixing occur?). So if a node does not have an element, and the dT is < 0 then isn't there a load on the node with an element? Just because a node doesn't have an element doesn't mean there is no load on nodes that do have an element. It only means Q=0 at that node. After saying this, if there are 2 elements in a tank with 2+ nodes, how is the tank load split between elements (i.e., if dT is used in line 8826 then which element does that Q apply to)?

rraustad: The previous calcs (before these changes) take into account the total load. Take these calcs for instance. Don't these look at a total tank load based on use/source/loss loads? That's not a load only at node that have a heater. So maybe integrating nodes without a heater may be exactly what the model should do.

    Qheat = Qoncycheat = tank_node.OnCycParaLoad * Tank.OnCycParaFracToTank;
    Qusl = Quse + Qsource + Qloss;
    Qneeded = max(-Qusl - Qoncycheat, 0.0);
    Qunmet = max(Qneeded - Qheaters, 0.0);

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8410,6 +8429,487 @@ namespace WaterThermalTanks {
...
+            Eheater1 += Qheater1 * dt;
+            Eheater2 += Qheater2 * dt;
+            Eunmet += max(Tank.UseEffectiveness * Tank.UseMassFlowRate * (UseOutletExpectedTemp - Tavg[Tank.UseOutletStratNode - 1]), 0.0) * dt;
+
+        } // end while TimeRemaining > 0.0
+
+        for (auto &e : Tank.Node) {
+            e.TempAvg = e.TempSum / SecInTimeStep;
+            e.TempSum = 0.0;
+        }
```
rraustad, line 8355: Why is this loop needed? Why not just sum:

    node.TempAvg += (Tavg[i] * dt) / SecInTimeStep;

at line 8795. Just tying to save clock cycles here.

rraustad: And isn't a += here or some average method of multiple nodes needed at 8831 if the loop is kept? Otherwise the last node is the average?

rraustad: Or is e.TempAvg an array?

rraustad: TempAvg is needed after here but this could still be calculated within the loop at 8792 to get rid of this loop.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8410,6 +8429,487 @@ namespace WaterThermalTanks {
...
+            Real64 const HPWHCondenserDeltaT = HPWHCondWaterOutletNode.Temp - HPWHCondWaterInletNode.Temp;
+            Tank.SourceInletTemp = Tank.SourceOutletTemp + HPWHCondenserDeltaT;
+        }
+
+        // Revised source outlet temperature to ensure energy balance. Assumes a constant CP. CR8341/CR8570
+        if (Tank.SourceOutletStratNode > 0) {
+            if (Tank.SourceMassFlowRate > 0.0) {
+                Tank.SourceOutletTemp = Tank.SourceInletTemp * (1.0 - Tank.SourceEffectiveness) + Tank.SourceOutletTemp * Tank.SourceEffectiveness;
+            }
+        }
```
rraustad, line 8395: Can this be moved up into the else if at line 8854?

rraustad: i.e., can line 8870-8872 be moved up to 8857.

rraustad:
My final comment is I would expect 0 diff's here (unless you fix something). I could be wrong but that's the goal.

src/EnergyPlus/WaterThermalTanks.hh
```diff
@@ -768,7 +768,13 @@ namespace WaterThermalTanks {
     void CalcWaterThermalTankStratified(int const WaterThermalTankNum // Water Heater being simulated
     );
 
-    void CalcNodeMassFlows(int const WaterThermalTankNum, // Water Heater being simulated
+    void CalcWaterThermalTankStratifiedOld(int const WaterThermalTankNum // Water Heater being simulated
+    );
+
```
EnergyArchmage: Is the old calc just here for testing and will eventually be removed completely?

src/EnergyPlus/WaterThermalTanks.cc
nmerket: Good point. I'll change expression in the `if` statement to be more clear.

As for your question about the test, yes, we are guaranteed to have a `Tank` here, but it could be a standalone stratified tank object or it can be connected to a HPWH. This test checks for that and if there's no heat pump there's not heat from the heat pump, which is why we need this check.

src/EnergyPlus/WaterThermalTanks.cc
nmerket: @rraustad The snippet you shared above illustrates what I'm talking about. It is being calculated inside a loop of the nodes. `Qusl` has the use, source, and loss heat rates _for the node being considered_, so the `Quse` and `Qsource` are only nonzero for the use and source inlet nodes respectively. So far so good. Same for the on or off cycle heat. However, `Qheaters` is sum of the heat output of both of the heaters during the timestep. Those heaters are assigned to act on specific nodes, but in this case they're being accounted for _on every node_. I can't figure out what that's supposed to physically represent. 

Is there a definition of what this unmet load rate is supposed to mean? If so, I can write code to it. Given the lack of information or definition in the IORef, I defined it in the way that made sense to me after discussing it with our guy who does water heaters. Which is to say that the unmet load of a water heater is the difference between the energy coming out of the water heater during a draw and the energy we expect to come out of the water heater during a draw given the setpoints. If the goal is to match what came out of the old model, I can do that, but I'm just not convinced that it has any meaning whatsoever.

src/EnergyPlus/WaterThermalTanks.cc
nmerket: You're right I could write it that way. I did it like this because initially I expected `dt` to be a variable timestep. In the future we may end up going that way, so I'd like to leave it as is. I don't expect it's adding a noticeable number of clock cycles.

src/EnergyPlus/WaterThermalTanks.cc
nmerket: No, because the source inlet temperature could be changed on line 8865.

src/EnergyPlus/WaterThermalTanks.hh
nmerket: Yes, that's the plan. It's made a nice way to switch between the two as I developed since this is more or less a rewrite of the old model and I wanted to compare results. At one point I thought maybe we could leave both in and let people choose which model they wanted with a flag in the idf object, but eventually talked myself out of it.

nmerket:
I'm feeling pretty pretty good about the changes here. I've been reviewing the diffs from CI and while there are some differences, for the most part they're small. The changes seem to be related to the fact that inversion mixing and control decisions are happening less frequently now (every minute instead of every second). For most simulations this doesn't matter at all. However, for some situations with high heat transfer and flow rates like we have in some of the test files, that can cause a control decision to happen in the next time step causing the results to be out of phase. Those are the diffs we're seeing, but fortunately they're not affecting overall energy use much. 

We're going to do some more testing of other buildings and draw profiles at different time steps this week, but overall I'm going to call the diffs here okay unless I hear otherwise.

nmerket:
@mjwitte @Myoldmopar @EnergyArchmage @rraustad Do any of you still have concerns about this? Do we think it's going to make it in?

mjwitte:
@nmerket There are some unit test failures showing. Please merge in develop to get a clean CI report. I'll review, but this week our priority is for the I/O Freeze - reviewing and merging PRs that had IDD or output format changes.

nmerket:
@mjwitte @Myoldmopar I think this is ready. The regression test diffs are expected. This mostly stems from control decisions being made less often in the tank (every minute vs every second) and mostly causing events to happen in a different time step. In our testing we're seeing reasonable agreement for some higher flow rate cases.

Myoldmopar:
@nmerket are you be able to run a constant-draw simulation case with both develop and this branch?  It would be reassuring to see that a case with no control decisions actually runs exactly the same.  I don't have any issues with the code changes, but this is a good chunk of changes on tank code that is used quite often, so I'm wanting to be extra diligent.

nmerket:
@Myoldmopar A constant draw would still instigate control decisions. At some point during that draw, the heater would come on within a system timestep. In theory that decision to start heating could happen up to 59 seconds later in the new model. For most systems where the flow rates and heat transfer rate from the heater are "normal" the differences are in the noise. The regression tests are showing that. Big numerical differences are more from events happening out of phase (in the next timestep). 

Here's some plots from one of our residential simulations using a heat pump water heater with a pretty typical but smoothed draw profile. (You can't see the blue line because it's underneath the orange one.) We also did another run with some aggressive draw profiles to try to see how bad it could get. The annual energy use difference in the water heater was ~3% for that case. 

image

Myoldmopar:
@nmerket there seems to be some issues here.  I appreciate you showing me the plots, that's more investigation that I would have time for as a reviewer on this.  However, CI is happily playing detective and finding a couple things:

 - At least a couple unit tests are throwing segfaults (index problems?) as revealed in the integration (or just the unit test) debug builds
 - Custom check is showing that Clang-Format needs to be reapplied to the tanks header file

The linux "build failure" will be gone once you pull develop into this branch again.  Let me know if you have any thoughts/concerns about this at this point.

nmerket:
@Myoldmopar I found the issue [here](https://github.com/NREL/EnergyPlus/blob/97ba5d61437b73ee5d01c6c23fb5ae995eecd7b6/src/EnergyPlus/WaterThermalTanks.cc#L8323) when there is no source outlet node set (i.e. `=0`). I'll put in a check for that. 

Strangely enough, this error doesn't show up when I build on my mac, which is concerning to me. I had to build on Ubuntu like the CI machine to hit the problem and then experience the joys of learning how to use `gdb` from the command line. Now I'm wondering whether Visual Studio on Windows would've caught this, but I don't know if I have the patience to set it all up and then wait for it to build in a VM.

nmerket:
@Myoldmopar This should fix the array out of bounds issues. It did on my testing anyway. We'll see what CI turns up.

nmerket:
@Myoldmopar should I be worried about the timeout failures on the integration tests? Or the linux build failure?

Myoldmopar:
The timeout failures are unrelated, as is the Mac unit test issue.  The build failure indicates this branch is out of date because that issue has been resolved in the develop branch.  It looks like CI is actually happy here.  Let me look it over again and see if there's anything remaining.  Do you have changes waiting to be pushed before I give it a final look?

nmerket:
I don't have any changes pending. Have at it.

Myoldmopar:
Oh! That Linux build failure is different!  It's almost certainly a LaTeX error even though it says it is Python related.  I'll build locally and see if it's a straightforward fix, if not I'll point it back to you.

Myoldmopar:
Found it: Aligns don't go inside other equation environments: https://github.com/NREL/EnergyPlus/pull/7146/files#diff-1231a3ce57fe53ce3200a60b36c7fc6aR605

I fixed it and I'm merging in develop and I'm going to push now.  If CI blesses it, we're done.

Myoldmopar:
This looks good.  Diffs look small, on the order of what I would expect for a change like this.  Some warnings cleaned up in the error files too; a side bonus.  Merging this.

PR #7148, opened 2019-01-31
Add support for QtIFW/Mac packaging

Myoldmopar:
Pull request overview
---------------------
This PR makes the changes (on the EnergyPlus repo side) to handle IFW as the new Mac packaging tool.  IFW has already been installed on the Mac CI box, with the binary directory added to the PATH so CMake can find it.  There were a couple minor [changes](https://github.com/Myoldmopar/decent_ci/pull/9) to the Decent CI code that were already merged in to handle this change: 

### Work Checklist
- [ ] Test a package build with this branch.  Should produce a Qt-based installer and be good to go.  Make sure installer does actually contain the entirety of E+ install, not just a 0.0 MiB version...
- [ ] Update CI instructions and such (maybe have a deployment folder in DecentCI itself) with the installation of IFW and adding to PATH


cmake/Install.cmake
```diff
@@ -243,6 +243,8 @@ endif()
 
 if( APPLE )
   set(CPACK_PACKAGE_DEFAULT_LOCATION "/Applications")
+  set(CPACK_PACKAGE_VERSION "${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
```
Myoldmopar, line 244: The Qt based Installer Framework (IFW) has a regex for the version string.  It requires only numerics.  Earlier in this file, we give CPack a version identifier that is like 8.9.0-abcd123456, and it will fail for the alpha characters in the SHA.  This changes the version string to just not have the SHA on it -- only for Mac installers -- and allows the installer to be built.

Myoldmopar:
@rpg777 so the dashboard did manage to build the DMG, and posted it to the [release page](https://github.com/NREL/EnergyPlus/releases/tag/TestMacIFW).  However, the DMG is only 13.5MB...so I think CI probably ran into the same problem we've already seen.  I think there is something we need to be setting (or unsetting) to make IFW work properly.  I would start investigating by looking at what CMake is doing to actually run the binarycreator command, and how it prepares the data and meta folders, etc.  The CMake definition to this binary path it is [here](https://github.com/Kitware/CMake/blob/07cfb18f9d29cfc0588ede928846a03ec5599c48/Modules/CPackIFW.cmake#L400), while the actual CMake C++ code references that variable [here](https://github.com/Kitware/CMake/blob/d75fec5a88f81a8c16cdeab46766e92a14d1d3cf/Source/CPack/IFW/cmCPackIFWGenerator.cxx#L244).  Let me know if you find anything or want to chat about it.

rpg777:
Yup, will look into it ASAP next week. Thanks!

Myoldmopar:
CI results are due to this branch being behind develop.  Everything is good here.  Merging.

cmake/Install.cmake
```diff
@@ -18,8 +18,6 @@ else()
   set(CPACK_PACKAGE_FILE_NAME "${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION}-${CMAKE_SYSTEM_NAME}-${TARGET_ARCH}-${CMAKE_BUILD_TYPE}")
 endif()
 
-set(CPACK_PACKAGING_INSTALL_PREFIX "/${CMAKE_PROJECT_NAME}-${CPACK_PACKAGE_VERSION_MAJOR}-${CPACK_PACKAGE_VERSION_MINOR}-${CPACK_PACKAGE_VERSION_PATCH}")
```
Myoldmopar, line 21: @jmarrec I wonder if this is the culprit for the change in Ubuntu installer structure.  If it is, and we want it back the original way, I think we can simply re-add this line, but it would be good to put it in an `IF ( UNIX AND NOT APPLE )` block with a comment on why it is needed.

If it's not this line, then I'm not really sure what it would have been that changed in this release.  I'd have to `git bisect` it to find when it changes.

rpg777: Entirely possible. Without this change, the file structure that got created on Mac was not compatible with QtIFW. I didn't "see" any ramifications on the other platforms when I did this, but I wasn't looking too closely. =( Sorry.

jmarrec: Pretty sure it's this line indeed.

I noticed the change isn't just on Ubuntu but on all platforms, so it'd make more sense to change to the OpenStudio FindEnergyPlus.cmake file


Just a side note, this change in structure may get pretty confusing for some (very limited numbers of) users: if you just try to extract the `tar.gz` to `/usr/local` with `tar -xzf EnergyPlus-9.1.0-d830726d9e-Darwin-x86_64.tar.gz --strip 1 -C /usr/local/`, it'll no longer end up in `/usr/local/EnergyPlus-9-1-0` but directly in `/usr/local/`

jmarrec: I modified OpenStudio's Cmake stuff to match the new structure, cf https://github.com/NREL/OpenStudio/pull/3427

PR #7154, opened 2019-02-05
Remove unnecessary versions from unit tests

Myoldmopar:
Pull request overview
---------------------
It was pointed out that a number of unit tests had IDF snippets with old versions.  Although the change_version script _does_ try to search through these files to update the versions, it doesn't always find them properly because of the way IDF snippets are entered as delimited strings in the code.

In any case, these versions are not used in the unit test anyway, except for unit tests that may specifically focus on the version itself.

This PR removes all the offending ones I could find.  Based on this command: 

` $ grep -rin 'Version,' * --include="*.cc" | grep -v '9.1' | grep -v 'EXPECT_EQ'`

the only remaining offenders are in the IdfParser unit test and the input processor unit test, which I decided to leave alone.

I built and ran all the unit tests and everything still passes so I think this should drop right in.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/tst/EnergyPlus/unit/AirTerminalSingleDuct.unit.cc
+++ b/tst/EnergyPlus/unit/AirTerminalSingleDuct.unit.cc
@@ -104,7 +104,6 @@ TEST_F(EnergyPlusFixture, AirTerminalSingleDuctCVReheat_GetInputTest)
     bool ErrorsFound(false);
 
     std::string const idf_objects = delimited_string({
-        "Version,8.4;",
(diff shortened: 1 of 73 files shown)


Myoldmopar:
lol, I started up my Decent CI VM to get some help on Linux builds, and it tries to build the custom_check here, however the VM year hadn't updated, and thought it was 2018.  The custom_check failures are invalid and I will now merge this.

Issue #6528, opened 2018-03-14
Reuse shading calculations between runs / implement external shading calculations


aaron-boranian:
Shading calculations are processing intensive and can take a long time for large models with a lot of shading surfaces.

Once all design decisions for the geometry and envelope properties of a building are made, the effect of shading on the building should be identical between simulations (valid assumption?).

If EnergyPlus could reuse these calculations between runs, a lot of processing time could be saved. Of course, EnergyPlus would have to detect that geometry and envelope have changed to force a recalculation.

Migrated from [UserVoice](http://energyplus.uservoice.com/forums/258860/suggestions/14170068)

**NOTE**: This seems to be related to PR #6388 and issue #6482. Would @Myoldmopar, @hongtz68, or @mjwitte like to chime in?

tkdogan:
We can't wait to see this implemented. I think the implementation could be quite simple. Just write out the shaded fractions to a separate file (and allow that to be overwritten by another shading tool). In the IDF you indicate whether shading calc should be performed or whether it can be loaded from the file. That way the user decides and you don't have to implement the "geometry has changed" check.

Myoldmopar:
https://github.com/NREL/EnergyPlus/pull/6388

aaron-boranian:
As stated in the feedback post “Reuse shading calculations between runs”, shading calculations are time consuming for large models with a lot of shading surfaces. In addition, the polygon clipping based algorithms that are used to determine shaded fractions also tends to be unreliable for complex geometry and results in several severe warnings. This limits applicability in shading device design and urban applications or require somewhat awkward workarounds with shading surfaces and transparency schedules such as suggested in this helpserve post (http://energyplus.helpserve.com/Tickets/Ticket/View/9310).

In computer graphics there are OpenGL based methods (such as: http://www.ibpsa.us/simbuild2012/Papers/SB12_TS09b_3_Jones.pdf) that could drastically speed up shading simulations and make them more robust for complex geometry. Similarly, users might wish to use ray-tracing based methods to include inter-reflections in the diffuse contribution.

[...]

Issue #6482, opened 2018-02-19
Revise file handling for imported and exported shading calcs


mjwitte:
Issue overview
--------------
- [ ] Per @Myoldmopar [here](https://github.com/NREL/EnergyPlus/pull/6388/files#r167880308) c++ file streams should be used to import the shading fractions. 

- [ ] Also, fix nan problem with shading file that has only 8760 data rows. SolarShadingTest_Shading_Data.csv was lengthened to 8784 rows to avoid test failures, but things *should* work with an 8760 file. More explanation - using the attached csv file, run SolarShadingTest_ImportedShading using a debug build. If you're on Windows, you need to add nan debugging options described [here](https://github.com/NREL/EnergyPlus/wiki/Debugging-Tips#catching-a-nan). 
[SolarShadingTest_Shading_Data.csv.txt](https://github.com/NREL/EnergyPlus/files/1740111/SolarShadingTest_Shading_Data.csv.txt)

It will crash inside the  minval function when it is called from [here in [ScheduleManager::CheckScheduleValueMinMax](https://github.com/NREL/EnergyPlus/blob/545c14151cf8a7c233209112117fddc39a8e9f16/src/EnergyPlus/ScheduleManager.cc#L3843) which was called from [here](https://github.com/NREL/EnergyPlus/blob/545c14151cf8a7c233209112117fddc39a8e9f16/src/EnergyPlus/ScheduleManager.cc#L2087).

The problem is nans in the data for DaySchedule(388).TSValue for East Side Tree day 366

![image](https://user-images.githubusercontent.com/4991075/36426006-5cffc76e-160e-11e8-8759-12c0934d9a70.png)
![image](https://user-images.githubusercontent.com/4991075/36426059-915fffe2-160e-11e8-8e2a-c9dec7c85799.png)

Looking at the logic for Schedule:File:Shading, the leap year logic seems wrong [here](https://github.com/NREL/EnergyPlus/blob/545c14151cf8a7c233209112117fddc39a8e9f16/src/EnergyPlus/ScheduleManager.cc#L1863-L1867)
If `ScheduleFileShadingLeapYear` is true that means the file has data for Feb29 so it doesn't need to duplicate any data. At first glance, it seems that when `ScheduleFileShadingLeapYear` is false then you need to duplicate the data for Feb28 and shift the rest of the year down one. 

[...]

PR #7157, opened 2019-02-06
Fix shading export and import issue #6482

xuanluo113:
Pull request overview
---------------------
Partially addresses #6482:

- [x] The run scripts and the ::command line interface:: should be modified to save the shading output alongside the idf file, different on all platforms. We need to decide on an output name, maybe call it InputFileNameShading.csv

- [x] The header of eplusshading.csv says “This file contains external shading fractions for all shading surfaces of all zones.” Based on the output in the file, it appears that this should say “This file contains external/sunlit/fractions for/all/surfaces.”

- [x] If the simulation has only design days, the eplusshading.csv only has the surface name headers, no fractions. Is that intended? (Exclude design day)

- [x] The I/O Ref description for ShadowCalculation “Output External Shading Calculation Results” needs updating to match the above items. Also clarify that it only writes data for each day that shadows are calculated, e.g. once every 20 days by default.

- [x] The header row “This file contains external shading fractions for all shading surfaces of all zones.” at the top of eplusshading.csv causes a crash. It appears that the file processing assumes that row 2 is the start of the data? 

- [ ] Additional comments from [@tkdogan](https://github.com/tkdogan) moved here from [6528](https://github.com/NREL/EnergyPlus/issues/6528#issuecomment-449364174) . Reference also [UnmetHours 35984](https://unmethours.com/question/35984/ep901-crashes-when-external-shading-is-imported-from-file/?answer=36055#post-id-36055) 
I am adding a reply to this thread to put in a requests to improve the implementation of this feature:
(1) Currently the ShadowCalculation override only considers direct sunlight. It would be great if in the same way the diffuse portion (skypatch - surface view-factors) could be overridden. We would like to use this feature to exclude complex shading context geometry from the IDF file.
(2) The shaded-fraction output file should match the input format. Currently a comment line is added to it that the input parser does not understand. Further, shadow calculation frequency and time-step influence the output changing the number of rows that are reported. This seems to be inconsistent with the way the input needs to be provided (8769 hourly values).

Additional/remaining items as of Mar 13:
- [x] ~Export/import should give same results.~ Same `Shading Summary` and output variables now.
- [ ] Missing surface warnings - generates lots of warnings for SolarShadingTest_ImportedShading.idf - should change this to a summary warning with DisplayExtraWarnings for details
- [x] ~Expand the warning message for an import file shorter than 8760 to say the import will assume that values start on Jan 01 and any values past the end will be zero.~ Throw error instead if file shorter than 8760 * timestep
- [x] Line endings for Windows in the export file

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - IDD changes n/a


xuanluo113:
@ #6482

xuanluo113:
@mjwitte 

Mike, we reviewed the issues in #6482 and fixed/addressed as follows:

1. In the previous implementation the command line interface was already modified for naming the shading output file as outputFilePrefix + "shading.csv" with the same pattern as other outputs. We document this in the I/O ref.

2. Fixed the design day and updates the I/O ref.

3. The header row is eliminated for importing back. However, in the current implementation, simulation should calculate shading in everyday frequency in order to reuse the previous exported results. We explained this in the I/O as well.

4. For the additional comments, we discussed and found that as diffuse solar radiation is more complicated in importing, in the future it would be nice to have a new feature to import overall solar radiation for each surface.

Thanks for your review.

mjwitte:
Ran a test with 5ZoneAirCooled, exported shading file, then imported it. The file ran and the table results are the same, but there are some warnings:
```
   ** Warning ** ProcessScheduleInput: Schedule:File:Shading="5ZoneAirCooled-exportShadingShading.csv" 8773 records had errors - these values are set to 0.
   **   ~~~   ** Use Output:Diagnostics,DisplayExtraWarnings; to see individual records in error.
```
DisplayExtraWarnings does not give any additional information. Files attached. 
[6482-FixShadingImport.zip](https://github.com/NREL/EnergyPlus/files/2908472/6482-FixShadingImport.zip)

mjwitte:
Also, I added lines to the various run scripts to save the shading.csv output.

mjwitte:
Also, the shading.csv file has unix line endings on Windows. @Myoldmopar ?

mjwitte (changes requested):
See earlier comments https://github.com/NREL/EnergyPlus/pull/7157#issuecomment-467711712 and https://github.com/NREL/EnergyPlus/pull/7157#issuecomment-467712452.

xuanluo113:
@mjwitte Fixed the importing issue by neglecting the first column and avoid using `auto` in the loop. Claimed this in the doc. With the attached idf, it's able to import and then export the same file now.
[5ZoneAirCooled-ImportExportShading.zip](https://github.com/NREL/EnergyPlus/files/2934089/5ZoneAirCooled-ImportExportShading.zip)

mjwitte (approved):
Retested this.
1. Reusing exported shading file does not give the same result (it did before the latest changes).
2. If some columns are deleted (missing surfaces) from the import file there is no warning - should be.
3. If the import file is shorter than 8760 hours there is a warning - regardless of the length of simulation or the shadow calculation frequency. And the warning says it's looking for "8760 hourly values" but it's really looking for 8760*TimestepsPerHour.
4. Question - if the exported file is a full year, but simulation that imports it runs for, say Jun-Jul, will it find the correct rows of shading data?
5. If the exported shading file is written at 4 timesteps per hour, but the importing run uses 10 timesteps per hour, there is no warning and there's no change in the shading summary results - how's that?
6. If shading is exported from a run with TimestepFrequency, the exported shading file is huge - likely being written too often. Oh yeah, the full 96 rows of Jan 1 are being written to the shading file 672 times. The other days are written 96 times.
7. Shading file is still being written with linux line endings on Windows.

It would be fantastic if timestepfrequency shading can be exported correctly.
If the imported shading file does not have the expected number of rows, this should be a severe/fatal error, because the results will be nonsense. If this means limiting some of the options that can be used to produce an exported file (e.g. require Calculation Frequency = 1), then that is ok.

The potential runtime improvement for imported shading is huge - so this will be fantastic and we'll want to promote it somehow to the users.

mjwitte:
@Myoldmopar Please review my [changes to the run scripts](https://github.com/NREL/EnergyPlus/pull/7157/commits/8c27eeb2027c732889bf83aea51a2f50bb22d2c7). And please advise on how to fix the linux line endings in the shading export csv file. I guess that's not a showstopper but would be nice for it to write proper windows line endings.

Myoldmopar:
Run script changes look OK.  I built locally and also saw the line ending issue, but for me it was that on Linux I see CRLF line endings (Windows-style).  I converted them to Linux and ran all the SolarShadingTest* tests (12) and they all passed.  I will push my line ending fix up, but will let @mjwitte merge at will.

mjwitte:
@Myoldmopar Thanks for the line-ending fix. There are other things to address before this merges.

xuanluo113:
@mjwitte 
Sorry for the delay as I ran to lots of fire drills these days. I took a look at this and I think the exported shading format was not designed to be reused right away (it should but I didn’t design it to coordinate these two features). And there were bugs in the exported format. To fix this in short run, I was thinking:

For export:
1. Fix the bug with `TimestepFrequency` shading triggered by writing during calculation. It can be fixed by writing after init heat balance with `BeginDayFlag && !WarmupFlag` (Is that right? I tested locally and it seems to solve this issue)
2. Write hourly shading fraction (instead of subhourly 8760*TimestepsPerHour)
3. Still, it is required that ’Calculation Frequency = 1’ if required to be reused.

For import:
1. Allow only 8760 or 8764 hourly `Schedule:File:Shading` and interpolate with `TimeStep`. 
2. Currently the `Schedule:File:Shading` is reading with the same way as `Schedule:File` - so read as [# surface * annual schedules] and populated each of them to weekly and daily schedules. 
Then each column is assigned with an `ExternalShadingSchInd`
When reading it uses `LookUpScheduleValue(Surface(SurfNum).ExternalShadingSchInd, iHour, iTimeStep)’.
So I guess if the export runs annually and import only runs for Jun-Jul it should be able to capture it - I’ll double check.
3. Warning if column missing (and the fraction of the missing surfaces are assigned to 1.0)

I'm working on this and please let me know if you have other thoughts. Thanks.

mjwitte:
@XuanLuoLBL The current export is writing some sort of values that change with timestep - where do those come from? Any reason they can't be preserved?

xuanluo113:
@mjwitte Two concerns over that:
1. What if the export `timestep` does not equal to the import simulation? Shall we throw error? (e.g. check if the file to import has = 8760* TimestepsPerHour number of rows.
2. Unlike `Shedule:File` I did not consider sub-hourly input (and I should have), so there's no 'Minutes Per Item' as a field. That would require an IDD change...

Or, is there any mechanism in EnergyPlus that would actually read the timestamp and do interpolation?

jasondegraw:
There's work planned to allow multi-year scheduling, so it'd be good if this didn't diverge too far from `Schedule:File`, if at all. If that means this has to get delayed to make IDD changes, that's a better solution than making changes that will need to be backed out to support multi-year calculations.

xuanluo113:
@jasondegraw My plan was to maintain this 8760 hourly import for this bug fixing PR (it's not `Schedule:File` but it's not diverging any further...) to make things right with a quick fix. I'll do another PR with IDD change and include the timestamp in `Schedule:File`, and perhaps coordinate with multi-year scheduling for next release. Wondering if that would work.

jasondegraw:
@XuanLuoLBL That sounds like a good plan to me, though I'm not quite sure what you mean by "include the timestamp in `Schedule:File`". If it's related to making sure the the timestep in the export run is the same as the timestep in the import run, that's going to be hard to do without embedding the timestep in the file itself. That's not a road to go down. This is also a potential problem with `Schedule:File`, so I'd suggest that probably this object needs to have similar fields to what `Schedule:File` has and there will need to be thought about interpolation and the like.

xuanluo113:
@jasondegraw I was saying the "Minute per hour" field is missing in `Schedule:File:Shading` and I will add it with another PR to be consistent with `Schedule:File`. 
But I agree it's harder to make sure the time steps are also consistent. I didn't see `Schedule:File` reading the time stamp either (otherwise why it needs "Minute per hour"). Please correct me if I'm missing something here.

jasondegraw:
@XuanLuoLBL OK, understood.

No, you're not missing anything as far as I know. With `Schedule:File`, it is possible to reinterpret the input data in various unpleasant ways if the inputs are modified "correctly". If you disallow interpolation and require 8760*TimestepsPerHour rows, that might help some but in the long run there'll still be ways around that, especially once we have multiyear scheduling. So I have my doubts that it'll be worth it to spend too much effort trying to lock it down. Better to document the feature and provide appropriate warnings (IMO).

mjwitte (approved):
Tested many combinations and everything looks good. Ran the two related example files. They run fine, but SolarShadingTest_ExternalFraction.idf take for----*ever* to process the Schedule:File objects. But that's not the fault of this PR.

xuanluo113:
@mjwitte That one is loading ~the same amount of shading data separately, column by column.

mjwitte:
Puzzled why only one of the solar shading files showed diffs, noticed that SolarShadingTest_ExternalFraction and SolarShadingTest_ImportedShading didn't have the same result, but they should be the same, just using different approaches to import the shading data. Realized that SolarShadingTest_Shading_Data.csv had changed but the column numbers in SolarShadingTest_ExternalFraction  were not update to match. Fixed SolarShadingTest_ExternalFraction and also added a leading space to the data rows in SolarShadingTest_Shading_Data.csv to match how shading export files are now written. If CI comes back clean, this will merge (finally).

mjwitte:
Diffs (or not diffs):
SolarShadingTest_ImportedShading - before this branch, import shading was not working at all. With the fixes in this branch, import shading is working, but it's not active for design days, and since CI only runs design days, there's no different in the regression output. But there is (as expected ) differences when running an annual simulation locally.

SolarShadingTest_ExternalFraction - several things have changed here, including scheduled shading *was* active for this one for design days, but now it's not. So, there are regression diffs.

SolarShadingTest_ImportedShading  vs SolarShadingTest_ExternalFraction  - These two now give the same result for a full annual simulation. That's how it should be.

These files, however, won't show any regression diffs in the future unless we change something. I'll leave that alone for now, but we may want to force these two to run, say, a couple of weeks on weather file on CI. But not now, because SolarShadingTest_ExternalFraction  is rather slow processing the schedules, regardless of how long the simulation is.

So, once again, assuming CI finishes clean, this will merge.

xuanluo113:
@mjwitte  Thanks, Mike.

PR #7158, opened 2019-02-06
Fix missed configure for new idd and report vars file

Myoldmopar:
Pull request overview
---------------------
This got missed when I incremented version numbers.  Just brings the right files into the build/Products directory.


Myoldmopar:
Not going to waste a couple hours of CI time on the integration build, this is merging.

PR #7163, opened 2019-02-07
Move array and allocate flag to namespace and include in clear_state

Myoldmopar:
Pull request overview
---------------------
@nmerket [noticed](https://github.com/NREL/EnergyPlus/commit/0ac3fd3a0b7260cced18ac4cf318bc2d9ef47a40#commitcomment-32223019) that a static variable was causing issues during unit testing.  This quick fix moves the variable and the allocate flag to the namespace level and includes them in the clear_state function so that it should be all good with unit tests again.

@nmerket if you are able to pull this branch into yours and verify it works well now, that would be helpful. If not, I think it's good to go anyway :upside_down_face: 

FYI @mjwitte

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ Energy+.schema.epJSON.in
 
 # now that we have python, ignore some things
 *.pyc
+venv
 
 # build folder, if the dev chooses to put it there
 #  could also do a wildcard for build-* if we suggest that naming approach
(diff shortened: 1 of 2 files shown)


nmerket (approved):
It works! Thank you for fixing this so quickly. 🥇

Myoldmopar:
CI's happy!  Merging.

PR #7167, opened 2019-02-08
Add ability to Transition to current iteration

jmarrec:
Pull request overview
---------------------

From the `Transition` README, there should be `CreateNewIDFUsingRulesVX_Y_Z.in.f90` file that would allow testing&transitioning from last official version to the current iteration version (from CMAKE_VERSION_XXX variables). It seems this doesn't exist.

So I added the ability to transition to current iteration by adding a CreateNewIDFUsingRules.in.f90 that will be configured by src/Transition/CMakeLists.txt by autofilling the to/from versions.

I started updating the README.md, but I'm not sure what's the best course of action, so welcoming discussion on this. @Myoldmopar 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
I don't really understand.  At any point in the development cycle, once we've setup the transition code for the next version, just building the latest code will always take the latest release files and transition them to the current codebase.  We purposely don't make transition available on a per-iteration basis because we don't want people attempting to use them in their workflows and ending up with halfway-files.  Files should either be on a release version or the latest tip of develop.  I may be a bit confused as to the intent here, though, as it's getting late on a Sunday night here.

jmarrec:
@Myoldmopar I might very well be the one that doesn't understand the current workflow that exists.

My reason for setting this up is that I have at least on two occasions made changes that required Transition, and never found a `CreateNewIDFUsingRulesVX_Y_Z.in.f90` (where X.Y.Z was superior to the latest official major release) in place that would allow me to **write and test** the transition rules at the same time as my PR.

This is meant as a development feature, where you could do transitions from for eg 9.1.0 to 9.1.1 (as soon as we bump the CMAKE VERSIONS variables), then 9.1.0 to 9.1.2, etc.
If you want to prevent the accidental shipping of these "interim" (non major) transition utilities, I could add a CMAKE Advanced variable or something that would need to be set in order to produce this utility. Another option is to only enable it for Debug builds.

Not sure if that helps in clarifying the intent?

nrel-bot-2b:
@jmarrec it has been 7 days since this pull request was last updated.

Myoldmopar:
I understand your intent @jmarrec, but we aren't going to add this right now.  We cognizantly don't increment the iteration version number, and until we have a full discussion about the changing that, I feel like your changes won't really be useful.  The whole team of developers is frequently adding new features that require transition, without encountering any issues, and I think you have just gotten unlucky with your timing in wanting to make transition changes.

For now, I think all we need is for me to make the next version transition files immediately after the release of a version.  That way we always have a vPreviousRelease-to-vNow version of the transition rules source for devs to write in.  That process generally takes less than an hour of dedicated time, and is a cathartic part of my cool-down once a release is out the door.  

I think I'm going to close this PR so CI stops badgering us about it.  If you feel I am still missing the point, then re-open it and let's keep chatting.  I don't want to miss out on useful changes for sure.  And also, you could leave the branch alive so that if we have a discussion and decide to start doing iteration release incrementing more seriously, we would have this ready.

jmarrec:
I must have gotten unlucky with the timings then yes.

If you bump the E+ version to the next minor (eg 9.1.0 to 9.2.0, and not 9.1.1) **and** create the transition file right away, then yeah I guess people can definitely test stuff right away so this is not necessary. If you're ok having this on your todo list right after a release, then I'm totally fine leaving this closed

Issue #7104, opened 2018-12-14
E+ 9.0 - huge change in energy due to Kiva


shorowit:
Issue overview
--------------
In testing a large number of residential buildings between E+ 8.9 and E+ 9.0, we found a huge energy discrepancy (for a small subset of buildings) that seem to be caused by Kiva changes in E+ 9.0. 

IDF files: [kiva.zip](https://github.com/NREL/EnergyPlus/files/2681143/kiva.zip)

We also tried setting the emissivity of the foundation walls to zero to remove radiation heat transfer and found it brought the results more closely in line with expectations. It's hard to believe that radiation heat exchange should have such a large impact.

![image](https://user-images.githubusercontent.com/5861765/50018448-9d0a4380-ff8c-11e8-8cf3-459db4e8a6a6.png)

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

PR #7169, opened 2019-02-12
Fix problem where Kiva foundation walls were improperly weighted

nealkruis:
Pull request overview
---------------------
Addresses #7104. Kiva walls results were getting weighted the same as the floors even though walls are never represented in multiple Kiva instances. As a result, Kiva wall surfaces were effected by a weights that did not add to 1.0 (e.g., the effective radiative temperature would be 0.7 * 290K = 203K = -94F .... Brrr!).

This fix pulls the weighting code out of EnergyPlus so that it is applied consistently and can be easily unit tested.

Diffs expected for cases where there are multiple wall constructions or wall heights.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Myoldmopar (approved):
This seems fine, the changes to E+ itself are minimal, just grabbing a different term from the structure.

Myoldmopar:
CI also looks happy here, with one certainly expected diff.  Merging this one in.

PR #7180, opened 2019-02-19
Remove duplicate call to GetHybridModelZone

mjwitte:
Pull request overview
---------------------
Fix problem with duplicate csv column headers introduced in #7094 .  CI will probably complain about the two newer HybridModel_4Zone test files, because develop will still have the duplicate headers.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
+++ b/src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
@@ -1316,9 +1316,6 @@ namespace ZoneContaminantPredictorCorrector {
 
         } // ContControlledZoneNum
 
-        // Get the Hybrid Model setting inputs
-        GetHybridModelZone();
-
         if (ErrorsFound) {
             ShowFatalError("Errors getting Zone Contaminant Control input data.  Preceding condition(s) cause termination.");
         }


src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
```diff
@@ -1316,9 +1316,6 @@ namespace ZoneContaminantPredictorCorrector {
 
         } // ContControlledZoneNum
 
-        // Get the Hybrid Model setting inputs
-        GetHybridModelZone();
-
```
Myoldmopar, line 1321: Whoa.  This wasn't the fix that I expected.  But OK!

mjwitte: Not 100% certain this is the proper fix. We could add a flag to make sure this only runs once, but I can't see any reason why this needs to be here when it's already getting called [here](https://github.com/NREL/EnergyPlus/blob/7a131682e99198678e82db06e8d60cb4097b267e/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L1777-L1778) in `GetZoneAirSetPoints`. I suppose it's possible that one could do a hybrid model on CO2 without any HVAC? Then would `GetZoneAirSetPoints` never be called? Hmmm.

mjwitte: `GetZoneAirSetPoints` is called from `ManageZoneAirUpdates` which will *always* get called, so I think this is OK.

mjwitte:
Ran the three hybrid zone test files with develop and with this branch. The duplicate columns are gone in this branch, and the results are the same as develop.

src/EnergyPlus/ZoneContaminantPredictorCorrector.cc
Myoldmopar: I'd suggest moving it even higher to ensure that it gets called once, and only once, every simulation.

mjwitte: Any suggestions? A quick scan says the only things that need to be done before this are Zones and ZoneAirMassFlow.EnforceZoneMassBalance but I may have missed something. Well, schedules, too.

xuanluo113: I'm not sure but `GetZoneAirSetPoints` already has a `GetZoneAirStatsInputFlag`, would this flag make sure that `GetHybridModelZone` under this be called only once?

Myoldmopar:
I do not know what happened to Windows there, but I can predict the outcome.  2 regression failures due to the column issue.  We've got to get two others in in short order so I'm merging this.

PR #7188, opened 2019-02-22
Refactor PlantPipingSystemsManager

Myoldmopar:
Pull request overview
---------------------
The plant piping systems manager is one location where the Objexx Arrays were used, but the objects were allocated with non-standard index zero.  The intention was actually to make it more C like at the time, but since the rest of the code uses the standard 1-base with Arrays, this module is now the oddball.  To avoid any complications in further array refactoring, I'm going to clean this manually out and use vectors instead.  I'll try to do some other cleanups along the way as well.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add to input rules file for interfaces
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Required documentation updates?
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #7194, opened 2019-02-26
Fix Plant Piping System Argument Order Problem

Myoldmopar:
Pull request overview
---------------------
Diffs were encountered when refactoring the plant piping system manager.  It took a while to track down but once I did it resolved to no diffs.  The problem was simply an argument order issue, mixing up first hvac and init loop equip.  Diffs aren't actually that big:

image

image

Off by about a degree for the most part.  

Anyway, I wanted to put this here because it's a two-line change that is clearly a correction and will be easy to resolve the diffs.  

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/PlantLoopEquip.cc
```diff
@@ -312,7 +312,7 @@ namespace PlantLoopEquip {
                 sim_component.compPtr->simulate(sim_component_location, FirstHVACIteration, CurLoad, RunFlag);
 
             } else if (EquipTypeNum == TypeOf_PipingSystemPipeCircuit) {
-                SimPipingSystemCircuit(sim_component.Name, sim_component.CompNum, InitLoopEquip, FirstHVACIteration);
+                SimPipingSystemCircuit(sim_component.Name, sim_component.CompNum, FirstHVACIteration, InitLoopEquip);
```
Myoldmopar, line 315: For reference, [here](https://github.com/NREL/EnergyPlus/blob/4bc63b707f3c1ea9291ee41fbf5f8b344998b4a7/src/EnergyPlus/PlantPipingSystemsManager.cc#L163-L166) is the signature of the function:

```
    void SimPipingSystemCircuit(std::string const &EquipName,
                                int &EqNum,
                                bool const EP_UNUSED(FirstHVACIteration),
                                bool const InitLoopEquip)
```

Obviously passing init loop equip first was **wrong**.

Myoldmopar:
The diffs here are exactly what I expected.  And the performance regression is also unfortunately exactly what I expected.  This incorrect behavior was causing the piping system to skip calculations sometimes, which allowed performance to look better, but was simply not correct.  Looking at the diffs, it didn't actually have much impact on the results of the simulation (they are out-of-range diffs, but not actually huge).  So that's a good sign that there is room for improvement in the codebase to clean up the calculations and gain that performance back in a proper way.  I'll be continuing to work on that over in the actual refactor branch #7188

mjwitte (approved):

Issue #6829, opened 2018-07-11
Dry Climate Water Cooling Coil Sizing Calculations Not Consistent for Leaving Humidity Ratio


EnergyArchmage:
Issue overview
--------------
Studying results in the new coil report table shows some inconsistent things with leaving humidity ratio where different parts of the cooling coil sizing in WaterCoils.cc use different leaving humidity ratios..  Usually the leaving humidity ratio is taken from the input in Sizing:System but that can be modified in dry climates, such as Denver.  There are two methods. First, if the mixed air humidity ratio is already lower than the target supply air humidity ratio the target is halved. Second, based on the entering chilled water temperature, there is a lower limit of humidity ratio that is possible and if entering or leaving humidity ratio is too low then the coil is running dry and the leaving humidity ratio adjusted upward to match the entering humidity ratio. These dry coil checks need to be implemented earlier in the cooling coil sizing routines and used consistently when calculating coil capacity and water flows.  

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #7196

PR #6830, opened 2018-07-11
fix cooling coil sizing issue for dry climates

EnergyArchmage:
Pull request overview
---------------------
address issue #6829 

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Code style (parentheses padding, variable names)
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check -- verify this
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add rules to spreadsheet
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Documentation changes in place
 - [ ] Changed docs build successfully
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #7196, opened 2019-02-28
Correct water coil sizing issue for dry climates where coil inlet humidity ratio is very low.

rraustad:
Pull request overview
---------------------
This change addresses issue #6829. Water cooling coil sizing did not correctly account for dry coil conditions. This affects the reported coil outlet temperature and resulting capacity calculated during sizing. These changes will affect design conditions used in dry climates where simulated coil entering air humidity ratio is lower than the supply air humidity ratio used during sizing. Water coil outlet air temperature and/or humidity ratio are also impacted by this change where Sizing:System or Sizing:Zone outlet air conditions are more reliably used. Fan heat is also correctly applied for draw-thru fan configurations.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### Diff's Review:

2ZoneDataCenterHVAC_wEconomizer: Zone Coils. Order change in reports to the eio, expected for all files. The only difference is water coil outlet air humrat (from 0.00847 to 0.0085) to match the design intent.

    Sizing:System,
      West Zone Air System,    !- AirLoop Name
      7.0,                     !- Preheat Design Temperature {C}
      0.0085,                  !- Preheat Design Humidity Ratio {kgWater/kgDryAir}
      11.0,                    !- Precool Design Temperature {C}
      0.0085,                  !- Precool Design Humidity Ratio {kgWater/kgDryAir}
      12.8,                    !- Central Cooling Design Supply Air Temperature {C}

    + Component Sizing Information, Coil:Cooling:Water, WEST ZONE CW COOLING COIL, Design Size Design Outlet Air Humidity Ratio, 8.50000E-003
    - Component Sizing Information, Coil:Cooling:Water, WEST ZONE CW COOLING COIL, Design Size Design Outlet Air Humidity Ratio, 8.47862E-003

5ZoneAirCooled: No change to OA coil sizing. Main coil sizing changed outlet air hum rat which trickled through to increase coil load and water flow rate.

    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 9.40590E-003
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 17383.26927
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 1.03533E-003
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 8.00000E-003
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 18481.06201
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 1.10072E-003

5ZoneCooledBeam; Sizing now picking up draw through fan heat. Cooling capacity increased by Fan Heat = 2080 W (Coil outlet T drops from 21 C to 18.5 C, not shown in eio file).

    - Component Sizing Information, Coil:Cooling:Water:DetailedGeometry, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 5274.44408
    + Component Sizing Information, Coil:Cooling:Water:DetailedGeometry, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 7354.00713

5ZoneDetailedIceStorage: Coil air outlet hum rat changed which trickled down to coil load and water flow rate.

    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 8.47862E-003
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 38661.07478
    - Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 1.59194E-003
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Outlet Air Humidity Ratio, 8.00000E-003
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Coil Load [W], 39823.46420
    + Component Sizing Information, Coil:Cooling:Water, MAIN COOLING COIL 1, Design Size Design Water Flow Rate [m3/s], 1.63980E-003


rraustad:
This branch has the same changes as in #6830. Only minor cosmetic differences exist.
Also, Clang reformatted SurfaceGeometry.cc and Vectors.cc when those files weren't modified??

src/EnergyPlus/VariableSpeedCoils.cc
```diff
@@ -4075,6 +4074,7 @@ namespace VariableSpeedCoils {
                         SupTemp -= FanCoolLoad / (CpAir * rhoair * VolFlowRate);
                     }
                     MixWetBulb = PsyTwbFnTdbWPb(MixTemp, MixHumRat, OutBaroPress, RoutineName);
+                    // need to use OutTemp for air-cooled and RatedInletWaterTemp for water-cooled
                     TotCapTempModFac =
                         CurveValue(VarSpeedCoil(DXCoilNum).MSCCapFTemp(VarSpeedCoil(DXCoilNum).NormSpedLevel), MixWetBulb, RatedInletWaterTemp);
```
rraustad, line 4079: I thought we had changed this line (for zone equipment and next line for air loop equipment) to account for air-cooled versus water-cooled. This needs to be corrected eventually. Need to compare diff's to #6830 before anything else happens.

rraustad:
So this is why it was so difficult to merge develop into #6830. What this chart shows is the flow of sizing for the water cooling coil. I only include sizing call functions and the Data* variables set by those calls to simplify the chart. This is worst case, if zone equip need this info or if supply equip, then its needed.

It appears that innocuous sizing methods were moved up in order to account for dry coil tests.

image

I did find one place (there may be more) where the water flow rate is needed to size TermIU CoolingWaterDesAirOutletTempSizing. I am still looking for other things like this. I would expect that there would be diff's in example files using TermIUs. It probably wouldn't hurt to still call for CoolingCapacitySizing and CoolingWaterFlowSizing at the same place they were before just to set DataCapacityUsedForSizing and DataWaterFlowUsedForSizing in case some methods needed that information. Only air loops need capacity to size water flow so that could be moved up into the if(CurSysNum) bracket. So when TermIU is sized is CurZoneEqNum > 0 or is CurSysNum > 0? I suspect the former. If so, might not even need to call capacity sizing early. @mjwitte please comment.

image

rraustad:
Another place where DataDesInletAirHumRat is needed is to calculate fan heat in CoolingWaterDesAirInletTempSizing. This is becoming circular. I suggest just setting these early on so this information is available even if it's wrong in certain instances. So far it's CoolingAirFlowSizing and CoolingWaterDesAirInletHumRatSizing that need to be called early (and then again later to account for dry coil). Comments?

    FanCoolLoad = DataAirSystems::calcFanDesignHeatGain(DataFanEnumType, DataFanIndex, DataAirFlowUsedForSizing);
    if (DataDesInletAirHumRat > 0.0 && DataAirFlowUsedForSizing > 0.0) {
        CpAir = PsyCpAirFnWTdb(DataDesInletAirHumRat, AutosizeDes);
        fanDeltaT = FanCoolLoad / (CpAir * StdRhoAir * DataAirFlowUsedForSizing);
    }

rraustad:
Or the fan heat calc could changed to read in the humrat just like temp is read in?

                    if (TermUnitIU) {
                        AutosizeDes = FinalZoneSizing(CurZoneEqNum).ZoneTempAtCoolPeak;
                    } else if (ZoneEqFanCoil) {
                        DesMassFlow = FinalZoneSizing(CurZoneEqNum).DesCoolMassFlow;
                        AutosizeDes =
                            setCoolCoilInletTempForZoneEqSizing(setOAFracForZoneEqSizing(DesMassFlow, zoneEqSizing), zoneEqSizing, finalZoneSizing);
                    } else {
                        AutosizeDes = FinalZoneSizing(CurZoneEqNum).DesCoolCoilInTemp;
                    }
                    Real64 fanDeltaT = 0.0;
                    if (DataSizing::DataFanPlacement == DataSizing::zoneFanPlacement::zoneBlowThru) {
                        // calculate fan heat to get fan air-side delta T
                        FanCoolLoad = DataAirSystems::calcFanDesignHeatGain(DataFanEnumType, DataFanIndex, DataAirFlowUsedForSizing);
                        if (DataDesInletAirHumRat > 0.0 && DataAirFlowUsedForSizing > 0.0) {
                            CpAir = PsyCpAirFnWTdb(DataDesInletAirHumRat, AutosizeDes);
                            fanDeltaT = FanCoolLoad / (CpAir * StdRhoAir * DataAirFlowUsedForSizing);
                        }
                    }

rraustad:
OK, my bad. CoolingAirFlowSizing is still called first out the gate. So why then am I seeing changes in inlet air temp and humrat for zone coils? Still looking to reduce these diffs.

rraustad:
Well this looks pretty good. 

2ZoneDataCenterHVAC_wEconomizer - Order change in reports to the eio, expected for all files. The only difference is water coil outlet air humrat (from 0.00847 to 0.0085) to match the design intent.

    Sizing:System,
      West Zone Air System,    !- AirLoop Name
      7.0,                     !- Preheat Design Temperature {C}
      0.0085,                  !- Preheat Design Humidity Ratio {kgWater/kgDryAir}
      11.0,                    !- Precool Design Temperature {C}
      0.0085,                  !- Precool Design Humidity Ratio {kgWater/kgDryAir}
      12.8,                    !- Central Cooling Design Supply Air Temperature {C}

image

5ZoneAirCooled - No change to OA coil sizing. Main air loop coil now looks correctly sized (red) versus incorrect previously (black).

image

image

Will do a few more spot checks and run the defect file.

rraustad:
I compared the coil sizing results in the html and found differences from values reported to the eio. I suspect this is due to lines of code like this that reset the Data* variables (e.g., DataDesInletAirHumRat was reset here to -99999.0). I have started removing those as necessary.

                // DataCapacityUsedForSizing = WaterCoil(CoilNum).DesWaterCoolingCoilRate; // used for warning messages
                // DataDesInletAirTemp = WaterCoil(CoilNum).DesInletAirTemp;
                // DataDesInletAirHumRat = WaterCoil(CoilNum).DesInletAirHumRat;
                // DataDesInletWaterTemp = WaterCoil(CoilNum).DesInletWaterTemp;
                RequestSizing(CompType, CompName, CoolingWaterDesAirOutletHumRatSizing, SizingString, TempSize, bPRINT, RoutineName);

rraustad:
Test using a defect file with Colorado weather shows dry coil condition. Warning isn't worded correctly so will change that.

    ** Warning ** SizeWaterCoil: Coil="SYS0SCC-1", Cooling Coil is running dry for sizing because minimum humidity ratio for water coil air inlet is > design air leaving humidity ratio.
    **   ~~~   **     Wair,in =  6.714843E-003 [KGWATER/KGDRYAIR]
    **   ~~~   **     Wair,out = 6.243859E-003 [KGWATER/KGDRYAIR]
    **   ~~~   **     Inlet chilled water temperature = 6.667 [C]
    **   ~~~   **     Minimum humidity ratio at saturation for inlet chilled water temperature = 7.642475E-003 [KGWATER/KGDRYAIR]
    **   ~~~   ** ....coil leaving humidity ratio will be reset to:
    **   ~~~   **     Wair,out = 6.714843E-003 [KGWATER/KGDRYAIR]

    Component Sizing Information, Coil:Cooling:Water, SYS0SCC-1, Design Size Design Inlet Air Humidity Ratio, 6.71484E-003
    Component Sizing Information, Coil:Cooling:Water, SYS0SCC-1, Design Size Design Outlet Air Humidity Ratio, 6.71484E-003

Warning text changed to:

    ** Warning ** SizeWaterCoil: Coil="SYS0SCC-1", Cooling Coil is running dry for sizing because minimum humidity ratio at saturation for inlet chilled water temperature > design air entering humidity ratio.

rraustad:
Everything I review now appears to make sense. This is ready for review by others.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -6257,10 +6257,10 @@ namespace SurfaceGeometry {
         using DataLoopNode::NodeConnectionType_Inlet;
         using DataLoopNode::NodeType_Air;
         using DataLoopNode::ObjectIsParent;
-        using DataSurfaces::SurfLocalEnvironment;
         using DataSurfaces::Surface;
-        using DataSurfaces::TotSurfLocalEnv;
+        using DataSurfaces::SurfLocalEnvironment;
```
mjwitte: This looks completely unrelated. Should this file be touched at all? Same for  Vectors.cc

rraustad: Yeah, I was also confused with this diff. Clang moved this line to put in alphabetical order. Not sure why since I didn't change this file nor vectors.cc

mjwitte: But if you didn't touch the file, clang shouldn't touch it either.

src/EnergyPlus/VariableSpeedCoils.cc
mjwitte: Is this still an open question?

> I thought we had changed this line (for zone equipment and next line for air loop equipment) to account for air-cooled versus water-cooled. This needs to be corrected eventually. Need to compare diff's to #6830 before anything else happens

rraustad: This needs correcting. I can do it here or elsewhere.

mjwitte: That's your call - just let me know if you plan to add it here.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -4127,6 +4151,7 @@ namespace ReportSizingManager {
             DesExitTemp =
                 max(FinalSysSizing(SysNum).CoolSupTemp, AvgZoneTemp - ZoneCoolLoadSum / (StdRhoAir * CpAir * FinalSysSizing(SysNum).DesCoolVolFlow));
             DesFlow = FinalSysSizing(SysNum).DesCoolVolFlow;
+            DesExitHumRat = Psychrometrics::PsyWFnTdbRhPb(DesExitTemp, 0.9, DataEnvironment::StdBaroPress, "GetCoilDesFlowT");
```
mjwitte, line 4169: Why 90% RH here?

rraustad: We don't know how close the supply air condition is to saturation at this point. Could choose 95% maybe but probably doesn't make much of a difference. @EnergyArchmage, any input to the reasoning of this choice?

mjwitte: @rraustad @EnergyArchmage Not clear in the discussions above if the 5ZoneAirCooled numbers are with Chicago weather (as it should be) or with Golden weather (like CI runs for whatever reason, won't change that until after release).  With Golden weather, the new warning appears and design conditions change, as well as the size, even though it's Chicago design days but with different altitude - that's all? 

Bigger question is: With Chicago weather, there is no coil sizing warning (develop or this branch) yet the sizing results for the Main Cooling Coil have changed. Wasn't expecting that.

rraustad: When the water coil sizes capacity it includes fan heat at the inlet/outlet for draw/blow through configurations. This isn't getting propagated to the DataDesOutlet/InletAirTemp variables.

mjwitte:
@rraustad Please merge in develop again.

mjwitte:
@rraustad @EnergyArchmage Here's the [new warning for 5ZoneAirCooled](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2019-03/a837bb7430496285acd0d88e081fb7bc3a0fa3f4-59f89904054fd3f800ffbbda817fbbe71d079136/5ZoneAirCooled/x86_64-MacOS-10.13-clang/eplusout.err.diff.html) main cooling coil:
```
+   ** Warning ** In calculating the design coil UA for Coil:Cooling:Water MAIN COOLING COIL 1
+   **   ~~~   ** the apparatus dew-point is below the coil design inlet water temperature;
+   **   ~~~   ** the coil outlet design conditions will be changed to correct the problem.
+   **   ~~~   ** The initial design conditions are: Tair,in = 22.1807
+   **   ~~~   **                                    Wair,in = 9.405898E-003
+   **   ~~~   **                                    Twater,in = 7.0000
+   **   ~~~   **                                    Tair,out = 11.9312
+   **   ~~~   **                                    Wair,out = 8.000000E-003
+   **   ~~~   ** The revised design conditions are: Tair,out = 10.2003
+   **   ~~~   **                                    Wair,out = 8.700460E-003
```
I understand raising Wair,out to something that the 7C entering ChW can handle, but why is Tair,out lowered?

mjwitte:
@rraustad I've resolved the conflicts locally. Will push up later after letting unit test run to be sure.

rraustad:
w is lower now that it is using the correct outlet air humrat

    } else if (SizingType == CoolingWaterDesAirOutletHumRatSizing) {
        AutosizeDes = FinalSysSizing(CurSysNum).CoolSupHumRat;

and both T and w get recalculated for that UA warning that uses enthalpic comparison.

    if (DesAirApparatusDewPtEnth <= DesSatEnthAtWaterInTemp) BelowInletWaterTemp = true;

        if (Ipass == 1 && (NoSatCurveIntersect || CBFTooLarge || BelowInletWaterTemp)) {
            // reset outlet conditions to 90% relative humidity at the same outlet enthalpy
            TOutNew = TdbFnHRhPb(DesOutletAirEnth, 0.9, StdBaroPress);
            WOutNew = PsyWFnTdbH(TOutNew, DesOutletAirEnth);

mjwitte:
I see. So where's the 90% come from?

rraustad:
Reach up to the sky, clench your hand, and pull if right down. We could do some investigation of water coil design conditions to see how close this is. Also, that wasn't changed as part of this effort.

rraustad:
Well, one place did add a line using 90%, but there are others I believe. Even if not, 90-95% is about right for a choice of outlet air RH. I hate it when I see SA RH close to 100%.

mjwitte:
OK. I'm satisfied. Ran 5ZoneAirCooled annual and the change in annual energy use is <1%.  Will merge after CI completes, just to be sure.

Also, note that tablediff failed on 5ZoneAirCooled_Annual - likely because the order of columns changed in the  Component Sizing Summary Coil:Cooling:Water subtable. But then why didn't that trip up tablediff on other files? It says something about an encoding error but I recall that's a header key difference. @Myoldmopar ? See [_5ZoneAirCooled_annual here](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-f6824136d31dbe40457cbb28e6d075017b454c56-x86_64-MacOS-10.13-clang.html).

mjwitte (approved):

src/EnergyPlus/DataSizing.cc
```diff
@@ -269,7 +269,7 @@ namespace DataSizing {
     Real64 DataCoilSizingAirOutHumRat(0.0);         // saves sizing data for use in coil object reporting
     Real64 DataCoilSizingFanCoolLoad(0.0);          // saves sizing data for use in coil object reporting
     Real64 DataCoilSizingCapFT(1.0);                // saves sizing data for use in coil object reporting
-    Real64 DataDesAccountForFanHeat(true);          // include fan heat when true
+    bool DataDesAccountForFanHeat(true);            // include fan heat when true
```
Myoldmopar, line 272: Well this is certainly an improvement already... :upside_down_face:

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1087,6 +1087,7 @@ namespace ReportSizingManager {
                                 AutosizeDes = max(ZoneEqSizing(CurZoneEqNum).AirVolFlow,
                                                   ZoneEqSizing(CurZoneEqNum).CoolingAirVolFlow,
                                                   ZoneEqSizing(CurZoneEqNum).HeatingAirVolFlow);
+                                // what is this IF block for? Already inside of (SizingType == CoolingAirFlowSizing).
```
Myoldmopar, line 1090: If this is true then the SizingType == HeatingAirflowSizing block should be removed from here.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -1908,8 +1909,8 @@ namespace ReportSizingManager {
                     } else {
                         AutosizeDes = FinalZoneSizing(CurZoneEqNum).CoolDesHumRat;
                     }
-                    if (AutosizeDes > DataDesInletAirHumRat && DataCapacityUsedForSizing > 0.0 &&
-                        UtilityRoutines::SameString(CompType, "COIL:COOLING:WATER")) { // zone coil uses mDot>0, sys coil uses cap > 0
+                    if (AutosizeDes > DataDesInletAirHumRat && (UtilityRoutines::SameString(CompType, "COIL:COOLING:WATER") ||
+                                                                UtilityRoutines::SameString(CompType, "COIL:COOLING:WATER:DETAILEDGEOMETRY"))) {
```
Myoldmopar, line 1913: Adding another string comparison seems counterproductive.  I wouldn't hold everything up, but it seems like we should be passing int's around, not strings, if we're going to be doing comparisons in here.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -2845,7 +2852,9 @@ namespace ReportSizingManager {
                     } else {                                                   // coil is in main air loop
                         if (PrimaryAirSystem(CurSysNum).NumOACoolCoils == 0) { // there is no precooling of the OA stream
                             AutosizeDes = FinalSysSizing(CurSysNum).MixTempAtCoolPeak;
-                        } else { // thereis precooling of the OA stream
+                        } else if (DataDesInletAirTemp > 0.0) {
```
Myoldmopar, line 2855: Is an air temperature compared to zero really correct here?  The flow greater than zero check, sure, but seems random to compare a non-absolute temperature to zero.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -3132,6 +3155,8 @@ namespace ReportSizingManager {
                                     CoilInHumRat = OutAirFrac * FinalSysSizing(CurSysNum).PrecoolHumRat +
                                                    (1.0 - OutAirFrac) * FinalSysSizing(CurSysNum).RetHumRatAtCoolPeak;
                                 }
+                                if (DataDesInletAirTemp > 0.0) CoilInTemp = DataDesInletAirTemp;
```
Myoldmopar, line 3158: I mean, it feels like temperature = 0.0 is being used as a flag for special operation.  Is there **any** chance that this could get to this point where the user actually entered a value of zero?

rraustad: Data* variables get initialized to 0, meaning they haven't been set yet. This is how we know they haven't been set. We could instead use -999.

rraustad: Oh, you knew that.

rraustad: Unlikely a user chooses freezing as a design condition. But what about low temp systems?

Myoldmopar: It may be worth considering at a later time.  Zero degrees Celsius has physical meaning, so it feels like it should be avoided as a flag-value.  -99999 is generally physically meaningless for values like autosized flow and capacity, so it works nicely.  Anyway, nothing to do for now.

src/EnergyPlus/ReportSizingManager.cc
rraustad: We'll integerize these on the inevitable sizing consolidation.

src/EnergyPlus/ReportSizingManager.cc
rraustad: These need a more defensible initialization.

src/EnergyPlus/ReportSizingManager.cc
rraustad: As long as others agree with my logic in seeing that it doesn't belong.

src/EnergyPlus/DataSizing.cc
rraustad: Fixed my original mistake. Thanks to the compiler it was working then as well.

Myoldmopar:
Looks like this is all ready to go.  Big diffs but I think we've all rigored over it and feel it is satisfactory.  Anyone opposed to the green button at this point?

mjwitte:
No objections here.

Issue #7152, opened 2019-02-04
Change debug configuration to catch NaNs by default on Windows


mjwitte:
Issue overview
--------------
There are special instructions to catch NaNs on Windows [here](https://github.com/NREL/EnergyPlus/wiki/Debugging-Tips).
It would be nice to add this to the debug cmake configuration so that:
1. EnergyPlus gets set for /fp:strict
2. except sqlite needs /fp:precise
3. main.cc needs to be configurable or add a compiler flag to include the `unsigned int fp_control_state` line

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Closed via #7198

PR #7144, opened 2019-01-30
Fix #7092 - Catch errors in GroundHeatExchanger:HorizontalTrench when pipe inner/outer diameters don't match

jmarrec:
Pull request overview
---------------------

Fix #7092 - Catch errors in GroundHeatExchanger:HorizontalTrench when pipe outer diameter isn't stritcly greater than inner diameter 

Changelog:
* Corrected small issue in existing reporting for the other object that has inner&outer diameters (`PipingSystem:Underground:PipeCircuit`)
* Added catching error + reporting for `GroundHeatExchanger:HorizontalTrench`
* Wrote a unit test to test both `PipingSystem:Underground:PipeCircuit` and `GroundHeatExchanger:HorizontalTrench`
* Modified `NodeInputManager::GetNodeListsInput(bool &ErrorsFound)` to not set `ErrorsFound` to false (found this problem while testing for `PipingSystem:Underground:PipeCircuit`)

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add to input rules file for interfaces
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Required documentation updates?
[...]

Issue #7092, opened 2018-12-07
Ground heat exchanger divide by zero if user inputs equal pipe inner and outer diameters


mjwitte:
Issue overview
--------------
User file with equal values for `GroundHeatExchanger:HorizontalTrench` "Pipe Inner Diameter" and "Pipe Outer Diameter" results in a divide-by-zero [here](https://github.com/NREL/EnergyPlus/blob/ae4c2ce0d208dff6982b852c40fccb78aa78c552/src/EnergyPlus/PlantPipingSystemsManager.cc#L6428-L6430) because [`XY_CrossSectArea`](https://github.com/NREL/EnergyPlus/blob/ae4c2ce0d208dff6982b852c40fccb78aa78c552/src/EnergyPlus/PlantPipingSystemsManager.hh#L257-L261) returns zero. In the release build, this propogates NANs until this familiar and confusing fatal error occurs:

```
   ** Severe  ** DualSetPointWithDeadBand: Unanticipated combination of heating and cooling loads - report to EnergyPlus Development Team
   **   ~~~   ** occurs in Zone=ZONE_1_14 During Warmup, Environment=WASHINGTON DC DULLES INTL AR ANN CLG 1% CONDNS DB=>MWB, at Simulation time=07/21 23:00 - 24:00
   **   ~~~   ** LoadToHeatingSetPoint=NAN, LoadToCoolingSetPoint=NAN
   **   ~~~   ** Zone Heating Set-point=21.11
   **   ~~~   ** Zone Cooling Set-point=22.78
   **   ~~~   ** Zone TempDepZnLd=NAN
   **   ~~~   ** Zone TempIndZnLd=NAN
   **   ~~~   ** Zone ThermostatSetPoint=22.78
   **  Fatal  ** Program terminates due to above conditions.
```

Input processing for `GroundHeatExchanger:HorizontalTrench` should probably require a minimal positive value for Outer Diameter - Inner Diameter, otherwise throw a severe/fatal error.

`GroundHeatExchanger:Pond` also has inputs for inner and outer diameter, so it may have the same issue.

Other ground heat exchangers have a diameter plus thickness which must be >0, or just one diameter(?), so they don't have the same problem.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Helpdesk ticket number 12268.

### Checklist
[...]

PR #7198, opened 2019-03-04
Catch NaNs in Windows debug build

mjwitte:
Pull request overview
---------------------
Attempt to automate settings to catch NaNs in windows debug builds. Addressing #7152.

If an NaN bug causes a user file to fail, sometimes the NaN propagates forward and shows in the text of an error message. On Mac and Linux and debug build will trap the NaN where it first occurs, but not so much on Windows. 

For years, [some manual steps](https://github.com/NREL/EnergyPlus/wiki/Debugging-Tips) have been required to catch an NaN when debugging a file on Windows. Kind of a pain, especially since SQLite won't compile with fp:strict because of some declarations with computed constants, requiring another manual step to build. So, this PR fixes all that with:

1. Add /fp:strict for Win debug build.
2. Override that with /fp:fast for SQLite.
3. Add a #def flag `MSVC_DEBUG` to enable a line of code in main.cc

The defect file from #7092 can be used to test this (at least until #7144 gets merged.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/cmake/CompilerFlags.cmake
+++ b/cmake/CompilerFlags.cmake
@@ -18,9 +18,6 @@ IF ( MSVC AND NOT ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" ) ) # Visual C++
     #  4996 Deprecated functions (/D_SCL_SECURE_NO_WARNINGS /D_CRT_SECURE_NO_WARNINGS /D_CRT_NONSTDC_NO_WARNINGS)
     #  4503 The decorated name was longer than the compiler limit (4096), and was truncated.
 
-    # need to figure out how to set this to avoid the major slow-down in debugging:
-    # Configuration Properties ->Debugging -> Environment, use drop-down list to choose <Edit> and type _NO_DEBUG_HEAP=1 then click OK
-
     # COMPILER FLAGS
     ADD_CXX_DEFINITIONS("/nologo")
     ADD_CXX_DEFINITIONS("/EHsc")
@@ -38,6 +35,8 @@ IF ( MSVC AND NOT ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" ) ) # Visual C++
     # ADDITIONAL DEBUG-MODE-SPECIFIC FLAGS
     ADD_CXX_DEBUG_DEFINITIONS("/Ob0") # Disable inlining
     ADD_CXX_DEBUG_DEFINITIONS("/RTCsu") # Runtime checks
+    ADD_CXX_DEBUG_DEFINITIONS("/fp:strict") # Floating point model
+    ADD_CXX_DEBUG_DEFINITIONS("/DMSVC_DEBUG") # Triggers code in main.cc to catch floating point NaNs
 
 ELSEIF ( CMAKE_COMPILER_IS_GNUCXX OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" ) # g++/Clang
     option(ENABLE_THREAD_SANITIZER "Enable thread sanitizer testing in gcc/clang" FALSE)
--- a/third_party/SQLite/CMakeLists.txt
+++ b/third_party/SQLite/CMakeLists.txt
@@ -1,6 +1,8 @@
 
 IF ( MSVC ) # visual c++ (VS 2013)
     ADD_DEFINITIONS("-W0")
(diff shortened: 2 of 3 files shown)


cmake/CompilerFlags.cmake
```diff
@@ -38,6 +35,8 @@ IF ( MSVC AND NOT ( "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" ) ) # Visual C++
     # ADDITIONAL DEBUG-MODE-SPECIFIC FLAGS
     ADD_CXX_DEBUG_DEFINITIONS("/Ob0") # Disable inlining
     ADD_CXX_DEBUG_DEFINITIONS("/RTCsu") # Runtime checks
+    ADD_CXX_DEBUG_DEFINITIONS("/fp:strict") # Floating point model
+    ADD_CXX_DEBUG_DEFINITIONS("/DMSVC_DEBUG") # Triggers code in main.cc to catch floating point NaNs
```
rraustad, line 39: Is the "D" supposed to be here? DMSVC_DEBUG

Line 54 shows: #ifdef MSVC_DEBUG

mjwitte: Yes, based on [other examples](https://github.com/NREL/EnergyPlus/blob/develop/cmake/CompilerFlags.cmake#L31-L32), /D sets a #def variable, followed by the variable name. When in VS, the variable name is blue if it's set. When it's not set, the code inside the #ifdef is greyed out.

Myoldmopar: Agree with this approach, it's /D on MSVC but -D on GCC-like compilers.  (https://docs.microsoft.com/en-us/cpp/build/reference/d-preprocessor-definitions?view=vs-2017)

third_party/SQLite/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 
 IF ( MSVC ) # visual c++ (VS 2013)
     ADD_DEFINITIONS("-W0")
+    ADD_CXX_DEBUG_DEFINITIONS("/fp:fast") # override fp:strict in debug builds
```
Myoldmopar, line 4: Good find here, I don't exactly understand why this is needed for SQLite, but I'm glad you got this worked out so it will build properly and catch NaNs for you.

Myoldmopar (approved):

jasondegraw:
@mjwitte Does this have any impact on the debug build or runtime?

mjwitte:
@jasondegraw Good question on runtime - will test that.

third_party/SQLite/CMakeLists.txt
mjwitte: Yeah, with fp:strict, the compiler chokes on this block, complaining "initializer is not a constant", starting at line 20910 in SQLite.cc.
```
static const struct {
  u8 eType;           /* Transformation type code */
  u8 nName;           /* Length of th name */
  char *zName;        /* Name of the transformation */
  double rLimit;      /* Maximum NNN value for this transform */
  double rXform;      /* Constant used for this transform */
} aXformType[] = {
  { 0, 6, "second", 464269060800.0, 86400000.0/(24.0*60.0*60.0) },
  { 0, 6, "minute", 7737817680.0,   86400000.0/(24.0*60.0)      },
  { 0, 4, "hour",   128963628.0,    86400000.0/24.0             },
  { 0, 3, "day",    5373485.0,      86400000.0                  },
  { 1, 5, "month",  176546.0,       30.0*86400000.0             },
  { 2, 4, "year",   14713.0,        365.0*86400000.0            },
};
```
I guess the alternative would be to change these, but I didn't want to touch that.

mjwitte:
Ran a timing test with 5ZoneAirCooled. Current Windows debug build took 03:44, this branch took 03:48. So, it doesn't appear to have a significant impact.

third_party/SQLite/CMakeLists.txt
mbadams5: Yeah I wouldn't touch sqlite library code.

Myoldmopar:
Runtime looks to check out, avoiding changing SQLite sounds good.  CI is happy, this is good to go in.

Issue #7173, opened 2019-02-14
Incorrect handling of weather file leap year flag in some situations


jasondegraw:
Issue overview
--------------
In some situations, setting the leap year flag in the weather file will cause a leap year to appear for non-leap years. Testing with 5ZoneAirCooled.idf and an EPW file with the leap year flag set to 'Yes' ended up adding an incorrect leap day for a variety of different situations, but not for simulations beyond a year in length. So something untoward is going on with handling the flag out of the EPW.

### Details
Some additional details for this issue (if relevant):
 - Version 9.0.1
 - Helpdesk ticket ID 13165

### Checklist
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Myoldmopar:
Fixed via #7199

PR #7199, opened 2019-03-04
Fix Leap Year Flag Issue

jasondegraw:
Pull request overview
---------------------
As described in #7173, the leap year flag in the weather file was overriding inputs in certain circumstances. The culprit was a bad bit of logic in WeatherManager introduced in the year awareness work.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types

### DIFF
--- a/src/EnergyPlus/WeatherManager.cc
+++ b/src/EnergyPlus/WeatherManager.cc
@@ -6102,8 +6102,8 @@ namespace WeatherManager {
             if (!lAlphaFieldBlanks(2)) { // Have input
                 auto result = weekDayLookUp.find(cAlphaArgs(2));
                 if (result == weekDayLookUp.end()) {
-                    ShowWarningError(cCurrentModuleObject + ": object=" + RunPeriodInput(Loop).title + cAlphaFieldNames(2) + " invalid (Day of Week) [" +
-                                     cAlphaArgs(2) + "] for Start is not valid, Sunday will be used.");
+                    ShowWarningError(cCurrentModuleObject + ": object=" + RunPeriodInput(Loop).title + cAlphaFieldNames(2) +
+                                     " invalid (Day of Week) [" + cAlphaArgs(2) + "] for Start is not valid, Sunday will be used.");
                     RunPeriodInput(Loop).startWeekDay = WeekDay::Sunday;
                 } else {
                     RunPeriodInput(Loop).startWeekDay = result->second;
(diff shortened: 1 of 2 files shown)


jasondegraw:
Fixes #7173. Let me know if you'd like me to undo the clang-format changes.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -9990,15 +10002,17 @@ namespace WeatherManager {
...
-                        LocalLeapYearAdd = 1;
+                    LocalLeapYearAdd = 0;
+                    if (isLeapYear(env.StartYear)) {
+                        // If a leap year is supported by the weather file, do it.
+                        if (WFAllowsLeapYears) {
+                            env.IsLeapYear = true; // explicit set, this might be unwise
+                            LocalLeapYearAdd = 1;
+                        } else {
+                            env.IsLeapYear = false; // explicit set, this might be unwise
+                        }
```
Myoldmopar, line 10015: OK, so the weather file leap year support is only checked if the current year is actually a leap year.  Makes sense.

tst/EnergyPlus/unit/RunPeriod.unit.cc
```diff
@@ -420,3 +420,47 @@ TEST_F(EnergyPlusFixture, SizingPeriod_WeatherFile)
...
+TEST_F(EnergyPlusFixture, RunPeriod_BadLeapDayFlagLogic)
+{
+    std::string const idf_objects = delimited_string({
+        "SimulationControl, NO, NO, NO, YES, YES;",
+        "Timestep,4;",
+        "RunPeriod,",
+        "RP3,                     !- Name",
+        "1,                       !- Begin Month",
+        "1,                       !- Begin Day of Month",
+        "2019,                    !- Begin Year",
```
Myoldmopar, line 433: So this unit test should _not_ be a leap year, but I'm guessing we'll set the weather file up as leap year capable.

tst/EnergyPlus/unit/RunPeriod.unit.cc
```diff
@@ -420,3 +420,47 @@ TEST_F(EnergyPlusFixture, SizingPeriod_WeatherFile)
...
+
+    EXPECT_FALSE(errors_in_input);
+
+    WeatherManager::Environment.allocate(1);
+    // These may already be set, but do it anyway
+    DataEnvironment::TotDesDays = 0;
+    WeatherManager::TotRunPers = 1;
+    WeatherManager::TotRunDesPers = 0;
+
+    WeatherManager::WFAllowsLeapYears = true; // This was hitting a bad bit of logic
```
Myoldmopar, line 459: So the run period should never even check the leap year capability of the weather file, meaning we won't have a chance to check this flag.  OK.

tst/EnergyPlus/unit/RunPeriod.unit.cc
```diff
@@ -420,3 +420,47 @@ TEST_F(EnergyPlusFixture, SizingPeriod_WeatherFile)
...
+    WeatherManager::SetupEnvironmentTypes();
+    
+    EXPECT_FALSE(WeatherManager::Environment[0].IsLeapYear);
```
Myoldmopar, line 462: Looks good to me!

Myoldmopar (approved):
Pulled latest develop and built locally.  Ran the new unit test and it passes and tests exactly what this fixes.  This is a nice compact fix and test.  Thanks @jasondegraw.  It should merge shortly, I just want to go back and make sure there weren't other comments on it.

Myoldmopar:
Even CI is happy.  Merging.

src/EnergyPlus/WeatherManager.cc
Myoldmopar: @jasondegraw hey, should we have put a warning in here for if the runperiod is a leap year but the weather file doesn't support it?  If so is this the right spot for it?  We'll add it within a different branch if so.  (Or you could add it to the namespace change branch on the next merge up)

jasondegraw: @Myoldmopar I'm not sure it is the right spot to add a warning. It would be good to have a warning, but my preference would be to wait until weather handling gets refactored. If we put in an issue on the lack of warning it'll be easier on whoever does the refactor. If we add the warning now, it'll be one more thing that someone will have to find. Previous behavior was to be silent, so I think it'd be alright to leave it alone for a bit longer, but if you'd rather add the warning that's OK too.

Myoldmopar: Given how late in the game we are right now, I support the idea of doing nothing but opening up a ticket.  Thanks @jasondegraw

Issue #6938, opened 2018-08-30
Sync Kiva convection algorithms with EnergyPlus user-specified algorithm


nealkruis:
Currently, Kiva uses only the TARP algorithm for convection. We can replace the TARP algorithm in Kiva with a virtual function that can be set by the calling program.

Requested by Carrier team.

JimPegues:
To elaborate on this ... Replacing the TARP algorithm with a virtual function set by the calling program allows the convention model used in KIVA to be synchronized with the convection model used by the room.  For example, suppose a room is supplied by a forced air HVAC system such as VAV.  The room would be modeled to use the 'ceiling diffuser' convection model since wall, ceiling, floor surfaces in the room will experienced forced convection.  When the room also contains ground connected elements like slab on grade floor, basement floor, or basement wall, modeled via KIVA those surfaces will automatically use the TARP model representing natural convection, and as a result will likely underestimate the convective heat flow between the surface and the room air.  If KIVA could use the same convection model as used for the rest of the room, that would synchronize convection modeling between KIVA and the rest of the room, and enhance the accuracy of load calculations.  Allowing the calling program (EnergyPlus) to specify convention via a virtual function allows this synchronization of convection algorithms.  (Jim Pegues, Carrier Corporation, 9/7/2018)

Myoldmopar:
Closed via #7202

PR #7202, opened 2019-03-06
Sync Kiva convection algorithms with EnergyPlus user-specified algorithm

nealkruis:
Pull request overview
---------------------
Addresses #6938. Kiva is currently set to use the TARP convection algorithm for cell-by-cell convection calculations. This pull request modifies Kiva to allow a calling program (e.g., EnergyPlus) to set the function used for convection calculations. Kiva will now utilize whichever algorithm is set by the user.

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


tst/EnergyPlus/unit/ConvectionCoefficients.unit.cc
```diff
@@ -596,7 +596,7 @@ TEST_F(EnergyPlusFixture, ConvectionCoefficientsTest_EvaluateIntHcModelsFisherPe
     // Test 2: Ceiling Diffuser Model
     ConvModelEquationNum = HcInt_FisherPedersenCeilDiffuserCeiling;
     Hc = 0.0;
-    HcExpectedValue = 6.333;
+    HcExpectedValue = 2.234;
```
Myoldmopar, line 599: How did this change so much?  I thought the only diff would be the tiny precision change due to the repeated code.

src/EnergyPlus/ConvectionCoefficients.hh
```diff
@@ -721,6 +809,13 @@ namespace ConvectionCoefficients {
                          Real64 const LongAxisOutwardAzimuth, // or Facing, Direction the surface outward normal faces (degrees)
                          int const SurfNum);
 
+    Real64 CalcClearRoof(Real64 const AirTemp,
+                         Real64 const WindAtZ,
+                         Real64 const WindDirect, // Wind direction measured clockwise from geographhic North
+                         Real64 const RoofArea,
+                         Real64 const RoofPerimeter,
+                         int const RoughnessIndex);
+
```
Myoldmopar, line 818: All these overloads make sense per your description on the call yesterday.

testfiles/CMakeLists.txt
```diff
@@ -562,10 +562,13 @@ ADD_SIMULATION_TEST(IDF_FILE WWHPSimpleAuto.idf EPW_FILE USA_IL_Chicago-OHare.In
...
 ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledGroundHTSlabOnGrade.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
-ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaBasement.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
-ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaRefBldgMediumOffice.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
 ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaSlab.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
+ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaBasement.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
 ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaWalkoutBasement.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
+ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaRefBldgMediumOffice.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
+ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaBasementAdaptiveConvection.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
+ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaConvection.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
+ADD_SIMULATION_TEST(IDF_FILE ZoneCoupledKivaConvectionAdaptiveSmallOffice.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
```
Myoldmopar, line 571: Adding to the test suite is appreciated given the changes here.

Myoldmopar (commented):
Overall it's fine here, but I am curious about why the unit test convection value changed so drastically.  I was expecting small changes due to the precision change, but that seems significant.  I'll also be looking at the diffs from the adaptive file to make sure they are OK, given the nature of your precision change.

Myoldmopar:
I looked at the adaptive file and I was a little surprised to see the surface temperatures changing more than a tenth of a degree:

image

I kept looking at other columns though, and I feel a little better about it when I see that the surface convection gain change is really small:

image

So I think the change in convection value is just causing a steady change in zone conditions and the surfaces and zone air are still together at essentially the same state, everything is just slightly elevated (or reduced).

I'm OK with it.  Can you please explain why the convection values changed so drastically in the unit tests though?

Also, CI is going a bit slow, but still churning through lots of active branches.  I'll definitely wait until CI makes a pass over this before merging anything.

tst/EnergyPlus/unit/ConvectionCoefficients.unit.cc
nealkruis: The same functions are called for the zonal level ceiling diffuser algorithm as the surface level algorithm in `EvaluateIntHcModels` except that they were calculating the ACH differently. [The version used in EvaluateIntHcModels](https://github.com/NREL/EnergyPlus/blob/a837bb7430496285acd0d88e081fb7bc3a0fa3f4/src/EnergyPlus/ConvectionCoefficients.cc#L5961-L5970) wasn't applying [the MinFlow constraint used at the zonal level](https://github.com/NREL/EnergyPlus/blob/a837bb7430496285acd0d88e081fb7bc3a0fa3f4/src/EnergyPlus/ConvectionCoefficients.cc#L4325).

The unit tests use a very small flow rate with a 1 m^3 volume which trigger the MinFlow constraint after I created a [common function to calculate ACH](https://github.com/NREL/EnergyPlus/pull/7202/commits/20b7c4c7d126f65f4c240f35c5581d82da25637f#diff-3edd28106d5270bbc5fee24bd22e637bR4265) for use in the Ceiling Diffuser algorithms.

Myoldmopar: Makes sense.  Thanks for the explanation.

Myoldmopar (approved):

tst/EnergyPlus/unit/ConvectionCoefficients.unit.cc
nealkruis: FWIW I think flow rates this low almost never happen in real models. This change didn't trigger any diffs in our test suite (though not many of the models use the ceiling diffuser algorithm).

Myoldmopar:
This is looking great on CI.  Once integration debug finishes and custom check follows immediately after, I anticipate this will merge in.

mbadams5:
I looked over the code and it seems reasonable. I don't have much more to add to this PR on top of what Edwin already asked.

Issue #7010, opened 2018-10-13
Transition IDF Version Updater Bug in v9.0


ff-kschlansker:
There appears to be an issue introduced between 8.9.0 and 9.0.1 having to do with updating IDFs to newer versions.  With 8.9.0 I had no problem updating IDF files.  With 9.0.1 I get the following error on every IDF I attempt to update:

```
At line 787 of file /Users/kbenne/ci_runs/EnergyPlus-v9.0.1-x86_64-clang/src/Transition/VCompareGlobalRoutines.f90
Fortran runtime error: Bad integer for item 1 in list input
```

The full output from the Transition program is as follows:
```
 Transition Starting
 Conversion 7.2 => 8.0
 Processing Old IDD -- V7-2-0-Energy+.idd
 Processing New IDD -- V8-0-0-Energy+.idd
 Will create new full IDFs
 Will create new IDF lines with units where applicable
 Will create new IDF lines leaving blank incoming fields as blank (no default fill)
 F
 Processing IDF -- /Users/kschlansker/work/ff/stack-validation/idf-conversion/test.ep90/idfs/RefBldgOutPatientPre1980_v1.4_7.2_1A_USA_FL_MIAMI.idf
At line 787 of file /Users/kbenne/ci_runs/EnergyPlus-v9.0.1-x86_64-clang/src/Transition/VCompareGlobalRoutines.f90
Fortran runtime error: Bad integer for item 1 in list input
```

mjwitte:
@ff-kschlansker I see the problem. There is a shared function that is used across all transition versions which was updated for v9.0 without recognizing that it will break earlier transition versions. 

### Workaround
Copy all of the older transition exe's from the v8.9.0 install into the v9.0.1 IDFVersionUpdater folder.

ff-kschlansker:
Thank you @mjwitte - using the the EP 9.0 binaries for only V8-9-0 ==> V9-0-0, and EP 8.0 binaries for everything <V9-0-0 appears to work.

Myoldmopar:
Closed via #7007

jasminaburek:
I got IDF updater working without errors but the EnergyPlus run shows severe error
Program Version,EnergyPlus, Version 9.0.1-bb7ca4f0da, YMD=2019.03.06 09:54,
   ** Severe  ** <root>[WaterHeater:Mixed][SWH Laundry][off_cycle_parasitic_fuel_type] - "Natural Gas" - Failed to match against any enum values.
[...]

PR #7210, opened 2019-03-07
Address byte order mark (BOM) in UTF-8 with BOM IDF files

mbadams5:
Pull request overview
---------------------
This issue was uncovered in #7010.

UTF-8 files can be encoded with byte order mark (BOM) but it is not encouraged. This change looks for those specific bytes and ignores them to continue parsing the IDF file.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/InputProcessing/IdfParser.cc
+++ b/src/EnergyPlus/InputProcessing/IdfParser.cc
@@ -198,6 +198,14 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success, 
         objectTypeMap.emplace(std::move(key), it.key());
     }
 
+    if (idf.size() > 3) {
+        // UTF-8 Byte Order Mark
+        if (idf[0] == '\xEF' && idf[1] == '\xBB' && idf[2] == '\xBF') {
+            index += 3;
+            index_into_cur_line += 3;
+        }
+    }
+
     int idfObjectCount = 0;
     while (true) {
         token = look_ahead(idf, index);
--- a/tst/EnergyPlus/unit/InputProcessor.unit.cc
+++ b/tst/EnergyPlus/unit/InputProcessor.unit.cc
@@ -335,6 +335,40 @@ TEST_F(InputProcessorFixture, decode_encode_3)
     EXPECT_EQ(expected, encoded);
 }
 
+TEST_F(InputProcessorFixture, byte_order_mark)
+{
+    auto const idf(delimited_string(
+            {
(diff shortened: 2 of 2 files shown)


src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -198,6 +198,14 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success,
         objectTypeMap.emplace(std::move(key), it.key());
     }
 
+    if (idf.size() > 3) {
+        // UTF-8 Byte Order Mark
+        if (idf[0] == '\xEF' && idf[1] == '\xBB' && idf[2] == '\xBF') {
```
rraustad, line 203: @mbadams5 do the initial BOM bytes have to be in this order? can they be swapped?

Myoldmopar: That's the [required order](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8) for a UTF-8 compliant BOM.  Other encodings (UTF-16) have others, but I don't think we need to concern ourselves with those.  Though, @mbadams5, if we are able to read [other encoding types](https://en.wikipedia.org/wiki/Byte_order_mark#Byte_order_marks_by_encoding) that use BOM, would it be worthwhile to add multiple catches here?

mbadams5: @Myoldmopar The JSON library only supports UTF-8 so I don't see the benefit of adding the other encodings BOM.

rraustad: The review is complete then.

rraustad: How would you know if an idf was formatted in a non-supported UTF encoding?

mbadams5: It would likely fail to parse and if it doesn't then the JSON parser will either reject or ignore the non-supported codepoints.

rraustad: Back to you guys on this one. I just wanted to talk out loud about this to see if something else was needed here. This change allows a step forward to anticipating UTF encoding so at least it's in the right direction. This could go in as-is for now. The CI failure for EnergyPlusFixture.InternalHeatGains_ElectricEquipITE_DefaultCurves doesn't look related.

Myoldmopar:
Yep, this looks good.  The unit test failure is unrelated.  Merging.

Issue #7211, opened 2019-03-07
Array bounds error in GetCoilDesFlowT if time of cooling peak is not set


mjwitte:
Issue overview
--------------
User file has one DOAS airloop serving one zone. Both the zone and system have specified airflow (Sizing:Zone Flow/Zone, Sizing:System Flow/System). There is no zone thermostat, which does not cause an error because the zone sizing method is `InpDesAirFlow`. But this results in all of the sequenced sizing values to be zero, which results in the cooling peak design day and time not being set, which results in a zero array index [here in `GetCoilDesFlowT`](https://github.com/NREL/EnergyPlus/blob/9768414a04e06361a0c0afedcb0a8b89e5a12b10/src/EnergyPlus/ReportSizingManager.cc#L3087-L3088).

Several things to fix:
1. In `GetCoilDesFlowT` need to trap for zero `DD*` and `TimeStep*` values before they used as an array subscript. OR figure out a way to set these by maybe setting sequenced flows to the specified user input flow?

2. If the peak day and time are not set, then `GetCoilDesFlow` can't size using the system sizing cooling capacity control methods VT or Bypass.

3. In [`ZoneEquipmentManager::SetupZoneSizingArrays`](https://github.com/NREL/EnergyPlus/blob/a310a424e4ea98bb8183137f0fd5f1012a6bd918/src/EnergyPlus/ZoneEquipmentManager.cc#L1003) the thermostat check should include both `FromDDCalc` and `DesAirFlowWithLim`.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #7215

Issue #7001, opened 2018-10-10
Improved chilled water coil humidity control


mjwitte:
Issue overview
--------------
This new feature request seeks more precise humidity control for chilled water coils.

User notes that a simple system with a fixed max humidity ratio setpoint over-dehumidifies the supply air, because the adjusted temperature setpoint appears to simply be the dew point temperature corresponding to the humidity ratio setpoint. 

Work was done in #5336 to improve humidity controls for chilled water and DX coils. When the chilled water coil controller is set for TemperatureAndHumidity control, it lowers the dry bulb temperature setpoint as needed in order to reach the maximum humidity ratio setpoint if possible. According to the code comments, the final solution was to adjust the setpoint only once per timestep in order to avoid controller iteration errors. This request would revisit the ChW controls.

It has been considered in the past that the simulation sequence for Controller:WaterCoil should be changed to be more local to the coil instead of the current airloop solution technique in `SolveAirLoopControllers` where the controllers are simulated first, then the airloops, then repeat. Not sure if that would help here or not.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Helpdesk ticket number 12653

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closed via #7215

Issue #6892, opened 2018-08-07
UnitarySystem set point based control of humidity not working correctly


rraustad:
Issue overview
--------------
Testing of new AirLoopHVAC:UnitarySystem shows incorrect logic when checking for outlet humidity level.

Line 10766 of UnitarySystem.cc

    if (PartLoadFrac == 0.0) {
        OutletHumRatDXCoil = NoLoadHumRatOut;
    } else {
        OutletHumRatDXCoil = DataLoopNode::Node(OutletNode).HumRat;
    }

However, SolveRoot iterates on PLR via function DOE2DXCoilResidual (doesn't use SimDXCoil) and does not set the outlet node humidity ratio which is set in Report for DXCoils. So here, the OutletHumRatDXCoil shows the last iterations humrat. This may also be happening for other coil types. I thought at one point that the node conditions were set at the end of the CalcDOE2 routine but don't see that now.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added (EnergyPlusDevSupport defect file folder 6892)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@rraustad For this, I see three options to fix:
1. In the various residual functions in unitarysystem, for the ones that call a `Calc*` function, add lines to set the outlet node temp and humrat so they are current at the end of each root solver call.
2. In the various residual functions in unitarysystem, for the ones that call a `Calc*` function, change them to call `Sim*` instead.
3. In the various `Calc*` functions called from these residuals, add lines to set the outlet node temp and humrat.
Which is better/safer?
I will add these to #7215

rraustad:
[...]

PR #7215, opened 2019-03-08
Better humidity control for chilled water coils and for AirloopHVAC:UnitarySystem

mjwitte:
Pull request overview
---------------------
- [x] 1. Addresses #7211 to avoid array bounds violation in `ReportSizingManager::GetCoilDesFlowT`.

- [x] 2. Adresses #7001 by modifying Controller:WaterCoil to switch gears from controlling on temperature setpoint to controlling on humidity ratio when needed. This replaces the existing concept of adjusting the temperature setpoint with an approach temperature to achieve the humidity control.

3. Addresses #6892 to fix humidity control in AirloopHVAC:UnitarySystem. This is also alternate solution to #7001, using AirloopHVAC:UnitarySystem to control the chilled water coil instead of Controller:WaterCoil. This required a few tweaks to avoid uninitialized variables and get things to control well, but this pegs the leaving dry-bulb setpoint or the leaving humidity ratio setpoint, and reliably switches between the two. (See EnergyPlusDevSupport\DefectFiles for an idf like this). At present, there are still some warnings of "Iteration limit exceeded calculating latent part-load ratio for unit", so may need some more tweaks, but that may just be how it is.
- [x] Chilled water coil
- [x] DX single speed

Defect Files
---------------------
7001 defect file no longer crashes in debug build and maintains the Node 27 humidity ratio within 9.8e-05 of the humidity ratio setpoint of 0.011388. The jagged profile comes from using a humidity ratio control tolerance of 1.0e-04 (arbitrarily chosen to be tight, but not so tight that there are max iteration problems). Maybe that should be tighter.

image

For 6892 there are three defect files, 6892-in_Copy-HumCtrl, DOAS_CWCoil-UseUnitarySystem, DOAS_DXCoil-UseUnitarySystem:
DOAS_CWCoil-UseUnitarySystem - crashes with develop debug build, runs fine and controls almost exactly to setpoint with develop release build. This branch, runs fine and controls well with both release and debug builds.

DOAS_DXCoil-UseUnitarySystem - crashes with develop debug build, runs fine and does not control humidity with develop release build. T
his branch, runs fine and controls well with both release and debug builds.
image

Other Diffs Summary
---------------------
**HospitalBaseline, HospitalBaselineReheatReportEMS, HospitalLowEnergy, RefBldgHospitalNew2004_Chicago** - small diffs in results, eio diffs are all in the warmup convergence details, and err diffs are either fewer or rearranged max iteration messages.

**5ZoneCoolingPanel\*, 5Zone\*Baseboard\*, 5ZoneWaterCooled\*, HeatPumpWaterHeater\*** - eio diffs are all in the warmup convergence details, no err diffs, big diffs in results. 

Reviewed 5ZoneCoolingPanelBaseboard results - Space2-1 humidity control is similar to before, still not pegged at 50%, main cooling coil load is 0.4% less, and main heating coil load is 0.8% less. Expect similar for the other cooling panel files.

Reviewed 5ZoneWaterCooled_Baseboard results - again humidity control is similar, coil loads 0.1%-0.4% less.

Reviewed HeatPumpWaterHeater results - similar humidity control, 0.04% higher main cooling coil load, 0.5% lower main heating coil load

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -4451,23 +4451,28 @@ namespace ReportSizingManager {
...
+                TimeStepAtPeak = TimeStepAtSensPeak;
+            }
         }
         if (CoolCapCtrl == VAV) {
             DesExitTemp = FinalSysSizing(SysNum).CoolSupTemp;
             DesFlow = FinalSysSizing(SysNum).MassFlowAtCoolPeak / StdRhoAir;
         } else if (CoolCapCtrl == OnOff) {
             DesExitTemp = FinalSysSizing(SysNum).CoolSupTemp;
             DesFlow = DataAirFlowUsedForSizing;
+        } else if (TimeStepAtPeak == 0) {
```
rraustad: TimeStepAtPeak can't be 0 if VAV or OnOff fan control is used?  If so then this line won't get hit. This IF test should be first in list?

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -4451,23 +4451,28 @@ namespace ReportSizingManager {
...
+            ShowFatalError("GetCoilDesFlow: AirLoopHVAC=" + SysSizInput(SysNum).AirPriLoopName +
+                           " Central Cooling Capacity Control Method requires zone thermostats to calculate timestep loads.");
```
rraustad: You can still have a no design load condition (and therefore a 0 for time step at peak) even though a Tstat is used. Remember when zone return temp at heating peak was 0 and TU coils weren't sizing correctly? It's the combination of Tstat temp (i.e., zone temp during sizing) and supply air temp that yields a non-zero zone mass flow rate. Isn't there another way to know if a Tstat is used?

src/EnergyPlus/HVACControllers.cc
```diff
@@ -1465,6 +1434,10 @@ namespace HVACControllers {
                 (0.001 / (2100.0 * max(ControllerProps(ControlNum).MaxVolFlowActuated, SmallWaterVolFlow))) * (HVACEnergyToler / 10.0);
             // do not let the controller tolerance exceed 1/10 of the loop temperature tolerance.
             ControllerProps(ControlNum).Offset = min(0.1 * HVACTemperatureToler, ControllerProps(ControlNum).Offset);
+            if (ControllerProps(ControlNum).ControlVar == HVACControllers::iTemperatureAndHumidityRatio) {
+                // for temperature and humidity control, need tighter tolerance if humidity control kicks in
+                ControllerProps(ControlNum).Offset = ControllerProps(ControlNum).Offset*0.1;
```
rraustad: Would be nice to be able to trigger a tighter tolerance only when HumRatCtrlOverride = true.

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -2960,6 +2963,31 @@ namespace SimAirServingZones {
                     // not converge
                     ControllerConvergedFlag = PrimaryAirSystem(AirLoopNum).ControlConverged(AirLoopControlNum);
                     IsUpToDateFlag = true;
+                } else {
```
rraustad: This will require multiple more iterations of the air loop components. Wouldn't it be better to use the lesser of T and TdpFw as the temperature set point up at line 1283 - 1289 of HVACControllers?

src/EnergyPlus/HVACControllers.cc
mjwitte: I tried that first but it didn't work. There's some initial setup step that I couldn't find that includes the offset.

src/EnergyPlus/ReportSizingManager.cc
rraustad: Oh, I see. You put this IF test before those that use TimeStepAtPeak.

src/EnergyPlus/SimAirServingZones.cc
mjwitte: I can try that on a separate branch, but I think it will over-dehumidify.

src/EnergyPlus/ReportSizingManager.cc
mjwitte: You're right, even with a tstat, there could be zero peak. So, should these methods throw a warning if there's no peak day/time and revert to one of the other methods?

src/EnergyPlus/SimAirServingZones.cc
rraustad: Or use T and TdpFw as a trigger to choose which set point to use, temperature or humidity ratio.

src/EnergyPlus/ReportSizingManager.cc
rraustad: Maybe change the warning text to say exactly what has happened (inside the else if's below) and revert to sizing with CoolSupTemp and MassFlowAtCoolPeak?

src/EnergyPlus/SimAirServingZones.cc
mjwitte: No, don't need to try that. The humidity ratio setpoint is 0.011388 which is roughly Tdp=15.9C. There are many hours where the coil can meet the humidity ratio setpoint with a leaving temperature of 16-17C, and there are many hours where the leaving air is saturated. That's why this needs to switch gears and can't just use Tdp.

rraustad: Hmm, how can Tdp come out lower than supply air T? If a dry coil then yes, but a wet coil? Not your problem at this point.

mjwitte: 'cause the water temp is well below Tdp, and some of the air bypasses the coil, so the resulting mix can be drier than saturation.

mjwitte:
So, the diffs are not good. The TemperatureAndHumidityRatio files with autosized controller tolerance now have a smaller tolerance by 10x which leads to max iteration problems, and the ones with a fixed controller tolerance of, e.g. 0.002 are too loose on the humidity control. Need to figure out how to have a variable tolerance - where that lives in the controller rootfinder stuff.

rraustad:
Here's some data from the HVAC lab that does show supply air Tdp is lower than T dry-bulb. The TXV modulates the refrigerant outlet temperature and superheat varies from around 5F to 13F.

image

mjwitte:
@Myoldmopar @rraustad This is ready for review. See diffs discussion at top of PR.

rraustad:
7001 looks good.

image

src/EnergyPlus/HVACControllers.cc
```diff
@@ -2324,52 +2241,33 @@ namespace HVACControllers {
...
+            if (IsConvergedFlag) {
+                if (thisController.ControlVar == iTemperatureAndHumidityRatio) {
+                    // For temperature and humidity control, after temperature control is converged, check if humidity setpoint is met
+                    if (!thisController.HumRatCtrlOverride) {
+                        // For humidity control tolerance, always use 0.0001 which is roughly equivalent to a 0.015C change in dewpoint
+                        if (Node(thisController.SensedNode).HumRat > (Node(thisController.SensedNode).HumRatMax + 1.0e-4)) {
+                            // Turn on humdity control and restart controller
+                            IsConvergedFlag = false;
+                            thisController.HumRatCtrlOverride = true;
+                            if (thisController.Action == iReverseAction) {
```
rraustad, line 2258: This seems redundant but I'm OK with it. Controller should spit out warnings that it can't control temperature much less humidity.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -10574,7 +10574,9 @@ namespace UnitarySystems {
                                (this->m_TESOpMode == PackagedThermalStorageCoil::OffMode ||
                                 this->m_TESOpMode == PackagedThermalStorageCoil::ChargeOnlyMode)) {
                         PartLoadFrac = 0.0;
-                    } else {
+                    } else if (!SensibleLoad) {
```
rraustad, line 10577: Highlights an issue with all coils types when no sensible load exists and a latent load does. Logic at top of function could skip sensible PLR calcs and go directly to latent calcs. Sensible calcs are inside of if(SensibleLoad || LatentLoad) block.

tst/EnergyPlus/unit/DXCoils.unit.cc
```diff
@@ -252,6 +252,9 @@ TEST_F(EnergyPlusFixture, DXCoils_Test1)
     DXCoilPartLoadRatio.allocate(1);
     DXCoilFanOpMode.allocate(1);
 
+    DataLoopNode::Node.allocate(1);
+    DXCoil(CoilIndex).AirOutNode = 1;
+
```
rraustad, line 257: Wow, previously no Node calls inside the Calc routine, I wouldn't have guessed that but I guess Init sets up node data variables and uses them in Calc. Wonder what the node data variable shows in the Calc routine. Oh well, unit test passes what it's intended to do.

tst/EnergyPlus/unit/ReportSizingManager.unit.cc
```diff
@@ -169,27 +173,85 @@ TEST_F(EnergyPlusFixture, ReportSizingManager_GetCoilDesFlowT)
...
+
+    DataSizing::SysSizInput(1).CoolingPeakLoadType = DataSizing::TotalCoolingLoad;
+    DataSizing::FinalSysSizing(1).CoolingPeakLoadType = DataSizing::TotalCoolingLoad;
+
+    // Single path for VAV
+    DataSizing::SysSizInput(1).CoolCapControl = DataSizing::VAV;
+    ReportSizingManager::GetCoilDesFlowT(sysNum, CpAir, designFlowValue, designExitTemp);
+    EXPECT_FALSE(has_err_output(true));
+    EXPECT_DOUBLE_EQ(DataSizing::FinalSysSizing(1).CoolSupTemp, designExitTemp);
+    EXPECT_DOUBLE_EQ(0.002, designFlowValue);
```
rraustad, line 230: Must be some 0's in the data used to calculate designFlowValue = 0.002, but it passes.

rraustad (approved):
All looks good. Reviewed code and unit tests.

src/EnergyPlus/PlantUtilities.cc
```diff
@@ -402,6 +402,12 @@ namespace PlantUtilities {
         // Set flow on node and branch while honoring constraints on actuated node
 
         auto &a_node(DataLoopNode::Node(ActuatedNode));
+        if (LoopNum == 0 || LoopSideNum == 0) {
+            // early in simulation before plant loops are setup and found
+            a_node.MassFlowRate = CompFlow;
+            return;
+        }
```
rraustad, line 409: This must have fixed something?

tst/EnergyPlus/unit/ReportSizingManager.unit.cc
mjwitte: Umm, this shouldn't be in here

src/EnergyPlus/PlantUtilities.cc
mjwitte: And neither should this be here.

tst/EnergyPlus/unit/ReportSizingManager.unit.cc
mjwitte: Oh wait - never mind - all good here. Forgot what was in here.

tst/EnergyPlus/unit/DXCoils.unit.cc
mjwitte: Other than the condenser inlet node, CalcDoe2DXCoil doesn't read or write to any nodes. It uses variables like `DXCoil(DXCoilNum).InletAirTemp`.

src/EnergyPlus/PlantUtilities.cc
mjwitte: Oh- right - this fixed an array bounds problem. https://github.com/NREL/EnergyPlus/commit/d136c4c8b92bba4f0f873770ff63c4b81679bfef
@rraustad Thanks for the review and merge.

rraustad:
@mjwitte @jmarrec @Myoldmopar my fault for not closing the issues, however, if the description had [one of these keywords](https://help.github.com/en/articles/closing-issues-using-keywords) then those issues closed just now would have automatically closed when the PR was merged. I was getting used to that happening and forgot to check to make sure the issues were closed when the PR was merged. "Addresses" is not one of those keywords.

mjwitte:
Yeah, well, sometimes I don't like the automatic closing, but I suppose I should conform.

rraustad:
I have no problem with closing issues as we have in the past. I just thought from recent past experience that we were doing it this way now and mistakenly thought those issues would automatically close. Do it either way, I just have to be vigilant to always check.

PR #7216, opened 2019-03-08
Sequential loading with unitary system sequentially loading again

### LINKED ISSUE #7186: Sequential Loading with Unitary System not working
Issue overview
--------------
The new sequential load fractions added in #7171 caused diffs in testfile House-2FurnaceAC-SequentialLoad. Need to correct that before v9.1.0 release.

### Details

### Checklist
 - [x] Defect file added (list location of defect file here) ... already in repo `testfiles/House-2FurnaceAC-SequentialLoad.idf`

nmerket:
Fixes #7186. This addresses the diffs in the House-2FurnaceAC-SequentialLoad file that were introduced in #7171 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


nmerket:
@mjwitte It looks like we're getting close on this. From what I can see, I need to 
- [x] Fix that unit test that's failing
- [x] Look closer at the results in `House-2FurnaceAC-SequentialLoad` and `House-2FurnaceAC-SequentialLoadFractions` to verify that they're doing what's expected.

nmerket:
@mjwitte @Myoldmopar This is ready to go in. There are diffs in 2 files and they are expected and an intentional improvement. 

## House-2FurnaceAC-SequentialLoad

This was the reason for doing this bugfix. As you can see below, previously the setpoints were being overshot. Now the equipment is sequencing correctly again.

sequential_nofrac

## House-2FurnaceAC-SequentialLoadFractions

This one wasn't quite working correctly before either. As you can see it was previously overshooting setpoints as well. This one has sequential load fractions so you can see the second equipment taking 25% of the load as specified.

sequential_frac

mjwitte:
Deleted failed linux results - hope it runs ok this time. Curious, however, why there are no eso diffs here?  Seems there should be.

nmerket:
When I run this on my own there _are_ eso diffs. And only in the two files I plotted above. Those are the plots of the ESO diffs. I'm it looks like we're behind develop again. I'm going to merge in and push to kick off another round of CI. What I'm seeing here doesn't make sense.

mjwitte:
@nmerket Make sure all of your code changes are committed?

nmerket:
They are.

mjwitte:
Still puzzled that there are no eso diffs here. Will test locally.

mjwitte:
I see why the eso diffs don't show up - the csv headers don't match. @Myoldmopar That's one more thing that would be nice to add to the CI reporting.

mjwitte (approved):
Oh - I see now - the headers align, but the rows do not, because the messed up version (current develop) downsteps and the reporting is for each call (detailed). The good news is that House-2FurnaceAC-SequentialLoad.idf in this branch matches what we had in v9.0.1.

House-2FurnaceAC-SequentialLoadFractions works well as-is. 

But flipping the order of the list (keeping the same sequence numbers, so now Furnace 2 comes before 1) does not give the same result. And adding an electric baseboard at the top of the list does not work for cooling - I suspect there may still be confusion between list order and sequenced load order. 

I'm inclined to merge this as-is and post a followup issue. @nmerket agree?

mjwitte:
Here are the two test files.
[House-2FurnaceAC-SequentialLoadFractions-AddBB.idf.txt](https://github.com/NREL/EnergyPlus/files/2971328/House-2FurnaceAC-SequentialLoadFractions-AddBB.idf.txt)
[House-2FurnaceAC-SequentialLoadFractions-After-FlipOrder.idf.txt](https://github.com/NREL/EnergyPlus/files/2971329/House-2FurnaceAC-SequentialLoadFractions-After-FlipOrder.idf.txt)

mjwitte:
@rraustad Please review the changes here in UnitarySystem and Furnaces. Some of the new zone equipment load distribution options require extra iterations to propagate the correct remaining load up the chain of sequenced loads. The hard iteration limits for the oscillation checks in furnace and unitary needed to be increased. Then I hit an array bounds limit, so I changed these again. Please confirm that these capture the original intent of the oscillation check.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -8776,22 +8775,26 @@ namespace UnitarySystems {
...
-                this->m_IterationMode[this->m_IterationCounter] = NoCoolHeat;
+            // push iteration mode stack and set current mode
+            this->m_IterationMode[2] = this->m_IterationMode[1];
+            this->m_IterationMode[1] = this->m_IterationMode[0];
+            if (CoolingLoad) {
+                this->m_IterationMode[0] = CoolingMode;
+            } else if (HeatingLoad) {
+                this->m_IterationMode[0] = HeatingMode;
+            } else {
+                this->m_IterationMode[0] = NoCoolHeat;
```
rraustad, line 8786: This looks all good and tidy. I just kept all previous iterations and this changed to just the 3 needed for the evaluation.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -8776,22 +8775,26 @@ namespace UnitarySystems {
...
                 HeatingLoad = false;
-            } else if (this->m_IterationCounter > 6) {                // attempt to lock output (air flow) if oscillations are detected
-                int OperatingMode = this->m_IterationMode[7];         // VS systems can take a few more iterations than single-speed systems
-                int OperatingModeMinusOne = this->m_IterationMode[6]; // previously tested 5th iteration, now tests 7th
-                int OperatingModeMinusTwo = this->m_IterationMode[5];
+            } else if (this->m_IterationCounter > (DataHVACGlobals::MinAirLoopIterationsAfterFirst + 6)) {
+                // attempt to lock output (air flow) if oscillations are detected
+                int OperatingMode = this->m_IterationMode[0]; // VS systems can take a few more iterations than single-speed systems
+                int OperatingModeMinusOne = this->m_IterationMode[1];
+                int OperatingModeMinusTwo = this->m_IterationMode[2];
```
rraustad, line 8797: The intent was to capture the previous iteration action and to test for an oscillating condition. This appears to still do this so I would agree when this change.

mjwitte:
@rraustad Thanks! Merging.

PR #7221, opened 2019-03-12
Fixes duplicate line in the desuperheater IO documentation.

shorowit:
Fixes duplicate line in the desuperheater IO documentation.

This can be seen at: https://bigladdersoftware.com/epx/docs/9-0/input-output-reference/group-heating-and-cooling-coils.html#field-heating-source-object-type-1


Myoldmopar:
Well, something went awry with the GitHub status API here while this PR was being tested.  It looks like none of the CI runs ever got set to completed, but when looking at the dashboard, this branch tests fine...as it should.  Merging.

PR #7226, opened 2019-03-14
Cleanup Documentation Build

Myoldmopar:
Pull request overview
---------------------
I built all the docs locally and found there were a chunk of errors and warnings that are flying under the CI radar, as Decent CI doesn't have a parser that is looking specifically for LaTeX errors.  This really got on the radar because it tries to flag a certain LaTeX error as a Python error...

Anyway, this cleans up a ton of the things I found in the LaTeX builds as well as some other cleanups.  I am hopeful that CI will like this set of changes and GCC will build OK here so this can drop in very easily.


Myoldmopar:
Yay!  No more divide by zero! All green (except the integration timeout(s) likely)

image

Myoldmopar:
@mjwitte if you are available, feel free to review and merge this in as soon as integration coverage is done (likely another 1-2 hours since it started like 45 minutes ago).  If it's clean and not merged after a while later, I'll go ahead and merge this in, as it will instantly start making other branches nice and shiny green again.

PR #7227, opened 2019-03-14
Fix daylighting crash caused by array allocation error

### LINKED ISSUE #7223: Different number of daylighting reference points in zone can cause crash in DaylightingManager
Issue overview
--------------
User file has multiple daylighting objects with different number of reference points. Simulation crashed in debug when a zone with more than 1 reference point was encountered.
This part of the code does not account for different number of reference points (NREFPT) when allocating the tmp* arrays. The allocation uses the first zones number of daylighting reference points to allocate these arrays. I added the +1 to get around the crash while debugging a file.

    void  DayltgInteriorIllum(int &ZoneNum) // Zone number

        NREFPT = ZoneDaylight(ZoneNum).TotalDaylRefPoints;

        // Three arrays to save original clear and dark (fully switched) states'
        //  zone/window daylighting properties.
        if (DayltgInteriorIllum_firstTime) {
            int const d1(max(maxval(Zone, &ZoneData::NumSubSurfaces), maxval(ZoneDaylight, &ZoneDaylightCalc::NumOfDayltgExtWins)));
            tmpIllumFromWinAtRefPt.allocate(d1, 2, NREFPT + 1);
            tmpBackLumFromWinAtRefPt.allocate(d1, 2, NREFPT + 1);
            tmpSourceLumFromWinAtRefPt.allocate(d1, 2, NREFPT + 1);
            DayltgInteriorIllum_firstTime = false;
        }

Assert failure occurred here:

        // Added TH 1/21/2010 - save the original clear and dark (fully switched) states'
        //  zone daylighting values, needed for switachable glazings
       tmpIllumFromWinAtRefPt(loop, IS, IL) = ZoneDaylight(ZoneNum).IllumFromWinAtRefPt(loop, IS, IL);
       tmpBackLumFromWinAtRefPt(loop, IS, IL) = ZoneDaylight(ZoneNum).BackLumFromWinAtRefPt(loop, IS, IL);
       tmpSourceLumFromWinAtRefPt(loop, IS, IL) = ZoneDaylight(ZoneNum).SourceLumFromWinAtRefPt(loop, IS, IL);

Here's a snippet of 2 daylighting objects:

### Zone with 1 ref point:

    Daylighting:Controls,
    Thermal Zone: 000_SEG DaylightingControls, !- Name
    Thermal Zone: 000_SEG,                  !- Zone Name
    ,                                       !- Daylighting Method
    ,                                       !- Availability Schedule Name
    Continuous,                             !- Lighting Control Type
    0.3,                                    !- Minimum Input Power Fraction for Continuous or ContinuousOff Dimming Control
    0.2,                                    !- Minimum Light Output Fraction for Continuous or ContinuousOff Dimming Control
    ,                                       !- Number of Stepped Control Steps
    ,                                       !- Probability Lighting will be Reset When Needed in Manual Stepped Control
    000_SEG daylighting control,            !- Glare Calculation Daylighting Reference Point Name
    0,                                      !- Glare Calculation Azimuth Angle of View Direction Clockwise from Zone y-Axis {deg}
    ,                                       !- Maximum Allowable Discomfort Glare Index
    ,                                       !- DElight Gridding Resolution {m2}
    000_SEG daylighting control,            !- Daylighting Reference Point Name 1
    1,                                      !- Fraction of Zone Controlled by Reference Point 1
    500;                                    !- Illuminance Setpoint at Reference Point 1 {lux}

    Daylighting:ReferencePoint,
    000_SEG daylighting control,            !- Name
    Thermal Zone: 000_SEG,                  !- Zone Name
    -3.65386581735227,                      !- X-Coordinate of Reference Point {m}
    -0.24523522484613,                      !- Y-Coordinate of Reference Point {m}
    0.7874;                                 !- Z-Coordinate of Reference Point {m}

### Zone with 2 ref points:

    Daylighting:Controls,
    Thermal Zone: 000_RESTAURANTE DaylightingControls, !- Name
    Thermal Zone: 000_RESTAURANTE,          !- Zone Name
    ,                                       !- Daylighting Method
    ,                                       !- Availability Schedule Name
    ContinuousOff,                          !- Lighting Control Type
    0.3,                                    !- Minimum Input Power Fraction for Continuous or ContinuousOff Dimming Control
    0.2,                                    !- Minimum Light Output Fraction for Continuous or ContinuousOff Dimming Control
    ,                                       !- Number of Stepped Control Steps
    1,                                      !- Probability Lighting will be Reset When Needed in Manual Stepped Control
    000_RESTAURANTE daylighting control,    !- Glare Calculation Daylighting Reference Point Name
    0,                                      !- Glare Calculation Azimuth Angle of View Direction Clockwise from Zone y-Axis {deg}
    ,                                       !- Maximum Allowable Discomfort Glare Index
    ,                                       !- DElight Gridding Resolution {m2}
    000_RESTAURANTE daylighting control,    !- Daylighting Reference Point Name 1
    1,                                      !- Fraction of Zone Controlled by Reference Point 1
    500,                                    !- Illuminance Setpoint at Reference Point 1 {lux}
    Daylighting Control 1,                  !- Daylighting Reference Point Name 2
    0,                                      !- Fraction of Zone Controlled by Reference Point 2
    500;                                    !- Illuminance Setpoint at Reference Point 2 {lux}

    Daylighting:ReferencePoint,
    000_RESTAURANTE daylighting control,    !- Name
    Thermal Zone: 000_RESTAURANTE,          !- Zone Name
    -6.56237537382633,                      !- X-Coordinate of Reference Point {m}
    -0.306084574094293,                     !- Y-Coordinate of Reference Point {m}
    0.7874;                                 !- Z-Coordinate of Reference Point {m}

    Daylighting:ReferencePoint,
    Daylighting Control 1,                  !- Name
    Thermal Zone: 000_RESTAURANTE,          !- Zone Name
    -6.51704177922846,                      !- X-Coordinate of Reference Point {m}
    -1.06464183917781,                      !- Y-Coordinate of Reference Point {m}
    0.762;                                  !- Z-Coordinate of Reference Point {m}

### Details

### Checklist
 - [x] Defect file added (EnergyPlusDevSupport\DefectFiles\7223\)

rraustad:
Pull request overview
---------------------
Fixes #7223. Correct allocation of Daylighting array that causes crash and move a few common arrays to a one time allocate. This issue appears to impact only simulations using switchable glazing.

Work around: Use the same number of daylighting reference points in each zone.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/DaylightingManager.cc
+++ b/src/EnergyPlus/DaylightingManager.cc
@@ -173,7 +173,6 @@ namespace DaylightingManager {
     bool ReportIllumMap_firstTime(true);
     bool SQFirstTime(true);
 
-
     // Surface count crossover for using octree algorithm
     // The octree gives lower computational complexity for much higher performance
     //  as the surface count increases but has some overhead such that the direct
@@ -549,10 +548,10 @@ namespace DaylightingManager {
         // na
 
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
-        int ZoneNum; // Zone number
-        int IHR;     // Hour of day counter
-        int IWin;    // Window counter
-        int loop;    // DO loop indices
+        int ZoneNum;     // Zone number
+        int IHR;         // Hour of day counter
+        int IWin;        // Window counter
+        int loop;        // DO loop indices
         Real64 DaylFac1; // sky daylight factor at ref pt 1
         Real64 DaylFac2; // sky daylight factor at ref pt 2
 
@@ -1132,18 +1131,17 @@ namespace DaylightingManager {
                 HourOfDay, {1, MaxSlatAngs + 1}, {1, 4}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
             ZoneDaylight(ZoneNum).DaylBackFacSky(
                 HourOfDay, {1, MaxSlatAngs + 1}, {1, 4}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
-            ZoneDaylight(ZoneNum).DaylIllFacSun(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) =
+            ZoneDaylight(ZoneNum).DaylIllFacSun(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
+            ZoneDaylight(ZoneNum).DaylIllFacSunDisk(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) =
                 0.0;
-            ZoneDaylight(ZoneNum).DaylIllFacSunDisk(
-                HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
-            ZoneDaylight(ZoneNum).DaylSourceFacSun(
-                HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
+            ZoneDaylight(ZoneNum).DaylSourceFacSun(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) =
+                0.0;
             ZoneDaylight(ZoneNum).DaylSourceFacSunDisk(
                 HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
             ZoneDaylight(ZoneNum).DaylBackFacSun(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) =
                 0.0;
-            ZoneDaylight(ZoneNum).DaylBackFacSunDisk(
-                HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) = 0.0;
+            ZoneDaylight(ZoneNum).DaylBackFacSunDisk(HourOfDay, {1, MaxSlatAngs + 1}, {1, numRefPts}, {1, ZoneDaylight(ZoneNum).NumOfDayltgExtWins}) =
+                0.0;
         }
 
         NRF = ZoneDaylight(ZoneNum).TotalDaylRefPoints;
@@ -6436,9 +6434,15 @@ namespace DaylightingManager {
         //  zone/window daylighting properties.
         if (DayltgInteriorIllum_firstTime) {
             int const d1(max(maxval(Zone, &ZoneData::NumSubSurfaces), maxval(ZoneDaylight, &ZoneDaylightCalc::NumOfDayltgExtWins)));
-            tmpIllumFromWinAtRefPt.allocate(d1, 2, NREFPT);
-            tmpBackLumFromWinAtRefPt.allocate(d1, 2, NREFPT);
-            tmpSourceLumFromWinAtRefPt.allocate(d1, 2, NREFPT);
+            tmpIllumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
+            tmpBackLumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
+            tmpSourceLumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
+
+            SetPnt.allocate(maxNumRefPtInAnyZone);
+            DaylIllum.allocate(maxNumRefPtInAnyZone);
+            GLRNDX.allocate(maxNumRefPtInAnyZone);
+            GLRNEW.allocate(maxNumRefPtInAnyZone);
+
             DayltgInteriorIllum_firstTime = false;
         }
         tmpIllumFromWinAtRefPt = 0.0;
@@ -6446,10 +6450,6 @@ namespace DaylightingManager {
         tmpSourceLumFromWinAtRefPt = 0.0;
 
         // FLOW:
-        SetPnt.allocate(NREFPT);
-        DaylIllum.allocate(maxNumRefPtInAnyZone);
-        GLRNDX.allocate(NREFPT);
-        GLRNEW.allocate(NREFPT);
 
         // Initialize reference point illuminance and window background luminance
         for (IL = 1; IL <= NREFPT; ++IL) {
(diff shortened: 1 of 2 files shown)


rraustad:
Unit test crashes when tmp arrays are allocated as before (i.e., change the 3 lines back to tmpIllumFromWinAtRefPt.allocate(d1, 2, NREFPT);  and unit test crashes). With correction to tmp array allocate, unit test passes. I don't recommend running the defect file in debug mode since it's very very slow. I have done that and it passes the point of the crash.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -6436,20 +6434,22 @@ namespace DaylightingManager {
         //  zone/window daylighting properties.
         if (DayltgInteriorIllum_firstTime) {
             int const d1(max(maxval(Zone, &ZoneData::NumSubSurfaces), maxval(ZoneDaylight, &ZoneDaylightCalc::NumOfDayltgExtWins)));
-            tmpIllumFromWinAtRefPt.allocate(d1, 2, NREFPT);
-            tmpBackLumFromWinAtRefPt.allocate(d1, 2, NREFPT);
-            tmpSourceLumFromWinAtRefPt.allocate(d1, 2, NREFPT);
+            tmpIllumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
+            tmpBackLumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
+            tmpSourceLumFromWinAtRefPt.allocate(d1, 2, maxNumRefPtInAnyZone);
```
rraustad, line 6439: Major change is here where tmp arrays are allocated to max number of reference points instead of number of reference points from the first zone called.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -6436,20 +6434,22 @@ namespace DaylightingManager {
...
+
+            SetPnt.allocate(maxNumRefPtInAnyZone);
+            DaylIllum.allocate(maxNumRefPtInAnyZone);
+            GLRNDX.allocate(maxNumRefPtInAnyZone);
+            GLRNEW.allocate(maxNumRefPtInAnyZone);
```
rraustad, line 6444: Minor change is moving these arrays up into one time init block.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -6383,8 +6381,8 @@ namespace DaylightingManager {
         static Array2D<Real64> SFSKHR(2, 4); // Sky source luminance factor for sky type (second index),
         //   bare/shaded window (first index)
         static Vector2<Real64> SFSUHR; // Sun source luminance factor for bare/shaded window
-        Array1D<Real64> GLRNDX;        // Glare index at reference point
-        Array1D<Real64> GLRNEW;        // New glare index at reference point
+        static Array1D<Real64> GLRNDX; // Glare index at reference point
+        static Array1D<Real64> GLRNEW; // New glare index at reference point
```
rraustad, line 6385: Also made these 3 arrays static. I'm sure there could be more work in this function but these changes meet the needs to avoid the crash.

rraustad:
Hoping to see all green from CI. If so, I feel this can go in. Unless, or course, there are review comments.

src/EnergyPlus/DaylightingManager.cc
Myoldmopar: This makes good sense.  Good catch.

src/EnergyPlus/DaylightingManager.cc
Myoldmopar: Is there any chance that the newly static nature of these arrays could cause problems for other unit tests?  These will stay allocated, etc., between multiple unit tests.

mbadams5: Static function members should really be avoided. I'd move them up to the namespace like we have for many other instances like this.

rraustad: Not sure what you guys mean. There were several local arrays that were already static inside this function (a function which is not static, the hh file shows `void DayltgInteriorIllum(int &ZoneNum);`. I just made these local arrays static as well. And the existing static arrays are not deallocated in the clear_state function. Should they be? So what is the recommendation here?

rraustad: Oh, and as far as the daylighting unit test go, none were calling this function but my change to one unit test does now, but it's the only one. So I think I get what you mean now. That means those existing static arrays would have caused a problem but unit testing never saw that since none called this function.

rraustad:
Can anyone tell me what was happening in release mode? The data was obviously not stored in the tmp arrays so the answer was wrong for the zone with excess daylighting reference points. So without the asserts used in debug mode, what really happens in release mode?

Myoldmopar:
Well, it was using uninitialized memory.  The Objexx array will have a variable pointing to a location of memory where it starts.  As you loop over the items in the array, it calculates the right memory location for element n as an offset from the original zeroth element.  In debug mode, there is also a trap to ensure that the index is less than or equal to the max.  In release mode, as you mentioned, there is no such check.  So the runtime still calculates an offset from the base location, and grabs the memory there.  But what is at that location?  Who knows?  Maybe a node temperature?  Maybe a refrigerated case name?  Maybe something else?  Depending on the calculated location, it could cause a segmentation fault because you accessed memory that wasn't in your process space, or it could cause a NaN if the bits cannot be converted into a meaningful floating point value.  Or, as in this case, it might've found some bits that could be converted into a number and it just ran with them.

rraustad:
The CI unit tests run in release mode so why then would this fail? I tested in debug mode, not release mode. This definitely makes me think! Maybe I should test the previous code with the new unit test.

Myoldmopar:
CI runs many different tests:
- Windows 64 bit release mode (runs integration, regression, and unit tests)
- WIndows 32 bit release mode (runs integration, regression, and unit tests)
- Linux GCC release mode (runs integration, regression, and unit tests)
- Linux Debug Unit Test Coverage (just runs unit tests in debug mode)
- Linux Debug Integration Test Coverage (just runs integration tests in debug mode)
- Linux CppCheck
- Linux CustomCheck
- Mac release mode (runs integration, regression, and unit tests)

With uninitialized memory due to an array bounds error, it has the potential to:
- Definitely fail in debug mode as an assertion will be thrown
- Potentially fail, indeterminately, in release mode, as the uninitialized memory could be anything

rraustad:
So then Linux debug would catch the assert and Windows/Mac would probably pass. I think access outside of the process space would be rare if it's accessing sequential memory.

Myoldmopar:
Rare, sure, but I mean, *some* data has to be right at the end of the road...

Myoldmopar:
Also, about the static data.  I think you nailed the understanding of why statics are generally bad.  Many unit testing problems we've had have stemmed from static variables not able to be reset within a function.  They should all just be hoisted out of the function and into the namespace so that a clear_state function can reset them.  It's usually quite a simple process, and can save many headaches down the road.

rraustad:
Gotcha. I am just waiting for CI results. Last few weeks CI has been processing like a banshee. As soon as I get back in and commit they came to a crashing halt?

Myoldmopar:
Yeah thanks a lot!!! 

:upside_down_face: @mjwitte and I are currently trying to figure out wtf happened this afternoon to CI.  Something weird...

rraustad:
So here's another question. Clear_state resets the one time init flag where these arrays are allocated. So with the second unit test, when these are again allocated, aren't they allocated correctly?

Myoldmopar:
@rraustad so we found the problem with CI.  It turns out it is stumbling on the branch name with the hash.  We had this working for a while, but Linux has been updated since then, and with it all the Ruby libraries that run CI were updated too.  And they've changed slightly in this respect.  I am going to create a branch off of this one without the hash at the beginning, then delete this branch.  CI should pick things up and run after that.

Myoldmopar:
And I'm glad the clear_state is managing it so it runs OK.  With that in place, there really shouldn't be any issues, and it boils back down to just the idea that we'd like to eliminate statics as we go along.

rraustad:
You mean my branch with #. Sorry.

Myoldmopar:
No big deal, it definitely used to work.  Just some coincidences about upgrades to CI and timing.  One less character for you to type when making branches!!!

rraustad:
This one was supposed to be a simple fix to get into the next release. #newBranchName.

rraustad:
Not sure which one of [these](http://www.asciitable.com/) I will try next as the first or even intermittent characters. Probably not anytime soon.

PR #7239, opened 2019-03-20
Have ceiling diffuser convection algorithm revert to natural convection when the system is off or at low air change rates.

### LINKED ISSUE #7208: CeilingDiffuser convection algorithm does not revert to natural convection algorithm when the system is off
Issue overview
--------------
The [documentation](https://bigladdersoftware.com/epx/docs/9-0/input-output-reference/group-simulation-parameters.html#field-algorithm-000) states:

> To avoid discontinuities in surface heat transfer rate calculations, all of correlations have been extrapolated beyond the lower limit of the data set (3 ACH) to a natural convection limit that is applied during the hours when the system is off.

However, in the source code, this only appears in [CalcCeilingDiffuserInletCorr](https://github.com/NREL/EnergyPlus/blob/a310a424e4ea98bb8183137f0fd5f1012a6bd918/src/EnergyPlus/ConvectionCoefficients.cc#L4430), a function that is never called ([though it looks like it was intended for use if it could be accommodated](https://github.com/NREL/EnergyPlus/blob/a310a424e4ea98bb8183137f0fd5f1012a6bd918/src/EnergyPlus/ConvectionCoefficients.cc#L4358)).

Instead, EnergyPlus uses [CalcCeilingDiffuserIntConvCoeff](https://github.com/NREL/EnergyPlus/blob/a310a424e4ea98bb8183137f0fd5f1012a6bd918/src/EnergyPlus/ConvectionCoefficients.cc#L4267), which never calls a natural convection algorithm.

### Checklist

nealkruis:
Pull request overview
---------------------
Fixes #7208 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types

### DIFF
--- a/src/EnergyPlus/ConvectionCoefficients.cc
+++ b/src/EnergyPlus/ConvectionCoefficients.cc
@@ -375,7 +375,7 @@ namespace ConvectionCoefficients {
                 // Ceiling Diffuser and Trombe Wall only make sense at Zone Level
                 // Interior convection coeffs are first calculated here and then at surface level
                 if (SELECT_CASE_var == CeilingDiffuser) {
-                    CalcCeilingDiffuserIntConvCoeff(ZoneNum);
+                    CalcCeilingDiffuserIntConvCoeff(ZoneNum, SurfaceTemperatures);
 
                 } else if (SELECT_CASE_var == TrombeWall) {
                     CalcTrombeWallIntConvCoeff(ZoneNum, SurfaceTemperatures);
@@ -4103,7 +4103,13 @@ namespace ConvectionCoefficients {
         return ACH;
     }
 
-    Real64 CalcCeilingDiffuserIntConvCoeff(Real64 Tilt, Real64 ACH)
+    Real64 CalcCeilingDiffuserIntConvCoeff(Real64 const ACH,  // [1/hr] air system air change rate
+                                           Real64 const Tsurf,
+                                           Real64 const Tair,
+                                           Real64 const cosTilt,
+                                           Real64 const humRat,
+                                           Real64 const height,
+                                           bool const isWindow)
     {
         // SUBROUTINE INFORMATION:
         //       AUTHOR         Rick Strand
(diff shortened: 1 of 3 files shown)


src/EnergyPlus/ConvectionCoefficients.cc
```diff
@@ -5741,12 +5751,15 @@ namespace ConvectionCoefficients {
 
         Real64 tmpHc = 0.0;
 
+        int const ZoneNum = Surface(SurfNum).Zone;
+        Real64& Tsurface = TH(2, 1, SurfNum);
+        Real64& Tzone = MAT(ZoneNum);
```
Myoldmopar, line 5756: :+1:

Myoldmopar (approved):
Overall the code changes are fine.  A little bigger than I expected, but it looks like things in the convection world are really starting to get unified.  Unit test changes help with confidence too.

Myoldmopar:
Looking at the CI results, I'm a little concerned about the magnitude of the changes.  For example, on the small office model, the [eio diffs](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2019-03/ed2b4a0b43f856349144b155ae3a798c0efc6db6-bfd70a8eb9ab849f8cebcddbe8947412dd9fab64/_SmallOffice_Dulles/x86_64-MacOS-10.13-clang/eplusout.eio.diff.html) show significantly different calculated design loads:

```
 ! <Zone Sizing Information>, Zone Name, Load Type, Calc Des Load {W} ...
- Zone Sizing Information, SOUTHZONE_1STFLOOR, Cooling, 4885.01562 ...
- Zone Sizing Information, SOUTHZONE_1STFLOOR, Heating, 9635.95393 ...
- Zone Sizing Information, NORTHZONE_1STFLOOR, Cooling, 3804.39581 ...
- Zone Sizing Information, NORTHZONE_1STFLOOR, Heating, 9619.63904 ...
- Zone Sizing Information, SOUTHZONE_2NDFLOOR, Cooling, 9530.03189 ...
- Zone Sizing Information, SOUTHZONE_2NDFLOOR, Heating, 14560.4150 ...
- Zone Sizing Information, NORTHZONE_2NDFLOOR, Cooling, 7833.80161 ...
- Zone Sizing Information, NORTHZONE_2NDFLOOR, Heating, 14568.3519 ...
+ Zone Sizing Information, SOUTHZONE_1STFLOOR, Cooling, 5673.85539 ...
+ Zone Sizing Information, SOUTHZONE_1STFLOOR, Heating, 9454.37757 ...
+ Zone Sizing Information, NORTHZONE_1STFLOOR, Cooling, 4398.55637 ...
+ Zone Sizing Information, NORTHZONE_1STFLOOR, Heating, 9469.01800 ...
+ Zone Sizing Information, SOUTHZONE_2NDFLOOR, Cooling, 10278.8394 ...
+ Zone Sizing Information, SOUTHZONE_2NDFLOOR, Heating, 14313.6896 ...
+ Zone Sizing Information, NORTHZONE_2NDFLOOR, Cooling, 8534.98458 ...
+ Zone Sizing Information, NORTHZONE_2NDFLOOR, Heating, 14324.2456 ...
```

I'm not saying it's wrong, I'm just saying I'm skiddish about dropping this in at the last minute.  I'm curious about how @mjwitte and maybe @RKStrand would feel about this amount of change just from switching the convection algorithm at off and low ACH conditions.

mjwitte:
Well, that does raise a question: does this convection method see an airflow during sizing? If not, it should.

nealkruis:
[ACH is set to zero during sizing](https://github.com/NREL/EnergyPlus/blob/bfd70a8eb9ab849f8cebcddbe8947412dd9fab64/src/EnergyPlus/ConvectionCoefficients.cc#L4017), meaning that the ceiling diffuser algorithms were limited by the constant term previous to this fix.

mjwitte:
Well, that explains the big diffs, but that seems wrong. There are flow rates during sizing - shouldn't we use them?

nealkruis:
BTW, this fix represents the way it was impemented in the unused [`CalcCeilingDiffuserInletCorr`](https://github.com/NREL/EnergyPlus/blob/bfd70a8eb9ab849f8cebcddbe8947412dd9fab64/src/EnergyPlus/ConvectionCoefficients.cc#L4181). I think it would be better to do some sort of superposition of forced and natural convection so that there isn't a discontinuity at 3.0 ACH.

nealkruis:
@mjwitte what if zone sizing is for a convective baseboard system. Are flow rates still used during sizing?

mjwitte:
Well, sizing is ignorant of the type of equipment, it calculates peak loads and peak airflow rates at the Sizing:Zone design supply conditions. So, with this convection correlation it won't know the right choice to make and the sizing peak load won't match the simulation peak load if the equipment type doesn't align with the sizing assumption. 👎

mjwitte:
What's interesting is that the cooling peak loads shown above have increased with this fix. Are the conv coeffs at zero flow now higher than they were?

nealkruis:
I think it depends on the magnitude of the temperature differences of the surfaces.

Myoldmopar:
I feel uncertain about this, but I'm trying to parse through this discussion to determine if the uncertainty lies within the context of the changes in this branch, or a more broad context.  My gut is telling me that the changes here are OK, but we may want to have bigger discussions later about convection and sizing.  With that in mind, I'm actually inclined to merge this in.  But if I'm missing something about the specific change in this branch, then please speak up and we'll push this until the next release.

nealkruis:
I also have some unease over the continuity at 3.0 ACH, but that's how the algorithm was originally framed. I think the fix as it stands now represents the intention described in the documentation. My feeling is that we should merge this now, and revisit both the sizing questions and the forced/natural transition as separate topics in the future.

mjwitte:
Really need buy-in from @RKStrand if possible

nealkruis:
@JimPegues any thoughts here?

RKStrand:
@mjwitte You asked for my "buy-in" so here are my thoughts after looking through the comments here.  I think proceed with the fix as is but I agree with @nealkruis that perhaps a larger discussion needs to take place regarding the discontinuity at ACH 3.0.  I'm not sure how many other issues like this are out there but it is worth investigating?

Myoldmopar:
I fully agree.  Merging this now, but we need to make sure to follow up with a larger discussion about a convection cleanup.

Issue #6323, opened 2017-09-19
Cooling and Heating Simultaneous Operation in VSHeatPumpWaterToAirWithRHControl.idf Causes Severe Error


mitchute:
From issue #6316 

Heating loads don't converge after 40 days due to the VS coils operating at the same time when dehumidification is needed.

### Details
Some additional details for this issue (if relevant):
 - Platform: Windows 7 SP2
 - Version of EnergyPlus: Pre 8.8. 0c736af1ca11d92f9d639de67236f3c6051eecfb

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@mitchute @rraustad @Myoldmopar This file is still a problem, caused a timeout error in the integration coverage tests on #7163. I've added a priority label in hopes this gets some attention soon.

mjwitte:
Adding a defect file here. The problem can be seen in the csv output, row 76,  01/03  03:57:30, the pertinent columns are all >0 which should never happen:
```
AY: SYS 1   HEAT PUMP COOLING MODE:Cooling Coil Total Cooling Rate [W](Each Call) = 24916
DP: SYS 1   HEAT PUMP HEATING MODE:Heating Coil Heating Rate [W](Each Call) = 23806
EA: HEAT PUMP DX SUPP HEATING COIL 1:Heating Coil Gas Rate [W](Each Call) = 390
```
[VSHeatPumpWaterToAirWithRHControl-OnlyRunPeriod1.idf.txt](https://github.com/NREL/EnergyPlus/files/2993255/VSHeatPumpWaterToAirWithRHControl-OnlyRunPeriod1.idf.txt)
Run with Miami TMY3.

mjwitte:
After the fix (to be pushed shortly) the heat pump heating mode is always zero, cooling mode runs to dehumidify , and supp heat offsets that plus adds requested heating load. No more downstepping either on 01/03.

Myoldmopar:
Closed via #7242

PR #7242, opened 2019-03-21
Fix WAHP simultaneous heat-cool with dehumid load plus heating load

mjwitte:
Pull request overview
---------------------
Address #6323. Control logic for water-to-air heat pump did not correctly meet the zone sensible and latent loads.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/Furnaces.cc
```diff
@@ -10639,7 +10639,7 @@ namespace Furnaces {
             }
 
             if (Furnace(FurnaceNum).FurnaceType_Num != UnitarySys_HeatCool) {
-                if (QZnReq > SmallLoad) {
+                if ((QZnReq > SmallLoad) && HeatingLoad) {
```
mjwitte, line 10639: When there is a dehumidification load, `HeatingLoad` is set to false in [InitFurnace](https://github.com/NREL/EnergyPlus/blob/272b4f052b19d426a0edbb4ff570d654cb6eba5b/src/EnergyPlus/Furnaces.cc#L5308-L5320)

mjwitte: Not sure if there may be other places that need this - perhaps for the drawthrough logic path?

rraustad: Well maybe that's a mistake? Or maybe the supplemental picks up all the sensible load? I'm fairly positive that the furnace can meet a heating load and dehumidification load at the same time.

rraustad: So for a heating load, MoistureLoad would be true and HeatingLoad would be false. Then the supplemental would pick up the sensible load. I don't recall it that way.

mjwitte: If there's a non-heatpump heating coil, then yes, the furnace can do the work, but if it's a heat pump, it can only cool or heat, it can't do both. That's why this logic happens only ` if (Furnace(FurnaceNum).FurnaceType_Num != UnitarySys_HeatCool)`. And the changes here are only in the varspd heat pump section.

rraustad: Oh yeah, forgot about the DX part. So yes, supplemental will pick up the sensible load.

mjwitte:
Never mind. This fix only fixes the modified defect file. The standard VSHeatPumpWaterToAirWithRHControl test file still fails to converge on warmup and the CI diffs shows an *additional* warmup failure. Punting this to after v9.1. <sigh>

mjwitte:
Well, I was looking at the wrong diffs, but in any case, VSHeatPumpWaterToAirWithRHControl still fails to converge on the same day as before this fix.
```
+   ** Severe  ** CheckWarmupConvergence: Loads Initialization, Zone="SPACE4-1" did not converge after 40 warmup days.
+   **   ~~~   ** See Warmup Convergence Information in .eio file for details.
+   **   ~~~   ** ...Environment(SizingPeriod)="MIAMI INTL AP ANN CLG 1% CONDNS DB=>MWB"
```
So, back to the drawing board.

nrel-bot-2:
@mjwitte it has been 13 days since this pull request was last updated.

nrel-bot-2c:

nrel-bot-2c:
@mjwitte it has been 8 days since this pull request was last updated.

nrel-bot-2b:
@mjwitte it has been 8 days since this pull request was last updated.

nrel-bot-2b:

nrel-bot-2c:

nrel-bot-2b:

nrel-bot-2:

nrel-bot-2b:

nrel-bot-2:

nrel-bot-3:

rraustad:
This appears to be a simple fix. And RH control has a detrimental impact on warmup convergence (reheat is on one warmup day and off the next or runs at different PLRs on successive days). So is this ready or on hold?

mjwitte:
@rraustad I gave up on this back in April and never picked it up again. The changes here didn't solve the initial problem. This hasn't been updated since.

mjwitte:
Feel free to pick it up.

rraustad:
This is interesting. This discussion is for defect file VSHeatPumpWaterToAirWithRHControl-OnlyRunPeriod1.idf.txt.

One thing I did is change the logic in ControlVSHPOutput to be more like the newer controls. This helped somewhat then digressed re the number of max iteration warnings. Then I turned on display extra warnings and see 5 of these:

```
   ************* SizeDirectAir: Potential issue with equipment sizing for AirTerminal:SingleDuct:Uncontrolled="ZONE1DIRECTAIR".
   **   ~~~   ** User-Specified Maximum Air Flow Rate of 1.40000 [m3/s]
   **   ~~~   ** differs from Design Size Maximum Air Flow Rate of 0.30904 [m3/s]
   **   ~~~   ** This may, or may not, indicate mismatched component sizes.
   **   ~~~   ** Verify that the value entered is intended and is consistent with other components.
```

Oddly, this doesn't affect the control zone fraction, all 5 are equal to 1.0. But I do see an impact on zone flow rate which does not make sense for continuous fan operation. Maybe this is an artifact of the number of speeds and therefore air flow rate that ultimately hits the zone inlet.

```
  AirLoopHVAC:UnitaryHeatPump:WaterToAir,
    Water-To-Air Heat Pump Unit 1,  !- Name
    1.4,                     !- Supply Air Flow Rate {m3/s}
    ContinuousFanSch,        !- Supply Air Fan Operating Mode Schedule Name
    CoolReheat;              !- Dehumidification Control Type
```

image

Another thing I see is that the max iteration warnings happen when the coils are off and the zone is in the deadband. System outlet node temp is set and that temp is passed back to the inlet? and then the mixed air temp changes and the system outlet node again changes. This seems to happen on the same iteration. I see this for all 5 systems.

I need to look closer at "System outlet node temp is set and that temp is passed back to the inlet?" because the system outlet node temp is not the temp that should be passed back to the system inlet. That air needs to pass through the zone first to determine the RA temp that hits the mixer (and it should be the end of previous time step zone temp?).

Myoldmopar:
This is good information.  I appreciate you digging into this and I already learned some things from that comment alone.

rraustad:
I'm trying to get the zone T and RH profiles to calm down. What I want to point out here for these 6 warmup days and 1 weather day (1/3) is how the T profile oscillates between cooling and heating set point from one warmup day to the next. Classic RH control.

image

rraustad:
There it is, the supplemental heater wasn't modeled correctly. I will push these changes to save them and have further discussions. SimHVAC max iteration warnings are still present (annual run) so that needs to be looked at next. It runs faster now since the zone temp isn't jumping around so much (1.5 min vs 3.5 min).

image

rraustad:
CI shows 3 large diff files. All 6 diff files are related to Furnace code.

1. UnitarySystem_VSHeatPumpWaterToAirEquationFit - big diff's
2. VSHeatPumpWaterToAirEquationFit - big diff's
3. VSHeatPumpWaterToAirWithRHControl - big diff's, err file shows max iter and convergence warnings are gone now

Need to compare zone T/RH before and after. Suspect better control caused diff's.

rraustad:
Big diff file 1. Zone T/RH and Sys 3 heating coil doesn't look to have changed significantly.

image

Then I take the results from develop and this branch and diff the results (develop - this branch). Plot all data on a single graph and look for the big diff's. In this case there are 2 obvious (red positive blip and blue negative blip). Then find those and plot them to be sure the correct data was found. Then plot each of those to see if anything really changed.

image

Then find the spike, point 662 by hovering the cursor over the graph data, then replot the data zoomed in to see the diff. The big diff's in the first graph can be deleted leaving the next biggest diffs.

image

In this branch the Zone Air System Sensible Heating Energy increased, possibly the supplemental heater, and the heating coil was delayed. Here it actually looks like the heating coil operation is smoother, i.e., it was off and gradually turned on.

Going back to the zone conditions at this same time shows the impact of the increase in Zone Air System Sensible Heating Energy. From this perspective it looks like a bit of a logic issue

image

which corresponds to this point on the zone T/RH plot.

image

At this point in time where the zone T is falling to the heating set point temp, this constant fan system is providing 231.8 W of heat while the predicted load is 232.6 W. So yes, the heating coil did need to turn on. From the figure above the heating coil is not on. Looking at the supplemental heater shows there is a slight difference.

image

rraustad:
Big diff file 3 shows much better RH control although it looks like it could be better.

RH schedule use by all systems in this example file:

```
  Schedule:Compact,
    Dehumidification Seasonal Relative Humidity Sch,  !- Name
    Any Number,              !- Schedule Type Limits Name
    Through: 03/31,          !- Field 1
    Until: 24:00,50.0,       !- Field 3
    Through: 09/30,          !- Field 5
    For: AllDays,            !- Field 6
    Until: 24:00,55.0,       !- Field 7
    Through: 12/31,          !- Field 9
    For: AllDays,            !- Field 10
    Until: 24:00,50.0;       !- Field 11
```

image

image

rraustad:
This is ready for review.

tst/EnergyPlus/unit/Furnaces.unit.cc
```diff
@@ -233,11 +237,102 @@ TEST_F(EnergyPlusFixture, SetVSHPAirFlowTest_VSFurnaceFlowTest)
...
+    EXPECT_EQ(Furnace(1).MdotFurnace, 0.2); // flow rate is at idle speed flow rate
+    EXPECT_EQ(DataLoopNode::Node(1).MassFlowRate, 0.2); // furnace inlet node mass flow rate is at idle speed flow rate
+    EXPECT_EQ(Furnace(1).CoolPartLoadRatio, 0.0);
+    EXPECT_EQ(Furnace(1).HeatPartLoadRatio, 0.0);
+
+    Furnace(1).HeatPartLoadRatio = 1.0;
+    Furnaces::HeatingLoad = true;
+    CalcNewZoneHeatCoolFlowRates(
+        FurnaceNum, firstHVACIteration, compOp, zoneLoad, moistureLoad, heatCoilLoad, reheatCoilLoad, onOffAirFlowRatio, hXUnitOn);
+    EXPECT_EQ(Furnace(1).HeatPartLoadRatio, 1.0);
```
rraustad, line 333: This is as far as I got on the unit test. It runs and does test some code but there is more to do to set up coil capacity, etc., to get meaningful results. I'm just out of time.

Myoldmopar (approved):
I have no problem with these changes, and I'm excited about this going in.  I'm currently building locally, but I am optimistic after playing with it a little that this can get merged into develop.

Myoldmopar:
> It runs faster now since the zone temp isn't jumping around so much (1.5 min vs 3.5 min).

Yeah, I'm not sure about your speed up calculations.  On the Mac I'm building with, the develop branch takes 63.12s to run, while this branch it runs in 2.27s.  Literally.  That's insane.  Merging.

Thanks @rraustad !

Issue #6727, opened 2018-04-23
UnitarySystem heating coil control problem with multi-speed fan - possible sizing issue


rraustad:
Issue overview
--------------
*Summary by @mjwitte based on testing with v9.0.1, Oct 2018*

### Issues

1. Unitary system with Coil:Gas:Heating and multispeed fan, control works well for larger autosized coil, but smaller coil results in max iterations and bogus results.
2. With Coil:Heating:Gas:MultiStage control is good, but airflow is wrong, either sits at high or low, but does not split the timestep between two speeds as it should.
3. Autosizing may be oversizing the heating coil, have not confirmed that.

3 v9.0 defect files in EnergyPlusDevSupport, run them with Chicago weather.

### Details
Defect file has three AirloopHVAC:UnitarySystem with load control, each serving a single zone (focus on CoreZone here) with multi-speed fan operation. 
For heating, the fan flow is 1.0.  For no-load the fan flow is 0.25. Fan operation is continuous during occupied hours.

With autosized Coil:Heating:Fuel (83.7kW), the system runs fine and controls well, even for an annual simulation. There are some DX cooling coil flow/capacity warnings, but no max iterations problems. During heating, the thermostat setpoint is met (within 0.01C) and there is no overheating. The fan speed outputs show for each timestep that the system is splitting the timestep between coil-on 100% flow and coil-off 25% flow to meet the current zone load.

But with fixed size Coil:Heating:Fuel (33.2kW), it struggles to control during heating. There are max iteration warnings, the zone temperature is often below setpoint, and there are times when the system is oscillating between 25% flow and 100% flow. It's interesting to note that some parts of the day, it controls well, but not in others. 

Results comparison below has the smaller heating coil on the left, the larger autosized heating coil on the right. The highlighted cells tell the story. The reported mixed air temp is not consistent with the flow rate, the fan speed is all high or all low, the zone setpoint is not met, and the heating coil is unnecessarily maxed out.

![image](https://user-images.githubusercontent.com/4991075/46907107-4ff9ea80-ced3-11e8-8744-e913efa86a16.png)

Background (from April 2018)
-------------
User noticed odd results when testing Coil:Heating:Fuel and Coil:Heating:Gas:MultiStage as described in HelpDesk Ticket [12594](http://energyplus.helpserve.com/staff/Tickets/Ticket/View/12594). Defect files are attached to that ticket.

Original post from HelpDesk ticket 12594:

[...]

PR #7249, opened 2019-03-29
NFP Changeover Bypass VAV enhancements

rraustad:
Pull request overview
---------------------
New Feature: Allow AirloopHVAC:UnitaryHeatCool:ChangeoverBypassVAV to be connected to AirloopHVAC:ReturnPlenum or AirloopHVAC:ZoneMixer to better model outside air system economizer controls. Also add a new control option for LoadPriority and include a minimum runtime before allowing operating mode change (e.g., from cooling to heating). 

Differences in example files ChangeoverBypass*.idf :

During development of this feature it was found that controls deviated from the user selection of Priority Control Mode. This was corrected. In this figure (winter then summer design days), the VAV damper position is shown in the middle plot. The orange line is original model damper position where the model would switch operating modes and the damper responded accordingly. The blue line is the corrected damper position. Note that this change has little impact on zone temperature and coil operation.

image

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
Viewing the markdown document in the branch shows the figures. See the document [here](https://github.com/NREL/EnergyPlus/blob/New-Feature-CBVAV-Mods/design/FY2019/VAV_ChangeoverBypass_Enhancement.md).

Myoldmopar:
I was reviewing this offline while flying, so my apologies that these notes are not more integrated with the changes themselves.  I can refine if needed:

- Like the DOAS to multiple AHU, I feel concerned about extra iteration required to handle this situation where you are coupling nodes together
- Presumably in figure 2 you don't need a zonemixer because the return plenum collects the zone return nodes
- in figure 2 it mentions the internal OA mixer will have flow set to zero when using an AirLoopHVAC:OASys -- is there already a precedent for this?  Does the solution algorithm rely on there being an OA mixer internal there?  Seems like it could be a chance to remove some calculations that aren't necessary.

rraustad:
Answers:
1. Well, it seems adding another node to the mixer or plenum is no different than adding another zone. I am not sure if additional iterations would be needed since if the flow to the zones converges then so does the bypass flow.
2. Figure 2 is missing the notation of the mixer (forgot to update the figure in the media folder). So the Changeover Bypass VAV model (CBVAV) can connect to either a mixer (when no plenum is used) or a plenum (when no mixer is needed).
3. In CBVAV the internal mixer is required because this is where the bypass was mixed and making the internal OA mixer optional would have made things harder [I think]). Setting internal OA mixer flow = 0 is more of a suggestion since sizing won't be able to handle both OA streams. Using both would also introduce excess outdoor air. User could specify both OA flow's if desired, it's just not how a unit in the field would be configured. Not sure yet how many, if any, calculations could be skipped.

nrel-bot-2:
@rraustad it has been 14 days since this pull request was last updated.

nrel-bot-2c:
@rraustad it has been 7 days since this pull request was last updated.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1172,7 +1187,7 @@ namespace HVACUnitaryBypassVAV {
             }
 
             if (CBVAV(CBVAVNum).DXCoolCoilType_Num > 0) {
-                ControlNodeNum = GetOnlySingleNode(
+                int ControlNodeNum = GetOnlySingleNode(
```
Myoldmopar: I'm getting that this is now an unused variable, was that your intention?

Myoldmopar:
@rraustad could you take a pass at the build warnings?  I checked this branch out locally with the intention of just trying to clean them up quickly myself.  But one is a confusing unused variable during get-input, that looks to be accidental.  Another is indeed a very simple one.  The third one is that an argument is now not used.  I'd prefer to just remove that argument entirely if it isn't used, but I won't make that decision right now.  After the build warnings are cleaned up, and the CI results are refreshed, I'll take another pass at this.  Thanks!

src/EnergyPlus/HVACUnitaryBypassVAV.cc
rraustad: My intention was to remove top of function declarations but if this now shows an unused variable I'll check it again and delete if needed.

rraustad:
Yes and thanks for the preliminary review.

nrel-bot-2c:
@rraustad it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@rraustad it has been 13 days since this pull request was last updated.

nrel-bot-2b:
@rraustad it has been 8 days since this pull request was last updated.

nrel-bot-2c:
@rraustad it has been 7 days since this pull request was last updated.

nrel-bot-2b:
@rraustad it has been 8 days since this pull request was last updated.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -595,94 +440,152 @@ namespace HVACUnitaryBypassVAV {
...
+            CBVAV(CBVAVNum).SplitterOutletAirNode = NodeInputManager::GetOnlySingleNode(SplitterOutletNodeName,
+                                                                                        ErrorsFound,
+                                                                                        CurrentModuleObject,
+                                                                                        Alphas(1),
+                                                                                        DataLoopNode::NodeType_Air,
+                                                                                        DataLoopNode::NodeConnectionType_Internal,
+                                                                                        1,
+                                                                                        DataLoopNode::ObjectIsParent);
+
+            if (NumAlphas > 19 && !lAlphaBlanks(20)) {
```
rraustad, line 529: New node here to connect to mixer or plenum

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -595,94 +440,152 @@ namespace HVACUnitaryBypassVAV {
...
+                                                                                              CurrentModuleObject,
+                                                                                              Alphas(1) + "_PlenumMixerInlet",
+                                                                                              DataLoopNode::NodeType_Air,
+                                                                                              DataLoopNode::NodeConnectionType_Outlet,
+                                                                                              1,
+                                                                                              DataLoopNode::ObjectIsParent);
+            }
+
+            CBVAV(CBVAVNum).plenumIndex = ZonePlenum::getReturnPlenumIndexFromInletNode(CBVAV(CBVAVNum).PlenumMixerInletAirNode);
+            CBVAV(CBVAVNum).mixerIndex = MixerComponent::getZoneMixerIndexFromInletNode(CBVAV(CBVAVNum).PlenumMixerInletAirNode);
```
rraustad, line 549: Find the mixer or plenum index

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1115,10 +1031,12 @@ namespace HVACUnitaryBypassVAV {
                 CBVAV(CBVAVNum).PriorityControl = HeatingPriority;
             } else if (UtilityRoutines::SameString(Alphas(18), "ZonePriority")) {
                 CBVAV(CBVAVNum).PriorityControl = ZonePriority;
+            } else if (UtilityRoutines::SameString(Alphas(18), "LoadPriority")) {
+                CBVAV(CBVAVNum).PriorityControl = LoadPriority;
```
rraustad, line 1035: New option for load priority control

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1171,9 +1089,8 @@ namespace HVACUnitaryBypassVAV {
                 ShowContinueError("In " + CurrentModuleObject + " \"" + CBVAV(CBVAVNum).Name + "\".");
             }
 
-            if (CBVAV(CBVAVNum).DXCoolCoilType_Num > 0) {
-                ControlNodeNum = GetOnlySingleNode(
-                    Alphas(7), ErrorsFound, CurrentModuleObject, Alphas(1), NodeType_Air, NodeConnectionType_Sensor, 1, ObjectIsParent);
+            if (NumNumbers > 8) {
+                CBVAV(CBVAVNum).minModeChangeTime = Numbers(9);
```
rraustad, line 1093: new input for minimum operating time in cooling or heating. If not entered, min time = -1, if field is present and blank default is read in as 0.25 hr, else read user input.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1430,13 +1356,18 @@ namespace HVACUnitaryBypassVAV {
                                 "System",
                                 "Average",
                                 CBVAV(CBVAVNum).Name);
+            SetupOutputVariable("Unitary System Operating Mode Index",
```
rraustad, line 1359: New report variable so users know operating mode

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1625,139 +1526,146 @@ namespace HVACUnitaryBypassVAV {
...
+            DataLoopNode::Node(InNode).MassFlowRateMaxAvail = max(CBVAV(CBVAVNum).MaxCoolAirMassFlow, CBVAV(CBVAVNum).MaxHeatAirMassFlow);
+            DataLoopNode::Node(InNode).MassFlowRateMin = 0.0;
+            DataLoopNode::Node(InNode).MassFlowRateMinAvail = 0.0;
+            DataLoopNode::Node(OutNode).Temp = DataLoopNode::Node(InNode).Temp;
+            DataLoopNode::Node(OutNode).HumRat = DataLoopNode::Node(InNode).HumRat;
+            DataLoopNode::Node(OutNode).Enthalpy = DataLoopNode::Node(InNode).Enthalpy;
+            DataLoopNode::Node(CBVAV(CBVAVNum).MixerReliefAirNode) = DataLoopNode::Node(MixerOutsideAirNode);
             MyEnvrnFlag(CBVAVNum) = false;
             CBVAV(CBVAVNum).LastMode = HeatingMode;
+            CBVAV(CBVAVNum).changeOverTimer = 0.0;
```
rraustad: initialize changeOverTimer (time when mode changed + new input for min operating time) to 0 on BeginEnvirnFlag

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1788,24 +1697,24 @@ namespace HVACUnitaryBypassVAV {
...
+                CBVAV(CBVAVNum).OpMode = DataHVACGlobals::CycFanCycCoil;
             } else {
-                CBVAV(CBVAVNum).OpMode = ContFanCycCoil;
+                CBVAV(CBVAVNum).OpMode = DataHVACGlobals::ContFanCycCoil;
             }
         }
 
         // Returns load only for zones requesting cooling (heating). If in deadband, Qzoneload = 0.
-        GetZoneLoads(CBVAVNum, QZnReq);
+        GetZoneLoads(CBVAVNum);
```
rraustad, line 1708: No longer need (never really needed) QZnReq. This is a temperature based control model so only need to know operating mode, not the actual load.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -2244,84 +2126,58 @@ namespace HVACUnitaryBypassVAV {
...
+            DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode) = DataLoopNode::Node(InletNode);
+        } else {
+            DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).Temp =
+                (1.0 - BypassDuctFlowFraction) * DataLoopNode::Node(InletNode).Temp + BypassDuctFlowFraction * DataLoopNode::Node(OutletNode).Temp;
+            DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).HumRat = (1.0 - BypassDuctFlowFraction) * DataLoopNode::Node(InletNode).HumRat +
+                                                                           BypassDuctFlowFraction * DataLoopNode::Node(OutletNode).HumRat;
+            DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).Enthalpy = Psychrometrics::PsyHFnTdbW(
+                DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).Temp, DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).HumRat);
+        }
+        MixedAir::SimOAMixer(CBVAV(CBVAVNum).OAMixName, FirstHVACIteration, CBVAV(CBVAVNum).OAMixIndex);
```
rraustad, line 2176: RE: don't need to call mixer when internal OA flow = 0 (i.e., when user sets internal OA flow = 0 or uses AirloopHVAC:Mixer or ReturnPlenum and sets internal OA flow = 0). I decided to leave this alone. I could have done this but thought it was overkill (but might be a time saver overall):

```
if (OAFlow == 0.0) {
  Node(MixedAirNode) = Node(MixerInletNode);
  Node(MixerExhNode) = Node(MixerInletNode);
  Node(MixerExhNode).MassFlowRate = 0.0;
} else{
  MixedAir::SimOAMixer();
}
```

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3422,100 +3328,86 @@ namespace HVACUnitaryBypassVAV {
...
-        CBVAV(CBVAVNum).BypassMassFlowRate = BypassDuctFlowFraction * Node(CBVAV(CBVAVNum).MixerInletAirNode).MassFlowRate;
+        int splitterOutNode = CBVAV(CBVAVNum).SplitterOutletAirNode;
+        DataLoopNode::Node(splitterOutNode).MassFlowRateSetPoint = DataLoopNode::Node(OutletNode).MassFlowRateSetPoint;
+        DataLoopNode::Node(OutletNode) = DataLoopNode::Node(splitterOutNode);
+        DataLoopNode::Node(OutletNode).MassFlowRate =
+            (1.0 - BypassDuctFlowFraction) * DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).MassFlowRate;
+        // report variable
+        CBVAV(CBVAVNum).BypassMassFlowRate = BypassDuctFlowFraction * DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).MassFlowRate;
+        // initialize bypass duct connected to mixer or plenum with flow rate and conditions
+        if (CBVAV(CBVAVNum).plenumIndex > 0 || CBVAV(CBVAVNum).mixerIndex > 0) {
```
rraustad, line 3359: When connected to a mixer or return plenum, place the mass flow rate on the mixer/plenum node. This means the toal return air to the air loop will be the full flow, not the flow to just the zone terminal units. Before, the bypass was handled internally and mixed after the OA mixer. This was the crux of the problem Carrier had with this model. Now bypass air is mixed as the AirloopHVAC:OutsideAirSystem and air conditions entering the air loop are more representative of a real system.

src/EnergyPlus/HVACManager.cc
```diff
@@ -2011,7 +2012,9 @@ namespace HVACManager {
                     // node mass flow max avail to what air loop can supply
                     SupplyNode = AirToZoneNodeInfo(AirLoopIndex).AirLoopSupplyNodeNum(SupplyIndex);
                     if (Node(SupplyNode).MassFlowRate > 0.0) {
-                        if ((Node(SupplyNode).MassFlowRateSetPoint - Node(SupplyNode).MassFlowRate) > HVACFlowRateToler * 0.01) {
+                        // must include bypass flow for ChangeoverBypass system so that terminal units are not restricted (e.g., MaxAvail is lowered)
+                        if ((Node(SupplyNode).MassFlowRateSetPoint - Node(SupplyNode).MassFlowRate - AirLoopFlow(AirLoopIndex).BypassMassFlow) >
+                            HVACFlowRateToler * 0.01) {
```
rraustad, line 2017: Subtracting off bypass mass flow here to allow convergence. I will look at this more to see if I can avoid this.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3542,61 +3434,68 @@ namespace HVACUnitaryBypassVAV {
...
-                        QZoneReq = QZoneReqHeat;
                         CBVAV(CBVAVNum).HeatCoolMode = HeatingMode;
                     } else if (std::abs(QZoneReqCool) == std::abs(QZoneReqHeat) && QZoneReqCool != 0.0) {
-                        QZoneReq = QZoneReqCool;
                         CBVAV(CBVAVNum).HeatCoolMode = CoolingMode;
-                    } else {
-                        QZoneReq = 0.0;
+                    }
+                }
+            } else if (SELECT_CASE_var == LoadPriority) {
```
rraustad, line 3475: Added logic for new control method.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3542,61 +3434,68 @@ namespace HVACUnitaryBypassVAV {
...
+                    } else if (QZoneReqHeat > 0.0) {
+                        CBVAV(CBVAVNum).HeatCoolMode = HeatingMode;
                     }
                 }
             }
         }
+
+        if (CBVAV(CBVAVNum).LastMode != CBVAV(CBVAVNum).HeatCoolMode) {
+            CBVAV(CBVAVNum).changeOverTimer = thisTime + CBVAV(CBVAVNum).minModeChangeTime;
+        }
```
rraustad, line 3505: Timer is updated to be the time of this mode change plus the minimum time until change over is allowed.

src/EnergyPlus/MixerComponent.cc
```diff
@@ -683,6 +683,34 @@ namespace MixerComponent {
         }
     }
 
+    int getZoneMixerIndexFromInletNode(int const &InNodeNum)
```
rraustad, line 686: New functions to get mixer or return plenum index.

rraustad:
Messy changes here so follow my comments on model changes. The rest is eliminating Using statements and top-of-function declarations. Had to get this done since no one normally works on this model.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
rraustad: I guess this could have been just one variable, mixerOrPlenumIndex, or even a bool since it's one or the other or 0. I didn't think of that when I started this ... hindsight. I only use this to trigger sending air to the bypass node, which doesn't care what it's connected to. It's getting late in the game so changing this might slow things down. A follow up would be better and could still go in after the IDD freeze.

rraustad:
I sent a release package to Tiejun Wu at Carrier today. He will exercise the new model and look for issues.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -2244,84 +2126,58 @@ namespace HVACUnitaryBypassVAV {
...
-            (1.0 - BypassDuctFlowFraction) * Node(InletNode).Temp + BypassDuctFlowFraction * Node(OutletNode).Temp;
-        Node(CBVAV(CBVAVNum).MixerInletAirNode).HumRat =
-            (1.0 - BypassDuctFlowFraction) * Node(InletNode).HumRat + BypassDuctFlowFraction * Node(OutletNode).HumRat;
-        Node(CBVAV(CBVAVNum).MixerInletAirNode).Enthalpy =
-            PsyHFnTdbW(Node(CBVAV(CBVAVNum).MixerInletAirNode).Temp, Node(CBVAV(CBVAVNum).MixerInletAirNode).HumRat);
-        SimOAMixer(CBVAV(CBVAVNum).OAMixName, FirstHVACIteration, CBVAV(CBVAVNum).OAMixIndex);
-
-        if (CBVAV(CBVAVNum).FanPlace == BlowThru) {
+        if (CBVAV(CBVAVNum).plenumIndex > 0 || CBVAV(CBVAVNum).mixerIndex > 0) {
+            Real64 saveMixerInletAirNodeFlow = DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).MassFlowRate;
```
rraustad, line 2165: Must save operating flow rate at mixer inlet node since the system inlet node will sum to the sum of the TU's plus the bypass. The first pass at FirstHVACIteration = true sets full flow and bypass = 0 so the sum of the TU's will be less than full flow (i.e., they won't be at full flow once they simulate). If lines 2167 and 2169 are removed, there are warnings of low flow per capacity warnings for DX coil, indicating that the flow through the coil is lower than expected, hence operating flow needs to be saved. Lines 2168 and 2169, pass the node conditions from inlet to the mixer inlet and reset mass flow.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3422,100 +3328,86 @@ namespace HVACUnitaryBypassVAV {
...
-
-        // SUBROUTINE PARAMETER DEFINITIONS:
-        // na
+        // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
+        Real64 ZoneLoad = 0.0; // Total load in controlled zone [W]
 
-        // INTERFACE BLOCK SPECIFICATIONS
-        // na
+        Real64 thisTime = (DataGlobals::DayOfSim - 1) * 24 + DataGlobals::HourOfDay - 1 + (DataGlobals::TimeStep - 1) * DataGlobals::TimeStepZone +
+                          DataHVACGlobals::SysTimeElapsed;
```
rraustad, line 3398: Time of this time step

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3422,100 +3328,86 @@ namespace HVACUnitaryBypassVAV {
...
 
-        // DERIVED TYPE DEFINITIONS
-        // na
+        if (thisTime <= CBVAV(CBVAVNum).changeOverTimer) return;
```
rraustad, line 3400: Compare time of this time step to time where change over can be allowed, if no change over is allowed return.

src/EnergyPlus/HVACUnitaryBypassVAV.hh
```diff
@@ -257,33 +259,38 @@ namespace HVACUnitaryBypassVAV {
         int CRDXIterationFailed;                 // Counter for cool reheat multimode DX coil messages
         int CRDXIterationFailedIndex;            // Counter for cool reheat multimode DX coil messages
         bool FirstPass;                          // used to determine when first call is made
+        int plenumIndex;                         // index to AirloopHVAC:ReturnPlenum
+        int mixerIndex;                          // index to AirloopHVAC:ZoneMixer
```
rraustad, line 263: these are the 2 ints that can be converted to a single bool

src/EnergyPlus/HVACUnitaryBypassVAV.hh
```diff
@@ -257,33 +259,38 @@ namespace HVACUnitaryBypassVAV {
...
+        Real64 changeOverTimer;                  // timer to avoid rapid change of operating modes (e.g., cooling to heating)
+        Real64 minModeChangeTime;                   // time limit before mode change is allowed (hr)
```
rraustad, line 265: New variables to avoid short cycling.

src/EnergyPlus/MixerComponent.cc
rraustad: These can be converted to bool functions but are still needed to verify connection is correct.

src/EnergyPlus/ZonePlenum.cc
```diff
@@ -1535,6 +1535,35 @@ namespace ZonePlenum {
         }
     }
 
+    int getReturnPlenumIndexFromInletNode(int const &InNodeNum)
```
rraustad, line 1538: Same int function as in MixerComponent.cc. Can be converted to bool.

rraustad:
OK, this should leave only non-ChangeoverBypass example files with diff's since this branch is behind develop.

rraustad:
**Comment from Carrier on 8/9/2019.**

Note the discussion of AirloopHVAC:OutdoorAirSystem controls with respect to AirloopHVAC: system operation. Seems like a similar problem to #6727 where OA system operation needs tighter linkage with air loop system operation.

```
Rich,
Thanks for fixing the problem so quick. I tested the new build and looks like it works great.
I tweaked the North zone heat source to 7.5 kW so that the unitary system would be able to
switch between cooling and heating mode throughout the winter designday. Looks like
both the Load Priority Control and the Minimum Run Time Control worked beautifully.

One thing I noticed is that every time it changes to cooling mode, for the first time step, the 
DX cooling coil would operate because the OA mixer outlet node setpoint was calculated 
based on the previous time step air loop outlet node temperature setpoint (which was in 
heating mode). So that the DX coil would need to turn on. Then at the next time step, the 
mixed air node stepoint follows the new cooling setpoint and economizer would bring 
enough cool outdoor air to meet the 10 C setpoint at air loop outlet node. So that the 
DX cooling coil would be off. I think this should be fine.

Next I will test a plenum return configuration and report the results.

Best,
Tiejun
```

rraustad:
@Myoldmopar @mjwitte I believe this latest version is the final unless Carrier or the review process shows something that needs to be addressed.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3714,40 +3588,38 @@ namespace HVACUnitaryBypassVAV {
...
 
         //   Account for floating condition where cooling/heating is required to avoid overshooting setpoint
-        if (CBVAV(CBVAVNumber).HeatCoolMode == 0 && CBVAV(CBVAVNumber).OpMode == ContFanCycCoil) {
+        if (CBVAV(CBVAVNumber).HeatCoolMode == 0 && CBVAV(CBVAVNumber).OpMode == DataHVACGlobals::ContFanCycCoil) {
             if (OutAirTemp > TSupplyToCoolSetPtMin) {
                 CalcSetPointTempTarget = TSupplyToCoolSetPtMin;
-                CBVAV(CBVAVNumber).HeatCoolMode = CoolingMode;
             } else if (OutAirTemp < TSupplyToHeatSetPtMax) {
                 CalcSetPointTempTarget = TSupplyToHeatSetPtMax;
-                CBVAV(CBVAVNumber).HeatCoolMode = HeatingMode;
```
rraustad, line 3736: Lines 3733, 3736, 3747 and 3750 here are responsible for diff's. This change was a last minute correction based on feedback from Carrier [7e90a6c](https://github.com/NREL/EnergyPlus/pull/7249/commits/7e90a6ce08066eefc589184374165605580d09c3). These lines would change the operating mode based on outdoor temperature and disregard the user specified Priority Control Mode (Cooling Priority, Heating Priority, ZonePriority, and now LoadPriority) and also the new user controlled minimum runtime. Carrier noticed this during the review process for this new feature (See discussion in Conversation tab of 8/9/2019).

rraustad:
@Myoldmopar regarding your initial review comment on Apr 1, 2019 on increased iterations. The number of iterations did increase. I would certainly take suggestions on how to minimize the extra iterations.

This figure compares the 2 design day example file ChangeoverBypassVAV.idf in develop (annual run = 6.6 sec) with the same file when connected to the mixer in this branch (annual run = 11.1 sec).

image

rraustad:
**Comment from Carrier on Plenum connected CBVAV on 8/9/2019**

```
Rich,

I tested plenum return configuration and looks like it shows same system behavior as the 
ducted return. From the results looks like the controls works fine.

When I looked at the results for both ducted return and plenum return. I see some behavior that 
I am not quite sure I understand clearly. Would like to get your comments. I am using the ducted
return configuration as an example:

1.       In the following plot for the system operating mode, the last time step running in heating
mode is 12:10, at 12:11:40 the system switched to heating mode. But it did not switch back to
heating mode until 12:55. The last timestep at cooling mode is 12:50. So it looks like instead of
the minimum run time of 30 minutes, the system stayed in cooling mode for 40 minutes (or 45
minutes). Though the combined heating load from West and East zone has exceeded the cooling
load of North zone after just a few time steps into cooling mode.

2.       When the system first switched into cooling mode at 12:11:40, it determined that the 
Unitary system outlet node setpoint should be 10.10 C to cool down the North zone. At this 
time step the North zone damper is fully open and it cooled down to 22.79 C. In the next 
time step, it maintained this setpoint of 10.10 C and the North zone is cooled to below its 
cooling setpoint 21.70 C  (setpoint is 22.30). And the damper closed to minimum opening 
of 0.25. This low supply air temperature triggered the west and East zone reheat coil to 
operate to maintain zone temperature. The unitary outlet setpoint of 10.10 C was maintained
throughout the cooling operation period.  Which caused the West and East zone reheat coil 
to operate for the period.  For this I thought that the Unitary system should have adjusted its 
outlet setpoint to a higher temperature to keep North zone damper fully open. (for example 
the unitary outlet setpoint should be reset to somewhere between 17-19 C, however if the 
unitary system is not able to meet this setpoint without operating the heating coil, it should 
just keep the heating coil off since the system is operating in cooling mode, this way the
economizer outlet node setpoint would be higher and it would not bring in so much cold 
outdoor air,  then if the supply  temperature is still  cooler than the North zone needed, then 
reduce the North zone damper opening to maintain zone temperature at or above heating 
setpoint of 20C). 

Thanks,
Tiejun
```

doc/input-output-reference/src/overview/group-unitary-equipment.tex
```diff
@@ -2843,6 +2843,12 @@ \subsection{AirLoopHVAC:UnitaryHeatCool:VAVChangeoverBypass}\label{airloophvacun
 \caption{Schematic of a CBVAV unitary system (blow through fan placement) \protect \label{fig:schematic-of-a-cbvav-unitary-system-blow}}
 \end{figure}
 
+\begin{figure}[hbtp]
+\centering
+\includegraphics[width=0.9\textwidth, height=0.9\textheight, keepaspectratio=true]{media/ChangeoverBypassVAV-ReturnPlenumOrMixer.png}
```
Myoldmopar, line 2848: I think this is causing the "build failure" on Linux.  The filename is subtly different:

```
media/ChangeoverBypassVAV-ReturnPlenumOrMixer.png    -- in this spot
media/ChangeoverBypassVAV-ReturnPlenumOrMixer.PNG   -- the actual file name
```

Keeping in mind that Linux is case sensitive.  Also, make sure this new image is added to the CMake rules [here](https://github.com/NREL/EnergyPlus/blob/5e66d36f7c13d2f0f794370eabd85f0dd90db55e/doc/input-output-reference/CMakeLists.txt#L117) so that if the image is changed, CMake picks up that change and will rebuild the doc, otherwise it doesn't know that the image is a dependency of this build.

Myoldmopar:
This looks to be all in good shape.  Thanks for the effort here @rraustad and those others who provided testing and validation along the way.  Merging.

rraustad:
Thanks. So after the merge, what do these new CI failures mean?

Myoldmopar:
CI appears to have gotten confused and thought those results still needed to be generated, so it tried to checkout the branch and failed because the branch was deleted.  Please ignore...

PR #7267, opened 2019-04-11
Streamline ManageSurfaceHeatBalance - Part 1

mjwitte:
Pull request overview
---------------------
Initial work on cleaning up `CalcHeatBalanceInsideSurf`.

- [x] Replace `any_eq` with simple bools for which algorithms are in use
- [x] Move out things that should only happen once, like building the heat transfer surface list, etc.
- [x] Move out things that should only happen at the end of the timestep (not repeatedly during iterations), like computing zone window heat gains, etc.
- [x] Build vectors with surface lists to drive loops in CalcHeatBalanceInsideSurf to reduce repeated checking if a surface is a heat transfer surface, or a window, etc.
- [x] Add unit tests (except none yet for CalcHeatBalanceInsideSurf, because more work is planned there)

*Stopping here for now. Will take up the remaining items and more in a new PR building on this one.*
- [ ] Move Kiva surface temp set and reset out of iteration loop - tried but got diffs - likely due to the every-30-iteration call to `InitInteriorConvectionCoeffs`
- [ ] Streamline calcs
- [ ] ??

### Diff Summary
WCE* small eso diffs in ROOM 102 WINDOW:Surface Inside Face Convection Heat Gain Rate. The new values appear to be better than before - they match h*DeltaT.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2:
@mjwitte it has been 22 days since this pull request was last updated.

nrel-bot-2c:

nrel-bot-2c:
@mjwitte it has been 8 days since this pull request was last updated.

mjwitte:
@Myoldmopar Code is ready for initial review, but still need to add unit tests.

src/EnergyPlus/DataSurfaces.cc
```diff
@@ -498,6 +498,12 @@ namespace DataSurfaces {
     Array1D<Real64> WinGapConvHtFlowRepEnergy;     // Energy of WinGapConvHtFlowRep [J]
     Array1D<Real64> WinHeatTransferRepEnergy;      // Energy of WinHeatTransfer [J]
 
+    std::vector<int> AllHTSurfaceList;  // List of all heat transfer surfaces
+    std::vector<int> AllIZSurfaceList;  // List of all interzone heat transfer surfaces
+
+    bool AnyHeatBalanceInsideSourceTerm(false);  // True if any SurfaceProperty:HeatBalanceSourceTerm inside face used
+    bool AnyHeatBalanceOutsideSourceTerm(false); // True if any SurfaceProperty:HeatBalanceSourceTerm outside face used
```
Myoldmopar, line 507: These flags and specialized vectors will surely be helpful.

src/EnergyPlus/DataSurfaces.hh
```diff
@@ -1093,6 +1099,8 @@ namespace DataSurfaces {
         int WindowModelType;           // if set to WindowBSDFModel, then uses BSDF methods
         BSDFWindowDescript ComplexFen; // Data for complex fenestration, see DataBSDFWindow.cc for declaration
 
+        int TDDPipeNum; // Tubular daylighting device pipe number for TDD domes and diffusers
```
Myoldmopar, line 1104: Yes, skipping looking this up each time will be an easy win for sure.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -184,40 +184,20 @@ namespace HeatBalanceSurfaceManager {
...
     // PUBLIC  CalcHeatBalanceInsideSurf   ! radiant systems need access to them in order to simulate
 
-    // Record Keeping/Utility Routines for Module
-
-    // Reporting routines for module
-
-    // MODULE SUBROUTINES:
-    //*************************************************************************
-
-    // Functions
```
Myoldmopar, line 220: I like all the cleanup.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -5826,8 +5781,55 @@ namespace HeatBalanceSurfaceManager {
 
     void CalcHeatBalanceInsideSurf(Optional_int_const ZoneToResimulate) // if passed in, then only calculate surfaces that have this zone
     {
+        // Pass correct list of surfaces to CalcHeatBalanceInsideSurf2
+        bool const PartialResimulate(present(ZoneToResimulate));
 
-        // SUBROUTINE INFORMATION:
+        if (!PartialResimulate) {
+            for (int zoneNum = 1; zoneNum <= NumOfZones; ++zoneNum) {
+                ZoneWinHeatGain = 0.0;
```
Myoldmopar: Is this assigning zero to each vector item each time this loop occurs?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -5856,59 +5858,17 @@ namespace HeatBalanceSurfaceManager {
...
-        using DataMoistureBalance::HMassConvInFD;
-        using DataMoistureBalance::HSkyFD;
-        using DataMoistureBalance::RhoVaporAirIn;
-        using DataMoistureBalance::RhoVaporAirOut;
-        using DataMoistureBalance::RhoVaporSurfIn;
-        using DataMoistureBalance::TempOutsideAirFD;
-        using DataMoistureBalanceEMPD::HeatFluxLatent;
-        using DataMoistureBalanceEMPD::RVSurface;
-        using DataMoistureBalanceEMPD::RVSurfaceOld;
-        using DataMoistureBalanceEMPD::RVSurfLayer;
```
Myoldmopar, line 5887: I'm liking all these usings going away :+1:

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -6099,81 +6013,66 @@ namespace HeatBalanceSurfaceManager {
...
         }
 
-        bool const useCondFDHTalg(any_eq(HeatTransferAlgosUsed, UseCondFD));
         Converged = false;
         while (!Converged) { // Start of main inside heat balance DO loop...
 
             TempInsOld = TempSurfIn; // Keep track of last iteration's temperature values
 
-            if (any_eq(HeatTransferAlgosUsed, HeatTransferModel_Kiva)) {
+            if (DataHeatBalance::AnyKiva) {
```
Myoldmopar, line 6069: This should be a really good gain.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -6495,30 +6383,27 @@ namespace HeatBalanceSurfaceManager {
...
-                        if (construct.SourceSinkPresent && calcHeatBalanceInsideSurfFirstTime)
-                            ShowSevereError("Windows are not allowed to have embedded sources/sinks");
-
                         if (SurfaceWindow(SurfNum).OriginalClass == SurfaceClass_TDD_Diffuser) { // Tubular daylighting device
                             // Lookup up the TDD:DOME object
-                            PipeNum = FindTDDPipe(SurfNum);
-                            SurfNum2 = TDDPipe(PipeNum).Dome;
-                            Ueff = 1.0 / TDDPipe(PipeNum).Reff;
+                            int pipeNum = SurfaceWindow(SurfNum).TDDPipeNum;
+                            int domeNum = DataDaylightingDevices::TDDPipe(pipeNum).Dome;
```
Myoldmopar: The compiler is probably already doing it, but I'd assign `const` where possible to be clear this is just assigned.

Myoldmopar (commented):
This is going to be really great.  I made some comments and had one question, but overall this is in good shape.

Myoldmopar:
I see there's a few percent gain on the current test suite, but do you have annual results?  I'd expect them to be even higher.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
mjwitte: Yes. Suggestions?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
mjwitte: OK - will do.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
Myoldmopar: Well, I mean it looks like ZoneWinHeatGain is an array allocated to the number of zones.  Just like the other arrays here.  Assigning a zero value to the ZoneWinHeatGain array here is essentially doing an internal loop over all the zones and assigning every ZoneWinHeatGain index equal to zero.  

I would either change this line to `ZoneWinHeatGain(zoneNum) = 0.0;`, like this:

```
for (int zoneNum = 1; zoneNum <= NumOfZones; ++zoneNum) {
  ZoneWinHeatGain(zoneNum) = 0.0;   // THIS IS NOW INDEXED
  ZoneWinHeatGainRep(zoneNum) = 0.0;
  ZoneWinHeatGainRepEnergy(zoneNum) = 0.0;
  ZoneWinHeatLossRep(zoneNum) = 0.0;
  ZoneWinHeatLossRepEnergy(zoneNum) = 0.0;
}
```

... or just take all these arrays out of the loop here and assign them all to zero at once:

```
ZoneWinHeatGain = 0.0;
ZoneWinHeatGainRep = 0.0;
ZoneWinHeatGainRepEnergy = 0.0;
ZoneWinHeatLossRep = 0.0;
ZoneWinHeatLossRepEnergy = 0.0;
```

In that case, the loop would occur for each, but you would rely on some possibly optimized operations inside of each array and avoid the array lookup operations.  It would be interesting to do a test to see which is faster.

OR

You tell me why I missed the point of this operation and it's actually correct as-is. 😄

mjwitte: Got it. These can all be simply `ZoneWinHeatGain* = 0.0;` and drop the for loop.  And this screams for a window surface lists. Adding that as well.

mjwitte:
@Myoldmopar I've addressed your comments and went a step further with separate surface lists for windows and non-windows.

Annual run comparison of 10x_incr_100zones100surfaces.idf (non-scientific, single-run stats on laptop, and yes, this file has 10,200 surfaces, but no windows):
Develop 9.2.0-26151c5477 24 min 54 sec
This branch 9.2.0-39ff270514 22 min 50 sec

So, would you like me to stop on this branch and start a new one for further work (just to keep things in reviewable chunks), or keep going here?

Myoldmopar:
I'd say leave it as-is for this branch.  This is a really nice cleanup.  I see the WCE diffs have been present for quite a while, have you already investigated those or should I dive in?

mjwitte:
Comment lurking at the top

> Diff Summary
WCE* small eso diffs in ROOM 102 WINDOW:Surface Inside Face Convection Heat Gain Rate. The new values appear to be better than before - they match h*DeltaT.

Would be prudent for you to look at one of these to see if you agree.

Myoldmopar:
OK, built this locally, pulled in latest develop and compared to develop. I reproduced the small diffs in the WCE files.  I calculated the convection rate manually in both files and found that while it may be closer, it's deep in the decimal points and insignificant.  This is good to go in.  I'm going to merge.

Issue #7233, opened 2019-03-19
AirflowNetwork floating point underflow


mjwitte:
Issue overview
--------------
Running AirflowNetwork_MultiAirLoops with Windows debug build fails with a floating point underflow [here](https://github.com/NREL/EnergyPlus/blob/a837bb7430496285acd0d88e081fb7bc3a0fa3f4/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L7001). Suspect there may be other places that need protection against underflow.

### Details
Some additional details for this issue (if relevant):
 - Platform Win64, debug
 - Version of EnergyPlus 9.1.0-https://github.com/NREL/EnergyPlus/commit/a837bb7430496285acd0d88e081fb7bc3a0fa3f4
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
@jasondegraw @lgu1234

lgu1234:
@mjwitte I tested the program and found underflow occurs at line 7001 in AirflowNetworkBalanceManager. 

                        Ei = std::exp(-UThermal * DuctSurfArea / (AirflowNetworkLinkSimu(i).FLOW2 * CpAir));
Then I found the value AirflowNetworkLinkSimu(i).FLOW2 is close 1.0E-11.

I looked at the website at http://www.enseignement.polytechnique.fr/informatique/INF478/docs/Cpp/en/cpp/numeric/math/exp.html and found the note below:

For IEEE-compatible type double, overflow is guaranteed if 709.8 < arg, and underflow is guaranteed if arg < -708.4

In order to prevent underflow for std:exp, we have to limit the value of argument < -708.4.

In addition, the same website also show return value:

If a range error occurs due to underflow, the correct result (after rounding) is returned.

Any comments?

lgu1234:
@mjwitte @jasondegraw A possible change:

Ei = std::exp(min(-UThermal * DuctSurfArea / (AirflowNetworkLinkSimu(i).FLOW2 * CpAir), -708.4));

jasondegraw:
[...]

PR #7274, opened 2019-04-18
Fix daylighting for BSDF complex fenestration with shading layer

### LINKED ISSUE #7128: Daylighting is broken for complex fenestration with a shading layer
Issue overview
--------------
1. Daylighting illuminance values are all zero for complex fenestration with a WindowMaterial:ComplexShade and SplitFlux method. Problem was introduced in v9.0, most likely in #6855 but that has not been confirmed.

2. Example files CmplxGlz_Daylighting_SouthVB45deg.idf and CmplxGlz_Daylighting_SouthVerticalVB45deg.idf both have `Daylighting:Controls` "Fraction of Zone Controlled by Reference Point 1" set to 0.0, so that daylighting controls are inactive. This should be changed to 1.0.

*Initial user post:*

> I am modelling a zone illuminated by a CFS window with the SplitFlux method. While the window transmittance and the solar gains seem to be calculated correctly (changing accordingly to the type of shading layer included), the illuminance over every Reference Point is zero anytime a shading layer is included. On the other hand, if I simulate the same window without shading layer, the illuminance is calculated correctly. I tried to model different shading layers, but the result is always the same. as a side note, I would like to mention that also in the example CmplxGlz_Daylighting_SouthVerticalVB45deg the same problem occurs.

*More from user*
>two critical points regarding the modelling of dynamic complex fenestration systems:

>1)DElight output (in all the versions I tested 8.5 8.9 &9.0) of a CFS including a complex shade and a double glazing (copied directly in the building idf) is 0 lux on any reference point. Note that if using DElight in the example EMSThermochromicWindow the output os produced: the problem apparently is the shading layer.

>2)Whenever actuating a CFS with through the EMS Contruction State actuator, the thermal balance and the solar gains are calculated correctly, while daylighting is not affected by the actuator (I read of similar problems in https://github.com/NREL/EnergyPlus/pull/6801). This occurs in any version (8.5, 8.9 9.0). You can notice it also in the example EMSThermochromicWindow v 8.9 if inserting a SplitFlux or DElight control in ZN1 (solar transmitted radiation affected by the actuator, but Illuminance not affected even if differently calculated by the two daylighting methods).

>Apparently those two bugs make impossible to properly simulate CFS (with more advanced daylighting calculation method) and to actuate any CFS dynamically (lights internal loads & daylighting are wrong).

### Details
 - Version of EnergyPlus 9.0.1
 - Helpdesk ticket number 13146

### Checklist
 - [x] Defect files: Example files CmplxGlz_Daylighting_SouthVB45deg.idf and CmplxGlz_Daylighting_SouthVerticalVB45deg.idf

mjwitte:
Pull request overview
---------------------
Fixes #7128. Diffs are expected in the CmplxGlz_Daylighting_* test files. The "Daylighting Reference Point 1 Illuminance" was all zeroes and now it has values which match the output from v8.9 and earlier.

ToDo
--------------

- [x] Added temporary underflow protection in [TarcogShading](https://github.com/NREL/EnergyPlus/pull/7274/files#diff-9fcefa849b6a7cedfc32d8a8a8b8e5e7R1039). Need to review.

- [x] Unit test `EnergyPlusFixture.DaylightingManager_GetDaylParamInGeoTrans_Test` fails locally on Win64 in VS2017, because the [`ArrayND` declarations like these in `DaylightingManager.cc`](https://github.com/NREL/EnergyPlus/blob/ec20f0b9713830463fb1641ccf052ae21bb00bdf/src/EnergyPlus/DaylightingManager.cc#L207-L223) with hard sizes do not size the arrays when running a unit test. The debugger shows these all sized at 1:0, 1:0 which causes this unit test to fail. For the unit test added in this PR, I added allocates as needed, but seems that shouldn't be necessary. Can't figure out what's missing for the unit test.
*`ArrayND` arrays which are sized in the declaration have no explicit allocate or dimension elsewhere in the code. These arrays should not be deallocated in `clear_state` functions because `EnergyPlusFixture` calls `clear_all_states` during setup which obliterates the arrays that were sized upon loading EnergyPlus.*

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/TarcogShading.cc
```diff
@@ -1036,7 +1036,12 @@ namespace TarcogShading {
                 H0 = (dens2 * cp2 * s * speed) / (4.0 * hc + 8.0 * speed);
 
                 P = -H / H0;
-                beta = std::pow(e, P);
+                if (P < -700.0) {
+                    beta = 0.0;
+                }
+                else {
+                    beta = std::pow(e, P);
```
mjwitte, line 1043: @lgu1234 @jasondegraw This bug fix work stumbled over an underflow in TarcogShading. Did you agree on a plan for exp underflow protection? I'll change this to match.

Myoldmopar: I don't recall if we ever actually agreed on what to do over on the AFN side.  @jasondegraw's [comment](https://github.com/NREL/EnergyPlus/issues/7233#issuecomment-474853258) indicates he has made some progress but we still need resolution.

vidanovic (approved):

src/EnergyPlus/TarcogShading.cc
jasondegraw: @Myoldmopar @mjwitte Oops, missed this. I did make progress, @lgu1234 believe we have a solution, but the problem is really testing. Let's talk this week.

nrel-bot-2b:

mjwitte:
@Myoldmopar CI looks as expected, diffs in CmplxGlz_Daylighting_SouthVB45deg and CmplxGlz_Daylighting_SouthVerticalVB45deg. Ready for your final review and merge.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -5848,7 +5864,9 @@ namespace DaylightingManager {
         for (loop = 1; loop <= ZoneDaylight(ZoneNum).NumOfDayltgExtWins; ++loop) {
             IWin = ZoneDaylight(ZoneNum).DayltgExtWinSurfNums(loop);
             IS = 1;
-            if ((SurfaceWindow(IWin).ShadingFlag >= 1 && SurfaceWindow(IWin).ShadingFlag <= 9) || SurfaceWindow(IWin).SolarDiffusing) IS = 2;
+            if ((SurfaceWindow(IWin).WindowModelType != WindowBSDFModel) &&
+                ((SurfaceWindow(IWin).ShadingFlag >= 1 && SurfaceWindow(IWin).ShadingFlag <= 9) || SurfaceWindow(IWin).SolarDiffusing))
+                IS = 2;
```
Myoldmopar, line 5869: If there are any last minute changes to make, it would be nice to clean this IF block to use braces.  The indentation of the conditions as well as the statement make it a bit funny to read and prone to confusion.  If there's no other changes to make then it's fine.

testfiles/CmplxGlz_Daylighting_SouthVB45deg.idf
```diff
@@ -86440,7 +86440,7 @@
 
   Output:Variable,EXTWALL,Surface Outside Face Incident Sky Diffuse Solar Radiation Rate per Area,Hourly;
 
-  Output:Variable,EXTWALL,Surface Inside Temperature,Hourly;
+  Output:Variable,EXTWALL,Surface Inside Face Temperature,Hourly;
```
Myoldmopar, line 86443: OK, this is just something that got missed when the variable changed at some point.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -9955,7 +9984,9 @@ namespace DaylightingManager {
                 for (loop = 1; loop <= ZoneDaylight(ZoneNum).NumOfDayltgExtWins; ++loop) {
                     IWin = ZoneDaylight(ZoneNum).DayltgExtWinSurfNums(loop);
                     IS = 1;
-                    if ((SurfaceWindow(IWin).ShadingFlag >= 1 && SurfaceWindow(IWin).ShadingFlag <= 9) || SurfaceWindow(IWin).SolarDiffusing) IS = 2;
+                    if ((SurfaceWindow(IWin).WindowModelType != WindowBSDFModel) &&
+                        ((SurfaceWindow(IWin).ShadingFlag >= 1 && SurfaceWindow(IWin).ShadingFlag <= 9) || SurfaceWindow(IWin).SolarDiffusing))
+                        IS = 2;
```
Myoldmopar, line 9989: All these changes seem fine, just more carefully avoiding BSDF windows when doing calculations if I'm reading it right.

tst/EnergyPlus/unit/DaylightingManager.unit.cc
```diff
@@ -1081,3 +1081,311 @@ TEST_F(EnergyPlusFixture, MapShadeDeploymentOrderToLoopNumber_Test)
...
+    SurfaceWindow(IWin).WindowModelType = WindowBSDFModel;
+    SurfaceWindow(IWin).ShadingFlag = DataSurfaces::NoShade;
+    DaylightingManager::DayltgInteriorIllum(ZoneNum);
+    EXPECT_NEAR(DaylightingManager::DaylIllum(1), 100.0, 0.001);
+    EXPECT_NEAR(DaylightingManager::DaylIllum(2), 10.0, 0.001);
+
+    SurfaceWindow(IWin).ShadingFlag = DataSurfaces::ExtBlindOn;
+    DaylightingManager::DayltgInteriorIllum(ZoneNum);
+    EXPECT_NEAR(DaylightingManager::DaylIllum(1), 100.0, 0.001);
+    EXPECT_NEAR(DaylightingManager::DaylIllum(2), 10.0, 0.001);
```
Myoldmopar, line 1390: :+1:

Myoldmopar (approved):

PR #7278, opened 2019-04-19
Use source mass flow rate in unmet energy calculation on stratified tank

### LINKED ISSUE #7275: Stratified tank source heat transfer using use flow rate instead of source
Issue overview
--------------
A very astute observer of the E+ source code noticed that we are [using the use flow rate in the source side energy calculation](https://github.com/NREL/EnergyPlus/blob/1afc16dac65705550eb9e08c5398d29f434f41ee/src/EnergyPlus/WaterThermalTanks.cc#L8294). Should be easy enough to fix. 

### Details
 - Version of EnergyPlus : 9.1

### Checklist

nmerket:
fixes #7275

Pull request overview
---------------------
This addresses an issue in the stratified tank where the use mass flow rate was used when the source mass flow rate should have been used. Fortunately, this doesn't affect the energy balance on the tank since the variable in question is only used for the unmet energy/rate reporting variable calculation. 

I branched from 9.1 in case this could go into a bugfix release. But given that it's a minor issue that only affects a single reporting variable it's not terribly urgent to me.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2:
@nmerket it has been 22 days since this pull request was last updated.

nmerket:
In my run of the regression tests, I saw one file with big mathdiffs: `WaterHeaterStandAlone`. These are the only diffs in that file:

stratified_tank_unmet_demand

This is to be expected because the correct flow rate is being used to calculate those unmet loads outputs now, which was the point of this bugfix. 

In my runs, I'm also seeing a few table diffs, but they're all the annual heat emissions and the difference is in the 10th significant digit. 

We'll see if CI turns up the same things.

nmerket:
Just looked at the CI results, it looks like the table diffs only showed up in my runs and the only difference was the ESO diffs plotted above.

mjwitte:
@nmerket CI only runs design days which supresses ABUPS and many of the table outputs, so those diffs don't come up on CI.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8291,7 +8291,7 @@ namespace WaterThermalTanks {
                             return Qheatpump * node.HPWHWrappedCondenserHeatingFrac;
                         }
                     } else {
-                        return node.UseMassFlowRate * Cp * (Tank.SourceInletTemp - Tavg[i]);
+                        return node.SourceMassFlowRate * Cp * (Tank.SourceInletTemp - Tavg[i]);
```
Myoldmopar, line 8294: Yeah, I mean this looks much better.

tst/EnergyPlus/unit/WaterThermalTanks.unit.cc
```diff
@@ -2051,5 +2051,134 @@ TEST_F(EnergyPlusFixture, StratifiedTankCalc)
...
+
+    WaterThermalTanks::CalcWaterThermalTankStratified(TankNum);
+
+    Real64 EnergySum = 0.0;
+    for (int i = 0; i < Tank.Nodes; ++i) {
+        auto &node = Tank.Node[i];
+        EnergySum += node.Mass * Cp * (node.Temp - 60.0);
+    }
+    Real64 Esource = Tank.SourceEffectiveness * Tank.SourceMassFlowRate * Cp * (Tank.SourceInletTemp - Tank.Node(Tank.SourceOutletStratNode).TempAvg) * TimeStepSys * SecInHour;
+    EXPECT_NEAR(Esource, EnergySum, EnergySum * 0.001);
```
Myoldmopar, line 2181: Very nice unit test :+1:

Myoldmopar (approved):

PR #7280, opened 2019-04-25
Correct load accounting for AirloopHVAC:UnitarySystem when used as zone equipment

### LINKED ISSUE #7264: Sequential load fraction bug w/ zonal UnitarySystem equipment
Issue overview
--------------
I'm testing a building in E+ 9.1 that has a zonal unitary system plus an ideal air system, each serving 50% of the heating load using either UniformLoad or SequentialLoad. However, the simulation results for the SequentialLoad simulations are incorrectly varying depending on Zone Equipment Sequence order. One of the SequentialLoad runs is perfectly matching the UniformLoad case while the other is not. (I've separately confirmed that the UniformLoad results are correct by comparing them to a building with the zonal unitary system meeting 100% of the heating load.)

Interestingly enough, I'm not seeing this issue with other zonal equipment (e.g., PTACs) or unitary systems on air loops. It's only showing up for zonal unitary systems.

Files:
[UniformLoad-good.idf.txt](https://github.com/NREL/EnergyPlus/files/3061448/UniformLoad-good.idf.txt)
[SequentialLoad-good.idf.txt](https://github.com/NREL/EnergyPlus/files/3061451/SequentialLoad-good.idf.txt)
[SequentialLoad-bad.idf.txt](https://github.com/NREL/EnergyPlus/files/3061452/SequentialLoad-bad.idf.txt)

IDF | Heating [GJ] | Heating Fraction
-- | -- | --
UniformLoad-good.idf | 17.91 | 0.50
SequentialLoad-good.idf | 17.91 | 0.50
SequentialLoad-bad.idf | 12.58 | 0.35

@nmerket @mjwitte 

### Details

### Checklist

rraustad:
Pull request overview
---------------------
This fixes #7264. When using AirloopHVAC:UnitarySystem as zone equipment the amount of load met by this equipment is not passed back to the building model (i.e., load met by this equipment = 0). For this reason, any other equipment in this zone will not meet the correct load if that equipment operates after the zonal UnitarySystem component. This has been corrected.

Work around:
In ZoneHVAC:EquipmentList, list the operation of the AirloopHVAC:UnitarySystem last in the list.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes n/a


src/EnergyPlus/UnitarySystem.cc
```diff
@@ -249,6 +249,31 @@ namespace UnitarySystems {
         if (designSpecMSHP.size() > 0) designSpecMSHP.clear();
     }
 
+    void UnitarySys::simulate(std::string const &Name,
```
rraustad, line 252: overload simulate function to accept sysOutputProvided and latOutputProvided used only for zone equipment

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -258,6 +283,33 @@ namespace UnitarySystems {
                               int const ZoneOAUnitNum,
                               Real64 const OAUCoilOutTemp,
                               bool const ZoneEquipment)
+    {
+        Real64 sysOutputProvided = 0.0;
```
rraustad, line 287: original function simply uses locals now

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -258,6 +283,33 @@ namespace UnitarySystems {
...
+                    HeatActive,
+                    CoolActive,
+                    ZoneOAUnitNum,
+                    OAUCoilOutTemp,
+                    ZoneEquipment,
+                    sysOutputProvided,
+                    latOutputProvided);
+    }
+
+    void UnitarySys::simulateSys(std::string const &Name,
```
rraustad, line 302: and each of those functions above now call the original simulate routine that now passes sysOutputRequired and latOutputRequired

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -3618,7 +3620,9 @@ namespace ZoneEquipmentManager {
                                        CoolingActive,
                                        OAUnitNum,
                                        OAUCoilOutTemp,
-                                       ZoneEquipFlag);
+                                       ZoneEquipFlag,
```
rraustad, line 3626: call to zone equipment now includes SysOutputProvided and LatOutputProvided just like all other zone equipment

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4499,85 +4503,84 @@ namespace ZoneEquipmentManager {
         // Sensible output updates
         auto &thisZEqList(DataZoneEquipment::ZoneEquipList(ctrlZoneNum));
         switch (thisZEqList.LoadDistScheme) {
-        case DataZoneEquipment::LoadDist::SequentialLoading:
```
rraustad: I didn't change anything here. Certainly don't merge this if it changes anything.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -2820,30 +2821,30 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_WaterCoilSPControl_Latent)
...
     DataLoopNode::Node(coolingCoilAirOutletNodeIndex).TempSetPoint = 20.0;
     thisSys->simulate(thisSys->Name, FirstHVACIteration, AirLoopNum, CompIndex, HeatActive, CoolActive, ZoneOAUnitNum, OAUCoilOutTemp, ZoneEquipment);
 
     // check that CW coil air outlet node is at set point
-    EXPECT_NEAR(DataLoopNode::Node( coolingCoilAirOutletNodeIndex ).Temp, DataLoopNode::Node( coolingCoilAirOutletNodeIndex ).TempSetPoint, 0.001 );
+    EXPECT_NEAR(DataLoopNode::Node(coolingCoilAirOutletNodeIndex).Temp, DataLoopNode::Node(coolingCoilAirOutletNodeIndex).TempSetPoint, 0.001);
     // check that CW coil air outlet node humrat is >= set point
     EXPECT_LE(DataLoopNode::Node(coolingCoilAirOutletNodeIndex).HumRat, DataLoopNode::Node(coolingCoilAirOutletNodeIndex).HumRatMax);
     // CW air inlet node temp is greater than CW air outlet node temp
-    EXPECT_GT( DataLoopNode::Node( coolingCoilAirInletNodeIndex ).Temp, DataLoopNode::Node( coolingCoilAirOutletNodeIndex ).Temp );
```
rraustad: I'm disliking automatic reformatting.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -4024,13 +4025,27 @@ TEST_F(EnergyPlusFixture, UnitarySystemModel_GetInput)
     thisSys->m_ControlType = UnitarySys::ControlType::CCMASHRAE; // control type = 3
     thisSys->m_ValidASHRAECoolCoil = true;
     thisSys->m_ValidASHRAEHeatCoil = true;
-
-    thisSys->simulate(thisSys->Name, FirstHVACIteration, AirLoopNum, CompIndex, HeatActive, CoolActive, ZoneOAUnitNum, OAUCoilOutTemp, ZoneEquipment);
+    Real64 sysOutputProvided = 0.0;
```
rraustad, line 4061: test new overload function

src/EnergyPlus/ZoneEquipmentManager.cc
rraustad: Looks like Clang just pulled up the open bracket and moved every line within that bracket left to realign the code. So this looks OK, except the Clang is flaky. Can anyone proficient with Clang tell us why this keeps happening?

rraustad:
Tested the defect file by reporting Zone Air System Sensible Heating Rate and zone supply inlet node mass flow rate and enthalpy and zone enthalpy. Made calculation for UnitarySystem and IdealLoads delivered capacity using Mdot*deltaH for each equipment type.

Light blue (Good Test File) and orange (Bad Test File) are the total Zone Air System Sensible Heating Rate reports for each defect file. The two equipment models in these files should each meet 1/2 the load. As shown by the green, yellow, dark blue and red lines, this is now happening. This issue is now corrected.

image

rraustad:
OK, I pulled the trigger too soon, so this isn't finished yet. I just recalled that I failed to include SysOutputProvided for set point based equipment. Comments are sill welcome because they will also apply to this next change which is identical to what I've done to date.

Top level control only passes down sysOutputRequired and latOutputRequired if it is load based control. This also needs to happen for set point based control.

                auto const SELECT_CASE_var(this->m_ControlType);
                if (SELECT_CASE_var == ControlType::Setpoint) {
                    if (ZoneEquipment) {
                        this->controlUnitarySystemtoSP(0, FirstHVACIteration, CompOn, OAUCoilOutTemp, HXUnitOn);
                    } else {
                        this->controlUnitarySystemtoSP(AirLoopNum, FirstHVACIteration, CompOn, OAUCoilOutTemp, HXUnitOn);
                    }
                } else if (SELECT_CASE_var == ControlType::Load || SELECT_CASE_var == ControlType::CCMASHRAE) {
                    if (ZoneEquipment) {
                        this->controlUnitarySystemtoLoad(
                            0, FirstHVACIteration, CompOn, OAUCoilOutTemp, HXUnitOn, sysOutputProvided, latOutputProvided);
                    } else {
                        this->controlUnitarySystemtoLoad(
                            AirLoopNum, FirstHVACIteration, CompOn, OAUCoilOutTemp, HXUnitOn, sysOutputProvided, latOutputProvided);
                    }
                }

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: I don't think it's being flaky, just fixing formatting that wasn't proper before. In this particular spot, that extra set of braces embracing the case block isn't necessary. You can either live with these changes as they come in, or turn off auto formatting in VS and just manually format selected sections that you've worked on.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -1325,7 +1342,17 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiStageElecHeatCoil_Only)
     DataGlobals::BeginEnvrnFlag = true; // act as if simulation is beginning
 
     // Heating mode
-    thisSys->simulate(thisSys->Name, FirstHVACIteration, AirLoopNum, CompIndex, HeatActive, CoolActive, ZoneOAUnitNum, OAUCoilOutTemp, ZoneEquipment);
```
rraustad, line 1328: test new overload function

shorowit (approved):
I didn't look at the code but ran the build against all of my various residential HVAC tests. It fixed the previously failing tests, so it gets a 👍 from me.

nmerket (approved):
This looks good to me. And if @shorowit is happy with the results coming out of it, I'm happy.

src/EnergyPlus/UnitarySystem.cc
nmerket: Also, this was a clever approach.

rraustad: @mjwitte suggested a new function with a different signature. I thought about that and this is what I came up with. Clever maybe, but if there is a better way I am always willing to learn. Thanks for the review.

rraustad:
The UnitarySystem model currently uses SequencedOutputRequiredToCooling/HeatingSP as the loads indicators. I was going to go back and make sure zone equipment used RemainingOutputRequired as suggested. However, it appears that this is what is happening now given the code in UpdateSystemOutputRequired which is called at the end of SimZoneEquipment. So for now I did not change that part of the code in UnitarySystem.cc (at line 7261 in this branch or 7201 in develop).

    if (PrioritySimOrder(EquipPriorityNum).CoolingPriority + 1 <= energy.NumZoneEquipment) {
        energy.SequencedOutputRequiredToCoolingSP(PrioritySimOrder(EquipPriorityNum).CoolingPriority + 1) = energy.RemainingOutputReqToCoolSP;

nrel-bot-2:
@mjwitte @rraustad it has been 14 days since this pull request was last updated.

nrel-bot-2c:
@mjwitte @rraustad it has been 7 days since this pull request was last updated.

mjwitte:
@rraustad Looking at the load assignment code for unitary as zone equipment [here](https://github.com/NREL/EnergyPlus/blob/e96d7c10b0d7b7172e4793d368cf2d35343a1799/src/EnergyPlus/UnitarySystem.cc#L7261-L7290).

```
if (this->m_ZoneSequenceCoolingNum > 0 && this->m_ZoneSequenceHeatingNum > 0) {
```
We do allow 0 (zero) as a valid sequence number in the zone equipment list. So, this will fall to the bottom if block if the user puts a zero for cooling or heating. Once upon a time, zero was supposed to mean don't run, but I don't see that happening anywhere (here or in SetZoneEquipmSimOrder) so I'm not sure why we still allow zero, but I'm not suggesting we change that, just suggesting this needs to be expected here as a possibility. So, the heating and cooling sequence need to be handled separately here.  I'll make a test file like that to see what happens.

When the above if is false, then this block executes:
```
                    ZoneLoad = DataZoneEnergyDemands::ZoneSysEnergyDemand(this->ControlZoneNum).RemainingOutputRequired;
                    QToCoolSetPt = DataZoneEnergyDemands::ZoneSysEnergyDemand(this->ControlZoneNum).OutputRequiredToCoolingSP;
                    QToHeatSetPt = DataZoneEnergyDemands::ZoneSysEnergyDemand(this->ControlZoneNum).OutputRequiredToHeatingSP;
                    MoistureLoad = DataZoneEnergyDemands::ZoneSysMoistureDemand(this->ControlZoneNum).OutputRequiredToDehumidifyingSP;
```
Shouldn't all of these be `Remaining...`?

mjwitte (commented):
Merged in develop, ran the defect files and confirmed (again) that the fix works for that case.

Tested cases with zeros in the equipment list. For uniformload, a zero heating sequence results in zero heating operation. For sequential, zero goes first and produces the same result (as long as the fractions are adjusted appropriately).

Ran `*Unitary*` unit tests locally and they all run fine. Running all unit tests sequentially fails in `EnergyPlusFixture.DesiccantDehum_RegenAirHeaterHWCoilSizingTest` but that's probably an uncleared variable in develop. (develop fails on a different test when running all tests sequentially).

rraustad:
Thanks @mjwitte.
 
 - [x] Convert load checks to something like:

```
if (this->m_ZoneSequenceCoolingNum > 0 && this->m_ZoneSequenceHeatingNum > 0) {
} else if (this->m_ZoneSequenceCoolingNum > 0 && this->m_ZoneSequenceHeatingNum == 0) {
} else if (this->m_ZoneSequenceCoolingNum == 0 && this->m_ZoneSequenceHeatingNum > 0) {
} else {
}
```
or
```
if (this->m_ZoneSequenceCoolingNum > 0) {
 process cooling load
} else {
 process cooling load
}

if (this->m_ZoneSequenceHeatingNum > 0) {
 process heating load
} else {
 process heating load
}
```

 - [x] review use of OutputRequiredToXXXXingSP and RemainingOutputRequiredToXXXXingSP (i.e., revfiew UpdateSystemOutputRequired again)

 - [ ] correct unit test failure in sequential runs for DesiccantDehum_RegenAirHeaterHWCoilSizingTest (not needed in latest branch?)

mjwitte:
@rraustad I was thinking you could just look at the cooling and heating loads separately, but maybe that doesn't fly. Also, the failing unit test was likely caused by another pull request. When you merge in current develop, it should be ok now.

rraustad:
I see your intent and want to make sure logic is sound. Will think more about when I implement.

rraustad:
I'm confused as to which RemainingOutputRequired(1-n) is associated with each piece of zone equipment because PrioritySimOrder gets sorted early on as necessary to put the equipment in sim order in SetZoneEquipSimOrder. I thought RemainingOutputRequired was zone indexed, not sequence order indexed, so then why is a sequence order used to access the load? This will take more time to weed out.

mjwitte:
`RemainingOutputRequired` *is* zone indexed, but it should only be used if there are no valid sequence numbers for this unit (which should never happen, or maybe happen only in an early pass?). The sequenced loads are necessary when the UnitarySystem is an airloop, because the systems get simulated before the zone equipment (the terminal unit). When UnitarySystem is zone equipment, you could just skip the sequenced stuff altogether and use the `Remaining...` loads. It should work either way for unitary as zone equipment.

rraustad:
Agreed. But the index to SequencedOutputRequired is 1-n (i.e., it's not a vector) for each zone and since these numbers can be input as 0 to n(-1?) then I am not sure yet how to access the correct load. It's not the sequence number I need it's the *ordered* equipment list number. When this was added to Furnace I don't know if the test for > 0 was to protect from a crash or to isolate a sequence of 0 as a special case. I agree that the sequenced loads are intended for air loop equipment but these could also be used for zone equipment. It's just the connection between the zone equipment (air loop terminal or zone HVAC equipment) and the load index that eludes me at this point  (Furnace(FurnaceNum).ZoneSequenceCoolingNum in this example could be 0 or 1 for the first piece of equipment intended to operate).
 
```
        if (Furnace(FurnaceNum).ZoneSequenceCoolingNum > 0 && Furnace(FurnaceNum).ZoneSequenceHeatingNum > 0) {
            ZoneLoadToCoolSPSequenced = ZoneSysEnergyDemand(Furnace(FurnaceNum).ControlZoneNum)
.SequencedOutputRequiredToCoolingSP(Furnace(FurnaceNum).ZoneSequenceCoolingNum);
         ...snip...
        } else {
            ZoneLoad = ZoneSysEnergyDemand(Furnace(FurnaceNum).ControlZoneNum).RemainingOutputRequired;
            MoistureLoad = ZoneSysMoistureDemand(Furnace(FurnaceNum).ControlZoneNum).OutputRequiredToDehumidifyingSP;
        }
```

Also, the sequence numbers don't have to be in numerical order without a gap. I ran with 0 and 2 and it ran fine (and by the way with the correct sequenced load index for UnitarySystem) and then changed it to 0 and 3 as shown here. Note the input field name is incorrect in the warning and I have fixed that. It also says the equipment sequence must be > 0. So 0 is invalid? which would correct the indexing problem above (except for input files that do use 0).

```
ZoneHVAC:EquipmentList,
  living Equipment List,                  !- Name
  SequentialLoad,                         !- Load Distribution Scheme
  ZoneHVAC:IdealLoadsAirSystem,           !- Zone Equipment Object Type 2
  res ideal heat,                         !- Zone Equipment Name 2
  0,                                      !- Zone Equipment Cooling Sequence 2
  0,                                      !- Zone Equipment Heating or No-Load Sequence 2
  0.25,                                      !- Zone Equipment Sequential Cooling Fraction 2
  0.25,                                      !- Zone Equipment Sequential Heating Fraction 2
  AirLoopHVAC:UnitarySystem,              !- Zone Equipment Object Type 1
  res unit heater unitary system,         !- Zone Equipment Name 1
  3,                                      !- Zone Equipment Cooling Sequence 1
  3,                                      !- Zone Equipment Heating or No-Load Sequence 1
  1,                                      !- Zone Equipment Sequential Cooling Fraction 1
  1;                                    !- Zone Equipment Sequential Heating Fraction 1

   ** Severe  ** GetZoneEquipmentData1: ZoneHVAC:EquipmentList="LIVING EQUIPMENT LIST".
   **   ~~~   ** invalid Zone Equipment Sequential Cooling Fraction=[3].
   **   ~~~   ** equipment sequence must be > 0 and <= number of equipments in the list.
   **   ~~~   ** only 2 in the list.
```

mjwitte:
You don't need the equipment number - that will confuse matters. The sequenced loads are ordered by the list that comes back from `SetZoneEquipSimOrder` which (I'm 99% certain) renumbers everything continuously from 1-n. Perhaps test with an airterminal set to 0, or skipping a number. The sequence numbers should come back complete with no zero and no skips. The more we discuss this, the more I think UnitarySystem as zone equipment should act like every other zone equipment and only use the zone-level remaining loads - no sequence  confusion.

rraustad:
The numbers that come back from `SetZoneEquipSimOrder` are the user inputs for Zone Equipment Cooling/Heating Sequence.

```
ZoneHVAC:EquipmentList,
  living Equipment List,                  !- Name
  SequentialLoad,                         !- Load Distribution Scheme
  AirLoopHVAC:UnitarySystem,              !- Zone Equipment Object Type 1
  res unit heater unitary system,         !- Zone Equipment Name 1
  0,                                      !- Zone Equipment Cooling Sequence 1
  2,                                      !- Zone Equipment Heating or No-Load Sequence 1
  1,                                      !- Zone Equipment Sequential Cooling Fraction 1
  1,                                    !- Zone Equipment Sequential Heating Fraction 1
  ZoneHVAC:IdealLoadsAirSystem,           !- Zone Equipment Object Type 2
  res ideal heat,                         !- Zone Equipment Name 2
  2,                                      !- Zone Equipment Cooling Sequence 2
  1,                                      !- Zone Equipment Heating or No-Load Sequence 2
  0.25,                                      !- Zone Equipment Sequential Cooling Fraction 2
  0.25;                                      !- Zone Equipment Sequential Heating Fraction 2
```
And I think that user input should point to 1-n here, so if the user enters 0 then there will be a crash.

SequencedOutput

You can see the user input here:

PrioritySimOrder

I think ZoneSysEnergyDemand(CntlZoneNum).SequencedOutputRequiredToCoolingSP(SequenceNum) is always in the same order such that if I can get past retrieving the CoolingPriority = 0 here this would work fine.

```
    thisSys.m_ZoneSequenceCoolingNum = DataZoneEquipment::ZoneEquipList(DataZoneEquipment::ZoneEquipConfig(ControlledZoneNum).EquipListIndex).CoolingPriority(EquipNum);
    thisSys.m_ZoneSequenceHeatingNum = DataZoneEquipment::ZoneEquipList(DataZoneEquipment::ZoneEquipConfig(ControlledZoneNum).EquipListIndex).HeatingPriority(EquipNum);
```

Not sure what to do here ... fix it for all equipment or just fix this in UnitarySystem. I vote to fix for all but not sure where to do that. I guess right where these input are read in.

mjwitte:
When UnitarySystem is used as zone equipment, it should *not* use the sequenced loads. It should use the remaining loads the same as other zone equipment. I would remove the code that's setting and/or checking priority when UnitarySystem is zone eqiupment.

When used as airloop equipment, it will error out (with a somewhat misleading message) if cooling priority comes back zero, but not if heating priority comes back zero. There's no crash, however, because there are checks in unitary and furnace for both cooling and heating priority > 0, otherwise do the wrong thing (use the full cooling or heating load). So, I would just change the check [here](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/UnitarySystem.cc#L8582-L8587) to check for cooling or heating priority == 0 and beef up the error message. Then make a similar change in the other furnaces that have similar code.

rraustad:
Well, I got the sequencing in order now from 1 to n and the ZoneEquipmentManager_DistributeUniformPLR unit test says there is something special about a user input for sequence = 0 for the UniformPLR case.  I guess I only put in order if not using UniformPLR. But I can't access SequencedOutputRequired(SequenceNum) if 0 is the sequence number ???? And I can't use RemainingOutputRequired if the load is supposed to be 0. So now what? I need another struct variable that holds the order 1 to n regardless of sequence magnitude or maybe I can return the correct index using getPrioritiesforInletNode?

```
        "ZoneHVAC:EquipmentList,",
        " Space Equipment,          !- Name",
        " UniformPLR,               !- Load Distribution Scheme",
        " ZoneHVAC:AirDistributionUnit,  !- Zone Equipment 1 Object Type",
        " Air Terminal 1 ADU,       !- Zone Equipment 1 Name",
        " 1,                        !- Zone Equipment 1 Cooling Sequence",
        " 1,                        !- Zone Equipment 1 Heating or No-Load Sequence",
        " ,                         !- Zone Equipment 1 Sequential Cooling Fraction",
        " ,                         !- Zone Equipment 1 Sequential Heating or No-Load Fraction",
        " ZoneHVAC:AirDistributionUnit,  !- Zone Equipment 2 Object Type",
        " Air Terminal 2 ADU,       !- Zone Equipment 2 Name",
        " 2,                        !- Zone Equipment 2 Cooling Sequence",
        " 2,                        !- Zone Equipment 2 Heating or No-Load Sequence",
        " ,                         !- Zone Equipment 2 Sequential Cooling Fraction",
        " ,                         !- Zone Equipment 2 Sequential Heating or No-Load Fraction",
        " ZoneHVAC:AirDistributionUnit,  !- Zone Equipment 2 Object Type",
        " Air Terminal 3 ADU,       !- Zone Equipment 3 Name",
        " 0,                        !- Zone Equipment 3 Cooling Sequence",
        " 3,                        !- Zone Equipment 3 Heating or No-Load Sequence",
        " ,                         !- Zone Equipment 3 Sequential Cooling Fraction",
        " ;                         !- Zone Equipment 3 Sequential Heating or No-Load Fraction",


    // UniformPLR Test 4 - Cooling, FirstHVACIteration = false, only 2 pieces of equipment are active for cooling
    firstHVACIteration = false;
    InitSystemOutputRequired(ZoneNum, firstHVACIteration);
    DistributeSystemOutputRequired(ZoneNum, firstHVACIteration);
    EXPECT_EQ(energy.SequencedOutputRequired(1), thisZEqList.CoolingCapacity(1) * plr);
    EXPECT_EQ(energy.SequencedOutputRequired(2), thisZEqList.CoolingCapacity(2) * plr);
    EXPECT_EQ(energy.SequencedOutputRequired(3), 0.0);
    EXPECT_EQ(energy.SequencedOutputRequiredToHeatingSP(1), thisZEqList.CoolingCapacity(1) * plr);
    EXPECT_EQ(energy.SequencedOutputRequiredToHeatingSP(2), thisZEqList.CoolingCapacity(2) * plr);
    EXPECT_EQ(energy.SequencedOutputRequiredToHeatingSP(3), 0.0);
    EXPECT_EQ(energy.SequencedOutputRequiredToCoolingSP(1), thisZEqList.CoolingCapacity(1) * plr);
    EXPECT_EQ(energy.SequencedOutputRequiredToCoolingSP(2), thisZEqList.CoolingCapacity(2) * plr);
    EXPECT_EQ(energy.SequencedOutputRequiredToCoolingSP(3), 0.0);
```

And here is where sequence = 0 is used (i.e., thisZEqList.HeatingPriority(equipNum) = 0):

```
                if (energy.TotalOutputRequired >= 0.0) {
                    if (thisZEqList.HeatingPriority(equipNum) > 0) {
                        energy.SequencedOutputRequired(equipNum) = energy.TotalOutputRequired * heatLoadRatio;
                        energy.SequencedOutputRequiredToHeatingSP(equipNum) = energy.OutputRequiredToHeatingSP * heatLoadRatio;
                        energy.SequencedOutputRequiredToCoolingSP(equipNum) = energy.OutputRequiredToCoolingSP * heatLoadRatio;
                        moisture.SequencedOutputRequired(equipNum) = moisture.TotalOutputRequired * heatLoadRatio;
                        moisture.SequencedOutputRequiredToHumidSP(equipNum) = moisture.OutputRequiredToHumidifyingSP * heatLoadRatio;
                        moisture.SequencedOutputRequiredToDehumidSP(equipNum) = moisture.OutputRequiredToDehumidifyingSP * heatLoadRatio;
                    } else {
                        energy.SequencedOutputRequired(equipNum) = 0.0;
                        energy.SequencedOutputRequiredToHeatingSP(equipNum) = 0.0;
                        energy.SequencedOutputRequiredToCoolingSP(equipNum) = 0.0;
                        moisture.SequencedOutputRequired(equipNum) = 0.0;
                        moisture.SequencedOutputRequiredToHumidSP(equipNum) = 0.0;
                        moisture.SequencedOutputRequiredToDehumidSP(equipNum) = 0.0;
                    }
                } else {
                    if (thisZEqList.CoolingPriority(equipNum) > 0) {
                        energy.SequencedOutputRequired(equipNum) = energy.TotalOutputRequired * coolLoadRatio;
                        energy.SequencedOutputRequiredToHeatingSP(equipNum) = energy.OutputRequiredToHeatingSP * coolLoadRatio;
                        energy.SequencedOutputRequiredToCoolingSP(equipNum) = energy.OutputRequiredToCoolingSP * coolLoadRatio;
                        moisture.SequencedOutputRequired(equipNum) = moisture.TotalOutputRequired * coolLoadRatio;
                        moisture.SequencedOutputRequiredToHumidSP(equipNum) = moisture.OutputRequiredToHumidifyingSP * coolLoadRatio;
                        moisture.SequencedOutputRequiredToDehumidSP(equipNum) = moisture.OutputRequiredToDehumidifyingSP * coolLoadRatio;
                    } else {
                        energy.SequencedOutputRequired(equipNum) = 0.0;
                        energy.SequencedOutputRequiredToHeatingSP(equipNum) = 0.0;
                        energy.SequencedOutputRequiredToCoolingSP(equipNum) = 0.0;
                        moisture.SequencedOutputRequired(equipNum) = 0.0;
                        moisture.SequencedOutputRequiredToHumidSP(equipNum) = 0.0;
                        moisture.SequencedOutputRequiredToDehumidSP(equipNum) = 0.0;
                    }
                }
```

I would need to return equipNum here:

    void EquipList::getPrioritiesforInletNode(int const inletNodeNum, // Zone inlet node number to match
                                              int &coolingPriority,   // Cooling priority num for matching equipment
                                              int &heatingPriority    // Heating priority num for matching equipment
    )
    {
        bool equipFound = false;
        for (int equipNum = 1; equipNum <= this->NumOfEquipTypes; ++equipNum) {
            if (this->EquipType_Num(equipNum) == AirDistUnit_Num) {
                if (inletNodeNum == DataDefineEquip::AirDistUnit(this->EquipIndex(equipNum)).OutletNodeNum) {
                    equipFound = true;
                }
            } else if (this->EquipType_Num(equipNum) == DirectAir_Num) {
                if (inletNodeNum == DirectAirManager::DirectAir(this->EquipIndex(equipNum)).ZoneSupplyAirNode) {
                    equipFound = true;
                }
            }
            if (equipFound) {
                coolingPriority = this->CoolingPriority(equipNum);
                heatingPriority = this->HeatingPriority(equipNum);
                break;
            }
        }

 since this is used as an index here and other places:

```
            if (ZoneEquipConfig(zoneNum).EquipListIndex > 0) {
                ZoneEquipList(ZoneEquipConfig(zoneNum).EquipListIndex).getPrioritiesforInletNode(zoneInlet, coolingPriority, heatingPriority);
                Furnace(FurnaceNum).ZoneSequenceCoolingNum = coolingPriority;
                Furnace(FurnaceNum).ZoneSequenceHeatingNum = heatingPriority;
            }
```

rraustad:
I just saw your previous comment. Will try that first and only use sequenced loads for air loops.

rraustad:
So why does a 0 load case when sequence = 0 apply only to zone equipment which DOES NOT access the SequencedLoads?

mjwitte:
> So why does a 0 load case when sequence = 0 apply only to zone equipment which DOES NOT access the SequencedLoads?

Probably because at the time I wrote the load distribution functions I was under the mistaken impression that a zero in the sequence list would remove a piece of equipment from being used for heating or cooling. The reality is that getting all of the load distribution to work as expected without breaking legacy options was a head-spinning exercise with lots of trial and error. So, while this all could use a revamp, we just got this stuff working (at least working correctly for well-formed equipment lists).

rraustad:
That's fine. I'm going with your suggestion. But I still think that if I returned equipNum in getPrioritiesforInletNode (which is only called by air loop equipment) it would have fixed things. equipNum is the index to the equipment in the ZoneHVAC:EquipmentList and this index is the same as the order used in the SequencedOutput variables. This also means that sequence = 0 would be allowed for air loop equipment and the warning I am updating would not be needed.

So now, zone equipment will not respond to `\key UniformPLR` since the information is in the SequencedOutput variables and zone equipment does not use those (unless I'm missing where RemainingOutput gets updated based on the SequencedOutput variables).

Any final thoughts?

mjwitte:
The zone-level remaining loads *do* get set to the proper sequenced load, both in [DistributeSystemOutputRequired](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/ZoneEquipmentManager.cc#L4386-L4395) and in [UpdateSystemOutputRequired](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/ZoneEquipmentManager.cc#L4587-L4607).

And the sequenced loads are ordered by - the sequence number? - which could be zero?  Ugh. That's the head-spinning part that needs to be simplified. Let's not touch that here, that should be a separate PR and will require re-testing with @shorowit series of different load distribution options. It could be a simple as adjusting the sequence numbers immediately on input processing for ZoneHVAC:EquipmentList - forcing them to be 1 to n in the same order they came in.

rraustad:
OK so the RemainingOutput is updated, that's good. I tried the 1-n thing and then found that 0 meant something so reverted that out.

In SimZoneEquipment:

            for (EquipTypeNum = 1; EquipTypeNum <= ZoneEquipList(ControlledZoneNum).NumOfEquipTypes; ++EquipTypeNum) {

                UpdateSystemOutputRequired(ActualZoneNum, SysOutputProvided, LatOutputProvided, EquipTypeNum);

The variable name and description look incorrect to me:

    void UpdateSystemOutputRequired(int const ZoneNum,
                                    Real64 const SysOutputProvided,     // sensible output provided by zone equipment (W)
                                    Real64 const LatOutputProvided,     // latent output provided by zone equipment (kg/s)
                                    Optional_int_const EquipPriorityNum // index in PrioritySimOrder for this update

I'm tinkering with changing getPrioritiesforInletNode but now thinking it won't work because the next equipment in the ZoneHVAC:EquipmentList will not necessarily be the next priority equipment to operate. So I'll probably change this back the way it was.

        for (int equipNum = 1; equipNum <= this->NumOfEquipTypes; ++equipNum) {
            if (this->EquipType_Num(equipNum) == AirDistUnit_Num) {
                if (inletNodeNum == DataDefineEquip::AirDistUnit(this->EquipIndex(equipNum)).OutletNodeNum) {
                    equipFound = true;
                }
            } else if (this->EquipType_Num(equipNum) == DirectAir_Num) {
                if (inletNodeNum == DirectAirManager::DirectAir(this->EquipIndex(equipNum)).ZoneSupplyAirNode) {
                    equipFound = true;
                }
            }
            if (equipFound) {
                coolingPriority = equipNum;
                heatingPriority = equipNum;
                break;
            }

We need a test file with multiple equipment in the list with priority order out of order to see what breaks. Or a unit test that calls all these functions and resolves the answer. ZoneEquipmentManager_DistributeUniformPLR is close to what is needed to sort this out. I'm wasting a lot of time on this one but starting to understand.

PrioritySimOrder is used to set sim order but index to equipment in ZoneEquipList is used set RemainingOutput order. I think there is a disjointed access to the RemainingOutput(n+1) variables since the next equipment in the list won't necessarily be the next priority. 

                if (EquipPriorityNum + 1 <= energy.NumZoneEquipment) {
                    energy.SequencedOutputRequired(EquipPriorityNum + 1) = energy.RemainingOutputRequired;
                    moisture.SequencedOutputRequired(EquipPriorityNum + 1) = moisture.RemainingOutputRequired;
                }

mjwitte:
EquipPriorityNum is not the order of the zone equipment list, it's the sorted order in PrioritySimOrder .

mjwitte:
And @shorowit has a set of test files that has things in different orders.

rraustad:
Look at what is passed into UpdateSystemOutputRequired at the top of my comment. It's the loop counter to the zone equipment list, not the PrioritySimOrder list. Even if someone passed in the index to PrioritySimOrder, the n+1 won't point back to the index in the SequencedOutput if the sequence is out of order. I'm over this for now and will try to address the issue and track these oddities.

mjwitte:
I agree to let this rest for now, but for completeness, as I see it in `SimZoneEquipment`, `EquipTypeNum` is the index for the `PrioritySimOrder` list [here](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/ZoneEquipmentManager.cc#L3478-L3510). The variable name is misleading which doesn't help. And that's what is passed in to `UpdateSystemOutputRequired`.

rraustad:
Maybe, but look again at what is passed into UpdateSystemOutputRequired.

ZoneEquipTypeNum (line 3510) is the EquipType_Num integer (as in SingleDuctConstVolNoReheat) from PrioritySimOrder (so the right equipment can be simulated in order), not the loop counter EquipTypeNum (line 3478) that is passed into UpdateSystemOutputRequired (line 3859). I made the same mistake several times. Things flip flop around here that will confuse the developer on equipment order in the various arrays. The arrays in UpdateSystemOutputRequired are based on the order of equipment in ZoneHVAC:EquipmentList. That's what I meant when I said the n+1 (supposed to be next equipment in PrioritySimOrder) isn't quite right (since it's the next equipment in the equipment list).

mjwitte:
The variable names in there are horrible, but I'm confident that the index passed to UpdateSystemOutputRequired is the correct one for sequencing (as long as nobody is sequence = 0).

tst/EnergyPlus/unit/HVACFourPipeBeam.unit.cc
```diff
@@ -1677,13 +1677,13 @@ TEST_F(EnergyPlusFixture, Beam_sizeandSimulateOneZone)
     Real64 NonAirSysOutput = 0.0;
     DataDefineEquip::AirDistUnit(1).airTerminalPtr->simulate(FirstHVACIteration, NonAirSysOutput);
 
-    EXPECT_NEAR(DataLoopNode::Node(1).MassFlowRate, 0.35251094469529615, 0.00001);
-    EXPECT_NEAR(DataLoopNode::Node(15).Temp, 19.191879243000013, 0.00001);
-    EXPECT_NEAR(DataLoopNode::Node(15).MassFlowRate, 0.046012222387687624, 0.00001);
+    EXPECT_NEAR(DataLoopNode::Node(1).MassFlowRate, 0.35251094469529615, 0.0003);
```
rraustad: Not sure why this unit test showed failure but the tolerance was very tight and I relaxed it slightly to correct.

rraustad:
Results haven't changed. Changed defect file to provide 25% of load for first equipment, second picked up remaining load. In this plot UnitarySystem and IdealLoadsAirSystem where switched.
First test (represents good test file) Ideal Loads met 25% of load (red) and UnitarySys met 75% (yellow).
Second test (represents bad test file) UnitarySys met 25% of load (dark blue) and IdealLoads met 75% (green).
Total delivered capacity is the same at 100% of predicted load (light blue and orange).

Delivered load on Y-axis, predicted load on X-axis. Heating only test file.

image

src/EnergyPlus/DataZoneEquipment.cc
```diff
@@ -597,7 +597,7 @@ namespace DataZoneEquipment {
                     if ((thisZoneEquipList.CoolingPriority(ZoneEquipTypeNum) < 0) ||
                         (thisZoneEquipList.CoolingPriority(ZoneEquipTypeNum) > thisZoneEquipList.NumOfEquipTypes)) {
                         ShowSevereError(RoutineName + CurrentModuleObject + "=\"" + AlphArray(1) + "\".");
-                        ShowContinueError("invalid " + cNumericFields(2 * ZoneEquipTypeNum - 1) + "=[" +
+                        ShowContinueError("invalid " + cNumericFields(4 * ZoneEquipTypeNum - 3) + "=[" +
```
rraustad, line 600: Warnings were reporting the wrong field description.

rraustad:
Documenting adjusting sequence number to be greater than 0 if it's ever needed even though it would be hard to find again. Repeat this for heating.
```
// DataZoneEnergyDemands::ZoneSysEnergyDemand(ControlZoneNum).SequencedOutputRequiredToCoolingSP(ZoneSequenceNum); requires
// ZoneSequenceNum to be > 0 so check cooling priority order to ensure sequence is 1 to NumOfEquipTypes if needed
// no errors means max n <= NumOfEquipTypes and no duplicates, warning was issued for 0 above
if (coolPriorityHasZero && !GetZoneEquipmentDataErrorsFound) {
    for (ZoneEquipTypeNum = 1; ZoneEquipTypeNum <= thisZoneEquipList.NumOfEquipTypes; ++ZoneEquipTypeNum) {
        thisZoneEquipList.CoolingPriority(ZoneEquipTypeNum) += 1; // increase values so that all are > 0
    }
    // find min/max element location
    auto minIndex = *std::min_element(thisZoneEquipList.CoolingPriority.begin(), thisZoneEquipList.CoolingPriority.end());
    auto maxIndex = *std::max_element(thisZoneEquipList.CoolingPriority.begin(), thisZoneEquipList.CoolingPriority.end());
    // only decrement if max value > num elements (i.e., there was a gap)
    if (thisZoneEquipList.CoolingPriority(maxIndex) > thisZoneEquipList.NumOfEquipTypes) {
        for (ZoneEquipTypeNum = 1; ZoneEquipTypeNum <= thisZoneEquipList.NumOfEquipTypes; ++ZoneEquipTypeNum) {
            if (ZoneEquipTypeNum == minIndex) continue; // decrement all elements except min location which is now 1
            thisZoneEquipList.CoolingPriority(ZoneEquipTypeNum) -= 1;
        }
    }
}
```

mjwitte:
@rraustad There's a custom check [error](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-0519652d010ced3ca7785a7dc6642b6d1c59caef-x86_64-Linux-Ubuntu-18.04-custom_check.html). It seems to be complaining that cool/heatactive are set to false before calling UnitarySys.Simulate, but this branch isn't changing any of that.
@Myoldmopar Any clues?

Also, @rraustad When running all unit tests serially, this branch fails in `EnergyPlusFixture.CalcVariableSpeedHP` with a fatal error in `ZoneTempPredictorCorrector::GetZoneAirSetpoints`.

mjwitte:
@rraustad Never mind about the unit test thing - that's in develop, not caused by the changes in this branch.

mjwitte:
And BTW, there is no problem in develop unit tests. That was my local problem. So, never mind.

rraustad:
Well, lines 364 and 366 are at the point in UnitarySystem.cc where those bools are set so those warnings must be describing that code. I'll try to comment out those lines, since those bools were already set to false in the calling routine. Not sure why that would cause a warning though.

mjwitte (approved):
Reviewed code changes, tested defect files, and tested some extra variants. All looks good.

@rraustad The unrelated unit test changes seemed fishy (4pipebeam and unitheater). I reverted these locally and the unit tests passed. So, I've pushed that up. If all is green after that, this will drop in.

Issue #3947, opened 2014-01-14
The water heater models were expanded to allow options for how the source side connections requests flow.  The program now supports seperate setpoint control for calls for indirect heating and passive storage tanks. (CR #9176)


axelstudios:
###### Added on 2013-03-18 10:48 by Brent Griffith
--
#### Description
The control logic for water heater tank's source side connection was developed for the use case of indirect heating.   A slightly different control may be desirable when the tank is used for thermal storage.  

Suggestion is to add another input field to water heater objects to indicate if tank is to be controlled as a 1) thermal storage tank or an 2) indirectly heated water heater.  If it is a storage tank, then do things differently
a.	Aggressively lodge flow requests even when no heating is needed by the tank
b.	but if tank temperature is at maximum limit, then donÆt request flow

moved over from CR 9082

[BG 9-10-2013] have also had request to allow indirect heating to one setpoint, and internal heater control to a second setpoint. Ticket #8517.

probably want two fields, one for control mode and another for alternate setpoint schedule. 

BG 9-17-2013] checked in code to add two new fields to mixed and stratified tanks.  can now select among three operation modes. 

pointing to edwin for verification.

--
Addressed in `13.10.03 V8.1.0.005` by Brent Griffith
Last build tested: `13.02.09 V8.0.0.004`

PR #7286, opened 2019-05-01
Tank Source Side Flow Control Mode IndirectHeatAlternateSetpoint Code Not Hit

yzhou601:
Pull request overview
---------------------
The source side flow set by calling function ``PlantMassFlowRatesFunc`` didn't actually work for source side flow control mode: IndirectHeatAlternateSetpoint for indirect water heater system and tankless coils. Relevant question on Unmet Hour: [Click Here](https://unmethours.com/question/30394/indirectly-heated-water-heater-with-a-backup-electric-heater-indirectheatalternatesetpoint/)
The [issue](https://github.com/NREL/EnergyPlus/issues/3947) and [CR9176](https://github.com/NREL/EnergyPlusArchive/commit/dfceb055639261df4354f7b9967d82dee1c9dc71) changes are made under condition of ``MaybeRequestingFlow`` which is not stepped at final iteration in every time step. 
So these codes are added to condition ``ThrottlingFlow`` to fix this problem.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


yzhou601:
This bug fix issue seems not have been covered in tests without any CI test fail. So it's necessary to show my local test results for indirect water heater system models (no backup water heater, 55C alternate setpoint schedule, 52.667C WH Setpoint, 1C deadband).

I expected the source side mass flow rate to be over 0 for time points when tank temperature is below 55C-1C. 

### Run in Develop

#### IndirectHeatAlternateSetpoint

image (4)

#### IndirectHeatPrimarySetpoint

image (6)

### Run in this branch

#### IndirectHeatAlternateSetpoint

image (5)

#### IndirectHeatPrimarySetpoint

image (7)

It shows more insights of this bug, the alternate option is not really working.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -10797,44 +10881,13 @@ namespace WaterThermalTanks {
...
-                            }
-                        } else if (WaterThermalTank(WaterThermalTankNum).TypeNum == StratifiedChilledWaterStorage) {
-                            NeedsCool = true;
-                        }
-
-                    } else if (OutletTemp <= SetPointTemp) {
-                        NeedsCool = false;
-                    }
-                }
+                NeedsHeatOrCool = SourceHeatNeed(WaterThermalTankNum,OutletTemp, DeadBandTemp,SetPointTemp);
```
Myoldmopar: I very much like this duplicated code is extracted out into a worker function.  The format is messy here though, need to include spaces after each comma in an argument list.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -10639,6 +10639,96 @@ namespace WaterThermalTanks {
...
+        // FUNCTION INFORMATION:
+        //       AUTHOR         Yueyue Zhou
+        //       DATE WRITTEN   May 2019
+        //       MODIFIED       na
+        //       RE-ENGINEERED  na
+
+        // PURPOSE OF THIS FUNCTION:
+        // Determine by tank type, tank temperature and control mode if source side flow is needed
+
+        // METHODOLOGY EMPLOYED:
```
Myoldmopar: If you aren't going to actually include a methodology, just eliminate these placeholders/stubs.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -10639,6 +10639,96 @@ namespace WaterThermalTanks {
...
+        // <description>
+
+        // REFERENCES:
+        // na
+
+        // Using/Aliasing
+        using ScheduleManager::GetCurrentScheduleValue;
```
Myoldmopar: I'd prefer you actually include `ScheduleManager::GetCurrentScheduleValue` in the code below rather than having a `using` statement at the top of the function.  In this case, it's not a big deal, but in many places within E+, the usage of `using` is huge and causes uncertainty around where a function or variable is being accessed.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -10639,6 +10639,96 @@ namespace WaterThermalTanks {
...
+
+        //return value initialization
+        bool NeedsHeatOrCool = false;
+
+        Real64 AltSetpointTemp;
```
Myoldmopar: These can be defined at the point of usage since they are only used within that scope, and also declared `const` for clarity.

tst/EnergyPlus/unit/WaterThermalTanks.unit.cc
```diff
@@ -2471,3 +2471,120 @@ TEST_F(EnergyPlusFixture, DesuperheaterTimeAdvanceCheck){
     EXPECT_GT(Tank.TankTemp, 50);
 
 }
+
```
Myoldmopar: I was expecting this unit test to include a direct call to `SourceHeatNeed` since you extracted it out into a worker function, but I guess it's called from within other calls below.

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -10639,6 +10639,96 @@ namespace WaterThermalTanks {
         return PLRResidualHPWH;
     }
 
+    bool SourceHeatNeed(int const WaterThermalTankNum,
```
Myoldmopar: A nice clean up here would be to either:
- Make this function a method on the tank data structure, that way you could just call `WaterThermalTank(WaterThermalTankNum).sourceHeatNeed(outletTemp, dbTemp, spTemp)` and eliminate the array lookups, or
- Pass in a const tank object as an argument instead of the tank num, again eliminating the array lookups.

Myoldmopar (changes requested):
The functional changes here look good, but I'd like to see some improvements made before this gets merged.  Take a look at my comments and let me know what you think.

src/EnergyPlus/WaterThermalTanks.cc
yzhou601: Thanks, Edwin, I passed the tank object as SourceHeatNeed function argument (option 2)

tst/EnergyPlus/unit/WaterThermalTanks.unit.cc
yzhou601: Yes it's calling ``PlantMassFlowRatesFun`` inside ``InitWaterThermalTank`` which is not straghtforward. It's a good idea to directly hit this function. I addressed this in the new commits below.

yzhou601:
~~@Myoldmopar Do you know why these commits that happened in develop branch all appeared in this PR once I pulled develop into this branch?~~
(seemed to be normal again.)

Myoldmopar:
Yeah I don't know what happened there.  But your [develop](https://github.com/yzhou601/EnergyPlus/tree/develop) branch is 8 commits ahead of the NREL/EnergyPlus develop branch.  So that's potentially problematic, but it seems not with the current issue.

yzhou601:
> Yeah I don't know what happened there. But your [develop](https://github.com/yzhou601/EnergyPlus/tree/develop) branch is 8 commits ahead of the NREL/EnergyPlus develop branch. So that's potentially problematic, but it seems not with the current issue.

Yeah, these 8 commits are all pulling the latest changes from NREL/EnergyPlus/develop, might be fine?

Myoldmopar:
No, your develop branch should never be _ahead_ of the NREL develop branch.  Pulling in the latest changes from NREL develop should cause a git fast-forward, never a merge, so it should never be ahead.  Something went wrong.  But again, I don't think it caused an issue here, just something you need to get addressed before it accidentally does.

mjwitte:
@Myoldmopar Um, there hasn't been a commit to this branch for 6 or 7 days, yet CI keeps making the rounds on this one. Working on it again now.

Myoldmopar:
Yeah, CI has been having a bit of trouble with some external PRs lately.  Nothing serious, just wastes some cycles running extra commits.  I'm currently unclear as to the reason, but will dig into it soon after the release, if not before.

tst/EnergyPlus/unit/WaterThermalTanks.unit.cc
```diff
@@ -1992,7 +1992,7 @@ TEST_F(EnergyPlusFixture, StratifiedTankCalc)
     for (int i = 0; i < Tank.Nodes; ++i) {
         NodeTemps[i] = Tank.Node[i].Temp;
     }
-
+
```
Myoldmopar, line 1995: I'm not sure what editor you are using to write C++, but it shouldn't be leaving these trailing whitespaces in the files.  In most editors there is an option to clean any trailing whitespace when you save a file, I'd look for that option.  I'm not going to hold up this PR just for that though.

Myoldmopar (approved):

Myoldmopar:
Code changes look much better now.  I'm testing locally.

Myoldmopar:
Merged up with develop, unit test suite passes, and ran a sampling of regression files and got no diffs.  I'm merging this in.

PR #7294, opened 2019-05-08
Air Thermal Boundary - Radiant Exchange

mjwitte:
Pull request overview
---------------------
Air wall Part 1 - Radiant exchange.
 - [x] Add new Construction:AirBoundary to specify an air thermal boundary between two zones.
 - [x] Implement air boundary as IRTSurface for radiant exchange.
 - [x] Implement the concept of grouped zones for radiant exchange which models all of the surfaces in the zones connected by air boundaries as if they were in a single zone.
 - [x] Fix ZoneProperty:UserViewFactors:bySurfaceName to work with grouped zones (enclosures)
 - [x] Implement distribution of radiant internal gains for grouped zones.
 - [x] Allow radiant equipment view factors to cross zone boundaries within an enclosure
 - [x] New example file: 5ZoneAirCooled_AirBoundaries.idf
 - [x] I/O Ref
 - [x] Engineering Ref
 - [x] Unit tests

*This PR implements only the radiant surface exchange and radiant internal gains part of the air boundary model. Solar gains (and daylighting), and air exchange are planned to be separate PRs building on this one.*

### Diffs Discussion
1. Files with eio diffs and big table diffs are all related to this header change:
"Zone Name" --> "Zone/Enclosure Name"
```
-! <Surface View Factor Check Values>,Zone Name,Original Check Value,Calculated Fixed Check Value,Final Check Value,Number of Iterations,Fixed RowSum Convergence,Used RowSum Convergence
+! <Surface View Factor Check Values>,Zone/Enclosure Name,Original Check Value,Calculated Fixed Check Value,Final Check Value,Number of Iterations,Fixed RowSum Convergence,Used RowSum Convergence
```
2. The file with err diffs is similar:
```
-   ** Warning ** Surfaces in Zone="ZONE ONE" do not define an enclosure.
+   ** Warning ** Surfaces in Zone/Enclosure="ZONE ONE" do not define an enclosure.
```

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Design ##
+
+The overall plan is to tackle radiant exchange first, then solar/daylighting, then air exchange.
+
+### Surface Class ###
+Anywhere that `Surface(n).Class` is used will be evaluated to add a case or if block for 
+the new `SurfaceClass_AirWall`.
+
+### SuperZones ###
```
macumber: Is it possible to not have to explicitly list super zones?  You should be able to determine the adjacency matrix for zones connected by air walls using [this algorithm](https://stackoverflow.com/questions/32164012/how-to-get-distance-matrix-from-adjacency-matrix-matlab).

macumber: Maybe I misread this, it sounds like a user does not need to explicitly enter in a SuperZone (there is no new IDD object)?  That would be good.  The only downside I can see is in reporting.  For that you could have a SuperZone report that lists the clustered SuperZones, variables could be reported for SuperZones.

macumber: I actually wonder if the concept of SuperZones could be extended or used for other purposes.  Specifically, could they be used to partition the building into mostly disconnected pieces that could be simulated in parallel.

shorowit: > 
> 
> Is it possible to not have to explicitly list super zones? You should be able to determine the adjacency matrix for zones connected by air walls using [this algorithm](https://stackoverflow.com/questions/32164012/how-to-get-distance-matrix-from-adjacency-matrix-matlab).

There are use cases where EnergyPlus doesn't always have a full, self-contained 3D geometry. We do this for some of our residential work. I believe Title 24 for commercial buildings does as well? So I'm in favor of the explicit list.

macumber: Should you be using airwalls if you don't have fully enclosed geometry?

shorowit: Maybe not often, but I wouldn't say never....

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+        for (ISurf = Zone(ZoneNum).SurfaceFirst; ISurf <= Zone(ZoneNum).SurfaceLast; ++ISurf) {
+```
+
+This example is from daylighting, so it will need to be changed. Other examples of this usage occur
+in convection coefficients and other places where the current method will not be impacted. 
+This can be addressed by using zone-level lists of surface numbers (which is already being 
+developed in a separate effort).
+
+### Reporting ###
+Various zone-level outputs will be reviewed to determine the proper accounting of
```
macumber: What are the cases that trigger warnings or errors?  Can you have AirWalls with no adjacent surface?  Can it be self-adjacent?  What about zone multipliers?  Can you have subsurfaces on an AirWall?  Zone mixing or airflow network in addition to the airwall?

mjwitte: Lots of good questions that I'll need to flesh out in the NFP. Some intitial clarifications:
1. The AirWall must have an adjacent surface.
2. The AirWall cannot be self-adjacent. I suppose it *could* be allowed but that wouldn't accomplish anything - it would have zero impact on the simulation. The whole idea is to have an opening that lets two zones transmit solar and radiant between them.
3. The intent is that the superzones would be determined simply by looking at the airwall surfaces which must all be an interzone surface. The zone which holds the airwall and the zone which holds it's adjacent surface would be grouped together.
4. Zone multipliers? Good question. The only way this could work would be if all of the zones grouped into the superzone had the same multiplier.
5. The airwall as proposed wouldn't do any mixing itself, so yes, zone mixing and/or airflow network are required to move air between the zones.

mjwitte: 6. No subsurfaces on an AirWall.
7. Perhaps an AirWall subsurface should be handled differently than an AirWall base surface to prevent the entire building from being connected into a single superzone. Maybe the new choice for fenestration objects (subsurfaces) should be "Opening" and it would behave as an interzone window with perfect transmittance. Something to consider.

Myoldmopar: Yeah, this superzone concept just makes me think the user should model the whole space as one zone instead...allow interfaces to deal with the space diversity issues that come with grouping the individual zones into one larger zone before handing it off to the energy calculation engine.

But alas.

If we are going to have an air wall, you've definitely covered a lot, laid out a lot of pitfalls, and clearly have a handle on how complex this will be.  I would say that, since we seem to have ramped up end-user-usage of the airflow network, special emphasis should be given there to ensure this works well with that.  I look forward to seeing this evolve into a full design.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+For simple air exchange (not AirflowNetwork) the user will be expected to input `ZoneMixing` or `ZoneCrossMixing`
+objects the same as always. For simulations using AirflowNetwork, it gets a little more complicated. For smaller openings,
+such as open doorways, the current `AirflowNetwork:MultiZone:Component:DetailedOpening`, 
+`AirflowNetwork:MultiZone:Component:SimpleOpening` and `AirflowNetwork:MultiZone:Component:HorizontalOpening`
+objects should be adequate. For very large openings (horizontal and vertical) another approach will be required. One
+possibility is to use a superzone for the AirflowNetwork calculations. Or perhaps a new opening type could be developed
+to approximate the connection.
+
+#### Pitfalls ####
+ - It's unclear what the values should be for the simple mixing models across an open boundary.
```
EnergyArchmage, line 199: Might consider an input value for Air Changes Per Hour for the opening.  I usually recommend 1 ACH based on the volume of the smaller zone as a basis for determining input for an airflow rate in zone mixing objects.  This might be automated to make it easier for the user with an ACH input to scale, such as higher for horizontal orientation.  I guess this would lead to an air wall construction object that gets referenced.  There might be other inputs that might go into a Construction:AirWall object.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+n/a
+
+## Approach ##
+
+The proposed approach is to remove the air wall surface entirely from most of the heat transfer calculations
+and create combined superzones for solar, daylighting, and radiant exchange calculations. Air exchange is
+more complicated.
+
+### Radiant Exchange ###
+For surface-to-surface radiant exchange, all surfaces in the zones connected by airwalls will be combined into
```
EnergyArchmage: This seems like a great approach, rather ambitious though. Impressive idea.  
Could offer speed improvements during looping if you can run the super zone once while skipping the individual zones.

design/FY2019/Interzone Air Wall NFP and Design.md
EnergyArchmage: The whole point is that the HVAC equipment and internal gains are different between the zones, so combining zones in the first place is not doable.

design/FY2019/Interzone Air Wall NFP and Design.md
macumber: For what it's worth this is kind of what OpenStudio does.  The way you tag an AirWall in OpenStudio is to assign a construction having a single layer of [AirWallMaterial](https://github.com/NREL/OpenStudio/blob/develop/openstudiocore/src/model/AirWallMaterial.hpp).  We added this as a material type with the intention to hang these types of properties onto it (e.g. ACH, transparent, etc).  That way you can also assign different properties to different airwalls in your model.

macumber: Construction:AirWall would be just as good, I can't remember why we used a material instead of a construction to indicate airwalls.

design/FY2019/Interzone Air Wall NFP and Design.md
mjwitte: As I think more about this, the superzone concept does not make sense for smaller airwall connections, like the serving window between a kitchen and dining area. So, airwall subsurfaces (fenestration) may be treated differently - no superzone connection for these smaller openings, but perhaps use the MRT of the adjacent zone as the airwall sub-surface temperature in the radiant exchange of each adjacent zone.

design/FY2019/Interzone Air Wall NFP and Design.md
mjwitte: Sure, I'll consider a construction and/or material type that could be used to control properties of the boundary. Perhaps this should replace Surface Type = AirWall as the trigger? Then the surfaces would still be Wall, Roof, Floor, etc. and the Construction with Material:AirWall would be the trigger to treat this surface differently.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+ will exaggerate that problem.
+ - Within the superzone it is likely that many surfaces will not have a direct view to all of the surfaces
+ in the combined zone. Presently there is no algorithm to exclude certain views. View factors could be determined
+ external to EnergyPlus and imported using `ZoneProperty:UserViewFactors:bySurfaceName`.
+ - Input processing will need to allow view factors (for radiant HVAC equipment) to cross zone boundaries.
+ - Re-simulation for radiant HVAC equipment will involve all zones in the superzone, causing extra computation time.
+ - Reporting for zone-level radiant gains may be confusing. For example, radiant gains which have a source in Zone A
+ may be distributed to other zones as well as Zone A. Even though the source of the gain is Zone A, the other zones
+ should be reporting their share of the distributed gain.
+
```
JasonGlazer, line 161: The nonconvex issue seems to be one that would make the superzone approach not work well.

I'm just imagining two simple rectangular zones next to each other. Between them, half is a regular wall and half is an airwall. Something as simple as the radiant distribution of heat from equipment or a person in one of the zones is no longer as easy since you can't assume all the radiant heat just gets fully distributed in that zone. The position of the equipment or person is now relevant to determine how much of the radiant heat goes to the other zone. Maybe it is just as much of an assumption as it ever was but it seems unintuitive to me. I kept expecting some additional inputs in your NFP that describe the assumptions of how heat, light, and air is being flung around.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+Field: Outside Boundary Condition
+
+. . . If Surface Type is *AirWall* the only valid options are *Surface* and *Zone*.
+
+**FenestrationSurface:Detailed**
+
+Field: Surface Type
+
+. . . *AirWall* indicates an open boundary between two zones. When this option is selected, Construction Name is ignored and the
+Outside Boundary Condition must be either *Surface* or *Zone*.
```
JasonGlazer: "AirWindow" seems funny to call it AirWall in the fenestration object.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+
+## E-mail and  Conference Call Conclusions ##
+
+n/a
+
+## Approach ##
+
+The proposed approach is to remove the air wall surface entirely from most of the heat transfer calculations
+and create combined superzones for solar, daylighting, and radiant exchange calculations. Air exchange is
+more complicated.
```
jasondegraw: @mjwitte Not sure why air exchange should be more complicated, since there's an easy solution - just disallow exchange and lump everything together. That should be fine as long as the zone conditions aren't super different, and that's something we can warn about if it happens. For AirflowNetwork it could easily be optional to treat the zones separately or lumped together (provided that there's something to attach a node to).

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+For surface-to-surface radiant exchange, all surfaces in the zones connected by airwalls will be combined into
+a superzone for the radiant exchange calculations. Radiant gains from lights, people, HVAC equipment, etc. will 
+be distributed to all surfaces in the superzone.
+
+#### Pitfalls ####
+ - Radiant exchange calculations become very slow for zones with a large number of surfaces. This approach
+ will exaggerate that problem.
+ - Within the superzone it is likely that many surfaces will not have a direct view to all of the surfaces
+ in the combined zone. Presently there is no algorithm to exclude certain views. View factors could be determined
+ external to EnergyPlus and imported using `ZoneProperty:UserViewFactors:bySurfaceName`.
```
jasondegraw: The internal view factor algorithm is definitely going to have trouble with the geometries that will result here. View3D should not, but it would make sense to add a script or two to simplify that calculation.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+ account for transmitted solar which passes through to another zone (and the reverse).
+
+### Air Exchange ###
+For simple air exchange (not AirflowNetwork) the user will be expected to input `ZoneMixing` or `ZoneCrossMixing`
+objects the same as always. For simulations using AirflowNetwork, it gets a little more complicated. For smaller openings,
+such as open doorways, the current `AirflowNetwork:MultiZone:Component:DetailedOpening`, 
+`AirflowNetwork:MultiZone:Component:SimpleOpening` and `AirflowNetwork:MultiZone:Component:HorizontalOpening`
+objects should be adequate. For very large openings (horizontal and vertical) another approach will be required. One
+possibility is to use a superzone for the AirflowNetwork calculations. Or perhaps a new opening type could be developed
+to approximate the connection.
```
jasondegraw: Using the superzone approach is the way to go, or at least make it an option. It's going to come down to which representation is best, and with large openings there's a higher probability of two-way flows. I don't think this is something that a new element is going to fix.

design/FY2019/Interzone Air Wall NFP and Design.md
jasondegraw: I'm not a fan of automating the "simple mixing models" or attaching it to the material. Construction, maybe, but shouldn't the air transfer be based on physics?

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,192 @@
...
+
+#### Pitfalls ####
+ - It's unclear what the values should be for the simple mixing models across an open boundary.
+ - AirflowNetwork reporting may be confusing if a superzone approach is used.
```
jasondegraw: I think this can be handled in a way that it won't be confusing.

hongtz68:
Mike,

A nice feature. The air wall helps on grouping (geometry side) multiple rooms or open spaces into a zone or larger room. Air walls are heat transparent - they don't receive solar radiation or internal radiant heat or participate in the long-wave radiation exchange. One tricky thing is to model the free-form air flow across the air wall. It would be good to provide some guidance (on using zone mixer to model the air exchange). There are places to skip the air walls, e.g., calculation of surface view factors need to skip the air walls.

design/FY2019/Interzone Air Wall NFP and Design.md
```diff
@@ -0,0 +1,444 @@
...
+    bool TypeIsAirBoundaryLumpedAirMass(false) // true for Construction:AirBoundary with grouped zones for air exchange
+    bool TypeIsAirBoundaryMixing(false) // true for Construction:AirBoundary with SimpleMixing for air exchange
+
+For the *IRTSurface* option, use existing `TypeIsIRT` as a guide, but there will be some differences for `TypeIsAirBoundaryIRTSurface`.
+For the *InteriorWindow* option, will likely use existing `TypeIsWindow` in combination with `TypeIsAirBoundaryInteriorWindow`.
+
+*I must say I'm not sure I want to add to the long list of `TypeIs...` variables in `ConstructionData`, but they get the job done.
+Several are used only for error checking, some are used only for inits, others are used in the simulation loops (primarily to skip a surface with `continue`. 
+In fact, the entire `ConstructionData` struct is full of stuff that's only used for one type of construction or another. There's some common stuff,
+but lots of extra baggage. I'm open to suggestions.*
```
Myoldmopar, line 336: For the specific TypeIs, wouldn't a few enums work instead of all bool flags?  Although, booleans are very efficient, so maybe leaving them as-is would be reasonable.  In the bigger picture, you could create a base class with all the common stuff and create little derived classes for each construction type.

Myoldmopar (commented):
This is well thought out.  While I have reservations about some of the aspects, I do see the gains that this can make and will add a capability that is requested frequently.  I'm not going to mark this approved right now just approving on the design work.

design/FY2019/Interzone Air Wall NFP and Design.md
mjwitte: Part of the problem with enums here is that these various bools are not all "peers"? of each other. They cover different aspects, some overlap and are mutually exclusive, some encompass a group of others, and some are just loners. So, all of that points to the opportunity for a better structure. I'll think about ways to improve this.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
```diff
@@ -1706,4 +1706,85 @@ TEST_F(EnergyPlusFixture, HeatBalanceManager_HeatBalanceAlgorithm_HAMT)
     EXPECT_EQ(DataHeatBalance::OverallHeatTransferSolutionAlgo, DataSurfaces::HeatTransferModel_HAMT);
 }
 
+TEST_F(EnergyPlusFixture, HeatBalanceManager_GetAirBoundaryConstructData)
```
mjwitte, line 1709: Almost there for [HeatBalanceManager::CreateAirBoundaryConstructions](http://energyplus.s3-website-us-east-1.amazonaws.com/coverage/2019-08-01-EnergyPlus-8229db616d-x86_64-gcc-7.4UTCD/EnergyPlus/HeatBalanceManager.cc.gcov.html) need another variation or two to cover the last few lines.

tst/EnergyPlus/unit/HeatBalanceIntRadExchange.unit.cc
```diff
@@ -216,4 +221,91 @@ TEST_F(EnergyPlusFixture, HeatBalanceIntRadExchange_UpdateMovableInsulationFlagT
     EXPECT_TRUE(!DidMIChange);
 }
 
+TEST_F(EnergyPlusFixture, HeatBalanceIntRadExchange_AlignInputViewFactorsTest)
```
mjwitte, line 224: Note the **COMPLETE** coverage for [HeatBalanceIntRadExchange::AlignInputViewFactors](http://energyplus.s3-website-us-east-1.amazonaws.com/coverage/2019-08-01-EnergyPlus-8229db616d-x86_64-gcc-7.4UTCD/EnergyPlus/HeatBalanceIntRadExchange.cc.gcov.html).

mjwitte:
Closing this for now - if #7455 fails dismally, will reopen this one.

PR #7295, opened 2019-05-09
Correct physics based VRF model to align inlet and outlet air flow rates

### LINKED ISSUE #6301: VRF Terminal unit flow balance problems

rraustad:
Pull request overview
---------------------
The physics based VRF model calculates the operating evaporator and condenser temperatures after all terminal units have been simulated. The method used adjusted the terminal unit simulated inlet node air flow rate such that it did not match the TU outlet node air flow rate. This change fixes #6301.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/doc/engineering-reference/src/simulation-models-encyclopedic-reference-002/variable-refrigerant-flow-heat-pumps.tex
+++ b/doc/engineering-reference/src/simulation-models-encyclopedic-reference-002/variable-refrigerant-flow-heat-pumps.tex
@@ -895,7 +895,7 @@ These features enhance the simulation accuracy of the VRF system performance in 
 
 Note that a number of calculation steps are coupled together in the VRF-FluidTCtrl model, for instance, the piping loss calculation and the system performance calculation. More specifically, the piping loss changes the operating conditions of the system, which may lead to a different control strategy and thus in reverse affect the amount of piping loss. This makes it difficult to obtain an analytical solution for a number of operational parameters (e.g., enthalpy of refrigerant entering the indoor unit), and therefore numerical iterations are employed to address this problem (refer to Figure VRF-FluidTCtrl-3 for more details). Therefore, the VRF-FluidTCtrl model can have a longer execution time to perform the simulation than the VRF-SysCurve model. 
 
-The object connections for VRF-FluidTCtrl model is similar to those for VRF-SysCurve model. The difference lies in the object used to describe a specific components. For example, VRF-SysCurve model uses \emph{AirConditioner:VariableRefrigerantFlow} object to describe the VRF outdoor unit performance, while in VRF-FluidTCtrl model the \emph{AirConditioner:VariableRefrigerantFlow} object is used.
+The object connections for VRF-FluidTCtrl model is similar to those for VRF-SysCurve model. The difference lies in the object used to describe a specific components. For example, VRF-SysCurve model uses \emph{AirConditioner:VariableRefrigerantFlow} object to describe the VRF outdoor unit performance, while in VRF-FluidTCtrl model the \emph{AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl} object is used.
 
 \begin{figure}[hbtp] % figure VRF-FluidTCtrl-1b
 \centering
--- a/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
+++ b/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
@@ -8543,7 +8543,7 @@ namespace HVACVariableRefrigerantFlow {
         }
     }
 
-    void VRFTerminalUnitEquipment::CalcVRFIUVariableTeTc(int const VRFTUNum, // Index to VRF terminal unit
+    void VRFTerminalUnitEquipment::CalcVRFIUVariableTeTc(int const EP_UNUSED(VRFTUNum), // Index to VRF terminal unit
                                                          Real64 &EvapTemp,   // evaporating temperature
                                                          Real64 &CondTemp    // condensing temperature
     )
@@ -8590,7 +8590,7 @@ namespace HVACVariableRefrigerantFlow {
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
         int const Mode(1);           // Performance mode for MultiMode DX coil. Always 1 for other coil types
         int CoolCoilNum;             // index to the VRF Cooling DX coil to be simulated
-        int FanOutletNode;           // index to the outlet node of the fan
+        //int FanOutletNode;           // index to the outlet node of the fan
         int HeatCoilNum;             // index to the VRF Heating DX coil to be simulated
         int OAMixerNum;              // OA mixer index
         int OAMixNode;               // index to the mix node of OA mixer
@@ -8614,12 +8614,12 @@ namespace HVACVariableRefrigerantFlow {
         Real64 EvapTempMin;          // Min evaporating temperature, correspond to the maximum cooling capacity (C)
         Real64 Garate;               // Nominal air mass flow rate
         Real64 H_coil_in;            // Air enthalpy at the coil inlet (kJ/kg)
-        Real64 QZnReqSenCoolingLoad; // Zone required sensible cooling load (W)
-        Real64 QZnReqSenHeatingLoad; // Zone required sensible heating load (W)
+        //Real64 QZnReqSenCoolingLoad; // Zone required sensible cooling load (W)
+        //Real64 QZnReqSenHeatingLoad; // Zone required sensible heating load (W)
         Real64 RHsat;                // Relative humidity of the air at saturated condition(-)
         Real64 SH;                   // Super heating degrees (C)
         Real64 SC;                   // Subcooling degrees (C)
-        Real64 temp;                 // for temporary use
+        //Real64 temp;                 // for temporary use
         Real64 T_coil_in;            // Temperature of the air at the coil inlet, after absorbing the heat released by fan (C)
         Real64 T_TU_in;              // Air temperature at the indoor unit inlet (C)
         Real64 Tout;                 // Air temperature at the indoor unit outlet (C)
@@ -8703,8 +8703,14 @@ namespace HVACVariableRefrigerantFlow {
         if ((Garate > 0.0) && ((!VRF(VRFNum).HeatRecoveryUsed && CoolingLoad(VRFNum)) ||
                                (VRF(VRFNum).HeatRecoveryUsed && TerminalUnitList(TUListIndex).HRCoolRequest(IndexToTUInTUList)))) {
             // 1.1) Cooling coil is running
+            // The TU has already been simulated so OutputRequiredToCoolingSP is load remaining, not zone load
             //QZnReqSenCoolingLoad = max(0.0, -1.0 * ZoneSysEnergyDemand(ZoneIndex).OutputRequiredToCoolingSP);
+            // What Tout is required here? Average over time step? Full outlet temp when coil is on?
+            // This equation would yield a false Tout if actual air flow is different from full air flow rate.
             //Tout = T_TU_in - QZnReqSenCoolingLoad * 1.2 / Garate / 1005;
+            // Tout for constant fan mode is the actual Tout over the time step and coil is on for entire time step
+            // Tout for cycling fan would be the full outlet temp while the coil operates at PLR
+            // so what Tout should be used to get true impact on compressor operation (e.g., PLR*Tout + 1-PLR*T_coil_in)?
             Tout = DXCoils::DXCoilOutletTemp(CoolCoilNum);
             Th2 = T_coil_in - (T_coil_in - Tout) / (1 - BFC);
             DeltaT = C3Tevap * SH * SH + C2Tevap * SH + C1Tevap;
--- a/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
+++ b/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
@@ -8602,7 +8602,6 @@ namespace HVACVariableRefrigerantFlow {
         int const Mode(1);           // Performance mode for MultiMode DX coil. Always 1 for other coil types
         int CoolCoilNum;             // index to the VRF Cooling DX coil to be simulated
         int HeatCoilNum;             // index to the VRF Heating DX coil to be simulated
-        int OAMixerNum;              // OA mixer index
         int OAMixNode;               // index to the mix node of OA mixer
         int IndexToTUInTUList;       // index to TU in specific list for the VRF system
         int TUListIndex;             // index to TU list for this VRF system
@@ -10651,7 +10650,6 @@ namespace HVACVariableRefrigerantFlow {
         // FUNCTION LOCAL VARIABLE DECLARATIONS:
         int const Mode(1);       // Performance mode for MultiMode DX coil. Always 1 for other coil types
         int CoilIndex;           // index to coil
-        int OAMixerNum;          // OA mixer index
         int OAMixNode;           // index to the mix node of OA mixer
         int VRFCond;             // index to VRF condenser
         int VRFTUNum;            // Unit index in VRF terminal unit array
--- a/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
+++ b/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
@@ -272,7 +272,7 @@ namespace HVACVariableRefrigerantFlow {
         // SUBROUTINE INFORMATION:
         // AUTHOR         Richard Raustad, FSEC
         // DATE WRITTEN   August 2010
-        // MODIFIED       Jul 2015, RP Zhang (LBNL), XF Pang (LBNL), Y Yura (Daikin Inc). Add a physics-based VRF model appliable for Fluid
+        // MODIFIED       Jul 2015, RP Zhang (LBNL), XF Pang (LBNL), Y Yura (Daikin Inc). Add a physics-based VRF model applicable for Fluid
         // Temperature Control RE-ENGINEERED  na
 
         // PURPOSE OF THIS SUBROUTINE:
@@ -371,7 +371,7 @@ namespace HVACVariableRefrigerantFlow {
         if (all(TerminalUnitList(TUListNum).IsSimulated)) {
 
             if (VRF(VRFCondenser).VRFAlgorithmTypeNum == AlgorithmTypeFluidTCtrl) {
-                // Algorithm Type: VRF model based on physics, appliable for Fluid Temperature Control
+                // Algorithm Type: VRF model based on physics, applicable for Fluid Temperature Control
                 VRF(VRFCondenser).CalcVRFCondenser_FluidTCtrl();
             } else {
                 // Algorithm Type: VRF model based on system curve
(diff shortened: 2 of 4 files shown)


src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8663,34 +8663,35 @@ namespace HVACVariableRefrigerantFlow {
         }
 
         // Set inlet air mass flow rate based on PLR and compressor on/off air flow rates
-        SetAverageAirFlow(VRFTUNum, 1.0, temp);
+        //SetAverageAirFlow(VRFTUNum, 1.0, temp);
```
rraustad: The TUs have already been simulated. At this point it appears the model want to know what the evap T and cond T are for use during the next time step. Anything that tried to resimulate the TUs is now commented out.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8663,34 +8663,35 @@ namespace HVACVariableRefrigerantFlow {
...
         VRFInletNode = this->VRFTUInletNodeNum;
         T_TU_in = Node(VRFInletNode).Temp;
         W_TU_in = Node(VRFInletNode).HumRat;
-        T_coil_in = T_TU_in;
-        W_coil_in = W_TU_in;
 
         // Simulation the OAMixer if there is any
         if (this->OAMixerUsed) {
-            SimOAMixer(this->OAMixerName, false, this->OAMixerIndex);
+            //SimOAMixer(this->OAMixerName, false, this->OAMixerIndex);
```
rraustad: Mixer has already been simulated.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8663,34 +8663,35 @@ namespace HVACVariableRefrigerantFlow {
...
-                HVACFan::fanObjs[this->FanIndex]->simulate(1.0 / temp, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
-                FanOutletNode = HVACFan::fanObjs[this->FanIndex]->outletNodeNum;
-            } else {
-                Fans::SimulateFanComponents("", false, this->FanIndex, FanSpeedRatio, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
-                FanOutletNode = Fans::Fan(this->FanIndex).OutletNodeNum;
-            }
-            T_coil_in = Node(FanOutletNode).Temp;
-            W_coil_in = Node(FanOutletNode).HumRat;
-        }
+        //if (this->FanPlace == BlowThru) {
```
rraustad: Fan has already been simulated. If the intent here is to simulate at full flow then why weren't the coils simulated again?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8702,8 +8703,9 @@ namespace HVACVariableRefrigerantFlow {
         if ((Garate > 0.0) && ((!VRF(VRFNum).HeatRecoveryUsed && CoolingLoad(VRFNum)) ||
                                (VRF(VRFNum).HeatRecoveryUsed && TerminalUnitList(TUListIndex).HRCoolRequest(IndexToTUInTUList)))) {
             // 1.1) Cooling coil is running
-            QZnReqSenCoolingLoad = max(0.0, -1.0 * ZoneSysEnergyDemand(ZoneIndex).OutputRequiredToCoolingSP);
-            Tout = T_TU_in - QZnReqSenCoolingLoad * 1.2 / Garate / 1005;
+            //QZnReqSenCoolingLoad = max(0.0, -1.0 * ZoneSysEnergyDemand(ZoneIndex).OutputRequiredToCoolingSP);
```
rraustad: Here's the big question. What Tout is needed here? The Tout when the coil is operating I assume. Line 8706 is wrong since for that zone, the TU operated and probably met the load, so OutputRequiredToCoolingSP is probably 0. Instead I use the actual coil outlet temp, but this is premised on needing actual Tout.

rraustad: I guess I should read the Eng. Ref. to get better acquainted.
@nealkruis @mbadams5 @mjwitte @Myoldmopar @jasondegraw @kbenne this will need some serious thought and review because I don't know this model. What the heck should it do? Does it need full output conditions while coils are operating? Yet the coils can have different PLR's so how to average correctly to get condenser performance? Anyone recognize old line 8706 above? Q*1.2/Mdot/1005 ? Figuring out what this function needs will probably correct this issue. If Tout at full air flow is really needed then the coils will need to be simulated again, or better yet that information needs to be saved when the TUs model full flow and then used here. Is Xiufeng Pang or RP Zhang or anyone at LBNL available for consult on this model?

rraustad:
Since the answers didn't line up well with develop I decided to mimic what the original model does. The answers don't line up perfectly because there was a mistake in the original code. I also don't understand why the model assumes full flow rate when determining suction/discharge temperature at the condenser when actual TU flow rates used are known and coil outlet temps are also known. Regardless, these changes mimic the original model.

Air flow rate is now consistent and zone temperature is roughly unchanged. Develop is left column, this branch is right column. In this first figure since zone temperature is held to set point, the TU outlet node flow rate was correct so in the right column, the new flows should roughly match the blue line in the left figure.

image

The compressor speed and cooling coil rate are also roughly the same.

image

The evap/condenser temps used are also similar but won't match because of the problem in code. Where full flow was set, the developer forgot to set the OA flow rate and therefore when SetAverageAirFlow was called it used the OA flow rate from the last TU simulated and wound up with a different mixed air temp.

image

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -7014,9 +7017,19 @@ namespace HVACVariableRefrigerantFlow {
         Real64 PartLoadRatio(1.0);
 
         if (VRF(VRFTU(VRFTUNum).VRFSysNum).VRFAlgorithmTypeNum == AlgorithmTypeFluidTCtrl) {
-            // Algorithm Type: VRF model based on physics, appliable for Fluid Temperature Control
+            // Algorithm Type: VRF model based on physics, applicable for Fluid Temperature Control
             VRFTU(VRFTUNum).ControlVRF_FluidTCtrl(VRFTUNum, QZnReq, FirstHVACIteration, PartLoadRatio, OnOffAirFlowRatio);
             VRFTU(VRFTUNum).CalcVRF_FluidTCtrl(VRFTUNum, FirstHVACIteration, PartLoadRatio, SysOutputProvided, OnOffAirFlowRatio, LatOutputProvided);
+            if ( PartLoadRatio == 0.0 ) { // set coil inlet conditions when coil does not operate. Inlet conditions are set in ControlVRF_FluidTCtrl when PLR=1
```
rraustad: Pick up temps used to calculate evap/cond temps later on. Here only for PLR=0.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: Note here that the original model did not set OA flow rate and therefore used the OA flow for the last TU simulated when SetAverageAirFlow was called. When I corrected this (lines 8671 and 8676), my new method matched the old method results.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -10046,23 +10064,23 @@ namespace HVACVariableRefrigerantFlow {
                 LimitTUCapacity(VRFCond,
                                 NumTUInList,
                                 TotalTUCoolingCapacity,
-                                TerminalUnitList(TUListNum).TotalCoolLoad,
-                                MaxCoolingCapacity(VRFCond),
-                                TotalTUHeatingCapacity,
-                                TerminalUnitList(TUListNum).TotalHeatLoad,
-                                MaxHeatingCapacity(VRFCond));
+TerminalUnitList(TUListNum).TotalCoolLoad,
```
rraustad: Oops, will fix this on next commit.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -10186,6 +10202,14 @@ namespace HVACVariableRefrigerantFlow {
         // Otherwise the coil needs to turn on. Get full load result
         PartLoadRatio = 1.0;
         this->CalcVRF_FluidTCtrl(VRFTUNum, FirstHVACIteration, PartLoadRatio, FullOutput, OnOffAirFlowRatio);
+        this->fanOutletT = Node(this->fanOutletNode).Temp;
```
rraustad: Here, when the TU operates the conditions resulting from PLR=1 (full flow to mimic what was done before) are captured and used in calculations for evap/cond temps.

src/EnergyPlus/HVACVariableRefrigerantFlow.hh
```diff
@@ -700,6 +700,10 @@ namespace HVACVariableRefrigerantFlow {
         int ATMixerSecNode;              // secondary air inlet node number for the air terminal mixer
         int ATMixerOutNode;              // outlet air node number for the air terminal mixer
         bool firstPass;                  // used to reset global sizing data
+        Real64 coilInNodeT;              // coil inlet node temp at full flow (C)
```
rraustad, line 703: new variables to hold data in order to eliminate the problem with flow mismatch from inlet to outlet.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: Also note that any calls to this and other functions updates the node flow rates. So data set when the TU was actually modelled is used here instead.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -3287,6 +3287,8 @@ namespace HVACVariableRefrigerantFlow {
                     if (errFlag) {
                         ShowContinueError("...occurs in " + cCurrentModuleObject + " = " + VRFTU(VRFTUNum).Name);
                         ErrorsFound = true;
+                    } else {
+                        VRFTU(VRFTUNum).fanOutletNode = Fans::Fan(VRFTU(VRFTUNum).FanIndex).OutletNodeNum;
```
rraustad, line 3291: Save fan outlet node for use in TeTc calculations.

rraustad:
I should probably show the marked difference between this branch and develop. Since the original model tried to use full flow in the TeTc calc's, and forgot to also set OACompOnMassFlow along with CompOnMassFlow, the resulting TeTc is now different. This impacts the air flow rate and will show diff's. Other than this, right or wrong the model still does what it originally did.

image

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8651,46 +8660,11 @@ namespace HVACVariableRefrigerantFlow {
...
-                FanOutletNode = HVACFan::fanObjs[this->FanIndex]->outletNodeNum;
-            } else {
-                Fans::SimulateFanComponents("", false, this->FanIndex, FanSpeedRatio, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
-                FanOutletNode = Fans::Fan(this->FanIndex).OutletNodeNum;
-            }
-            T_coil_in = Node(FanOutletNode).Temp;
-            W_coil_in = Node(FanOutletNode).HumRat;
-        }
+        T_coil_in = this->coilInNodeT;
+        W_coil_in = this->coilInNodeW;
```
rraustad, line 8662: Change here is that coil inlet temps are now used so it doesn't matter if blow thru fan or OA mixer is used. Gets rid of a lot of unnecessary code.

rraustad:
The 2 new diff's are from the merge of #7258 into develop. No need to spin off CI again just for that.

Myoldmopar:
But there are build warnings.  It looks like just one unused "using" statement (which is interpreted as an unused variable in a function).  The error message is difficult to parse, so the CI message doesn't always point you to the right spot easily, but I can track it down for you if you need help diagnosing where it is.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-002/variable-refrigerant-flow-heat-pumps.tex
```diff
@@ -895,7 +895,7 @@ \subsubsection{Overview}\label{VRF-FluidTCtrl-HP-overview}
...
 
-The object connections for VRF-FluidTCtrl model is similar to those for VRF-SysCurve model. The difference lies in the object used to describe a specific components. For example, VRF-SysCurve model uses \emph{AirConditioner:VariableRefrigerantFlow} object to describe the VRF outdoor unit performance, while in VRF-FluidTCtrl model the \emph{AirConditioner:VariableRefrigerantFlow} object is used.
+The object connections for VRF-FluidTCtrl model is similar to those for VRF-SysCurve model. The difference lies in the object used to describe a specific components. For example, VRF-SysCurve model uses \emph{AirConditioner:VariableRefrigerantFlow} object to describe the VRF outdoor unit performance, while in VRF-FluidTCtrl model the \emph{AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl} object is used.
```
Myoldmopar, line 898: Good find.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Myoldmopar: Seems reasonable enough, if you are going to have to query what the fan is doing, you can check the outlet node.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8526,7 +8537,7 @@ namespace HVACVariableRefrigerantFlow {
             for (int i = 1; i <= TerminalUnitList(TUListNum).NumTUInList; i++) {
                 int VRFTUNum = TerminalUnitList(TUListNum).ZoneTUPtr(i);
                 // analyze the conditions of each IU
-                VRFTU(VRFTUNum).CalcVRFIUVariableTeTc(VRFTUNum, EvapTemp(i), CondTemp(i));
+                VRFTU(VRFTUNum).CalcVRFIUVariableTeTc(EvapTemp(i), CondTemp(i));
```
Myoldmopar, line 8540: :+1: :+1: :+1: :+1:

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Myoldmopar: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1:

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -2440,16 +2440,143 @@ TEST_F(EnergyPlusFixture, VRF_FluidTCtrl_CalcVRFIUTeTc)
...
+    VRFTU(1).coilInNodeW = Node(3).HumRat;
+    VRFTU(2).coilInNodeT = Node(3).Temp;
+    VRFTU(2).coilInNodeW = Node(3).HumRat;
+    // system is on in heating mode
+    VRF(IndexVRFCondenser).CalcVRFIUTeTc_FluidTCtrl();
+    // default value, coil inlet temps higher than default
+    EXPECT_EQ(VRF(IndexVRFCondenser).IUEvaporatingTemp, 15);
+    // fan heat added to coil inlet temp, coil surface temp should also be lower to account for less heat tranfer
+    EXPECT_LT(VRF(IndexVRFCondenser).IUCondensingTemp, saveHeatingCondTemp);
+
```
Myoldmopar, line 2579: Great unit test.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -2440,16 +2440,143 @@ TEST_F(EnergyPlusFixture, VRF_FluidTCtrl_CalcVRFIUTeTc)
...
     // Clean up
     VRF.deallocate();
```
Myoldmopar, line 2581: Ideally these are cleaned up as part of the clear_state after every single unit test is run, but I understand it doesn't hurt to have them here.

Myoldmopar (approved):

Myoldmopar:
Alright, I fixed that unused variable issue, merged up from develop, and will let this run over the next couple hours.  I'm anticipating very clean CI except for the actual expected diff file.  If so I'll merge in the morning.

rraustad:
This:

    // Clean up
    VRF.deallocate();

should have been removed. I was so focused on testing this that I missed this. We should not allow deallocate in unit tests? New python script warranted?

rraustad:
And this:

image

made me think it was something other than what I worked on (why would I think that if this warning never showed up before?). Thanks for catching that.

Myoldmopar:
> New python script

I was already half asleep and this phrase woke me right back up.  I like your thinking but I would worry that in this particular case, it may hit too many false alarms.  There are certainly times where you want to deallocate/reallocate within a unit test, and I feel like we may end up hitting too many of those cases to warrant a script.  But your line of thinking here makes me smile.

Myoldmopar:
Yeah, the full warning message text is:

```
$ make HVACVariableRefrigerantFlow.cc.o
Building CXX object src/EnergyPlus/CMakeFiles/energypluslib.dir/HVACVariableRefrigerantFlow.cc.o
In file included from /eplus/repos/4eplus/src/EnergyPlus/HVACVariableRefrigerantFlow.cc:86:
/eplus/repos/4eplus/src/EnergyPlus/MixedAir.hh: In member function ‘void EnergyPlus::HVACVariableRefrigerantFlow::VRFTerminalUnitEquipment::CalcVRFIUVariableTeTc(Real64&, Real64&)’:
/eplus/repos/4eplus/src/EnergyPlus/MixedAir.hh:414:34: warning: unused variable ‘EnergyPlus::MixedAir::OAMixer’ [-Wunused-variable]
     extern Array1D<OAMixerProps> OAMixer;
                                  ^~~~~~~
```

Which makes it pretty clear in the third or so line that the error occurs in `In member function ‘void EnergyPlus::HVACVariableRefrigerantFlow::VRFTerminalUnitEquipment::CalcVRFIUVariableTeTc`.  However, Decent CI has trouble with parsing messages that run this long and detecting which part of it is the actual useful error content.  It's easy enough for me to build the code and find it locally.

rraustad:
Shoot, now I have to think about a valid deallocate in a unit test...

Myoldmopar:
> Shoot, now I have to think about a valid deallocate in a unit test...

Yeah, maybe there's not one, I may be concerned at that possibility for nothing.  For better or worse (worse) there definitely _are_ a ton of deallocate calls in tst/EnergyPlus/unit though.

rraustad:
"definitely are a ton of deallocate calls in tst/EnergyPlus/unit"

Of course there are, we are old and lazy!!!

Myoldmopar:
Everything looks great.  Merging.

Issue #6817, opened 2018-07-10
CeilingDiffuser inside convection model sets ACH to zero during sizing


mjwitte:
Issue overview
--------------
User noted a very large difference between zone sizing peak load and the loads seen during a sizing period simulation with ideal loads. These *should* be very close to the same value. With the default TARP inside convection, they are, but with CeilingDiffuser convection they differ significantly, on the order of 30%-50%. Code inspection reveals that the CeilingDiffuser models all [set the ACH to zero during sizing](https://github.com/NREL/EnergyPlus/blob/8a7f7d6f776766877ad6790104bde49a6bc29894/src/EnergyPlus/ConvectionCoefficients.cc#L4353-L4355). Since zone sizing is calculating air mass flow rates based on the load and supply temperature, this flow rate should be available to the convection coefficient calculation.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v8.9.0
 - Helpdesk ticket 12897

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [x] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte The defect file that you added has ideal loads air systems in it.  How did you generate it and to what does it compare?  In other words, if I run the file that you added, which other input file will have the 30-50% differences that you mentioned in the description?

mjwitte:
The defect file is the HVACTemplate-5ZonePurchAir example file modified to use `SurfaceConvectionAlgorithm:Inside,CeilingDiffuser`. When this file is run for sizing periods (with Chicago weather), compared the zone sizing calculated design load with the peak ????? from the csv output.  For example, for Space1-1
```
! <Zone Sizing Information>, Zone Name, Load Type, Calc Des Load {W}
SPACE1-1, Cooling, 3434.83414
```
From the csv output, the max of  SPACE1-1   IDEAL LOADS AIR SYSTEM:Zone Ideal Loads Zone Sensible Cooling Rate   [W](Hourly) is 4841.8.

The same file using TARP inside convection results in 3435W Calc Des Cooling Load and 3438W peak ideal loads sensible cooling.

RKStrand:
[...]

PR #7300, opened 2019-05-17
Correction of ACH Set to Zero for CeilingDiffuser Convection Model

RKStrand:
Pull request overview
---------------------
Addresses #6817.

In the CeilingDiffuser option for interior convection, the ACH value which is critical to the calculation of convection coefficients in this particular method was being set to zero during zone and system sizing.  This defaults the convection coefficient calculations back to the TARP Natural Convection model which results in significantly different convection coefficients.  This resulted in sizing calculations for peak load to be between 30-50% different than the peak load during a design day run.  These two values for the same convection model should be almost identical.  When the TARP model is used for both sizing and the design day simulation, the results are almost identical.  This issue was noticed by a user and this CR work corrects this problem.  ACH is no longer set to zero for sizing calculations and the peak load and the sized load are now nearly identical as would be expected.  As per CR protocol, a unit test was added to test this fix.  In the test suite, the only file that shows a difference is a file that uses the CeilingDiffuser model and the differences are expected due to this fix.

### Work Checklist
   - Defect: This pull request repairs a github defect issue 6817

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


mjwitte (approved):
Merged in develop, ran defect file before/after. With this change, the peak sensible cooling load during the design day simulation is within 1% of the design cooling load reported in the zone sizing section of the eio output. The design cooling loads have change by as much as 50% in Space 2-1, so this will have a significant impact on sizing for simulations using the ceiling diffuser algorithm. Unit tests all run ok. Will wait for one more CI round.

mjwitte:
Also, changes in _SmallOfficeDulles look good. Warmup convergence errors on the cooling design day are gone now and sizing results are different as expected. Otherwise, the annual simulation is quite similar, with a small decrease in hours cooling setpoint not met, and a roughly 5% increase in electricity consumption.

Issue #6743, opened 2018-05-13
Bug report for Unused Schedules


mldichter:
Using Energyplus 8.9 and
Output:Diagnostics,
    DisplayUnusedSchedules

for linked idf file, the schedules
AIR VELOCITY SCHEDULE
CLOTHING SCHEDULE
WORK EFFICIENCY SCHEDULE

are listed as unused, but they are referenced by the People object.
People,
    Warehouse Office People, !- Name
    Warehouse Office,        !- Zone or ZoneList Name
    Warehouse BLDG_OCC_SCH,  !- Number of People Schedule Name
    People/Area,             !- Number of People Calculation Method
    ,                        !- Number of People
    0.0210972644167511,      !- People per Zone Floor Area {person/m2}
    ,                        !- Zone Floor Area per Person {m2/person}
    0.3,                     !- Fraction Radiant
    ,                        !- Sensible Heat Fraction
    Warehouse Office Activity Schedule,  !- Activity Level Schedule Name
    ,                        !- Carbon Dioxide Generation Rate {m3/s-W}
    ,                        !- Enable ASHRAE 55 Comfort Warnings
    ,                        !- Mean Radiant Temperature Calculation Type
    ,                        !- Surface Name/Angle Factor List Name
    Work Efficiency Schedule,!- Work Efficiency Schedule Name
    ,                        !- Clothing Insulation Calculation Method
    ,                        !- Clothing Insulation Calculation Method Schedule Name
    Clothing Schedule,       !- Clothing Insulation Schedule Name
    Air Velocity Schedule;   !- Air Velocity Schedule Name

And deleting those three schedules from the idf file results in a severe error.

idf file google drive link
https://drive.google.com/file/d/14ofPMDyjMqCMG7NG3LIofWQiwJAzpL5O/view?usp=sharing

err file google drive link
https://drive.google.com/file/d/1tAXULNflC-yfvLuX_yANoVMoT_E21kRN/view?usp=sharing

JasonGlazer:
Originally reported:

https://unmethours.com/question/31458/bug-report-for-unused-schedules/

RKStrand:
[...]

PR #7301, opened 2019-05-17
Additional Warnings for Thermal Comfort Schedules

RKStrand:
Pull request overview
---------------------
A user noted that some unused schedules were being reported for People objects relating to thermal comfort models.  In the user input file, the user had specified schedules for air velocity, clothing insulation, and work efficiency, but the ERR file noted that these schedules were NOT being used.  Investigation of the issue uncovered that the user had not selected a thermal comfort model (Fanger, Pierce, or KSU) and in fact the schedules were not being used as a result.  However, it wasn't clear to the user why the schedules were not being used.  In fact, when the user deleted definitions of the schedules that were supposedly unused from the input file, the user got an error message that those schedules were missing.  Testing showed that if a thermal comfort model was added to the People object that everything was fine and the unused schedule errors went away.  The solution included testing for the case where the schedules were included in the user's People input statement but no thermal comfort model was selected.  In those cases, an additional warning was added noting the problem and how to solve it.  A unit test was added to check a new subroutine added as part of the solution to this problem.  In addition, the I/O Reference was updated to note this potential issue (noting that the user must select a thermal comfort model if they add schedules for these items) and one of the IDF examples in the I/O Reference was corrected because it was missing some recent things that were added to the People syntax.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #6743 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer (commented):
Everything works as expected.

JasonGlazer:
@mjwitte @Myoldmopar Everything looks fine for the pull request. Should I merge this or do one of you want to review this further?

Myoldmopar:
Merge at will @JasonGlazer

PR #7305, opened 2019-05-22
Incorrect EvaporativeCooler:Direct:ResearchSpecial water pump power modifier curve use

### LINKED ISSUE #7118: Cannot Produce I/O doc statement on EvaporativeCooler:Direct:ResearchSpecial Water Pump Modifier Curve
Issue overview
--------------
I/O documentation section 1.50.2.1.8 Field: Water Pump Power Modifier Curve Name states that:

This alpha input field is the name of a dimensionless normalized pump power modifying curve. This curve modifies the pump electric power in the previous field by multiplying the design power by the result of this curve. The normalized curve is a function of the primary air flow fraction as independent variable. The curve shall yield a value of 1.0 at a flow fraction of 1.0. The flow fraction is the ratio of the primary air during current operation divided by primary air Design Air Flow Rate. If this input field is left blank, the pump power is assumed to linearly vary with the load. Any curve or table with one independent variable can be used: Curve:Linear, Curve:Quadratic, Curve:Cubic, Curve:Quartic, Curve:Exponent, Curve:ExponentialSkewNormal, Curve:Sigmoid, Curve:RectuangularHyperbola1, Curve:RectangularHyperbola2, Curve:ExponentialDecay, Curve:DoubleExponentialDecay, and Table:OneIndependentVariable.

Within above lines, I can see "If this input field is left blank, the pump power is assumed to linearly vary with the load."

I've tested a simple curve that has a linear change of Y against X with X ranges [0, 1], but eso files show different results especially from "Evaporative Cooler Electric Power" variable.

Changing the curve to a constant one (always 1) does not solve the problem (this test was based on the suspect that there might be a typo/mistake in the documentation).

So now I have a concern that the source codes don't treat this curve object well. As a contrast, the Indirect ResearchSpecial operative cooler has been tested well between using and not using curves generated by the guidance defined in the I/O documentation, where for Water Pump Power Modifier Curve Name field it says "If this input field is left blank, the pump power is assumed to be constant."

### Details
Some additional details:
 - Tested on Windows 7 64-bit computer.
 - Using EnergyPlus V8.9

Nigusse:
Pull request overview
---------------------
This resolves #7118. The direct evaporative cooler research special water pump modifier curve was using incorrect flow rate unless "Effectiveness Flow Ratio Modifier Curve Name" input field is not blank. This fix modified the logic such that pump power uses the correct flow ratio regardless of the Effectiveness modifier curve presence. This is a defect. Diffs are expected/justified if the "Effectiveness Flow Ratio Modifier Curve Name" input field is blank.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -3270,16 +3270,17 @@ namespace EvaporativeCoolers {
...
+            MassFlowRateSys = EvapCond(EvapCoolNum).InletMassFlowRate;
+            MassFlowRateSysDesign = Node(InletNode).MassFlowRateMax;
+            if (MassFlowRateSysDesign > 0.0) {
+                if (MassFlowRateSys > 0.0) {
+                    FlowRatio = MassFlowRateSys / MassFlowRateSysDesign;
+                } else {
+                    FlowRatio = 1.0;
                 }
+            }
+            if (EvapCond(EvapCoolNum).WetbulbEffecCurveIndex > 0) {
```
Myoldmopar, line 3315: I see, so the mass flow stuff should've just always been outside the curve index IF block.  You are still checking for divide by zero, but I don't see any handling of the MassFlowRateSysDesign == 0.0.  I realize this would mean things are going badly, but I want to make sure this case is at least gracefully handled and not going to divide by zero somewhere else down the line.

Myoldmopar: OK, I checked it out.  FlowRatio is initialized properly to zero, and is used as a curve input variable below.  This is fine.

tst/EnergyPlus/unit/EvaporativeCoolers.unit.cc
```diff
@@ -761,4 +762,56 @@ TEST_F(EnergyPlusFixture, DefaultAutosizeDirEvapCoolerTest)
...
+    EvaporativeCoolers::CalcDirectResearchSpecialEvapCooler(EvapCoolNum);
+    EXPECT_DOUBLE_EQ(200.0, thisEvapCooler.RecircPumpPower);
+    EXPECT_DOUBLE_EQ(200.0, thisEvapCooler.EvapCoolerPower);
+
+    // reduce primary air flow rate by half
+    thisEvapCooler.InletMassFlowRate = 0.5;
+    // check water pump power at half primary air flow
+    EvaporativeCoolers::CalcDirectResearchSpecialEvapCooler(EvapCoolNum);
+    EXPECT_DOUBLE_EQ(200.0, thisEvapCooler.RecircPumpPower);
+    EXPECT_DOUBLE_EQ(100.0, thisEvapCooler.EvapCoolerPower);
```
Myoldmopar, line 814: Great!

Myoldmopar (approved):

Myoldmopar:
Merged develop in, resolved the conflict, and ran the unit test suite.  This is good to go.  Merging it in.

Issue #6707, opened 2018-03-30
Allow evaporative coolers to cycle


aaron-boranian:
Problem: Many residential evaporative cooling technologies (e.g., swamp coolers) are controlled to cycle on and off to meet the load. As stated in the I/O reference, these units cannot cycle, meaning that their pumps run continuously even when the required airflow rate is close to zero.

Solution: Give zone evaporative coolers the capability to cycle like unitary equipment.

Context: See the second paragraph [here](http://bigladdersoftware.com/epx/docs/8-3/input-output-reference/group-zone-forced-air-units.html#zonehvacevaporativecoolerunit)

Migrated from [UserVoice](http://energyplus.uservoice.com/forums/258860/suggestions/9118099) feedback from @nealkruis

mjwitte:
I see this as more of a defect. Other systems such as unit heater were upgraded a while back to do a proper cycling fan (operating just long enough to meet the load in a given timestep).
Also, when operating with variable flow fan, the pump power stays at full power. Seems it should vary or have an option to vary.
See [unmethours 33251](https://unmethours.com/question/33251/why-my-zone-have-heating-load-in-summer)

nmerket:
@mjwitte This looks like it might be important to us in the near term. Is this something on your (or someone's) list to get to soon?

mjwitte:
@nmerket Not that I'm aware of.

mjwitte:
Closed via #7319.

PR #7319, opened 2019-05-29
Allows ZoneHVAC:EvaporativeCooler unit to cycle

Nigusse:
Pull request overview
---------------------
Addresses issue #6707. This fix allows ZoneHVAC:EvaporitveCooler unit to cycle when the cooling load is less than the available capacity and the fan type specified is either Fan:OnOff or Fan:SystemModel. Previously, it assumed continuous fan flow for all fan types. Allowing the unit to cycle avoids over cooling of the zone. 

Diffs are expected/justified based on this change for these fan types.  The unit may cycle when the "Cooler Unit Control Method" specified is either "ZoneTemperatureDeadbandOnOffCycling" or "ZoneCoolingLoadOnOffCycling". Cycling is supported for all five Evaporative Cooler Object Type currently allowed.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - [x] Documentation


Nigusse:
This initial commit is code refactor to simplify code changes that allows ZoneHVAC:EvaporativeCooler to cycle. Diffs are not expected in this commit.

Nigusse:
Diffs are expected in the following three example files: 
- StripMallZoneEvapCooler.idf
- StripMallZoneEvapCoolerAutosized.idf
- Fault_FoulingEvapCooler_StripMallZoneEvapCooler.idf

There are 10 ZoneHVAC Evap Cooler units in each test file. Each evap cooler unit serves a single zone. Evap cooler serving "LGstore2" zone has OnOff fan type hence it is cycling now but not prior to this fix.
Hence large diffs are expected in thermal zone "LGstore2" and evap cooler performance serving it. The diffs observed in the other thermal zones is a secondary effect.

Nigusse:
The two plots below show ZoneHVAC EvapCooler unit cycling operation before and after this fix. After the fix (orange color line) shows the zone air is temperature is not over-cooled and the evaporator cooling rate is also scaled by PLR to match zone cooling load. Where the air temperature is over-cooled (before the fix), the unit PLR is less than  1.0. Where the cooling rate and zone air temperature values after the fix match the values prior to the fix, the PLR is always equal to 1 as expected.


EvapCoolerUnitCycling

Nigusse:
@Myoldmopar Diffs are expected in three example files. Otherwise, this branch is ready for review.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -146,6 +146,16 @@ namespace EvaporativeCoolers {
...
 
+    namespace {
+        // These were static variables within different functions. They were pulled out into the namespace
+        // to facilitate easier unit testing of those functions.
+        // These are purposefully not in the header file as an extern variable. No one outside of this should
+        // use these. They are cleared by clear_state() for use by unit tests, but normal simulations should be unaffected.
+        // This is purposefully in an anonymous namespace so nothing outside this implementation file can use it.
+        bool InitZoneHVACEvapCoolerOneTimeFlag(true);
+        bool InitEvapCoolerMyOneTimeFlag(true);
+    } // namespace
```
Myoldmopar: :+1: :+1:

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -958,12 +968,11 @@ namespace EvaporativeCoolers {
         int OutNode;
         int EvapUnitNum;
         static bool MySetPointCheckFlag(true);
```
Myoldmopar, line 970: Do you need to get rid of these statics at the same time?

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4031,6 +4048,14 @@ namespace EvaporativeCoolers {
...
                                 "Average",
                                 ZoneEvapUnit(UnitLoop).Name);
+            if (ZoneEvapUnit(UnitLoop).ControlSchemeType != ZoneCoolingLoadVariableSpeedFan) {
+                SetupOutputVariable("Zone Evaporative Cooler Unit Part Load Ratio",
+                                    OutputProcessor::Unit::None,
+                                    ZoneEvapUnit(UnitLoop).UnitPartLoadRatio,
+                                    "System",
+                                    "Average",
+                                    ZoneEvapUnit(UnitLoop).Name);
+            }
```
Myoldmopar, line 4058: Can you clarify why this variable isn't relevant to a ZoneCoolingLoadVariableSpeedFan configuration?

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4085,21 +4110,21 @@ namespace EvaporativeCoolers {
...
-        if (MyOneTimeFlag) {
+        if (InitZoneHVACEvapCoolerOneTimeFlag) {
             MySizeFlag.dimension(NumZoneEvapUnits, true);
             MyEnvrnFlag.dimension(NumZoneEvapUnits, true);
             MyFanFlag.dimension(NumZoneEvapUnits, true);
             MyZoneEqFlag.allocate(NumZoneEvapUnits);
             MyZoneEqFlag = true;
-            MyOneTimeFlag = false;
+            InitZoneHVACEvapCoolerOneTimeFlag = false;
+            ZoneEquipmentListChecked = false;
```
Myoldmopar: I know you didn't add any arrays here, but this is another place where the Evap structure could just have booleans:

```
struct Evap {
  bool mySize = true;
  bool myEnvrn = true;
  bool myFan = true;
  bool myZoneEq = true;
}
```

Then we don't have a bunch of extra static arrays to manage.  Nothing for you to necessarily do here just saying it is a welcomed change whenever you are working on code like this.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
+                    if (ZoneEvapUnit(UnitNum).OpMode == DataHVACGlobals::ContFanCycCoil) {
+                        PartLoadRatio = 1.0;
+                        ZoneEvapUnit(UnitNum).UnitPartLoadRatio = PartLoadRatio;
+                        CalcZoneEvapUnitOutput(UnitNum, PartLoadRatio, SensibleOutputProvided, LatentOutputProvided);
+                    } else {
+                        ZoneCoolingLoad = ZoneSysEnergyDemand(ZoneNum).RemainingOutputReqToCoolSP;
+                        // calculate part load ratio for cycling fan/unit first
+                        ControlZoneEvapUnitOutput(UnitNum, ZoneCoolingLoad);
+                        PartLoadRatio = ZoneEvapUnit(UnitNum).UnitPartLoadRatio;
+                        CalcZoneEvapUnitOutput(UnitNum, PartLoadRatio, SensibleOutputProvided, LatentOutputProvided);
```
Myoldmopar, line 4473: I was getting nervous about where all this code went but I see you moved it to separate functions.  This is good, just makes it a little difficult to spot the changes you specifically made to the functionality.  Can you make some comments in the code here on the lines you may have changed to get this working?

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -188,6 +188,7 @@ set( test_src
   XingGroundTemperatureModel.unit.cc
   ZoneContaminantPredictorCorrector.unit.cc
   ZoneEquipmentManager.unit.cc
+  ZoneHVACEvaporativeCooler.unit
```
Myoldmopar, line 193: It's become rare for entirely new unit test files!

Myoldmopar (commented):
@Nigusse everything I've seen seems OK, but I'd like for you to comment up the changes a little bit so I can confirm that I understand the impact.  Thanks!

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: The variable speed fan modulates the supply air flow to meet the cooling load, so the zone evaporative unit does not need to cycle.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -3771,6 +3780,14 @@ namespace EvaporativeCoolers {
...
                     ZoneEvapUnit(UnitLoop).FanAvailSchedPtr = HVACFan::fanObjs[ZoneEvapUnit(UnitLoop).FanIndex]->availSchedIndex;
                 }
+
+                if (ZoneEvapUnit(UnitLoop).FanType_Num == DataHVACGlobals::FanType_SimpleOnOff ||
+                    ZoneEvapUnit(UnitLoop).FanType_Num == DataHVACGlobals::FanType_SystemModelObject) {
+                    ZoneEvapUnit(UnitLoop).OpMode = DataHVACGlobals::CycFanCycCoil;
+                } else {
+                    ZoneEvapUnit(UnitLoop).OpMode = DataHVACGlobals::ContFanCycCoil;
+                }
+
```
Nigusse, line 3790: This is new code added in Get Input to catch and set the fan operating mode.  Later the "OpMode" is used whether to allow or not allow the unit to cycle.

src/EnergyPlus/EvaporativeCoolers.cc
Myoldmopar: Yeah OK makes good sense :smile:

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
-                                                        ZoneEvapUnit(UnitNum).DesignFanSpeedRatio,
-                                                        ZoneCompTurnFansOn,
-                                                        ZoneCompTurnFansOff); // DesignFanSpeedRatio used by OnOff fan for power calc
-                        } else {
-                            HVACFan::fanObjs[ZoneEvapUnit(UnitNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
-                        }
-                    }
-
-                    if (ZoneEvapUnit(UnitNum).EvapCooler_1_AvailStatus) {
-                        SimEvapCooler(ZoneEvapUnit(UnitNum).EvapCooler_1_Name, ZoneEvapUnit(UnitNum).EvapCooler_1_Index);
```
Nigusse, line 4482: This code section is is duplicate for the three ControlSchemeType available. I moved and created a new function called CalcZoneEvapUnitOutput(). This was the first step I did and made sure it did not break any thing. After confirming CI machine no diffs, then I broke the code for cycling and continuous run operating mode depending on the fan type.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
+                    if (ZoneEvapUnit(UnitNum).OpMode == DataHVACGlobals::ContFanCycCoil) {
+                        PartLoadRatio = 1.0;
+                        ZoneEvapUnit(UnitNum).UnitPartLoadRatio = PartLoadRatio;
+                        CalcZoneEvapUnitOutput(UnitNum, PartLoadRatio, SensibleOutputProvided, LatentOutputProvided);
```
Nigusse, line 4467: This section is for continuous fan operating mode (continuous Zone Evap Cooler). I set the PLR to 1.0 such that is maintains the original operating mode. The "ELSE" section is the one that allows the evap unit to cycle if the condition is favourable. Under this new section I determine a PLR such that evap unit meet the zone cooling load. For this I added a new function  ControlZoneEvapUnitOutput() that calculates PLR.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
+                    } else {
+                        ZoneCoolingLoad = ZoneSysEnergyDemand(ZoneNum).RemainingOutputReqToCoolSP;
+                        // calculate part load ratio for cycling fan/unit first
+                        ControlZoneEvapUnitOutput(UnitNum, ZoneCoolingLoad);
```
Nigusse, line 4471: new function added for calculating PLR using root solver.

mjwitte:
I don't see any input changes here. How would one run the fan continuously, assuming anyone would want to?

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: In this function what changed is the supply flow rate is multiplied by PLR to determine the actual delivered out put. For Continuous fan operating mode, the PLR is always set to 1.0.

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: The function CalcZoneEvapUnitOutput() is now runs with PLR determined for current cooling load. When the unit is allowed to cycle, it delivers just the requested zone cooling load, instead of the available capacity.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
-                        SimEvapCooler(ZoneEvapUnit(UnitNum).EvapCooler_2_Name, ZoneEvapUnit(UnitNum).EvapCooler_2_Index);
-                    }
-                    if (ZoneEvapUnit(UnitNum).FanLocation == DrawThruFan) {
-                        if (ZoneEvapUnit(UnitNum).FanType_Num != DataHVACGlobals::FanType_SystemModelObject) {
-                            Fans::SimulateFanComponents(
-                                ZoneEvapUnit(UnitNum).FanName, false, ZoneEvapUnit(UnitNum).FanIndex, _, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
-                        } else {
-                            HVACFan::fanObjs[ZoneEvapUnit(UnitNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
-                        }
-                    }
```
Nigusse, line 4549: This ELSE section was meant to set the flow to zero when the evap unit is off. Now I can do the same with PLR set to zero and running the same function CalcZoneEvapUnitOutput(). This function also set these mass flow rates to zero if the PLR is zero.  I consolidated the On and Off operating modes of the evap unit using the same function CalcZoneEvapUnitOutput() and works for the three ControlSchemeType.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4453,99 +4476,23 @@ namespace EvaporativeCoolers {
...
+                    PartLoadRatio = 0.0;
+                    ZoneEvapUnit(UnitNum).UnitPartLoadRatio = PartLoadRatio;
+                    CalcZoneEvapUnitOutput(UnitNum, PartLoadRatio, SensibleOutputProvided, LatentOutputProvided);
                 }
 
             } else if (SELECT_CASE_var == ZoneCoolingLoadOnOffCycling) {
```
Nigusse, line 4483: The same comment applies to the ControlSchemeType ZoneCoolingLoadOnOffCycling.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4659,104 +4529,210 @@ namespace EvaporativeCoolers {
...
+        // calculate sensible and latent outputs delivered
+        MinHumRat = min(Node(ZoneNodeNum).HumRat, Node(OutletNodeNum).HumRat);
         SensibleOutputProvided =
-            Node(ZoneEvapUnit(UnitNum).UnitOutletNodeNum).MassFlowRate * (PsyHFnTdbW(Node(ZoneEvapUnit(UnitNum).UnitOutletNodeNum).Temp, MinHumRat) -
-                                                                          PsyHFnTdbW(Node(ZoneEvapUnit(UnitNum).ZoneNodeNum).Temp, MinHumRat));
-        LatentOutputProvided = Node(ZoneEvapUnit(UnitNum).UnitOutletNodeNum).MassFlowRate *
-                               (Node(ZoneEvapUnit(UnitNum).UnitOutletNodeNum).HumRat - Node(ZoneEvapUnit(UnitNum).ZoneNodeNum).HumRat);
+            Node(OutletNodeNum).MassFlowRate * (PsyHFnTdbW(Node(OutletNodeNum).Temp, MinHumRat) - PsyHFnTdbW(Node(ZoneNodeNum).Temp, MinHumRat));
+        LatentOutputProvided = Node(OutletNodeNum).MassFlowRate * (Node(OutletNodeNum).HumRat - Node(ZoneNodeNum).HumRat);
+    }
```
Nigusse, line 4622: This function CalcZoneEvapUnitOutput() is factored from an existing code to handle the On and Off operating modes and scale the supply flow rate as a function of PLR. Also set the flow rates to zero when the unit is off (PLR=0). Factoring out this code removed duplicate code sections.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4659,104 +4529,210 @@ namespace EvaporativeCoolers {
...
+        int const OutletNodeNum = ZoneEvapUnit(UnitNum).UnitOutletNodeNum;
+        int const ReliefNodeNum = ZoneEvapUnit(UnitNum).UnitReliefNodeNum;
+        int const FanInletNodeNum = ZoneEvapUnit(UnitNum).FanInletNodeNum;
+        int const FanOutletNodeNum = ZoneEvapUnit(UnitNum).FanOutletNodeNum;
+        int const EvapCooler_1_Index = ZoneEvapUnit(UnitNum).EvapCooler_1_Index;
+        int const EvapCooler_2_Index = ZoneEvapUnit(UnitNum).EvapCooler_2_Index;
+
+        // calculate unit sensible cooling output
+        if (PartLoadRatio > 0) {
+            Node(OAInletNodeNum).MassFlowRate = ZoneEvapUnit(UnitNum).DesignAirMassFlowRate * PartLoadRatio;
```
Nigusse, line 4550: PLR is used to modulate the supply air flow. Prior to this fix, the avp units always run at full flow, and some times over-cooled the space. This is one of the main request of the user, avoiding over cool.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4659,104 +4529,210 @@ namespace EvaporativeCoolers {
...
+            Node(EvapCond(EvapCooler_1_Index).OutletNode).MassFlowRate = 0.0;
+            Node(EvapCond(EvapCooler_1_Index).OutletNode).MassFlowRateMaxAvail = 0.0;
+
+            if (EvapCooler_2_Index > 0) {
+                Node(EvapCond(EvapCooler_2_Index).InletNode).MassFlowRate = 0.0;
+                Node(EvapCond(EvapCooler_2_Index).InletNode).MassFlowRateMaxAvail = 0.0;
+                Node(EvapCond(EvapCooler_2_Index).OutletNode).MassFlowRate = 0.0;
+                Node(EvapCond(EvapCooler_2_Index).OutletNode).MassFlowRateMaxAvail = 0.0;
+            }
+        }
```
Nigusse, line 4575: Set all flow rates to zero if the unit is off (PLR = 0).

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4659,104 +4529,210 @@ namespace EvaporativeCoolers {
...
+                    "Zone Evaporative Cooler unit part load ratio control failed (limits exceeded) for ZoneHVAC:EvaporativeCoolerUnit =\"" +
+                        ZoneEvapUnit(UnitNum).Name,
+                    ZoneEvapUnit(UnitNum).UnitLoadControlLimitsErrorIndex);
+            }
+
+        } else {
+            PartLoadRatio = 1.0;
+        }
+        ZoneEvapUnit(UnitNum).UnitPartLoadRatio = PartLoadRatio;
+    }
```
Nigusse, line 4680: The above function ControlZoneEvapUnitOutput() solves for the PLR.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4659,104 +4529,210 @@ namespace EvaporativeCoolers {
...
+        Real64 QLatOutputProvided;  // latent output at a given PLR
+
+        UnitNum = int(Par(1));
+        LoadToBeMet = Par(2);
+
+        CalcZoneEvapUnitOutput(UnitNum, PartLoadRatio, QSensOutputProvided, QLatOutputProvided);
+
+        Residual = QSensOutputProvided - LoadToBeMet;
+
+        return Residual;
```
Nigusse, line 4702: new residual function ZoneEvapUnitLoadResidual().

Nigusse:
@Myoldmopar I added inline comments about functionality and why I made the changes.

Nigusse:
> I don't see any input changes here. How would one run the fan continuously, assuming anyone would want to?

Now that depends on the Fan Type only. Only OnOff and SystemModel fan types are allowed to cycle. All other fan types will run continuously. We do not have input field for fan operating mode in ZoneHVACEvap unit for now. So the unit will run only in one of the two operating modes through out the simulation.

src/EnergyPlus/EvaporativeCoolers.cc
mjwitte: Oh I see, onoff is triggered by fan type. This won't fly after we deprecate all of the old fans and move to Fan:SystemModel as the only fan. This unit probably needs a Supply Fan Operating Mode Schedule Name, similar to PTAC, PTHP, FanCoil, etc.

Myoldmopar: @mjwitte are you requesting he add this on this PR?  If not, are you satisfied with the current state?

Myoldmopar:
> @Myoldmopar I added inline comments about functionality and why I made the changes.

Thanks!  That's a great walkthrough of the code changes!

src/EnergyPlus/EvaporativeCoolers.cc
mjwitte: I think we should add the field now. It does require some thought for the variable fan case. Does it currently check the settings inside Fan:SystemModel to determine if it's VAV?

Nigusse: OK. I will append new field at the end. But we do not want to stretch this issue into "new feature".

mjwitte: Before you code it, let's think through all the use cases and make sure it's clean. e.g. If someone currently has VAV, should this schedule be always 1?

Nigusse: Here are the different option to consider:

(1) if the fan is VAV (Fan:VariableVolume or Fan:SystemModel with continuous Speed Control Method) then the schedule should be always 1 (no cycling) regardless of the fan op schedule. This requires to scan the Fan:SystemModel object. I do not know if there a method to tell or how easy it is to find out.

(2) Fan:OnOff and Fan:SystemModel (with number of speed = 1 with discrete "Speed Control Method:) should default to cycling fan unless the FanOpschedule tells us differently

(3)  Fan:ConstantVolume defaults to continuous fan always.

(4) What should we do if the input field is blank? The default operating mode should be dependent on the fan type, shouldn't it?

mjwitte: <sigh> nothing simple. Question is - how important is it to reproduce past behavior? If the field is blank, then it should default to cycling fan, regardless of fan type.
What does, say, PTAC do if you specify Fan:ConstantVolume with a zero fan operating mode schedule? Does it reject that, or go ahead and cycle the Fan:ConstantVolume?

Nigusse: If you specify Fan:ConstantVolume with a zero fan operating mode schedule value, then EPlus throws severe error message and fatal out. See code [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/PackagedTerminalHeatPump.cc#L1381-L1406). This seems very restrictive. On the other hand if you instead leave the supply air fan operating schedule name field blank, then EPlus defaults to cycling mode and simulation continues. See the code [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/PackagedTerminalHeatPump.cc#L1304-L1311).

Nigusse: Current code allows zone Evap cooler to run in continuously mode regardless of the fan type supported.

Nigusse: If we implement with approach used in PTAC and the supply air fan operation schedule is left bank, then the Zone Evap Cooler cycles regardless of the fan type for ZoneTemperatureDeadbandOnOffCycling and ZoneCoolingLoadOnOffCycling Unit Control Methods but has no effect when the Unit Control Method is ZoneCoolingLoadVariableSpeedFan.

But we can always reproduce the the past behavior if set the supply air fan operation mode schedule value is set to 1 but requires transition rule.

Nigusse: @mjwitte OK. I will proceed with Fan Op Mode implementation similar to PTAC but with possibility of providing transition rule to reproduce past results. Let me know what you think?

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: @Myoldmopar updated Evap structure.

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: @MJW @Myoldmopar Second look at allowing the fan operating mode for zone Evap Unit: 

The existing implementation is not a "Continuous Fan Flow", rather it is FULL fan flow if the Evap Unit is ON based on the Evap Unit controllers decision. If the Evap Unit controller decides to turn OFF the evaporative cooler, then the Evap Unit and the fan are OFF. So, in the existing implementation, the fan and the Evap Unit do not run unless the controller decides to turn ON the Evap cooler (Evap Unit). Therefore, the exiting implementation is "some sort" of cycling fan (cycling unit) but with FULL fan flow for ZoneTemperatureDeadbandOnOffCycling and ZoneCoolingLoadOnOffCycling Unit Control Methods. ZoneCoolingLoadVariableSpeedFan Unit Control Method modulates the fan flow as the name implies.

What this fix allowed is to modulate the fan flow rate for the ZoneTemperatureDeadbandOnOffCycling and ZoneCoolingLoadOnOffCycling Unit Control Methods to avoid over cooling. This fix makes Evap Unit control similar to how we cycle other zone equipment depending on fan and load.

Now the question is do we want to allow "Continuous Fan Flow" just like the other  zone equipment for Evap Cooler Unit? This will require to turn OFF the evaporative-cooler but allow FULL flow outside air to the zone. This change would be quite different from what the user requested.

mjwitte: @Nigusse Ohhhhh - I hadn't realized this unit has a "Cooler Unit Control Method" field. Maybe that simplifies things. 

So, now, let's not add new schedule field. It probably doesn't make sense for a unit like this - who would want a continuous flow of air that gets hot when the water cycles off?

So, it seems the "Cooler Unit Control Method" should dictate the operation, regardless of fan type. If the fan settings are not compatible with the selected control type, then either severe/fatal out, or throw a warning and flip the control choice?

Nigusse: @mjwitte Thanks. Will look into "Cooler Unit Control Method" and fan type combination.

Nigusse: Existing code of the zone Evap Cooler Model supports five fan types:  "Fan:OnOff", "Fan:SystemModel", "Fan:ComponentModel", "Fan:ConstantVolume" and "Fan:VariableVolume".

If the Cooler Unit Control Method is ZoneCoolingLoadVariableSpeedFan, "Fan:OnOff" and "Fan:ConstantVolume" currently are not allowed and EPlus fatal out with severe error message. 

However, with the other two Cooler Unit Control Methods (ZoneCoolingLoadOnOffCycling and ZoneTemperatureDeadBandOnOffCycling) existing code allows all the five fan types.

Below is the proposed change for ZoneCoolingLoadOnOffCycling and ZoneTemperatureDeadBandOnOffCycling Cooler Unit Control Methods:
(1) fan types allowed will be restricted to "Fan:OnOff" and "Fan:SystemModel"
(2) fatal out with severe error message for the other three fan types
(3) Update existing example files based on the first two items of proposed changes

Nigusse: I see only two example files that need to change the fan type in multiple places for the two cooler unit control methods. I am planning to replace the :OnOff, :ConstantVolume, and :VariableVolume fan types with :SystemModel used with the two control methods only. @mjwitte Let me know if you have any objection.

mjwitte: @Nigusse Why make the fan types more restrictive now? If that's necessary, then we'll need transition rules to change the fan types.
Also, if it's made more restrictive, Fan:ComponentModel should also be valid.

Nigusse: @mjwitte I am just trying to allowing fan types I thought appropriate for these two controller types. Otherwise, I do not have any specific reason. Are you suggesting that we allow all the five fan types for the ZoneCoolingLoadOnOffCycling and ZoneTemperatureDeadBandOnOffCycling Cooler Unit Control Methods as it is done in the existing code?

mjwitte: Yes, as long as they produce reasonable results (i.e. power consumption that makes sense).

Nigusse: OK. I will revise the code accordingly and push to remote. And I will review the regression results.

Nigusse: @mjwitte I created a single zone model to investigate electric power consumption. see below the fan and evap unit electric power for "EvaporativeCooler:Direct:CelDekPad". The Unit's electric power for this evap cooler is the electric power of the Water Pump only.  The supply fan power reported is consistent with the operation mode of the fan type (cycling for Fan:OnOff and Full Flow for Fan:ConstantVolume) when the evap unit is ON depending the controller action. The control type tested is ZoneCoolingLoadOnOffCycling. I see a problem with the electric power (water Pump Electric Power), it shouldn't have been constant for cycling fan/Unit (the first graph) when the PLR is less than 1.0. This should be fixed, but I am afraid this may be the case for the other evap cooler types as well.

CyclingUnit_FullFlowUnit

mjwitte: But I'm trying to get away from the fan type affecting operation. Why aren't these two cases the same?

Nigusse: Do we allow Fan:ConstantVolume fan to cycle? The original code allowed FULL flow regardless of the fan type when the unit ON and causing zone over-cool. Now, this fix allowed the unit/fan to cycle but only with these three fan types (Fan:OnOff, Fan:SystemModel, Fan:ComponentModel).  I can change the fix to allow the unit to cycle regardless of the fan type, if you agree?

mjwitte: That's what I thought we agreed on. That's assuming Fan:ConstantVolume will work that way. The IDD notes for it imply otherwise:
```
Fan:ConstantVolume,
       \memo Constant volume fan that is intended to operate continuously based on a time schedule.
       \memo This fan will not cycle on and off based on cooling/heating load or other control
       \memo signals.
```
If that's true, then we should throw an error if it's used with a cycling control option and add a transition rule to replace it with Fan:OnOff when use with the evap cooler.
@Myoldmopar It's time we dump all of these old fans. They should all be marked obsolete for this release. And possibly transitioned away in this release and the next.

Nigusse:
Diffs are expected based on the code changes made and is ready for final review. . Also added defect files.

doc/input-output-reference/src/overview/group-zone-forced-air-units.tex
```diff
@@ -1484,9 +1484,9 @@ \subsubsection{Inputs}\label{inputs-4-040}
...
+  \textbf{ZoneCoolingLoadOnOffCycling}. This control method determines whether or not to operate the cooler based on the predicted zone load to cooling setpoint.~ If there is a cooling load (and the unit is available), then the unit is operated for the entire timestep at the design air mass flow rate. If there is no cooling load, then the unit is not operated at all.~ The magnitude of the cooling load that will trigger the unit to operate can be controlled using the field below called Cooling Load Control Threshold Heat Transfer Rate. If the unit is On, then the unit is cycled such that it meets the current cooling load and avoids over-cooling the zone.
```
mjwitte: For this control type, seems it should read "If there is a cooling load (and the unit is available), then the unit is cycled at the design air mass flow rate such that it meets the current cooling load . . . " (essentially merging the last sentence with the second sentence).

tst/EnergyPlus/unit/ZoneHVACEvaporativeCooler.unit.cc
```diff
@@ -0,0 +1,527 @@
...
+    ASSERT_FALSE(ErrorsFound);
+
+    DataGlobals::BeginEnvrnFlag = true;
+    DataZoneEquipment::ZoneEquipInputsFilled = true;
+
+    auto &thisZoneEvapCooler(ZoneEvapUnit(UnitNum));
+
+    DataZoneEquipment::ZoneEquipConfig(1).ExhaustNode(1) = thisZoneEvapCooler.UnitReliefNodeNum;
+
+    DataLoopNode::Node.redimension(NumOfNodes);
```
mjwitte, line 214: Hmm, redimension after `GetInputZoneEvaporativeCoolerUnit`? Not following why that should happen here.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -3630,8 +3652,8 @@ namespace EvaporativeCoolers {
         // Using/Aliasing
         using BranchNodeConnections::SetUpCompSets;
         using DataGlobals::NumOfZones;
-        using DataHVACGlobals::FanType_SimpleConstVolume;
-        using DataHVACGlobals::FanType_SimpleOnOff;
+        // using DataHVACGlobals::FanType_SimpleConstVolume;
+        // using DataHVACGlobals::FanType_SimpleOnOff;
```
mjwitte: If these are other changes needed, these should be deleted.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4660,104 +4526,210 @@ namespace EvaporativeCoolers {
...
-                    Node(EvapCond(ZoneEvapUnit(UnitNum).EvapCooler_1_Index).OutletNode).MassFlowRateMaxAvail = 0.0;
-
-                    if (ZoneEvapUnit(UnitNum).EvapCooler_2_Index > 0) {
-                        Node(EvapCond(ZoneEvapUnit(UnitNum).EvapCooler_2_Index).InletNode).MassFlowRate = 0.0;
-                        Node(EvapCond(ZoneEvapUnit(UnitNum).EvapCooler_2_Index).InletNode).MassFlowRateMaxAvail = 0.0;
-                        Node(EvapCond(ZoneEvapUnit(UnitNum).EvapCooler_2_Index).OutletNode).MassFlowRate = 0.0;
-                        Node(EvapCond(ZoneEvapUnit(UnitNum).EvapCooler_2_Index).OutletNode).MassFlowRateMaxAvail = 0.0;
-                    }
+        // METHODOLOGY EMPLOYED:
+        // na
```
mjwitte: We don't add unused boiler plate comments anymore to new functions, here and below.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4660,104 +4526,210 @@ namespace EvaporativeCoolers {
...
-                    }
+        // REFERENCES:
+        // na
 
-                    if (ZoneEvapUnit(UnitNum).EvapCooler_1_AvailStatus) {
-                        SimEvapCooler(ZoneEvapUnit(UnitNum).EvapCooler_1_Name, ZoneEvapUnit(UnitNum).EvapCooler_1_Index);
-                    }
+        // Using/Aliasing
+        using DataHVACGlobals::ZoneCompTurnFansOff;
+        using DataHVACGlobals::ZoneCompTurnFansOn;
```
mjwitte: Minor, but prefer to avoid `using` in new functions.

mjwitte (commented):
Getting started on final review. I've merged in the latest develop branch and will push that here.

When running all unit tests serially,  ZoneHVACEvapCoolerUnitTest.DirectResearchSpecial_CyclingUnit_Sim fails with an array bounds error. I don't see any deallocates in `EvapCoolers::clear_state` for `EvapCond, ZoneEvapUnit, or ZoneEvapCoolerUnitFields`.

Some review comments below. I haven't tested idf files yet, so more next week.
Does this need any changes to the Engineering Reference?

doc/input-output-reference/src/overview/group-zone-forced-air-units.tex
Nigusse: Updated as suggested.

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: Done.

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: Done

src/EnergyPlus/EvaporativeCoolers.cc
Nigusse: updated as suggested.

tst/EnergyPlus/unit/ZoneHVACEvaporativeCooler.unit.cc
Nigusse: Required to include the zone related nodes defined in the EnergyPlusFixture::SetUp() [here](https://github.com/NREL/EnergyPlus/blob/166330897_Issue6707/tst/EnergyPlus/unit/ZoneHVACEvaporativeCooler.unit.cc#L99-L114);

Nigusse:
Addressed failed unit test by adding module variables to clear function and moving a static local variable to module level namespace.

Nigusse:
@mjwitte Made minor modification in the Engineering Document for "ZoneTemperatureDeadbandOnOffCycling" and "ZoneCoolingLoadOnOffCycling" control type to reflect the cycling operation added.

tst/EnergyPlus/unit/ZoneHVACEvaporativeCooler.unit.cc
rraustad: The name of the evap cooler is different from the base fixture and the zone inlet node name looks unique, so number of nodes will probably be different from what was expected based on the base fixture setup. But I agree that the need for reallocation of nodes is suspect. Didn't `GetInputZoneEvaporativeCoolerUnit` allocate new nodes as they were read in (e.g., DataZoneEquipment::ZoneEquipConfig(1).ZoneNode should be a valid node)? What was NumOfNodes compared to the allocated Node array size?

Nigusse: Get input does not allocate all nodes required for zone connection. That is why I re-dimension the "Node" array.

Nigusse: The number of nodes allocated during get input depends on the unit test but it was some where between 4 - 7, I think.  I reserved node number 9 and 10 for return and zone air node.

Myoldmopar:
Wow.  This has been a lengthy and informative read.  Thanks to @Nigusse @mjwitte and @rraustad for all the effort in making code changes and reviewing.  Can I get a feel for where I could step in at this point to provide review support to get this merged in?  Running test files and checking diffs?  Something else?

mjwitte:
@Myoldmopar Next steps here, run unit tests to make sure clear_state solves the noted problem, run test files, try some disallowed options to see if they are trapped correctly, make sure fan operation and tstat control are as expected.

mjwitte (approved):
Tested this with all fan types. New temperature control is excellent. Since it's been a few weeks, merging in develop and letting CI make another pass. Otherwise, this is great.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -1727,8 +1736,12 @@ namespace EvaporativeCoolers {
             //***************************************************************************
             //                  ENERGY CONSUMED BY THE RECIRCULATING PUMP
             // Add the pump energy to the total Evap Cooler energy comsumption
-            EvapCond(EvapCoolNum).EvapCoolerPower += EvapCond(EvapCoolNum).RecircPumpPower;
-
+            Real64 FlowFraction = 1.0;
+            Real64 MassFlowRateMax = Node(EvapCond(EvapCoolNum).InletNode).MassFlowRateMax;
+            if (MassFlowRateMax > 0) {
+                FlowFraction = EvapCond(EvapCoolNum).InletMassFlowRate / MassFlowRateMax;
```
mjwitte: @Nigusse @Myoldmopar @rraustad So, I was puzzled by the meter diffs in 1ZoneEvapCooler and 1ZoneParameterAspect, because these have airloop evap coolers, no zoneHVAC. 

It turns out that 1ZoneParameterAspect is hard-sized with the airloop, fan, and OA controller all at 3.0 but the terminal unit is at 2.5. So, the evap cooler power is always 2.5/3.0 on this branch compared to develop. (1ZoneEvapCooler is similar with 2.0 vs 1.5). I don't see anything in the comments to indicate that the different flow rates are intentional. So, one option here is to simply fix those to use the terminal unit flow rate throughout and then the diffs would go away. 

Since there is no design flow rate input for EvaporativeCooler:*:CelDekPad, using the node max flow rate is a clever solution to calculate a "plr" for the evap cooler, but will this always produce the desired evap cooler power consumption? What if the evap cooler is in a VAV system, does that still make sense?

Thoughts?

rraustad: 1) I would expect MassFlowRateMax = branch max flow so this max looks accurate.
2) Does EvapCooler have a VS pump? Probably no since the misters need high pressure to mist water.
3) if EvapCooler cycles then this seems reasonable for pump power, if EvapCooler runs entire time step then previous pump power calc was accurate.
4) Don't care if it's VAV. How long does the pump run this time step is more important.

mjwitte: So, perhaps instead of using flow rate as a proxy for PLR, the PLR should default to 1.0 and ZoneHVAC:EvapCooler needs to pass a PLR to the evap cooler component?

rraustad: The outlet conditions are correct now so I was thinking more like:

If (constant fan){
 same power equation as before
} else {
 new PLR equation
}

mjwitte: The base evap cooler component doesn't know anything about fan type - only the ZoneHVAC:EvapCooler knows that. If the eval cooler component is loose on a branch, then it doesn't know what the fan will do.

rraustad: @Nigusse you need to figure out a way to know when the fan is cycling vs constant so you can get the correct pump power.

rraustad: If evap cooler is loose on the air loop branch you can use this bool and probably also if AirLoopNum > 0:
AirLoopControlInfo(AirLoopNum).AnyContFan, or could pass PLR = 1 as @mjwitte suggested.

If evap cooler is in the ZoneHVAC:EvapCooler then you need to pass info to the evap cooler to tell it if it is cycling or constant.

Passing PLR as an argument is probably easiest unless you can think of something else.

Nigusse: @mjwitte @rraustad I am thinking to pass the ZoneHVAC:EvapCooler unit PLR as an optional argument to SimEvapCooler() function and cascade it to CalcDirectEvapCooler() and CalcDryIndirectEvapCooler() functions. If the  PLR argument is not present, then I will reset it to 1.0.  The later will be used for air loop.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -193,7 +193,7 @@ namespace EvaporativeCoolers {
 
     // Functions
 
-    void SimEvapCooler(std::string const &CompName, int &CompIndex)
+    void SimEvapCooler(std::string const &CompName, int &CompIndex, Optional<Real64 const> PartLoadRatio)
```
mjwitte: @Nigusse We're trying to avoid using optional arguments (for performance). A default value for the argument is better (if the argument is left off then it's set to the default value). See [discussion here](https://github.com/NREL/EnergyPlus/pull/7230#issuecomment-517287379). Then you don't need `if (present . . .)`. You can just use and pass `PartLoadRatio` directly.

Nigusse: @mjwitte Okay, will modify the code accordingly. Thanks.

Nigusse: done.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -4966,11 +4908,11 @@ namespace EvaporativeCoolers {
         }
 
         if (ZoneEvapUnit(UnitNum).EvapCooler_1_AvailStatus) {
-            SimEvapCooler(ZoneEvapUnit(UnitNum).EvapCooler_1_Name, ZoneEvapUnit(UnitNum).EvapCooler_1_Index);
+            SimEvapCooler(ZoneEvapUnit(UnitNum).EvapCooler_1_Name, ZoneEvapUnit(UnitNum).EvapCooler_1_Index, FanSpeedRatio);
```
rraustad, line 4911: As @mjwitte found with a lower max TU flow rate than the branch max flow, this isn't really the PLR of the evap cooler. I assume the diff's found yesterday would still be there.

Nigusse: PLR is passed to zone evaporative coolers only. When an evap cooler in airloop the PLR defaults to 1.0. The fanspeedratio is used when the zone HVAC evaporative cooler control type is ZoneCoolingLoadVariableSpeedFan.

rraustad: Ok then, as long as this is the actual PLR based on time.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -2029,24 +2037,29 @@ namespace EvaporativeCoolers {
...
+            Real64 MassFlowRateMax = Node(EvapCond(EvapCoolNum).InletNode).MassFlowRateMax;
+            if (MassFlowRateMax > 0) {
+                FlowFraction = EvapCond(EvapCoolNum).InletMassFlowRate / MassFlowRateMax;
+            }
             //                  POWER OF THE SECONDARY AIR FAN
             if (EvapCond(EvapCoolNum).IndirectFanEff > 0.0) {
-                EvapCond(EvapCoolNum).EvapCoolerPower +=
-                    EvapCond(EvapCoolNum).IndirectFanDeltaPress * EvapCond(EvapCoolNum).IndirectVolFlowRate / EvapCond(EvapCoolNum).IndirectFanEff;
+                EvapCond(EvapCoolNum).EvapCoolerPower += FlowFraction * EvapCond(EvapCoolNum).IndirectFanDeltaPress *
+                                                         EvapCond(EvapCoolNum).IndirectVolFlowRate / EvapCond(EvapCoolNum).IndirectFanEff;
```
mjwitte, line 2048: The indirect fan power is still based on FlowFraction. Is that appropriate or should this also be based on PLR?

Nigusse: I think, it should be based on PLR. It is an oversight. I will make the changes.

mjwitte:
@Nigusse Comparing the latest vs yesterday's code, there is a change in StripMallZoneEvapCooler (and likely similar in StripMallZoneEvapCoolerAutosized but I haven't looked). SMSTORE7 DIRECT EVAPORATIVE COOLER:Evaporative Cooler Electric Energy [J] is now varying every hour, whereas yesterday it was always at full power. Since SMSTORE7 INDIRECT EVAPORATIVE COOLER:Evaporative Cooler Electric Energy [J](Hourly) varies both yesterday and today, I'm concluding that today's result is correct for SMSTORE7 Direct. Would you agree?

Nigusse:
@mjwitte I run the StripMallZoneEvapCooler test file with latest code of this branch, and I see the Evaporative Cooler Electric Energy and Evaporative Cooler Electric Power "proportional" to PLR. I agree varying values should be the trend expected.

mjwitte:
Nice! Now we're down to diffs in just the 3 files with ZoneHVAC:EvaporativeCoolerUnit. Will let CI finish up and then merge.

src/EnergyPlus/EvaporativeCoolers.cc
yzhou601: @rraustad @mjwitte @Nigusse  Hi, sorry for jumping into the conversation which I might haven't understood thoroughly. So, is there a way to model cycling fan and cycling evap cooler pump in loose air loop? So far I didn't find any approach to directly put cycling fan into air loop and make it cycling properly. And as far as I've learned in this conversation, there's no way to communicate PLR between fan and evap cooler (this PR is going to pass 1.0 for air loop), correct? Because my recent work involves modeling evap cooler in a loose air loop, it would be good to know.

mjwitte: There's no way presently to model cycling fans on built-up branches. EvaporativeCooler:*:ResearchSpecial is able to control itself to a setpoint.

rraustad: That's true for purely component built-up branches, but when evap cooler is used with parent objects, the fan operating mode schedule is known as AirLoopControlInfo(AirLoopNum).CycFanSchedPtr (as long as this variable is set correctly everywhere) and could, in the future, be used to cycle the evap cooler.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -240,11 +250,11 @@ namespace EvaporativeCoolers {
...
             if (SELECT_CASE_var == iEvapCoolerDirectCELDEKPAD) {
-                CalcDirectEvapCooler(EvapCoolNum);
+                CalcDirectEvapCooler(EvapCoolNum, ZoneEvapCoolerPLR);
             } else if (SELECT_CASE_var == iEvapCoolerInDirectCELDEKPAD) {
-                CalcDryIndirectEvapCooler(EvapCoolNum);
+                CalcDryIndirectEvapCooler(EvapCoolNum, ZoneEvapCoolerPLR);
             } else if (SELECT_CASE_var == iEvapCoolerInDirectWETCOIL) {
-                CalcWetIndirectEvapCooler(EvapCoolNum);
+                CalcWetIndirectEvapCooler(EvapCoolNum, ZoneEvapCoolerPLR);
             } else if (SELECT_CASE_var == iEvapCoolerInDirectRDDSpecial) {
```
yzhou601, line 258: @Nigusse Hi, Bereket, is there any specific reason not to allow researchspecial types cycling with fan as well?  In [these lines](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/EvaporativeCoolers.cc#L3367-L3373), it confused me that it reported power consumption by `FlowRatio` if there's modifier curve (regardless of `PartLoad` which is calculated to meet outlet setpoint), but by `PartLoad` if there's not(regardless of flow ratio). Actually I am not sure what should it be though, the current approach looks confusing to me. @rraustad @mjwitte

Nigusse: The research special evaporative coolers models are different from the other models. They have pump power modifier curve that accounts for pump power calculation as a function of flow fraction. The PartLoad is used only if user did not provide pump power correction curve.

yzhou601: @Nigusse Yes that's correct. So if there's no modifier curve, the pump power would only be calculated based on `PartLoad` ignoring mass flow ratio in air loop, so whatever amount of supply air the thermal zones need, it just cycling to meet outlet temperature controlled by setpoint manager (more like supply air temperature). Where it is different with other types is that others always assume a full load outlet temperature. At least in this case ( the `else` condition), it seems to be close to this PR's purpose to fix it. 

In my personal opinion (might be wrong), it seems only consider one of them in each situation looks not enough to include both outlet temperature control (PLR) and mass flow rate control (PLR).

yzhou601: I knew that type is different so you didn't touch that in this PR. I mean maybe we could rethink about if we need to change something there as well. 😛

PR #7289, opened 2019-05-02
Refactoring Plant Components

Myoldmopar:
Pull request overview
---------------------
This continues the refactor of plant components to use the new PlantComponent interface.  The idea is to get effectively no diffs, though some diffs due to different calling sequences may cause tiny diffs or different row ordering in the tabular output.  The specific components being refactored in this branch will depend on how the diffs come about.  As long as the diffs are easy to justify, more components will be added, but once they become unwieldy, a new branch will be started to continue refactoring.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add to input rules file for interfaces
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Required documentation updates?
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #7322, opened 2019-06-03
Plant component continued - without boilers

Myoldmopar:
Pull request overview
---------------------
The effort to convert plant components into the new PlantComponent interface continues in this pull request.  I had originally tried to add in boilers also in #7289 but that started introducing tons of diffs, so I pulled this back to a simpler state and expect it to merge in easily.  Then I can focus on just specifically boilers in that PR.


Myoldmopar:
The last set of results look pretty good.  The three files with diffs are the GSHP related files, and the only math related diffs are small diffs in the MTR file...that don't even show up on Windows.  The custom_check is an encoding problem but otherwise I think this could be ready to go in.

Myoldmopar:
Resolved the conflicts, rebuilt and everything was good.  Reran the unit test suite just for fun and it went fine.  I'll let CI check it overnight here and probably merge in the morning.

PR #7323, opened 2019-06-03
Update Mac Installer

### LINKED ISSUE #5410: El Capitan (Mac OSX) breaks EnergyPlus
Due to new System Integrity Protection, various locations are no longer user writable, even as root. This breaks the runenergyplus script and potentially the installation script since it installs symlinks to `/usr/bin`.

See [UnmetHours post regarding this](https://unmethours.com/question/13844/el-capitan-breaking-energyplus/).

### LINKED ISSUE #7298: IDFVersionUpdater  and EP-Launch-Lite not working on Mac for EnergyPlus V9.1
Issue overview
--------------
I have EnergyPlus V9.1 installed on a Mac running  MacOS Mojave Version 10.14.4 (18E226).  When I attempt to launch IDFVersionUpdater.app, I get the following error message:

"The application “IDFVersionUpdater.app” can’t be opened."

The application then does not run.

### Details
 - Platform (Operating system, version): MacOS Mojave V10.14.4
 - Version of EnergyPlus (if using an intermediate build, include SHA): Version 9.1
 - Unmethours link or helpdesk ticket number: NA

### Checklist

jmarrec:
Pull request overview
---------------------

Fix #5410: Symlink to `/usr/local/bin` on Mac 
Fix #7298: Mac EP-Launch-Lite and IDFVersionUpdater work on mac.

The Gist here is that `darwinpostflight.sh` is something only used by PackageMaker (the old packaging program used for Mac), and now we use QtIFW, so instead you should use some install operations (cf [doc](https://doc.qt.io/qtinstallerframework/operations.html)).

I also added actual components that you can choose to install or not (all enabled by default):
* Create Symlinks: if enabled, requires admin rights (you have to type your password), used to symlink to /usr/local/bin and to symlink the man page too. If disabled, then you don't need admin right (it won't ask for your password)
* DataSets, WeatherData and ExampleFiles: if disabled, you won't get these folders and their content.

Screen Shot 2019-06-03 at 4 37 18 PM

I have tested the installer and the various combinations of stuff you can install and diff'ed the directory structure compared to an installer of current develop, and everything does make sense. If you enable eveyrthing, the only difference is that now you don't have "Uninstall Energyplus.app" (since it's superseeded by the "Maintenance Tool" created by QtIFW that will correctly remove the symlinks etc))

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2b:

CMakeLists.txt
```diff
@@ -59,8 +59,6 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
-set(CPACK_PACKAGE_CONTACT "Edwin Lee <edwin.lee@nrel.gov>")
-
```
Myoldmopar, line 63: Take that, Edwin Lee!

cmake/Install.cmake
```diff
@@ -1,4 +1,6 @@
-set( CPACK_PACKAGE_VENDOR "US Department of Energy" )
+set(CPACK_PACKAGE_VENDOR "US Department of Energy" )
+set(CPACK_PACKAGE_CONTACT "Edwin Lee <edwin.lee@nrel.gov>")
```
Myoldmopar, line 2: Oh nevermind, there he is.

cmake/Install.cmake
```diff
@@ -146,13 +155,15 @@ install(FILES "${CMAKE_SOURCE_DIR}/release/ep.gif" DESTINATION "./")
...
-install( FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/" RENAME "V9-2-0-Energy+.idd" )
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/app_g_postprocess.py" DESTINATION "workflows/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/calc_soil_surface_temp.py" DESTINATION "workflows/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/coeff_check.py" DESTINATION "workflows/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/coeff_conv.py" DESTINATION "workflows/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/energyplus.py" DESTINATION "workflows/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/workflows/transition.py" DESTINATION "workflows/")
+install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/" RENAME "V9-2-0-Energy+.idd" )
+
+# Workflow stuff, takes about 40KB, so not worth it proposing to not install it
```
Myoldmopar, line 160: Agreed, and it would stink for a user to have to add it in later if they use EP-Launch-3

cmake/Install.cmake
```diff
@@ -292,23 +302,72 @@ configure_file("${CMAKE_SOURCE_DIR}/cmake/CMakeCPackOptions.cmake.in"
...
+)
+
+cpack_add_component(ExampleFiles
+  DISPLAY_NAME "Example Files"
+  DESCRIPTION "IDF Example Files"
+)
+
+cpack_add_component(WeatherData
+  DISPLAY_NAME "Weather Data"
+  DESCRIPTION "A few EPW files"
```
Myoldmopar: I'm going to make a change here to "A set of EPW files" rather than "A few..."

Myoldmopar (approved):
From a coding standpoint, this looks good.  The IFW installer is so much nicer than PackageMaker.  I built this locally and tested it out successfully.  Thanks for cleaning out so much stuff.  I'll push a small commit to change one string and then this can merge in.

cmake/Install.cmake
rraustad: Maybe "EPW Weather Files"?

Myoldmopar:
> * Create Symlinks: if enabled, requires admin rights (you have to type your password), used to symlink to /usr/local/bin and to symlink the man page too. If disabled, then you don't need admin right (it won't ask for your password)

On my work machine, it still asks for an admin password if I install it into /Applications even if I don't have symlinks installed.  It probably wouldn't have asked if I had installed it into a user directory.

cmake/Install.cmake
rraustad: Maybe "EPW Weather Files"? to mimic "IDF Example Files"

jasondegraw: Wouldn't that expand to "EnergyPlus Weather Weather Files"?

Myoldmopar: @jasondegraw EPW is actually a recursive acronym where the E stands for EnergyPlus Weather Data Files, so the full name is "EnergyPlus Weather Data Files Plus Weather Weather Files ... the 3rd Baron of Townsville"

As bad as having a PIN number or a VIN number or and IDF file is, I think "EPW Weather Files" will make perfect sense to the users.

rraustad: Well, yes. Using that same logic it would be "Input Data File Examples" and "EnergyPlus Weather Files".

PR #7326, opened 2019-06-05
AFN Contaminant Transport NFP

jasondegraw:
NFP for the new contaminant transport feature. Some description of the design is already in the NFP, but a fuller design document may be added later.

### DIFF
--- a/design/FY2019/NFP-AFN-Contaminant-Transport.md
+++ b/design/FY2019/NFP-AFN-Contaminant-Transport.md
@@ -31,16 +31,15 @@ For simulation of transport of trace contaminants, conservation of mass for a
 transported material is written for each zone as:
 
 ```
-dM/dt = sum(FF_j) + G - R*M
+dM/dt = sum(dot(M)) + G - R*M
 ```
 
-where `M` is the mass of material in a zone, the `FF_j`s are the fluxes of mass
+where `M` is the mass of material in a zone, the `dot(M)`s are the fluxes of mass
 of material into and out of the zone, `G` is the generation of material in the
 zone, and `R` is the removal rate of material (by non-flux processes) from the
-zone. Filtration
-processes are embedded in the `FF_j`s and the equation is typically divided by
-mass of air in the zone to obtain an equation in terms of concentrations. The
-resulting equation is
+zone. Filtration processes are embedded in the `dot(M)`s and the equation is
+typically writte in terms of zonal concentrations rather than mass. The resulting
+equation is
 
 ```
 dC/dt = sum(F_j*(1-n_j)C_j) + G - R*C = RHS
@@ -171,9 +170,11 @@ ZoneContaminantSourceAndSink:Generic:Constant,
        \required-field
        \type object-list
        \object-list ZoneNames
+       \note This field is ignored for the AirflowNetwork transport calculation
   A3 , \field AirflowNetwork Material
        \type object-list
        \object-list AirflowNetworkMaterials
+       \note This field is only used for the AirflowNetwork transport calculation and is ignored for the generic calculation
   N1 , \field Design Generation Rate
        \units m3/s
        \type real
--- a/design/FY2019/NFP-AFN-Contaminant-Transport.md
+++ b/design/FY2019/NFP-AFN-Contaminant-Transport.md
@@ -56,7 +56,7 @@ upon the airflows.
 
 To advance the conservation equations through time, there are a number of
 different options. For now, the first order finite difference discretization in
-time will be used in three different ways. First, the implicit Euler method will
+time will be used in two different ways. First, the implicit Euler method will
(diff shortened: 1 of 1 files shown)


design/FY2019/NFP-AFN-Contaminant-Transport.md
```diff
@@ -0,0 +1,262 @@
...
+
+```
+M_(t+h) = M_t + h*RHS_(t)  
+```
+
+This approach is limited in step size, while the other two are stable for all
+step sizes (though in some cases one will do better than the other). The main
+advantage of the explicit Euler approach is that it does not require solution
+of simulataneous equations.
+
```
lgu1234, line 87: You show 3 different approaches to solve the first order differential equation: explicit, semi and implicit. I don't like explicit, because results are not stable with a large time step. If you solve the equation individually, you may try an analytical solution approach.

design/FY2019/NFP-AFN-Contaminant-Transport.md
```diff
@@ -0,0 +1,262 @@
...
+      \note The zone name is where AirflowNetwork:Distribution:Component:Duct is exposed. Leave this field blank if the duct
+      \note conduction loss is ignored.
+ A6 , \field Filter 1
+      \begin-extensible
+      \type object-list
+      \object-list AirflowNetworkFilters
+ A7 , \field Filter 2
+      \type object-list
+      \object-list AirflowNetworkFilters
+ ...
```
lgu1234, line 157: I think that a filter is not used for every linkage. Instead, it will be better to define a filter as a component and use a linkage to link this component.

design/FY2019/NFP-AFN-Contaminant-Transport.md
```diff
@@ -0,0 +1,262 @@
...
+      \required-field
+      \type alpha
+      \reference AirflowNetworkMaterials
+      \note A unique name for the material.
+ N1;  \field Ambient Concentration
+      \required-field
+      \type real
+      \minimum 0
+      \note The default/ambient concentration of the material.
+```
```
lgu1234, line 248: I am not clear definition of Ambient Concentration. Is it outdoor? If it is outdoor, it is better to use other general objects, because it may be used in general (outside of AirflowNetwork)?

In addition, I prefer to use a schedule, instead of a constant.

design/FY2019/NFP-AFN-Contaminant-Transport.md
jasondegraw: I don't quite follow what you mean here. Do you mean do more inside the filter object?

design/FY2019/NFP-AFN-Contaminant-Transport.md
jasondegraw: Will the analytical solution really work here? I included the explicit procedure for situations in which the timestep is small,  but I would be OK eliminating it if you think it is too dangerous.

design/FY2019/NFP-AFN-Contaminant-Transport.md
jasondegraw: This would be the outdoor concentration and the default initial indoor concentration. I semi-intentionally left schedules out of the first draft. There are a lot of places where schedules could (and probably should) be added. Interaction with the current capability is a bit difficult, I'd like to (especially the source and sinks types) but I'm not sure how. This is something we should discuss further.

design/FY2019/NFP-AFN-Contaminant-Transport.md
```diff
@@ -0,0 +1,262 @@
...
+      \key CrankNicolson
+      \default None
+```
+
+The new IDD objects are:
+
+* A material object that describes a transported material
+
+```
+AirflowNetwork:Material,
```
jasondegraw, line 236: @mjwitte suggests that this name is too generic. Suggested options:

 * AirborneMaterial - I'm going to veto this one, the materials aren't always airborne
* Constituent - Vetoing this one too, to me this one implies that it is part of something
* Species - This is what CONTAM uses
* Substance - This is the one I like the best, if nobody chimes in and stops me I'm going to go with this one

nrel-bot-3:
@jasondegraw @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@jasondegraw @lgentile it has been 29 days since this pull request was last updated.

Myoldmopar:
@lgu1234 you had made a few comments on this NFP to which @jasondegraw replied.  Could you resolve those conversations so we can call the NFP complete on here?  Thanks!

lgu1234:
https://github.com/Myoldmopar Jason addressed my comments in the updated NFP. Yes, you can call the NFP complete from my point of view.

Myoldmopar:
Thank you @lgu1234, I am going to merge this PR as an NFP complete.

PR #7334, opened 2019-06-13
Remove Documentation build from default make rule

### LINKED ISSUE #7315: Doc builds should not be in default make rule
Issue overview
--------------
Right now, if you turn on BUILD_DOCS, and you just run make, it will build all the doc PDFs.  It would be nice if you could leave BUILD_DOCS turned on but not incur that build time penalty just for running the default make rule.  The default make rule should only build E+ and its dependencies, along with maybe a few key supporting tools (readvars, e.g., if BUILD_FORTRAN is turned on).  The docs should be a required dependency when building a package, but not when just building EnergyPlus.

This may not be as big of an issue on Windows, where you are often already focused on the singular EnergyPlusLib project for building.  I'm not actually sure whether that triggers doc builds or not...

### Details
 - Platform: Linux, but would be good to tighten up all the platforms

### Checklist
 - [x] Defect file added: N/A

jmarrec:
Pull request overview
---------------------

Fix #7315. This adds a new target called `documentation` that will build the PDF. This target isn't included in the default target (ALL), so even if you have enabled `BUILD_DOCS`, you'll have this target but will have to call it explicitly to build it.
When you build the `package` target, it will trigger build of `documentation` target too if you have  `BUILD_DOCS` enabled.

The only caveat is that, just like other things in cmake/Install.cmake that relies on `install(CODE ...)` (such as calling `parse_output_variables.py`) it will build the `documentation` target for each Cpack package generator.

Tested on Ubuntu and Windows.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


nrel-bot-2:

Myoldmopar:
@jmarrec I'm getting something weird:

```
$ make -j 7 package
[  0%] Built target gmock_main
[  0%] Built target sqlite
...
[ 79%] Built target energyplus
[100%] Built target energyplus_tests
Run CPack packaging tool...
CPack: Create package using STGZ
CPack: Install projects
CPack: - Run preinstall target for: EnergyPlus
CPack: - Install project: EnergyPlus
 +++ AutoDocs: Completed processing output variable audit: processed 3935 calls
 +++ AutoDocs: Completed processing example file summary: processed 663 example files
 +++ AutoDocs: Completed processing example file links summary: processed 663 example files
 +++ AutoDocs: Completed processing changelog: processed 89 merged pull requests
make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
[ 50%] Generating ../UsingEnergyPlusForCompliance.pdf
This is XeTeX, Version 3.14159265-2.6-0.99999 (TeX Live 2019/dev/Debian) (preloaded format=xelatex)
 restricted \write18 enabled.
entering extended mode
This is XeTeX, Version 3.14159265-2.6-0.99999 (TeX Live 2019/dev/Debian) (preloaded format=xelatex)
 restricted \write18 enabled.
```
Check out the make warning there in the middle.  This is causing the docs to be built serially instead of in parallel, and since they are made for each package type, it's causing package builds to get really slow.

That said, it is very nice to have them built intentionally with the documentation target.  And since we don't built package very often, it's not like adding 15 minutes is *that* big of a deal.  But I'd like to see if you have any thoughts on that issue and how you could maybe fix the issue and get the package build time at least back to reasonable before considering this complete.

jmarrec:
@Myoldmopar I hadn't seen this since I use ninja and not make. Ninja has one difference compared to make: `ninja` is equivalent to `ninja -j$(nproc)` while `make` is equivalent to `make -j1`. So it takes a minute on my machine. I'll try to see what I can do for make.

(I would suggest you consider switching to ninja on Unix (mac + ubuntu) regardless)

jmarrec:
@Myoldmopar Tried about everything I could think of and find. None really worked, so I settled for this: If generator=make, then specify -j`nproc` directly. This isn't ideal because it ignores the "-jN" option actually passed, but there's no way to get that one, and the latex generations doesn't take too long so it's probably fine.

Myoldmopar (approved):
Your latest changes make sense to me, and the build is now working great, with parallel doc builds at package time.  Cut down on the build time dramatically.  This can go in.  

(Note to self: remember that this is a change to the CMake rules, so if for _some_ reason CI package builds act funny, this is one potential, but extremely unlikely, culprit.)

jmarrec:
Follow up in https://github.com/NREL/EnergyPlus/pull/7468

PR #7338, opened 2019-06-17
Fix #7328 - Crash if AFN zone doesn't have inlet or return node connection

### LINKED ISSUE #7328: Crash if AFN zone doesn't have inlet or return node connection
Issue overview
--------------

If an airflow network zone (AirflowNetwork:MultiZone:Zone) is defined, and the AFN:SimulationControl object is using "MultiZoneWithDistribution", but the corresponding Zone has no inlet or return node connections defined in ZoneHVAC:EquipmentConnections, EnergyPlus will crash. After assigning an inlet or return node connection, there is a severe error message telling the user to hook up the necessary AFN objects (AFN:Distribution:Node, AFN:Distribution:Linkage, AFN:Distribution:Component:Duct, etc.). There should be a similar severe error letting the user know that for "MultiZoneWithDistribution", zones must have inlet or return node connections.

Attached defect file is a slightly modified version of the AirflowNetwork_MultiAirLoops.idf example file. The user added a baseboard heater to the unconditioned attic zone (new objects at top of file). Their goal was to simulate a zone that only has a baseboard heater (no conditioned supply air), then mixes air into an adjacent zone before returned back to air system.

[AirflowNetwork_MultiAirLoops MOD.txt](https://github.com/NREL/EnergyPlus/files/3261983/AirflowNetwork_MultiAirLoops.MOD.txt)

### Details
 - Version 9.1.0

### Checklist
 - [x] Defect file added: [EnergyPlusDevSupport - 7328](https://github.com/NREL/EnergyPlusDevSupport/blob/master/DefectFiles/7000s/7328/AirflowNetwork_MultiAirLoops.MOD.idf)

jmarrec:
Pull request overview
---------------------

Fix #7328 - Crash if AFN zone doesn't have inlet or return node connection

Initially I was throwing a severe (than a fatal) but realized it wasn't actually needed and @lgu1234 was kind enough to confirm this. I caught another potential hard crash during testing too.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -9754,8 +9754,14 @@ namespace AirflowNetworkBalanceManager {
                 for (j = 1; j <= NumOfZones; ++j) {
                     if (!ZoneEquipConfig(j).IsControlled) continue;
                     if (ZoneEquipConfig(j).ZoneNode == i) {
-                        NodeFound(i) = true;
-                        AirflowNetworkNodeData(ZoneEquipConfig(j).ActualZoneNum).EPlusNodeNum = i;
+                        if (ZoneEquipConfig(j).ActualZoneNum > AirflowNetworkNumOfNodes) {
```
Myoldmopar, line 9757: OK, so if we find a zone with this matching node number but it's not an AFN distribution node then error out.  Got it.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -9754,8 +9754,14 @@ namespace AirflowNetworkBalanceManager {
...
+                            ShowSevereError(RoutineName + NodeID(i) + " is not defined as an AirflowNetwork:Distribution:Node object.");
```
Myoldmopar: If more work is to be done on this PR I'd suggest quoting the `NodeID(i)` since it could be named in a variety of ways and may look funny in part of the sentence.  If no more work is needed I won't require that change.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -9855,12 +9863,12 @@ namespace AirflowNetworkBalanceManager {
...
             NodeFound.deallocate();
 
-            // Asign AirLoop Number to every node and linkage
+            // Assign AirLoop Number to every node and linkage
             // Zone first
             for (i = 1; i <= AirflowNetworkNumOfZones; i++) {
                 for (j = 1; j <= NumOfZones; j++) {
                     if (!ZoneEquipConfig(j).IsControlled) continue;
-                    if (MultizoneZoneData(i).ZoneNum == j) {
+                    if ((MultizoneZoneData(i).ZoneNum == j) && (ZoneEquipConfig(j).NumInletNodes > 0)) {
```
Myoldmopar, line 9871: And so we don't need any ELSE blocks to accommodate this change in logic?  We don't need (If zoneNum is J but numInletNodes is zero)?

tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
```diff
@@ -15280,4 +15282,152 @@ TEST_F(EnergyPlusFixture, TestAFNFanModel)
...
+    AirflowNetwork::AirflowNetworkNodeData(1).EPlusZoneNum = 1;
+
+    AirflowNetworkBalanceManager::SplitterNodeNumbers.allocate(2);
+    AirflowNetworkBalanceManager::SplitterNodeNumbers(1) = 0;
+    AirflowNetworkBalanceManager::SplitterNodeNumbers(2) = 0;
+
+
+    // MixedAir::NumOAMixers.allocate(1);
+
+    ASSERT_NO_THROW(ValidateDistributionSystem());
```
Myoldmopar: Hmm, is the assert_no_throw wrapper necessary?

tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
```diff
@@ -15280,4 +15282,152 @@ TEST_F(EnergyPlusFixture, TestAFNFanModel)
...
+    DataZoneEquipment::ZoneEquipConfig(1).NumInletNodes = 0;
+    DataZoneEquipment::ZoneEquipConfig(1).NumReturnNodes = 0;
+    DataZoneEquipment::ZoneEquipConfig(1).IsControlled = true;
+
+    ASSERT_THROW(ValidateDistributionSystem(), std::runtime_error);
+
+    std::string const error_string = delimited_string({
+        "   ** Severe  ** ValidateDistributionSystem: ATTIC ZONE AIR NODE is not defined as an AirflowNetwork:Distribution:Node object.",
+        "   **   ~~~   ** This Node is the zone air node for Zone 'ATTIC ZONE'.",
+        "   ** Severe  ** ValidateDistributionSystem: ATTIC ZONE AIR NODE is not defined as an AirflowNetwork:Distribution:Node object.",
```
Myoldmopar: Why two different severes?  What leads to that?

Myoldmopar (commented):
I think this is close, if not done, but I'd like to converse a little to make sure I understand the changes and impact fully.

mjwitte:
@Myoldmopar @jmarrec @jasondegraw These new unit tests fail when all unit tests are run serially (on Windows). Looks like this [`OneTimeFlag`](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L9655) may need to be lifted up and added to `clear_state`? There are lots of other one-time-flags lurking in AFN as well.

Myoldmopar:
I encountered a unit test failure on my schedule manager refactor branch and just temporarily disabled it so I could keep working.  @jmarrec after you address the installer issues, can you tackle this next?  If not I'll take a pass at it soon.

jasondegraw:
I'm against creating another global. One possible workaround is to move the flag up into ManageAirflowNetworkBalance and skip even making the call to the validation function after the first pass through. It's not great, but personally I'd revert this all out before creating yet another another global.

jmarrec:
So @mjwitte has the correct identification of the problem, this flag is definitely false after the first time a test hits it.

@jasondegraw, is the pattern of setting a bool **but inside an anonymous namespace** (not defined as extern in header) ok in this case?  The pattern I'm talking about can be seen as can be seen here for eg: https://github.com/jmarrec/EnergyPlus/blob/develop/src/EnergyPlus/PoweredInductionUnits.cc#L152:L157

----

**Site Notes**:

* I notice there's a block of code in `ValidateDistributionSystem` that's actually not inside the `if OneTimeFlag` block. https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L10432:L10455
I'll have to separate this one out.
     * **Edit**: Actually I think I got thrown off by https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L10438.
IIUC, It appears that `FanFlow = Node(j).MassFlowRate` isn't being used, so this block itself just sets the MaxAirMassFlowRate, so it could be called only once instead of at each iteration. 

* `ValidateDistributionSystem` is also using static error flags, which isn't needed and causing other problems with serially run unit tests:
``` 
static bool ErrorsFound(false);
static bool IsNotOK(false);
static bool errFlag(false);
```

jmarrec:
cf #7430 for my proposed solution / review / comments.

Issue #7318, opened 2019-05-29
FluidCooler:SingleSpeed cannot autosize


mjwitte:
Issue overview
--------------
1. User file with autosized FluidCooler:SingleSpeed fails to reject any heat. Stepped through `SizeFluidCooler` and found that `SimpleFluidCooler(FluidCoolerNum).HighSpeedFluidCoolerUAWasAutoSized` is false [here](https://github.com/NREL/EnergyPlus/blob/e5e9977e455951877332892a2062c093dfee08d6/src/EnergyPlus/FluidCoolers.cc#L1228-L1229) which causes the fluid cooler capacity to be zero.

`GetFluidCoolerInput` needs [this code](https://github.com/NREL/EnergyPlus/blob/e5e9977e455951877332892a2062c093dfee08d6/src/EnergyPlus/FluidCoolers.cc#L453-L455) added above in the singlespeed input processing.

2. Also, setting the capacity method to NominalCapacity causes a fatal error if the UA value is not blank - that should be a warning or simply silently ignored.
```
   ** Severe  ** FluidCooler:SingleSpeed= "FLUIDCOOLER_SINGLESPEED 1". Nominal fluid cooler capacity has been specified and design fluid cooler UA is being autosized.
   **   ~~~   ** Design fluid cooler UA field must be left blank when nominal fluid cooler capacity performance input method is used.
   **  Fatal  ** Errors found in getting fluid cooler input.
```
3. This file fails in debug with a floating point underflow [here](https://github.com/NREL/EnergyPlus/blob/e5e9977e455951877332892a2062c093dfee08d6/src/EnergyPlus/HVACInterfaceManager.cc#L695). Yet another `exp(<largenegativenumber>)` issue? Could we trap just overflows and allow underflows to go through - would they be zero?

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus 9.2.0-e5e9977e45
 - Helpdesk ticket number 15178.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added to EnergyPlusDevSupport\DefectFiles - 
190517_FluidCooler_Original-MJW - see col CJ FLUIDCOOLER_SINGLESPEED 1:Cooling Tower Heat Transfer Rate [W](Each Call) - always zero. After fix, there should be non-zero values, similar to 190517_Original-MJW which uses a cooling tower.
[...]

PR #7354, opened 2019-06-21
Addressed FluidCooler:SingleSpeed UA value autosize problem

Nigusse:
Pull request overview
---------------------
Addressed issue #7318. Fixed FluidCooler:SingleSpeed UA value autosize problem and avoids fatal out when the "NominalCapacity" is specified as performance input method and non-zero UA value also specified.  Diffs are expected/justified based on this change.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
Addressed the three problems described in issue 7318: (1) UA value autosizing problem, (2) setting the capacity method to NominalCapacity causes a fatal error if the UA value is not blank. Now throws error message but the simulation continuous with the UA value set to zero. (3) "exp(largenegativenumber)" issue. Now, traps such case and uses a reduced form of the equation. Found this problem in two places [here](https://github.com/NREL/EnergyPlus/blob/e5e9977e455951877332892a2062c093dfee08d6/src/EnergyPlus/HVACInterfaceManager.cc#L693-L701) and [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/CondenserLoopTowers.cc#L6551-L6554).

Nigusse:
This branch is ready for review.

src/EnergyPlus/FluidCoolers.cc
```diff
@@ -715,7 +718,8 @@ namespace FluidCoolers {
                 }
                 ShowContinueError(
                     "Design fluid cooler UA field must be left blank when nominal fluid cooler capacity performance input method is used.");
-                ErrorsFound = true;
+                ShowContinueError("Design fluid cooler UA value will be reset to zero and the simulation continuous.");
+                SimpleFluidCooler(FluidCoolerNum).HighSpeedFluidCoolerUA = 0.0;
```
Myoldmopar: @Nigusse 

Since you are now allowing this code to continue without issuing a fatal, the severe error a couple lines above should be converted to a warning.

src/EnergyPlus/FluidCoolers.cc
```diff
@@ -2503,6 +2507,8 @@ namespace FluidCoolers {
 
     void clear_state()
     {
+        NumSimpleFluidCoolers = 0;
+        SimpleFluidCooler.clear();
```
Myoldmopar, line 1834: 👍 👍 👍

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -6549,8 +6552,13 @@ namespace CondenserLoopTowers {
             NumTransferUnits = UAactual / CapacityRatioMin;
             // calculate heat exchanger effectiveness
             if (CapacityRatio <= 0.995) {
-                effectiveness = (1.0 - std::exp(-1.0 * NumTransferUnits * (1.0 - CapacityRatio))) /
-                                (1.0 - CapacityRatio * std::exp(-1.0 * NumTransferUnits * (1.0 - CapacityRatio)));
+                Real64 Exponent = NumTransferUnits * (1.0 - CapacityRatio);
+                if (Exponent >= 700.0) {
```
Myoldmopar, line 6556: This 700 is documented in the engineering reference and/or elsewhere, right?

Myoldmopar: Also, can you reuse this evaluation?  Like actually use the exponent variable in the expression?  If you needed you could just set `Real64 Exponent = negative value` and then change the logic to `if (Exponent <= -700)` and then use the value directly inside `std::exp`.

src/EnergyPlus/HVACInterfaceManager.cc
```diff
@@ -691,13 +691,24 @@ namespace HVACInterfaceManager {
...
+                } else {
+                    TankFinalTemp = (LastTankOutletTemp - (MassFlowRate * Cp * TankInletTemp + PumpHeat) / (MassFlowRate * Cp)) *
+                                        std::exp(-(MassFlowRate * Cp) / (ThisTankMass * Cp) * TimeStepSeconds) +
+                                    (MassFlowRate * Cp * TankInletTemp + PumpHeat) / (MassFlowRate * Cp);
+
+                    TankAverageTemp = ((ThisTankMass * Cp) / (MassFlowRate * Cp) *
+                                           (LastTankOutletTemp - (MassFlowRate * Cp * TankInletTemp + PumpHeat) / (MassFlowRate * Cp)) *
+                                           (1.0 - std::exp(-(MassFlowRate * Cp) / (ThisTankMass * Cp) * TimeStepSeconds)) / TimeStepSeconds +
+                                       (MassFlowRate * Cp * TankInletTemp + PumpHeat) / (MassFlowRate * Cp));
+                }
```
Myoldmopar, line 710: This is really messy.  I wonder if there is anything you could evaluate to clean this up.  Maybe make a couple intermediate variables?

`Real64 const mdotCp = MassFlowRate * Cp;`

That alone may reduce the visual complexity of this code.  If there is another obvious choice, that would be great, but I understand it's a balance.

tst/EnergyPlus/unit/FluidCoolers.unit.cc
```diff
@@ -244,3 +244,78 @@ TEST_F(EnergyPlusFixture, SingleSpeedFluidCoolerInput_Test3)
...
+        "     autosize;                !- Design Air Flow Rate Fan Power {W}",
+    });
+
+    ASSERT_TRUE(process_idf(idf_objects));
+
+    GetFluidCoolerInput();
+    auto &thisFluidCooler = FluidCoolers::SimpleFluidCooler(FluidCoolerNum);
+    EXPECT_TRUE(thisFluidCooler.HighSpeedFluidCoolerUAWasAutoSized);
+    EXPECT_EQ(thisFluidCooler.HighSpeedFluidCoolerUA, DataSizing::AutoSize);
+    EXPECT_EQ(thisFluidCooler.FluidCoolerNominalCapacity, 0.0);
```
Myoldmopar, line 274: 👍 👍

tst/EnergyPlus/unit/FluidCoolers.unit.cc
```diff
@@ -244,3 +244,78 @@ TEST_F(EnergyPlusFixture, SingleSpeedFluidCoolerInput_Test3)
...
+    AlphArray(4) = "NominalCapacity";
+    thisFluidCooler.FluidCoolerNominalCapacity = 5000.0;
+    thisFluidCooler.HighSpeedFluidCoolerUA = 500.0;
+    thisFluidCooler.HighSpeedFluidCoolerUAWasAutoSized = false;
+    // test input error check, if the nominal capacity specified and UA value is not zero, then it does not fatal out
+    bool testResult = TestFluidCoolerSingleSpeedInputForDesign(cCurrentModuleObject, AlphArray, cNumericFieldNames, cAlphaFieldNames, FluidCoolerNum);
+    EXPECT_FALSE(testResult); // no error message triggered
+    EXPECT_EQ(thisFluidCooler.PerformanceInputMethod_Num, PIM_NominalCapacity);
+    // UA value is reset to zero if nominal capacity is specified and input method is "NominalCapacity"
+    EXPECT_EQ(thisFluidCooler.HighSpeedFluidCoolerUA, 0.0);
```
Myoldmopar, line 320: Compact, efficient unit tests, these are awesome @Nigusse !

Myoldmopar (changes requested):
Requesting a couple minor cleanups; overall this is looking great!

src/EnergyPlus/FluidCoolers.cc
Nigusse: Changed the Severe Error message to Warning Error message.

src/EnergyPlus/CondenserLoopTowers.cc
Nigusse: The 700 is not documented in the eng. ref. This value is used else where in the code to avoid underflow/overflow.  I think there is an issue for this limit #7233 and addressed in PR #7414. Now, @lgu1234 proposed 70.0 as the limit instead of 700.0.

Nigusse: Modified the code to use the Exponent local variable directly in the equations.

src/EnergyPlus/HVACInterfaceManager.cc
Nigusse: Implemented suggestion by defining local variables.

Nigusse:
@Myoldmopar Addressed your review comments and resolved merge conflict.

Myoldmopar (approved):

Myoldmopar:
Comments are addressed.  CI is happy.  Defect file runs.  Built locally with updated develop branch and ran unit tests successfully.  This is good to go in.

PR #7356, opened 2019-06-21
Fix #7346 - AirLoop/Facility Component Load Summary can apply conversion to IP twice

### LINKED ISSUE #7346: EnergyPlus is converting from SI to IP twice in the eplustbl.htm for some areas
Using EnergyPlus v9.1.0 via OpenStudio 2.8.0 I find that in parts of the eplustbl.htm output, the SI to IP conversion happens twice when the following object is added to the IDF:

OutputControl:Table:Style,
  All,                                    !- Column Separator
  InchPound;                              !- Unit Conversion

I have attached both idfs, as well as the htm reports generated for convenience. An example of where it can be seen is in the AirLoop Component Load Summary in the heating peak conditions table for "DOAS NONEHWCHWHWVV" (sorry about the name). The SI version has an OADB of -17.4C, which should be 0.68F, but rather it is reported as 33.22F, which is the Celsius to Farenheit conversion again (0.68 * 1.8 + 32). It can also be seen in the airflows.

[IP-SI.zip](https://github.com/NREL/EnergyPlus/files/3303385/IP-SI.zip)

jmarrec:
Pull request overview
---------------------

Fix #7346 - AirLoop/Facility Component Load Summary can apply conversion to IP twice

The problem was that the Zone Peaks are calculated before, and the IP conversion is already applied.

When you check the AirLoop Load Summary, you check the peak for each zone at the loop's peak, but if the loop's peak happens to be the same as the zone peak, you don't compute it, you just reuse what was previously calculated (which has the conversion done already). Then you combine all the zone for the loop, and apply the conversion again, so potentially twice.

Added unit test + fix which is to basically recompute all the zone if IP even if loop's peak is same as zone's peak.

Found a typo in the Facility one that was comparing loop's time at peak to the zone design day num, so facility should be a lot less affected than Airloop.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@JasonGlazer I am trying to recollect the iteration call a week ago.  Did you volunteer to take a look at this at some point?  It's only a bug fix, not a new feature, so there's not a crazy urgent push, but I'm trying hard to clean out the massive pull request list.  If I'm just dreaming all that then I'll dig in and take a look!

JasonGlazer:
@Myoldmopar yep, I took this one on. Haven't done anything yet. It was probably code that I originally wrote that was broken so I wanted to take a look.

JasonGlazer:
@jmarrec The change that you are making would mean that the GetDelaySequences() and ComputeTableBodyUsingMovingAvg() would be always called anytime the user chose IP units. Those are time-consuming routines and that is why the logic is there to try to avoid them when the day and timestep match. I would suggest instead that the LoadSummaryUnitConversion() calls are put near the end or else that an unconverted version of the array is stored. Could you take another pass at fixing this?

jmarrec:
@JasonGlazer See latest commit where I have tried to address your comments.

JasonGlazer:
@Myoldmopar and @mjwitte this fix changes the order of the ComponentLoadSummary tables. It used to be the order Zone, AirLoop, Facility and now it is AirLoop, Facility, Zone. How big of a deal is that?

Myoldmopar:
I would hope it's not a big deal at all.  Users should not be relying on an exact order of things and instead finding them more carefully than something as fragile as that.  We can add it to the output rules document.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -13011,7 +13012,8 @@ namespace OutputReportTabular {
                 if (!ZoneEquipConfig(iZone).IsControlled) continue;
                 mult = Zone(iZone).Multiplier * Zone(iZone).ListMultiplier;
                 if (mult == 0.0) mult = 1.0;
-                if (displayZoneComponentLoadSummary && (timeCoolMax == ZoneCoolCompLoadTables(iZone).desDayNum) &&
+                // The ZoneCoolCompLoadTables already hasn't gotten a potential IP conversion yet, so we won't convert it twice.
```
JasonGlazer, line 13015: Thanks for catching this timeCoolMax vs coolDesSelected typo. It might make a big difference.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -13097,6 +13099,20 @@ namespace OutputReportTabular {
...
+                if (!ZoneEquipConfig(iZone).IsControlled) continue;
+                if (allocated(CalcFinalZoneSizing)) {
+                    LoadSummaryUnitConversion(ZoneCoolCompLoadTables(iZone));
+                    LoadSummaryUnitConversion(ZoneHeatCompLoadTables(iZone));
+
+                    OutputCompLoadSummary(zoneOuput, ZoneCoolCompLoadTables(iZone), ZoneHeatCompLoadTables(iZone), iZone);
+                }
+            }
+        }
+
```
JasonGlazer, line 13115: Thanks for structuring the code this way so that those time consuming routines are called for any IP unit

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -13512,14 +13528,14 @@ namespace OutputReportTabular {
...
+                // Outside Dry Bulb Temperature
                 compLoad.outsideDryBulb = CalcFinalZoneSizing(zoneIndex).CoolOutTempSeq(timeOfMax);
 
-                // Outside  Wet Bulb Temperature
+                // Outside Wet Bulb Temperature
                 // use standard sea level air pressure because air pressure is not tracked with sizing data
                 if (CalcFinalZoneSizing(zoneIndex).CoolOutHumRatSeq(timeOfMax) < 1.0 &&
                     CalcFinalZoneSizing(zoneIndex).CoolOutHumRatSeq(timeOfMax) > 0.0) {
-                    compLoad.outsideWebBulb = PsyTwbFnTdbWPb(CalcFinalZoneSizing(zoneIndex).CoolOutTempSeq(timeOfMax),
+                    compLoad.outsideWetBulb = PsyTwbFnTdbWPb(CalcFinalZoneSizing(zoneIndex).CoolOutTempSeq(timeOfMax),
```
JasonGlazer, line 13538: Funny how a typo just proliferates when you copy and paste it :-)
Thanks for fixing these.

JasonGlazer (approved):

JasonGlazer:
When running the unit tests locally, I'm getting a failure of AFN_CheckMultiZoneNodes_NoZoneNode but it also fails when on run the unit tests on develop. I asked @mjwitte and he said @jmarrec and @Myoldmopar were aware of the problem. Given this. This is ready to merge.

Issue #7048, opened 2018-11-06
Construction:WindowEquivalentLayer using WindowMaterial:SimpleGlazingSystem as a material fatals out silently


mjwitte:
Issue overview
--------------
User file has a `Construction:WindowEquivalentLayer` which references a `WindowMaterial:SimpleGlazingSystem` as a material layer. The run fatals out with the following err output:
```
Program Version,EnergyPlus, Version 9.0.1-bb7ca4f0da, YMD=2018.11.06 16:36,
   ************* Warning:  Node connection errors not checked - most system input has not been read (see previous warning).
   ************* Fatal error -- final processing.  Program exited before simulations began.  See previous error messages.
   ************* EnergyPlus Warmup Error Summary. During Warmup: 0 Warning; 0 Severe Errors.
   ************* EnergyPlus Sizing Error Summary. During Sizing: 0 Warning; 0 Severe Errors.
   ************* EnergyPlus Terminated--Fatal Error Detected. 0 Warning; 0 Severe Errors; Elapsed Time=00hr 00min  0.08sec
```

More info would be useful here to inform the user that SimpleGlazingSystem is not valid here. Use `Construction`.

### Details
Some additional details for this issue (if relevant):
 - Win64
 - Version of EnergyPlus v9.0.1
 - Helpdesk ticket number 13102

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Nigusse:
@mjwitte I added a check for valid Layers Type for Construction:WindowEquivalentLayer object. But when I run the defect in develop (V9.2), JSON Validator catches the invalid layer  and throws the following error before it gets to CONSTRUCTION object getinput function:

   ** Severe  ** <root>[Construction] - Object contains properties that could not be validated using 'properties' or 'additionalProperties' constraints
   ** Severe  ** <root>[Construction] - Object name is required and cannot be blank or whitespace
   **  Fatal  ** Errors occurred on processing input file. Preceding condition(s) cause termination.

I tested the defect file using V9.0 and I duplicated the error message that you posted. I am wondering why JSON did not catch it before in V9.1. Is it worth fixing the code to check for layers if JSON already does it?

mjwitte:
[...]

PR #7357, opened 2019-06-21
Construction:WindowEquivalentLayer using invalid layer material fatal out with error message

Nigusse:
Pull request overview
---------------------
Addresses issue #7048. Invalid material layers in Construction:WindowEquivalentLayer is checked and if found throws appropriate severe error message before fatal out. Diffs are not expected based on this change.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)


src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -4502,6 +4502,17 @@ namespace HeatBalanceManager {
...
                     ErrorsFound = true;
                 } else {
+                    MaterialLayerGroup = Material(Construct(TotRegConstructs + ConstrNum).LayerPoint(Layer)).Group;
+                    if (!((MaterialLayerGroup == GlassEquivalentLayer) || (MaterialLayerGroup == ShadeEquivalentLayer) ||
+                          (MaterialLayerGroup == DrapeEquivalentLayer) || (MaterialLayerGroup == BlindEquivalentLayer) ||
+                          (MaterialLayerGroup == ScreenEquivalentLayer) || (MaterialLayerGroup == GapEquivalentLayer))) {
+                        ShowSevereError("Invalid material layer type in window " + CurrentModuleObject + " = " +
+                                        Construct(TotRegConstructs + ConstrNum).Name);
+                        ShowContinueError("...Window layer material type = " + ConstructAlphas(Layer) +
+                                          " is not allowed in Construction:WindowEquivalentLayer window object.");
```
mjwitte, line 4508: @Nigusse This is good, except `ConstructAlpahs(Layer)` is the material name, not the material type, so the error message is a little confusing:
```
   ** Severe  ** Invalid material layer type in window Construction:WindowEquivalentLayer = SERRAMENTI PVC FINESTRE DOPPIA CAMERA 7CM
   **   ~~~   ** ...Window layer material type = VETRO DOPPIA CAMERA 4-9-4 NON TRATTATO SUPERFIC. is not allowed in Construction:WindowEquivalentLayer window object.
```
Maybe just delete "material type" above. And add "Only materials of type Material:*:EquivalentLayer are allowed."

mjwitte (commented):
Tested this with modified defect file and it now traps the illegal material with a meaningful error message. Ran unit tests - all OK. Please fix the error message as noted below and merge in develop and then this can go in once CI comes back all green.

src/EnergyPlus/HeatBalanceManager.cc
rraustad: But then the user has to identify which material is wrong. Small effort I know but wouldn't it be better to point to it?

mjwitte: My suggestion was to keep the name, just delete two words:
** ... Window layer ~material type~ = VETRO VETRO DOPPIA CAMERA 4-9-4 NON TRATTATO SUPERFIC. is not allowed ....
**  Only materials of type Material:*:EquivalentLayer are allowed.

Myoldmopar:
OK, I updated the error message per @mjwitte request.  I took the defect file and ran it with both develop and this branch.  As expected, with the degree symbol it fails in both with an incorrectly "blank" name.  When I clean up the object name, on develop the results go wild, while on this branch I get a nice error message:

```
   ** Severe  ** Invalid material layer type in window Construction:WindowEquivalentLayer = SERRAMENTI PVC FINESTRE DOPPIA CAMERA 7CM
   **   ~~~   ** ...Window layer = VETRO DOPPIA CAMERA 4-9-4 NON TRATTATO SUPERFIC. is not allowed in Construction:WindowEquivalentLayer window object.
   **   ~~~   ** Only materials of type Material:*:EquivalentLayer are allowed
   **  Fatal  ** GetSurfaceData: Errors discovered, program terminates.
```

I'm pushing up my changes and merging in.

PR #7360, opened 2019-06-24
Changing sequential load fractions to schedules

nmerket:
Pull request overview
---------------------
This implements Option A of [this NFP](https://github.com/NREL/EnergyPlus/blob/0762639f813cfcb1083a56c6bbca89a4d5ce2acd/design/FY2019/SequentialLoadFractionScheduleEMS.md). It changes the Sequential Load Fractions on `ZoneHVAC:EquipmentList` into schedules. That allows them to be seasonally altered or actuated via EMS. 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
   - [x] If transition, add to input rules file for interfaces
   - [x] If transition, add transition source
   - [x] If transition, update idfs
 - [x] ExpandObjects changes?


src/Transition/CreateNewIDFUsingRulesV9_2_0.f90
```diff
@@ -480,6 +485,28 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
               ! If your original object starts with W, insert the rules here
 
               ! If your original object starts with Z, insert the rules here
+              CASE('ZONEHVAC:EQUIPMENTLIST')
+                nodiff = .false.
```
nmerket, line 513: My FORTRAN is a bit rusty, but in my testing this worked. You might want to verify I'm doing all this right.

.gitignore
```diff
@@ -48,6 +48,7 @@ cmake-build-*
 .idea
 .se
 .vs
+.vscode
```
mjwitte, line 51: @Myoldmopar I'm assuming this change is OK?

tst/EnergyPlus/unit/UnitaryHybridUnitTest_DOSA.idf
```diff
@@ -579,14 +579,14 @@
     MUNTERSEPX5000,          !- Zone Equipment 1 Name
     1,                       !- Zone Equipment 1 Cooling Sequence
     2,                       !- Zone Equipment 1 Heating or No-Load Sequence
-    ,                        !- Zone Equipment 1 Sequential Cooling Fraction
-    ,                        !- Zone Equipment 1 Sequential Heating Fraction
+    ,                        !- Zone Equipment 1 Sequential Cooling Load Fraction Schedule Name
+    ,                        !- Zone Equipment 1 Sequential Heating Load Fraction Schedule Name
```
mjwitte, line 583: Very minor - Cooling ~Load~ Fraction ... Only if other changes are needed. And below. I've fixed these.

src/Transition/CreateNewIDFUsingRulesV9_2_0.f90
```diff
@@ -480,6 +485,28 @@ SUBROUTINE CreateNewIDFUsingRules(EndOfFile,DiffOnly,InLfn,AskForInput,InputFile
...
+                  ELSE IF (MOD((CurField - 2) - 6, 6) == 0) THEN
+                    zeqHeatingOrCooling = 'Heating'
+                  ELSE
+                    zeqHeatingOrCooling = 'Neither'
+                  END IF
+                  IF (InArgs(CurField) /= Blank .AND. (zeqHeatingOrCooling == 'Cooling' .OR. zeqheatingOrCooling == 'Heating')) THEN
+                    zeqNum = (CurField - 3) / 6 + 1
+                    WRITE(zeqNumStr, *) zeqNum
+                    OutArgs(CurField) = TRIM(InArgs(1)) // ' ' // zeqHeatingOrCooling // 'Frac' // TRIM(ADJUSTL(zeqNumStr))
+                    CALL WriteOutIDFLines(DifLfn,'Schedule:Constant',3,[OutArgs(CurField), '', InArgs(CurField)],['Name                     ', 'Schedule Type Limits Name', 'Hourly Value             '],['', '', ''])
```
mjwitte: This is well-crafted. I've mode a few modifications. Transitioned file comes out almost exactly the same. Added a schedule type object to keep the warnings down.

mjwitte (approved):
Ran unit tests serially.
Ran transitioned example files.
Perfectionist-ized the transition code.
Added transition rules to input rule doc.
Merged in develop.
This should drop in once CI come back as green as the hulk.

src/EnergyPlus/DataZoneEquipment.cc
```diff
@@ -1848,6 +1877,17 @@ namespace DataZoneEquipment {
...
 
+    Real64 EquipList::SequentialHeatingFraction(const int equipNum)
+    {
+        return ScheduleManager::GetCurrentScheduleValue(SequentialHeatingFractionSchedPtr(equipNum));
+    }
+
+    Real64 EquipList::SequentialCoolingFraction(const int equipNum)
+    {
+        return ScheduleManager::GetCurrentScheduleValue(SequentialCoolingFractionSchedPtr(equipNum));
+    }
```
mjwitte, line 1888: @nmerket @Myoldmopar This is all ready to go in except for the small performance degradation.

Performance Test Time | 339774256113 | 339348886410 | -0.1253487841083034%
-- | -- | -- | --

For the default case where `SequentialCooling/HeatignFraction` was a simple value of 1.0, it's now a 2-layer function call out to `GetCurrentScheduleValue`.
Too small to worry about? 
Let the schedule manager refactor recover some of the performance?
Split the difference and put an if in these functions to return 1.0 if the `...SchedPtr = ScheduleAlwaysOn`?
Other?

nmerket: We can live with the 0.1% performance degradation. My bet is that the slow down comes from looking looking up schedule values instead of reading a single value off a struct. I'd be surprised if the extra layer of function call makes that much of a difference.

src/Transition/CreateNewIDFUsingRulesV9_2_0.f90
nmerket: Thanks for fixing my iffy transition code.

src/EnergyPlus/DataZoneEquipment.cc
Myoldmopar: A 0.1% performance degradation may not seem that important, but we are looking at trying to scrape every bit of performance we can, and small steps backwards can add up.

That said, I'm not sure where we get if we add the IF logic in there to see if it is always on.  We may still have a performance degradation of like 0.05%.

In this particular case, I'd say let it go in because of the schedule manager refactor work.

Issue #7345, opened 2019-06-18
ASHWAT_Thermal not converging causes simultaneous heating and cooling load error


aaron-boranian:
Issue overview
--------------
The EquivalentLayerWindow.idf example file showcases how to use various `WindowMaterial:###:EquivalentLayer` objects combined into assemblies with the `Construction:WindowEquivalentLayer` object. The `WindowMaterial:Gap:EquivalentLayer` objects are all set to "Sealed" for the **Gap Vent Type** input field. The example file is set up for only simulating design days, but also running the annual run period yields no warnings or errors. Changing the innermost gap equivalent layers to use the "VentedIndoor" option instead of "Sealed" also runs without warnings or errors. However, when the "VentedOutdoor" option is applied to the outermost gap equivalent layers, the annual simulation fails (no warnings or errors over the design days). 

The first severe error is related to [ASHWAT_Thermal radiation analysis not converging](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/WindowEquivalentLayer.cc#L5469), but that message doesn't stop the simulation. The final severe error is related to an [unanticipated combination of heating and cooling loads](https://github.com/NREL/EnergyPlus/blob/f43e577cc4735e48577bdca46a73cfd09404f746/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L4298), which is the message that does stop the simulation. However, various components of the message are set to "NAN" and don't give much help to the user. This seems like a bug - perhaps divide by zero somewhere results in NAN and triggers this message?

Defect IDF and resulting ERR file after running a simulation with Atlanta EPW are attached. There is also an [Unmet Hours post](https://unmethours.com/question/38902/naturally-vented-double-skin-glazing-model/) with the same series of severe errors using the equivalent layer objects.

On this topic, it would be helpful if the time stamp of when the ASHWAT_Thermal error occurs was included in the error file.

### Details
Some additional details for this issue (if relevant):
 - Version 9.1.0 

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [X] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

[...]

PR #7363, opened 2019-06-26
Addressed equivalent layer window open channel model converging problem

Nigusse:
Pull request overview
---------------------
Addresses issue #7345. Corrected how the interior layer inside facing (facing the room) temperature is calculated for the sealed and open channel cavity model. The fix eliminated the convergence problem that led to simultaneous cooling and heating operation type error. The fix was verified against sealed channel equivalent layer window model and WIN6 model. Diffs are expected/justified based on this change. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


src/EnergyPlus/DataWindowEquivalentLayer.hh
```diff
@@ -232,15 +232,16 @@ namespace DataWindowEquivalentLayer {
...
+        int NL;                   // number of layers
+        Array1D<CFSLAYER> L;      // layer array, L(1) is outside layer
+        Array1D<CFSGAP> G;        // gap array, G(1) is outside-most, betw L(1) and L(2)
+        bool ISControlled;        // CFS is not controlled, or has no controlled VB layer
+        int VBLayerPtr;           // Venetian blind layer pointer
+        int WEQLSolverErrorIndex; // recurring error index
 
         // Default Constructor
-        CFSTY() : NL(0), L(CFSMAXNL), G(CFSMAXNL - 1), ISControlled(false), VBLayerPtr(0)
+        CFSTY() : NL(0), L(CFSMAXNL), G(CFSMAXNL - 1), ISControlled(false), VBLayerPtr(0), WEQLSolverErrorIndex(0)
```
Myoldmopar, line 244: No problem here, just another member variable.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -446,8 +446,8 @@ namespace WindowEquivalentLayer {
         if (CFSHasControlledShade(CFS(EQLNum)) > 0) CFS(EQLNum).ISControlled = true; // is controlled
     }
 
-    void CalcEQLWindowUvalue(CFSTY const &FS, // CFS to be calculated
-                             Real64 &UNFRC    // NFRC U-factor, W/m2-K
+    void CalcEQLWindowUvalue(CFSTY &FS,    // CFS to be calculated
```
Myoldmopar: Why can't this be const?  I don't see changes here that would restrict that away, but maybe I'm missing something.  Please clarify.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -1025,12 +1022,12 @@ namespace WindowEquivalentLayer {
...
         //  Solve energy balance(s) for temperature at each node/layer and
         //  heat flux, including components, between each pair of nodes/layers
-        ASHWAT_ThermalR = ASHWAT_Thermal(
-            CFS(EQLNum), TIN, Tout, HcIn, HcOut, TRMOUT, TRMIN, 0.0, QAllSWwinAbs({1, NL + 1}), TOL, QOCF, QOCFRoom, T, Q, JF, JB, H, UCG, SHGC);
-        // long wave radiant power to room not including reflected
-        QRLWX = JB(NL) - (1.0 - LWAbsIn) * JF(NL + 1);
-        // nominal surface temp = effective radiant temperature
-        SurfInsideTemp = TRadC(QRLWX, LWAbsIn);
+        ASHWAT_ThermalR =
+            ASHWAT_Thermal(CFS(EQLNum), TIN, Tout, HcIn, HcOut, TRMOUT, TRMIN, 0.0, QAllSWwinAbs({1, NL + 1}), TOL, QOCF, QOCFRoom, T, Q, JF, JB, H);
```
Myoldmopar: Two arguments gone.  OK, I like that though I am curious why they were needed before but aren't now.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -1025,12 +1022,12 @@ namespace WindowEquivalentLayer {
...
+
+        // effective surface temperature is set to surface temperature calculated
+        // by the fenestration layers temperature solver
+        SurfInsideTemp = T(NL) - KelvinConv;
```
Myoldmopar, line 1031: So QRLWX is just not set anymore?  I need to see exactly what TRadC was doing.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -5006,7 +5003,7 @@ namespace WindowEquivalentLayer {
         }
     }
 
-    bool ASHWAT_Thermal(CFSTY const &FS,  // fenestration system
+    bool ASHWAT_Thermal(CFSTY &FS,        // fenestration system
```
Myoldmopar: Alright, another removal of const, and I see that below in this function you are assigning the value of the error index.  That's fine, but what I can't see (from this diff) is why the `FS` variable can't be constant.  It looks like the first argument being passed into here is called `CFS(EQLNum)`, not `FS` in the function above.  I think I'm still just missing something because the diff is only showing part of the story.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -5023,8 +5020,8 @@ namespace WindowEquivalentLayer {
                         Array1A<Real64> JF,           // returned: front (outside facing) radiosity of surfaces, W/m2
                         Array1A<Real64> JB,           // returned: back (inside facing) radiosity, W/m2
                         Array1A<Real64> HC,           // returned: gap convective heat transfer coefficient, W/m2K
-                        Real64 &UCG,                  // returned: center-glass U-factor, W/m2-K
-                        Real64 &SHGC,                 // returned: center-glass SHGC (Solar Heat Gain Coefficient)
+                        Optional<Real64> UCG,         // returned: center-glass U-factor, W/m2-K
+                        Optional<Real64> SHGC,        // returned: center-glass SHGC (Solar Heat Gain Coefficient)
```
Myoldmopar: I'm pretty opposed to the use of optionals here.  Is there a reason they can't just be passed in dummy values?  Or if this function is used for two separate purposes, maybe you can find the common code and break it off into a separate one that's called from two different workers.  Then you only have to call the function with the arguments relevant to your current use case.

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -5489,6 +5496,9 @@ namespace WindowEquivalentLayer {
         // New code follows from here - for calculating Ucg and SHGC
         // NOTE: This code can be bypassed if
         //       indices of merit are not needed
+        if (!(present(UCG) && present(SHGC))) {
```
Myoldmopar: I'm not sure how `IM_ON` got set before, but can you do it without requiring optionals?

Myoldmopar (changes requested):
I'm a little confused on parts of this, I'll have to look outside of GitHub's diff view for clarification.  If you can provide some commentary that would be helpful.  And I'd like you to find a way to achieve your changes without requiring the Optional variables.

Myoldmopar:
I just checked the EquivalentLayerWindow test file diffs.  There is a time where the `ZN001:WALL001:WIN003:Surface Inside Face Temperature [C](Hourly)` value is off by as much as 5°C.  I think this will need further justification.  

image

Do you have the results comparisons with WIN6?  @vidanovic are you available to review this change?

vidanovic:
@Myoldmopar this seems to be ASHWAT model which unfortunately we do not have in WINDOW and also I still did not implement that in Windows-CalcEngine as well. Yes, 5 degrees difference on window surface is huge. One thing to note is that the biggest difference is during solar radiation which indicates some big changes there. My availability these days is close to zero due to some deadlines coming up in October, but maybe I can get some time to take a look into this. Will let you know?

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: These two variables (UCG and SHGC) are rated window performance values. They do not need to be calculated for each time step. That is why I made them optional arguments.

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: QRLWX was used to re-calculate the inside surface temperature. I was hesitant about the use this variable when I first implemented the model. The inside surface temperature is already calculated using layer by layer thermal balance function ASHWAT_Thermal() and has average room radiant boundary temperature as an input to this function. I always thought that QRLWX calculation was unnecessary step. It is the re-calculation of the inside temperature using QRLWX that resulted in unrealistic inside surface temperature and lead to simultaneous heating and cooling situation for this defect. That is why I removed it.

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: The only reason that I removed the constant was I cannot update the error index variable (WEQLSolverErrorIndex) that I added for this fix for reporting recurring errors. If I can pass as const and still be able to update the error index variable, I am happy to to do it.

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: Let me think If I can avoid requiring optional variables.

Nigusse:
ZN001:WALL001:WIN003 ASHWAT model has six solid layers and I don't think I can create WIN6 equivalent for that. This makes the the verification difficult. But I created WIN6 equivalent of  ZN001:WALL001:WIN001, which has two 6mm clear glass with 12 mm air gap. See below the summer and winter days inside surface temperature comparison: ASHWAT before and after fix and WIN6 model.

EQLWIN_and_WIN6_Comparison

Nigusse:
Added WIN6 equivalent of the defect file. Only one of the three window files was modified.

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: Broke off ASHWAT_Thermal() function into two functions and eliminated the two optional arguments.

src/EnergyPlus/WindowEquivalentLayer.cc
Nigusse: Removed the optional arguments.

Myoldmopar (approved):

Myoldmopar:
Merging this in conjunction with the other equivalent layer changes.  Thanks @Nigusse !

PR #7370, opened 2019-07-02
addressed low surface temperatures for WindowEquivalentLayer with interior blind

### LINKED ISSUE #7367: Very low surface temperatures for WindowEquivalentLayer with interior blind
Issue overview
--------------
Comparisons between Winkelmann and the WindowEquivalentLayer (ASHWAT) methods for a window with interior blinds show large differences. From the user's post:

> Transmitted solar radiation is consistent for both models but the equivalent layer approach gives extremely low indoor surface temperatures (lower than indoor and outdoor temp), no radiative exchange and very high convective exchange.

See [here for plots](https://unmethours.com/question/38948/inconsistency-in-results-from-winkelman-and-equivalent-layer-blinds-models/).

### Details
 - Version of EnergyPlus v9.1
 - Unmethours [38948](https://unmethours.com/question/38948/inconsistency-in-results-from-winkelman-and-equivalent-layer-blinds-models/)
### Checklist
 - [x] Defect files added - v9.1 files for horizontal 2cm blinds and v8.8 files for wide vertical blinds are in EnergyPlusDevSupport\DefectFiles. 

Nigusse:
Pull request overview
---------------------
Fixes #7367. The low surface temperature is caused by zero thermal emissivity value set when the thermal transmittance input field of a blind is left blank. Furthermore, the effective thermal emissivity calculated is not properly updated for use in the fenestration assembly layer by layer heat balance calculation. Diffs are expected/justified based on this change.

### Work Checklist
   - Refactoring: This pull request includes code changes that don't change the functionality of the 

### Review Checklist
 - [ ] Unit Test(s)


src/EnergyPlus/HeatBalanceIntRadExchange.cc
```diff
@@ -301,6 +301,11 @@ namespace HeatBalanceIntRadExchange {
                                 (ShadeFlagPrev != IntBlindOn && ShadeFlag == IntBlindOn) ||
                                 (ShadeFlagPrev == IntShadeOn && ShadeFlag != IntShadeOn) || (ShadeFlagPrev == IntBlindOn && ShadeFlag != IntBlindOn))
                                 IntShadeOrBlindStatusChanged = true;
+                            if (SurfaceWindow(SurfNum).WindowModelType == WindowEQLModel &&
```
Myoldmopar, line 309: Is there something funny about the indentation here?  I'll have to check it locally because on GitHub it looks like it is off.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
```diff
@@ -323,6 +328,10 @@ namespace HeatBalanceIntRadExchange {
                             HeatBalanceMovableInsulation::EvalInsideMovableInsulation(SurfNum, HMovInsul, AbsInt);
                             zone_info.Emissivity(ZoneSurfNum) = Material(Surface(SurfNum).MaterialMovInsulInt).AbsorpThermal;
                         }
+                        if (surface_window.WindowModelType == WindowEQLModel &&
+                            DataWindowEquivalentLayer::CFS(Construct(ConstrNum).EQLConsPtr).ISControlled) {
+                            zone_info.Emissivity(ZoneSurfNum) = EQLWindowInsideEffectiveEmiss(ConstrNum);
+                        }
```
Myoldmopar, line 338: OK, so setting the emissivity was just completely skipped for the equivalent layer window. Yikes. And OK.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -3583,8 +3583,10 @@ namespace HeatBalanceManager {
                 Material(MaterNum).ReflFrontDiffDiffVis = MaterialProps(14);
                 Material(MaterNum).ReflBackDiffDiffVis = MaterialProps(15);
             }
-            if (!lNumericFieldBlanks(19) && !lNumericFieldBlanks(20) && !lNumericFieldBlanks(21)) {
+            if (!lNumericFieldBlanks(19)) {
                 Material(MaterNum).TausThermal = MaterialProps(19);
+            }
```
Myoldmopar, line 3616: I agree with splitting this out, makes good sense.  How about the thermal front and back emissivity though?  Does this same problem happen if just one of them is provided in input and the other is blank?

src/EnergyPlus/WindowEquivalentLayer.cc
```diff
@@ -444,6 +444,9 @@ namespace WindowEquivalentLayer {
         CalcEQLWindowStandardRatings(ConstrNum);
 
         if (CFSHasControlledShade(CFS(EQLNum)) > 0) CFS(EQLNum).ISControlled = true; // is controlled
+
+        // set internal face emissivity
+        Construct(ConstrNum).InsideAbsorpThermal = EffectiveEPSLB(CFS(EQLNum));
```
Myoldmopar, line 449: Yes, I can see this would be important...

Myoldmopar (changes requested):
It looks like at a minimum there is an indentation problem to be addressed.  Functionally it seems reasonable.  I'm building and running locally.

Myoldmopar:
This is showing even more drastic differences in window surface temperature.  While I understand that this was a blatant error in the code, and this fixes it, we'll need to see more detailed comparisons to ensure this is a step in the right direction.

These changes may need to be combined with the other equivalent layer window PR to ensure the results are expected once they are both merged down.

src/EnergyPlus/HeatBalanceManager.cc
Nigusse: Split the check for front and back thermal emissivity.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
Nigusse: Correct.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
Nigusse: I do not see the issue after merging.

Nigusse:
Resolved merge conflict and split thermal emissivity get input check. Successfully built locally.

Myoldmopar:
I'm going to test locally where I pull in this branch as well as #7363 and compare the combination of them against develop.  I want to see the combined effects of these two changes to the equivalent layer model.

Myoldmopar:
I ran that little experiment of pulling both branches in and looking at the diffs.  As expected, the equivalent layer file is the only one with diffs.  However, those diffs are really really big now.

image

I'm really uncomfortable with this much difference.  I understand the code is making it better, but it would be really great to have some more justification.  @Nigusse can you pull the two branches together and run the equivalent layer file and do some deeper analysis of the surface temperature?  It's going from 6 degrees to 16 degrees overnight.  Is it more physical of a temperature given the heat rates and other surface temperatures and boundary conditions?

I really want to get this in, but I need to be confident it's the right set of changes.

Nigusse:
@Myoldmopar Sure will look into it.

Nigusse:
OK, I pulled in both branches (Issues #7345 and #7367) locally, fixed merge conflict and investigated the defect file deeply. My finding is summarized as follows:

I looked at the WIN3 inside face temperature for winter and summer design days. Note that WIN3 has six solid layers. The plot below shows inside face temperature before and after the fix including the outside air and zone air temperatures.  Justifications for the fix:

- If you look at the winter design day (01/21) results, the before and after the fix WIN3 inside face temperatures are bounded by the outside air and zone air temperature (expected for no solar input hours). Furthermore, the after fix inside face temperature is more biased to the zone air temperature as expected for six-solid-layer window. 

- If you look at the summer design day (07/21) the before the fix WIN3 inside face temperatures are NOT BOUNDED by outside air and zone air temperature for off solar hours. This violates laws of thermodynamics hence the before the fix results are incorrect. However, for the after the fix WIN3 inside face temperatures are BOUNDED by outside air and zone air temperatures during off solar hours (this is what it should be). Also note that the after the fix WIN3 inside face temperatures during summer steadily approaches the zone air temperature during the evening hours. The after the fix WIN3 inside face temperature is behaving correctly. 

DesignDaysWindowInsideFaceTemperature

Nigusse:
@Myoldmopar Again I looked at the WIN3 inside face temperature for ZERO SOLAR summer design day. The plot below shows inside face temperature before and after the fix including the outside air and zone air temperatures. Justifications for the fix:

- the winter design day (01/21) results are the same as the previous plot, so no need repeat the explanation here. 

- If you look at the summer design day (07/21) with zero solar input. The solar was turned off deliberately to check whether the heat balance holds or not. The before the fix WIN3 inside face temperatures are NOT BOUND by outside air and zone air temperature for all day. This clearly shows the  before the fix results violates thermodynamics law. However, for the after the fix WIN3 inside face temperatures are BOUND by outside air and zone air temperatures during the entire day (as it should be). Also note that the after the fix WIN3 inside face temperatures during summer with zero solar input are very close to the zone air temperature as expected. Again the after the fix results are behaving correctly.

ZeroSolarDesignDaysWindowInsideFaceTemperature

Nigusse:
@Myoldmopar This branch includes the fix for issue #7367 as well.

rraustad:
That's a great explanation and shows it's working correctly. It would be nice to see inside and outside face temps on that last plot (both before and after).

Nigusse:
@Myoldmopar @rraustad WIN3 outside and inside faces temperature for winter and ZERO SOLAR summer design days. The plot below shows the OUTSIDE and INSIDE face temperatures before and after the fix including the outside air, zone air and sky temperatures. Justifications for the fix:

- If you look at the summer design day (07/21) with zero solar input, the before and after the fix WIN3 outside face temperatures are similar and consistent trend because this fix does not impact the outside temperature calculation. But the very small differences observed between before and after the fix outside face temperature can be attributed to a secondary effect from the change in inside face temperature. For most part in summer design day (07/21) they are BOUND by outside air and zone air temperatures but in early morning hours during summer the outside face temperature drops below the outside and zone air temperature due to colder sky temperature as shown. In general the outside face temperature is driven by the outside air temperature when there is no solar.

- If you look at the winter design day (07/21) both outside face temperature slightly drops below the outside air temperature due to colder sky temperature (and when the outside and zone air temperatures are very close). The sky temperature is actually much colder than the outside air temperature (Atlanta Georgia Weather). This explains why the WIN3 outside face temperatures during winter (01/21) slightly drifts below the outside air temperature. Again the after the fix results are behaving correctly.

ZeroSolarDesignDaysWindow_OutsideFace_and_InsideFaceTemperatures

Myoldmopar (approved):

Myoldmopar:
This investigation is *so* good.  Thank you for pushing that through @Nigusse.  I feel very confident in these changes now and will merge these equivalent layer window changes in directly.

Nigusse:
I also compared EquivalentLayer and Winkelmann (BuiltInWindowsModel) window models for two pane clear glazing. I modified the Zn001:Wall001:Win001 of the defect file to create Winkelmann windows equivalent model. The after fix equivalent window model and the Winkelmann window model show good match for outside and inside face temperatures. The before fix inside face temperature for some hours during summer is slightly lower (this the kind of defect stated in issue #7367). The plot below shows the various temperatures for zero solar input test case.

ZeroSolarDesignDaysWindow_OutsideFace_and_InsideFaceTemperatures_Intermadel

PR #7376, opened 2019-07-06
Addressed Fan models need to protect against negative fan power due to bad EMS override values

### LINKED ISSUE #7365: Fan models need to protect against negative fan power due to bad EMS override values
Issue overview
--------------
User file failed with a fatal psychrometric error due to very cold temperatures in the HVAC system. The problem comes from negative values set by EMS for fan pressure , which caused negative values for fan power and fan heat gain, leading to very cold node temperatures, which then cascaded until the psych error tripped.

All of the fan models need a trap for this. I'd be find with simply putting a max(power,0.0) around the first power values calculated in each fan model - just clip it silently without warning.

### Details
 - Version of EnergyPlus 9.1.0-08d2e308bb
 - Helpdesk ticket number 15206

### Checklist

Nigusse:
Pull request overview
---------------------
Fixes #7365 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:

### DIFF
--- a/src/EnergyPlus/HVACFan.cc
+++ b/src/EnergyPlus/HVACFan.cc
@@ -812,8 +812,8 @@ namespace HVACFan {
                             localFanTotEff = m_fanTotalEff;
                             locHiSpeedFanRunTimeFrac = locRunTimeFraction * locFlowRatio;
                             m_fanRunTimeFractionAtSpeed[0] += locHiSpeedFanRunTimeFrac;
-                            m_fanPower +=
-                                locHiSpeedFanRunTimeFrac * m_maxAirMassFlowRate * localPressureRise[mode] / (localFanTotEff * m_rhoAirStdInit);
+                            m_fanPower += max(
+                                0.0, locHiSpeedFanRunTimeFrac * m_maxAirMassFlowRate * localPressureRise[mode] / (localFanTotEff * m_rhoAirStdInit));
                         } else if (m_numSpeeds > 1) { // multi speed
 
                             // find which two speed levels bracket flow ratios and calculate runtimefraction at each speed
@@ -826,7 +826,7 @@ namespace HVACFan {
                                 locHiSpeedFanRunTimeFrac = locFlowRatio * locRunTimeFraction / m_flowFractionAtSpeed[0];
                                 m_fanRunTimeFractionAtSpeed[0] += locHiSpeedFanRunTimeFrac;
                             } else {
-                                lowSideSpeed = 0;  // hush up cppcheck
+                                lowSideSpeed = 0; // hush up cppcheck
                                 hiSideSpeed = 0;  // hush up cppcheck
                                 for (auto loop = 0; loop < m_numSpeeds - 1; ++loop) {
                                     if ((m_flowFractionAtSpeed[loop] <= locFlowRatio) && (locFlowRatio <= m_flowFractionAtSpeed[loop + 1])) {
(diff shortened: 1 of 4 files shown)


tst/EnergyPlus/unit/Fans.unit.cc
```diff
@@ -98,3 +98,156 @@ TEST_F(EnergyPlusFixture, Fans_FanSizing)
...
+    EXPECT_DOUBLE_EQ(Result_FanPower, thisFan.FanPower); // expects 300 W
+
+    // negative fan pressure rise set using EMS
+    thisFan.EMSFanPressureOverrideOn = true;
+    thisFan.EMSFanPressureValue = -300.0;
+    // simulate the fan with negative pressure rise
+    // set using fans EMS actuator for Pressure Rise
+    Fans::SimOnOffFan(FanNum);
+    Real64 Result2_FanPower = max(0.0, thisFan.MaxAirMassFlowRate * thisFan.EMSFanPressureValue / (thisFan.FanEff * thisFan.RhoAirStdInit));
+    EXPECT_DOUBLE_EQ(Result2_FanPower, thisFan.FanPower); // expects zero
```
Myoldmopar, line 194: :+1: makes sense based on the desired changes here.

src/EnergyPlus/HVACFan.cc
```diff
@@ -866,8 +868,8 @@ namespace HVACFan {
                             localFanTotEff = m_fanTotalEff;
                             locHiSpeedFanRunTimeFrac = locFanRunTimeFraction;
                             m_fanRunTimeFractionAtSpeed[0] += locHiSpeedFanRunTimeFrac;
-                            m_fanPower +=
-                                locHiSpeedFanRunTimeFrac * m_maxAirMassFlowRate * localPressureRise[mode] / (localFanTotEff * m_rhoAirStdInit);
+                            m_fanPower += max(
```
Myoldmopar, line 871: I understand the intent here, but seeing all these spread out has me concerned about a performance impact.  Is there a more central location we can check for this and have less max calls later on?  Probably not but I feel I should ask.

Nigusse: One central location for Fan:SystemModel check is [here](https://github.com/NREL/EnergyPlus/blob/167109939_Issue7365/src/EnergyPlus/HVACFan.cc#L976). But, will not work for "SpeedControlMethod::Continuous" due to the code [here](https://github.com/NREL/EnergyPlus/blob/167109939_Issue7365/src/EnergyPlus/HVACFan.cc#L939-L943) used to update fan outlet air condition.

The other possibility is to reset EMS fan pressure rise value [here](https://github.com/NREL/EnergyPlus/blob/167109939_Issue7365/src/EnergyPlus/HVACFan.cc#L751) but I do not know how significant would the performance impact be until we test it.

Nigusse: @Myoldmopar @mjwitte Thinking about it, adding a check for the EMS pressure rise only is executed only when the EMS variable EMSFanPressureOverrideOn is true. This reduces the computation effort, but negative fan power will not be captured if it happen for other reasons, for instance when the Fan(FanNum).EMSAirMassFlowValue is negative.

Myoldmopar:
This still looks like a good set of changes.  I'm going to pull the code down, merge up with develop and re-run unit tests just to be sure it's all set.

Myoldmopar:
Yep, everything looks good locally.  Merging.

PR #7380, opened 2019-07-09
Addressed FanCoil with CyclingFan control using Fan:SystemModel has different fan power than Fan:OnOff

### LINKED ISSUE #7086: FanCoil with CyclingFan control using Fan:SystemModel has different fan power than Fan:OnOff
Issue overview
--------------
Tried to switch HVACTemplate:Zone:FanCoil to always use Fan:SystemModel, but this caused diffs in HVACTemplate-5ZoneFanCoil-DOAS because this uses the CyclingFan control. The defect files below have more output and no DOAS, just fancoils with CyclingFan control.

### Details
 - Version of EnergyPlus v9.0.1
 - Reference #7062 

### Checklist
 - [x] Defect files added here:
[7086-HVACTemplate-5ZoneFanCoil-CyclingFan-exp-FanSystemModel.idf.txt](https://github.com/NREL/EnergyPlus/files/2641071/7086-HVACTemplate-5ZoneFanCoil-CyclingFan-exp-FanSystemModel.idf.txt)
[7086-HVACTemplate-5ZoneFanCoil-CyclingFan-exp-FanOnOff.idf.txt](https://github.com/NREL/EnergyPlus/files/2641072/7086-HVACTemplate-5ZoneFanCoil-CyclingFan-exp-FanOnOff.idf.txt)

Nigusse:
Pull request overview
---------------------
Fixes #7086. Fixed performance differences between a fan coil unit with cycling fan using Fan:OnOff and Fan:SystemMode. Now, fancoil units with cycling fan operating mode, will give the same result using either Fan:OnOff or Fan:SystemMode. Diffs are expected when a FanCoil unit uses  Fan:SystemModel fan type and the capacity control method is CyclingFan.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types

### DIFF
--- a/src/EnergyPlus/FanCoilUnits.cc
+++ b/src/EnergyPlus/FanCoilUnits.cc
@@ -3594,8 +3594,7 @@ namespace FanCoilUnits {
                                                 ZoneCompTurnFansOn,
                                                 ZoneCompTurnFansOff);
                 } else {
-                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(
-                        FanCoil(FanCoilNum).LowSpeedRatio, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
                 }
             } else if (FanCoil(FanCoilNum).SpeedFanSel == 2) {
 
@@ -3607,15 +3606,22 @@ namespace FanCoilUnits {
                                                 ZoneCompTurnFansOn,
                                                 ZoneCompTurnFansOff);
                 } else {
-                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(
-                        FanCoil(FanCoilNum).MedSpeedRatio, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
                 }
-            } else { // using 1.0 here for fan speed ratio seems wrong if FCU max flow rate is different than the fan maximum flow rate
+            } else if (FanCoil(FanCoilNum).SpeedFanSel == 3) {
+
                 if (FanCoil(FanCoilNum).FanType_Num != DataHVACGlobals::FanType_SystemModelObject) {
                     Fans::SimulateFanComponents(
                         FanCoil(FanCoilNum).FanName, FirstHVACIteration, FanCoil(FanCoilNum).FanIndex, 1.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
                 } else {
-                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(1.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                }
+            } else { // using 1.0 here for fan speed ratio seems wrong if FCU max flow rate is different than the fan maximum flow rate
+                if (FanCoil(FanCoilNum).FanType_Num != DataHVACGlobals::FanType_SystemModelObject) {
+                    Fans::SimulateFanComponents(
+                        FanCoil(FanCoilNum).FanName, FirstHVACIteration, FanCoil(FanCoilNum).FanIndex, 0.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
+                } else {
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(0.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
                 }
             }
             if (FanCoil(FanCoilNum).CCoilType_Num == CCoil_HXAssist) {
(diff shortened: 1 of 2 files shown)


Nigusse:
Diffs are expected in the example file "HVACTemplate-5ZoneFanCoil-DOAS.idf", because the FanCoil units has Fan:SystemModel fan type  and the capacity control method is CyclingFan.

Myoldmopar:
Yes, diffs are indeed expected and acceptable.  Testing locally now.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -3594,8 +3594,7 @@ namespace FanCoilUnits {
                                                 ZoneCompTurnFansOn,
                                                 ZoneCompTurnFansOff);
                 } else {
-                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(
-                        FanCoil(FanCoilNum).LowSpeedRatio, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
```
Myoldmopar, line 3597: Can you walk me through this change?  By not passing a speed ratio, does this force the fan object to operate at a single discrete mode?  Is that what fixes the fan to operate just like the Fan:OnOff?

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -3607,15 +3606,22 @@ namespace FanCoilUnits {
...
                 } else {
-                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(1.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(_, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
+                }
+            } else { // using 1.0 here for fan speed ratio seems wrong if FCU max flow rate is different than the fan maximum flow rate
+                if (FanCoil(FanCoilNum).FanType_Num != DataHVACGlobals::FanType_SystemModelObject) {
+                    Fans::SimulateFanComponents(
+                        FanCoil(FanCoilNum).FanName, FirstHVACIteration, FanCoil(FanCoilNum).FanIndex, 0.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff);
+                } else {
+                    HVACFan::fanObjs[FanCoil(FanCoilNum).FanIndex]->simulate(0.0, ZoneCompTurnFansOn, ZoneCompTurnFansOff, _);
```
Myoldmopar, line 3624: Is this change required for fixing this defect or is this an unrelated change?

tst/EnergyPlus/unit/FanCoilUnits.unit.cc
```diff
@@ -2883,4 +2883,407 @@ TEST_F(EnergyPlusFixture, FanCoil_CyclingFanMode)
...
+    ZoneSysEnergyDemand(1).RemainingOutputReqToCoolSP = -2500.0;
+    ZoneSysEnergyDemand(1).RemainingOutputRequired = -2500.0;
+    QZnReq = -2500.0;
+    Sim4PipeFanCoil(FanCoilNum, ZoneNum, ZoneNum, FirstHVACIteration, QUnitOut, QLatOut);
+    // expect fan speed 2 and moderate air and water flow and meet capacity
+    EXPECT_EQ(2, FanCoil(1).SpeedFanSel);
+    EXPECT_NEAR(FanCoil(1).PLR, 0.753, 0.001);
+    EXPECT_NEAR(QZnReq, QUnitOut, 5.0);
+    EXPECT_NEAR(Node(1).MassFlowRate, thisFanCoil.PLR * thisFanCoil.MaxAirMassFlow * thisFanCoil.MedSpeedRatio, 0.0000000001);
+}
```
Myoldmopar, line 3287: I'm not asking you to make a change to the unit test right now but I will say it would be nice to set up a unit test with both fan onoff and system model connected to identical but separate systems.  Then you could simulate both of them and check that the values are equal.  This would avoid having to hardwire expected values in the unit test.  Just a note for future tests like this.

Myoldmopar:
I just ran the OnOff and the SystemModel defect files with both develop and this branch.  I see the big difference in results in the develop branch between the two files.  With this branch I see the results come together for the two fan types.  I think this is ready to go in, but I'd still like a clarification of the changes on my review above so I fully understand it and so they are documented here for the future.

Thanks @Nigusse .

Myoldmopar:
(Also, I just ran unit test suite locally from this branch with develop merged in and it ran fine.)

src/EnergyPlus/FanCoilUnits.cc
Nigusse: We do not want to pass the fan speed ratio for cycling operating mode. FanCoilUnits cycling mode are designed such that the fan coil unit cycles at each speed level. Thus passing speed ratio will result incorrect fan power calculation. By not passing the fan speed ratio, the fansystem model locally calculates average flow fraction instead, for the fan power calculation.

src/EnergyPlus/FanCoilUnits.cc
Nigusse: If you referring to the last line where I set 0.0 for speed ratio, is to turn off the fan. Simple because the unit is off if no fan speed is specified.

tst/EnergyPlus/unit/FanCoilUnits.unit.cc
Nigusse: Ok. @Myoldmopar Thanks.

src/EnergyPlus/FanCoilUnits.cc
Nigusse: The fancoil unit is expected to operate at one of the three fan speed levels, if the fan speed is not set yet, then the fan coil unit must be off. Otherwise, what does the fancoil unit do if no fan speed set yet? I observed such a case when I was debugging the defect.

Myoldmopar:
Thanks for the clarifications @Nigusse.  Merging now.

PR #7382, opened 2019-07-11
Addressed Economizer BypassWhenOAFlowGreaterThanMinimum control needs a mass flow tolerance

### LINKED ISSUE #7379: Economizer BypassWhenOAFlowGreaterThanMinimum control needs a mass flow tolerance
Issue overview
--------------
Two very similar user files which differ only slighting in one value have very different HX bypass results on the winter design day. The OA controller has an enthalpy economizer with no lockout and BypassWhenOAFlowGreaterThanMinimum. Watching in the debugger shows that the OA flow is on the order of 2e-16 > than the minimum, so it locks out. Look at `OAMassFlowActual` and `OAMassFlowMin` in red below.

image

This code was modified in #6861 by @Nigusse  to address a similar issue, #6796 .

One solution would be to add a small tolerance to the > check?

`if (OAMassFlowActual > (OAMassFlowMin + SmallMassFlow) )`

Also, output variable "Air System Outdoor Air Heat Recovery Bypass Heating Coil Activity Status" is always zero in this file. Seems that it should be 1 when the heating coil is operating.

### Details
 - Win64, Windows 10
 - Version of EnergyPlus 9.2.0-26151c5477
 - Direct e-mail from PNNL, 03 Jul 2019

### Checklist
 - [x] Defect file added to EnergyPlusDevSupport\DefectFiles:
case_1-moreoutput.idf is the base case with some Kitchen HX operation during the early morning hours of the winter design day.
case_2-moreoutput.idf is the problem case where the Kitchen HX never operates on the winter design day.
case_2-moreoutput-WinterFirst - winter design day first to simplify debugging

Nigusse:
Pull request overview
---------------------
Fixes #7379. 
This fixes actual OA mass flow rate calculation in air economizer and added a small air mass flow a threshold (DataHVACGlobals::SmallMassFlow = 0.001 kg/s) when comparing actual OA mass flow with minimum OA mass flow for Heat Recovery Bypass Status determination. Also add a log that traps when the mixed air flow rate exceeds the design flow rate due to large exhaust air mass flow rate. This is a defect and diffs are expected/justified based on this change. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
@MJWitte While investigation the operation of Economizer, Heat Recover HX and Heating coil I found that the mixed air flow rate can be greater than the Design Supply Air Flow Rate. This happens when the zone exhaust air mass flow rate exceeds that design supply air flow rate and the offending code is [HERE](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3302).  At the results of this actual OA mass flow rate [HERE](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4707) exceeds the minimum OA mass flow rate and forces to bypass heat recovery heat exchanger. Now, the heat recovery HX is not bypassed when it should not.

Nigusse:
Looking at the code, the output variable "Air System Outdoor Air Heat Recovery Bypass Heating Coil Activity Status" is set always zero regardless of the heating coils operation. I could not find any where in the code that sets the heat exchanger status, the "*.HRHeatingCoilActive" variable is always set to zero. This needs further investigation.

mjwitte:
Hmmm, excess zone exhaust should *not* cause higher system flow. It's a bit circular, but ` thisOAController.ExhMassFlow = max(0.0, AirLoopFlow(AirLoopNum).SupFlow - AirLoopFlow(AirLoopNum).SysRetFlow)` and ` Node(thisOAController.RetNode).MassFlowRat` *should* be the same as `AirLoopFlow(AirLoopNum).SysRetFlow`.  Is `AirLoopFlow(AirLoopNum).SupFlow > AirLoopFlow(AirLoopNum).DesSupply`?

Nigusse:
I saw the Node(thisOAController.RetNode).MassFlowRate is not the same as AirLoopFlow(AirLoopNum).SysRetFlow when the thisOAController.MixMassFlow flow was greater than the AirLoopFlow(AirLoopNum).DesSupply. In fact the Node(thisOAController.RetNode).MassFlowRate was > 0 while the AirLoopFlow(AirLoopNum).SysRetFlow was 0 when the exhaust exceeded the design flow rate. But the AirLoopFlow(AirLoopNum).SupFlow is the same as AirLoopFlow(AirLoopNum).DesSupply.

mjwitte:
Hmmm, if zone exhaust exceeds supply, the I would expect zero return, and both of those values to be zero. What are the return node flows from the zones?

Nigusse:
See below, debug snapshot of the airloop serving kitchen. The return node flow rate is 1.6742286435554297 kg/s and the zone exhaust mass flow rate is 1.6742286435554301, so according the existing code (before the fix) the mixed air mass flow rate would have been 3.34845... kg/s. But as you can see the snapshot I capped it by the design flow rate. 

DebugSnipetForAirLoopServingKitchen

Nigusse:
I agree the zone return node mass flow rate should be zero when the zone exhaust flow rate exceed the design flow rate and that is the case when I look at the output variables. But in this code the zone return node flow rate changes and sometime it is non zero and pushes the mixed air mass flow rate above the design flow rate according to this equation [HERE](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3302). In return this causes the actual OA mass flow rate to exceed minimum OA mass flow rate [HERE](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L4707) and thereby turning the heat recovery HX bypass status ON when it should not.

Nigusse:
I am looking at the code regarding output variable "Air System Outdoor Air Heat Recovery Bypass Heating Coil Activity Status". The variable HRHeatingCoilActive is always 0. I do not see any where in the code that sets this variable to 1.0.  I am not sure since when this variables is always set to 0.  This must have been unintended change @mjwitte @rraustad

rraustad:
I'm fairly sure it used to work since I added that variable to help users identify why the HR bypass was off. I assume this report is not used that often.

Nigusse:
This defect file has kitchen exhaust mass flow rate greater than the air loop design supply air flow rate. Air economizer does not allow the OA mass flow rate to be less than the exhaust mass flow rate [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/MixedAir.cc#L3690-L3693) even if the economizer controller prefers to run at the minimum OA mass flow. As the result the OA mass flow rate is greater than the minimum flow and the heat recovery is bypassed and makes the Heating Coil Active. Also, now the fix sets the HRHeatingCoilActive to 1.0 if the "*.HeatingActiveFlag" variable is true (i.e., any heating coil in the air loop is active).

Nigusse:
There are ten example files that failed due to large diffs and reviewed a couple of them that has large diffs. I did not find any unexpected operation. This branch is ready for review.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3782,6 +3786,8 @@ namespace MixedAir {
                 } else {
                     curAirLoopControlInfo.ResimAirLoopFlag = false;
                 }
+            } else if (curAirLoopControlInfo.HeatingActiveFlag) {
+                this->HRHeatingCoilActive = 1;
```
Myoldmopar, line 3818: Yeah, this should help that variable go to 1 sometimes.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3304,7 +3304,11 @@ namespace MixedAir {
                     // if flow rate has been specified by a manager, set it to the specified value
                     thisOAController.MixMassFlow = AirLoopFlow(AirLoopNum).ReqSupplyFrac * AirLoopFlow(AirLoopNum).DesSupply;
                 } else {
-                    thisOAController.MixMassFlow = Node(thisOAController.RetNode).MassFlowRate + thisOAController.ExhMassFlow;
+                    // found this equation results in flow that exceeds the design flow rate when there is exhaust flow rate is greater than
+                    // the design supply air flow rate. Capped the mixed air flow rate at design supply air flow rate, issue #77379
+                    // thisOAController.MixMassFlow = Node(thisOAController.RetNode).MassFlowRate + thisOAController.ExhMassFlow;
+                    thisOAController.MixMassFlow =
+                        min(Node(thisOAController.RetNode).MassFlowRate + thisOAController.ExhMassFlow, AirLoopFlow(AirLoopNum).DesSupply);
```
Myoldmopar: It looks like the issues around this calculation have been pretty well resolved based on conversation on the PR.  I'll do testing to verify the defect file is now functioning as intended.  @mjwitte or @rraustad, it would be great to get your approval on this cap on the mass flow rate.

src/EnergyPlus/MixedAir.cc
```diff
@@ -4710,7 +4716,7 @@ namespace MixedAir {
                 } else if (this->HeatRecoveryBypassControlType == BypassWhenOAFlowGreaterThanMinimum) {
                     Real64 OAMassFlowMin = OutAirMinFrac * AirLoopFlow(AirLoopNum).DesSupply;
                     Real64 OAMassFlowActual = OASignal * this->MixMassFlow;
-                    if (OAMassFlowActual > OAMassFlowMin) {
+                    if (OAMassFlowActual > (OAMassFlowMin + DataHVACGlobals::SmallMassFlow)) {
```
Myoldmopar: When I saw the correction was that a tolerance needed to be added, I assumed it was an equality.  This is usually a problem:

```
if (floatingPointA == floatingPointB) { ...
```

needs to be:

```
if (abs(floatingPointA - floatingPointB) < tolerance) { ...
```

But this is already an inequality.  I don't have a problem adding the padding on there, I guess I just wonder if you could allow for a padding on the other side of OAMassFlowMin?

Myoldmopar (commented):
Please respond to my comments.  I ran this locally with develop merged up and it passed all unit tests.  I still need to run the defect file(s), but I'd like to wrap up the commentary here first.

src/EnergyPlus/MixedAir.cc
mjwitte: I'm not sure - we had some conversation a while back but then I dropped the ball. I never expected this fix to be anything more than a one-liner to add a tolerance. I'm not sure if I'll have time to look at this again today/tomorrow.

src/EnergyPlus/MixedAir.cc
mjwitte: The problem that this is band-aid-ing is that OAMassFlowActual in the defect file was ever so slightly greater than OAMassFlowMin which triggered the economizer to behave differently when it shouldn't. Since the control decision is only for when OAMassFlowActual is meaningfully-greater-than OAMassFlowMin this is an appropriate padding.

mjwitte: Perhaps there's a better way to control the bypass than comparing mass flow rates - i.e. yet another bool flag? - that really indicates if the economizer really wants to economize with extra fresh air.

Myoldmopar:
FYI @mjwitte @lgu1234 we should discuss this today on the call.

Myoldmopar:
Thanks for the discussion on the call today.  It looks like there are two action items to get this immediately merged, and then we can continue discussions as needed later.

- `DataHVACGlobals::SmallMassFlow` seems too large here.  But VerySmallMassFlow is too small.  I will make a new local constant at 1e-6 that should suffice.
- The change to constrain the flow to the design supply flow seems to overstep things here.  I will pull that out of here.

I am going to make those changes and rebuild locally.  I'll then test a defect file I have here to ensure this still fixes it.  I'll also run a full regression and unit test suite locally to make sure we're still good there.  Then this can merge in.  After that, we can open a new issue to continue discussion.

@Nigusse let me know if you have any issues with me doing these tasks to get this merged in right away.

Myoldmopar:
Made those two fixes locally:

- Defect file I have is still addressed by the changes
- Unit test suite still passes
- Regressions running locally now

Myoldmopar:
I grabbed a few of the files that were showing big diffs in that earlier commit and ran them through regressions first.  No big diffs.  I think we'll be in good shape.  I'm quickly running the full regression suite and I'll also push my commit.  If everything is good then this will go in.

Myoldmopar:
In my regression results, there are no files with any math diffs.  However, there are 6 files showing big table diffs:

- CoolingCoilFreezingPrevention
- HeatPumpwithBiquadraticCurves
- HeatPumpVRP_DCV
- HeatPumpAirToAirWithRHcontrol
- UnitarySystem_MultiSpeedCoils_SingleMode
- HP_wICSSolarCollector

The only diff in the table is the "Heat Emission by Components" table, and the only column with a difference is the "HVAC Reject Heat" value.  But it's significant.  Taking HeatPumpAirToAirWithRHcontrol as an example, in the develop branch the value is 5.31 GJ, while in this branch it is zero.  As another example, with the UnitarySystem_MultiSpeedCoils_SingleMode file, in develop branch the value is -39099999.5 GJ, but in this branch the value is zero.  

I understand diffs might be expected, but is it expected that these values will go fully to zero?

Myoldmopar:
Hmm... CI is happier about this branch than my regression tool.  Curious.

Myoldmopar:
OK, only diffs are due to the PVWatts branch getting merged into develop.  If this continues to be clean I'll merge it tonight.

Myoldmopar:
I'm feeling confident.  I'm going to merge this now.  I'll leave the branch behind again so CI will run Windows on it just in case.

Issue #7285, opened 2019-05-01
Unit ventilator control problems


mjwitte:
Issue overview
--------------
ZoneHVAC:UnitVentilator controls do not work as advertised.

1. When there is no coil, VariablePercent is supposed to increase the OA to max when there is a cooling load and the outdoor air is colder than the zone. But the if logic is backwards for this case.

2. Where there is a cooling coil, VariablePercent currently stays at minimum OA when there is a cooling load, as described in the I/O Reference. But this isn't really useful - with VariablePercent, the OA should increase to provide free cooling if there is a cooling load and conditions are favorable.

3. If the user really wants a fan coil unit with economizer, the I/O Ref suggests using the unit ventilator. It would be better to just add some rudimentary economizer controls to the fan coil unit, because the fan coil unit has better control options in general.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus 9.1.0
 - Helpdesk ticket number 15147.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

Nigusse:
@mjwitte @Myoldmopar defect items 1 and 2 can be addressed here, but item 3 needs a separate issue, or even may be enhancement. What do you think?

RKStrand:
@Nigusse I agree with you.  Items 1 and 2 are defect issues while Item 3 is a separate thing--more of a development project.  By the way, are you thinking of taking this on?  If not, I am interested in working on this.

Nigusse:
@RKStrand You can have it as you are more familiar with this object.

RKStrand:
Thanks, Bereket. I’ll try to look into this more next week.


On Jul 5, 2019, at 1:34 PM, Bereket Nigusse <notifications@github.com<mailto:notifications@github.com>> wrote:


@RKStrand<https://github.com/RKStrand> You can have it as you are more familiar with this object.

RKStrand:
@mjwitte Do you agree with @Nigusse and @RKStrand that this should be separated into two separate issues (items 1 and 2 in your original problem statement stay here and item 3 gets moved to something else, possible a new feature)?

RKStrand:
@mjwitte So, in item 1, you wrote:

[...]

PR #7386, opened 2019-07-12
Unit Ventilator Variable Fraction OA Control Correction and Fix Fan Mode Schedule with Fan:SystemModel

RKStrand:
Pull request overview
---------------------
This work fixes an issue with the unit ventilator variable fraction outside air control.  Previously, unit ventilator, when in cooling mode would toggle between the minOA and maxOA fraction without a cooling coil and with a cooling coil would simply default to the minOA.  This does not allow appropriate variation of the OA fraction.  Both the with and without cooling coil options during cooling mode have now been adjusted to allow any variation between the minOA and maxOA fraction to meet the cooling load.  When the outside air temperature is above the zone return temperature, the OA fraction is still set to the minimum as before.  What is new is the ability to vary the OA fraction when the outside air temperature is cooler to meet the load.  Previously, it was possible to overcool the space with outside air.  This addresses GitHub Issue #7285 and includes both a new unit test and modified documentation.

Also, this fixes the Supply Fan Operating Mode Schedule to work correctly with Fan:SystemModel. Previously, the fan mode was always continuous regardless of the schedule values.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7285 

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes n/a


RKStrand:
@mjwitte I tagged you as a reviewer for this one since we have already had discussions about it.  There are small differences in three test cases--all ones that should see changes because they have unit ventilators that use variable percent OA control (which was the focus of this work).

src/EnergyPlus/UnitVentilator.cc
```diff
@@ -3991,6 +3982,35 @@ namespace UnitVentilator {
...
+        if (Tinlet <= Toutdoor) {
+            
+            ActualOAMassFlowRate = MinOAFrac * MassFlowRate;
+            
+        } else { // Tinlet > Toutdoor
+            // Use cooler outside air to provide "free" cooling without over-cooling.
+            // First, use a simple load equals mass flow times Cp time Delta T equation to find OA Mass Flow Rate.
+            // Then, limit the OA Mass Flow Rate between the MinOA flow and the MaxOA flow.
+            
+            ActualOAMassFlowRate = std::abs(QZnReq) / (PsyCpAirFnWTdb(DataEnvironment::OutHumRat,Toutdoor) * (Tinlet-Toutdoor));
```
mjwitte: Not sure why, but this is off by about 30W from what the cooling coil is doing. I'll attach a test idf that shows the cooling coil operating at about 30W whenever the OA flow is between min and max.
Oh - it's probably the fan heat?
Also, this probably needs to be sensitive to `OpMode == CycFanCycCoil` - then set to max?

src/EnergyPlus/UnitVentilator.cc
```diff
@@ -3048,22 +3048,11 @@ namespace UnitVentilator {
...
-                            } else { // Tinlet > Toutdoor
-
-                                MaxOAFrac = GetCurrentScheduleValue(UnitVent(UnitVentNum).MaxOASchedPtr);
-                                OAMassFlowRate = MaxOAFrac * Node(OutsideAirNode).MassFlowRate;
-                            }
-
+                            
+                            OAMassFlowRate = SetOAMassFlowRateForCoolingVariablePercent(MinOAFrac,Node(OutsideAirNode).MassFlowRate,
+                                                                                        GetCurrentScheduleValue(UnitVent(UnitVentNum).MaxOASchedPtr),
+                                                                                        Tinlet,Toutdoor);
```
mjwitte, line 3054: Nice to pull this out into a separate function.

mjwitte:
Here's my test file. Look at Unit Ventilator 1, columns BM:BN and CW:CX on 04/10 and 04/11.
[UnitVent5ZoneAuto-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt](https://github.com/NREL/EnergyPlus/files/3407550/UnitVent5ZoneAuto-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt)

src/EnergyPlus/UnitVentilator.cc
RKStrand: @mjwitte Before I make a change, what you are asking for is:

When OpMode == CycFanCycCoil, then set:

ActualOAMassFlowRate = MaxOAFrac * MassFlowRate

Correct?

src/EnergyPlus/UnitVentilator.cc
RKStrand: Thanks--makes unit testing cleaner as well.

src/EnergyPlus/UnitVentilator.cc
mjwitte: That was my initial thought. I'm fairly certain that's how the airloop economizer functions with cycling fan. Otherwise, you end up with two variables - fan runtime and OA fraction to mess with to meet the load. And I would expect you want to maximize the beneficial outdoor air.

mjwitte: But I'm not sure setting OAMassFlowRate works - You want 100% OA that cycles on enough to meet the load.

RKStrand: @mjwitte Elsewhere in this model, I am seeing the following code:

        if (FanOpMode == CycFanCycCoil) {
            OutAirMassFlowRate = min(OAMassFlowRate, Node(InletNode).MassFlowRate);
        }

which basically limits the outside air mass flow rate if the fan is cycling if I am understanding it correctly. Do you feel if I do something similar in the new function that this will resolve the concern?

mjwitte: Not sure - this has to be done at the appropriate point in the process in order for the mixing (if any) and cycling control to all work from the correct value.

RKStrand: But it seems like SimUnitVentOAMixer (which is where that reassignment that I showed in my previous comment) should be run and then it would take care of the cycling case?

mjwitte: Hopefully, test files will tell . . .

RKStrand:
@mjwitte So, looking at the test file that you supplied, what I think that I am seeing is that there is this leftover 30-ish Watt cooling coil load that hits for a while (presumably fan energy as you suggested?) and then finally the cooling coil starts actually producing some real numbers. On 4/10, it appears that before that 30W cooling coil load shows up that the system is running with min OA, meaning it has more than enough cooling to use outside air.  Then, once it starts to modulate, it isn't factoring in the fan heat so the cooling coil is forced to run a small fraction.  It varies the OA amount until the maximum at which point you see the cooling coil load go beyond 30W.  I'm seeing the same pattern on 4/11.  So, based on this, it seems like the problem here is that my new algorithm is not factoring in fan heat and this is causing some cooling coil load that isn't necessary.  So, I should factor the fan heat in and this should eliminate the 30W cooling coil loads.  Does this summarize the concern you had and the potential solution?  Or are there more issues?

mjwitte:
Yes, that summarizes it.

RKStrand:
@mjwitte Just committed changes which address the fan heat issue and also expand the unit test cases from 4-7 to test the modified code.  This should be ready for re-review when the CI machines finish doing their thing.

mjwitte:
@RKStrand The results for UnitVent5ZoneAuto-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt look good now. I tried a variation on this with the Unit Ventilator 1 Supply Fan Operating Mode Schedule set to always = 0 to test cycling fan operation and the results are identical to the continuous fan case. If I leave the fan operating mode schedule blank, then I get cycling fan operation and the control looks good. Can you check if the fan mode schedule is being used correctly to switch the fan mode every timestep?

RKStrand:
@mjwitte Here is the section of code that I think addresses your question:

        if (UnitVent(UnitVentNum).FanSchedPtr > 0) {
            if (GetCurrentScheduleValue(UnitVent(UnitVentNum).FanSchedPtr) == 0.0 && UnitVent(UnitVentNum).FanType_Num == FanType_SimpleOnOff) {
                UnitVent(UnitVentNum).OpMode = CycFanCycCoil;
            } else {
                UnitVent(UnitVentNum).OpMode = ContFanCycCoil;
            }
        }

This is done every timestep in InitUnitVentilator.  Does that answer your question fully?

mjwitte:
@RKStrand The fan type in the test file is Fan:SystemModel. This fan cycling if wasn't modified to allow to that fan type. Since the input processing checks for illegal combination of fan and cycling schedule, I would just remove the `&& UnitVent(UnitVentNum).FanType_Num == FanType_SimpleOnOff` here.

RKStrand:
@mjwitte Ok, I changed that line and made another commit.  Hopefully the CI comes back clean and this is now ready to go.

mjwitte (approved):
@RKStrand Tested the fan operating mode schedule fix - all looks good now.
Discovered a new problem while reviewing the cycling fan results, see #7420.

PR #7398, opened 2019-07-17
Provide correct outlet air properties

lgu1234:
Pull request overview
---------------------
The current calculation of outlet air properties are only correct at speed 1 when no load flow rate = supply heat air flow rate. When 0<PLR<1 and no load flow rate != supply heat air flow rate, the heating load value (PLR * Cap) is not equal to the value = mdot * (outlet h - inlet h), since no humidity changes at outlet. The fix provides correct calculations, so that a linear relationship between PLR and heating load is valid.

A test file is uploaded.  

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/HeatingCoils.cc
```diff
@@ -2342,9 +2342,13 @@ namespace HeatingCoils {
                     OutletAirHumRat = FullLoadOutAirHumRat;
                     OutletAirTemp = FullLoadOutAirTemp;
                 } else {
-                    OutletAirEnthalpy = PartLoadRat * FullLoadOutAirEnth + (1.0 - PartLoadRat) * InletAirEnthalpy;
-                    OutletAirHumRat = PartLoadRat * FullLoadOutAirHumRat + (1.0 - PartLoadRat) * InletAirHumRat;
-                    OutletAirTemp = PartLoadRat * FullLoadOutAirTemp + (1.0 - PartLoadRat) * InletAirDryBulbTemp;
+                    OutletAirEnthalpy =
```
Myoldmopar, line 2345: It looks like this is protected above for nonzero AirMassFlow variable, but I don't see in this GitHub diff exactly where it is protected from zero HeatingCoil(CoilNum).InletAirMassFlowRate.  I'll need to check it out locally.

src/EnergyPlus/HeatingCoils.cc
```diff
@@ -2342,9 +2342,13 @@ namespace HeatingCoils {
...
+                        PartLoadRat * AirMassFlow / HeatingCoil(CoilNum).InletAirMassFlowRate * (FullLoadOutAirEnth - InletAirEnthalpy) +
```
Myoldmopar, line 2346: So this was calculated as `h_out = (PLR)*h_out_full_load + (1-PLR)*h_in`, which essentially scales the enthalpy as a weighted fraction of outlet enthalpies for the time when the unit was on or off based on part load.  This seems to make sense if it was runtime fraction, but not with PLR.

This is now calculated as `h_out = h_in + PLR * (AirMassFlow/InletAirMassFlow) * (h_out_full_load - h_in)`, which is more of an enthalpy balance application and seems to make more sense.

tst/EnergyPlus/unit/HeatingCoils.unit.cc
```diff
@@ -107,4 +112,58 @@ TEST_F(EnergyPlusFixture, HeatingCoils_FuelTypePropaneGas)
...
+    HeatingCoils::CalcMultiStageGasHeatingCoil(CoilNum, 0.0, 1.0, 1, 2);
+    Real64 HeatLoad10 =
+        HeatingCoils::HeatingCoil(CoilNum).InletAirMassFlowRate *
+              (Psychrometrics::PsyHFnTdbW(HeatingCoils::HeatingCoil(CoilNum).OutletAirTemp, HeatingCoils::HeatingCoil(CoilNum).OutletAirHumRat) -
+               HeatingCoils::HeatingCoil(CoilNum).InletAirEnthalpy);
+    EXPECT_NEAR(HeatLoad10, HeatingCoils::HeatingCoil(CoilNum).HeatingCoilLoad, 0.0001);
+
+    // check linear relationship at PLR = 0.5
+    EXPECT_NEAR(HeatLoad05, 0.5 * HeatingCoils::HeatingCoil(CoilNum).MSNominalCapacity(1), 0.0001);
+}
```
Myoldmopar, line 167: A very nice unit test!

Myoldmopar (approved):
Will build and test locally, but this looks good so far.

Myoldmopar:
Pulled in develop locally, built fine, unit tests run fine.  CI looks pretty happy, one file with modest diffs (2W).  I'm merging this.

PR #7414, opened 2019-07-25
Fix AFN floating point underflow

lgu1234:
Pull request overview
---------------------
A global epexp function was generated. It set very small value to zero, so that floating underflow using debug is fixed.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


lgu1234:
@mjwitte @jasondegraw 

I created epexp function in General, suggested by Jason, and replace std::exp() by General::exexp().
    Real64 epexp(Real64 x)
    {
        if (x < -708.0) {
            return 0.0;
        }
        return std::exp(x);
    }

When I tested AirflowNetwork_MultiAirLoops.idf using the debug exe, the program still crashed shown below:

Capture

The main reason is that the small value (e-305) from a new function times a small value (e-10) generated  a denormalized value.

Here is my test code in the unit test:

    x = -700.0;
    Real64 z = 1.0e-15 * std::exp(x);  

std::exp(x) = 9.8596765437597708e-305 and z = 9.859574028408e-320#DEN. That was why the program crashed.

In order to avoid underflow, I made the maximum value is -70, instead of -708. The value of std::exp(-70) = 1.0806392777072785e-30, which is small enough for E+ to be considered as zero. Here is my modification:

    Real64 epexp(Real64 x)
    {
        if (x < -70.0) {
            return 0.0;
        }
        return std::exp(x);
    }

I ran AFN examples file. No diffs were found. There is no crash using the debug exe.

Please let me know if my change is OK with you or not. My code revision only replaced std::exp() in AFN. Do you think I need to change other modules?

jasondegraw (approved):
These changes work for me.

src/EnergyPlus/General.cc
```diff
@@ -3922,6 +3922,23 @@ namespace General {
...
+    {
+        if (x < -70.0) {
+            return 0.0;
+        }
+        return std::exp(x);
+    }
+
+    Real64 epexp(Real64 x, Real64 defaultHigh)
+    {
+        if (x < -708.4) {
```
mjwitte: @lgu1234 I'm confused here. Why does this form use -708.4 as the lower limit while the one above uses -70.0? Also, the unit test does not exercise this form.

jasondegraw: If I recall correctly, we discussed having one function with both high and low tests. So this second one is more for completeness and potential use outside AFN. Could also do it like this:

```c++
    Real64 epexp(Real64 x, Real64 defaultLow=-70.0)
    {
        if (x < defaultLow) {
            return 0.0;
        }
        return std::exp(x);
    }

    Real64 epexp(Real64 x, Real64 defaultHigh, Real64 defaultLow=-70.0)
    {
        if (x < defaultLow) {
            return 0.0;
        } else if (x > defaultHigh) {
            return std::exp(defaultHigh);
        }
        return std::exp(x);
    }
```

But it probably should be the same if we leave it as is.

lgu1234: @mjwitte This is my mistake. The value should be -70.0 in both functions. In addition, I will add a unit test for the second function.

@jasondegraw I like your suggestion. However, I got compile problem, because polymorphism may not allow different types of arguments: overloaded in the first equation, and normal in the second equation. Then, I changed both equations slightly with argument order:
    Real64 epexp(Real64 x, Real64 defaultLow=-70.0)
    {
        if (x < defaultLow) {
            return 0.0;
        }
        return std::exp(x);
    }

    Real64 epexp(Real64 x, Real64 defaultLow=-70.0, Real64 defaultHigh = 700.0)
    {
        if (x < defaultLow) {
            return 0.0;
        } else if (x > defaultHigh) {
            return std::exp(defaultHigh);
        }
        return std::exp(x);
    }
Visual Studio still gives me an error.

Since we agree -70 for low end, I am going to hard-wired the value in both functions:

    Real64 epexp(Real64 x)
        if (x < -70.0) {
            return 0.0;
        }
        return std::exp(x);
    }

    Real64 epexp(Real64 x, Real64 defaultHigh)
    {
        if (x < -70) {
            return 0.0;
        } else if (x > defaultHigh) {
            return std::exp(defaultHigh);
        }
        return std::exp(x);
    }

Myoldmopar:
These changes still seem appropriate for this defect's scope.  Merging this in.

Issue #7425, opened 2019-07-30
Wrong PLR in chiller IPLV calculation


lymereJ:
Issue overview
--------------
The wrong Part Load Ratio (PLR) is used in chiller IPLV calculation.

### Details
In the current implementation of the chiller IPLV calculation it is assumed that the chiller PLR is equal to the different loading percentage prescribed by AHRI 550-590 (% of full load capacity). The PLR at each different loading should be recalculated as the PLR is function of the operating capacity, not the full load capacity.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Pull request created

Myoldmopar:
Closed via #7426

PR #7426, opened 2019-07-30
Correct PLR in chiller IPLV calculation

lymereJ:
Pull request overview
---------------------
In the current implementation of the chiller IPLV calculation it is assumed that the chiller PLR is equal to the different loading percentage prescribed by AHRI 550-590 (% of full load capacity). The PLR at each different loading should be recalculated as the PLR is function of the operating capacity, not the full load capacity. This is reflected in the proposed changes.

Reference issue: #7425.


rraustad:
@Nigusse could you review this change. It's either a correction to the IPLV calculation or the documentation is wrong.

Nigusse:
Reviewing the current source code indicates that percent loading is used as chiller part load ratio. Looking at the ASHRAE Std 550-590 documentation, the percent loading shall be used for calculating the load for each loading point and then use the load to calculate PLR. Thus, I agree with @lymereJ proposed fix.

rraustad:
@lymerej the engineering documentation is correct as it stands and now agrees with the code?

lymereJ:
@rraustad, yes.

rraustad:
Only thing missing here is a unit test. Not sure who will do that. In StandardRatings.unit.cc there is only 1 test for a heating coil so at minimum there needs to be a unit test for a cooling coil.

@lymereJ do you want to give that a go or should someone else do that?

lymereJ:
@rraustad, if it's not too urgent, I'd like to give it a shot.

rraustad:
Several weeks yet (mid Sep) before they lock down the next release. Easiest to copy the heating coil unit test, change the test name, and change as necessary to create a cooling coil test. If you need help along the way just ask.

lymereJ:
@rraustad, thanks, will do!

rraustad:
@lymereJ are you making any progress on a unit test?

lymereJ:
Yes, I should have something early next week.

lymereJ:
@rraustad, I've added a unit test for the IPLV calcs. I tested it locally, it passes. Please let me know if something needs to be revised. Thanks.

src/EnergyPlus/ChillerElectricEIR.cc
```diff
@@ -1847,7 +1848,8 @@ namespace ChillerElectricEIR {
                                 ElectricEIRChiller(EIRChillNum).ChillerCapFT,
                                 ElectricEIRChiller(EIRChillNum).ChillerEIRFT,
                                 ElectricEIRChiller(EIRChillNum).ChillerEIRFPLR,
-                                ElectricEIRChiller(EIRChillNum).MinUnloadRat);
+                                ElectricEIRChiller(EIRChillNum).MinUnloadRat,
+                                IPLV);
```
rraustad, line 1852: @Myoldmopar do you have any reservations about adding a function argument to enable unit testing? I don't see anything wrong with this.

rraustad (approved):
Reviewed code and unit test.

rraustad:
I'm good with these changes. Can merge at will.

src/EnergyPlus/ChillerElectricEIR.cc
Myoldmopar: In general, adding functionality purely for the sake of a unit test is frowned upon.  However, in this case, it doesn't have to be purely for a unit test.  Having a function called CalcIPLV return a IPLV value is justified.

src/EnergyPlus/StandardRatings.cc
```diff
@@ -226,6 +226,7 @@ namespace StandardRatings {
                          int const EIRFTempCurveIndex,           // Index for the energy input ratio modifier curve
                          int const EIRFPLRCurveIndex,            // Index for the EIR vs part-load ratio curve
                          Real64 const MinUnloadRat,              // Minimum unloading ratio
+                         Real64 &IPLV,
```
Myoldmopar, line 229: Although the function could simply have returned the IPLV value rather than requiring another argument...

src/EnergyPlus/StandardRatings.cc
```diff
@@ -380,9 +380,10 @@ namespace StandardRatings {
 
                     ChillerEIRFT = CurveValue(EIRFTempCurveIndex, EvapOutletTemp, CondenserInletTemp);
 
-                    if (ReducedPLR(RedCapNum) >= MinUnloadRat) {
-                        ChillerEIRFPLR = CurveValue(EIRFPLRCurveIndex, ReducedPLR(RedCapNum));
-                        PartLoadRatio = ReducedPLR(RedCapNum);
+                    PartLoadRatio = ReducedPLR(RedCapNum) / ChillerCapFT;
```
Myoldmopar, line 383: So this just corrects the PartLoadRatio calculation before comparing it to the minimum unload PLR.  Sounds good.

src/EnergyPlus/StandardRatings.cc
```diff
@@ -489,6 +491,7 @@ namespace StandardRatings {
 
         // Writes the IPLV value to the EIO file and standard tabular output tables
         ReportChillerIPLV(ChillerName, ChillerType, IPLV, IPLV * ConvFromSIToIP);
+        return;
```
Myoldmopar: I don't think this is needed.  We're already returning here.  Unless I'm misreading something on the GitHub diff view.

tst/EnergyPlus/unit/StandardRatings.unit.cc
```diff
@@ -236,4 +239,94 @@ TEST_F(EnergyPlusFixture, SingleSpeedHeatingCoilCurveTest)
...
+                    ChillerElectricEIR::ElectricEIRChiller(1).CondenserType,
+                    ChillerElectricEIR::ElectricEIRChiller(1).ChillerCapFT,
+                    ChillerElectricEIR::ElectricEIRChiller(1).ChillerEIRFT,
+                    ChillerElectricEIR::ElectricEIRChiller(1).ChillerEIRFPLR,
+                    ChillerElectricEIR::ElectricEIRChiller(1).MinUnloadRat,
+                    IPLV);
+
+    EXPECT_DOUBLE_EQ(round(IPLV * 100) / 100, 3.87); // 13.20 IPLV
+
+}
```
Myoldmopar, line 331: This appears to be a good unit test, though I'll have to rebuild it with develop merged in to see if the performance curve work affected it.

Myoldmopar (changes requested):
There is a problem.  You changed the CalcChillerIPLV function signature, adding a new argument in between other arguments.  You then changed the call from ChillerElectricEIR, which is great.  But you didn't change the call from ChillerReformulatedEIR.  This is causing hard crashes in 20 input files.  You need to either fix up that function call, or, what I would prefer, would be for you to just make IPLV be a return value from this function and get rid of the new reference argument.

Myoldmopar:
Definitely don't merge - it's causing hard crashes in 20 test files.

Myoldmopar:
Pulled develop in, ran unit test fine.  I then fixed the code issues I mentioned above and a failing file is now working.  I think this will actually be ready to go in if CI comes in clean now.

Myoldmopar (approved):

Myoldmopar:
As an aside, @lymereJ, please use a different branch when making any submissions.  Thanks for this submission.

rraustad:
CI is clean now thanks to @Myoldmopar update.

PR #7430, opened 2019-07-30
AFN tests fail to run serially following up addition of new tests in #7338

jmarrec:
Pull request overview
---------------------

Following up on this comment: https://github.com/NREL/EnergyPlus/pull/7338#issuecomment-513941939, the unit tests will fail to run serially after adding a few in #7338 because there is OneTimeFlag that isn't being reset.

This PR aims to address this by putting the OneTimeFlag is an anynomous namespace at the global level (so you cannot access it / reset it outside of this modules) and have it being reset in the `AirflowNetworkBalanceManager::clear_state` function.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/AirflowNetworkBalanceManager.cc
+++ b/src/EnergyPlus/AirflowNetworkBalanceManager.cc
@@ -497,9 +497,9 @@ namespace AirflowNetworkBalanceManager {
         if (AirflowNetworkFanActivated && SimulateAirflowNetwork > AirflowNetworkControlMultizone) {
             if (ValidateDistributionSystemFlag) {
                 ValidateDistributionSystem();
+                ValidateFanFlowRate();
                 ValidateDistributionSystemFlag = false;
             }
-            ValidateFanFlowRate();
         }
         CalcAirflowNetworkAirBalance();
 
@@ -10447,14 +10447,12 @@ namespace AirflowNetworkBalanceManager {
             {
                 auto const SELECT_CASE_var(AirflowNetworkCompData(AirflowNetworkLinkageData(i).CompNum).CompTypeNum);
                 if (SELECT_CASE_var == CompTypeNum_CVF) { // 'CVF'
-                    int nodeNum = AirflowNetworkNodeData(AirflowNetworkLinkageData(i).NodeNums[1]).EPlusNodeNum;
                     int typeNum = AirflowNetworkCompData(AirflowNetworkLinkageData(i).CompNum).TypeNum;
-                    // Will potentially be passed by ref to Fans::GetFanVolFlow so make a copy instead of using a ref
-                    Real64 FanFlow = Node(nodeNum).MassFlowRate;
                     if (DisSysCompCVFData(typeNum).FanTypeNum == FanType_SimpleVAV) {
                         if (DisSysCompCVFData(typeNum).FanModelFlag) {
                             DisSysCompCVFData(typeNum).MaxAirMassFlowRate = HVACFan::fanObjs[DisSysCompCVFData(typeNum).FanIndex]->designAirVolFlowRate * StdRhoAir;
                         } else {
+                            Real64 FanFlow; // Return type
                             GetFanVolFlow(DisSysCompCVFData(typeNum).FanIndex, FanFlow);
                             DisSysCompCVFData(typeNum).MaxAirMassFlowRate = FanFlow * StdRhoAir;
                         }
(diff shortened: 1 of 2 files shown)


src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -233,6 +233,14 @@ namespace AirflowNetworkBalanceManager {
     // Report variables
 
     // MODULE VARIABLE DECLARATIONS:
+
+     namespace {
+        // These are purposefully not in the header file as an extern variable. No one outside of this should
+        // use these. They are cleared by clear_state() for use by unit tests, but normal simulations should be unaffected.
+        // This is purposefully in an anonymous namespace so nothing outside this implementation file can use it.
+        bool ValidateDistributionSystemFlag(true);
```
jasondegraw, line 241: As I indicated on the previous PR, this is NOT the preferred solution and is only exacerbating the problem.

jasondegraw (changes requested):
This PR solves one smaller problem by making others worse and includes a bunch of unrelated changes. Fix the problem at hand and leave the cosmetic changes for another time.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -9652,796 +9666,795 @@ namespace AirflowNetworkBalanceManager {
...
-        static bool ErrorsFound(false);
         bool LocalError;
         Array1D_bool NodeFound;
-        Real64 FanFlow;
-        static bool IsNotOK(false);
-        static bool errFlag(false);
+
+        bool ErrorsFound(false);
+        bool IsNotOK(false);
+        bool errFlag(false);
```
Myoldmopar, line 9674: Making these static is great.  There was no need for ErrorsFound, IsNotOK, or errFlag to be static in the first place.  Having a static OneTimeFlag in the function is the core issue with this function, and I am glad to see it is gone.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -9652,796 +9666,795 @@ namespace AirflowNetworkBalanceManager {
...
-                        ErrorsFound = true;
-                    }
-                    if (AirflowNetworkNodeData(AirflowNetworkLinkageData(i).NodeNums[1]).EPlusTypeNum == EPlusTypeNum_MIX) {
-                        ShowSevereError(RoutineName + "An " + CurrentModuleObject + " object (" + AirflowNetworkLinkageData(i).CompName + ')');
-                        ShowContinueError("does not allow a AirLoopHVAC:ZoneMixer node = " +
-                                          AirflowNetworkNodeData(AirflowNetworkLinkageData(i).NodeNums[1]).Name);
-                        ErrorsFound = true;
-                    }
-                    if (AirflowNetworkNodeData(AirflowNetworkLinkageData(i).NodeNums[0]).EPlusNodeNum > 0) {
-                        ShowSevereError(RoutineName + "An " + C
```
Myoldmopar, line 10443: Moving these checks to their own one-time-use function is good from an organization standpoint.

src/EnergyPlus/AirflowNetworkBalanceManager.hh
```diff
@@ -228,6 +228,8 @@ namespace AirflowNetworkBalanceManager {
 
     void ValidateDistributionSystem();
 
+    void ValidateFanFlowRate(); // Catch a fan flow rate from EPlus input file and add a flag for VAV terminal damper
```
Myoldmopar, line 231: I don't even really think it needs to be exposed in the header; you could leave it contained in the implementation file, although that's rarely done in EnergyPlus...for now.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
Myoldmopar: So I do agree that adding to the global state here is not the best solution, and it would be better to just ensure we only execute the one-time code...well...once.  However, I will say that pulling out static variables from inside functions is a very important incremental step.  I'm not as opposed to this as a band-aid, and I am now encountering this in my testing and want this to be fixed.  In future fixes, we'll push for more formal solutions, but I'm going to merge this in.

Myoldmopar (approved):

jasondegraw:
It's already merged, but for the sake of posterity:

* the cosmetic changes will just make a lot of things harder later on and result in work that could have been avoided
* this all could have been avoided had the original bug been properly fixed

Issue #7174, opened 2019-02-15
Standard Effective Temperature (SET) outputs from Energyplus not available


Arezaei558:
Issue overview
--------------
SET calculations occur in the background Energyplus code calculations but not SET is not a direct output from simulations that modeler can request. Please see the Unmethours link below for more information. 

https://unmethours.com/question/37129/standard-effective-temperature-set-outputs-from-energyplus/

SET is a necessary output from simulations to calculate the Passive Survivability LEED Pilot Credit: https://www.usgbc.org/credits/passivesurvivability. Not having SET output directly reported from the results of an Energyplus simulation limits the modeler's use of the entire OS measures ecosystem to visualize the results in measures like ViewData. 

Although Ladybug tools have these calculations streamlined (https://mostapharoudsari.gitbooks.io/ladybug-primer/content/text/components/PMV_Comfort_Calculator.html ---- example file: https://github.com/ladybug-tools/ladybug/blob/comfort-utility/ladybug/comfort/pmv.py#L207-L411   -----source code: https://github.com/ladybug-tools/ladybug/blob/comfort-utility/ladybug/comfort/pmv.py#L207-L411 -----) it will be helpful to have SET as a direct output from E+ to feed into OS ecosystem. 

### Details
N/A

RKStrand:
@Arezaei558 As pointed out in the unmethours discussion, there is a SET variable calculated for the Pierce model.  I will add the ability to output that variable.  To get SET you will have to select the Pierce model, but at least this will give you what you need.  I will add to the documentation so that it is clear you have to select the Pierce model to get SET.  The other models do not calculate SET internally so adding that for other models would require quite a bit more work.

Arezaei558:
Any updates on when this output for SET will be added to a next release? Or has it already been added?

Arezaei558:
Never mind, just realized it was released with 9.2 version. Thank you all!

mjwitte:
. . . and this issue should have been closed. Addressed in #7431.

PR #7431, opened 2019-07-30
Addition of Standard Effective Temperature (SET) as an Output Variable

RKStrand:
Pull request overview
---------------------
Based on discussion on unmethours.com, there was a request to add Standard Effective Temperature (SET) as an output variable.  This is calculated as part of the Pierce two-node thermal comfort model.  This work fulfills this request and it is now possible to select SET as an output variable in EnergyPlus.  This includes code changes, a unit test, documentation updates, and a modified test suite input file to show the use of this new output variable.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7174 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes n/a


tst/EnergyPlus/unit/ThermalComfort.unit.cc
```diff
@@ -948,3 +948,527 @@ TEST_F(EnergyPlusFixture, ThermalComfort_CalcIfSetPointMetWithCutoutTest)
...
+        " ",
+        
+    });
+    
+    ASSERT_TRUE(process_idf(idf_objects));
+    
+    OutputProcessor::TimeValue.allocate(2);
+    DataGlobals::DDOnlySimulation = true;
+    
+    ManageSimulation();
```
mjwitte: Would prefer a much more lightweight unit test. Seems you could just have zone, people, and schedules in the idf snippet and then just call the functions to process zone and people data and then CalcThermalComfortPierce.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -467,6 +467,12 @@ namespace ThermalComfort {
                                     "Zone",
                                     "State",
                                     People(Loop).Name);
+                SetupOutputVariable("Zone Thermal Comfort Pierce Model Standard Effective Temperature",
+                                    OutputProcessor::Unit::None,
```
mjwitte: OutputProcessor::Unit::C ??

tst/EnergyPlus/unit/ThermalComfort.unit.cc
```diff
@@ -948,3 +948,527 @@ TEST_F(EnergyPlusFixture, ThermalComfort_CalcIfSetPointMetWithCutoutTest)
...
+        
+        "Output:Table:SummaryReports,",
+        "  AllSummary; !- Report 1 Name",
+        " ",
+        
+    });
+    
+    ASSERT_TRUE(process_idf(idf_objects));
+    
+    OutputProcessor::TimeValue.allocate(2);
```
mjwitte: And when you merge in develop, this line will fail. Simply delete it.

RKStrand: Will delete it.

tst/EnergyPlus/unit/ThermalComfort.unit.cc
mjwitte: Moderately important. This should get your schedule in place, if done before processing the people:
```
    ScheduleManager::ProcessScheduleInput();
    ScheduleManager::ScheduleInputProcessed = true;
```

RKStrand: @mjwitte Ok, it's now much more lightweight.  Doesn't call anything except the Pierce subroutine.  Sets every manually in the unit test now.  This should be up in GitHub as of now.

rraustad:
New report variable added but no changes to table reports.

rraustad:
- Final code and unit test review, removed clear_state access from unit test
 - Compiled and reviewed IO Ref., small edits, no new figures to add to CMake.
 - Ran sequential unit tests in release mode without failure
 - Ran example file, no warnings in err file. Same for annual run.
 - Verified new report variable added to existing example file (TermReheatSurfTC.idf)
 - Pulled develop and uploaded final changes
 - Plotted new report variable for design days

image

rraustad (approved):
Minor editorial mods to IO Ref, correct unit test, uploaded changes.

rraustad:
This is ready to merge once CI has a chance to process changes.

rraustad:
@RKStrand can you comment on the SET being that much above the summer design day zone temp in the figure above (e.g., occupants have on too many clothes, etc.)? The summer zone humidity ratio is actually lower than winter and zone temp is not that much different. No other code changes will be requested, just wondering if this is working as expected (i.e., if new issue should be opened).

RKStrand:
@rraustad I see what you are saying in the results, but I think this is actually what one would expect.  SET includes both radiative and convective aspects as well as a bunch of other things.  I would expect that SET would exhibit somewhat similar behavior to MAT but the surface temperatures in the space will dampen the response and make SET more closely related to outside air temperature (since that will in turn affect inside surface temperatures).  So, I think that what you are seeing is actually correct--SET is lower than MAT for the most part for winter (because MRT is lower as a result of the cold exterior conditions) and SET is generally higher than MAT for summer.  The "heavy" surfaces are damping the response of SET so when MAT changes rapidly like it does at the end of the work day in the summer data MAT can flip so that it is higher than SET because MAT in that situation is actually higher temporarily for that time.  Basically, if one were to graph MAT, MRT, and SET, I would guess that in general SET should track between these two for the most part.  Hopefully that makes sense, but if it doesn't, please let me know.  Overall, I think what you are seeing is the right trend and I don't think there is a problem here that needs to be logged as a defect.  Thanks for asking and for diligently checking results to see if there are additional problems!

rraustad:
Followup graph based on results discussion.

image

rraustad:
I can't think of a reason why this should not merge except for a complete CI assessment...

RKStrand:
@Myoldmopar This has met with approval from @rraustad but it sounds like there wasn't a complete CI result?  Does it need to be relaunched or something?

Myoldmopar:
The CI machine itself needed to be relaunched.  And it has been done.  CI results should be coming in soon.

rraustad:
Diff's from Coil Sizing report fixes #7435 are creeping in here. Merged that one into develop 17 hrs ago.

rraustad:
Well, I was just going to merge this anyway since I understood the diff's. But it can't hurt to have a fresh clean CI result.

RKStrand:
@rraustad Oops--sorry to get in the way.  Should have left it alone.  But as you pointed out, a fresh CI can't (shouldn't) hurt.

Myoldmopar:
We've had an issue with the CI box and had to take it down.  I'm working to get some extra support stood back up but this is likely to need to wait until after the weekend.

rraustad:
CI machines now match with expected results.

PR #7432, opened 2019-07-30
Stratified tank adaptive internal sub timestep

nmerket:
Pull request overview
---------------------
The stratified tank solves a set of ODEs within the tank every system timestep. It is set up as an initial value problem and marches forward through time from the beginning to the end of the system timestep via shorter sub timesteps. In the original stratified tank model that used the forward Euler method, it stepped forward at 1 second intervals. In #7146 I introduced a different solution method for the stratified tank. Because that method solved the ode using a simplified analytical solution it allowed larger timesteps (60 seconds) to be taken within the tank. This yielded a 5-6x speed improvement on the entire simulation at the expense of some accuracy.

One of the issues with the new solution in #7146 was that in cases of high flow rates and/or heat transfer rates it could overshoot or undershoot control decisions for the heaters because those decisions are only made at the beginning of each sub timestep, which was every 60 seconds. Conversely when a tank is sitting with no draws during the system timestep, a 60 second sub timestep is overly conservative and a nearly equivalent result could be achieved with fewer computations. 

This PR introduces an adaptive sub timestep within the stratified tank. Instead of being fixed at 60 seconds, it can float between 10 seconds and 10 minutes. Each sub timestep in the model makes a linear prediction of how long it can run before a control decision is made and the sub timestep is set to approach that time. When the heat transfer rates are high in the tank, it will slow down to capture the dynamics. When the heat transfer rates are low it speeds up to save time. With this model change I am seeing simulation times ~2x faster than the current `develop` branch with the fixed sub timestep. When compared to the original stratified tank model it is ~10x faster.

The regression testing is showing differences in all models that include a stratified tank. To get a better sense of the changes introduced I have been running and comparing three branches:

- `develop`
- `stratified_tank_adaptive_interal_dt`
- `old_stratified_tank` which is a private branch I created that is identical to `develop` except that I retrieved the old `CalcWaterThermalTankStratified` function from the git history and swapped it in.

After spending time looking at the results of all three of those together, I am feeling pretty confident that this is working correctly. Many of the regressions from `develop` are actually bringing the results closer in line with the old model because this new model can now capture the high speed dynamics a little bit better. I am also running this on a few annual test cases of residential heat pump water heaters and finding very good agreement. I'm happy to share any specific plots of results if that would help with a review. 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes n/a


src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8112,8 +8124,95 @@ namespace WaterThermalTanks {
...
+                }
+
+                // Set the sub timestep (dt)
+                dt = TimeRemaining;
+                for (int i = 0; i < nTankNodes; ++i) {
+                    const Real64 Denominator = fabs(A[i] * Tavg[i] + B[i]);
+                    if (Denominator != 0.0)
+                        dt = min(dt, dT_max / Denominator);
+                }
+                dt = max(min(SubTimestepMin, TimeRemaining), dt);
```
rraustad, line 8166: How often will TimeRemaining be less than SubTimestepMin (10 sec)? How much impact would it have to allow TimeRemaining to go down to 1 sec as the previous model allowed?

rraustad: I have that backwards. TimeRemaining is less than SubTimestepMin when these iterations are near completion. It's hard to understand this at first glance. Nearing the end of the iterations, TimeRemaining is approaching 0 and dT_max is also approaching 0. So the max(min()) above simply caps dt at 10 sec or TimeRemaining, whichever is less and takes the greater of that or dt. Then the line below limits dt to SubTimeStepMax. So when is dt a small number?

nmerket: `dT_max` may or may not approach zero by the end of `TimeRemaining`. It's a measure of how much we'll allow the tank temperature to change before doing the control decisions and looping around again. We want to loop and recalculate about the time a heater kicks on or off. Every once in a while that happens near the end of a system time step. This logic makes sure we don't go past the end of the system timestep but right up to it. This line is confusing. Maybe I should rewrite it to be more clear. 

I messed around with `dt` being allowed to go down to 1 second, but it counteracted our performance gains. The estimation of the next `dt` tends to undershoot it because it is a linear approximation of a non-linear solution. This means that without a lower bound on `dt` it would never get there. A 1 second lower bound works, but it takes a lot of decreasingly small steps before being pushed over. Each time around the sub timestep loop is more expensive now as there are more computations than with the old model, so it canceled it out. 

I'd argue that the 10 second minimum sub timestep is at least as good as a solution as the previous 1 second because it uses the analytical solution over that longer time rather than the linear forward Euler approximation the old model used. The only place where we'd see a difference is in making a control decision for a heater. It may happen a few seconds later with this model. 

That being said this is an improvement over what we have in 9.1, which is that the sub timestep is _always_ 60 seconds.

nmerket:
ping @mjwitte

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8065,8 +8077,95 @@ namespace WaterThermalTanks {
...
+                        B[i] -= (- node.OffCycParaLoad + node.OnCycParaLoad + (node.OnCycLossCoeff - node.OffCycLossCoeff) * Tank.AmbientTemp) / NodeCapacitance;
+                    }
+                }
+
+                // Set the sub timestep (dt)
+                dt = TimeRemaining;
+                for (int i = 0; i < nTankNodes; ++i) {
+                    const Real64 Denominator = fabs(A[i] * Tavg[i] + B[i]);
+                    if (Denominator != 0.0)
+                        dt = min(dt, dT_max / Denominator);
```
rraustad, line 8164: If nTankNodes > 1, then how is dt adjusted? It appears the nth node will set dt. Shouldn't there be a += here and then a divide by nTankNodes after the for loop? Or min dt of all nodes? Or just use the nth node if that is the intent?

nmerket: `dt` is determined by solving

image

for dt. We determine the maximum allowable temperature change, `dT_max`,  for any node above. In these lines we estimate the time it would take for each node to experience that much of a temperature change. Then I take the minimum of those so that we only go as far forward as the fastest changing node temperature, which isn't necessarily `n`. That's what the `min` here does. It's equivalent to writing:

```c++
if (dT_max / Denominator < dt) 
    dt = dT_max / Denominator
```

That way we're calculating the min dt of all the nodes.

rraustad: OK, I missed the "min(dt," part of it. This looks good then with using the smallest dt.

min(dt, dT_max / Denominator);

rraustad (approved):
I've been over this several times. I can't think of a reason why this methodology should not be employed, I can't dispute the math, and the answers are closer than before. No reason to hold this up.

mjwitte (commented):
@nmerket This all looks good. I've merged in develop for one more round of CI testing.
But I don't see a unit test or extension of an existing one?

mjwitte:
And I don't see any doc changes - engineering ref?

nmerket:
@mjwitte 

### Unit Tests

There are several unit tests of the stratified tank that I broke and got working in the process of implementing this feature. I couldn't think of any additional tests that were needed to verify this was working properly.

### Engineering Reference

Fair point. The documentation currently states:

> The system time step is divided into one minute substeps at which control decisions are evaluated that allow the simulation to capture events that occur on a short time scale.

That should be replaced with a discussion of the adaptive substeps. I'll take care of that.

mjwitte:
Well, there is high coverage of `CalcWaterThermalTankStratified`in the unit tests. There are a few lines here and there that aren't covered. If it's easy to add another case and pick up one or two, that would be great. Otherwise, it sounds like the unit tests did their job here.

mjwitte (approved):
Doc changes look good. Will merge after CI completes.

PR #7435, opened 2019-07-30
Coil Sizing report fixes

EnergyArchmage:
Pull request overview
---------------------
A number of values reported to the Coil Sizing Details summary report table were found to be incorrect.  This pull request fix the following:

- For cases of central air system cooling coils that are sized using VentilationRequirement as the the type of load to size on, the values for (weighted average) zone air drybulb, relative humidity and humidity ratio were not being filled properly.  This PR corrects a logic problem that allows that case to work as intended. 

- The peak design day and time of peak during that design day for central air system cooling was not being properly discerned when the type of load to size on is changed between peak total cooling load and peak sensible cooling load.  This PR corrects that situation.

- For certain types of Coils in Zone HVAC equipment, the concurrent zone air temperature was being reported as the return air temperature instead of the zone air temperature. This PR corrects that situation.

- Coil entering and leaving conditions were sometimes being reported incorrectly because they were getting overwritten by the values from the sizing structures. In many cases the coil's parent had already made adjustments to those values for things like OA mixing or fan heat.  This PR corrects that situation. 

- Values reported for heating coils for Coil Total Capacity at Ideal Loads Peak [W] were often incorrect and did not match Coil Sensible Capacity at Ideal Loads Peak [W]. This was corrected. 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
Only seeing table diff's now. I tracked down the first one in 2ZoneDataCenterHVAC_wEconomizer in the Coil Sizing Summary report. At end of table shows lower Zone Sensible Heat Gain at Ideal Loads Peak for all zones in this branch. These are the only diff's in this example file. I assume all diff's are related to this field but haven't checked others yet.

CoilSizingDiffs

rraustad:
Oh beautiful, the results now match what is shown in the eio file !!! The question is should the Calc Des Load (this branch) or User Des Load (develop) show up here in the Coil Sizing Summary? It would seem we would want the unadjusted zone load.

image

rraustad:
And it's a factor of 1.3 as shown here:

```
  Sizing:Parameters,
    1.3,                     !- Heating Sizing Factor
    1.3,                     !- Cooling Sizing Factor
    ;                        !- Timesteps in Averaging Window
```

rraustad:
Then next table diff file is 5ZoneAirCooled. The heating coils table is shown first, there were no diff's in this table. The Coil Sizing Summary results for Coil Total Capacity at Ideal Loads Peak has changed. Now it matches the sensible capacity as it should, and the data match the Design Coil Load from the Heating Coils table instead of the Nominal Total Capacity. This makes sense but @EnergyArchmage will have to confirm.

5ZoneAirCooled_CoilSizingDiffs

rraustad:
Also in 5ZoneAirCooled the Coil and Fan Net Total Capacity at Ideal Loads Peak also now match the above (there were the same diff's here), but this is supposed to include fan heat and looks to not include it, @EnergyArchmage ?

image

EnergyArchmage:
In 5ZoneAirCooled's reheat coils, the Supply Fan Air Heat Gain at Ideal Loads Peak [W] values are also coming out zero, so the combined coil and fan capacity is coming out the same.  This is expected when the coil sizing is not actually including the fan heat.  The idea with the report is to show what is actually happening with the sizing calculations and unexpected values usually point out where the sizing calculations could be improved.  I don't think fan heat is universally included in coil sizing in the DOE version.

rraustad:
@Myoldmopar I'm not seeing any glaring problems yet. Changes have moved sizing data in the right direction. Not saying all the problems are solved, but the changes here have corrected some mistakes found over time.

rraustad:
Table diff's only as changes are related to Coil Sizing Summary table.

Issue #7381, opened 2019-07-10
Coil:Heating:DX:MultiSpeed & Coil:Heating:WaterToAirHeatPump:EquationFit values missing from ComponentSizes when hard-sized


shorowit:
Issue overview
--------------
I ran EnergyPlus 9.1 simulations with single-speed and multi-speed DX coils. When I look in the ComponentSizes table in eplusout.sql, I don't see capacity/airflow values for the `Coil:Heating:DX:MultiSpeed` object. The `Coil:Cooling:DX:MultiSpeed`, `Coil:Heating:DX:SingleSpeed`, and `Coil:Cooling:DX:SingleSpeed` objects appear correct.

### Coil:Heating:DX:SingleSpeed ComponentSizes

CompType | CompName | Description
-- | -- | --
Coil:Heating:DX:SingleSpeed | ASHP HTG COIL | User-Specified Rated Air Flow Rate
Coil:Heating:DX:SingleSpeed | ASHP HTG COIL | User-Specified Gross Rated Heating Capacity
Coil:Cooling:DX:SingleSpeed | ASHP CLG COIL | User-Specified Rated Air Flow Rate
Coil:Cooling:DX:SingleSpeed | ASHP CLG COIL | User-Specified Gross Rated Total Cooling Capacity
Coil:Cooling:DX:SingleSpeed | ASHP CLG COIL | User-Specified Gross Rated Sensible Heat Ratio

### Coil:Heating:DX:MultiSpeed ComponentSizes

CompType | CompName | Description
-- | -- | -- 
Coil:Heating:DX:MultiSpeed | ASHP HTG COIL | Design Size Resistive Defrost Heater Capacity
Coil:Cooling:DX:MultiSpeed | ASHP CLG COIL | User-Specified Speed 2 Rated Air Flow Rate
Coil:Cooling:DX:MultiSpeed | ASHP CLG COIL | Design Size Speed 1 Rated Air Flow Rate
Coil:Cooling:DX:MultiSpeed | ASHP CLG COIL | User-Specified Speed 1 Rated Air Flow Rate
Coil:Cooling:DX:MultiSpeed | ASHP CLG COIL | Speed 2 User-Specified Total Cooling Capacity
Coil:Cooling:DX:MultiSpeed | ASHP CLG COIL | Speed 2 User-Specified Rated Sensible Heat Ratio
[...]

PR #7436, opened 2019-07-30
Addressed Coil:Heating:DX:MultiSpeed & Coil:Heating:WaterToAirHeatPump:EquationFit values missing from ComponentSizes when hard-sized

Nigusse:
Pull request overview
---------------------
Addressed issue #7381. Summary of changes in this fix:
- added missing heating capacity reporting for  COIL:HEATING:WATERTOAIRHEATPUMP:EQUATIONFIT heating coil
- added missing calculation, and flow and heating capacity reporting for  multi-speed DX heating coil
- corrected wrong rated total cooling capacity reporting for multi-speed DX cooling coil
- added missing rated sensible cooling capacity reporting for multi-speed DX cooling coil

Diffs are expected based on this change in SQL output file and HTML tables for MultiSpeed DX coils. The eio and audit files diffs are expected due to addition of new sizing variables for multi-speed DX heating coil and diffs in html table output is due to filling missing values. There are 10 example files with multi-speed DX coils that are expected to show diffs. Similarly heating capacity is reported to EIO file for COIL:HEATING:WATERTOAIRHEATPUMP:EQUATIONFIT coil, hence diffs are expected.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
The eio and audit files diffs are expected due to addition of new sizing variables for multi-speed DX heating coil. There are nine example files with multi-speed DX coils that are expected to show diffs.

Nigusse:
while reviewing the tabular outputs of this defect file the multi-speed DX cooling coils "Coil Final Gross Total Capacity [W]" is 10128.536 W, which corresponds to the lowest speed (speed 1) user specified cooling capacity. I was expecting to see the highest speed (speed 2) user specified cooling capacity of 14067.411 W.

Nigusse:
Also addressed 0.0 rated sensible cooling capacity reporting issue for multi-speed DX cooling coil.

rraustad:
html table from V9.1:
image

html table from this branch:
image

rraustad:
@shorowit do you want to test this branch and defect file for sql accuracy? I don't use sql but since the tables are now correct I assume the sql tables are as well.

src/EnergyPlus/DXCoils.cc
```diff
@@ -7988,48 +7907,68 @@ namespace DXCoils {
...
+                    FieldNum = 10 + (Mode - 1) * 5;
+                    SizingString = DXCoilNumericFields(DXCoilNum).PerfMode(1).FieldNames(FieldNum) + " [W]";
+                    DataTotCapCurveIndex = DXCoil(DXCoilNum).MSCCapFTemp(Mode);
+                    if (IsAutoSize || !HardSizeNoDesRun) {
+                        // Heating capacity is assumed to be equal to the cooling capacity
+                        SizingMethod = AutoCalculateSizing;
+                        DataFractionUsedForSizing = 1.0;
+                        if (DXCoil(DXCoilNum).CompanionUpstreamDXCoil > 0) {
+                            NumOfSpeedCompanion = DXCoil(DXCoil(DXCoilNum).CompanionUpstreamDXCoil).NumOfSpeeds;
+                            DataConstantUsedForSizing = DXCoil(DXCoil(DXCoilNum).CompanionUpstreamDXCoil).MSRatedTotCap(NumOfSpeedCompanion);
```
rraustad, line 7927: This code is inside the if (Mode == DXCoil(DXCoilNum).NumOfSpeeds) { section. You should not use DataConstantUsedForSizing and DataFractionUsedForSizing on the highest speed. The reason is that you will not get the Design vs User capacities if those are different by 10%. For the lower speeds using the Data* variables is OK (lines 7940-7958 below). For sizing air flows you did it correctly (lines 7822-7849 above).

shorowit:
Yes, I would like to test this. I can probably do so next week...

src/EnergyPlus/DXCoils.cc
Nigusse: @rraustad I did not change the heating capacity calculation here albeit adding reporting using request sizing. The code is setting the cooling capacity to the heating capacity. Is there a better was of doing it?

rraustad: OK then. For the heating coil this is probably the way to size the coil to be the same size as the cooling capacity. I can think of one other method but I don't want to do that here.

rraustad:
@Nigusse can you resolve these conflicts in the unit tests. This looks ready to go.

rraustad:
@shorowit don't forget to test this branch, thanks.

shorowit:
Ok, I just tested it. It gives all the values I needed and these values are as expected. So it meets my needs.

I did notice a few inconsistencies though. I'll let you decide if they are worth tackling or fall outside the scope of this PR.
1. Coil:Cooling:DX:MultiSpeed lists, e.g., "**Speed 2** User-Specified ... Capacity" while Coil:Heating:DX:MultiSpeed lists, e.g., "User-Specified **Speed 2** ... Capacity".
2. Coil:Cooling:DX:MultiSpeed lists the capacity at only max speed (e.g., 4) while Coil:Heating:DX:MultiSpeed lists the capacity at all speeds (e.g., 1, 2, 3, and 4).
3. Coil:Cooling:DX:MultiSpeed only lists the total cooling capacity while Coil:Cooling:WaterToAirHeatPump:EquationFit lists both sensible and total cooling capacities.

Nigusse:
The first item of inconsistencies listed above is due to difference how the RequestSizing() function and local sizing reporting work. I think, this can be addressed easily but it may impact the eio output variable names. I will post new issue that addresses all the three inconsistencies @shorowit pointed out above.

Nigusse:
Posted new issue #7477 that will address additional reporting inconsistencies.

Nigusse:
This branch was expected to show diffs in only ten example files but I see diffs more than that. I will merge in develop and re-run.

rraustad:
@Myoldmopar this does actually change table output as shown from my test 7 days ago. So I don't know if this should go in before or after release. See results table above for Coil:Heating:DX:Multispeed. Need to be tagged as Output Change?

Myoldmopar:
Since this is completely changing the multispeed table it will need to wait until after release.  I'll alter the milestone and it can drop in as soon as we've tagged 9.2.0

nrel-bot-2b:

Issue #6767, opened 2018-05-31
Coil:WaterHeating:AirToWaterHeatPump OutputVariable names for heating reference cooling coils 


MatthewSteen:
Issue overview
--------------
The variable names for the heat pump water heater (HPWH) heating coil reference cooling coils instead of heating coils and must be requested with the Cooling Coil names.

__Input Output Reference__
* [Coil:WaterHeating:AirToWaterHeatPump:Pumped](https://github.com/NREL/EnergyPlus/blob/develop/doc/input-output-reference/src/overview/group-heating-and-cooling-coils.tex#L5711-L5736)
* [Coil:WaterHeating:AirToWaterHeatPump:Wrapped](https://github.com/NREL/EnergyPlus/blob/develop/doc/input-output-reference/src/overview/group-heating-and-cooling-coils.tex#L5931-L5960)

---

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

YanfeiNREL:
@MatthewSteen Hey, would you please point me where did u find the issue? It is in the code of DXCoils.cc? Or WaterThermalTanks.cc? Or the code to generate the Input/Output References? Or from the EnergyPlus simulation outputs?Thanks.

MatthewSteen:
Based on the links I provided, it was a documentation issue.

MatthewSteen:
Based on the link, I believe it was a documentation issue. See the last four OutputVariable names.

    HVAC,Average,Cooling Coil Total Cooling Rate [W]

    HVAC,Sum,Cooling Coil Total Cooling Energy [J]

    HVAC,Average,Cooling Coil Sensible Cooling Rate [W]

    HVAC,Sum,Cooling Coil Sensible Cooling Energy [J]

    HVAC,Average,Cooling Coil Latent Cooling Rate [W]

    HVAC,Sum,Cooling Coil Latent Cooling Energy [J]

    HVAC,Average, Cooling Coil Runtime Fraction []

[...]

PR #7439, opened 2019-07-31
Fix the OutputVariables reference issue for cooling coils for #6767 and #7440

YanfeiNREL:
Fix issue for #6767. 
The Output Variables of two coils are referencing the cooling coils, instead of heating coils:
Coil:WaterHeating:AirToWaterHeatPump:Pumped
Coil:WaterHeating:AirToWaterHeatPump:Wrapped.

Heating Coils have no sensible/latent calculated, therefore the Output
Variables are updated accordingly for the two coils.

Pull request overview
---------------------

### Work Checklist

### Review Checklist
 - [ X ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/doc/input-output-reference/src/overview/group-heating-and-cooling-coils.tex
+++ b/doc/input-output-reference/src/overview/group-heating-and-cooling-coils.tex
@@ -5725,84 +5725,61 @@ Coil:WaterHeating:AirToWaterHeatPump:Pumped,
 
 \begin{itemize}
 \item
-  HVAC,Average,Cooling Coil Total Cooling Rate {[}W{]}
+  HVAC,Average,Heating Coil Total Heating Rate {[}W{]}
 \item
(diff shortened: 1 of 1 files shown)


mjwitte:
This is on hold until the discussion in #6767 comes to some conclusion. Then this PR can be modified accordingly.

rraustad:
This got merged? I thought this was on hold and then in #6767 decided there should be no change.

Myoldmopar:
Good call.  If we need to revert this out, we can definitely do it.

YanfeiNREL:
I updated the DOC of [EnergyPlus Input/Output Reference] with the explanations why we call it Cooling Coil. This is a minimal effort to fix the issue in #6767 and #7440. This is not the best solution. If you have a better solution, i would love to reopen it and work on it. @rraustad Thanks.

Issue #7384, opened 2019-07-12
Material:InfraredTransparent issues


mjwitte:
Issue overview
--------------
1. Noticed this code in the setup for Material:InfraredTransparent in [HeatBalanceManager::GetMaterialData](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HeatBalanceManager.cc#L1832-L1836):
```
            if (MaterialNumProp >= 1) {
                Material(MaterNum).Resistance = MaterialProps(1);
                Material(MaterNum).ROnly = true;
            } else {
                Material(MaterNum).Resistance = 0.01;
            }
```
It appears that the object at one time had more input fields, but currently, the object only has a name field. So MaterialNumProp will always be zero, and ROnly will not be set true.

Not sure what the implications of this may be. Will need to track down what happens with ROnly and see if this changes results for runs with an IRT material.

2. Also, since the object has no other input fields, this if and following could be reduced to just the default assignments.

3. "Surface Inside Face Net Surface Thermal Radiation Heat Gain Rate" does not balance. In `InitInteriorRadExahange` the area used in the radiant exchange calculations is [doubled for IRT surfaces](https://github.com/NREL/EnergyPlus/blob/d74896b46f53cc71953c54f5166c37384e3432d4/src/EnergyPlus/HeatBalanceIntRadExchange.cc#L593-L600). Ultimately `NetLWRadToSurf` [W/m2] is calculated in `CalcInteriorRadExchange` using the 2xArea, but then this flux is applied to the surface heat balance on an equal basis with other perArea values. `NetLWRadToSurf` needs to be multiplied by 2 at the end of `CalcInteriorRadExchange`.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v9.2.0-https://github.com/NREL/EnergyPlus/commit/3e91d130a5724d86a9172f4156001e69b10127ad
 - Notice while coding for Construction:AirBoundary

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
[...]

PR #7441, opened 2019-08-01
Correction of Infrared Transparent Material Code and Interior Radiation Exchange

RKStrand:
Pull request overview
---------------------
This pull request addresses some issues with the infrared transparent material code.  First, it did not properly assign that this was an R-only layer and cleaned up the code in the get routine for this material type.  In addition, it also corrected an issue with the interior radiation balance that resulted when using an infrared transparent material.  This imbalance was the result of doubling the surface area without doubling the potential impact on the zone.  As there is no rationale for doubling the impact of this surface type on the rest of the zone, the original surface area is now maintained and the radiation on the interior side balances.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7384 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


RKStrand:
@mjwitte The two test files that have differences are ones that use the IRT material.

mjwitte:
@EnergyArchmage You ok with these changes?

mjwitte (changes requested):
Merged in develop and resolved the conflicts. Tested before and after with StackedZonesWithInterzoneIRTLayers example file (with additional outputs). Surface Inside Face Net Surface Thermal Radiation Heat Gain Energy across all surfaces balances to zero now.

Unit test looks good.

But the engineering reference needs changes to remove or alter the 2x area discussion.

RKStrand:
@mjwitte I missed the eng ref changes that were needed and as I look back at the code I have to admit that I'm having a bit of "coder's remorse".  The description in the Eng Ref sort of makes sense with the simple case that it presents.  Though I am still having a hard time agreeing with it fully given that E+ is not as simple as the case presented in the Eng Ref.  So perhaps the second approach would be better--leaving the original doubling of area and then adjusting the NetLWRadToSurf.  I'd like to switch to that.  This would not require any changes to Eng Ref but to the code.  I will look into that asap...but for some reason, I'm not seeing any of the commits that you made on this branch.

@Myoldmopar Any idea why I'm not seeing the commits that Mike made?

RKStrand:
@Myoldmopar Never mind...I got Mike's commits to show up.  User (my) error.

mjwitte:
@RKStrand I tried the 2xNetLWRadToSurf a while back during the airwall development and ran into some instability. But I may have done it incorrectly back then. So, let's see what you come up with.

RKStrand:
@mjwitte Ok, that's interesting.  Doubling the NetLWRadToSurf, the test input file does yield a net radiation balance in all three zones from what I saw.  However, the net radiation to individual surfaces is REALLY high (45kW for some surfaces like the IRT) and the temperatures for the IRT surfaces are not reasonable (-30C).  So, while I think I have done the correct change to double the NetLWRadToSurf for just the IRT surfaces, the end result is not what we want.  The net radiation to surfaces numbers for what I initially did in this are much closer to what the numbers for develop are and so are the temperatures for those "fake" surfaces.  I didn't see anything go unstable and cause a crash, but under other conditions that happening is a definite possibility.

So...I will adjust the language in the Eng Ref.  I think the example presented their works in a simple case, but the surface heat balance is not a simple case.  I will do what I can to adjust what's there.

RKStrand:
@mjwitte I've made some changes to the Eng Ref to reflect the changes that were made to the code with respect to the radiation exchange and committed them.  Let me know if you have any questions or concerns.

mjwitte (approved):

mjwitte:
Not waiting for more CI cycles here. Merging.

PR #7443, opened 2019-08-02
Unit Ventilator with Cycling Fan Operation Resulting in Negative Cooling Coil Loads

### LINKED ISSUE #7420: Unit ventilator with cycling fan results in simultaneous heating-cooling and negative cooling coil loads
Issue overview
--------------
This was noticed while testing #7386. A test file with variable percent outdoor air and cycling fan operation has 1000s of warnings like this:
```
   ** Warning ** GetSpecificHeatGlycol: Temperature is out of range (too low) for fluid [WATER] specific heat supplied values **
   **   ~~~   ** ..Called From:UpdatePlantLoopInterface,Temperature=[-0.67], supplied data range=[0.00,125.00]
   **   ~~~   **  Environment=RUN PERIOD 1, at Simulation time=01/01 00:00 - 00:15
```

Looking at the output reveals times when the heating and cooling coils are both operating and times when the cooling coil output is negative. For example, look at the variables output, columns BJ and BM on 01/01.

### Details
 - Version of EnergyPlus 9.2.0-7a21f42855
 - Ref #7386 
### Checklist
 - [x] Defect file attached:
[UnitVent5ZoneAuto-CycFan2-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt](https://github.com/NREL/EnergyPlus/files/3443963/UnitVent5ZoneAuto-CycFan2-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt)

RKStrand:
Pull request overview
---------------------
It was discovered that there were not only cooling coil loads happening when the unit ventilator should be in heating mode but also negative cooling coil loads when cycling fan operation was used.  It was discovered that there was flow rate past the cooling coil whenever the heating coil was on and this was somehow prompting the cooling coil to run.  This bug has now been fixed so the cooling coil does not run when there is a heating load.  A new input file that has a unit ventilator that runs in fan cycling mode and a unit test were also added.  This resolves #7420.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7420 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
@RKStrand I just pulled this branch and tried to compare the defect file to develop(ish) and this branch. I don't see any differences. Can you please post a figure such as this that shows a correction to the results. I still see negative numbers and warnings as described in #7420. Are these the same results you see on your end?

image


Also, I see these warnings in the error file:

```
   ** Warning ** GetSpecificHeatGlycol: Temperature is out of range (too low) for fluid [WATER] specific heat supplied values **
   **   ~~~   ** ..Called From:UpdatePlantLoopInterface,Temperature=[-0.67], supplied data range=[0.00,125.00]
   **   ~~~   **  Environment=RUN PERIOD 1, at Simulation time=01/01 00:00 - 00:15
   ** Warning ** GetDensityGlycol: Temperature is out of range (too low) for fluid [WATER] density **
   **   ~~~   ** ..Called From:NodeReportingCalc:CHW SUPPLY OUTLET NODE,Temperature=[-0.67], supplied data range=[0.00,100.00]
   **   ~~~   **  Environment=RUN PERIOD 1, at Simulation time=01/01 00:00 - 00:15
   ************* Simulation Error Summary *************
   ************* There are 1 unused schedules in input.
   ************* There are 1 unused week schedules in input.
   ************* There are 1 unused day schedules in input.
   ************* Use Output:Diagnostics,DisplayUnusedSchedules; to see them.
   *************
   ************* ===== Recurring Error Summary =====
   ************* The following recurring error messages occurred.
   *************
   *************  ** Warning ** GetSpecificHeatGlycol: Temperature out of range (too low) for fluid [WATER] specific heat **
   *************  **   ~~~   **   This error occurred 87595 total times;
   *************  **   ~~~   **   during Warmup 0 times;
   *************  **   ~~~   **   during Sizing 0 times.
   *************  **   ~~~   **   Max=-1.514965E-003 {C}  Min=-12.493293 {C}
```

rraustad:
I kept thinking about why these would be identical. I switched back to VS and it said the project had been modified do I want to reload. This is after I compiled this branch and ran the defect file. I clicked OK and recompiled. Not much recompilation so I thought nothing changed. Then I ran the defect file and here is what I got. So is this what you see on your end? And no errors!!!

image

rraustad:
Looking a little closer I see this. Top plot is the coil output in this branch, bottom plot simply changes the y-axis limits to better see neg numbers. Notice the little negative blip that is barely noticable at x-axis 5000 in the plot above.

image

rraustad:
Also think about this sentence in the PR description. The coils are in series...

It was discovered that there was flow rate past the cooling coil whenever the heating coil was on and this was somehow prompting the cooling coil to run.

RKStrand:
@rraustad Hmmm...this is strange.  I could have sworn that I checked all the output on the defect file for the cooling coil and that there were no negative numbers.  For the defect input file, I visually checked by scrolling through and also did a MIN on that column in the csv.  So, I thought the issue was resolved.  I will have to go back and look at this again since there shouldn't be any negative coil loads.  Which file did you run for these results?

rraustad:
I ran the defect file attached to issue #7420.
https://github.com/NREL/EnergyPlus/files/3443963/UnitVent5ZoneAuto-CycFan2-Variable-MinOAOn-Annual-SelectMonths-Detailed.idf.txt

RKStrand:
@rraustad Thanks, Rich.  I will definitely have to take a look at this again.  I'll get back to you once I've had a chance to investigate this further (hopefully over the next week?).

RKStrand:
@rraustad Ok, what I am seeing: when the fan is assigned a schedule of always_1, then the negative cooling loads don't show up.  When the fan schedule is blank for UnitVent1, then these negative loads reappear.  So, I'll look at this some more and figure out how to prevent this...

rraustad:
Some thoughts. How does a fan schedule affect the water cooling coil report for total cooling rate Q = Mdot*deltaH? Is the water coil outlet air temp really higher than the inlet air temp? Or I guess it's really the enthalpy difference, is outlet air H higher than inlet air H?

RKStrand:
@rraustad You make some good points.  There was a situation where because Mdot was not being zeroed out.  When in heating mode, the flow past the cooling coil should be zeroed out but was not for a certain fan type.  So, that is how fan type resulted in large negative cooling loads when clearly the cooling coil shouldn't be running.  Now, there are negative cooling coil loads when it appears that the unit ventilator can meet the load with outside air.  I agree with you that these very small negative cooling coil loads seems more like a reporting issue.  That's the next thing I'm going to look at.  Thanks for your suggestions and staying engaging in this defect/PR.

RKStrand:
@rraustad Ok, I think I finally have the small negative cooling coil loads eliminated.  The change I made before eliminated cooling coil when there was a heating load.  This latest change eliminates the cooling coil when outside air is handling the cooling load which for some reason I missed before.  Now, I'm getting no negative cooling coil loads.  I have committed this latest change and also updated the branch to include the latest from develop in a separate commit.  Both commits were made just now.  Hopefully the integration tests come back without any issues.

src/EnergyPlus/WaterCoils.cc
```diff
@@ -3449,8 +3449,8 @@ namespace WaterCoils {
             }
 
             // Set the outlet conditions
-            WaterCoil(CoilNum).TotWaterCoolingCoilRate = TotWaterCoilLoad * 1000.0;
-            WaterCoil(CoilNum).SenWaterCoolingCoilRate = SenWaterCoilLoad * 1000.0;
+            WaterCoil(CoilNum).TotWaterCoolingCoilRate = max(0.0, TotWaterCoilLoad * 1000.0);
+            WaterCoil(CoilNum).SenWaterCoolingCoilRate = max(0.0, SenWaterCoilLoad * 1000.0);
```
rraustad: I can't say that I like this change. Simply capping the report variable does not account for the model setting the outlet air and water temperatures. So now the report variable is correct but the node data does not correspond to this report.

```
                //        total cooling = change in air enmthalpy across coil
                TotWaterCoilLoad = AirMassFlow * (InletAirEnthalpy - OutletAirEnthalpy);
                //        conservation of energy on water stream gives water outlet
                //        temperature
                TempWaterOut = WaterMassFlowRate * Cp * ConvK; // Temp for next calc
                TempWaterOut = min(TempWaterIn + TotWaterCoilLoad / TempWaterOut, TempAirIn);
```

```
            if (WaterCoil(CoilNum).SurfAreaWetFraction == 0) {
                //       dry coil
                TempAirOut = TempAirIn - TotWaterCoilLoad / (AirMassFlow * MoistAirSpecificHeat);
                OutletAirHumRat = InletAirHumRat;
                SenWaterCoilLoad = TotWaterCoilLoad;
```

rraustad: There are many places in code like this. I would rather see a max(0,TotWaterCoilLoad) in a few places so that everything agrees with TotWaterCoilLoad = 0. Now of course if there is water flow through the coil then it is possible water and air temp is reversed of what is expected. However, when there is no cooling load the water flow rate is supposed to be 0.

RKStrand: @rraustad I see what you are saying.  I will look at this further and get back to you.

RKStrand: @rraustad Ok, I undid the previous set of changes that simply zeroed out the output variables.  Now, during the simulation of the coil type in question, if it says that the cooling coil is heating (negative load) or the air temperature or humidity ratio increase across the cooling coil (also a sign that it is heating), then it zeroes out the coil load and transfers inlet conditions to the outlet side for consistency.  Testing shows that this does eliminate the negative cooling coil loads in the defect input file.  Hope this takes care of everything!

rraustad:
OK, this change is acceptable and gives time to think of a better approach if one exists. Defect file now shows no negative cooling.

image

src/EnergyPlus/UnitVentilator.cc
```diff
@@ -3433,7 +3433,7 @@ namespace UnitVentilator {
 
             if (UnitVent(UnitVentNum).CCoilPresent) {
 
-                mdot = UnitVent(UnitVentNum).MaxColdWaterFlow * PartLoadRatio;
+                mdot = CalcMdotCCoilCycFan(QZnReq,UnitVentNum,PartLoadRatio);
```
rraustad: What is the argument for changing a 1 line calculation to a multiple line function call? QZnReq is tested outside of CalcUnitVentilatorComponents and PartLoadRatio is passed in accordingly. Did you find something while testing that warranted this? Was this a first try that is no longer needed?

src/EnergyPlus/WaterCoils.cc
```diff
@@ -3676,6 +3676,18 @@ namespace WaterCoils {
                 }
             }
 
+            if ((TotWaterCoilLoad <= 0) || (OutletAirTemp >= WaterCoil(CoilNum).InletAirTemp) ||
+                (OutletAirHumRat >= WaterCoil(CoilNum).InletAirHumRat)) {
```
rraustad: This seems reasonable. I thought that testing TotWaterCoilLoad would be sufficient but this doesn't appear to hurt anything (except speed). Can OutletAirTemp or OutletAirHumRat show heating with TotWaterCoilLoad > 0? We need to think about whether we want a cooling coil to provide heating, for example if the plant locks water  flow > 0 when there is no cooling load. It's possible via the SetComponentFlowRate call. Beyond the scope here but something to think about for the future.

rraustad:
There is still time while CI runs through.
1) Can the new function be pulled out for speed consideration?
2) Can "if (TotWaterCoilLoad <= 0)" be used to spawn variable reset instead of the more detailed test?
3) Unit test would need re-write if 1) is reverted

rraustad:
I ran the defect file with a break point inside the if test to see when this happens.

Here's a problem when there is a cooling load, TotWaterCoilLoad = 1845 W.
OutletAirHumRat = WaterCoil(5).InletAirHumRat since it's a dry coil. 
I think the test here is overly restrictive and should just be if (TotWaterCoilLoad < 0.0). A TotWaterCoilLoad = 0 should work itself out. If TotWaterCoilLoad > 0, OutletAirTemp will be less than inlet and OutletAirHumRat will be <= inlet (and even if it's not it's what the model predicted, a different issue).

image

src/EnergyPlus/UnitVentilator.cc
RKStrand: @rraustad It was the first attempt and it eliminated some very large negative cooling coil loads that were happening when there was actually a heating load.  The point of the function was to simplify the code here with the introduction of a function.  It could have been done here rather than in a function.  Is it still needed?  I think so because the original condition was large negative cooling loads and this definitely eliminated that problem.  Maybe with the new code in the cooling coil this won't be necessary anymore, but I haven't tested that.  Didn't seem necessary since it did fix one of the bad conditions.

RKStrand:
@rraustad I can definitely make the conditions on the new code less restrictive and only look at if TotWaterCoilLoad <= 0.  I was trying to be more complete in the if condition because it seemed like other weird conditions should also be.  As for undoing the first code, I guess I can try that and see what it yields.  If that gets backed out, then yes a new unit test will be needed.

rraustad:
I ran this again with your new if test commented out. When the break point stops at TotWaterCoilLoad < 0 I see that outlet air conditions are good so what caused the coil load to go negative?

image

rraustad:
So I looped back around to the coil model calls and find that inlet air temp < inlet water temp. What do you think Q will be?
The model shouldn't even be here under those conditions. So this can probably be caught well before TotWaterCoilLoad < 0.

image

rraustad:
This seems to me to be a unit ventilator control problem. I'm seeing inlet air temp at 23C and outdoor air temp of 3C with a high OA fraction. Nowhere do I see a check on inlet water temp. I'm not saying the water coil model could not be beefed up, but that model has been around for decades. The variable speed control of the unit ventilator is new and could have logic issues.

rraustad:
More thoughts after I saw this last night. None of the systems models (Unit vent, UnitarySystem, etc.) check the water temp vs air temp as a guide for operating the coils. They just model with PLR=0 and PLR = 1 and compare results to QZnReq. So in this case the water coil entering air temp < water coil entering water temp and the water coil model shows an outlet air temp slightly higher than the entering air temp. This is correct, at least as far as a HX component model goes. Now all this happens before the system model decides what water flow rate to operate at this time step.

So it seems to me that the parent models need to decide if the coil model should even operate under those conditions (I hate that because it means more logic in all the system models, please think of something else). If not, the parent would set coil water flow rate = 0 and the problem would disappear. What would a real system do in this case? Real water coils are typically set point based control (whereas most of the system models are load based) and the cooling valve controller would be reverse acting, so the valve would shut down (outlet air set point higher than inlet air temp).

I'm not sure if lumping this in the water coil model would work correctly because the parent models would set PLR=1 if the coil could not meet the load at PLR=0. So that leaves the parent model logic changes.

rraustad:
... and this happened in Unit Ventilator because of the OA fraction (I think). I wonder if the OA flow rate is adjusted for cycling fan? I do recall a very low PLR to meet QZnReq = -32 W when the zone temp was 23C and OA temp was 3C. I don't recall what the OA flow rate was compared to the system flow rate. Just something else to look at.

RKStrand:
@rraustad Thanks for all of your thoughts on this.  I will dig a little deeper.  I'm not sure that a big solution that goes across several different equipment types is the solution.  This seems to be a problem in the unit ventilator so I will look at that further.  If there were other systems that were exhibiting this behavior, then I would think that this is a bigger deal that requires a more comprehensive solution.  If it gets to that point, I will have to move on at least temporarily.

rraustad:
Certainly not saying to fix other system types (that would be a follow up issue), just pointing out that this issue may affect more than the Unit Ventilator because us developers never thought of this before.

RKStrand:
@rraustad Well, I think I have a solution now.  I backed out the changes to the water coil because that isn't where the problem originates.  It is a unit ventilator problem for the cycling fan.  It was trying to load the cooling coil based on the zone load remaining but wasn't factoring in the possibility that outside air was actually providing some cooling.  It seems like that was resulting in some small negative cooling coil loads.  I've adjusted the logic and now I'm getting no negative coil loads.  Strangely enough, when there is a heating load and heating coil, there was logic to handle the impact of outside air already there.  Maybe an oversight in implementation?

RKStrand:
@rraustad Ok, this is ready for the next review.

rraustad:
Review step 1: Are negative coil loads reported? No!

image

rraustad:
Review step 2 - Release version of unit tests run to completion. Check!
However, the Mac CI shows a unit test failure that cannot be reproduced on Windows. For this reason, this PR is on temporary hold unit the Mac issue for unit testing is identified/resolved.
UnitarySystemModel_TwoSpeedDXCoolCoil_Only

RKStrand:
@Myoldmopar Any reason why the Mac would have an issue?  Is this one of these blips that will go away if it gets re-run?

Myoldmopar:
That is **not** the test failure that occasionally pops up.  Seems legit for now.  I'll build on my Mac right now and see if I can verify.

rraustad:
Review Step 3 - the code change addresses the underlying issue. I would rather see a test for coil inlet air vs water temp happen sooner, however, the PLR is unknown and mixing of OA with return can't happen unless PLR is known. So this seems a reasonable approach. I get an eerie feeling about discontinuity with this change (e.g., coil on above PLR = 0.1, off otherwise, or vice versa) but have no better solution and not really sure if a discontinuity of this kind is a real problem. Code changes approved.

rraustad:
Review Step 4 - diff's review pending, waiting on CI results.

Myoldmopar:
I am indeed getting a unit test failure with this branch.  However, it is failing in ZoneUnitarySysTest.Test_UnitarySystemModel_factory.  It's failing saying that unitarySys.size() is a wildly invalid number.  Something is going awry here.  If you run the unit tests locally do you see this?

rraustad:
All run fine here. And these changes have nothing to do with UnitarySystem, although that failure needs to be understood.

image

rraustad:
Diff's from Coil Sizing report fixes #7435 are creeping in here.

rraustad:
Oops, wrong PR. I meant to say this for the SET PR where I see coil sizing diff's.

Myoldmopar:
Mac is reporting all happy again here.  If Windows and the Linux integration builds come back clean this will merge.

rraustad:
Agreed. Leaving in your hands.

rraustad (approved):
CI issue now appears to be resolved. Code/Diff review complete.

PR #7444, opened 2019-08-06
Ceiling Diffuser Convection Coefficient Transition from Natural to Forced Convection

### LINKED ISSUE #7423: Forced to natural convection transition for "Ceiling Diffuser" convection model
Issue overview
--------------
EnergyPlus v9.1 fixed a defect in the "Ceiling Diffuse" convection model (https://github.com/NREL/EnergyPlus/issues/7208).  The intended operation of this model was to use forced convection coefficients for air circulation rates of more than 3 ACH, and to use natural convection coefficients for air circulation rates less than 3 ACH.  Previously the model was using forced convection coefficients for all air circulation levels, but apparently this was not intended.
    There are two issues with this return to "intended" operation of the model.  First the 3 ACH threshold needs to be validated.  Second, the change from forced to natural convection is currently a step change rather than a smooth transition.  Its unclear whether this represents actual building physics.  
   Validation: The basis for the 3 ACH threshold is not clear.  Talking with EnergyPlus developers, a source for the 3 ACH threshold could not be cited.  A large fraction of VAV air systems have VAV box minimum flow rates less than 3 ACH so heating season and heating design operation is affected by the 3 ACH threshold.  Heating loads and therefore heating energy consumption is significantly lower when assuming natural convection rather than forced convection in these zones.  Having a strong, documented basis for the 3 ACH threshold is therefore very important for the accuracy of EnergyPlus and the credibility of the engine as EnergyPlus gains wider use among engineers doing "production" work - that is creating equipment specifications for commercial projects and doing energy modeling to optimize system designs.
    Transition:  The second aspect of this is whether the change from forced to natural convection is actually a step change that occurs a particular ACH threshold, or is a smooth transition that occurs across an interval between two ACH levels.  Having a strong, documented basis for the transition modeling is also important for accuracy and credibility of EnergyPlus results.
    It is possible initial FlexLAB test results can help with validation of the model.  However, talking with the researchers, it appears much more test work is needed to fully validate the threshold and to understand how transition between convection modes occurs.

### Details

### Checklist

RKStrand:
Pull request overview
---------------------
In the ceiling diffuser algorithm, there is a change from the actual ceiling diffuser algorithm to natural convection when the ACH drops below 3.  The problem is that the ceiling diffuser algorithm does not yield the same value as the natural convection algorithm at ACH=3 leading to a discontinuity in the convection coefficient calculated.  This work resolves this by using an interpolation between the two models that starts at ACH=2.5 and ends at ACH=3.  This eliminates the discontinuity.  This resolves #7423 

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7423 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


RKStrand:
@Myoldmopar Suggested you as a reviewer but feel free to assign to someone else if you wish.  The only diffs are in the _SmallOffice_Dulles input file that uses the CeilingDiffuser model.  The input file defaults to monthly and annual results which accentuates the diffs in the files.  The diffs are fairly minor though and when I looked at detailed results, the diffs in surface temperatures initially at least are around 0.01C.  So, I think that's ok given the change, but I will wait for a reviewer to weigh in.

rraustad (approved):
Straight-forward update and good unit test.

rraustad:
Reviewed and approved.

Issue #7269, opened 2019-04-12
User file with Site:GroundDomain:Basement failing with NaN


mjwitte:
Issue overview
--------------
User file has Site:GroundDomain:Basement fails with a thermostat error message showing NaNs in the error. The problem is with the horizontal insulation material which is Material:NoMass which results in a zero thickness domain region during mesh development. This could likely be a problem with any of the insulation materials in Site:GroundDomain:*

Workaround
------------
Delete the insulation, or use a Material object which has thickness.

### Details
Some additional details for this issue (if relevant):
 - Win64
 - Version of EnergyPlus v9.1.0
 - Helpdesk ticket number 15127.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added in EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

RKStrand:
@mjwitte It definitely cannot use a no mass material because of the grid spacing issue as you suggested.  Though on the Mac, it never gets to the NaN problem but crashes on line 3983 of PlantPipingSystemsManager.cc:

            assert(g.Max > g.Min);

where it fails this assert.  Changing the material to a non-no mass material allows it to run.  Two potential solutions include:

1. Flagging this on input and forcing the user to pick a material with mass.
2. Check the materials for thickness and assigning a nominal thickness if it is a no mass material.
3. Altering the ground heat transfer model to do something differently when a no mass material is selected.

My thoughts: I think 3 is way too ambitious and would probably require more effort than this is worth.  2 is probably not awful but a bit artificial.  I believe that 1 is the best option for right now and should be implemented in both Site:GroundDomain get input routines.  I will move forward with that solution unless I heard concerns about that approach.

I'm "CC-ing" @Myoldmopar here since his name is on at least some of the code in this area in case he had any additional thoughts.

Myoldmopar:
Closed via #7446

PR #7446, opened 2019-08-08
Resolve Site:GroundDomain:* Issues with R-Only Materials

RKStrand:
Pull request overview
---------------------
EnergyPlus was failing during situations where a user chose a "no mass" material (defined by only an R-Value) as the insulation material for the underground slab or basement model.  This is tied to the fact that the model uses the thickness of the insulation material to define a solution mesh.  The solution here was to not allow users to use no mass materials for horizontal or vertical insulation in these slab or basement models and require them to using a regular material which includes a designation of thickness, specific heat, etc.  The new error message alerts the user to the need to use a regular material rather than a no mass material.  This resolve GitHub Issue #7269.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #7269 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/PlantPipingSystemsManager.cc
```diff
@@ -946,6 +946,10 @@ namespace EnergyPlus {
                                 thisDomain.HorizInsMaterialNum).SpecHeat;
                         thisDomain.HorizInsProperties.Conductivity = DataHeatBalance::Material(
                                 thisDomain.HorizInsMaterialNum).Conductivity;
+                        if (SiteGroundDomainUsingNoMassMat(thisDomain.HorizInsThickness,thisDomain.HorizInsMaterialNum)) {
+                            ErrorsFound = true;
+                            SiteGroundDomainNoMassMatError(DataIPShortCuts::cAlphaFieldNames(8),DataIPShortCuts::cAlphaArgs(8),thisDomain.Name);
+                        }
```
Myoldmopar, line 952: At first I was going to suggest that the no-mass check function and the error function should be methods on the domain class itself so that you didn't have to pass args.  However, I see now that the function would need to switch based on which condition it is checking, and would be overly complicated.  On the other hand, in this case, the function is super simple and you just pass in a couple arguments and it is all good.

Myoldmopar (approved):
I am verifying this locally, but this looks like a good fix.  Thanks.

Myoldmopar:
Verified the NaN defect with develop branch locally.  Verified this branch provides a meaningful error message instead and gracefully exits.  I also fixed a couple small things that I am pushing up now and this branch will be merged.  Thanks @RKStrand

PR #7447, opened 2019-08-08
Correct AirflowNetwork zone Outdoor Air conditions for local environment when no schedules exist in OutdoorAir:Node object

### LINKED ISSUE #7308: Outdoor air node conditions are not used correctly for zone local environment and for airflow network
Issue overview
--------------
User file attempts to adjust infiltration heat gain by using EMS to change the temperature of an `OutdoorAir:Node` which is used as the external node on an `AirflowNetwork:MultiZone:Surface`. The node temperature changes as expected, but there is little or no impact on the infilration heat gain/loss. Investigation into this revealed several issues:

1. Calculation of [AFN Zone Infiltration Sensible Heat Gain/Loss](https://github.com/NREL/EnergyPlus/blob/7a7082a357ad875ac3f0d9dd66564a85cae454d2/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L8252-L8272) uses `Zone(ZN1).OutDryBulbTemp` instead of the external node temp. This appears to be just a reporting issue - these values aren't used in the heat balance.

2. The [AFN heat balance calculation](https://github.com/NREL/EnergyPlus/blob/7a7082a357ad875ac3f0d9dd66564a85cae454d2/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L6835-L6838) uses `Node.OutAirDryBulb`. Why? Why not `Node.Temp`? Don't understand why `Node.OutAirDryBulb` is even a variable - there's no output:variable associated with it. `SetOANodeValues` sets `Node(NodeNum).Temp = Node(NodeNum).OutAirDryBulb` so it's probably ok, but why confuse things by having two dry bulb temp variables on the same node structure?

3. So, tried to use `ZoneProperty:LocalEnvironment` to associated the same `OutdoorAir:Node` with the zone in question (R01). There was some impact on infiltration heat gain/loss, but the reported values for R01:Zone Outdoor Air Drybulb Temperature are all zero. This is partly due to [this code for local environment](https://github.com/NREL/EnergyPlus/blob/7a7082a357ad875ac3f0d9dd66564a85cae454d2/src/EnergyPlus/HeatBalanceManager.cc#L5335-L5346) which is grabbing from the outdoor air node schedules rather than taking the OA node conditions directly.  This block should look like the [equivalent one for surfaces](https://github.com/NREL/EnergyPlus/blob/7a7082a357ad875ac3f0d9dd66564a85cae454d2/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L448-L454).

### Details
 - Helpdesk ticket number 15158.

### Checklist
...SmartWindow-TVent1VeryCold-ZoneOANode.csv - Col A shows "R01:Zone Outdoor Air Drybulb Temperature [C](Hourly)" always zero.
...SmartWindow and ..._SmartWindow-TVent1VeryCold - Col F "R01:Zone Air Heat Balance Outdoor Air Transfer Rate" is similar in both, but there should be much more heat loss in "VeryCold" file.  These both actuate the outdoor air node only.
...SmartWindow-TVent1VeryCold-ZoneOANode-ActuateBoth - Col F shows much more heat loss.

After this is fixed, I would expect these three files to have the same results:
..SmartWindow-TVent1VeryCold
...SmartWindow-TVent1VeryCold-ZoneOANode.csv 
...SmartWindow-TVent1VeryCold-ZoneOANode-ActuateBoth (this one shouldn't change with this fix, 90% certain of this?)

lgu1234:
Pull request overview
---------------------
The correction fixes AirflowNetwork zone outdoor air properties for local environment when there are no inputs of outdoor air property schedules in OutdoorAir:Node. If no scheduled outdoor air properties exist in the outdoor air nodes, the air conditions (T and w) are assigned to zone outdoor air properties. This fixes #7308.

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


lgu1234:
@mjwitte I fixed 3rd issue to assign outdoor air node properties into zone outdoor air properties if there are no inputs of outdoor air property schedule names in the OutdoorAir:Node.  The main problem is that there no inputs of schedule name, so that schedule point = 0 gets 0 values under local environment. After the fix, the value of SmartWindow-TVent1VeryCold-ZoneOANode.csv - Col A shows "R01:Zone Outdoor Air Drybulb Temperature C" are correct. The results in both test files (.SmartWindow-TVent1VeryCold-ZoneOANode.csv and 
...SmartWindow-TVent1VeryCold-ZoneOANode-ActuateBoth) are the same. However, differences between SmartWindow-TVent1VeryCold and SmartWindow-TVent1VeryCold-ZoneOANode are expected, because SmartWindow-TVent1VeryCold uses real outdoor conditions in Zone Outdoor Air Drybulb Temperature, while SmartWindow-TVent1VeryCold-ZoneOANode uses EMS overidden values for Zone Outdoor Air Drybulb Temperature.

Your 2nd issue is when both values are the same, why should we use a single variable, instead of 2? I think @XuanLuoLBL introduced LocalEnvironment, she should have some justification for this.

lgu1234:
The previous check-in address 1st issue to use external node temperature to calculate infiltration gains or losses, instead of zone outdoor temperature for local environment.

It is ready for review.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -8306,8 +8316,14 @@ namespace AirflowNetworkBalanceManager {
...
-                    Tamb = Zone(ZN2).OutDryBulbTemp;
-                    CpAir = PsyCpAirFnWTdb(OutHumRat, Tamb);
+                    if (Surface(MultizoneSurfaceData(i).SurfNum).HasLinkedOutAirNode) {
+                        Tamb = Surface(MultizoneSurfaceData(i).SurfNum).OutDryBulbTemp;
+                        CpAir = PsyCpAirFnWTdb(
+                            Psychrometrics::PsyWFnTdbTwbPb(Tamb, Surface(MultizoneSurfaceData(i).SurfNum).OutWetBulbTemp, OutBaroPress), Tamb);
+                    } else {
+                        Tamb = Zone(ZN2).OutDryBulbTemp;
+                        CpAir = PsyCpAirFnWTdb(OutHumRat, Tamb);
+                    }
```
rraustad, line 8324: So now code will use the linked OAT, and if not linked then the weather OAT.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5345,10 +5345,26 @@ namespace HeatBalanceManager {
...
+                    if (Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindSpeedSchedNum > 0) {
+                        Zone(ZoneNum).WindSpeed = GetCurrentScheduleValue(Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindSpeedSchedNum);
+                    } else {
+                        Zone(ZoneNum).WindSpeed = Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindSpeed;
+                    }
+                    if (Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindDirSchedNum > 0) {
+                        Zone(ZoneNum).WindDir = GetCurrentScheduleValue(Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindDirSchedNum);
+                    } else {
+                        Zone(ZoneNum).WindDir = Node(Zone(ZoneNum).LinkedOutAirNode).OutAirWindDir;
+                    }
```
rraustad, line 5363: And here the schedule value will be used, else the weather file data.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
```diff
@@ -1278,6 +1278,24 @@ TEST_F(EnergyPlusFixture, HeatBalanceManager_TestZonePropertyLocalEnv)
...
+
+    // Add a test for #7308 without inputs of schedule names
+    DataLoopNode::Node(1).OutAirDryBulbSchedNum = 0;
+    DataLoopNode::Node(1).OutAirWetBulbSchedNum = 0;
+    DataLoopNode::Node(1).OutAirWindSpeedSchedNum = 0;
+    DataLoopNode::Node(1).OutAirWindDirSchedNum = 0;
+    DataEnvironment::OutDryBulbTemp = 25.0;
+    DataEnvironment::OutWetBulbTemp = 20.0;
+    DataEnvironment::WindSpeed = 1.5;
+    DataEnvironment::WindDir = 90.0;
```
rraustad, line 1290: Poor unit test, the outdoor values (25, 20, 1.5 and 90) are the same as what the schedule values were. How do you know that the correct values were used?

```
    ScheduleManager::Schedule(1).CurrentValue = 25.0;
    ScheduleManager::Schedule(2).CurrentValue = 20.0;
    ScheduleManager::Schedule(3).CurrentValue = 1.5;
    ScheduleManager::Schedule(4).CurrentValue = 90.0;
```

rraustad:
@lgu1234 if you would resolve the conflicts and address my comment this can be merged.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
rraustad: OK, so the schedule values were 0'd out so this is sort of OK.

```
    // Add a test for #7308 without inputs of schedule names
    DataLoopNode::Node(1).OutAirDryBulbSchedNum = 0;
    DataLoopNode::Node(1).OutAirWetBulbSchedNum = 0;
    DataLoopNode::Node(1).OutAirWindSpeedSchedNum = 0;
    DataLoopNode::Node(1).OutAirWindDirSchedNum = 0;
    DataEnvironment::OutDryBulbTemp = 25.0;
    DataEnvironment::OutWetBulbTemp = 20.0;
    DataEnvironment::WindSpeed = 1.5;
    DataEnvironment::WindDir = 90.0;
```

rraustad: Next time think about using unique numbers so there is no ambiguity in the unit test results.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
lgu1234: Yes.

src/EnergyPlus/HeatBalanceManager.cc
lgu1234: Yes.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
lgu1234: OK.

rraustad:
CI status is all green, which is expected. Code review looks good, unit test looks good.

rraustad:
@lgu1234 I don't see an issue related to this pull request or see a reference to an issue in the PR description. Is there an issue that needs to be closed?

rraustad:
Never mind, I found the issue, #7308.

Issue #6950, opened 2018-09-07
VRF Temperature Control outputs don't add up


mjwitte:
Issue overview
--------------
Defect file has AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl:HR with several terminal units with no outdoor air.
Cooling Coil Total Cooling Rate *should* equal Zone VRF Air Terminal Total Cooling + Fan Electric Power but it does not.

For AirConditioner:VariableRefrigerantFlow, these do add up.  Have not tested AirConditioner:VariableRefrigerantFlow:FluidTemperatureControl.

Also, the I/O Ref description for VRF Heat Pump Total Cooling Rate is incorrect.
"This value should match the sum of the individual zone terminal unit output variables for ~Zone VRF Air Terminal Total Cooling Rate.~"
should be
""This value should match the sum of the individual zone terminal unit *cooling coil* output variables for *Cooling Coil Total Cooling Rate.*"

### Details
Some additional details for this issue (if relevant):
 - Platform Win64
 - Version of EnergyPlus v9.0.0-111432f7f1
 - Helpdesk ticket number 12357

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

rraustad:
Closed via #7449.

PR #7449, opened 2019-08-08
VRF Temperature Control outputs don't add up

Nigusse:
Pull request overview
---------------------
Addressed issue #6950. This fix addressed mismatch between the VRF unit DX Coils output energy rates and the sum of the VRF air terminal ouput energy rate and fan power when there is OA flow rates. The fix impacts system curve and physics based VRF models. Diffs are expected both VRF models but physics based model is expected to show large diffs because of another issue in the DX cooling coil calculation. The ten example files with VRF model in the test suite are impacted by this change.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
Two different causes were identified for mismatch between Cooling Coil Total Cooling Rate and the sum of Zone VRF Air Terminal Total Cooling + Fan Electric Power when there is no OA air. These issues were:
-  the sensible and latent split calculation of the total cooling rate delivered. This issue exists in all VRF models.
- incorrect total cooling rate calculation in DX cooling coil model (CalcVRFCoolingCoil_FluidTCtrl). This impacted physics based model only.

Nigusse:
Added defect files created from energyplus example files.

rraustad:
@Nigusse this branch is now conflicted. Please address that and also post a figure showing the reports now add up.

Nigusse:
@rraustad The unit tests demonstrate those that they add up.

rraustad:
Good point, I'll accept that as proof that the VRFTU and DXCoil outputs now match.

rraustad:
I was worried that changing the sensible calculation would mean that the TU would not hit the zone temperature set point so I tested both versions of VRF TU. Things still look good.

image

rraustad:
There are unacceptable diff's. Please review the diff's and provide an explanation.

DOASDXCOIL_wADPBFMethod now shows a negative sensible capacity (-4539) for the Space 1 coil.

image

Nigusse:
@rraustad I am going to look into it today.

Nigusse:
@rraustad This branch addresses VRF coils and VRF air terminal units. The sizing issue you pointed out here is for a different type of coil (Coil:Cooling:DX:TwoStageWithHumidityControlMode). Also I just run the example file using the latest develop and I see the same sizing problem. I do not think this issue is related to this fix.

rraustad:
The only thing I can think to do is to post a new issue and track down why this coil type (and possibly others) has changed. I assume it's a recent change but I'm not sure how to determine what code change caused this. @mjwitte @Myoldmopar any suggestions?

rraustad:
Yes, this is new. I do not see this same sizing problem in V9.1.
image

mjwitte:
@rraustad To track down when it changed, if the relevant files don't show an obvious commit that might be the culprit, then you'll have to resort to good ol' brute force interval-halving on develop.

rraustad:
No, develop shows a positive value and matches what was reported for V9.1. It's this coil that has a negative number.
image

the 3rd coil down the list at the far right for Coil Sensible Capacity at Rating Conditions = -4539.14. Develop shows this value as +4534.64. This data is 10 columns from the end of the Coil Sizing Details table.

image

Nigusse:
This branch is the offending code, I know where the problem is and will fix it shortly.

Nigusse:
@rraustad Now after the fix, I see rated sensible cooling coil capacity difference in the range from 3.0 to 6.0 W (between develop and this branch). This magnitude of differences are expected due to this fix.

rraustad:
This looks better now. Diff's are related to change in sensible cooling calculation as expected. Approved for merge.

rraustad (approved):
Reviewed code and unit test.

rraustad:
@Nigusse I looked at the only diff's with big MTR diff's, DOAToVRF.idf. Why would changing the sensible calculation have such a large impact on peak demand? About 60 W for heating and 50 W for cooling.

BigDiffs

rraustad:
It looks like DOAToVRF is controlling worse now in Zone 4 and 5. So maybe that would explain the diff's on kW?

image

rraustad:
@Nigusse I got the data switched and have corrected the plot above. The zone temperature control is affected by these changes. Maybe correct now or there could be a problem, not sure of the cause. I suspect the supply side AirTerminal:SingleDuct:Mixer connected to space 4 and 5.

rraustad:
@Nigusse there is also descriptions of psychrometric functions in the ModuleDeveloper guide. The 2 new psyc functions need to be described there.

Nigusse:
Updated zone air temperatures. The three zones Space1-1 through Space3-1 were connected to inlet-side air terminal VRF unit. Shows good match.

InletSideConnected_VRFUnit

Nigusse:
Updated zone air temperatures. The two zones Space4-1 and Space5-1 were connected to supply-side air terminal VRF unit. Shows good match, after correcting mass flow rate used for load met calculation. This was inadvertently changed variable.

SupplySideConnected_VRFUnit

rraustad:
DOAToVRF example file annual run increases energy use by about 0.5% and peak demand increase of same magnitude of 0.5% or around 50 W. This seems acceptable for a change in calculated sensible capacity.

rraustad:
It would be nice to know the math behind the data in eplusmtr.csv.diffsummary.csv and this:
Max absolute diff: 500.445608139, field: Electricity:Facility [J](Monthly), time: July, relative: 4.83337655435e-07

If the absolute difference is 500 J, then how can the relative diff be 5E-7 ?

rraustad:
Now I see only eio diff's for warmup convergence and table diff's of 10's of change in Coil Sensible Capacity at Rating Conditions with 100's of change in the new VRF model for sensible and total capacity at rated conditions.

rraustad:
This single change likely responsible for the 100's change in sensible/total capacity at rated conditions.

```
DXCoil(DXCoilNum).SensCoolingEnergyRate = AirMassFlow * 1005.0 * (InletAirDryBulbTemp - OutletAirTemp);
```

to:

```
DXCoil(DXCoilNum).SensCoolingEnergyRate = AirMassFlowRate * PsyDeltaHSenFnTdb2W2Tdb1W1(InletAirDryBulbTemp, InletAirHumRat, OutletAirTemp, OutletAirHumRat); // sensible {W};
```

rraustad:
Not sure if CI is complete. This looks ready to go.

rraustad:
Yeah, CI is done (refreshed). Merging.

PR #7452, opened 2019-08-08
7445 Allow evap cooler in air loop to cycle with air loop flow fraction

### LINKED ISSUE #7445: Evap cooler in air loop not cycling using RTF of on/off fan built in unitary system
Issue overview
--------------
If an evaporative cooler is put in AirLoopHVAC system with an on/off fan built inside a dummy unitary system (because cycling on/off fan is not currently supported to be built directly in AirLoopHVAC  [#6613](https://github.com/NREL/EnergyPlus/issues/6613)), evaporative cooler is not able to cycle using RTF of cycling fan. 

@Nigusse @rraustad @mjwitte Might have thought or suggestion?

[Defect test file](https://drive.google.com/file/d/1ns95Wv2mQA49qpDD3OuIQUsPq0qJ5DLD/view?usp=sharing)

### Details
 - Platform (Windows, 10)
 - Version of EnergyPlus (9.1.0)

### Checklist

yzhou601:
Pull request overview
---------------------
fix   #7445 

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12853,7 +12853,7 @@ namespace UnitarySystems {
                 if (this->m_FanOpMode == DataHVACGlobals::CycFanCycCoil) {
                     DataAirLoop::AirLoopFlow(AirLoopNum).FanPLR = this->FanPartLoadRatio;
                 } else {
-                    DataAirLoop::AirLoopFlow(AirLoopNum).FanPLR = 0.0;
+                    DataAirLoop::AirLoopFlow(AirLoopNum).FanPLR = 1.0;
```
rraustad, line 12856: Also should default to 1 in DataAirLoop.hh

```
        // Default Constructor
        AirLoopFlowData()
            : DesSupply(0.0), DesReturnFrac(1.0), SysToZoneDesFlowRatio(0.0), ReqSupplyFrac(1.0), MinOutAir(0.0), MaxOutAir(0.0), OAMinFrac(0.0),
              Previous(0.0), SupFlow(0.0), ZoneRetFlow(0.0), ZoneRetFlowRatio(1.0), SysRetFlow(0.0), RecircFlow(0.0), LeakFlow(0.0),
              ExcessZoneExhFlow(0.0), FanPLR(0.0), OAFrac(0.0), OAFlow(0.0), FlowError(false), BypassMassFlow(0.0)
        {
        }
```

yzhou601:
**EvaporativeCooler:DirectCelDekPad**

image
image


**EvaporativeCooler:DirectResearchSpecial**

image
image

In current E+, research special evap coolers are cycling based on meeting setpoint temperature, not following mass flow ratio if there's no modifier curve specified. These types of evap coolers are skipped when fixing cycling issue in [PR7319](https://github.com/NREL/EnergyPlus/pull/7319). Do we want to change pump energy calculations for those types as well? @rraustad @mjwitte @Nigusse

src/EnergyPlus/UnitarySystem.cc
yzhou601: Done

rraustad:
I would think if the evap cooler cycles with the air system, then yes the power should modulate. But doesn't this code already do that?

```
void CalcDirectResearchSpecialEvapCooler(int const EvapCoolNum)

            if (EvapCond(EvapCoolNum).PumpPowerModifierCurveIndex > 0) {
                PumpPowerModCurveValue = CurveValue(EvapCond(EvapCoolNum).PumpPowerModifierCurveIndex, FlowRatio);
            } else {
                // if no pump power modifier curve specified, then assume linear variation with part-load
                PumpPowerModCurveValue = PartLoad;
            }
            EvapCond(EvapCoolNum).EvapCoolerPower = EvapCond(EvapCoolNum).RecircPumpPower * PumpPowerModCurveValue;
```

yzhou601:
> I would think if the evap cooler cycles with the air system, then yes the power should modulate. But doesn't this code already do that?
> 
> ```
> void CalcDirectResearchSpecialEvapCooler(int const EvapCoolNum)
> 
>             if (EvapCond(EvapCoolNum).PumpPowerModifierCurveIndex > 0) {
>                 PumpPowerModCurveValue = CurveValue(EvapCond(EvapCoolNum).PumpPowerModifierCurveIndex, FlowRatio);
>             } else {
>                 // if no pump power modifier curve specified, then assume linear variation with part-load
>                 PumpPowerModCurveValue = PartLoad;
>             }
>             EvapCond(EvapCoolNum).EvapCoolerPower = EvapCond(EvapCoolNum).RecircPumpPower * PumpPowerModCurveValue;
> ```

@rraustad Yes if there's a modifier curve specified for evap cooler, it would account for mass flow ratio, in this test file, it is the `else` case where a `PartLoad` is used which is calculated based on how much should evap cooler run to meet outlet setpoint temperature (not for the purpose of addressing mass flow ratio of air system). 
However in modifier curve case, it seems to be ignoring the `PartLoad`. These two PLRs are confusing, didn't know which one(or two) to use for pump power reporting...

rraustad:
The changes in this branch look correct to me. You are now passing the air loop PLR which will be used to modify power. If a modifier curve is used, then the user has control over how power changes. Fix the conflicts and this is probably good to go.

rraustad:
Wait a minute. You don't even need to pass PLR to the research special evap cooler. The function CalcResearchSpecialPartLoad calculates the PLR which is used to calculate power. So now I think there is no need for this change.

yzhou601:
> Wait a minute. You don't even need to pass PLR to the research special evap cooler. The function CalcResearchSpecialPartLoad calculates the PLR which is used to calculate power. So now I think there is no need for this change.

@rraustad Yes it calculated a PLR but is for meeting the outlet setpoint temperature. So it's actually not cycling with air loop flow ratio but for guaranteeing the outlet temperature controlled by setpoint manager, we could think that other types always run in full load in terms of the outlet temperature, and then adjusted by **air flow ratio** as PLR. Does that make sense? 
Actually we are trying to build models with ResearchSpecial evaporative cooler in the air loop and wish pump to report correct energy consumption, so it's very important  for us to dig into this point. 🙏

rraustad:
Well, the PLR needed to meet the outlet temp set point IS the cycling PLR, if the evap cooler is cycling. If the evap cooler runs the entire time step then the only way to meet the set point is to reduce water flow rate, which also seems to be this same PLR. So if this is not the PLR you are interested in, then what is?

rraustad:
Or asked another way, how does the air loop PLR differ from the evap cooler internal calculated PLR?

yzhou601:
> Or asked another way, how does the air loop PLR differ from the evap cooler internal calculated PLR?

@rraustad , the *air loop fanPLR* we are passing is based on **load** or **thermostat** of conditioned zones, however this *evap cooler internal calculated PLR* is to control **evap cooler outlet temperature** (which is very similar to supply air temperature, 18C in above case), so this is the reason the above plots show different shapes. Only those hours evap cooler has capacity to cool air to below 18C, it would give an *internal PLR* below 1.0. 
The air loop and the "water loop"(actually not exist, in evap cooler, it just calculates pump power and water consumption, there's no real plant loop) are working separately. So do these two PLRs, I want to find a way to better combine these two to make pump power more representative to its cycling operation.

rraustad:
OK, so I ask again, which PLR do you wish to control the evap cooler power? And if air loop PLR < internal evap cooler PLR, then will you reduce the moisture evap rate?

yzhou601:
> OK, so I ask again, which PLR do you wish to control the evap cooler power? And if air loop PLR < internal evap cooler PLR, then will you reduce the moisture evap rate?

I think what @Nigusse was doing for other types is to use air loop PLR, which means pump is cycling in the same way as fan. But for researchspecial type if the outlet temperature is not conditioned by full load (as others do), maybe multiply with both? Is that making sense? I am not sure so I am actually asking for suggestions. 😝 

What I want to say is that they're all cycling now, but the researchspecial types are not following the same control logic as others, so my question is that if research special evap cooler pump should also operate by air loop mass flow ratio? Or it would look like evap cooler power consumption would be not that relevant to zone load.

rraustad:
I think this is correct way:

"use air loop PLR, which means pump is cycling in the same way as fan.'

rraustad:
If the fan is not cycling then the fan runs the entire time step (even when fan PLR < 1). In this case I would think the internal PLR should be used. This would represent the evap cooler cycling to meet the supply air T set point. If the fan is cycling, then I would think the fan PLR should be used as the evap cooler would be full on when the fan was on and off when the fan was off. If the fan is cycling, and the outlet air temp is also modulated, then power should probably be calculated as Full Power * internal PLR * fan PLR.

rraustad:
So this boils down to:
Cycling fan: use Full Power * internal PLR * fan PLR
Constant fan: use Full Power * internal PLR

Have to think more about what to do when the flow fraction curve is used with respect to cycling or constant fan mode.

yzhou601:
> So this boils down to:
> Cycling fan: use Full Power * internal PLR * fan PLR
> Constant fan: use Full Power * internal PLR
> 
> Have to think more about what to do when the flow fraction curve is used with respect to cycling or constant fan mode.

That's what I am thinking too, totally agree! I think the constant fan maybe could also follow the same equation with `fan PLR = 1` all the time?

rraustad:
Yes, always use Power = Full Power * internal PLR * fan PLR; but make sure fan PLR = 1 for constant fan mode.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -2150,7 +2150,7 @@ namespace EvaporativeCoolers {
             {
                 auto const SELECT_CASE_var(EvapCond(EvapCoolNum).EvapCoolerType);
                 if (SELECT_CASE_var == iEvapCoolerInDirectRDDSpecial) {
-                    CalcIndirectResearchSpecialEvapCooler(EvapCoolNum);
+                    CalcIndirectResearchSpecialEvapCooler(EvapCoolNum, 1.0);
```
yzhou601: This didn't do anything with power reporting, just passed default value 1.0.

yzhou601:
EvaporativeCooler:Direct:ResearchSpecial results now show pump cycling with fan along with setpoint control:
image
image
image
Figure 1&2 show evap cooler follows very similar trend with fan PLR, but if the hand-calculated evap cooler PLR was divided by fan PLR to zoom up the difference (Figure 3), it came out the same shape as previous evap cooler power consumption (evap cooler internal PLR). 
I think it's now working as what I intended.

src/EnergyPlus/EvaporativeCoolers.hh
```diff
@@ -401,9 +401,9 @@ namespace EvaporativeCoolers {
                               Real64 const FlowRatio   // secondary air flow fraction
     );
 
-    void CalcIndirectResearchSpecialEvapCooler(int const EvapCoolNum);
+    void CalcIndirectResearchSpecialEvapCooler(int const EvapCoolNum, Real64 const FanPLR);
 
-    void CalcDirectResearchSpecialEvapCooler(int const EvapCoolNum);
+    void CalcDirectResearchSpecialEvapCooler(int const EvapCoolNum, Real64 const FanPLR);
```
rraustad: The unit tests are now failing since the FanPLR is a required argument. You also added a 1.0 at line 2165 above so that the code would compile. Look at [this default argument](https://github.com/NREL/EnergyPlus/pull/7230#issuecomment-517287379) declaration to see if you would rather go that route.

Real64 const FanPLR = 1.0

and then you don't need to pass FanPLR when it is 1.0.

src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -3222,16 +3222,17 @@ namespace EvaporativeCoolers {
             if (EvapCond(EvapCoolIndex).FanPowerModifierCurveIndex > 0) {
                 FanPowerModCurveValue = CurveValue(EvapCond(EvapCoolIndex).FanPowerModifierCurveIndex, FlowRatio);
             } else {
-                FanPowerModCurveValue = EvapCond(EvapCoolIndex).PartLoadFract;
+                // linearly scale fan power using part-load-fraction and fan flow ratio when fan power modifier curve is not specified
+                FanPowerModCurveValue = EvapCond(EvapCoolIndex).PartLoadFract * FlowRatio;
```
rraustad, line 3225: This should now be multiplied by FanPLR, not FlowRatio, correct?

rraustad:
The new plots look correct now. 

I'm still a little concerned about the power performance curve as a function of flow fraction and how that curve would interact with a cycling fan. It "feels" as if FanPowerModCurveValue should also be multiplied by FanPLR when the curve is used. Consider that when the developer created this model, cycling fan was not an option (or on their radar). So now that cycling fan is allowed/included, what should happen to fan power when the flow fraction curve is used? What does the flow fraction curve represent? I believe it is used to modify pump power when the air flow through the evap cooler is less than max, not quite sure why, but that's how it's used. So when using the performance curve, if air flow is less than max and the fan cycles, then it seems fan power also needs to be multiplied by FanPLR. Or maybe the performance curve should only be used when FanPLR = 1? 

This is a hard one to get right because average air flow over the time step can be less than max when the fan cycles, but when the fan is on it's at the max air flow. So this would mean:

```
FanPowerModCurveValue = CurveValue(EvapCond(EvapCoolIndex).FanPowerModifierCurveIndex, FlowRatio/FanPLR);
```

but only when the fan is cycling (and when it's constant then FanPLR = 1) and then you would modify fan power by FanPLR:

```
EvapCoolertotalPower += EvapCond(EvapCoolIndex).IndirectFanPower * FanPowerModCurveValue * FanPLR;
```

So as you can see I am not quite sure how to address this.

rraustad:
The Water Pump Power Modifier Curve models a VS pump so that when air flow is reduced, less water is needed to get to the same outlet condition and therefore pump power is reduced. So in this case you would need to know the air flow fraction when the fan is on. And then the power would need to also be multiplied by FanPLR.

src/EnergyPlus/EvaporativeCoolers.cc
yzhou601: Actually this indirect research special calculation confused me a lot, this `FlowRatio` is the secondary fan flow ratio, so I think pump might need to be consistent with secondary fan?

rraustad: If this is the indirect model then maybe the power calculation should not change until it is better understood.

src/EnergyPlus/EvaporativeCoolers.hh
yzhou601: Done, thanks!

yzhou601:
For indirect object without running advanced control, the fanPLR could be passed in the same way. Run StripMallZoneEvapCooler, the changes make SMSTORE5 and SMSTORE 8 INDIRECT EVAPORATIVE COOLER cycle (with variable volume fan):

 
image

image

rraustad:
@yzhou601 can you show the zone RH on the same plot (y2 axis) to see the impact this change has on zone control?

rraustad:
Well, that was dumb, this change only modifies power. But it's always good to double check controls.

tst/EnergyPlus/unit/EvaporativeCoolers.unit.cc
```diff
@@ -163,59 +163,6 @@ TEST_F(EnergyPlusFixture, EvapCoolers_IndEvapCoolerOutletTemp)
     EvapCond.deallocate();
 }
 
-TEST_F(EnergyPlusFixture, EvapCoolers_IndEvapCoolerPower)
```
yzhou601, line 164: This test is the same as `EvaporativeCoolers_IndEvapCoolerPower`, deleted.

yzhou601:
> Well, that was dumb, this change only modifies power. But it's always good to double check controls.

@rraustad I checked the CI for commits yesterday, only evap cooler power & energy show diffs. And I checked these diff test files, they did compare zone temperature and predicted loads, so they should be good.

rraustad:
I don't see anything wrong with this control.

image

rraustad:
@yzhou601 this looks good. Only thing left is the docs update.

 - power equations in Eng. Ref.
 - description of controls (cycling fan vs constant fan) or how a user would get constant power if they so desired (use performance curve)
 - other?

rraustad:
@Myoldmopar @mjwitte I am merging this in once CI comes back. Last chance for review.

rraustad (approved):
Code reviewed and tested. Doc updates are complete.

tst/EnergyPlus/unit/EvaporativeCoolers.unit.cc
```diff
@@ -918,4 +868,55 @@ TEST_F(EnergyPlusFixture, DirectEvapCoolerAutosizeWithoutSysSizingRunDone)
...
+    DataLoopNode::Node(EvapCond(EvapCoolNum).InletNode).Press = DataEnvironment::OutBaroPress;
+
+    DataGlobals::BeginEnvrnFlag = true;
+    
+    //Simulate air loop component calls SimEvapCooler
+    //SimEvapCooler calls InitEvapCooler(EvapCoolNum) and CalcDirectEvapCooler
+    SimAirServingZones::SimAirLoopComponent(EvapCond(EvapCoolNum).EvapCoolerName, Evap_Cooler_CompType, false, AirLoopNum, EvapCoolNum, 0);
+    
+    //air loop FanPLR successfully passed for pump power calculation
+    EXPECT_EQ(EvapCond(EvapCoolNum).EvapCoolerPower, 60 * 0.8);
```
rraustad, line 918: Evap cooler power (60 W) is now proportional to air loop FanPLR of 0.8.

rraustad:
@Myoldmopar the Win64-Windows-10-VisualStudio-16 CI had a build failure the last 2 CI passes. I'm not comfortable merging this until that build failure is understood. This CI machine built fine on another branch 10  hours ago (albeit with different diff's than the other CIs). I'm deleting those results and trying again to see what happens.

Myoldmopar:
It looks like it is working ok now.  I logged onto it and it was complaining about low disk space, but it actually had plenty.  I'm thinking that either an antivirus run or a backup interfered with something and caused a temporary issue.  It looks like it is about 20 minutes into building one of the cache branches right now, but that everything is OK.

rraustad:
Now I'm confused. The Win64-Windows-10-VisualStudio-16: Build Failed shows 1 file passed and 668 failed. When you look at the results it's 9 diff's and 2507 passed. So this seems like a CI issue, not related to these changes?

rraustad:
I'm inclined to merge this. Any thoughts as to why not?

Myoldmopar:
@rraustad the confusion here is that I deleted the Windows results with the build failure.  The CI then re-ran and posted new results (the final CI commit above).  So the old results just need to be ignored.  I'm striking them right now.

rraustad:
Now the 3 diff's files are all evap coolers and differences have been explained. This can be merged once CI completes.

rraustad:
CI looks to have 1 more test. Delete branch after CI completes.

Issue #5000, opened 2015-06-29
Verify and possibly update definition of leeward for exterior convection coefficients


mjwitte:
As an outcome of current ASHRAE Std 140 development work, it has been noted that EnergyPlus uses a test of angle >100degrees to determine if a surface is windward or leeward.   The same test is used for all exterior convection algorithms that are sensitive to wind direction.  However, Yazdanian and Klems (1994, p.4) indicate their data tabulations correspond to “windward and leeward hemispheres”, interpretable as leeward being definable as > 90 degrees from normal wind incidence.  If this interpretation is correct, then should EnergyPlus be changed to use >90 degrees, and should this apply to all convection models or just for DOE-2 and MoWiTT.

Myoldmopar:
Magic issue #5000!

RKStrand:
@mjwitte @Myoldmopar Ok, so the documentation agrees with the code as far as the >100 degrees test, but I have to say that I find no logical reason why 100 is used as the cutoff.  >90 seems more rational as per Yazdanian and Klems.  While someone might argue that we should handle anything related to Walton with the >100 cutoff and DOE-2 and MoWiTT with a cutoff of >90, the reality is that only the Sparrow, DOE-2, and MoWiTT models use this check as far as I can tell.  Why use different values for the windward cutoff?  Seems like >90 has clear logical reasoning while >100 seems strange.  Planning on changing this to use >90 for all models and let the convection fall where it may.

mjwitte:
Closed via #7458

RKStrand:
Thanks, Mike!

On Aug 25, 2019, at 7:19 AM, Michael J. Witte <notifications@github.com<mailto:notifications@github.com>> wrote:


Closed via #7458<https://github.com/NREL/EnergyPlus/pull/7458>

PR #7458, opened 2019-08-16
Corrected Cutoff between Windward and Leeward for Exterior Convection Models

RKStrand:
Pull request overview
---------------------
Legacy code used a 100 degree difference between the wind direction and the surface outward normal as the cutoff between windward and leeward.  More recent publications and common sense say that the cutoff should be 90 degrees.  This work changed the cutoff to 90 degrees and updated the documentation to reflect this.  This resolves GitHub Issue #5000 

### Work Checklist
   - Defect: This pull request repairs a github defect issue #5000 

### Review Checklist
 - [x] Unit Test(s)
 - IDD changes n/a


RKStrand:
In looking at the differences caused by this change, the files that show differences are generally fairly small and all of them use the DOE-2 exterior convection model which does in fact use the check that was modified in this work.  The file that shows the biggest difference (other than differences in J) was in the temperatures of surfaces in StackedZonesWithInterzoneIRTLayers.idf, particularly on the south side.  When looking at the file, the winter design day has a wind direction that will make it be windward in develop and leeward after the mod.  So, I tried to run this with the modified code and changing the wind direction for the winter design day so that south will again be windward.  When I make that change, the temperatures line back up again exactly.  Note also that this example file is a bit unusual in that it uses the infrared transparent surface to artificially separate a single tall zone into three zones.  So, I felt like the differences in output were fairly reasonable given the change.

mjwitte (approved):
Straightforward change. Nice tight unit test (which I made a little tighter).  Merged in develop for one more round of testing.

PR #7459, opened 2019-08-16
Add "with escalation" tables to life cycle cost report

### LINKED ISSUE #6144: LifeCycleCost:UsePriceEscalation should impact the Total Cost column of the Present Value by Year table
The LifeCycleCost:UsePriceEscalation should impact the Total Cost column of the Present Value by Year table. The LifeCycleCost:UsePriceEscalation does seem to impacting numbers but its impact is not directly shown in any of the existing tables. 

 - Version of EnergyPlus 8.6.0

### Checklist
 - [x] Defect file added (in DevSupport repo)

JasonGlazer:
Add escalated energy cost tables and a column to the present value by year table that shows the escalated energy costs. This resolves #6144

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


JasonGlazer:
@mjwitte or @Myoldmopar would one of you like to review this or suggest someone that would be a good reviewer?

Myoldmopar:
I'm happy to review this week.

JasonGlazer:
Thanks

src/EnergyPlus/EconomicLifeCycleCost.cc
```diff
@@ -1496,7 +1496,13 @@ namespace EconomicLifeCycleCost {
 
          for (int iCashFlow = 1; iCashFlow <= numCashFlow; ++iCashFlow) {
             if (CashFlow(iCashFlow).pvKind == pvkEnergy) {
-                curResource = CashFlow(iCashFlow).Resource - ResourceTypeInitialOffset;
+                // make sure this is not water
+                int curResource_iRT = CashFlow(iCashFlow).Resource;
+                if (CashFlow(iCashFlow).Resource == iRT_Water ||
+                    (CashFlow(iCashFlow).Resource >= iRT_OnSiteWater && CashFlow(iCashFlow).Resource <= iRT_Condensate)) {
+                    continue;
```
rraustad, line 1503: Just a thought. It would be a little faster if these iRT variables were reordered, with water types at the end, so that only
```
if (CashFlow(iCashFlow).Resource >= iRT_Water) continue;
```

could be used. But probably wouldn't save that much since this code is only executed at end? of simulation.

src/EnergyPlus/EconomicLifeCycleCost.cc
```diff
@@ -1477,6 +1488,47 @@ namespace EconomicLifeCycleCost {
...
+                }
+                curResource = curResource_iRT - ResourceTypeInitialOffset;
+                if ((curResource >= 1) && (curResource < NumOfResourceTypes)) {
+                    int found = 0;
+                    for (nUsePriceEsc = 1; nUsePriceEsc <= numUsePriceEscalation; ++nUsePriceEsc) {
+                        if (UsePriceEscalation(nUsePriceEsc).resource - ResourceTypeInitialOffset == curResource) {
+                            found = nUsePriceEsc;
```
rraustad, line 1510: Once found, break out of for loop?

tst/EnergyPlus/unit/EconomicLifeCycleCost.unit.cc
```diff
@@ -415,3 +421,172 @@ TEST_F(EnergyPlusFixture, EconomicLifeCycleCost_ProcessMaxInput)
...
+    EscalatedEnergy.allocate(lengthStudyYears, 1);
+    EscalatedTotEnergy.allocate(lengthStudyYears);
+    EscalatedTotEnergy = 0.0;
+
+    ComputeEscalatedEnergyCosts();
+    EXPECT_NEAR(EscalatedEnergy(1, 1), 100., 0.001);
+    EXPECT_NEAR(EscalatedEnergy(2, 1), 100., 0.001);
+    EXPECT_NEAR(EscalatedEnergy(3, 1), 100., 0.001);
+    EXPECT_NEAR(EscalatedEnergy(4, 1), 100., 0.001);
+    EXPECT_NEAR(EscalatedEnergy(5, 1), 100., 0.001);
```
rraustad: Probably not a good choice for a unit test (using all 100s) where you don't know if the correct CashFlow(1).yrAmount(x) was used. Would rather see 100, 200, 300... or 101, 102, 103... Might not be worth the effort.

rraustad:
I have reviewed with only minor comments. Hard to understand mods without reviewing results and comparing with hand calcs.

JasonGlazer:
Thanks for the review @rraustad

src/EnergyPlus/EconomicLifeCycleCost.cc
JasonGlazer: It would make sense to optimize this if it was in a timestep but since this is not it probably is not worth the effort.

src/EnergyPlus/EconomicLifeCycleCost.cc
JasonGlazer: Thanks

tst/EnergyPlus/unit/EconomicLifeCycleCost.unit.cc
JasonGlazer: Good idea.

rraustad (approved):
Comments addressed, Things look good, requesting clarification on results before merging.

rraustad:
I don't see any diff's related to these changes.

rraustad:
I'm not seeing any reason for further review. Merging.

Myoldmopar:
Have at it, @rraustad

rraustad:
Ran 5ZoneEconomicsTariffAndLifeCycleCosts.idf in this branch:

image


and in develop:

image

rraustad:
New Present Value By Year table now shows escalated Total Costs and those costs vary by year. @JasonGlazer since I don't have time to review how the escalated Total Costs should change over time as compared to the non-escalated costs, I'll ask a few questions before merging.

 - Is the Total Cost columns the cost for that year alone? I mean, not a cumulative cost.
 - Shouldn't the Total escalated cost for Jan 2013 be different from $51,500 ? Why was this number not affected by escalation?
 - and in Jan 2014, a difference of ~$2,000 (non escalated vs escalated) seems rather large for 3 years of escalation. Any explanation for that?
 - and if the year values are present value in today's dollars, then should the values increase year over year, not decrease? OK, well the numbers do increase year over year, I meant the $10,113.35 number for Jan 2014, shouldn't it be higher than$11,941.56.

JasonGlazer:
Let me take a look. That does seem like a large difference.

rraustad:
In the example file I see escalation rates of around 1%, that's $100 per year (based on the 11,000 number), not $1000 per year.

mjwitte:
@Myoldmopar Since this adds a column to an output table, is it still ok to merge for 9.2?

Myoldmopar:
That's structural, so technically no, this should wait until after release.  I'll add the proper milestone and you can review and approve the changes still.  Then as soon as the release is done this can drop in.

JasonGlazer:
Darn. I did this one first in hopes that it would be in before the I/O freeze.

mjwitte:
The OutputChange label would have helped prioritize it - sorry.

rraustad:
Technically no? As if a new column in an LCC table, would cause havoc. I'd rather see correct data in the tables. Oh well, it'll drop in 2 seconds after release as long as reviewer questions are addressed.

JasonGlazer:
The BND diffs seem completely unrelated. I'm going to merge develop again and hopefully they go away.

JasonGlazer:
@rraustad and @Myoldmopar I think this ready to be merged as soon as the release goes out. CI had some spurious warnings but that was just because it wasn't caught up to develop.

PR #7463, opened 2019-08-20
Remove extraneous epJSON fields when outputting converted file

EricDGarrison:
Added function to clean the output epJSON file to no longer have "idf_order", "idf_max_fields", and "idf_max_extensive_fields" and modified existing code to use this function to modify the ouput epJSON file.

The current epJSON structure with the cleaned fields ("idf_order", "idf_max_fields", and "idf_max_extensive_fields") is still available to be used at run time, but is no longer saved to an epJSON file.


src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -238,6 +238,21 @@ void InputProcessor::markObjectAsUsed(const std::string &objectType, const std::
     }
 }
 
+void cleanEPJSON(json &epjson)
+{
+    if (epjson.type() == json::value_t::object) {
+
+        epjson.erase("idf_order");
+        epjson.erase("idf_max_fields");
+        epjson.erase("idf_max_extensible_fields");
```
Myoldmopar, line 246: This seems fine, I was worried about what happened if that key wasn't in the object, but I see from the docs that it will pass normally.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -280,12 +295,17 @@ void InputProcessor::processInput()
...
             //			if ( !success || hasErrors ) {
             //				ShowFatalError( "Errors occurred on processing input file. Preceding condition(s) cause termination." );
             //			}
+
             if (DataGlobals::outputEpJSONConversion) {
-                input_file = epJSON.dump(4, ' ', false, json::error_handler_t::replace);
+                json epJSONClean = epJSON;
+                cleanEPJSON(epJSONClean);
+                input_file = epJSONClean.dump(4, ' ', false, json::error_handler_t::replace);
+                //input_file = epJSON.dump(4, ' ', false, json::error_handler_t::replace);
```
Myoldmopar, line 305: No unit test for this new function?  Seems like it would be worth passing in a full epJSON instance and verifying it handles all of it correctly.

Myoldmopar (commented):
I really like these changes, just need a unit test and it can drop in.

Myoldmopar (approved):

Myoldmopar:
I just tested this locally, and it verified the changes, and it sure worked perfectly.  CI seems to really hate it though...I'll have to understand why that is happening.

Myoldmopar:
Or...if I just take a quick look at the CI failure I see the unit tests won't build because of an obvious compilation failure.

- Occasionally I understand pushing up code that is broken as just an incremental step during development, but this was you writing a unit test, apparently not building it locally, and pushing it up.  That adds to our review time.
- Please use a branch name other than `develop` to push changes for bug fixes or new features, even within your own fork, if you intend on submitting a pull request back to the upstream repo.

Myoldmopar:
I fixed the build issue and I'm not going to wait for CI to pick this up to re-run, it was obviously the problem.  I'm merging this now.

PR #7473, opened 2019-08-26
Fix #7176: Non-zero zone origins warning is wrong with simple rectangular surface

### LINKED ISSUE #7176: Non-zero zone origins warning is wrong with simple rectangular surfaces
Issue overview
--------------
Got unexpected eio diffs when cleaning up zone origin warnings in HybridZoneModel.idf for #7094. The GlobalGeometryRules say World coordinates, but the zone origins were non-zero which generated a warning. Setting the zone origins to zero shouldn't have changed anything, but there were diffs in the zone centroid and max/min x/y in the eio output. Reproduced this using the 4ZoneWithShading_Simple.idf testfile.

And it caused eso diffs, which are likely due to the z-centroid being incorrect for zones 3 and 4. This needs more investigation and is likely to cause many diffs for any file with non-zero zone origins with world coordinates!
And also shd diffs?

Possibly related to #5849?

**CORRECTION**
Turns out this file uses the simple rectangular surfaces which have their own coordinate system input in GlobalGeometryRules which defaults to Relative. So, the eio info is not wrong, but this warning is incorrect for this case:
```   
** Warning ** GetSurfaceData: World Coordinate System selected.  Any non-zero Building/Zone North Axes or non-zero Zone Origins are ignored.
**   ~~~   ** These may be used in daylighting reference point coordinate calculations but not in normal geometry inputs.
```

So, the warning should be gone after this is fixed for 4ZoneWithShading_Simple_1.idf as-is.  But the last two fields should be added to this so that it shows the real picture:
```
  GlobalGeometryRules,
    UpperLeftCorner,         !- Starting Vertex Position
    CounterClockWise,        !- Vertex Entry Direction
    World,                   !- Coordinate System
    Relative,                !- Daylighting Reference Point Coordinate System
    Relative;                !- Rectangular Surface Coordinate System
```

### Details
 - Version of EnergyPlus 9.1.0-https://github.com/NREL/EnergyPlus/pull/7094/commits/1d97699f9548727042a451c0761eeb9008c8bfdb

### Checklist
 - [x] Defect file attached:
[4ZoneWithShading_Simple_1.idf.txt](https://github.com/NREL/EnergyPlus/files/2877735/4ZoneWithShading_Simple_1.idf.txt)

dareumnam:
Pull request overview
---------------------

Issue #7176 

- SurfaceGeometry.cc
   - Updated the comment for **bool RectSurfRefWorldCoordSystem**:
      The previous comment was appropriate to DataSurfaces::WorldCoordSystem, not to RectSurfRefWorldCoordSystem
   - Fixed the code:
      Warning messages were added in case of 
      - Coordinate system = **World** &&
      - Rectangular Surface Coordinate System = **Relative (or default)** &&
      - **Non-zero** zone origins
 
- SurfaceGeometry.unit.cc
   - Added a unit test

- Inputs of 12 regression test files were fixed so that the error messages aren't shown
   - "World" coordinate was selected for "Rectangular Surface Coordinate System" field
   - "4ZoneWithShading_Simple_1" was fixed: fixed each zone's starting X,Y,Z with world coordinate system so that it has the same numeric result with original IDF file (relative coordinate with each zone origin)

### Notes to Reviewer
- Expected diffs: Diffs will be expected in the error outputs including .eio for 12 regression tests
- CI results: Other than expected diffs, CI should be green

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


Myoldmopar:
@dareumnam thanks for this.  Could you please add a section to the pull request text that is something like:

--------------------------------
### Notes to Reviewer
* Expected diffs: Diffs will be expected in the error output of 2 files
* CI results: Other than expected diffs, CI should be green
--------------------------------

Something like that anyway...it would make it very easy to review and merge once CI is done with it!

dareumnam:
@Myoldmopar Thanks for letting me know. Just added the section to the PR text.

Myoldmopar:
@dareumnam there are still a couple unit tests failing.  It looks like they are expecting error output to be blank but there is error content.  I'd debug in and see what the error stream looks like at that point in the unit test to see what's going on.

Myoldmopar:
@dareumnam this is much better now, no unit test failures and only 12 files with error diffs.  I think this is really close.  We do need to address those diffs one way or another though.

For one thing, the error message has an odd blank string in it, presumably because the input field is defaulted to blank or something?

image

Second, we should just try to fix up our input files so that the error message isn't thrown.  Then you won't get any diffs and this can drop in.

dareumnam:
@myoldmopar Okay. I’ll try to fix up input files. And yes, the odd blank string in the error message is because the input field is the default. Do you think it is better to separate the cases that using “default” / using “relative” for Rectangular surface coordinate system, and then make different warning messages for each case?

Myoldmopar:
The ZoneUnitarySysTest failure is unrelated to your branch, it just sometimes fails due to something not being cleaned up during unit testing.  

The error diffs are still showing up, let me know if you need any tips to wrap this up.

Myoldmopar:
@dareumnam so this is looking really close.  However, the IDF changes you made used tabs as the indentation, and it should be plain spaces.  So we'll need to clean that up.  

As for CI, this is kinda what I was expecting, with most of the IDFs reverting back to no diffs once you set them to World, but one showing diffs.  The EIO diffs are showing lots of different surface information and small changes to warmup convergence.  There are large ESO and MTR diffs, and they do indeed look large.  We need to decide how to proceed forward at this point.  

One option is to simply accept the changes as they are and that will be the new results for that file with diffs.  If we choose to do this, we need to spend some more time understanding the diffs before accepting them.  This could prove a bit difficult since we appear to be moving surfaces around.  So another option would be to actually manually adapt the origins, vertices or coordinate system inputs so that we use the World coordinates, with rectangular surfaces, and yet we still achieve the same numeric results from the simulation.  It may be as simple as changing the origin coordinates or something, but this would be a great solution that would eliminate the significant diffs and yet provide a full fix and clean error files.

dareumnam:
@Myoldmopar Thanks so much for the review. I'll fix the IDFs with spaces instead of using tabs. I'll also try to do the second option.

Myoldmopar:
👍

dareumnam (commented):
Zone origins were set to the same with the original idf file. (instead of 0,0,0)

Myoldmopar:
This is great now, thanks @dareumnam.  It's great to have the input file fixed up and no new warnings or diffs.  This is good to go in.

Issue #4420, opened 2014-08-20
Surface Window Front and Back Face Temperature report zeroes for plain windows - output not set up correctly


BobHenninger:
Helpdesk Ticket 9049

Zone,Average,Surface Window Front Face Temperature Layer 1 [C]
 Zone,Average,Surface Window Back Face Temperature Layer 1 [C]

 These should only get setup for windows which are Construction:ComplexFenestrationState, unless we can figure out how to report these for all window types. It appears there are values of thetas that are the layer-by-layer temps in the standard window model. See also Ticket 9013

 Michael J. Witte on 28 January 2014 10:06 PM (Edited by Michael J. Witte on 28 January 2014 10:09 PM)

Posted on: 28 January 2014 10:04 PM

I have confirmed that these output variables for front and back face layer temperatures are only active when using complex fenestration (Construction:ComplexFenestrationState). They should not even appear in the output for other window types. I will submit a request to get this fixed. For standard windows, you can use these output:variables to report the innermost and outermost glass face temperature.

 Surface Inside Face Temperature [C]
 Surface Outside Face Temperature [C]

 Mike

> > I try to run EP in the basic case ( A room with 1 window+Venetian
> > blind on the west wall).Some output variables can be figured out
> > but
> > "Surface Window Front and back Face Temperature" show "0 value" in
> > anytime. I don't know how to show the true values of them.
> 
> Which type of window model are you using? I am not sure these
> variables are active for all of the window models. It would be helpful
> to see your idf file to confirm.
> 
> Mike

Naparat
User
Posted on: 26 January 2014 08:03 AM

I try to run EP in the basic case ( A room with 1 window+Venetian blind on the west wall).Some output variables can be figured out but "Surface Window Front and back Face Temperature" show "0 value" in anytime. I don't know how to show the true values of them.

User files
01clear6mm_1.idf

mjwitte:
Came up again in helpdesk ticket 11027.  Adding priority to get this fixed.

RKStrand:
Just verified that this is still a bug.  Will be working to fix it for the upcoming release.

Myoldmopar:
Closed via #7474

PR #7474, opened 2019-08-26
Correction of Window Face Temperature Reporting for Non-Complex Windows

RKStrand:
Pull request overview
---------------------
A user noted that the window face temperatures were reporting as zero.  Further investigation showed that these variables were only being produced for complex (BSDF) windows.  This pull request includes a fix that allows the outer most surface outward facing temperature and the inner most surface inward facing temperature to be produced.  For other layers of non-BSDF windows, the output will not be produced at all.  In addition, the variable Surface Window Total Absorbed Shortwave Radiation Rate Layer will only be produced for BSDF windows, not for other types.  This addresses GitHub Issue #4420.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #4420 

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
This all looks good. My only question would be why the inner/outer surface temps could not also be reported for non-BSDF windows? To be clear, this change addresses the issue, but for the future shouldn't all window surface temps be available?
The 2 example files that show err diff's are for the "unused" report variable that is no longer available for non-BSDF windows.

mjwitte:
Added OutputChange label - at first glance this may have to wait until after release due to changes in output variables? Not sure though.

RKStrand:
@rraustad But this change specifically did allow the reporting of non-BSDF windows--at the outer most outward facing layer and the inner most inward facing layer.  The unit test documents the setting of those variables and the documentation also shows that.  So, you get all layers in both directions for BSDF windows because it is calculating things layer by layer.  For non-BSDF windows, you just get what gets calculated at the inside and outside surface heat balance.  Those variables were already available like they would be for any other surface so this can be verified pretty easily.  So, I think this work addresses your concern already...or is something missing???

rraustad:
Maybe I was just confused by the description. Reading again I guess it says outer/inner layers of non-BSDF windows will be reported but others will not. 

"For other layers of non-BSDF windows, the output will not be produced at all."

mjwitte:
Removed outputchange label - this is just fixing existing variables.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -2632,15 +2632,16 @@ \subsection{Window Output Variables}\label{window-output-variables}
 
 \subsubsection{Surface Window Total Absorbed Shortwave Radiation Rate Layer \textless{}x\textgreater{} {[}W{]}}\label{surface-window-total-absorbed-shortwave-radiation-rate-layer-x-w}
 
-This will output shortwave radiation absorbed in a window layer. The key values for this output variable are the surface name. Layers are numbered from the outside to the inside of the surface. The full listing will appear in the RDD file.
+This will output shortwave radiation absorbed in a window layer. The key values for this output variable are the surface name. Layers are numbered from the outside to the inside of the surface. The full listing will appear in the RDD file. Note that this variable is only defined for constructions defined by a \hyperref[constructioncomplexfenestrationstate]{Construction:ComplexFenestrationState}.
```
Myoldmopar, line 2635: 👍

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5608,6 +5608,10 @@ namespace HeatBalanceManager {
         for (SurfNum = 1; SurfNum <= TotSurfaces; ++SurfNum) {
             DataSurfaces::Surface(SurfNum).MovInsulIntPresentPrevTS = DataSurfaces::Surface(SurfNum).MovInsulIntPresent;
         }
+        
+        // For non-complex windows, update a report variable so this shows up in the output as something other than zero
+        UpdateWindowFaceTempsNonBSDFWin();
```
Myoldmopar, line 5613: No problem here.

src/EnergyPlus/SolarShading.cc
```diff
@@ -1316,24 +1316,30 @@ namespace SolarShading {
...
-                                            "Average",
-                                            Surface(SurfLoop).Name);
+                        if (Construct(Surface(SurfLoop).Construction).WindowTypeBSDF) {
+                            SetupOutputVariable("Surface Window Total Absorbed Shortwave Radiation Rate Layer " + RoundSigDigits(I) + "",
+                                                OutputProcessor::Unit::W,
+                                                QRadSWwinAbsLayer(I, SurfLoop),
+                                                "Zone",
+                                                "Average",
+                                                Surface(SurfLoop).Name);
+                        }
```
Myoldmopar, line 1326: OK, shortwave absorbed is only defined for BSDF.

src/EnergyPlus/SolarShading.cc
```diff
@@ -1316,24 +1316,30 @@ namespace SolarShading {
...
+                        if (Construct(Surface(SurfLoop).Construction).WindowTypeBSDF || (I == NumOfLayers)) {
+                            SetupOutputVariable("Surface Window Back Face Temperature Layer " + RoundSigDigits(I) + "",
+                                                OutputProcessor::Unit::C,
+                                                FenLaySurfTempBack(I, SurfLoop),
+                                                "Zone",
+                                                "Average",
+                                                Surface(SurfLoop).Name);
+                        }
```
Myoldmopar, line 1342: But the front and back layers are available for all types, and for every layer in BSDF.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
```diff
@@ -1878,4 +1878,59 @@ TEST_F(EnergyPlusFixture, HeatBalanceManager_GetMaterialData_IRTSurfaces)
     
 }
 
+TEST_F(EnergyPlusFixture, HeatBalanceManager_UpdateWindowFaceTempsNonBSDFWin)
+{
+
+    DataHeatBalance::clear_state();
+    DataSurfaces::clear_state();
+    DataHeatBalSurface::clear_state();
```
Myoldmopar, line 1886: Definitely should _not_ be necessary.  These should've been cleaned up by the cleanup from the unit test that ran before this.  It's OK that they are here right now, but just note this shouldn't be needed.  If it is then something else is wrong.

tst/EnergyPlus/unit/HeatBalanceManager.unit.cc
```diff
@@ -1878,4 +1878,59 @@ TEST_F(EnergyPlusFixture, HeatBalanceManager_GetMaterialData_IRTSurfaces)
...
+
+    HeatBalanceManager::UpdateWindowFaceTempsNonBSDFWin();
+    
+    // First surface is NOT a window so these should NOT be set
+    EXPECT_NEAR(FenLaySurfTempFront(1,1),ZeroResult,0.0001);
+    EXPECT_NEAR(FenLaySurfTempBack(1,1),ZeroResult,0.0001);
+    
+    // Second surface is a window so these should be set
+    EXPECT_NEAR(FenLaySurfTempFront(1,2),DataHeatBalSurface::TH(1,1,2),0.0001);
+    EXPECT_NEAR(FenLaySurfTempBack(SurfsForRegWindow,2),DataHeatBalSurface::TH(2,1,2),0.0001);
```
Myoldmopar, line 1928: OK

Myoldmopar (approved):

Myoldmopar:
@RKStrand this needs the example files cleaned up to get rid of the [error diffs](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2019-08/1992e565ecd81ab3945a7d9f390397485970dc01-0c1a3e8bdf6f0a4ddfb54374e1989629d8bf4dc5/WCE_Diffuse_Shade/x86_64-MacOS-10.13-clang/eplusout.err.diff.html) due to the warning message about unused output variable.  While making that change, might as well go ahead and fix up the unnecessary clear_state calls at the beginning of the new unit test.  I'm going to build this locally and verify anyway, so I'll likely just make those changes and push the fixes up.

RKStrand:
@Myoldmopar Thanks for sorting that out (if you end up doing it).  My apologies for some sloppy clear_state calls in the unit test.  I was operating under the mode that each unit test has to perform on its own and not rely on anything from any previous unit test.  My thought was that if the previous unit test changes or gets deleted (would that ever happen?) that would then potentially cause a problem with this one.  I understand that could be somewhat wasteful though with lots and lots of calls to clear_state routines that aren't really needed.

Myoldmopar:
You are generally right, a unit test should operate on their own.  However, clear_state calls are handled at the end of every single unit test automatically by the EnergyPlusFixture base class.  It is expected that through these clear_state calls, the entire global state of EnergyPlus will be reset fully to its raw state again.  If a clear_state needs to be called within a unit test, it indicates that something wasn't properly added to the fixture base class.

I pulled this branch and tested locally.  With develop pulled in, and the unneeded clear_state calls in place, the unit test suite runs successfully.  I also re-ran the example files after cleaning out the output variable issue and the error files are fully clean again.  Merging this.

PR #7479, opened 2019-08-29
Simplify the moist air specific heat capacity psychrometric function

### LINKED ISSUE #7476: Simplify the moist air specific heat capacity psychometric function
## Issue overview

The equation used to calculate specific heat of moist air can be reduced to simple form and avoid the call for enthalpy of moist air. Currently energyplus uses function PsyCpAirFnWTdb(), which is based on the definition of specific heat = dh/dT, is given by:

            CpAir = ((PsyHFnTdbW(T + 0.1, w) - PsyHFnTdbW(T, w)) * 10.0);

The above expression requires to call enthalpy function twice but we can substitute the expression for enthalpy of moist air in the equation above and reduce the moist air specific heat function to simple form (or derive definition of specific heat = dh/dT) as shown below:

           CpAir = cpa {dryAir} + w * cpv {water vapor) =  1.00484e3 + w * 1.85895e3 

In fact the specific heat moist air should be a function of HUMIDITY RATIO only and we should not pass temperature to this function. This simplification may reduce specific heat computation burden. There are more than 1500 places that call the function PsyCpAirFnWTdb().

### Details
The simplified specific heat expression for moist air has been verified manually.

### Platform (Operating system, version)
N/A

### Checklist
Add to this list or remove from it as applicable. This is a simple templated set of guidelines.

Nigusse:
Pull request overview
---------------------
Addresses issue #7476. This fix replaces moist air specific heat calculating numerical function by analytical function. The temperature argument is redundant if the analytical function is used per the definition of moist air enthalpy currently used. This is not a defect but anticipates reducing the computation burden of moist air specific heat calculation. The diffs observed are may be rounding numerical error.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Myoldmopar:
@Nigusse I like the simplification here, but as a starter, you'll need to address the failing unit tests.  It will be interesting to see what kind of changes you need to make to get them passing again.  Hopefully it is just very small rounding differences.

src/EnergyPlus/Psychrometrics.hh
```diff
@@ -334,8 +334,8 @@ namespace Psychrometrics {
         return 1.00484e3 * TDB + dW * (2.50094e6 + 1.85895e3 * TDB); // enthalpy {J/kg}
     }
 
-    inline Real64 PsyCpAirFnWTdb(Real64 const dw, // humidity ratio {kgWater/kgDryAir}
-                                 Real64 const T   // input temperature {Celsius}
+    inline Real64 PsyCpAirFnWTdb(Real64 const dw,          // humidity ratio {kgWater/kgDryAir}
+                                 Real64 const EP_UNUSED(T) // input temperature {Celsius}
```
Myoldmopar: My preference would be to go ahead and just remove this argument entirely if it isn't used anymore.

src/EnergyPlus/Psychrometrics.hh
```diff
@@ -356,45 +356,41 @@ namespace Psychrometrics {
...
         static Real64 cpaSave(-100.0);
 
         // check if last call had the same input and if it did just use the saved output
-        if ((Tsave == T) && (dwSave == dw)) return cpaSave;
+        if (dwSave == dw) return cpaSave;
 
         // compute heat capacity of air
         Real64 const w(max(dw, 1.0e-5));
-        Real64 const cpa((PsyHFnTdbW(T + 0.1, w) - PsyHFnTdbW(T, w)) * 10.0); // result => heat capacity of air {J/kg-C}
+        Real64 const cpa((1.00484e3 + w * 1.85895e3)); // result => heat capacity of moist air {J/kg-C}
```
Myoldmopar, line 473: Is there a reference for this change?  The PR description is empty, and with lots of diffs to work through, it will be important to document where this change came from.

nrel-bot-3:
@Nigusse @lgentile it has been 30 days since this pull request was last updated.

Nigusse:
Updated branch and resolved merge conflict.

src/EnergyPlus/Psychrometrics.hh
Nigusse: Removed the unused argument (TEMPERATURE) from PsyCpAirFnWTdb() function. Run unit tests locally. All happy.

Nigusse:
The diffs / failed test observed between develop and this branch is caused by difference in moist air cp value calculation method. Develop uses numerical function (delta_h/delta_t = PsyCpAirFnWTdb(W, T)) to determine Cp value while this branch uses analytical function (PsyCpAirFnW(W)). The analytical function requires humidity ratio only as an input while the numerical function requires two parameters Humidity Ratio and Temperature. Furthermore, the numerical function computes  enthalpy first to determine the Cp value while the analytical function determines the cp value directly.

Nigusse:
The difference in cp values observed in wide range of test conditions is in the order of 1.0e-10 kgWater/kgDryAir. Error stats determined between the analytical and numerical function over a wide range of humidity ratio and temperature range that covers the entire psychometric chart grid is summarized below:
    EXPECT_DOUBLE_EQ(Error_min, -2.8808244678657502e-10);
    EXPECT_DOUBLE_EQ(Error_max, 2.5875124265439808e-10);
    EXPECT_DOUBLE_EQ(Error_avg, 1.5508032789728189e-09);
    EXPECT_DOUBLE_EQ(StdError, 6.7111413639467468e-10);

Positive MBE (Error_avg) indicates the numerical method slightly over predicts the cp values. However, I am puzzled why the simplified analytical solution used in this branch does not improve computational performance. Some of the regression test suite files show increased warmup errors. This could be one of the explanation for increased computational time in the regression results and the large diffs reported for some of test files. Needs further investigation to find out what causes the increased warmup errors for a few of the regression test files.

Nigusse:
Investigated one of the test files failed (5ZoneCoolingPanelBaseboard.idf). This test file has a humidity   control (SetpointManager:SingleZone:Humidity:Maximum) based on thermal zone, SPACE2-1. The Zone Air Temperatures and Zone Predicted Loads align very well between develop and this branch even though diffs were reported for other variables. The large diffs were reported for the main heating and cooling coils due to shift in coils operation timestep around 3 am on July 21 (07/21) run period caused by changing predicted moisture loads. Note that there is no predicted sensible loads around this time.

ZoneAirTemp_PredictedLoad

See below the main air loop heating and coiling water coils heat transfer rates. Note that around 3 am 07/21 shows large diffs in the heating and cooling rates between develop and this branch (#7476). It is clearly evident that the diffs oscillate between negative and positive values. This oscillation can be  directly explained by changing predicted moisture loads in the csv file snapshots around 3 am.

MainHeating_CoolingCoils_HeatRate
 

See changing predicted moisture loads around 3 am 07/21.

ZonesPredicedLoads


Also see how the coils operation shifted between develop and this branch as reported in the CSV output file, snapshot round 3 am 07/21.

CoilsHeatRateCSVOutput

Nigusse:
Annual energy use comparison for failed test file 5ZoneCoolingPanelBaseboard.idf does not show diffs.

AnnualEnergyUse

Nigusse:
@Myoldmopar This branch is ready for review.

nrel-bot-3:

nrel-bot-2b:

mitchute:
@Nigusse It looked like most of the conflicts here were due to my recent refactoring tasks, so I went ahead and pulled in develop and resolved the conflicts. Let's see what the results look like after CI has had a chance to run.

Myoldmopar:
@Nigusse I agree that the changes are looking better.  There are 17 files with big ESO diffs and 8 with big MTR diffs.  The files with big MTR diffs are the ones I am most concerned with.  As long as they are actually small then the oscillatory time series diffs are acceptable.  The branch is again conflicted so I'll pull develop in here soon.

Nigusse:
@Myoldmopar @mitchute would you please resolve this conflicts so that I can look at the big MTR diffs?

mitchute:
@Nigusse ready for you again.

Nigusse:
@mitchute Thanks.

Nigusse:
I run one of the test suite file (5ZoneCoolingPanelBaseboard.idf) with big MTR diffs. The annual run shows that the Energy End Uses in html outputs files of this branch and develop are identical. Similary the meter variables sum (at detailed timestep reporting frequency) are almost the same as shown below. The absolute difference for Electricity:Facility [J] is about 119.5J and the percent diffs is tiny (in the order of 1.0E-8). 

AnnualEnergyUseNew

Note that all the eight test files with big MTR diffs has 2 days run period (one in winter and another in Summer). Next I will compare meter outputs at detailed time step for the two days run period only.

Nigusse:
I traced back the big MTR energy use difference in the cooling and heating energy end uses, i.e, cooling energy use of the chiller and heating energy use of the gas heating coils. This problem happens during summer design day run only (07/21, 2am - 5am) while the other three days run show very good match. The chiller energy use difference is caused by time shift in the main water cooling coil operation and the heating energy end use difference is caused by the shift in the gas heating coil operation.  This time shift is documented in my earlier comments and is caused by predicted moisture load shifts in magnitude and sign for late night hours of summer sizing period run. The cooling and heating coils operation is triggered by a humidistat. Seven of the eight files with big MTR diffs have humidistat.

ElectricFacility

Nigusse:
See below the chiller cooling rate and heating coil operation shift during early morning hour of summer sizing period run. 

CoolingHeatingCoils

Nigusse:
Figures below show the predicted moisture load around the time shift and the predicted sensible load for summer sizing period run. It is evident that there is no sensible heating and cooling load around the time shift. The predicted moisture load shift and sometimes the change from de-humidification to humidification is the one that triggers the main cooling and heating coils operation to shift and hence the MTR energy use diffs. Note that I did not not see the time shift anomaly or the MTR energy use big diffs during annual run.

PredictedLoad

Nigusse:
Review summary of the BIG MTR diffs:
- the big MTR diffs is caused by the time shift in predicted moisture load. The shift in predicted moisture load and hence the diffs in MTR chiller energy use and heating coil gas use were observed only during 2am to 5am hours of summer sizing period run.
- this BIG MTR diffs wash out during annual run. Also, MTR diffs are tiny if the test file is run for the two days run period only. The trigger for the BIG MTR diffs is the summer sizing period run.

Nigusse:
I stumbled into @mjwitte comment regarding increasing plant minimum iteration limit (changing  from default value of 2 to 10) impact described in issue #7653. Out of curiosity I run one of the test files (5ZoneCoolingPanelBaseboard.idf) with big MTR diffs by increasing the plant minimum iteration limit to 10 from default value of 2. Now this branch and develop show almost now diffs. See in figures below, diffs for facility electricity and heating gas energy uses vanished. See also the facility electricity energy end uses in my earlier plot with default plant minimum iteration limit of 2. It is clearly evident that convergence problem at the plant min iteration limit affected this branch and develop.


FacilityELectricAndGasUse

mitchute:
@Nigusse Excellent find with regard to the plant iterations. @mjwitte @Myoldmopar any thoughts on how to proceed with this? Perhaps potentially bump the plant iterations up on the files with diffs and let CI take a pass, then revert them back out so we can clearly show the issues?

This has fallen behind develop *again*. I'll pull develop in one more time and let CI take one more pass, but I think this should be able to go in soon.

mitchute:
@Nigusse @mjwitte @Myoldmopar 

I spent some time this morning looking at the latest CI results on this. It looks like the plant iterations issue referenced in #7653 may be at least partially causing the diffs, but that doesn't totally explain it. I ran the following files (which have big meter diffs) locally and looked at the results before and after Min Plant Iterations was set to 10.

- 5ZoneCoolingPanelBaseboard
- 5ZoneCoolingPanelBaseboardAuto
- 5ZoneCoolingPanelBaseboardTotalLoad
- 5ZoneCoolingPanelBaseboardVarOff
- 5ZoneWaterCooled_Baseboard
- 5ZoneWaterCooled_BaseboardScalableSizing
- MultiSpeedHP_StagedThermostat

With Min Plant Iterations set to 10, the MultSpeedHP_StagedThermostat is the one with the largest relative meter diffs of about 0.2% on the Electricity:HVAC \[J](Monthly) meter. See the following plots from this file.

Var_16
Var_17
Var_18
Var_19
Var_20
Var_21
Var_22
Var_23

It looks like the reported variables for this zone, with the minor exception of the humidity ratio, are all behaving nearly identical to the develop branch. The humidity ratio during a portion of the interval plotted is off by about 1%. That seems a *little* odd, however, this psychrometric function is used all over the place and this is the file with the biggest difference. So, IMO, if this is the biggest problem this work is ready to be merged.

Take a look and let me know. Unless I hear objections, I'll merge this in by the end of the week.

mitchute:
This is ready to go. Merging.

PR #7483, opened 2019-09-02
Revert fluid inlet temp change for DX heating coil

rraustad:
Pull request overview
---------------------
I think your review eye found a change that did not belong. Code changes include accounting for compressor amb temp for crankcase heater operation and accounting for condenser inlet temp for limit testing of operational limits. For crankcase heater, the CompAmbTemp variable should be used. For operational limits, the condenser inlet temp should be used, a.k.a. OutdoorDryBulb.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?

### DIFF
--- a/src/EnergyPlus/DXCoils.cc
+++ b/src/EnergyPlus/DXCoils.cc
@@ -10375,7 +10375,7 @@ namespace DXCoils {
         }
 
         if ((AirMassFlow > 0.0) && (GetCurrentScheduleValue(DXCoil(DXCoilNum).SchedPtr) > 0.0) && (PartLoadRatio > 0.0) &&
-            CompAmbTemp > DXCoil(DXCoilNum).MinOATCompressor) {
+            OutdoorDryBulb > DXCoil(DXCoilNum).MinOATCompressor) {
             // for cycling fan, reset mass flow to full on rate
             if (FanOpMode == CycFanCycCoil) AirMassFlow /= PartLoadRatio;
             if (FanOpMode == ContFanCycCoil) AirMassFlow *= AirFlowRatio;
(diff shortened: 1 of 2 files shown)


rraustad:
So let's see what CI thinks. No diff's previously means no diff's now. So that means no water cooled DX example files or unit tests.

src/EnergyPlus/DXCoils.cc
```diff
@@ -10375,7 +10375,7 @@ namespace DXCoils {
         }
 
         if ((AirMassFlow > 0.0) && (GetCurrentScheduleValue(DXCoil(DXCoilNum).SchedPtr) > 0.0) && (PartLoadRatio > 0.0) &&
-            CompAmbTemp > DXCoil(DXCoilNum).MinOATCompressor) {
+            OutdoorDryBulb > DXCoil(DXCoilNum).MinOATCompressor) {
```
rraustad, line 10378: Revert this one change.

src/EnergyPlus/VariableSpeedCoils.cc
```diff
@@ -7084,19 +7084,19 @@ namespace VariableSpeedCoils {
 
         // PURPOSE OF THIS FUNCTION:
         // This function looks up the the min oat for the cooling coil compressor and returns it.  If
-        // incorrect coil type or name is given, ErrorsFound is returned as true and value is returned
-        // as negative.
+        // incorrect coil index is given, ErrorsFound is returned as true and value is returned
+        // as negative 1000.
```
rraustad, line 7088: And address your other comments.

rraustad:
So the justification here is that I purposely changed CompAmbTemp as the local ambient temp for the crankcase heater. Then I mistakenly changed the one test for condenser inlet temp to use this same variable which I think should only affect DX heating coil water-cooled. I know that water-cooled inlet temp for DX coils has been corrected in the past so it is important that the previous correction is not corrupted.

src/EnergyPlus/DXCoils.cc
rraustad: This code is now back to what it was previously. Looks good to me.

Issue #4397, opened 2014-08-19
interzone surface material checks fail for a single pane of glass with asymmetrical properties (different front/back side properties).


BobHenninger:
Helpdesk Ticket 9441

Log of User and MJW comments

Here is an exemple,

 It’s some testing I’ve done in order to model reflective heat shield in a greenhouse.
 Even if it’s not the model I retain, it’s possible to see how implementing such a situation.

 Florian

> After other testing, I can simulate defining two different
> constructions. I have a warning message, but not a fatal error.
> 
> The results should be correct this way, but there is still a problem
> if the user define the Outside Boundary Condition as zone and let E+
> defining the surface in the other zone : reversing the layer will not
> respect the back face and front face property.
> 
> Florian

 Thank you for pointing this out. The case of a single layer of glass with asymmetrical properties is a special case that was not considered when implementing the interzone surface checks and the automated interzone surface creation. We will need to add some special logic for this. It would be helpful if you would reply and attach your working idf (the one with two different constructions that gives a warning).
 Thanks,
 Mike

User IDF file
Modele2ZonesEcran.idf

RKStrand:
@mjwitte Looking at this now after finally getting the user IDF to work (some special characters and other issues were causing a crash).  Note that the error is actually only a warning message so I think it actually runs the file properly, just complains about it.  Right now, all it does is check to see if the material names line up.  There is no checking regarding properties.  So, regarding multi-pane interzone windows, if they have different properties on front/back that is not being checked.  That could get a bit tricky but I'll see what I can do.

mjwitte:
Closed via #7487

PR #7487, opened 2019-09-04
Correction of Reversed Interzone Construction Check for Asymmetric Material Layers

RKStrand:
Pull request overview
---------------------
Most material layers are homogeneous and have the same properties throughout.  Constructions that consist of these layers are easy to check when they are used for interzone surfaces because the layers just have to be matched in reverse order.  For certain materials that have properties that differ on the front and back side, a new material layer has to be created that has these front and back side properties reversed to make sure that the construction is in fact a reverse of the original construction.  This enhanced check was not done previously, leading to error messages when the user did this properly and also no error messages when in fact there should have been a problem because one construction was not the mirror of its interzone pair.  The solution to this problem includes additional check for glass materials that have front and back side properties that are different.  This resolves GitHub Issue #4397.

### Work Checklist
   - Defect: This pull request repairs a github defect issue #4397

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [x] Argument types
 - IDD changes: n/a


src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -13264,6 +13260,92 @@ namespace SurfaceGeometry {
         Surface(SurfNum).Width = WidthEff;
         Surface(SurfNum).Height = HeightEff;
     }
+    
+
+    void CheckForReversedLayers(bool &RevLayerDiffs,    // true when differences are discovered in interzone constructions
```
mjwitte, line 13281: Would prefer for this to be a bool function and drop the first argument, but only if there are other changes needed.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -13264,6 +13260,92 @@ namespace SurfaceGeometry {
...
+                                int const ConstrNum,    // construction index
+                                int const ConstrNumRev, // construction index for reversed construction
+                                int const TotalLayers   // total layers for construction definition
+    )
+    {
+        
+        int LayerNo;
+        RevLayerDiffs = false;
+        
+        for (LayerNo = 1; LayerNo <= TotalLayers; ++LayerNo) {
```
mjwitte: Again, nitpicking, prefer to declare the int here and delete it above:
for (int LayerNo = 1 . . . .

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -13264,6 +13260,92 @@ namespace SurfaceGeometry {
...
+                        (thisMatLay.SolarDiffusing != revMatLay.SolarDiffusing) ||
+                        (abs(thisMatLay.YoungModulus - revMatLay.YoungModulus) > SmallDiff) ||
+                        (abs(thisMatLay.PoissonsRatio - revMatLay.PoissonsRatio) > SmallDiff)) {
+                        RevLayerDiffs = true;
+                        break; // exit when diff
+                    }   // If none of the above conditions is met, then these should be the same layers in reverse (RevLayersDiffs = false)
+                } else if ((thisMatLay.Group == GlassEquivalentLayer) && (revMatLay.Group == GlassEquivalentLayer)) {
+                    auto &thisMatLay(Material(thisConstLayer));
+                    auto &revMatLay(Material(revConstLayer));
+                    Real64 const SmallDiff = 0.0001;
```
mjwitte: all three of these appear to be duplicate declarations?

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -13264,6 +13260,92 @@ namespace SurfaceGeometry {
...
+        for (LayerNo = 1; LayerNo <= TotalLayers; ++LayerNo) {
+            auto &thisConstLayer(Construct(ConstrNum).LayerPoint(LayerNo));
+            auto &revConstLayer(Construct(ConstrNumRev).LayerPoint(TotalLayers - LayerNo + 1));
+            auto &thisMatLay(Material(thisConstLayer));
+            auto &revMatLay(Material(revConstLayer));
+            if ((thisConstLayer != revConstLayer) ||    // Not pointing to the same layer
+                (thisMatLay.Group == WindowGlass) ||    // Not window glass or glass equivalent layer which have
+                (revMatLay.Group == WindowGlass) ||     // to have certain properties flipped from front to back
+                (thisMatLay.Group == GlassEquivalentLayer) ||
+                (revMatLay.Group == GlassEquivalentLayer)) {
```
mjwitte, line 13299: These long ORs were hard for me to follow. I think I understand them now. Not sure how to suggest making this more clear, however.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -4432,3 +4432,158 @@ TEST_F(EnergyPlusFixture, SurfaceGeometry_CreateInternalMassSurfaces)
...
+    Construct(1).LayerPoint(2) = 2;
+    Construct(1).LayerPoint(3) = 3;
+    Construct(2).TotLayers = 3;
+    Construct(2).LayerPoint(1) = 3;
+    Construct(2).LayerPoint(2) = 2;
+    Construct(2).LayerPoint(3) = 1;
+    RevLayerDiffs = true;
+    ExpectResult = false;
+    CheckForReversedLayers(RevLayerDiffs, 1, 2, 3);
+    EXPECT_EQ(RevLayerDiffs, ExpectResult);
```
mjwitte: Simpler to use EXPECT_FALSE(RevLayerDiffs);

mjwitte (commented):
Tested this with the user defect file, and with some variations adding more layers. The checks are working as expected.
I've noted a few nit-picky code things, but the duplicate declarations should be fixed for sure.
I've merged in develop and resolved the conflict and pushed that up to github.
And the unit tests all run ok serially.

src/EnergyPlus/SurfaceGeometry.cc
RKStrand: @mjwitte Yeah, had done that early on and ran into some sort of weird problems.  I don't remember what the issue was anymore, but doing this resolved the issue.  My preference would have been a bool function, too.  I'll see what shakes out with the other comments...

src/EnergyPlus/SurfaceGeometry.cc
RKStrand: The problem here is that there are a lot of input parameters that need to be checked and some have to agree with the reverse side.  I agree that these are pretty complex.  I don't know how to make them less complex since all of those checks have to be made.

src/EnergyPlus/SurfaceGeometry.cc
RKStrand: @mjwitte Hmmm...looks like I forgot to eliminate these declarations after moving them up a level.  I will get rid of those.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
RKStrand: My bad...I should have used the EXPECT_TRUE and EXPECT_FALSE functions rather than these.

src/EnergyPlus/SurfaceGeometry.cc
RKStrand: @mjwitte Now I remember...it was some weird complaint about setting the value of a bool function and then resetting it.  I also found the logic here a bit more complex than a simple set it and return immediately.  So, I did this because of the complexity.  I'd like to leave this as is if you don't mind.  I will fix a couple of the other things though (see other comments).

src/EnergyPlus/SurfaceGeometry.cc
RKStrand: Will fix this.

mjwitte (approved):

RKStrand:
Thanks, Mike!

On Sep 18, 2019, at 8:50 PM, Michael J. Witte <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #7487<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_NREL_EnergyPlus_pull_7487&d=DwMCaQ&c=OCIEmEwdEq_aNlsP4fF3gFqSN-E3mlr2t9JcDdfOZag&r=rxT5qGz4Fgu2V4OupeO-GHp4q_5AzQNUBLnrS4uzhd8&m=-7Ku3JFXPvAfnaCQf43bjYTWlqT9OU1VU79KBo_vE2U&s=fIx1WCf14pbXldIHe0mqIKUtQy9JbCpQsU1b5c5Sp6c&e=> into develop.

PR #7488, opened 2019-09-05
Boiler:Steam PlantComponent Refactor

Myoldmopar:
Pull request overview
---------------------
Continued effort to convert plant components to new interface, this time Boiler:Steam.  Should be no diffs except bnd files.


Myoldmopar:
Of the 7 files with steam boilers, I'm seeing annual runtime reductions on the order of ~2.2% on average.  5ZoneSteamBaseboard is consistently much higher, almost 10%, while PackagedTerminalAirConditioner is always slightly slower with this branch, which is quite unexpected.  But these changes are not expected to produce dramatic gains, this is just a step to get the interface fleshed out to all the components.  This is good to go in.

tidecoder:
Thanks. This is a extremely good demo for EnergyPlus refactoring.

PR #7494, opened 2019-09-10
Cooling tower plant component refactor

Myoldmopar:
Pull request overview
---------------------
Continued plant component refactor, this time the cooling towers


.github/pull_request_template.md
```diff
@@ -1,34 +1,30 @@
...
-   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
-   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog
+ - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
+ - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
+ - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
+ - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
+ - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
+ - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
+ - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
+ - [ ] If structural output changes, add to output rules file and add OutputChange label
```
rraustad, line 18: This is a beautiful thing!

- [ ] If figures are included with documents, include figure name in CMakeLists.txt (e.g., ...\EnergyPlus\doc\input-output-reference\CMakeLists.txt)
- [x] I'm sure there are other requirements like this ... and other things I don't even know about

Myoldmopar: Thanks @rraustad, I think we can make better use of this checklist as it gets tidied up.  I'll add your figure name entry here as well.

rraustad: And let's not forget those critical PR tags.
- [ ] If table columns are added include OutputChange tag so that PR review process can begin before deadline
- [ ] If IDD is changed include IDDChange tag so that PR review process can begin before deadline

etc...

rraustad: Oh, you already have OutputChange label. Also, should reviewer run develop and this branch using the defect file, or the developer? It's seems the developer would have already done that and could easily provide a plot of the results. We don't want to duplicate effort.

Myoldmopar: So the reason a reviewer would also want to do that is twofold: one it gives someone a chance to evaluate the fix with a fresh set of eyes, and also if the develop branch has moved quite a bit since the PR was issued, the combination of this branch and the new develop could have side effects that cause unit tests to fail or IDFs to fail.  It's a good idea to have a reviewer pull updated develop branch in and build and run.  The reviewer shouldn't have to do any investigating though, just reproducing what was already seen.

rraustad: Ripe for a Wiki page...eventually...nice work.

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -2363,8 +2363,9 @@ namespace CondenserLoopTowers {
 
         // PURPOSE OF THIS SUBROUTINE:
         // Initialize the simulation variables.
-
-
+        for (auto & thisTower : SimpleTower) {
+            thisTower.__AirFlowRateRatio = 0.0;
+        }
```
Myoldmopar, line 2197: This variable must need to be initialized at the beginning of the calc or init routines.  The other variables didn't have this issue, so I just need to unify the usage.

Myoldmopar:
5ZoneWaterSystems is the failing regression file.  The table diffs in that file are due to shifted row order, the values did not change at all, which is why you don't see any mathdiffs.

- BND diffs expected due to reordered initialization
- MTD, MDD and RDD diffs expected due to delayed setup output variable calls
- ERR diffs are due to increased warnings that were being muted accidentally via a shared flag across all towers, now each tower holds their own flag and can individually warn as needed.

This is good to go in.

PR #7495, opened 2019-09-10
Corrected MultiSpeed DX Cooling Coil sizing variable names and capacity reporting issues

### LINKED ISSUE #7477: MultiSpeed DX Cooling Coil sizing variable names and capacity reporting issues
## Issue overview
--------------
Multispeed DX cooling coil has the following reporting inconsistencies:

- Coil:Cooling:DX:MultiSpeed lists, e.g., "Speed 2 User-Specified ... Capacity" while  Coil:Heating:DX:MultiSpeed lists, e.g., "User-Specified Speed 2 ... Capacity".
- Coil:Cooling:DX:MultiSpeed lists the capacity at only max speed (e.g., 4) while Coil:Heating:DX:MultiSpeed lists the capacity at all speeds (e.g., 1, 2, 3, and 4).
- Coil:Cooling:DX:MultiSpeed only lists the total cooling capacity while Coil:Cooling:WaterToAirHeatPump:EquationFit lists both sensible and total cooling capacities.

### Details
 -  @shorowit commented in pull request https://github.com/NREL/EnergyPlus/pull/7436#issuecomment-525591009

### Checklist

Nigusse:
Pull request overview
---------------------
Fixes #7477. This fixes inconsistent reporting in multi-speed DX cooling and heating coils:
- corrected inconsistent gross rated total capacity, SHR and rated flow rates names for different speeds
- now the reporting includes for all speed level for multi-speed DX cooling coils
- diffs are expected/justified based on this change in tabular reporting.
- resolved design size values difference when DX coils are fully and partially autosized
- SHR values now sized using the autosided capacities when design size values are available

### Work Checklist

### Review Checklist
 - [x] Unit Test(s)
 - C++ checks:
   - [ ] Argument types


Nigusse:
This is ready for review.

Nigusse:
@Myoldmopar @rraustad Note that I have merged in issue #7381 to this branch.

Myoldmopar:
Putting this as milestone 9.2 for now since it doesn't appear at first glance to change I/O in a structural manner, but I need to look at the details before making a final decision on that.  We'll be freezing the code next week so it may be too late anyway, but if it is clean it would be nice to drop it in.

Myoldmopar:
Unlikely to get this reviewed before we make final cutoff tomorrow afternoon.  I'm going to shift the milestone back to future.

Nigusse:
SQL output for develop: the rated sizing parameters (Capacity, flow rates, ...) are not reported for all speeds. Note also that the user specified variable naming is not consistent.

Develop_MultiSpeedDXCoils

Nigusse:
SQL output for this branch: the rated sizing parameters (Capacity, flow rates, ...) are reported for all speeds of cooling and heating coils. Note also that the user specified sizing variables naming is now consistent.

ThisBranch_MultiSpeedDXCoils

Myoldmopar:
@Nigusse I believe this branch is good to go, but with the merge of #7436, there are conflicts.  If you can resolve them I'll get this merged in.

Nigusse:
@Myoldmopar working on it now.

Nigusse:
@Myoldmopar Merged in develop and run the unit test suite. All happy.

Nigusse:
The multi-speed DX cooling coil capacity sizing related report variable names were modified for consistency. Also the design rated capacity at lower speeds are now corrected, the lower speed design cooling capacity are fraction of the design capacity at maximum speed, instead of the user specified maximum speed capacity.

mjwitte (changes requested):
1. Tested this vs develop and saw the expected changes using multispeed.idf as-is (moved to v9.2).
2. The name re-ordering in the sizing labels warrants an entry in \src\Transition\OutputRulesFiles\OutputChanges9-2-0-to-9-3-0.md
3. Added sizing objects to the defect file and expected to see a reported "Design Size . . . " for every "User-Specified . . .". This was true for all except "Speed 2 Gross Rated Heating Capacity", the design size output is missing for this one. See After-multispeed-AddSizing.eio
4. Also made a fully autosized version of the defect file and compared to the hard-sized file with sizing. I was surprised that the Coil:Cooling:DX:MultiSpeed, ASHP CLG COIL, Design Size Speed 2 Gross Rated Total Cooling Capacity was not the same in both files. So maybe that's because the capacity design size is being calculated based on the hard airflow rate? So, I made the airflow rate and the shr autosize in both files and the design gross cooling capacity is still different between the two. Is it broken, or am I missing something?  Compare the eio output for After-multispeed-AddSizing-AllAutosize vs After-multispeed-AddSizing-AutoSizeSHRAndAirFlow.

Test files attached: [7477-MulitSpeedDX.zip](https://github.com/NREL/EnergyPlus/files/3719478/7477-MulitSpeedDX.zip)

Nigusse:
"Added sizing objects to the defect file and expected to see a reported "Design Size . . . " for every "User-Specified . . .". This was true for all except "Speed 2 Gross Rated Heating Capacity", the design size output is missing for this one. See After-multispeed-AddSizing.eio"

@mjwitte I moved the cooling capacity sizing calc to Requestsizing() function. I think, this function does not report the design capacity (may be any design variable) unless the user specified value is different from the design value by more than 10%. This is a quick reply to item 3.

mjwitte:
@Nigusse I may be wrong, but my understanding is that both values should always be reported, but the >10% difference triggers a warning (with DisplayExtraWarnings?).

rraustad:
If not autosized, and difference is greater than 0.1, both are reported, otherwise only user specified is reported.

```
            if (PrintWarningFlag) {
                if (IsAutoSize && SizingResult > 0.0) {
                    ReportSizingOutput(CompType, CompName, "Design Size " + SizingString, SizingResult);
                } else if (SizingResult > 0.0) {
                    AutosizeUser = SizingResult;
                    if ((std::abs(AutosizeDes - AutosizeUser) / AutosizeUser) > AutoVsHardSizingThreshold) {
                        ReportSizingOutput(
                            CompType, CompName, "Design Size " + SizingString, AutosizeDes, "User-Specified " + SizingString, AutosizeUser);
                    } else {
                        ReportSizingOutput(CompType, CompName, "User-Specified " + SizingString, AutosizeUser);
                    }
```

mjwitte:
@rraustad @Nigusse I didn't realize the reporting was suppressed for <10% difference. So, OK, if I change that to exceed 10% then I see design reported for every user value. So, that part's ok.

Any thoughts on the other question about fully autosized capacity being different from the design values reported when user-specified?

Nigusse:
@mjwitte addressed item 2. This branch was meant to include a final fix for issues #7381 and #7477. Now looking at item 4.

Nigusse:
@mjwitte @rraustad The main reason why these two defect files show design size cooling capacity difference is due to differences in sizing method. When it is fully auto-sized, the capacity determined  using the RequestSizing() function from the parent object (UnitarySystem) and is set for the DX coil. When there is hard sizing value specified for the DX coil capacity, the design capacity is determined using the autosizing calculation in the DXCoil object. These two sizing methods are slightly different.  The RequestSizing() function accounts for fan heat effect in the sizing calculation, while the other is not. I think the way forward in resolving this problem is to use the RequestSizing() in both places. I think, this defect must have been there longer.

Nigusse:
@mjwitte As you have pointed out partial autosizing of multi-speed DX coils fields has been broken. I have found partial autosizing problem for Rated Total Cooling/Heating Capacity, Rated Air Flow Rates, Rated Sensible Heat Ratio, Evaporative Condenser Air Flow Rate, etc. All capacity and flow rate related sizing variables were broken or did not work in the first place when there is partial autosizing. Fixing these issues has broken else where, so reverted some of the changes I made. The remaining issues  will be addressed after reviewing the this push, perhaps this issue should be divided into multiple bug issues.

Nigusse:
Fully autosized Multispeed DX coils results:

FullyAutosizeOfMSpeedDXCoils

Nigusse:
Partially autosized Multispeed DX coils results:

PartialAutosizeOfMSpeedDXCoils

Nigusse:
This fix resolves the mismatch between design size values when fully and partially autosized.

mjwitte (approved):
@Nigusse Thanks for addressing *all* of the comments. Output looks exactly as expected now. Merging in develop for one more CI round, then this can drop in.

PR #7498, opened 2019-09-12
Follow up to 7249   changeover bypass vav enhancement

rraustad:
Pull request overview
---------------------
Carrier's review of new CBVAV mod's to allow connection to OA system showed odd behavior. The cooling coil would turn on briefly when the operating mode changed from heating to cooling.

### Work Checklist

### Review Checklist
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
 - IDD changes:
 - [ ] ExpandObjects changes?


rraustad:
Carrier test file showed odd spikes from cooling coil when operating mode change from heating to cooling. Red line shows operating mode, heating = 2, cooling = 1, off = 0. During the winter day the cold outdoor air is sufficient to meet the cooling load. The spike occurs because the OA system mixed air SP isn't in sync with the CBVAV controls, on the next time step the OA system controls catch up and the coil turns off.

image

After hours of review, the mixed air node set point temperature was not updated within the HVAC iteration loop and caused the OA system to NOT meet the correct set point (using SetPointManger:MixedAir based on CBVAV outlet node SP). Calling the specific SPM from within CBVAV corrected the problem. The OA system now hits the mixed air SP and the coil does not need to operate.

image

src/EnergyPlus/HVACManager.cc
```diff
@@ -702,7 +702,6 @@ namespace HVACManager {
         using PlantUtilities::CheckPlantMixerSplitterConsistency;
         using PlantUtilities::SetAllPlantSimFlagsToValue;
         using SetPointManager::ManageSetPoints;
-        using SimAirServingZones::ManageAirLoops;
```
rraustad, line 705: not used in this function

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -87,6 +87,7 @@
 #include <Psychrometrics.hh>
 #include <ReportSizingManager.hh>
 #include <ScheduleManager.hh>
+#include <SetPointManager.hh>
```
rraustad, line 90: will be calling specific SPM from CBVAV, this may come in handy until ManageSetPoints is moved into the HVAC iteration loop.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -311,6 +312,14 @@ namespace HVACUnitaryBypassVAV {
...
             CalcCBVAV(CBVAVNum, FirstHVACIteration, PartLoadFrac, QSensUnitOut, OnOffAirFlowRatio, HXUnitOn);
         }
+        if (CBVAV(CBVAVNum).modeChanged) {
+            // set outlet node SP for mixed air SP manager
+            DataLoopNode::Node(CBVAV(CBVAVNum).AirOutNode).TempSetPoint = CalcSetPointTempTarget(CBVAVNum);
+            if (CBVAV(CBVAVNum).OutNodeSPMIndex > 0) { // update mixed air SPM if exists
+                SetPointManager::MixedAirSetPtMgr(CBVAV(CBVAVNum).OutNodeSPMIndex).calculate(); // update mixed air SP based on new mode
+                SetPointManager::UpdateMixedAirSetPoints(); // need to know control node to fire off just one of these, do this later
+            }
+        }
```
rraustad, line 322: Note the modeChanged flag, that will be apparent later. Here the outlet node SP is set and the specific SPM is called. I didn't go to the length needed to Update the specific SPM and just update them all.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1511,6 +1520,9 @@ namespace HVACUnitaryBypassVAV {
             //   Set UnitarySys flag to FALSE and let the heating coil autosize independently of the cooling coil
             DataAirLoop::AirLoopControlInfo(AirLoopNum).UnitarySys = false;
             DataAirLoop::AirLoopControlInfo(AirLoopNum).FanOpMode = CBVAV(CBVAVNum).OpMode;
+            // check for set point manager on outlet node of CBVAV
+            CBVAV(CBVAVNum).OutNodeSPMIndex =
+                SetPointManager::getSPMBasedOnNode(OutNode, SetPointManager::iCtrlVarType_Temp, SetPointManager::iSPMType_MixedAir, SetPointManager::CtrlNodeType::reference);
```
rraustad, line 1525: grab the specific SPM index for a mixed air SPM if connected to the CBVAV outlet node.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1705,6 +1717,7 @@ namespace HVACUnitaryBypassVAV {
         }
 
         // Returns load only for zones requesting cooling (heating). If in deadband, Qzoneload = 0.
+        if (FirstHVACIteration) CBVAV(CBVAVNum).modeChanged = false;
```
rraustad, line 1720: reset the modeChanged flag on FirstHVACIteration so SPM is not called unnecessarily.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -1804,19 +1817,13 @@ namespace HVACUnitaryBypassVAV {
                 DataLoopNode::Node(CBVAV(CBVAVNum).MixerInletAirNode).MassFlowRate = CompOnMassFlow;
                 DataLoopNode::Node(CBVAV(CBVAVNum).MixerOutsideAirNode).MassFlowRate = OACompOnMassFlow;
                 DataLoopNode::Node(CBVAV(CBVAVNum).MixerReliefAirNode).MassFlowRate = OACompOnMassFlow;
-                BypassDuctFlowFraction = 0.0;
```
rraustad: clean up some places where duplicate code sets BypassDuctFlowFraction

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -2207,8 +2214,7 @@ namespace HVACUnitaryBypassVAV {
                                                                                 CBVAV(CBVAVNum).CoolCoilCompIndex,
                                                                                 DataHVACGlobals::ContFanCycCoil,
                                                                                 HXUnitOn);
-                        } else if (DataLoopNode::Node(CBVAV(CBVAVNum).DXCoilOutletNode).Temp < CBVAV(CBVAVNum).CoilTempSetPoint &&
-                                   DataLoopNode::Node(CBVAV(CBVAVNum).DXCoilInletNode).Temp > CBVAV(CBVAVNum).CoilTempSetPoint) {
+                        } else if (DataLoopNode::Node(CBVAV(CBVAVNum).DXCoilOutletNode).Temp < CBVAV(CBVAVNum).CoilTempSetPoint) {
```
rraustad, line 2225: Cleaning up unnecessary logic, inlet node T must be greater than CoilTempSP here, otherwise it's caught in the IF above this line

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -2276,7 +2282,7 @@ namespace HVACUnitaryBypassVAV {
                                            DataHVACGlobals::ContFanCycCoil,
                                            PartLoadFrac,
                                            OnOffAirFlowRatio);
-                        if (DataLoopNode::Node(CBVAV(CBVAVNum).DXCoilInletNode).Temp < CBVAV(CBVAVNum).CoilTempSetPoint) {
+                        if (DataLoopNode::Node(CBVAV(CBVAVNum).DXCoilInletNode).Temp <= CBVAV(CBVAVNum).CoilTempSetPoint) {
```
rraustad, line 2293: similar clean up to avoid turning on the coil at very low PLR if inlet T = CoilTempSP

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3351,6 +3356,7 @@ namespace HVACUnitaryBypassVAV {
         int splitterOutNode = CBVAV(CBVAVNum).SplitterOutletAirNode;
         DataLoopNode::Node(splitterOutNode).MassFlowRateSetPoint = DataLoopNode::Node(OutletNode).MassFlowRateSetPoint;
         DataLoopNode::Node(OutletNode) = DataLoopNode::Node(splitterOutNode);
+        DataLoopNode::Node(OutletNode).TempSetPoint = CBVAV(CBVAVNum).OutletTempSetPoint;
```
rraustad, line 3367: Go ahead and actually set node TempSetPoint since now the CBVAV can be used with OASystems. OutletTempSetPoint was used internally before and now a mixed air SPM could be used.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3397,7 +3403,10 @@ namespace HVACUnitaryBypassVAV {
         Real64 thisTime = (dayOfSim - 1) * 24 + DataGlobals::HourOfDay - 1 + (DataGlobals::TimeStep - 1) * DataGlobals::TimeStepZone +
                           DataHVACGlobals::SysTimeElapsed;
 
-        if (thisTime <= CBVAV(CBVAVNum).changeOverTimer) return;
+        if (thisTime <= CBVAV(CBVAVNum).changeOverTimer) {
+            CBVAV(CBVAVNum).modeChanged = true;
+            return;
+        }
```
rraustad, line 3417: Now here, the mode has changed and this function is not needed so a return is used. But this defect showed that the SPMs needed to be called so the modeChanged flag also needs to stay true during the min time that the operating mode must be active (e.g., for at least 0.5 hours) so the SPM can be called at line 315

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3502,6 +3511,7 @@ namespace HVACUnitaryBypassVAV {
         if (CBVAV(CBVAVNum).LastMode != CBVAV(CBVAVNum).HeatCoolMode) {
             CBVAV(CBVAVNum).changeOverTimer = thisTime + CBVAV(CBVAVNum).minModeChangeTime;
             CBVAV(CBVAVNum).LastMode = CBVAV(CBVAVNum).HeatCoolMode;
+            CBVAV(CBVAVNum).modeChanged = true;
```
rraustad, line 3522: At the bottom of this function the timer is set and the modeChanged flag also needs to be set so the SPM can be called.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3565,12 +3575,12 @@ namespace HVACUnitaryBypassVAV {
 
             Real64 CpSupplyAir = Psychrometrics::PsyCpAirFnWTdb(OutAirHumRat, OutAirTemp);
 
-            //     Find the supply air temperature that will force the box to full flow
+            // Find the supply air temperature that will force the box to full flow
             if (BoxOutletNodeNum > 0) {
-                if (CpSupplyAir * DataLoopNode::Node(BoxOutletNodeNum).MassFlowRateMax == 0.0) {
+                if (DataLoopNode::Node(BoxOutletNodeNum).MassFlowRateMax == 0.0) {
```
rraustad, line 3588: Don't need CpAir in the conditional, it's just a constant. If Mdot = 0, then CpAir * Mdot also = 0 so don't need CpAir here

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3603,23 +3613,24 @@ namespace HVACUnitaryBypassVAV {
...
         //   Account for floating condition where cooling/heating is required to avoid overshooting setpoint
-        if (CBVAV(CBVAVNumber).HeatCoolMode == 0 && CBVAV(CBVAVNumber).OpMode == DataHVACGlobals::ContFanCycCoil) {
-            if (OutAirTemp > TSupplyToCoolSetPtMin) {
-                CalcSetPointTempTarget = TSupplyToCoolSetPtMin;
-            } else if (OutAirTemp < TSupplyToHeatSetPtMax) {
-                CalcSetPointTempTarget = TSupplyToHeatSetPtMax;
-            } else {
-                CalcSetPointTempTarget = OutAirTemp;
+        if (CBVAV(CBVAVNumber).HeatCoolMode == 0) {
+            if (CBVAV(CBVAVNumber).OpMode == DataHVACGlobals::ContFanCycCoil) {
```
rraustad, line 3625: This code hasn't changed, I just moved the additional conditional inside so it doesn't have to be evaluated every time through unless HeatCoolMode = 0. Speed up change.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
```diff
@@ -3833,18 +3844,13 @@ namespace HVACUnitaryBypassVAV {
...
-                BypassDuctFlowFraction = 0.0;
-            } else {
-                OnOffAirFlowRatio = 1.0;
-                auto &cbVAVBoxOut(CBVAV(CBVAVNum).CBVAVBoxOutletNode);
-                Real64 boxOutletNodeFlow = 0.0;
-                for (int i = 1; i <= CBVAV(CBVAVNum).NumControlledZones; ++i) {
-                    boxOutletNodeFlow += DataLoopNode::Node(cbVAVBoxOut(i)).MassFlowRate;
-                }
-                BypassDuctFlowFraction = max(0.0, 1.0 - (boxOutletNodeFlow / AverageUnitMassFlow));
+            OnOffAirFlowRatio = 1.0;
```
rraustad, line 3854: So here the BypassDuctFlowFration was set = 0 and the model needed to iterate enough to get to the correct bypass fraction. I figured to just set it based on the TU box flow rate and always know the correct fraction.

src/EnergyPlus/SetPointManager.cc
```diff
@@ -916,6 +916,7 @@ namespace SetPointManager {
             }
             AllSetPtMgr(AllSetPtMgrNum).Name = SchSetPtMgr(SetPtMgrNum).Name;
             AllSetPtMgr(AllSetPtMgrNum).SPMType = iSPMType_Scheduled;
+            AllSetPtMgr(AllSetPtMgrNum).SPMIndex = SetPtMgrNum;
```
rraustad, line 919: The AllSetPtMgr array now knows which specific SPM is associated with this AllSetPtMgrNum index

src/EnergyPlus/SetPointManager.cc
```diff
@@ -1721,8 +1729,10 @@ namespace SetPointManager {
             }
             AllSetPtMgr(AllSetPtMgrNum).Name = MixedAirSetPtMgr(SetPtMgrNum).Name;
             AllSetPtMgr(AllSetPtMgrNum).SPMType = iSPMType_MixedAir;
+            AllSetPtMgr(AllSetPtMgrNum).SPMIndex = SetPtMgrNum;
             AllSetPtMgr(AllSetPtMgrNum).CtrlTypeMode = MixedAirSetPtMgr(SetPtMgrNum).CtrlTypeMode;
             AllSetPtMgr(AllSetPtMgrNum).NumCtrlNodes = MixedAirSetPtMgr(SetPtMgrNum).NumCtrlNodes;
+            AllSetPtMgr(AllSetPtMgrNum).RefNode = MixedAirSetPtMgr(SetPtMgrNum).RefNode;
```
rraustad, line 1735: And the mixed air SPM now holds the reference node in the AllSetPtMgr array

src/EnergyPlus/SetPointManager.cc
```diff
@@ -8429,6 +8459,43 @@ namespace SetPointManager {
         }
     }
 
+    int getSPMBasedOnNode(int const NodeNum, int const SetPtType, int const SPMType, CtrlNodeType ctrlOrRefNode) {
```
rraustad, line 8462: new function allows any model to find a specific SPM based on either a control node or a reference node.

tst/EnergyPlus/unit/HVACUnitaryBypassVAV.unit.cc
```diff
@@ -777,6 +777,7 @@ TEST_F(CBVAVSys, UnitaryBypassVAV_NoOASys)
     DataLoopNode::Node(MixedAir::OAMixer(1).InletNode).Enthalpy = 71299.267;
 
     DataLoopNode::Node(cbvav.CBVAVBoxOutletNode(1)).MassFlowRateMax = 0.61;
+    DataLoopNode::Node(cbvav.CBVAVBoxOutletNode(1)).MassFlowRate = 0.61;
```
rraustad, line 780: Now that BypassMassFlowFrac is only based on box flow, each unit test needed to include the box node flow rate.

src/EnergyPlus/SetPointManager.cc
```diff
@@ -8429,6 +8459,43 @@ namespace SetPointManager {
...
+                case CtrlNodeType::reference: {
+                        // SPMType is type of set point manager, e.g., iSPMType_Scheduled, iSPMType_MixedAir, etc.
+                        if (NodeNum == AllSetPtMgr(SetPtMgrNum).RefNode && SPMType == AllSetPtMgr(SetPtMgrNum).SPMType) {
+                            getSPMBasedOnNode = AllSetPtMgr(SetPtMgrNum).SPMIndex; // SPMIndex is index to specific type of SPM
+                        }
+                        break;
+                }
+                }
+            }
+            if (getSPMBasedOnNode > 0) break;
```
rraustad, line 8493: I wondered if I should break here or continue to find a duplicate if a user puts 2 of the same SPM on the same node. Since it's not a fatal if a user puts 2 SPMs on the same node (a warning is currently issued) then no need to continue (at this point in time, maybe needed later)

src/EnergyPlus/SetPointManager.hh
```diff
@@ -61,7 +61,10 @@ namespace EnergyPlus {
 
 namespace SetPointManager {
 
-    // Using/Aliasing
+    enum class CtrlNodeType: int {
+        control,
+        reference
+    };
```
rraustad, line 67: I'm not up to speed on these enums so critique away. Used in new function above in the switch/case.

rraustad:
This is what I get for trying to clean things up along the way. The zone VAV dampers are not opening as much as they were before. Before seems correct to minimize fan static pressure. So now the question is why is the CBVAV supply air temp higher now than before. These results are for example file ChangeoverBypassVAV.idf. Maybe I can revert something simple to get back to where it was but I thought I only got rid of things that weren't necessary.

image

rraustad:
Results are back to where they were before these changes. And the other test file with cooling coil spikes hasn't changed. You can see a slight hint of blue behind the orange line in the middle row for VAV 1 and 3 damper position but this seems trivial compared to the defect fix. Top plots, example file ChangeoverBypass.VAV, bottom plot is other test file that no longer has cooling spikes in winter.

image

image

rraustad:
Here's a snap shot of data from big diff # 2 file, ChangeoverBypassVAV_AirToAir. We might need a better way to denote real diff's.

image

And the CBVAV outlet node T and Mdot as a comparison of system performance between versions.

image

rraustad:
Big diff file 3 with a big diff shown in lower right figure.

image

With annual simulation results for end uses:
CBVAV_Annual

rraustad:
This is ready for review.

src/EnergyPlus/SetPointManager.hh
Myoldmopar: 👍 👍 the int type specifier isn't necessarily, but perfectly fine.

src/EnergyPlus/SetPointManager.cc
Myoldmopar: Seems reasonable for now and potentially changing later.

src/EnergyPlus/HVACUnitaryBypassVAV.cc
Myoldmopar: What if mdot is greater than zero but cp is zero?

🤣 🤣 🤣

Myoldmopar (approved):
I _really_ appreciate your walkthrough of these changes, it's very helpful in understanding your intent.  This all looks good to go.

Myoldmopar:
Defect file changes look good, analysis of test file difference is very helpful.  No build warnings anymore.  Good to go, thanks @rraustad !

PR #7517, opened 2019-09-20
Fix infinite loop in input processor

### LINKED ISSUE #7516: Alpha9_2 memory consumption
This could be an issue related to Alpha9_2 release and it's ready being addressed but 
@jmarrec discovered a process consuming excessive memory that you may want to take a look at. Here is comments by Julien. 

 I don't know what's going on, suddenly airterminal_fourpipebeam_osm (and _rb) are failling. energyplus uses all of my RAM (16GB on this machine) while running a single simulation, I don't think I've seen this ever before...
I've got a eplusout.dbg in the run folder which shouldn't happen.
Could the Alpha2 asset be a debug build of E+ or something?
Lookig at the in.idf that's generated, there's some funky business witht eh Table Objects
```
Table:IndependentVariable,
  CapModFuncOfWaterFlow 7_IndependentVariable_0, !- Name
  Cubic,                                  !- Interpolation Method
  Constant,                               !- Extrapolation Method
  0,                                      !- Minimum Value {BasedOnField A4}
  1.333333,                               !- Maximum Value {BasedOnField A4}
  ,                                       !- Normalization Reference Value {BasedOnField A4}
  Dimensionless,                          !- Unit Type
  ,                                       !- External File Name
  ,                                       !- External File Column Number
  ,                                       !- External File Starting Row Number
  0&#44000000,                            !- Value 1
  0&#44050000,                            !- Value 2
  0&#44333330,                            !- Value 3
  0&#44500000,                            !- Value 4
  0&#44666667,                            !- Value 5
  0&#44833333,                            !- Value 6
  1&#44000000,                            !- Value 7
  1&#44333333;                            !- Value 8
```
running the very similar airterminal_cooledbeam_osm test takes 120 MB of memory...
I  am using the binary posted by CI on Ubuntu 18.04.

mbadams5:
Pull request overview
---------------------
 - Fixes #7516 
 - This fixes an infinite loop in the input processor when invalid numbers are input in extensible fields. See defect file in #7516

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/InputProcessing/IdfParser.cc
+++ b/src/EnergyPlus/InputProcessing/IdfParser.cc
@@ -395,6 +395,7 @@ json IdfParser::parse_object(
             auto const size = legacy_idd_extensibles_array.size();
             std::string const &field_name = legacy_idd_extensibles_array[extensible_index % size];
             auto const val = parse_value(idf, index, success, schema_obj_extensions->at(field_name));
+            if (!success) return root;
             extensible[field_name] = std::move(val);
             was_value_parsed = true;
             extensible_index++;
@@ -415,6 +416,7 @@ json IdfParser::parse_object(
                 }
             } else {
                 auto const val = parse_value(idf, index, success, find_field_iter.value());
+                if (!success) return root;
                 root[field] = std::move(val);
             }
             if (!success) return root;
@@ -497,24 +499,19 @@ json IdfParser::parse_number(std::string const &idf, size_t &index, bool &succes
 
 json IdfParser::parse_value(std::string const &idf, size_t &index, bool &success, json const &field_loc)
 {
+    Token token;
     auto const &field_type = field_loc.find("type");
     if (field_type != field_loc.end()) {
         if (field_type.value() == "number" || field_type.value() == "integer") {
-            return parse_number(idf, index, success);
+            token = Token::NUMBER;
         } else {
-            auto const parsed_string = parse_string(idf, index, success);
-            auto const &enum_it = field_loc.find("enum");
-            if (enum_it == field_loc.end()) return parsed_string;
-            for (auto const &s : enum_it.value()) {
-                auto const &str = s.get<std::string>();
-                if (icompare(str, parsed_string)) {
-                    return str;
-                }
-            }
-            return parsed_string;
+            token = Token::STRING;
         }
     } else {
-        switch (look_ahead(idf, index)) {
+        token = look_ahead(idf, index);
+    }
+
+    switch (token) {
         case Token::STRING: {
             auto const parsed_string = parse_string(idf, index, success);
             auto const &enum_it = field_loc.find("enum");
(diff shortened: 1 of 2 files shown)


src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -395,6 +395,7 @@ json IdfParser::parse_object(
             auto const size = legacy_idd_extensibles_array.size();
             std::string const &field_name = legacy_idd_extensibles_array[extensible_index % size];
             auto const val = parse_value(idf, index, success, schema_obj_extensions->at(field_name));
+            if (!success) return root;
```
Myoldmopar, line 398: OK, so while parsing an object, just be more careful to catch malformed values.  Got it.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -415,6 +416,7 @@ json IdfParser::parse_object(
                 }
             } else {
                 auto const val = parse_value(idf, index, success, find_field_iter.value());
+                if (!success) return root;
```
Myoldmopar, line 419: And same thing here :heavy_check_mark:

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -497,24 +499,19 @@ json IdfParser::parse_number(std::string const &idf, size_t &index, bool &succes
 
 json IdfParser::parse_value(std::string const &idf, size_t &index, bool &success, json const &field_loc)
 {
+    Token token;
     auto const &field_type = field_loc.find("type");
     if (field_type != field_loc.end()) {
         if (field_type.value() == "number" || field_type.value() == "integer") {
-            return parse_number(idf, index, success);
+            token = Token::NUMBER;
```
Myoldmopar, line 506: So here you're more careful to not blindly parse a number, instead first setting what type you expect then parsing later.

third_party/valijson/valijson/validation_visitor.hpp
```diff
@@ -1026,8 +1026,37 @@ class ValidationVisitor: public constraints::ConstraintVisitor
...
+                    output_target = false;
+                }
+
+                if (output_target) {
+                    results->pushError(context,
+                                       "Value type \"" + type + "\" for input \"" + target.asString() + "\" not permitted by 'type' constraint.");
+                } else {
+                    results->pushError(context,
+                                       "Value type \"" + type + "\" not permitted by 'type' constraint.");
+                }
```
Myoldmopar, line 1058: Extra careful reporting here makes sense.

Myoldmopar (approved):

src/EnergyPlus/InputProcessing/IdfParser.cc
mbadams5: It is more than just that. Originally there was duplicate code for parsing number or string depending if it found "type" or not. So it first looked for "type" and would then parse directly based on that type in the schema. If it didn't find a "type" then it would try to determine if it was a number or string and parse it. The code for parsing number or string was different in these two approaches subtly. This refactors the code to determine if it is probably a string or number first then there is the code to parse number or string. So now both approaches use the more robust number and string parsing code.

Myoldmopar: :+1: thanks for the clarification!

Myoldmopar:
Verified this fix by building develop and witnessing the infinite loop, and then building this branch and getting meaningful messages in the error file with a graceful fatal error.  This is good to go, but I'll give CI a little longer to do a couple more builds.  Thanks for the prompt turnaround on this priority fix @mbadams5.  And thanks @jmarrec @tijcolem for testing out the alpha with sufficient time to get this fixed up.

PR #7528, opened 2019-09-25
AirloopHVAC:UnitarySystem used with other zone equipment does not meet the load when the zone thermostat is in the deadband

### LINKED ISSUE #7507: Load introduced by ZoneHVAC:EnergyRecoveryVentilator not always met by AirLoop systems
Issue overview
--------------
When modeling an ERV with AirLoopHVAC:UnitarySystem, Even with zone equipment fraction set to 1.0, in some hours, hvac still didn't meet 100% load.

    ZoneHVAC:EquipmentList,
    living Equipment List,                     !- Name
    SequentialLoad,                            !- Load Distribution Scheme
    ZoneHVAC:AirDistributionUnit,              !- Zone Equipment 1 Object Type
    ADU central ac and furnace living terminal,  !- Zone Equipment 1 Name
    2,                       !- Zone Equipment 1 Cooling Sequence
    2,                       !- Zone Equipment 1 Heating or No-Load Sequence
    1,                       !- Zone Equipment 1 Sequential Cooling Fraction
    1,                       !- Zone Equipment 1 Sequential Heating Fraction
    ZoneHVAC:EnergyRecoveryVentilator,  !- Zone Equipment 2 Object Type
    mech vent erv,           !- Zone Equipment 2 Name
    1,                       !- Zone Equipment 2 Cooling Sequence
    1,                       !- Zone Equipment 2 Heating or No-Load Sequence
    1,                       !- Zone Equipment 2 Sequential Cooling Fraction
    1;                       !- Zone Equipment 2 Sequential Heating Fraction

image

If added an ideal system at the last sequence as a backup :

    ZoneHVAC:EquipmentList,
      living Equipment List,                  !- Name
      SequentialLoad,                         !- Load Distribution Scheme
      ZoneHVAC:AirDistributionUnit,           !- Zone Equipment Object Type 1
      ADU central ac and furnace living terminal, !- Zone Equipment Name 1
      2,                                      !- Zone Equipment Cooling Sequence 1
      2,                                      !- Zone Equipment Heating or No-Load Sequence 1
      1,                                      !- Zone Equipment Sequential Cooling Fraction 1
      1,                                      !- Zone Equipment Sequential Heating Fraction 1
      ZoneHVAC:EnergyRecoveryVentilator,      !- Zone Equipment Object Type 2
      mech vent erv,                          !- Zone Equipment Name 2
      1,                                      !- Zone Equipment Cooling Sequence 2
      1,                                      !- Zone Equipment Heating or No-Load Sequence 2
      1,                                      !- Zone Equipment Sequential Cooling Fraction 2
      1,                                      !- Zone Equipment Sequential Heating Fraction 2
      ZoneHVAC:IdealLoadsAirSystem,           !- Zone Equipment Object Type 2
      ideal,                                  !- Zone Equipment Name 2
      3,                                      !- Zone Equipment Cooling Sequence 2
      3,                                      !- Zone Equipment Heating or No-Load Sequence 2
      1,                                      !- Zone Equipment Sequential Cooling Fraction 2
      1;                                      !- Zone Equipment Sequential Heating Fraction 2

There's always some load( 0.9-1.0 GJ running in Denver) distributed to ideal loads air system no matter how large the airloop system is.

image

So I guess zone equipment could handle ERV load while air loop is falling behind. 
Test files are ERV of 0 effectiveness (no heat recovery) with and without residual ideal loads air system, attached [here](https://drive.google.com/drive/folders/1zFN2dFPQg2JB9AvhGN_6CGbVygoI6ln7?usp=sharing)

### Details
 - Platform (Windows 10)
 - Version of EnergyPlus (both E+ 9.1 & 9.2)

### Checklist

rraustad:
Pull request overview
---------------------
 - Fixes #7507
 - When using AirloopHVAC:UnitarySystem with other zone equipment, the air loop will not pick up the load of other zone equipment (e.g., zone ERV) when the zone is initially in the thermostat deadband.
 - Corrected a sensible capacity equation in AirloopHVAC:UnitarySystem to include duct losses when AirTerminal:SingleDuct:Mixer is used.

### Pull Request Author
        Updated defect files on [DevSupport](https://github.com/NREL/EnergyPlusDevSupport/tree/master/DefectFiles/7000s/7507)

### Reviewer
 - [ ] ~~If feature, test running new feature, try creative ways to break it~~
 - [ ] ~~Check any new function arguments for performance impacts~~
 - [ ] ~~Verify IDF naming conventions and styles, memos and notes and defaults~~
 - [ ] ~~If new idf included, locally check the err file and other outputs~~


mjwitte:
@shorowit It would be great if you can pull this branch and test it with your suite of various equipment combinations to make sure this doesn't break anything.

rraustad:
Before and after results of defect file balanced-erv-object.idf:

image

rraustad:
Before and after comparison of second defect file: balanced-erv-residualideal.idf. The need for IdealLoadsAirSystem is mostly eliminated, except for about 30W a few times a year. Note change in Y2 axis limit in these figures. In this figure, the develop branch shows good control of zone temperature because the ideal loads system picked up any remaining zone load. Not sure if the minimal need for ideal loads in this branch (second figure) is another issue with UnitarySystem or ERV model, or simply a result of the tolerances used for convergence, or the methods used for calculating sensible delivered capacity between the models (UnitarySystem, AirTerminal, EnergyRecoveryVentilator, or IdealLoadsAirSystem).

image

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -8814,11 +8814,11 @@ namespace UnitarySystems {
         CoolingLoad = false;
         HeatingLoad = false;
 
-        if (QZnReq > Small5WLoad / this->ControlZoneMassFlowFrac && !DataZoneEnergyDemands::CurDeadBandOrSetback(this->ControlZoneNum)) {
+        if (QZnReq > Small5WLoad / this->ControlZoneMassFlowFrac) { // no need to check deadband flag, QZnReq is correct.
```
mjwitte: @rraustad A small suggestion for future efficiency improvement (not now - might introduce tiny diffs):
`Small5WLoad/this->ControlZoneMassFlowFrac` is constant over the course of a simulation, correct? So, could compute that one upfront and store in something like `this->SmallLoadTolerance` instead of recomputing this on every pass.

mjwitte (approved):
Ran defect files before and after - can see impact of fix.
No unit test - but this is a hotfix with minimal code changes and zero ci diffs.
So, I'm ok with this.
Permission to merge, @Myoldmopar ?

src/EnergyPlus/UnitarySystem.cc
rraustad: Got it.

rraustad:
I did a unit test today, but was only a small change to an existing unit test and deletion of CurDeadBandOrSetback from all but one unit test. I was also looking at why the IdealLoadsAirSystem even needs to run since the UnitarySystem model should be very close to the predicted load. I am OK with you merging this in now since it does correct a problem with UnitarySystem. If you do merge I'll just open another PR with other changes I find and what I have done today. No worries, I know your in a time crunch.

rraustad:
Re: the 30W spikes from the IdealLoadsAirSystem. I took the time of the first large spike, 3/4 12:00 and debugged to that point. I see the zone load in the deadband (not really part of the problem) and the UnitarySystem is off. Then the ERV simulates and causes a small remaining load, about 27.13 W. The next time around the UnitarySystem hits 27.13 W exactly with a mass flow rate of 0.0009xxx. This is a small flow rate < 0.001 (Real64 const SmallMassFlow(0.001)). So then the zone equipment is simulated. The ERV is modeled and then the air terminals, except the air terminal has a 0 inlet node mass flow rate (while the air loop outlet is 0.0009xxx) and provides 0 W to the zone. And thus the ideal loads system must turn on. So I suspect somewhere a SmallMassFlow check is turning off flow to the TUs because it's so small. I also suspect that there is node flow discontinuity which I will check tomorrow. So this is real and also not related to this issue. But it is a problem because if I disable the UnitarySystem using SmallMassFlow then the ideal loads system will still be needed (which is OK in my opinion). If I enable very low flows in the TUs then we get into very small mass flow rate issues. At least I know why these small loads still exist.

mjwitte:
@rraustad Since v9.2-RC3 has been tagged, this PR will wait until after release. So you can add the unit test here at your leisure.

rraustad:
Here's a snapshot of air system node flow data at time when ideal loads system turns on. Look at Row 1501, Node 4 (Supply side outlet node) is non-zero while the rest of the air loop has 0 flow and the ideal loads system is active. Apparently, ideal loads system is also on (row 1496-1498) when air loop is operating so convergence tolerance or sensible delivered capacity calculations differ by model and don't match what predictor needs.

```
AirLoopHVAC,
  central ac and furnace airloop,         !- Name
  Node 3,                                 !- Supply Side Inlet Node Name
  Node 6,                                 !- Demand Side Outlet Node Name
  Node 5,                                 !- Demand Side Inlet Node Names
  Node 4;                                 !- Supply Side Outlet Node Names

AirLoopHVAC:SupplyPath,
  central ac and furnace airloop Node 5 Supply Path, !- Name
  Node 5,                                 !- Supply Air Path Inlet Node Name
  AirLoopHVAC:ZoneSplitter,               !- Component Object Type 1
  central ac and furnace zone splitter;   !- Component Name 1

AirLoopHVAC:ZoneSplitter,
  central ac and furnace zone splitter,   !- Name
  Node 5,                                 !- Inlet Node Name
  Node 9;                                 !- Outlet Node Name 1

AirLoopHVAC:ReturnPath,
  central ac and furnace airloop Return Path, !- Name
  Node 6,                                 !- Return Air Path Outlet Node Name
  AirLoopHVAC:ReturnPlenum,               !- Component Object Type 1
  central ac and furnace airloop return plenum, !- Component Name 1
  AirLoopHVAC:ZoneMixer,                  !- Component Object Type 2
  central ac and furnace zone mixer;      !- Component Name 2

AirLoopHVAC:ReturnPlenum,
  central ac and furnace airloop return plenum, !- Name
  central ac and furnace airloop ret air zone, !- Zone Name
  Node 22,                                !- Zone Node Name
  Node 24,                                !- Outlet Node Name
  ,                                       !- Induced Air Outlet Node or NodeList Name
  Node 8;                                 !- Inlet Node Name 1

AirLoopHVAC:ZoneMixer,
  central ac and furnace zone mixer,      !- Name
  Node 6,                                 !- Outlet Node Name
  Node 24;                                !- Inlet Node Name 1
```

image

rraustad:
Correcting the discontinuity between the air loop UnitarySystem model and the remaining models (e.g., TUs, ReturnPlenum, etc.) of the air loop shows yet another discontinuity. 

In the top figure the same ideal loads system operation is shown (i.e., 30W spikes of ideal loads air system) along with the difference between the air loop mass flow rate of the supply side outlet node (Node 4) and the supply side inlet node (Node 3). Note the node flow difference lines up with the need for ideal loads air system to meet remaining zone load (i.e., the TUs were turned off as previously described).

After this was corrected (UnitarySystem is now off below Mdot < SmallMassFlow), the bottom figure shows the remaining node discontinuity. This now only happens during cooling mode (and ideal loads heating still occurs but is not shown in this figure). In addition, it happens when the air loop system model is operating at mass flow rates > SmallMassFlow. The air loop supply outlet node is the only node that differs from the other nodes constituting the remainder of the air loop as shown in the following table. This model uses AirTerminal:SingleDuct:ConstantVolume:NoReheat and suspected cause is some form of limiting of MassFlowRateMaxAvail in other equipment models (this may be harder to find).

image

image

rraustad:
The debugger shows the TU is limiting flow (in Init). This is after the air loop simulates, the final iteration before the next time step, and lines up with the data shown in the table above at 6/6 19:00. The simulation moves to the next time step which means flow convergence was reached so this may be normal.

image

rraustad:
- I question whether the AirTerminal:SingleDuct:ConstantVolume:NoReheat, the old direct air model, should check against maxavail or just use it's inlet node flow rate.
 - This discontinuity happens on the 3rd iteration, after the 2nd iteration the maxavail is set in ResolveAirLoopFlowLimits, and then on the 3rd iteration the TU gives the same air flow as before and the air loop converges. But the air loop requested a slightly higher flow rate. So how to allow the air loop flow to propagate to the TUs?

rraustad:
As a try to get the flow rates to match I had to do just one thing. Change the tolerance on the air loop convergence from 0.01 to 0.0001 kg/s. So since this was initially set at 20 CFM, I guess we need to decide if this is still a good limit for convergence. The simulation still executed very fast but this is a constant volume unitary system with cycling fan. I wonder how much impact a tighter flow tolerance would have on a VAV system? I added back in the ideal loads sensible heating energy to see the whole picture.

Compared to this same plot before this change, the ideal loads air system is still operating but at <30W, and since the flow rate of the unitary system is making it all the way to the zone (within 1E-16) then the need for the ideal loads air system (albeit only 30W) is due to some other issue. So I guess one question is whether the HVACFlowRateToler convergence limit should be changed or to just leave this as is.

    Real64 const HVACFlowRateToler(0.0001); // Tolerance for mass flow rate convergence (in kg/s) [~20 CFM]


image

rraustad:
OK, I blew that last evaluation. The ideal loads air system still comes on because the air loop is off when air flow rate < SmallMassFlow. So this is mostly resolved except for the question on HVACFlowRateToler.

rraustad:
Latest status since last commit.
 - CurDeacBandOrSetback has been removed from UnitarySystem since internal checks account for deadband
 - Ideal loads air system operates when UnitarySystem is off due to very small air mass flow rate
 - difference in UnitarySystem inlet/outlet node mass flow rate is due to HVACFlowRateToler(0.01) and the fact that the change in flow rate of final iteration was much less than this value

image

image

src/EnergyPlus/UnitarySystem.cc
rraustad: Done

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -6972,6 +6973,17 @@ namespace UnitarySystems {
...
+            CoolingLoad = false;
+            HeatingLoad = false;
+            MoistureLoad = 0.0;
+            this->m_CoolingPartLoadFrac = 0.0;
+            this->m_HeatingPartLoadFrac = 0.0;
+            if (this->CoolCoilFluidInletNode > 0) DataLoopNode::Node(this->CoolCoilFluidInletNode).MassFlowRate = 0.0;
+            if (this->HeatCoilFluidInletNode > 0) DataLoopNode::Node(this->HeatCoilFluidInletNode).MassFlowRate = 0.0;
+            this->setAverageAirFlow(this->m_CoolingPartLoadFrac, OnOffAirFlowRatio);
+            // anything else need to be reset here when system is shut down on low flow?
+        }
```
rraustad, line 7201: Shut down unit if air mass flow rate is below SmallMassFlow tolerance.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -9758,7 +9766,7 @@ namespace UnitarySystems {
                 int ATMixOutNode = this->ATMixerOutNode;
                 SensOutput =
                     DataLoopNode::Node(ATMixOutNode).MassFlowRate * (Psychrometrics::PsyHFnTdbW(DataLoopNode::Node(ATMixOutNode).Temp, MinHumRatio) -
-                                                                     Psychrometrics::PsyHFnTdbW(RefTemp, MinHumRatio));
+                                                                     Psychrometrics::PsyHFnTdbW(RefTemp, MinHumRatio) - this->m_SenLoadLoss);
```
rraustad, line 9973: noticed this when playing with SensOutput calculation (i.e., using new Bereket psyc function). Left psyc calc's as is to not muddy the water but noticed that this variable was missing.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -8735,6 +8742,7 @@ namespace UnitarySystems {
             m_initLoadBasedControlAirLoopPass = false;
         }
         if (!DataGlobals::BeginEnvrnFlag) {
+            this->m_MyEnvrnFlag2 = true; // this does not appear to be needed, only initializes autosized coil fluid flow rates
```
rraustad, line 8959: Just moved this from one location to another, no need to test !DataGlobals::BeginEnvrnFlag twice.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -203,14 +203,14 @@ namespace UnitarySystems {
...
-          NodeNumOfControlledZone(0), FanPartLoadRatio(0.0), CoolCoilWaterFlowRatio(0.0), HeatCoilWaterFlowRatio(0.0), ControlZoneNum(0),
-          AirInNode(0), AirOutNode(0), MaxCoolAirMassFlow(0.0), MaxHeatAirMassFlow(0.0), MaxNoCoolHeatAirMassFlow(0.0), DesignMinOutletTemp(0.0),
-          DesignMaxOutletTemp(0.0), LowSpeedCoolFanRatio(0.0), LowSpeedHeatFanRatio(0.0), MaxCoolCoilFluidFlow(0.0), MaxHeatCoilFluidFlow(0.0),
-          CoolCoilInletNodeNum(0), CoolCoilOutletNodeNum(0), CoolCoilFluidOutletNodeNum(0), CoolCoilLoopNum(0), CoolCoilLoopSide(0),
-          CoolCoilBranchNum(0), CoolCoilCompNum(0), CoolCoilFluidInletNode(0), HeatCoilLoopNum(0), HeatCoilLoopSide(0), HeatCoilBranchNum(0),
-          HeatCoilCompNum(0), HeatCoilFluidInletNode(0), HeatCoilFluidOutletNodeNum(0), HeatCoilInletNodeNum(0), HeatCoilOutletNodeNum(0),
-          ATMixerExists(false), ATMixerType(0), ATMixerOutNode(0), ControlZoneMassFlowFrac(0.0), m_CompPointerMSHP(nullptr)
+          m_SuppCompNotSetYet(true), m_OKToPrintSizing(false), m_SmallLoadTolerance(5.0), UnitarySystemType_Num(0), MaxIterIndex(0),
```
rraustad, line 208: initialize new small load tolerance to 5W

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -2512,6 +2512,7 @@ namespace UnitarySystems {
                         ShowContinueError(" The Fraction of Supply Air Flow That Goes Through the Controlling Zone is set to 1.");
                         this->ControlZoneMassFlowFrac = 1.0;
                     }
+                    this->m_SmallLoadTolerance = 5.0 / this->ControlZoneMassFlowFrac; // adjust 5W load tolerance by control zone fraction
```
rraustad, line 2643: adjust new tolerance by control zone fraction

src/EnergyPlus/UnitarySystem.cc
rraustad: @lgu1234 do these sensible capacity calculations need to include the SensLoadLoss variable?

lgu1234: @rraustad The SensLoadLoss is duct loss for a whole distribution system. If this system has a single terminal, it needs to add it. However, if there are more terminals or more zones are served, the SensLoadLoss should be added. However, the loss should be added weighted by mass flow rates.

nrel-bot-2b:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-2b:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

lgu1234:
I tested both defect files. There are no unmet load in the results from balanced-erv-object.idf. No extra Ideai Zone Heating energy was found. The highest heating energy during the same period is 47,000 J. The fix makes proper results.

lgu1234 (approved):

nealkruis:
@lgu1234 @rraustad @mjwitte What's the status of this PR? One of our fixes (#7827) depends on this fix.

mjwitte:
@nealkruis Looks like this needs one more review pass. In the queue.

Myoldmopar:
I am encouraged by reading through the long conversation here, how happy CI is, and how minimal the code changes are.

mjwitte:
@Myoldmopar @rraustad Not sure why this didn't get merged a while ago. Re-tested defect file for good measure. This all works as advertised. Ran unit tests serially. All pass. Not waiting for more CI platforms.

PR #7535, opened 2019-09-30
NFP for HAMT Enhancements Part 1

jasondegraw:
Pull request overview
---------------------
This is the NFP for the first of a set of enhancements to the HAMT model to add capabilities. See the document for further justification.

### Pull Request Author

### Reviewer

### DIFF
--- a/design/FY2019/NFP-HAMT-Enhancements-Pt1.md
+++ b/design/FY2019/NFP-HAMT-Enhancements-Pt1.md
@@ -1,0 +1,307 @@
+# NFP: HAMT Model Airflow Enhancements Part I #
+
+## Justification for New Feature ##
+The current HAMT model does not include air leakage in its calculation of the moisture deposition. Air movement as a moisture transport mechanism is typically far more important than vapor diffusion. Air leaks through walls, roofs, and floors can have the most damaging effect on the durability of a house. Uncontrolled airflow through the shell not only carries moisture into framing cavities, causing mold and rot, but it also can account for a huge portion of a home's energy use and can cause indoor air quality problems.
+
+## Overview ##
+The modification adds an internal moisture source in HAMT components due to air flow through the component. The moisture source is attached to a user-selected layer inside the component. Input options to determine the air flow are added, including a user-defined air flow, air flow determined from stack height, air permeance of the component and ventilation overpressure as well as are flow computed with the AirflowNetwork model.
+
+These modifications to EnergyPlus are part of a larger effort to enhance the moisture transport capabilities of EnergyPlus. This NFP describes the first two steps of this process:
+
+Step 1 - Enable air leakage to be applied as a source/sink at any arbitrary point inside the simulation model. Manual inputs of air flow rates is required.
+
+Step 2 - Allow air leakage rates to be controlled by air pressure gradients induced by wind forces, temperature differences and ventilation system.
+
+The necessary inputs will be developed to allow both user specified airflows and airflows calculated using the AirflowNetwork feature. The HAMT model calculations will be modified to account for this additional mode of transport.
+
+## Approach ##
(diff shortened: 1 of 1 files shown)


design/FY2019/NFP-HAMT-Enhancements-Pt1.md
```diff
@@ -0,0 +1,307 @@
...
+
+## I/O Reference ##
+The I/O Reference will be modified to describe the additional moisture source inputs. A moisture source object is attached to a particular layer in a manner similar to the way that heat sources are specified. Two options are being considered.
+
+### Modify Existing Object ###
+
+The first option modifies the existing internal sources object to include the needed moisture-related parameters. This is the preferred approach as it makes fewer changes to the IDD.
+
+```
+Construction:InternalSources,
```
mjwitte: Questions:
1. Is it feasible for HAMT to have more than one moisture source in a given construction?
2. Would more than one moisture source be useful?
3. Does HAMT currently support internal heat sources?
4. Could HAMT support both a moisture source and a heat source in the same construction? (I guess the proposed IDD object says Yes to this.)

design/FY2019/NFP-HAMT-Enhancements-Pt1.md
```diff
@@ -0,0 +1,307 @@
...
+       \type object-list
+       \object-list MaterialName
+```
+
+### Modify Existing Object, Add Two New Objects ###
+
+This option modifies the existing sources object to move the source information (both thermal and moisture) out into separate objects.
+
+```
+Construction:InternalHeatSource,
```
mjwitte: I lean toward this format, but suggest ConstructionProperty:InternalHeatSource, have it reference the construction name and material layer number and then the existing Construction:InternalSource transitions to plain old Construction plus the new ConstructionProperty:InternalHeatSource. 

OR if you don't want to touch Construction:InternalSource, how about two new object just for HAMT that are parallel to the existing HAMT objects:
MaterialProperty:HeatAndMoistureTransfer:HeatSource
MaterialProperty:HeatAndMoistureTransfer:MoistureSource

design/FY2019/NFP-HAMT-Enhancements-Pt1.md
jasondegraw: @mjwitte Thanks for the review, here are some answers:

1. I believe so, yes. I don't think there are any technical complications that would prevent it from being done.
2. Possibly, that's a point of active discussion right now. For situations where moisture accumulates near the surface of a material we might want to be able to have separate sources to account for which direction the air is going.
3. I think so, but I'll need to check to be sure.
4. Same as for number 3. The IDD changes were mainly driven by the possibility of multiple sources, but I'd like to be flexible in any case.

design/FY2019/NFP-HAMT-Enhancements-Pt1.md
jasondegraw: OK, I'll make some edits with your naming and go with the "separate objects" plan.

Myoldmopar:
@jasondegraw just to tie up this on my end, I don't have anything to add from what @mjwitte already commented.  This will be a nice addition.

jasondegraw:
@mjwitte @Myoldmopar So can we merge this then?

Myoldmopar:
It's more typical to just leave it open and have the code, idd, idf, tests, and docs all go into develop at the same time.  Is it anticipated that this NFP will build into the full implementation?

jasondegraw:
I'm not with you on it being typical, but in this case there's already a PR (or two) and a branch for the code. So I'd prefer to merge this PR and revive the (closed) PR for the code.

Myoldmopar:
Understood.  Merging this in.

PR #7536, opened 2019-10-01
Indoor Ice Rink codes

punyasloka95:
Indoor Ice Rink
---------------------

 -Initial Stage codes for the indoor ice rink


doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -520,6 +520,10 @@ \subsection{Surface(s)}\label{surfaces}
...
+\end{callout}
```
Myoldmopar, line 525: Why is this here?  This has nothing to do with ice rinks so I'm not sure why this is showing up.

scripts/dev/find_byref_bool_override.py
```diff
@@ -322,6 +322,11 @@
             "CannotShade"
         ]
     },
+    "SurfaceGeometry.cc": {
+        "CheckForReversedLayers": [
+                                    "RevLayerDiffs"
+                                    ]
+    },
```
Myoldmopar, line 329: This is also a bit surprising to see this added in your branch, but maybe it's necessary.

src/EnergyPlus/IndoorIceRink.hh
```diff
@@ -0,0 +1,299 @@
...
+    // MODULE VARIABLE DECLARATIONS:
+    // Standard, run-of-the-mill variables...
+    extern int NumOfDirectRefrigSys;
+    extern int NumOfIndirectRefrigSys;
+    extern int TotalNumRefrigSystem;
+    extern int NumOfResurfacer;
+    extern int OperatingMode;                    // Used to keep track of whether system is in heating or cooling mode
+    extern Array1D<Real64> QRadSysSrcAvg;        // Average source over the time step for a particular radiant surface
+    extern Array1D<Real64> ZeroSourceSumHATsurf; // Equal to SumHATsurf for all the walls in a zone with no source
+    extern Array1D_bool CheckEquipName;
```
Myoldmopar: This shouldn't need to be here.  The component should be created using the PlantComponent interface, not the old-style plant types.  We have several dozen examples in the plant code now of objects which use the new interface, and it makes them much sleeker and more easily testable than the old style types.  We can have a chat to discuss getting your code over to the new structure.

src/EnergyPlus/IndoorIceRink.hh
```diff
@@ -0,0 +1,299 @@
...
+        Real64 EHeatingWater;
+        Real64 QHumidity;
+        // Default Constructor
+        ResurfacerData() : CompIndex(0), GlycolIndex(0), ResurfacingSchedPtr(0), 
+            NoOfResurfEvents(0), ResurfacingWaterTemp(0.0), InitWaterTemp(0.0),
+            TankCapacity(0.0), QResurfacing(0.0), EHeatingWater(0.0), QHumidity(0.0), 
+            ResurfacingHeatLoad(0.0)
+        {
+        }
+    };
```
Myoldmopar, line 215: So there is no single struct for an ice rink?  That's unusual.  I'm not saying it has to be changed, but I don't exactly understand what the plant solver is going to do here.  The plant usually has a single index to a single typenum, but in this case, it almost looks like there is a possibility of two different structs being used, and so how would plant know to keep track of them.  I'll keep looking here, this is just suspicious.

src/EnergyPlus/IndoorIceRink.hh
```diff
@@ -0,0 +1,299 @@
...
+    void UpdateIndoorIceRink(bool const FirstHVACIteration, int const RadSysNum, int const SystemType);
+
+    void UpdateRinkRadSysSourceValAvg(bool &FloorRadSysOn);
+
+    void CheckForOutOfRangeTempResult(
+        int const SystemType, int const SysNum, Real64 const outletTemp, Real64 const inletTemp, Real64 const EP_UNUSED(mdot));
+
+    Real64 SumHATsurf(int const ZoneNum);
+
+    void ReportIndoorIceRink(int const SysNum, int const SystemType);
```
Myoldmopar: These functions look nice and split up, once this gets converted to the new plantcomponent it will look even better.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -13280,6 +13276,88 @@ namespace SurfaceGeometry {
         Surface(SurfNum).Width = WidthEff;
         Surface(SurfNum).Height = HeightEff;
     }
+    
+
+    void CheckForReversedLayers(bool &RevLayerDiffs,    // true when differences are discovered in interzone constructions
```
Myoldmopar, line 13281: Why did you have to add this?

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -4563,3 +4563,159 @@ TEST_F(EnergyPlusFixture, WorldCoord_with_RelativeRectSurfCoord_test4)
...
+    EXPECT_FALSE(RevLayerDiffs);
+
+    // Case 3a: Single layer constructs using Equivalent Glass, front/back properties NOT properly switched (should get a "true" answer)
+    Material(7).EmissThermalFront = 0.888;
+    RevLayerDiffs = false;
+    // ExpectResult = true;
+    CheckForReversedLayers(RevLayerDiffs, 5, 6, 1);
+    EXPECT_TRUE(RevLayerDiffs);
+
+}
```
Myoldmopar, line 4721: Unit tests look to be coming along nicely, that's great!  But it looks like your unit tests are only covering about 15% of the ice rink source file...that's really low and it would be great, and important, to get that unit test coverage up much higher.

http://energyplus.s3-website-us-east-1.amazonaws.com/coverage/2019-11-07-EnergyPlus-e44d834f73-PR7536-x86_64-gcc-7.4UTCD/EnergyPlus/IndoorIceRink.cc.gcov.html

Myoldmopar: And it doesn't look like you have an example file with the ice rink in it either, so that will be necessary.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
+// C++ Headers
```
Myoldmopar, line 48: You'll need a license header at the top of all your cc and hh files.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+
+    // Object Data
+    Array1D<DirectRefrigSysData> DRink;
+    Array1D<IndirectRefrigSysData> IRink;
+    Array1D<RefrigSysTypeData> RefrigSysTypes;
+    Array1D<ResurfacerData> Resurfacer;
+
+    // Functions:
+
+    void SimIndoorIceRink(std::string const &CompName,       // name of the floor refrigeration system
```
Myoldmopar: Once we get this converted over to the new plant component interface, this will be much cleaner, and I feel like you may want to have two different plant types, one for direct and one for indirect.  I'm not 100% sure of it right now, but that's what I feel like at this point.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+        // Using/Aliasing
+        using BranchNodeConnections::TestCompSet;
+        using DataGlobals::ScheduleAlwaysOn;
+        using DataHeatBalance::Zone;
+        using DataSurfaces::SurfaceClass_Floor;
+        using DataSurfaces::SurfaceClass_Window;
+        using FluidProperties::FindGlycol;
+        using FluidProperties::FindRefrigerant;
+        using NodeInputManager::GetOnlySingleNode;
+        using ScheduleManager::GetScheduleIndex;
```
Myoldmopar, line 173: Don't use `using` statements, just put the full namespace where you use the variable.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+        int MaxAlphas;                  // Maximum number of alphas for these input keywords
+        int MaxNumbers;                 // Maximum number of numbers for these input keywords
+        Array1D<Real64> Numbers;        // Numeric items for object
+        int NumAlphas;                  // Number of Alphas for each GetObjectItem call
+        int NumArgs;                    // Unused variable that is part of a subroutine call
+        int NumNumbers;                 // Number of Numbers for each GetObjectItem call
+
+        int SurfNum;                 // DO loop counter for surfaces
+        int SurfFloor;               // Surface which is a floor
+        int BaseNum;                 // Temporary number for creating RadiantSystemTypes structure
```
Myoldmopar: You don't have to define variables at the top of the function like in Fortran.  Define them right where they are used.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+            }
+
+            IRink(Item).RefrigConc = Numbers(12);
+            if ((IRink(Item).RefrigConc < 25.00) || (IRink(Item).RefrigConc > 30.0)) {
+                ShowWarningError(RoutineName + CurrentModuleObject + "=\"" + Alphas(1) +
+                                 " was entered with invalid refrigerant concentration. This is not allowed");
+                ShowContinueError("The refrigerant concentration has been reset to 25%.");
+                IRink(Item).RefrigType = 25.0;
+            }
+        }
```
Myoldmopar: Your error handling looks good here.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+                                _,
+                                _,
+                                "System");
+            SetupOutputVariable(
+                "Fluid Mass Flow Rate", OutputProcessor::Unit::kg_s, DRink(Item).RefrigMassFlow, "System", "Average", DRink(Item).Name);
+            SetupOutputVariable(
+                "Refrigerant Inlet Temperature", OutputProcessor::Unit::C, DRink(Item).RefrigInletTemp, "System", "Average", DRink(Item).Name);
+            SetupOutputVariable(
+                "Refrigerant Outlet Temperature", OutputProcessor::Unit::C, DRink(Item).RefrigOutletTemp, "System", "Average", DRink(Item).Name);
+        }
```
Myoldmopar: You will want to split out the setup output variable calls to a separate function, but that can happen later on, this looks fine for now.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+        static std::string const RoutineName("InitIndoorIceRink");
+
+        // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
+        int LoopCounter;
+        int SurfNum;
+        int SurfFloor;
+        static Array1D_bool MyEnvrnFlagDRink;
+        static Array1D_bool MyEnvrnFlagIRink;
+        static bool MyEnvrnFlagGeneral(true);
+        static bool ZoneEquipmentListChecked(false); // True after the Zone Equipment List has been checked for items
```
Myoldmopar: No statics.  And no local arrays used like this.  Just add a `bool myEnvrnFlag` to the actual direct or indirect struct, and access it like you do all the other variables on the struct.  Also, don't do anything in this init function that affects *all* the systems...only do initialize of the current system.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+                    }
+                }
+                MyPlantScanFlagIRink(SysNum) = false;
+            } else if (MyPlantScanFlagIRink(SysNum) && !AnyPlantInModel) {
+                MyPlantScanFlagIRink(SysNum) = false;
+            }
+        }
+
+        if (BeginEnvrnFlag && MyEnvrnFlagGeneral) {
+            ZeroSourceSumHATsurf = 0.0;
```
Myoldmopar: These should be defined for each object, there shouldn't be just global variables for SumHAT, QRad, etc.  They just need to be like all the other variables on the struct.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+                                           IRink(SysNum).RefrigBranchNum,
+                                           IRink(SysNum).RefrigCompNum);
+                    }
+                }
+                MyEnvrnFlagIRink(SysNum) = false;
+            }
+        }
+        if (!BeginEnvrnFlag && SystemType == IndirectSystem) MyEnvrnFlagIRink(SysNum) = true;
+
+        if (BeginTimeStepFlag && FirstHVACIteration) {
```
Myoldmopar: Begin time step looks suspicious here.  Did you know that the HVAC time step loop can step back and re-run the same iteration at a new time step length?  You probably need to check whether the current time has advanced to a new simulation time and only then step forward (and if it hasn't then you need to clear out the current results and start anew)

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+
+                    ShowSevereError("Refrigeration system entered without specification of type: Direct or Indirect?");
+                    ShowContinueError("Occurs in Refrigeration System=" + DRink(SysNum).Name);
+                    ShowFatalError("Preceding condition causes termination.");
+                }
+            }
+        }
+
+        {
+            auto const SELECT_CASE_var(SystemType);
```
Myoldmopar: I know these select case var blocks are present in the code, but there's no reason to include them in new code.  Just use a switch statement if possible, or at least a regular IF block.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+        } else if (SystemType == IndirectSystem) {
+            SetPointTemp = IRink(SysNum).IceSetptTemp;
+            FloodWaterTemp = IRink(SysNum).FloodWaterTemp;
+            RhoWater = GetDensityGlycol(fluidNameWater, FloodWaterTemp, IRink(SysNum).WaterIndex, RoutineName);
+            CpWater = GetSpecificHeatGlycol(fluidNameWater, FloodWaterTemp, IRink(SysNum).WaterIndex, RoutineName);
+            Length = IRink(SysNum).LengthRink;
+            Width = IRink(SysNum).WidthRink;
+            Depth = IRink(SysNum).IceThickness;
+            Volume = Length * Width * Depth;
+        }
```
Myoldmopar: These two blocks look *really* similar.  I think that splitting up the direct and indirect is leading to duplicated code and excessive effort on your part.  This should be considered moving forward because we don't want to end up with code that is difficult to maintain.

src/EnergyPlus/IndoorIceRink.cc
```diff
@@ -0,0 +1,2797 @@
...
+
+        QFreezing = RhoWater * Volume * ((CpWater * FloodWaterTemp) + (QFusion) - (CpIce * SetPointTemp
```
Myoldmopar: This is a surprising amount of data in here.  I would've preferred using std::vector instead of Array1D, but I don't think we need to make that change right now.

Myoldmopar, line 616: With this many potential code paths, this function will need to be tested...a lot.

Myoldmopar: Do you need to clean up anything else here before leaving?  Maybe you do below.

Myoldmopar, line 784: This looks _a lot_ like the function above it.  Again, I'm concerned about duplicated code here.

Myoldmopar: Yeah, duplicated blocks....

Myoldmopar (commented):
This still needs lots of work.  I would say the primary things that need to be changed are:
 - We need to discuss the design of the structs.  I think having two separate structs that are not connected at all is causing a problem.  You could create a single base class that has all the common data and then inherit that base class for direct and indirect classes.  Then you could share all the code between the direct and indirect, except for what is specifically different about them.
 - We need to get this connected to plant.  It looks like right now you aren't connected at all.  And before you attempt it, we need to get your class inheriting PlantComponent and operating in the new style.  This relies on setting up the classes that I mentioned in the prior bullet.

nrel-bot-2c:
@punyasloka95 @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@punyasloka95 @lgentile it has been 28 days since this pull request was last updated.

mjwitte:
@punyasloka06 Is this targeted for 9.3.0 release (Feb 15 I/O freeze date, March 31 release) or later?

nrel-bot-2:
@punyasloka95 it has been 27 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 8 days since this pull request was last updated.

nrel-bot-2:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@punyasloka95 it has been 8 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 9 days since this pull request was last updated.

nrel-bot-2:
@punyasloka95 it has been 8 days since this pull request was last updated.

mitchute:
Can we close this PR, or at least convert it to a draft? @Myoldmopar @mjwitte @punyasloka95

mjwitte:
@mitchute Converted to draft. Haven't seen any new code since November.

nrel-bot-2c:
@punyasloka95 it has been 17 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 7 days since this pull request was last updated.

nrel-bot-2c:
@punyasloka95 it has been 8 days since this pull request was last updated.

nrel-bot-2b:
@punyasloka95 it has been 9 days since this pull request was last updated.

nrel-bot-2:
@punyasloka95 it has been 13 days since this pull request was last updated.

mitchute:
No updates here for 9 months. I'll close this but feel free to reopen when ready to pick it back up. The PR is from a fork, so there won't be any branches to preserve on this side.

PR #7543, opened 2019-10-03
Trim down #include paths for EnergyPlus

Myoldmopar:
Pull request overview
---------------------
 The #include declarations in EnergyPlus are a complete mixed bag.  Sometimes quotes, sometimes brackets.  And then, because we have so many INCLUDE paths in CMake, we have varying bracket include types.  Sometimes it has <EnergyPlus/ and sometimes it is just the filename.

This cleans up the EnergyPlus-specific includes, not all the third party code.

Should be a clean build and no diffs.  If not then something is wrong or out of date.


tst/EnergyPlus/unit/OutdoorAirUnit.unit.cc
```diff
@@ -52,28 +52,28 @@
...
-#include <OutputReportPredefined.hh>
-#include <Psychrometrics.hh>
-#include <ScheduleManager.hh>
+#include <EnergyPlus/Fans.hh>
+#include <EnergyPlus/FluidProperties.hh>
+#include <EnergyPlus/General.hh>
+#include <EnergyPlus/HeatBalanceManager.hh>
+#include <EnergyPlus/OutputReportPredefined.hh>
+#include <EnergyPlus/Psychrometrics.hh>
+#include <EnergyPlus/ScheduleManager.hh>
```
Myoldmopar, line 76: In this block alone, three different include options are used.  The new pattern is simple, to import EnergyPlus header files, just `#include <EnergyPlus/HeaderFile.hh>`

Myoldmopar:
This is really an either it's broken or it worked type of PR.  It just eliminates one extraneous include directory.  I'm going to merge this in now so it can build overnight and not add runs to CI during the workweek.  Any PRs that have new/modified include statements will need to resolve conflicts and make sure they start with <EnergyPlus/Foo.hh>, instead of just <Foo.hh>.

PR #7544, opened 2019-10-03
Require Python 3

Myoldmopar:
Pull request overview
---------------------
Moves us to Python 3.6 or higher.  All developers will need to ensure this is available on their machines.  Initially, I expect the Windows CI box to fail, at which point I will install Python 3 and it should pass after a re-run.

Developers will likely need to either clear their CMake cache, or start a new build folder, as the old PYTHON_EXECUTABLE path will persist and since it will be Python 2, it will fail.


Myoldmopar:
OK, installed Python3 on Windows, then installed the required dependencies on Linux and Windows.  I set CI to rerun those and am hopeful those will be clean now.  I need to get Mac Python3 updated tomorrow.

Myoldmopar:
OK, Mac has Python 3 installed along with the dependencies.  I had a little trouble with explicitly requiring at least 3.6 even on my own machine, so I opened it up to just any Python 3.  I'll wait and see what the latest CI results are before attempting any other changes.

nrel-bot-3:

nrel-bot-2:

Myoldmopar:
Finally got a chance to check on this, and I waited just long enough.  Looks like the BeautifulSoup package had this [defect](https://bugs.launchpad.net/beautifulsoup/+bug/1838903) that was **just** fixed late 2019.  They did a release of 4.8.2 a couple weeks ago, so I'm going to log in and `pip install --upgrade` and rerun this.

jasondegraw:
@Myoldmopar When I was doing the license update, I ran on my Windows machine with Python 3.7 and got some strange behavior out of file finding code. It's finding way more files than it should, it looks like the 2.7 way of recursively finding files goes a little haywire in 3. I just verified that it does the same thing on 3.6 (Ubuntu 18.04). Do you want fixes here, or in a separate PR?

Myoldmopar:
@jasondegraw here is a suitable place.  ~~I'm impressed you're still having fun with the Lucid Lynx.~~

jasondegraw:
I'm not sure that's even possible (typo corrected). In any case, I'll get some changes together to clean things up and push them up here.

jasondegraw:
@Myoldmopar OK, that's done. I made a lot of edits to clean things up, but the "real" changes that were needed were (1) change how the files were being found to use glob and (2) use regular `open` instead of the codecs version.

Myoldmopar:
Great, thanks man, I'm continuing to work through tablediff encoding issues but hopefully this will be wrapped up tonight.

Myoldmopar:
@jasondegraw OK, I merged in a change to the regression scripts which will hopefully fix the unicode issue -- of course that may simply lead to the next issue to fix.  Anyway, I triggered a new build of the Linux release configuration so we'll see soon.

There was a failure on the custom_check for license_check, complaining that the output is not JSON consumable line-by-line output messages from the script.  Any chance there is something new coming out of the script?

jasondegraw:
@Myoldmopar Yes, I could have messed that up. Let me give it a try locally to make sure I got the output right.

jasondegraw:
@Myoldmopar I had previously only tested on Windows and verified visually that the output was OK. I redid some checks on Ubuntu 18.04, and it all seems OK. When it finds something, Python does accept the output as legal JSON (on a per line basis), and when it doesn't find anything it writes out nothing. I'm using `json.dumps` to format things, see [here](https://github.com/NREL/EnergyPlus/blob/b7bb47e194474cd37f32423a1508eea133077058/scripts/dev/licensetext.py#L243). But since the year change is in this branch I'm not sure what is going on. It shouldn't be finding anything... can you try running the script locally on the CI VM to see what happens?

Myoldmopar:
@jasondegraw I _think_ I found the problem.  I'm guessing when you run this locally, you execute: `python3 ./scripts/dev/license-check.py` or something equivalent.  When CI runs it, it is executed by bash as just `./scripts/dev/license-check.py`, so the first thing bash does is check the shebang line and sees that it needs to be executed by whatever `python` is available from `env`.  On the Ubuntu boxes, the system python defaults to Python 2, so the script gives a Python error due to the new call to `super()` without an argument.  I changed the shebang line to search for `/usr/bin/env python3` and it works on my machine, hopefully it will also work find on the CI box.  We'll see soon.

Myoldmopar:
> MoveToPython3 (Myoldmopar) - x86_64-Linux-Ubuntu-18.04-custom_check: OK (11 of 11 tests passed, 0 test warnings)
> 
> [Build Badge](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-cfc07e58f8140432a039a1af37adfe45f6e30a10-x86_64-Linux-Ubuntu-18.04-custom_check.html)

Hooray!!

jasondegraw:
@Myoldmopar Good catch, I was indeed running as you suggested. Are we going to leave Python 2 on the VMs after this changeover?

Myoldmopar:
Python2.7 is a core dependency on Ubuntu, so we can't remove that.  We could potentially remove it from Windows and Mac, but I doubt I'll do that right away.  The Windows VM is actually set for a fresh image soon, so I will probably just _not_ install Python 2 after that.

Myoldmopar:
We are all green!  This is merging in, and EnergyPlus is officially requiring Python 3.

jasondegraw:
@Myoldmopar It may be a core dependency, but it's not in my path on my new VMs, that's all I'm saying.

Myoldmopar:
> @Myoldmopar It may be a core dependency, but it's not in my path on my new VMs, that's all I'm saying.

Hmm.  So are you saying you have an Ubuntu 18.04 image, and if you run `which python2`, it doesn't find one?

jasondegraw:
That's what I'm saying. Both `python` and `python2` come up empty. If I try to run them I get the apt message about how to get them.

mjwitte:
@Myoldmopar @jasondegraw On my Win10 machine, I have three versions of Python installed, 2.7, 3.4, and 3.7. I can't build now:
```
CMake is re-running because C:/MJW Git/EnergyPlus/Build/CMakeFiles/generate.stamp is out-of-date.
1>  the file 'C:/MJW Git/EnergyPlus/CMakeLists.txt'
1>  is newer than 'C:/MJW Git/EnergyPlus/Build/CMakeFiles/generate.stamp.depend'
1>  result='-1'
1>CMake Error at C:/Program Files/CMake/share/cmake-3.15/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
1>  Could NOT find PythonInterp: Found unsuitable version "2.7.15", but
1>  required is at least "3" (found C:/Python27/python.exe)
```
Does this require uninstalling older Pythons? Should cmake be smarter about this? Or?

mbadams5:
@mjwitte You might need to delete your CMake cache and run configuration again (from within CMake). It has probably cached the python2 path and thus isn't looking for python3 but still requires it.

mjwitte:
@mbadams5 Yes, I found that's necessary but not sufficient. Also had to remove any Python2.x lurking in the system path.

mbadams5:
@mjwitte Interesting. I thought CMake was smart enough to figure it out without removing python2 from the path.

mjwitte:
Not 100% certain that I always cleared the cache as I was messing with the path, but I think I did.

Myoldmopar:
> @mbadams5 Yes, I found that's necessary but not sufficient. Also had to remove any Python2.x lurking in the system path.

@mjwitte So was Python 3 available on PATH too?

> That's what I'm saying. Both python and python2 come up empty. If I try to run them I get the apt message about how to get them.

@jasondegraw what version of Ubuntu is installed?  Is it a minimal image or something?  The Ubuntu desktop [manifest](http://old-releases.ubuntu.com/releases/bionic/ubuntu-18.04-desktop-amd64.manifest) lists both python2.7 and python3 dependencies, so if you try to apt remote it, it _should_ also try to remove the Ubuntu meta-package itself.  Unless I'm missing something.

mjwitte:
@Myoldmopar At one point, I had both Python 3 and 2 in the system path. I don't recall if 3 was in the path when I started all of this. And I'm not sure if I deleted the cache every single time I changed the path (which requires a restart to reset the path variable). I'll go back and add python 2 into the path along with 3, restart, delete cache and see what happens.

mjwitte:
@Myoldmopar @mbadams5 @jasondegraw After some experimenting, the conclusion is that cmake goes with whatever Python it finds first in the path list. I've got 2.7, 3.4, and 3.7 on this machine. If 3.4 or 3.7 comes first in the path, then cmake runs with that. If 2.7 is first, then it errors out.

PR #7566, opened 2019-10-17
MicroTurbine Generator Plant Comp Refactor

mitchute:
Pull request overview
---------------------

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/ChillerExhaustAbsorption.cc
+++ b/src/EnergyPlus/ChillerExhaustAbsorption.cc
@@ -124,7 +124,6 @@ namespace ChillerExhaustAbsorption {
     using DataHVACGlobals::SmallWaterVolFlow;
     using General::RoundSigDigits;
     using General::TrimSigDigits;
-    using MicroturbineElectricGenerator::SimMTGenerator;
     using Psychrometrics::CPCW;
     using Psychrometrics::CPHW;
     using Psychrometrics::RhoH2O;
@@ -329,7 +328,6 @@ namespace ChillerExhaustAbsorption {
         using CurveManager::GetCurveCheck;
         using DataSizing::AutoSize;
         using GlobalNames::VerifyUniqueChillerName;
-        using MicroturbineElectricGenerator::GetMTGeneratorExhaustNode;
         using NodeInputManager::GetOnlySingleNode;
         using OutAirNodeManager::CheckAndAddAirNodeNumber;
 
@@ -338,7 +336,6 @@ namespace ChillerExhaustAbsorption {
         int NumAlphas;        // Number of elements in the alpha array
         int NumNums;          // Number of elements in the numeric array
         int IOStat;           // IO Status when calling get input subroutine
-        int MTExhaustNodeNum; // Exhaust node number passed from MicroTurbine
         std::string ChillerName;
         bool Okay;
 
@@ -520,9 +517,8 @@ namespace ChillerExhaustAbsorption {
                 ExhaustAbsorber(AbsorberNum).CompType_Num = iGeneratorMicroturbine;
                 ExhaustAbsorber(AbsorberNum).ExhuastSourceName = cAlphaArgs(18);
 
-                GetMTGeneratorExhaustNode(
-                    ExhaustAbsorber(AbsorberNum).CompType_Num, ExhaustAbsorber(AbsorberNum).ExhuastSourceName, MTExhaustNodeNum);
-                ExhaustAbsorber(AbsorberNum).ExhaustAirInletNodeNum = MTExhaustNodeNum;
+                auto thisMTG = MicroturbineElectricGenerator::MTGeneratorSpecs::factory(ExhaustAbsorber(AbsorberNum).ExhuastSourceName);
+                ExhaustAbsorber(AbsorberNum).ExhaustAirInletNodeNum =dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->CombustionAirOutletNodeNum;
             }
         }
 
@@ -1515,7 +1511,6 @@ namespace ChillerExhaustAbsorption {
         using DataPlant::SingleSetPoint;
         using FluidProperties::GetDensityGlycol;
         using FluidProperties::GetSpecificHeatGlycol;
-        using MicroturbineElectricGenerator::SimMTGenerator;
         using PlantUtilities::SetComponentFlowRate;
         using Psychrometrics::PsyCpAirFnWTdb;
 
@@ -2004,7 +1999,6 @@ namespace ChillerExhaustAbsorption {
         using DataPlant::SingleSetPoint;
         using FluidProperties::GetDensityGlycol;
         using FluidProperties::GetSpecificHeatGlycol;
-        using MicroturbineElectricGenerator::SimMTGenerator;
         using PlantUtilities::SetComponentFlowRate;
         using Psychrometrics::PsyCpAirFnWTdb;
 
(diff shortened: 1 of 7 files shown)


mitchute:
@Myoldmopar Ready for review.

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -520,9 +517,8 @@ namespace ChillerExhaustAbsorption {
                 ExhaustAbsorber(AbsorberNum).CompType_Num = iGeneratorMicroturbine;
                 ExhaustAbsorber(AbsorberNum).ExhuastSourceName = cAlphaArgs(18);
 
-                GetMTGeneratorExhaustNode(
-                    ExhaustAbsorber(AbsorberNum).CompType_Num, ExhaustAbsorber(AbsorberNum).ExhuastSourceName, MTExhaustNodeNum);
-                ExhaustAbsorber(AbsorberNum).ExhaustAirInletNodeNum = MTExhaustNodeNum;
+                auto thisMTG = MicroturbineElectricGenerator::MTGeneratorSpecs::factory(ExhaustAbsorber(AbsorberNum).ExhuastSourceName);
+                ExhaustAbsorber(AbsorberNum).ExhaustAirInletNodeNum =dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->CombustionAirOutletNodeNum;
```
Myoldmopar, line 521: This line made me think about our factories.  Right now our factories return a pointer to a plant component.  That means that as a part of returning from the factory, the code adapts the component struct pointer to a plant component pointer.  It may not actually do that step, but just making a point here.  If our factories instead returned component struct pointers, then when we call the factory from plant, we would still cast it to a plant component pointer for storage.  But we could then call the factory and get a living reference to a direct struct instance.  Then we wouldn't have to do this casting you are doing here.  So my point would be potentially changing:

```
static PlantComponent *factory(std::string const &objectName);
```
to

```
static BoilerSpecs *factory(std::string const &objectName);
```

It seems to compile happily on my machine.  Then this section would just be:

```
auto thisMTG = MicroturbineElectricGenerator::MTGeneratorSpecs::factory(ExhaustAbsorber(AbsorberNum).ExhuastSourceName);
ExhaustAbsorber(AbsorberNum).ExhaustAirInletNodeNum =thisMTG->CombustionAirOutletNodeNum;
```

What do you think?

src/EnergyPlus/ElectricPowerServiceManager.cc
```diff
@@ -2156,12 +2156,18 @@ void GeneratorController::simGeneratorGetPowerOutput(bool const runFlag,
...
+        // dummy vars
+        PlantLocation L(0,0,0,0);
+        Real64 tempLoad = myElecLoadRequest;
+
+        // simulate
+        dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->InitMTGenerators(runFlag, tempLoad, FirstHVACIteration);
+        dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->CalcMTGeneratorModel(runFlag, tempLoad);
+        dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->UpdateMTGeneratorRecords();
+        electricPowerOutput = dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->ElecPowerGenerated;
+        thermalPowerOutput = dynamic_cast<MicroturbineElectricGenerator::MTGeneratorSpecs*> (thisMTG)->QHeatRecovered;
```
Myoldmopar, line 2170: And these lines could be cleaned up too by the factory change, I'm feeling like that's a good option now as I continue reviewing...

tst/EnergyPlus/unit/ChillerExhaustAbsorption.unit.cc
```diff
@@ -60,7 +60,9 @@
 using namespace EnergyPlus;
 using namespace EnergyPlus::ChillerExhaustAbsorption;
 
-TEST_F(EnergyPlusFixture, ExhAbsorption_GetInput_Test)
+class ExhAbsorptionFixture : public EnergyPlusFixture {};
+
+TEST_F(ExhAbsorptionFixture, ExhAbsorption_GetInput_Test)
```
Myoldmopar, line 65: Not quite sure why this was necessary, but no problem :+1:

src/EnergyPlus/MicroturbineElectricGenerator.cc
```diff
@@ -996,193 +842,137 @@ namespace MicroturbineElectricGenerator {
...
-    //******************************************************************************
+    void MTGeneratorSpecs::simulate(const PlantLocation &EP_UNUSED(calledFromLocation),
+                                    bool EP_UNUSED(FirstHVACIteration),
+                                    Real64 &EP_UNUSED(CurLoad),
+                                    bool EP_UNUSED(RunFlag))
+    {
+        // empty function to emulate current behavior as of conversion to using the PlantComponent calling structure.
+        // calls from the plant side... do nothing.
+        // calls from the ElectricPowerServiceManger call the init, calc, and update worker functions
+    }
```
Myoldmopar, line 964: This is a key note.  By changing the factory to return a pointer to the struct instead of a pointer to a plant component, it allows other managers to also get a reference to the same object.  Later, this component struct could also inherit some other base class, like ElectricPowerServiceComponent.  This new base class may have some function like: `eps_simulate(...)`, and the same plant approach can be used over there.  In this particular case, I do wonder if we even _need_ to inherit plant component, but we can leave it for now.

src/EnergyPlus/MicroturbineElectricGenerator.cc
```diff
@@ -996,193 +842,137 @@ namespace MicroturbineElectricGenerator {
...
+        MaxLoad = 0.0;
+        MinLoad = 0.0;
+        OptLoad = 0.0;
+    }
 
-    void InitMTGenerators(int const GenNum,
-                          bool const RunFlag,
-                          Real64 const MyLoad, // electrical load in W
-                          bool const FirstHVACIteration)
+    void MTGeneratorSpecs::InitMTGenerators(bool const RunFlag,
```
Myoldmopar, line 973: Just an editorial note, feel free to rename these functions to simple names like `initialize()` and `calculate()` as you wrap up the final touches.  Not necessary for this one.

src/EnergyPlus/MicroturbineElectricGenerator.cc
```diff
@@ -2106,163 +1812,36 @@ namespace MicroturbineElectricGenerator {
...
+        if (this->FuelEnergyUseRateLHV > 0.0) {
+            this->ElectricEfficiencyLHV = this->ElecPowerGenerated / this->FuelEnergyUseRateLHV;
+            this->ThermalEfficiencyLHV = this->QHeatRecovered / this->FuelEnergyUseRateLHV;
         } else {
-            ExhaustOutletNodeNum = MTGenerator(CompNum).CombustionAirOutletNodeNum;
+            this->ElectricEfficiencyLHV = 0.0;
+            this->ThermalEfficiencyLHV = 0.0;
         }
+        this->AncillaryEnergy = this->AncillaryPowerRate * DataHVACGlobals::TimeStepSys * DataGlobals::SecInHour;
+        this->StandbyEnergy = this->StandbyPowerRate * DataHVACGlobals::TimeStepSys * DataGlobals::SecInHour;
```
Myoldmopar, line 1842: This entire file -- so much better.  Thanks!

Myoldmopar (commented):
This is acceptable as-is, but I'd like to chat regarding my factory notes below before we proceed.

Myoldmopar:
No CI warnings or failures related to this, code changes look good.  I'll still pull develop in to make sure nothing breaks after the merge.

Myoldmopar:
Nevermind, this branch is only forward of develop.  All good.  Let's just drop this in for now and we can still discuss the factory method for later refactors.

PR #7584, opened 2019-10-28
Modify End Use by Subcategory tables in SQL so they can be queried

### LINKED ISSUE #7481: Difficult to query End Uses By Subcategory tables in SQLite
It's extremely difficult to query data from the End Uses By Subcategory tables using the SQLite file. This is because you can only query based on RowName (first value in row) and ColumnName; however RowName can be blank for many rows in these tables.

For example, imagine trying to obtain the fridge electricity use from the `TabularDataWithStrings` SQLite table for results like these:

  | Subcategory | Electricity [GJ] | Natural Gas [GJ] | Additional Fuel [GJ] | District Cooling [GJ] | District Heating [GJ] | Water [m3]
-- | -- | -- | -- | -- | -- | -- | --
Heating | General | 0.00 | 26.03 | 0.00 | 0.00 | 0.00 | 0.00
Cooling | General | 2.29 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00
Interior Equipment | misc plug loads | 8.84 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00
  | fridge | 2.34 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00
Water Systems | clothes washer | 0.00 | 0.00 | 0.00 | 0.00 | 0.00 | 5.22
  | General | 4.88 | 8.95 | 0.00 | 0.00 | 0.00 | 0.00
  | Other | 1.33 | 0.00 | 0.00 | 0.00 | 0.00 | 0.00

It essentially can't be done. Thus, you have to resort to querying results from the `TabularData` table in the SQLite file, which is far harder to deal with.

One possible solution is to combine the first two columns in the table. So you end up with RowName values that are, e.g., "Heating: General", "Cooling: General", "Interior Equipment: misc plug loads",  "Interior Equipment: fridge", etc.

Another possible solution is to add a "SubRowName" field to the `TabularDataWithStrings` table. For most tables the value would be NULL, but the End Uses By Subcategory tables could take advantage of it.

jmarrec:
Pull request overview
---------------------
 - Fixes #7481
 - There are two tables in TabularDataWithStrings that show end uses (either energy or demand) by subcategories. They are currently impossible to properly query using SQL

This Pull Request will properly fill the rowHead (forward fill the values of End Uses basically) before ouputing to SQL but after the HTML file is already written so we don't mess with this.

-----

**Before**:

If you look at the HTML result, you see that rowHead is empty for display purposes:

image

This means the SQL one looks very similar, with empty "RowName" cells:

image

---- 

**After**: All RowNames are filled

image

It's still not terribly straightforward to get a specific entry for a subcategory end use, but it's doable for sure (and quite fast). I personally like this better than combining both End Use and End Use Subcat in the same RowName, or worse adding a new column (which would create way too many "NULLs").

image

```
SELECT * FROM TabularDataWithStrings
  WHERE TableName = "End Uses By Subcategory"
  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
  AND RowName = "Interior Lighting"
  AND (TabularDataIndex - (SELECT TabularDataIndex FROM TabularDataWithStrings
							  WHERE TableName = "End Uses By Subcategory"
							  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
							  AND ColumnName = "Subcategory"
							  AND RowName = "Interior Lighting"
							  AND Value = "GeneralLights"))
	  % (SELECT COUNT(Value) FROM TabularDataWithStrings
							  WHERE TableName = "End Uses By Subcategory"
							  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
							  AND ColumnName = "Subcategory") = 0
```

### Pull Request Author
 - [x] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [x] If structural output changes, add to output rules file and add OutputChange label

### Reviewer


tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -7653,3 +7654,206 @@ TEST_F(SQLiteFixture, WriteSourceEnergyEndUseSummary_TestPerArea) {
...
+                          "  AND ReportName = '" + reportName + "'"
+                          "  AND Value = '" + endUseSubCategoryName + "'");
+
+        auto result = queryResult(query, "TabularDataWithStrings");
+
+        ASSERT_EQ(1ul, result.size()) << "Query crashed for reportName=" << reportName;
+
+        // Add informative message if failed
+        EXPECT_EQ(endUseName, result[0][0]) << "Failed for reportName=" << reportName;
+    }
```
jmarrec, line 7784: Check that the "RowName" isn't empty anymore

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -7653,3 +7654,206 @@ TEST_F(SQLiteFixture, WriteSourceEnergyEndUseSummary_TestPerArea) {
...
+                "                              AND ReportName = 'AnnualBuildingUtilityPerformanceSummary'"
+                "                              AND ColumnName = 'Subcategory'"
+                "                              AND RowName = '" + endUseName + "'"
+                "                              AND Value = '" + endUseSubCategoryName + "'))"
+                "       % (SELECT COUNT(Value) FROM TabularDataWithStrings"
+                "            WHERE TableName = 'End Uses By Subcategory'"
+                "            AND ReportName = 'AnnualBuildingUtilityPerformanceSummary'"
+                "            AND ColumnName = 'Subcategory') = 0",
+                "TabularDataWithStrings");
+        ASSERT_EQ(7u, result.size());
```
jmarrec: This query works in one step to query all entries for a specific end use subcategory.

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -7653,3 +7654,206 @@ TEST_F(SQLiteFixture, WriteSourceEnergyEndUseSummary_TestPerArea) {
...
+
+        ASSERT_EQ(7u, result.size()) << "Failed for query: " << query;
+
+        // Now specifically get the electricity one, and make sure it's the right number that's returned
+        query = (
+            "SELECT Value FROM TabularDataWithStrings"
+            "  WHERE TableName = 'End Uses By Subcategory'"
+            "  AND ReportName = 'AnnualBuildingUtilityPerformanceSummary'"
+            "  AND RowName = '" + endUseName + "'"
+            "  AND ColumnName = 'Electricity'"
```
jmarrec: This is the addition here, query for ONE fuel. And Return only "Value" not "*"

shorowit:
@jmarrec Thank you! I'll be able to get rid of my super complicated, hacky code...

nrel-bot-3:

Myoldmopar:
@shorowit have you tested this to confirm it fixes your issue?  @kbenne I'd normally contact DMac with things like this to give the OS team a heads up.  Can you advise who should be notified now?

shorowit:
I have not tested it, only looked at @jmarrec's screenshots.

That said, now that I get fresh eyes on this... the PR is an incremental improvement that makes it slightly easier to query, but the query is still pretty ugly. It's fine to have this drop in, but I don't feel that the issue has been substantially addressed. The query to obtain this kind of information should not have to be so complicated.

Compare this PR's query:
```
SELECT * FROM TabularDataWithStrings
  WHERE TableName = "End Uses By Subcategory"
  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
  AND RowName = "Interior Lighting"
  AND (TabularDataIndex - (SELECT TabularDataIndex FROM TabularDataWithStrings
                              WHERE TableName = "End Uses By Subcategory"
                              AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
                              AND ColumnName = "Subcategory"
                              AND RowName = "Interior Lighting"
                              AND Value = "GeneralLights"))
      % (SELECT COUNT(Value) FROM TabularDataWithStrings
                              WHERE TableName = "End Uses By Subcategory"
                              AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
                              AND ColumnName = "Subcategory")
```
with the two queries I was envisioning as possibilities:
```
SELECT * FROM TabularDataWithStrings
  WHERE TableName = "End Uses By Subcategory"
  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
  AND RowName = "Interior Lighting"
  AND SubRowName = "GeneralLights"
```
or
```
SELECT * FROM TabularDataWithStrings
  WHERE TableName = "End Uses By Subcategory"
  AND ReportName = "AnnualBuildingUtilityPerformanceSummary"
  AND RowName = "Interior Lighting:GeneralLights"
```

jmarrec:
Extracting the relevant comment from my original PR message:

> It's still not terribly straightforward to get a specific entry for a subcategory end use, but it's doable for sure (and quite fast). I personally like this better than combining both End Use and End Use Subcat in the same RowName, or worse adding a new column (which would create way too many "NULLs").

The `TabularDataWithStrings` is a SQL heresy (note that I do understand why it is like that): it's a View that's aiming to be generic enough to store all kinds of data, and bends the rules of database design. This kind of problem wouldn't occur if this data was in its own SQL TABLE.

mjwitte:
@jmarrec Would it be useful to duplicate the end-use information in a second well-designed table? Not necessarily as part of this PR.

shorowit:
Personally, I don’t see why adding a new End Use SubCat column that is often null is so bad. It seems like the most generic and least obtrusive change. And it would be non-breaking. 

Adding a separate table would be fine, but seems silly to query end use data one place and end use by subcategory data somewhere else. 

My two cents.

jmarrec:
I don't like adding 10,000 NULLS for what's really a corner case.

Ok, would this make you happier?

image


I deleted the "ColumnName==Subcategory" (should I leave it?), then for each RowName is "EndUse:Subcategory"

That'd make your query work:

image

shorowit:
Not sure how others would feel, but I would be happy with this solution, yes.

jmarrec:
@shorowit see updated solution. check the markdown, I included example queries and table outputs too

shorowit:
I built this locally and confirmed that the results look correct and the data is now easy to query. @jmarrec, thanks so much for making this change.

Myoldmopar:
This is really great.  Verified functional improvement from @shorowit, unit test coverage, and documented output changes in the markdown.  Thanks @jmarrec; merging this in.

PR #7603, opened 2019-11-07
Input file conversion utility IDF->epJSON and reverse

### LINKED ISSUE #6835: No way to just convert IDF to JSON without running simulation
Issue overview
--------------
I would like to use `energyplus -c my.idf` to convert an IDF to JSON.  However, when I do this, EnergyPlus also attempts to run a simulation.  This takes additional time and generates many unwanted files.  Can we add a way to only do the conversion and then exit?

### Details
 - Version of EnergyPlus - 8.9

### Checklist

mbadams5:
Pull request overview
---------------------
Fixes #6835

This pull request adds a flag to only convert the input file (`--convert-only`) and not perform a simulation as well as adds an external input file conversion utility program.

The CLI `--convert-only` flag will now exit EnergyPlus after converting the input file format from IDF->epJSON or epJSON->IDF. This new flag reflects developer and user feedback that input file conversion should not require running a full simulation. Some output files will still be generated in this approach (for example, error file since there could still be errors during conversion).

Additionally, this PR adds a new input file conversion utility program. This program utilizes a new, small library for input parsing and validation, energyplusparser, from EnergyPlus InputProcessing code. The conversion program has a CLI that takes 1 or more input files and/or a newline delimited file of input file paths. If the code is compiled with OpenMP, this process is parallelized using a configurable number of threads. 

### Pull Request Author

### Reviewer

### DIFF
--- a/scripts/Epl-run.bat
+++ b/scripts/Epl-run.bat
@@ -316,7 +316,7 @@ IF EXIST SLABSurfaceTemps.TXT COPY "%epout%.expidf"+SLABSurfaceTemps.TXT "%epout
 IF EXIST SLABSurfaceTemps.TXT DEL SLABSurfaceTemps.TXT
 
 :  4. Execute EnergyPlus
-"%program_path%EnergyPlus" -c
+"%program_path%EnergyPlus"
 if %pausing%==Y pause
 
 :  5. Copy Post Processing Program command file(s) to working directory
--- a/src/ConvertInputFormat/main.cpp
+++ b/src/ConvertInputFormat/main.cpp
@@ -45,6 +45,7 @@
 // OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#include <string>
 #ifdef _OPENMP
 #include <omp.h>
 #endif
@@ -203,42 +204,42 @@ bool processInput(std::string const & inputFilePath, json const & schema, Output
         return false;
     }
 
-    std::ifstream input_stream(inputFilePath, std::ifstream::in);
+    std::ifstream input_stream(inputFilePath, std::ifstream::in | std::ofstream::binary);
     if (!input_stream.is_open()) {
         displayMessage("Input file path " + inputFilePath + " not found");
         return false;
     }
 
-    std::string input_file;
-    std::string line;
-    while (std::getline(input_stream, line)) {
-        input_file.append(line + EnergyPlus::DataStringGlobals::NL);
-    }
-
-    if (input_file.empty()) {
-        displayMessage("Failed to read input file: " + inputFilePath);
-        return false;
-    }
-
     try {
         if (!isEpJSON) {
+            std::string input_file;
+            std::string line;
+            while (std::getline(input_stream, line)) {
+                input_file.append(line + EnergyPlus::DataStringGlobals::NL);
+            }
+            if (input_file.empty()) {
+                displayMessage("Failed to read input file: " + inputFilePath);
+                return false;
+            }
+
             bool success = true;
             epJSON = idf_parser->decode(input_file, schema, success);
             cleanEPJSON(epJSON);
         } else if (isCBOR) {
-            epJSON = json::from_cbor(input_file);
+            epJSON = json::from_cbor(input_stream);
         } else if (isMsgPack) {
-            epJSON = json::from_msgpack(input_file);
+            epJSON = json::from_msgpack(input_stream);
         } else if (isUBJSON) {
-            epJSON = json::from_ubjson(input_file);
+            epJSON = json::from_ubjson(input_stream);
         } else if (isBSON) {
-            epJSON = json::from_bson(input_file);
+            epJSON = json::from_bson(input_stream);
         } else {
-            epJSON = json::parse(input_file);
+            epJSON = json::parse(input_stream);
         }
     } catch (const std::exception &e) {
         displayMessage(e.what());
         displayMessage("Errors occurred when processing input file. Preceding condition(s) cause termination.");
+        return false;
     }
 
     bool is_valid = validation->validate(epJSON);
@@ -247,6 +248,7 @@ bool processInput(std::string const & inputFilePath, json const & schema, Output
 
     if (!is_valid || hasErrors) {
         displayMessage("Errors occurred when validating input file. Preceding condition(s) cause termination.");
+        return false;
     }
 
     if (isEpJSON && !versionMatch) {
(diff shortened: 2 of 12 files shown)


mjwitte:
@mbadams5 If you're adding a standalone utility, why change the CLI behavior? 
This breaks the windows ep-launch run script which has "-c" hard-wired [here](https://github.com/NREL/EnergyPlus/blob/0e6e9c08a048eb9bb04387a61384bd842816c52f/scripts/Epl-run.bat#L319) (a promotional decision back when epJSON was added).
Other windows .bat scripts look ok. 
Not sure where to find run scripts for Mac and Linux.
Need to check other scripts and the new ep-launch workflows to see if any are using -c. @JasonGlazer

mbadams5:
@mjwitte I can reverse out the breaking change to the CLI behavior if that is what the EnergyPlus team thinks is best. I included the change because not everyone may want to include a standalone utility in their workflow to convert files. This way, if you have the energyplus executable then you can quickly convert a file. That said, I am open to reverting that change, but I would be curious to hear more opinions.

I am also open to new names for the standalone utility executable FYI.

Myoldmopar:
Can we just add a `--convert-only` flag?  I don't mind having a separate utility, but in this case, EnergyPlus behavior wouldn't break, we would just add another option.

mbadams5:
@Myoldmopar We can do that as well.

mjwitte:
@mbadams5 I get that. I'll commit the epl-run.bat change here now.

https://github.com/NREL/EP-Launch/issues/123 posted to add support for the conversion utility to EP-Launch 3.0.

mbadams5:
@mjwitte If we change to use a `--convert-only` flag, I will revert my original changes as well as your latest commit.

mjwitte:
Either way, I think it's best to remove the hard-wired "-c" option in Epl-run.bat since the users can't control it and it generates extra files. Any users interested in seeing the converted files can use the new utility. This was an undocumented "feature" anyway.

mbadams5:
@Myoldmopar @mjwitte I added the `--convert-only` flag so the existing flag functionality stays the same.

Myoldmopar:
CI looks very happy, reviewing the changes now.

scripts/Epl-run.bat
```diff
@@ -316,7 +316,7 @@ IF EXIST SLABSurfaceTemps.TXT COPY "%epout%.expidf"+SLABSurfaceTemps.TXT "%epout
 IF EXIST SLABSurfaceTemps.TXT DEL SLABSurfaceTemps.TXT
 
 :  4. Execute EnergyPlus
-"%program_path%EnergyPlus" -c
+"%program_path%EnergyPlus"
```
Myoldmopar, line 319: Pulling this back out is good.

src/ConvertInputFormat/main.cpp
```diff
@@ -0,0 +1,526 @@
...
+    auto schema = json::from_cbor(embeddedEpJSONSchema.first, embeddedEpJSONSchema.second);
+
+    auto number_files = files.size();
+    std::size_t fileCount = 0;
+
+#ifdef _OPENMP
+    omp_set_num_threads(number_of_threads);
+#endif
+
+#ifdef _OPENMP
```
Myoldmopar, line 499: :+1:

src/EnergyPlus/CMakeLists.txt
```diff
@@ -645,6 +664,16 @@ add_subdirectory(AirflowNetwork)
 #message( $<TARGET_PROPERTY:airflownetworklib,INTERFACE_INCLUDE_DIRECTORIES> )
 #include_directories($<TARGET_PROPERTY:airflownetworklib,INTERFACE_INCLUDE_DIRECTORIES>)
 
+add_library( energyplusparser STATIC ${INPUTPARSING_SRC} )
```
Myoldmopar, line 667: :+1:

src/EnergyPlus/CMakeLists.txt
```diff
@@ -645,6 +664,16 @@ add_subdirectory(AirflowNetwork)
...
+add_dependencies( energyplusparser GenerateEmbeddedEpJSONSchema )
+target_link_libraries( energyplusparser re2 )
+if(UNIX AND NOT APPLE)
+  target_link_libraries( energyplusparser dl )
+endif()
+if (WIN32)
+  target_link_libraries( energyplusparser Shlwapi )
+endif()
```
Myoldmopar, line 675: Hmm I expected E+ to depend on this DLL.  It's fine to continue building E+ as-is with the input processing built into it, I just sorta expected the E+ DLL to depend on this.  The code isn't duplicated anywhere, just some extra build rules, and negligible extra build time, so all good.

src/EnergyPlus/FileSystem.cc
```diff
@@ -134,10 +131,10 @@ namespace FileSystem {
             if (pathTail.size() == 0)
                 return absoluteParentPath;
             else
-                return absoluteParentPath + pathChar + pathTail;
+                return absoluteParentPath + DataStringGlobals::pathChar + pathTail;
 
         } else {
-            DisplayString("ERROR: Could not resolve path for " + path + ".");
+            std::cout << "ERROR: Could not resolve path for " + path + "." << std::endl;
```
Myoldmopar, line 137: Unexpected, but OK.

Myoldmopar (approved):

Myoldmopar:
No issues on code review, and CI is happy.  I'll build it locally and test it out and it can surely go in soon.

src/EnergyPlus/CMakeLists.txt
mbadams5: E+ could depend on this library. I don't have strong feelings one way or another.

src/EnergyPlus/FileSystem.cc
mbadams5: This was to avoid the `#include <EnergyPlus/DisplayRoutines.hh>` include since that has `#include <EnergyPlus/EnergyPlus.hh>`, which then requires ObjexxFCL includes. I could look to see if `DisplayRoutines.hh` ACTUALLY requires `EnergyPlus.hh` header or not.

mbadams5: I forgot that `DisplayRoutines.cc` has the following headers as well.
```cpp
// ObjexxFCL Headers
#include <ObjexxFCL/Fmath.hh>

// EnergyPlus Headers
#include <EnergyPlus/DataGlobals.hh>
#include <EnergyPlus/DataSystemVariables.hh>
#include <EnergyPlus/DisplayRoutines.hh>
```

src/EnergyPlus/CMakeLists.txt
Myoldmopar: I don't think we should make that change, not yet anyway.

src/EnergyPlus/FileSystem.cc
Myoldmopar: That's what I figured.  I wasn't expecting it when I saw it, but I figured it was to avoid a dependency. :+1:

Myoldmopar:
Output needs a small tweak:

```
$ ./ConvertInputFormat 1ZoneUncontrolled.idf 
Input file converted to  successfully | 1/1 | 1ZoneUncontrolled.idf
```

It has a blank on the *converted to* clause.

mbadams5:
@Myoldmopar I just pushed up a change that should fix that issue.

Myoldmopar:
Just built and re-ran, output looks good now.  Merging.  Thanks @mbadams5

Issue #7613, opened 2019-11-12
EnergyPlus as cmake sub project


kbenne:
Issue overview
--------------
EnergyPlus does not lend itself to be used as a cmake sub project, meaning if you try to add_subdirectory(EnergyPlus) from a parent cmake project it will typically fail. This is because the variable CMAKE_SOURCE_DIR is often used, where a better choice would be PROJECT_SOURCE_DIR. The former will resolve to the top level source directory, whereas the latter will always resolve to the root of the EnergyPlus project source directory. 

Reference
--------------
https://stackoverflow.com/questions/32028667/are-cmake-source-dir-and-project-source-dir-the-same-in-cmake

I will propose a PR to address this issue.

jmarrec:
Fixed via https://github.com/NREL/EnergyPlus/pull/7616

PR #7616, opened 2019-11-13
Make EnergyPlus suitable for use as a cmake sub project

kbenne:
EnergyPlus does not lend itself to be used as a cmake sub project, meaning if you try to add_subdirectory(EnergyPlus) from a parent cmake project it will typically fail. This is because the variable CMAKE_SOURCE_DIR is often used, where a better choice would be PROJECT_SOURCE_DIR. The former will resolve to the top level source directory, whereas the latter will always resolve to the root of the EnergyPlus project source directory.

#7613


Myoldmopar (approved):
I have no problem with this change.  If CI is happy with it across platforms, and no one else speaks up, it will drop in directly.

Myoldmopar:
So I am a little concerned that **none** of our _actual_ builds finished here.  Is there something about this change that requires a super new version of CMake or anything?

nrel-bot-2c:

nrel-bot-3:

Myoldmopar:
Windows and Mac have _still_ never reporting on this.  Windows is nearly caught up after all the build madness it went through, just one build plus the one it is currently running.  @kbenne there are a few conflicts after the IO freeze craze, but they look pretty easy to resolve.  If you can pull develop in one more time and push, we'll see if CI finally runs it across the board, and if not, we can investigate why CI hates this branch so much.

PR #7619, opened 2019-11-14
Add FinalizeSimulation function

kbenne:
This change adds a FinalizeSimulation function. This function is useful
for external programs (such as Spawn) that are calling EnergyPlus. There
is no expected change in normal program execution. The
FinalizeSimulation function contains the code from the end of the
ManageSimulation function and simply bottles it up into a new stand
alone function. This is useful in contexts such as spawn where the
client may end the simulation before the run period is complete.


nrel-bot-2c:

kbenne:
never mind. I'm going to cancel this request. 

giphy

PR #7627, opened 2019-11-19
Correct desuperheater reclaimed heat calculation and fix its delivery to tank source side

### LINKED ISSUE #7407: (Desuperheater heating energy) / (cooling coil cooling energy + cooling coil electric energy) never hit Heat Reclaim Recovery Efficiency
Issue overview
--------------
In running a model with an desuperheater object, whenever ``WATER HEATER:Water Heater Heating Energy [J](Hourly)`` is not 0, I would expect:

 ``DESUPERHEATER:Water Heater Heating Energy [J]`` / (``ASHP CLG COIL:Cooling Coil Electric Energy [J]`` + ``ASHP CLG COIL:Cooling Coil Total Cooling Energy [J]``) = 0.25,

which is the Rated Heat Reclaim Recovery Efficiency defined in idf.
IDF desuperheater snippet:

    Coil:WaterHeating:Desuperheater,
    desuperheater,            !- Name
      ,                                 !- Availability Schedule Name
    Desuperheater Setpoint Temp,  !- Setpoint Temperature Schedule Name
    5,                       !- Dead Band Temperature Difference {deltaC}
    0.25,                    !- Rated Heat Reclaim Recovery Efficiency
    50,                      !- Rated Inlet Water Temperature {C}
    35,                      !- Rated Outdoor Air Temperature {C}
    60,                      !- Maximum Inlet Water Temperature for Heat Reclaim {C}
    ,                          !- Heat Reclaim Efficiency Function of Temperature Curve Name
    DesuperheaterIn,         !- Water Inlet Node Name
    DesuperheaterOut,        !- Water Outlet Node Name
    WaterHeater:Mixed,       !- Tank Object Type
    water heater,                  !- Tank Name
    Coil:Cooling:DX:SingleSpeed,  !- Heating Source Object Type
    ashp clg coil,         !- Heating Source Name
    0.001,                   !- Water Flow Rate {m3/s}
    ,                        !- Water Pump Power {W}
    0.2;                     !- Fraction of Pump Heat to Water

Results:

 DESUPERHEATER:Water Heater Heating Energy [J] / (ASHP CLG COIL:Cooling Coil Electric Energy [J] + ASHP CLG COIL:Cooling Coil Total Cooling Energy [J]) :

image

Water heating energy:

image

The highest reclaimed ratio is just about 0.1+. Why couldn't we reclaim more to save water heating energy?

### Details
 - Platform (Windows 10)
 - Version of EnergyPlus (E+ 9.1.0)

### Checklist

yzhou601:
Pull request overview
---------------------
 - Fixes issue discussed in #7251 , specifically [this comment](https://github.com/NREL/EnergyPlus/pull/7251#issuecomment-555277394)
 - Also fixes issue #7407 

### Pull Request Author

### Reviewer


yzhou601:
image

Cooling coil source side energy is now balanced for gshp + stratified tank + desuperheater test file.

For a gshp model with `Coil:Cooling:WaterToAirHeatPump:EquationFit` (no heat reclaim):

image

This first commit fixed (Cooling Coil Electric Energy + Cooling Coil Total Cooling Energy) to be equal to Cooling Coil Source Side Heat Transfer Energy. Its annual energy impact:

image

yzhou601:
@rraustad Looking at these comments, it makes me confused if multiple desuperheater users are desired scenarios: https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HeatingCoils.cc#L1108-L1110
I am wondering which of the followings should it be to solve the reclaimed heat issue:
1. One source coil can only serve one desuperheater (water heating or heating), then should add error checking to avoid duplicate heat source.
2. One source coil at maximum is able to serve one water heating desuperheater + one heating desuperheater;
3. Multiple desuperheater users are allowed to use the same source coil with documentation giving some notice eg. the sum of reclaimed portions from the same source coil shouldn't exceed suggested values. 

The current stage looks to be designed for at least scenario 2 or scenario 3 because it decremented available reclaimed heat capacity across namespaces.

rraustad:
Well, that error checking was added so that there was no more than 1 desuperheater per source. The reason is that a desuperheater can condense refrigerant which would change the performance of the source. 

    Coil:Heating:Desuperheater,
       \memo Desuperheater air heating coil. The heating energy provided by this coil is reclaimed
       \memo from the superheated refrigerant gas leaving a compressor and does not impact the
       \memo performance of the compressor.

If you run the hot gas off a compressor to 2 desuperheaters, and each consumed 30% of the available heat reclaim energy, it is likely that the refrigerant would condense (either in the desuperheater or the refrigeration condenser) and the refrigeration system would have extra sub-cooling and result in extra cooling on the air-side of the refrigeration system at the cooling coil. It's also likely that 1 desuperheater would condense refrigerant but this would be a small (model assumption) amount of sub-cooling if max heat reclaim is limited and not grossly affect the refrigeration system performance. 

I couldn't find in the IDD (and did not look at the code) where a max reclaim efficiency is used (i.e., which input limits max reclaim energy). 

    Coil:Heating:Desuperheater,
      N1 , \field Heat Reclaim Recovery Efficiency
              \type real
              \minimum 0.0

    Coil:WaterHeating:Desuperheater,
      N2 , \field Rated Heat Reclaim Recovery Efficiency
             \type real
             \minimum> 0.0

I recall 30% when this object was originally added to E+. You have looked at the code and might know how limits to recovered heat is addressed. You are right that if 2 desuperheaters consumed no more than 30% of the total available heat reclaim energy then that should be OK (i.e., same as 1 desuperheater consuming 30%). Since the energy bucket is decremented each time it is used then more than 1 desuperheater per source should be OK, but would the user be aware if the 2nd desuperheater did not having enough heat available to do the job expected? So using more than 1 desuperheater per source does push the envelope. I do like allowing users to model whatever they like, but we can't allow users to provide extra sub-cooling without including that sub-cooling in the cooling coil model (which we do not do now and only a component model could handle this without additional inputs from the user). I do like your items number 2 and 3 and we have to be sure to model these correctly so that we do not do something that would impact cooling coil performance without accounting for the change in performance.

I hope this clarifies a few things as you move forward.

src/EnergyPlus/DataHeatBalance.hh
```diff
@@ -2164,24 +2164,17 @@ namespace DataHeatBalance {
...
-        std::string Name;       // Name of DX Coil
-        std::string SourceType; // SourceType for DX Coil
+        std::string Name;       // Name of Coil
+        std::string SourceType; // SourceType for Coil
         Real64 AvailCapacity;   // Total available heat reclaim capacity
+        Real64 ReclaimEfficiencyTotal;   // Total reclaimed portion
+        Real64 WaterHeatingDesuperheaterReclaimedHeatTotal;    // total reclaimed heat by water heating desuperheater coils
+        Real64 HVACDesuperheaterReclaimedHeatTotal;    // total reclaimed heat by water heating desuperheater coils
+        Array1D<Real64> WaterHeatingDesuperheaterReclaimedHeat; // heat reclaimed by water heating desuperheater coils
+        Array1D<Real64> HVACDesuperheaterReclaimedHeat; // heat reclaimed by water heating desuperheater coils
```
yzhou601, line 2177: Added waste heat user arrays to track them separately. Use this instead of decrementing AvailCapacity to avoid repeatedly extract heat in iterations and to keep data consistent through namespaces.

src/EnergyPlus/DataHeatBalance.hh
```diff
@@ -2648,11 +2623,11 @@ namespace DataHeatBalance {
     extern Array1D<ScreenTransData> ScreenTrans;
     extern Array1D<ZoneCatEUseData> ZoneIntEEuse;
     extern Array1D<RefrigCaseCreditData> RefrigCaseCredit;
-    extern Array1D<HeatReclaimRefrigeratedRackData> HeatReclaimRefrigeratedRack;
+    extern Array1D<HeatReclaimDataBase> HeatReclaimRefrigeratedRack;
     extern Array1D<HeatReclaimRefrigCondenserData> HeatReclaimRefrigCondenser;
-    extern Array1D<HeatReclaimDXCoilData> HeatReclaimDXCoil;
-    extern Array1D<HeatReclaimDXCoilData> HeatReclaimVS_DXCoil;
-    extern Array1D<HeatReclaimHPCoilData> HeatReclaimSimple_WAHPCoil;
+    extern Array1D<HeatReclaimDataBase> HeatReclaimDXCoil;
```
yzhou601, line 2631: Don't need so many data structs. Base struct is now good for all these types except for condenser.

src/EnergyPlus/HeatingCoils.cc
```diff
@@ -984,38 +984,155 @@ namespace HeatingCoils {
...
+                        ShowSevereError(CurrentModuleObject + ", \"" + HeatingCoil(CoilNum).Name +
+                                        "\" heat reclaim recovery efficiency must be >= 0 and <=0.9");
+                        InputErrorsFound = true;
+                    }
+                }
+            } else {
+                if (lNumericBlanks(1)) {
+                    HeatingCoil(CoilNum).Efficiency = 0.25;
+                } else {
+                    HeatingCoil(CoilNum).Efficiency = Numbers(1);
```
yzhou601, line 1002: Move this part ahead to use efficiency number later to check total reclaim efficiency (by multiple users, from the same source coil) not exceeding limitation.

src/EnergyPlus/HeatingCoils.cc
```diff
@@ -2611,16 +2724,29 @@ namespace HeatingCoils {
...
             } else if (HeatingCoil(CoilNum).ReclaimHeatingSource == CONDENSER_REFRIGERATION) {
-                HeatReclaimRefrigCondenser(SourceID).UsedHVACCoil = HeatingCoilLoad;
+                HeatReclaimRefrigCondenser(SourceID).HVACDesuperheaterReclaimedHeat(DesuperheaterNum) = HeatingCoilLoad;
+                HeatReclaimRefrigCondenser(SourceID).HVACDesuperheaterReclaimedHeatTotal = 0.0;
+                for (auto& num : HeatReclaimRefrigCondenser(SourceID).HVACDesuperheaterReclaimedHeat) 
+                    HeatReclaimRefrigCondenser(SourceID).HVACDesuperheaterReclaimedHeatTotal += num;
             } else if (HeatingCoil(CoilNum).ReclaimHeatingSource == COIL_DX_COOLING ||
                        HeatingCoil(CoilNum).ReclaimHeatingSource == COIL_DX_MULTISPEED ||
                        HeatingCoil(CoilNum).ReclaimHeatingSource == COIL_DX_MULTIMODE) {
-                HeatReclaimDXCoil(SourceID).AvailCapacity -= HeatingCoilLoad;
```
yzhou601, line 2622: If bug successfully fixed in these lines, CI results should reflect large diffs in desuperheaters connected with DX coils, the same applying to water heating desuperheaters + DX coils.

yzhou601:
Desuperheater was now using reclaimed heat and **saved tank source side outlet temperature**(temp from last time step) to calculate tank source side inlet temperature for current time step, then called tank calculation functions. But later tank calculation function was using **timestep-averaged tank tamperature** to recalculate source side heat transfer. This caused energy discrepancy between desuperheater heater rate and tank source side heat transfer rate. I think desuperheater should be called multiple times using updated tank temperature to ensure reclaimed heat was correctly fed into the tank source side (achieve some convergence). 
This issue caused a big energy discrepancy of a test file with desuperheater connected to a storage tank: 

  | WATER HEATER STORAGE:Water Heater   Source Side Heat Transfer Energy [J](Hourly) | COIL WATER HEATING DESUPERHEATER   1:Water Heater Heating Energy [J](Hourly)
-- | -- | --
Annual sum | 1.87E+09 | 2.12E+09

src/EnergyPlus/WaterThermalTanks.cc
```diff
@@ -8965,137 +9031,28 @@ namespace WaterThermalTanks {
...
-                                                                       WaterHeaterDesuperheater(DesuperheaterNum).RegulaFalsiFailedIndex1,
-                                                                       PartLoadRatio,
-                                                                       PartLoadRatio);
-                                    }
-                                }
-                            }
-                            NewTankTemp = WaterThermalTank(WaterThermalTankNum).TankTemp;
-                        } else {
+                        // Calculate twice for consistency of desuperheater and tank source side energy transfer
+                        for (int i = 0; i < 2; i++){
```
yzhou601: I added a loop of two iterations for desuperheater to get updated tank temperature (whenever desuperheater is in heat mode) and recalculate once, the energy discrepancy is cut a lot (from 1.87GJ vs 2.12GJ to 2.10GJ vs 2.13GJ). However the git diff here is messy and make it hard to review. @mjwitte @rraustad do you have any suggestion?

yzhou601:
@rraustad Do you mind taking a look at this? It's pretty close to completion, looks like it has fixed all targeted bugs. Some tests showed large diffs due to Qsource fix, another few due to desuperheater fixes. Any suggestion to improve these changes will be greatly appreciated!

src/EnergyPlus/WaterThermalTanks.cc
yzhou601: Looping twice is not enough. Should use some label to indicate convergence.

rraustad:
Your comment [here](https://github.com/NREL/EnergyPlus/pull/7627#issuecomment-558771235) about using last time step tank temp or this iteration tank temp scares me a little. When the time step starts, what is the tank temp provided to the desuperheater? It's the tank temp at the end of the last time step. I understand completely that there may be a need to use the current "something" within an iteration but in this case using this iteration tank temp will lead to a tank temp that continually changes during the iterations because the return water temp from any component will also change. We have been here before with other issues like this. The decision was to always use the starting temp to try to calculate a final tank temperature. I am sure this is not happening everywhere in the plant loop so this makes fixing this even harder.

yzhou601:
> Your comment [here](https://github.com/NREL/EnergyPlus/pull/7627#issuecomment-558771235) about using last time step tank temp or this iteration tank temp scares me a little. When the time step starts, what is the tank temp provided to the desuperheater? It's the tank temp at the end of the last time step. I understand completely that there may be a need to use the current "something" within an iteration but in this case using this iteration tank temp will lead to a tank temp that continually changes during the iterations because the return water temp from any component will also change. We have been here before with other issues like this. The decision was to always use the starting temp to try to calculate a final tank temperature. I am sure this is not happening everywhere in the plant loop so this makes fixing this even harder.

Thanks @rraustad for insights! I am not sure if this fix will be in the same situation as the previous one you mentioned. It might be necessary to clarify this fix here since git diff is hard to review.

I've added a loop in CalDesuperheaterWaterHeater to update tank source inlet temperature until desuperheater and tank are on the same page, to make sure reclaimed heat is delivered to tank. The logic of modeling a desuperheater is that `SimWaterThermalTanks` will call `CalDesuperheaterWaterHeater` instead of directly call `CalWaterThermalTankMixed` whenever there's a desuperheater connected (`CalDesuperheaterWaterHeater` calls `CalWaterThermalTankMixed` internally), so the added loop will always execute when top function `SimWaterThermalTanks` is called, to make sure the convergence between dsh and tank will always be achieved. 

I think your concern comes from potentially updating tank conditions continually in iterations. The only thing I update in this loop is the tank source inlet node temperature, as mentioned, previously it's based on previous time step tank temp, but now it's updated with pre-simulated tank temp in the loop, that means tank will always start at its saved tank temperature from last time step, just with different source side temperatures. This will only cause its final conditions changed (because of it making use of all reclaimed heat) but not reinitializing tank in iterations. For example, if saved tank temp (last time step) is 20C, desuperheater reclaimed heat is 1000w, its calculation of source side inlet temperature is using 20C, 1000w and source side mass flow rate, assumed to be 25C, after calling `CalWaterThermalTankMixed`, tank avg temp in current time step is 30C, which is over 25C, the tank source side heat transfer then, will report a negative value instead of 1000w. What I did is recalculating tank source inlet temperature to be based on 30C, 1000w and mfr again, assumed to be 35C this time, and rerun `CalWaterThermalTankMixed`. It happens several times until pre-simulated tank is no longer changing is final conditions (what I think is convergence). Within `CalWaterThermalTankMixed`, it's still picking up 20C as starting temperature, but source side is able to provide heat at a higher temperature to ensure, in current time step, reclaimed heat 1000w is correctly delivered to tanks, thus reduce the energy discrepancy. 

This issue is not very obvious if tank temperature is well maintained at specific values, but for tanks like preheat tank, its temperature changes significantly between time steps, this will be a problem. Therefore I wrote a unit test to target on such situations where temperature was greatly recovered in one time step, and it looks like this fix is working well. 

Let me know if your concern is gone or not. Feel free to talk whenever you have further concerns!

yzhou601:
And the updated source side inlet node is only used in tank calculation and desuperheater, without a sophisticated plant loop, nor other plant component. So in theory it won't introduce additional discrepancy except for a new tank final condition with desuperheater heat thoroughly reclaimed.

nrel-bot-3:
@yzhou601 @lgentile it has been 28 days since this pull request was last updated.

yzhou601:
@mitchute I resolved conflicts between master and this branch. Even though, ci is complaining of the format. Can you tell me how to apply clang? I am using VS2017 with Windows 10.

mitchute:
@yzhou601 I don't use VS all that much. However, I believe you have to add the [ClangFormat VS Extension](https://marketplace.visualstudio.com/items?itemName=LLVMExtensions.ClangFormat), then run the commands as shown.

mjwitte:
Actually, clang is built-in with VS2017, but there are options to activate clang, and to format automatically or manually. I use manual and apply as needed to avoid excess formatting diffs.
Tools - Options - Text Editor - C/C++ - Formatting - General. 

If you go with manual formatting, then apply it from Edit - Advanced - Format Selection.

yzhou601:
@mjwitte Thanks! So this manual method is applied to each file, right? Is there any way to run clang on the entire solution? For example, after making changes or resolving merge conflicts from different files, it will be great if I can run clang once across all source code files before push them up, and ci clang check won't be a concern.

mitchute:
@yzhou601 At most, only run it on the main files you have modified. If you have added a line or two in other files, **do not** run it there. If you run it across the entire code base, you're likely to cause merge conflicts with most of the dev team.

yzhou601:
> @yzhou601 At most, only run it on the main files you have modified. If you have added a line or two in other files, **do not** run it there. If you run it across the entire code base, you're likely to cause merge conflicts with most of the dev team.

Oh ok, I thought other files are already formatted with the same clang. Good to know, then I think I'm fine so far.

yzhou601:
@mitchute Do you mind starting looking at this PR whenever you're available. I was kind of stuck not knowing what's the next step. If there's any issue, let me know and we can talk about how to address. I'm kind of overwhelmed by getting CI results the same as before pulling in new Master branch, I think they're looking much closer now. This PR solves following issues:
1. Heat reclaim efficiency is not correctly applied. It was caused by the "-=" operated multiple times in one timestep on heat reclaim capacity. It got fixed by restructuring the related data structure and tracking each heat-reclaim user independently. This fix will basically impact results of desuperheaters connected with DX cooling coils, introducing ci diffs.
2. Qsource didn't account for the startup loss electric energy for WaterToAirHeatPump:EquationFit coils. This issue caused coil cooling energy + electric energy != coil source side heat transfer energy.  This fix might impact coil performance, introducing ci diffs.
3. The output `Desuperheater Water Heater Heating Energy` is not the same, or even close to` Water Heater Source Side Heat Transfer Energy`. There's a discrepancy between desuperheater and water thermal tank parts, caused calculated reclaim heat not totally applied to tank side, thus made reported variables look energy-unbalanced. This was fixed by adding a loop to iterate between these two functions more times until they're on the same page. This fix might have a performance impact(not quite sure how much).

mitchute:
Oh yeah, I was going to look at this but had forgotten. Thanks for the reminder.

For starters, could you come up with a better PR title? This one appears to have come straight from the GitHub default, which is the initial git commit message on this branch. The PR titles get summarized on the changelog for each release, as shown here for example: https://github.com/NREL/EnergyPlus/releases/tag/v9.2.0

I'll work on the review.

mitchute:
@yzhou601 are the example files linked over on #7251 still valid test cases for this issue? See files linked here: https://github.com/NREL/EnergyPlus/pull/7251#issuecomment-513366445

mitchute:
There's one unit test failure after pulling in develop, however, I can't replicate it on mac or linux. Let's see what happens after CI has a chance to run again.

mitchute:
Not sure what's going on here. I've run the failing unit test on Mac and Linux in debug and release modes and I don't see any problems. I guess I'll look at the rest of the results for now.

yzhou601:
> @yzhou601 are the example files linked over on #7251 still valid test cases for this issue? See files linked here: [#7251 (comment)](https://github.com/NREL/EnergyPlus/pull/7251#issuecomment-513366445)

Yes this is what's supposed to be fixed here. So the example files are valid to test. 
Though I've been using another test file for my local test. Let me know if those two get fixed as expected if you want to test them.

yzhou601:
> Not sure what's going on here. I've run the failing unit test on Mac and Linux in debug and release modes and I don't see any problems. I guess I'll look at the rest of the results for now.

I ran with Windows debug mode, it also passed.

mitchute:
I've checked the ```SingeSpeed``` file referenced here (https://github.com/NREL/EnergyPlus/pull/7251#issuecomment-513366445) and the results are looking good. The```MultiSpeed``` coil appears to be broken, so I didn't dig into that one.

The heat reclaimed now matches what is being specified in the IDF.

>     0.25,                    !- Rated Heat Reclaim Recovery Efficiency

Screen Shot 2020-03-05 at 12 01 52 PM

Previously, the ```Water Heater Source Side Heat Transfer Energy``` was not matching the Desuperheater ```Water Heater Heating Energy``` as shown in the next plot.

Screen Shot 2020-03-05 at 12 04 21 PM

Annual energy between the two were also off by about 8%.

Screen Shot 2020-03-05 at 12 06 53 PM

This now appears to be fixed by adding some additional iterations between these components and the results look much better.

Screen Shot 2020-03-05 at 12 05 42 PM

Annual energy totals also match up more closely.

Screen Shot 2020-03-05 at 12 07 59 PM

I also compared ```DOAToWaterToAirHeatPumpInlet.idf``` which is one of the other "big" diff files. The results are nearly identical. The condenser inlet temperatures looks to be the variable with the largest diffs.

Screen Shot 2020-03-05 at 12 15 23 PM

As noted, there are going to be diffs due to changes during the warmup periods, and there are a few other assorted power diffs, but overall, they look reasonable.

As far as CI results go, this looks good with the exception of the ```ZoneUnitarySysTest.UnitarySystemModel_FlowPerCoolingCapacityTest``` unit test failure which I can't replicate locally.

I'm going to pull in develop one more time in hopes of that being fixed. We will look into it after that if it still fails.

mitchute:
@Myoldmopar @mjwitte @rraustad would one of you take a look at this when you get a minute? The numeric diffs appear to be justified. However, I want to get a second look at the err diffs. Most of the err diffs are related to changes to the number of times the plant loop temperature limits are exceeded. Admittedly, most of these errors were there before, but now the number of occurrences for *most* of these errors has gone up, say, for example from 6 to 15.

And then there's this error message from ```ASHRAE9012016_OfficeLarge_Denver```:

```
    *************  ** Warning ** GetSpecificHeatGlycol: Temperature out of range (too low) for fluid [WATER] specific heat **
-   *************  **   ~~~   **   This error occurred 575 total times;
-   *************  **   ~~~   **   during Warmup 0 times;
-   *************  **   ~~~   **   during Sizing 0 times.
-   *************  **   ~~~   **   Max=-6.689049E-002 {C}  Min=-562.450726 {C}
+   *************  **   ~~~   **   This error occurred 551 total times;
+   *************  **   ~~~   **   during Warmup 0 times;
+   *************  **   ~~~   **   during Sizing 0 times.
+   *************  **   ~~~   **   Max=-0.135791 {C}  Min=-562.353545 {C}
```

Huh??? Sub absolute zero temperatures and we don't fatal out?

I'm going to hold off on this until we can take a little closer look at it.

mjwitte:
Hmm, those numbers hint that a large amount of heat is being extracted from a small flow rate. The question is whether that's coming from the desuperheater portion of the simulation, or elsewhere that's unrelated to this fix.

rraustad:
The only thing I can suggest is to break at the warning and then use the call stack to see what calculation/condition caused the issue.

mitchute:
Thanks @mjwitte and @rraustad for the comments.

@yzhou601 could you dig in here a little and see whether this is your code or just some other lingering issue?

yzhou601:
@mitchute The `ASHRAE9012016_OfficeLarge_Denver` doesn't have any desuperheater object, so it shouldn't be related to those changes to desuperheater functions. 
I broke at the warning and checked the call stack, the issue basically came from `CoolSys1`, where there're chillers on supply side + some water cooling coils on demand side. The supply side is where most warnings came from. 
I can't see why my changes increased the occurrences of these errors by tracking breakpoints for limited times, because almost all the breaks were caused by chiller side of CoolSys1 within scope of my observation. But I did see that this file has `Coil:Cooling:WaterToAirHeatPump:EquationFit`, to which I made some changes [here](https://github.com/NREL/EnergyPlus/pull/7627/files#diff-42185d37cb060b7dc404b9325c3f5693R2384) to include startup losses. The changes are fairly straightforward per [Richard's suggestion](https://github.com/NREL/EnergyPlus/pull/7251#issuecomment-555516623).

mitchute:
@Myoldmopar any last thoughts before this goes in? Numeric diffs all look justified. The only one that I'm not sure about is the ```UtilityRoutines_appendPerfLog2``` unit test failure.

Myoldmopar:
I have no issues with this going in.  I'm looking at the convection issue again, and don't worry about that other failure -- it's nothing related to this.  Merge at will.

PR #7635, opened 2019-11-25
Addresses Set DOAS Set point temp to default if fields are blank

### LINKED ISSUE #7471: Sizing:Zone DOAS design temperatures autosize to zero
Issue overview
--------------
In Sizing:Zone, there are inputs for DOAS sizing, e.g.
```
    ColdSupplyAir,           !- Dedicated Outdoor Air System Control Strategy
    13.8,                    !- Dedicated Outdoor Air Low Setpoint Temperature for Design {C}
    15.6;                    !- Dedicated Outdoor Air High Setpoint Temperature for Design {C}
```
The two temperatures default to autosize, but a user file with `NeutralSupplyAir` (unconditioned DOAS air) and the temperature fields left off (so they should default) results in zeros which undersizes the cooling equipment (and presumably would oversize the heating equipment, but here it's a heat pump, so cooling takes precedent):
```
    Yes,                     !- Account for Dedicated Outdoor Air System
    NeutralSupplyAir;        !- Dedicated Outdoor Air System Control Strategy

! <Zone Sizing DOAS Inputs>, Zone Name, DOAS Design Control Strategy, DOAS Design Low Setpoint Temperature {C}, DOAS Design High Setpoint Temperature {C} 
 Zone Sizing DOAS Inputs, BLOCKS-9436_MULTIFAMILY_1_CORE, NeutralSupplyAir, 0.000, 0.000
 Zone Sizing DOAS Inputs, BLOCKS-9436_MULTIFAMILY_2_CORE, NeutralSupplyAir, 0.000, 0.000
```
This results in undersized cooling equipment and unmet load hours.

Note that there is not an example file that uses the NeutralSupplyAir option, probably should add one or modify one?

### Workaround
Enter the desired temperatures. For example, if the DOAS does not condition the supply air, then the low temperature should be the winter design day low, and the high temperature should be the summer design day high. If the DOAS conditions the air, then enter the high and low setpoints for the DOAS supply.

### Details
 - Version of EnergyPlus v9.2.0--c00e682ec8
 - Direct e-mail to @mjwitte on 23 Aug 2019
### Checklist
 - [x] Defect file attached
[in_V2_MZ-DOAS_Edit-5-more output-v92.idf.txt](https://github.com/NREL/EnergyPlus/files/3537428/in_V2_MZ-DOAS_Edit-5-more.output-v92.idf.txt)

08/23 is an example day with unmet cooling loads

Nigusse:
Pull request overview
---------------------
 - Fixes #7471
 - Sets default values to "DOAS Design Low Setpoint Temperature {C}" and "DOAS Design High Setpoint Temperature {C}" fields in zone sizing object if the fields are left blank and the filed "Account for Dedicated Outdoor Air System" is set to "YES".  
 - added new example file that demonstrates DOAS with neutral supply air condition providing ventilation air and fancoil units serving each zone.

### Pull Request Author

### Reviewer


src/EnergyPlus/SizingManager.cc
```diff
@@ -3166,8 +3166,20 @@ namespace SizingManager {
...
+                        {
+                            ZoneSizingInput(ZoneSizIndex).DOASLowSetpoint = rNumericArgs(17);
+                        }
+                        if ( lNumericFieldBlanks( 18 ) ) {
+                            ZoneSizingInput(ZoneSizIndex).DOASHighSetpoint = DataSizing::AutoSize;
+                        } else
+                        {
+                            ZoneSizingInput(ZoneSizIndex).DOASHighSetpoint = rNumericArgs(18);
+                        }
+
```
Nigusse: If these two fields are left blank and ZoneSizingInput(*).AccountForDOAS is "true" ("Account for Dedicated Outdoor Air System" is set to "YES"), then the "DOASLowSetpoint" and "DOASHighSetpoint" values are set to Autosize.

Nigusse:
Example of inputs for DOAS sizing and results after the fix:

     Sizing:Zone,
       blocks-9436_multifamily_1_Core,  !- Zone or ZoneList Name
       ...
       blocks-9436_multifamily_1_Core Design Spec Zone Air Dist,  !- Design Specification Zone Air Distribution Object Name
       Yes,                     !- Account for Dedicated Outdoor Air System
       NeutralSupplyAir;        !- Dedicated Outdoor Air System Control Strategy

      After the fix blank fields are correctly set to default values:

      ! <Zone Sizing DOAS Inputs>, Zone Name, DOAS Design Control Strategy, DOAS Design Low Setpoint Temperature {C}, DOAS Design High Setpoint Temperature {C} 
      Zone Sizing DOAS Inputs, BLOCKS-9436_MULTIFAMILY_1_CORE, NeutralSupplyAir, 21.100, 23.900
      Zone Sizing DOAS Inputs, BLOCKS-9436_MULTIFAMILY_2_CORE, NeutralSupplyAir, 21.100, 23.900

Nigusse:
Added new example file that demonstrates a central DOAS with neutral supply air condition providing ventilation air to the zones and each zone is served with a FanCoil unit.

nrel-bot-2:
@Nigusse @lgentile it has been 30 days since this pull request was last updated.

Myoldmopar:
CI is super happy with this fix.  I'll take a look at the code changes now.  This is ready to go, right @Nigusse ?

tst/EnergyPlus/unit/SizingManager.unit.cc
```diff
@@ -244,3 +249,74 @@ TEST(SizingManagerTest, TimeIndexToHrMinString_test)
     EXPECT_EQ("16:39:00", TimeIndexToHrMinString(333));
     EXPECT_EQ("24:00:00", TimeIndexToHrMinString(480));
 }
+
+TEST_F(EnergyPlusFixture, SizingManager_DOASControlStrategyDefaultSpecificationTest)
```
Myoldmopar, line 253: I :heart: this unit test

src/EnergyPlus/SizingManager.cc
```diff
@@ -3166,8 +3166,19 @@ namespace SizingManager {
                                 ErrorsFound = true;
                             }
                         }
-                        ZoneSizingInput(ZoneSizIndex).DOASLowSetpoint = rNumericArgs(17);
-                        ZoneSizingInput(ZoneSizIndex).DOASHighSetpoint = rNumericArgs(18);
+                        // blank fields are set to autosize
```
Myoldmopar: I feel like there is probably a valid reason due to different configurations, etc., but I should at least ask: couldn't the IDD just have a default /autosize on these fields?

Myoldmopar (commented):
The code changes do make sense, and I appreciate the new integration test along with a great, succinct unit test.  However, I would like to understand why the code changes are necessary and not just an IDD change to use a default of autosize before I accept it.  Thanks @Nigusse

src/EnergyPlus/SizingManager.cc
rraustad: There are 2 things. If the fields are present and blank then yes, the default = autosize would work. Second is if the fields are not present and the previous field for Dedicated Outdoor Air System Control Strategy is present, what would the result be with either method?

Myoldmopar: So maybe you also need to also check NumNums or whatever the var is here?

Nigusse: As @rraustad said leaving to default won't work if these fields don't existing as blank fields in the object.  I will add a check for NumNumbers as well.

mjwitte: Back in the old days all fields in the range of \min-fields would get their defaults inserted. I'm not sure if this is still the case since the input processor rewrite for JSON. The current \min-fields for this object is ridiculously short. Probably should include all fields because of the importance of these. It's not like this has extensible repeating fields at the end. Every field has a unique function.

rraustad: We discussed this. It looks like lNumericFieldBlanks is true when the input field is not present in the object and when the input field is present then the default of AutoSize is used and this code also works. So it looks like this code covers all cases and no need to check for NumNumbers. Formatting does need to be corrected though.

rraustad: @Myoldmopar I thought you added a script to check for formatting? If so, then why are there no warnings from the CI?

Nigusse: @mjwitte @Myoldmopar The current \min-field for Sizing:Zone is set to 18, this will become 27 if we include all fields.

Nigusse: @mjwitte I tested with the \min-field for Sizing:Zone in the idd changed to 27 but did not make any difference. For some reason the JSON idd file created during build did not use the \min-field value I set in the idd, instead the \min-field value in JSON idd is defaulted to 18. I am not sure what the problem is. I suggest that this JSON \min-field value issue investigated separately.

mjwitte: did you rebuild EnergyPlus after changing the IDD? That's necessary, it doesn't read the idd directly anymore.

mjwitte: I do see various values of `"min_fields":` throughout Energy+.schema.epJSON but I don't know if the input processor still uses these for anything. Either way, given the behavior of IDF Editor, transition, and other tools, I think we should increase min-fields for this object to get these fields visible in the idfs.

Nigusse: Yes, I did. I verified that the Energy+IDD copied to build/run directory is updated but not the Energy+JSON Idd.

mbadams5: @mjwitte The [IdfParser](https://github.com/NREL/EnergyPlus/blob/7ff93c0f885e7955217a876c59b1bb82d1546d90/src/EnergyPlus/InputProcessing/IdfParser.cc#L337) uses `"min_fields"` to then set `idf_max_fields` and/or `idf_max_extensible_fields`, which are then used by the InputProcessor::getObjectItem(). I don't see a reason from a cursory glance why the epJSON schema wouldn't reflect the correct min-fields value.

rraustad: @mjwitte @mbadams5 what we did was change the min-fields in Energy+.idd.in to 27 and then recompiled (we did not change min_fields in Energy+.schema.epJSON.in which was still 18). The new Energy+.idd had 27 as min-fields but Energy+.schema.epJSON still had 18. Does the Energy+.schema.epJSON.in also need to be changed to 27?

Since we were confused on this, we changed the input file to have 17 fields in one Sizing:Zone object and the file ran successfully. This also confused us.

mjwitte: Look at Build\Products\Energy+.schema.epJSON And check the date on it. This rebuilds whenever Energy+.idd.in changes.

rraustad: Yes the dates were the same. We even deleted both files (idd and json) in the build folder and saw that they were there again, and with current time stamp, after the compile. So does the Energy+.schema.epJSON.in in the repository every need to be edited? We couldn't figure out where the 18 in the schema came from once we changed min-fields in the idd.

mjwitte: Hmm, I can't reproduce that here. I changed sizing:zone min-fields to 25 as a test, in \IDD\Energy+.idd.in. Then in VS, Build EnergyPlus. Both IDD\Energy+.schema.epJSON.in and Build\Products\Energy+.schema.epJSON now have min-fields 25 (line 64409).
```
        "Sizing:Zone": {
            "patternProperties": {

< . . . >
                        "dedicated_outdoor_air_high_setpoint_temperature_for_design"
                    ]
                }
            },
            "type": "object",
            "memo": "Specifies the data needed to perform a zone design air flow calculation. The calculation is done for every sizing period included in the input. The maximum cooling and heating load and cooling, heating, and ventilation air flows are then saved for system level and zone component design calculations.",
            "min_fields": 25.0
        },
        "DesignSpecification:ZoneHVAC:Sizing": {

rraustad: So that may mean then that @Nigusse has multiple versions of the idd/json files and not sure which one is being used. I can attest that we checked the date after the build and those were new files. At least someone can get the right min fields in both places so suspect this is something on our end.

Nigusse: One problem that I saw was that the IDD\Energy+.schema.epJSON.in is not updated after build even if I changed the sizing:zone min-fields in \IDD\Energy+.idd.in. Then, I removed my build folder and run CMake to create a new one. This resolved the problem.

Nigusse: Now I will extend the Sizing:Zone object min-fields to 27 to cover all fields as suggested and I may not need to check these two fields (rNumericArgs(17) and rNumericArgs(18)) are blank or not. I will test the code locally and push the updated branch.

rraustad: We discussed this again, at length, min-fields = 18 vs 27. I don't see the need and it seems the user still does not need to enter blank commas at the end of the input object if those fields are not used. I don't understand this but if the defect file runs (with < min-fields) and all example files run, then I suggested go ahead and use 27 and see if anything breaks.

Nigusse: Apparently, the min-field rule some how has changed. So, now unless the required-field rules are not violated, then E+ runs regardless of the min-fields value, right? @mjwitte

mjwitte: min-fields does not mean that the user file has to include all of those - it never has meant that. It's supposed to mean that those fields will be populated on input with defaults if defaults are declared in the IDD. Also, transition and IDF Editor will fill out any shorter object to that number of fields anytime they touch an idf file.

Nigusse:
@Myoldmopar This branch is ready for final review or merge into develop.

Myoldmopar:
CI is happy, and I'm glad the conversation above looks to have resolved with everyone in agreement that the min-fields change was all that was needed.  I'm ready to merge this in.  @mjwitte anything else?

mjwitte:
Nope. Merge away.

Issue #7601, opened 2019-11-05
Unsuccessful simulation using PerformancePrecisionTradeoffs


shorowit:
Issue overview
--------------
When I add `PerformancePrecisionTradeoffs, Yes;` to my IDF, the simulation just stops after "Initializing Simulation" and the err file is empty. If I change the object to `PerformancePrecisionTradeoffs, No;`, it runs successfully.

File to reproduce:
[in.idf.txt](https://github.com/NREL/EnergyPlus/files/3811168/in.idf.txt)

@XuanLuoLBL 

### Details
Some additional details for this issue (if relevant):
 - Windows 10
 - EnergyPlus 9.2.0

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

shorowit:
My expectation was that the direct coil solution would be used for PLR calculation for all supported coils in the model (`AirLoopHVAC:UnitarySystem` with `Coil:Heating:Fuel` and `Coil:Cooling:DX:SingleSpeed` in the attached file) and not used for all unsupported coils/systems (perhaps `ZoneHVAC:IdealLoadsAirSystem` in the attached file?).

I can't tell if our file is supposed to run but there's a bug, or supposed to throw an error message.

yzhou601:
With the same object turned on for another model with `Coil:Cooling:DX:SingleSpeed` and `Coil:Heating:DX:SingleSpeed`, I can run 1 successfully over 5 simulation trials (without changing anything in the model or weather file). [Defect file](https://drive.google.com/file/d/16ffa0KWQuk3lyPr3dmr5xwaaQyR6oyJC/view?usp=sharing)

xuanluo113:
@lgu1234 Lixing, it seems it's not anything related to the object input. Do you mind checking the coil implementation? Thanks

lgu1234:
@XuanLuoLBL I will take a look next week.

shorowit:
https://github.com/NREL/EnergyPlus/pull/7636

Myoldmopar:
Closed via #7636

PR #7636, opened 2019-11-25
Call DXCoils::DisableLatentDegradation after getting cooling coil index

lgu1234:
Pull request overview
---------------------
The program was crashed due to 0 cooling coil index, because DXCoils::DisableLatentDegradation is called before getting cooling coil index. The function is moved to a right location after getting cooling coil index. 

### Pull Request Author

### Reviewer


shorowit:
I can confirm that this fixes the crash.

I also see the simulation results change very slightly when using this object, which suggests that it's doing something, however the err file suggests that it's unused:
```
   ** Warning ** The following lines are "Unused Objects".  These objects are in the input
   **   ~~~   **  file but are never obtained by the simulation and therefore are NOT used.
   **   ~~~   **  Only the first unused named object of an object class is shown.  Use Output:Diagnostics,DisplayAllWarnings; to see all.
   **   ~~~   **  See InputOutputReference document for more details.
   ************* Object=PerformancePrecisionTradeoffs=PerformancePrecisionTradeoffs 1
```

lgu1234:
@shorowit I ran the test file and unable to get your message. Could you upload the your test file?

shorowit:
@lgu1234 Here you go:
[in.idf.txt](https://github.com/NREL/EnergyPlus/files/3898492/in.idf.txt)
[eplusout.err.txt](https://github.com/NREL/EnergyPlus/files/3898497/eplusout.err.txt)

lgu1234:
@shorowit Thanks for your input file. I checked the cause. Since PerformancePrecisionTradeoffs is a unique object and does not have an object name field, it is unnecessary to check the object name and confuses the program to provide a wrong message.

                auto const &thisObjectName = UtilityRoutines::MakeUPPERCase(instance.key());
                inputProcessor->markObjectAsUsed("PerformancePrecisionTradeoffs", thisObjectName);

The fix is easy by removing thisObjectName check and checking uniqueness.

This issue is nothing to do with "Unused Objects" and the fix addressed the crash. Please close this issue. I am going to create a new issue to address "Unused Objects".

Myoldmopar:
@lgu1234 the hash symbol (octothorpe, #) in the branch name is causing CI to skip running this branch.  I can review this when it is ready and provide regression results manually so that you don't have to create a new branch and pull request.  Let me know when you feel this is completely ready to go.

lgu1234:
@Myoldmopar Thanks for letting me know. I will not create a branch with # from now on. The branch is ready for review.

Myoldmopar:
@shorowit thanks for verifying the fix worked.  Also thanks for noting the error file issue.  This is indeed merely that in the new input processing approach, where we grab the inputs directly from the internal JSON representation, we need to call to mark the object as "used".  This needs to be cleaned up for this object.

@lgu1234 I tried to verify the bug fix also locally, but when I transitioned up the defect file, it ended up truncating the EMS code due to long lines, so it would no longer run.  I will base the verification on @shorowit's prior verification.  Also, it looks like we don't have an example file with the tradeoff object, so that's fun.

There were a few build warnings due to unused variables.  I fixed them and will push that up then merge this into develop right away.

Myoldmopar:
Was there a defect to close with this PR?

shorowit:
https://github.com/NREL/EnergyPlus/issues/7601

lgu1234:
@Myoldmopar I am going to make an example file for the object.

PR #7637, opened 2019-11-26
Chiller Plant Component Refactor

Myoldmopar:
Pull request overview
---------------------
This PR introduces a newly refactored set of plant chiller models.  This is refactoring the PlantChillers.cc file, which houses the electric chiller, engine driven chiller, gas turbine chiller, and constant COP chiller.  This refactor will ultimately be a major streamlining of these models.

Diffs started showing up at step 7, when I started moving a few module level variables to structs.  This step is important in ensuring that there is no component cross-talk.  When diffs show up here, it is indicative that in some manner, multiple chillers are accidentally sharing data.  In this case, I found diffs, went back through and more carefully stepped through the step, doing at one point one variable at a time, to find out which caused the issue.  As an example of this, see commit 40cff82, where I am simply taking the shared condenser outlet temperature module level variable and putting on as a member variable on the chiller struct.  This should definitely cause zero diffs, but definitely causes some diffs.

The diffs I see locally are in 3 files: CompSetPtControl, RadLoTempCFloHeatCool, and TermReheatScheduledPump.

- RadLoTempCFloHeatCool has exceedingly small math diffs, with this meter file diff:
image
- TermReheatScheduledPump shows more significant diffs at 2100 monthly joules, which works out to about 0.001 Watts
image
- CompSetPtControl has big diffs, and they actually are significant.  But they appear to be significant in a good way.  The chiller heat transfer is a good match:
image
But yet, the chilled water and condenser water demand rates are all wonky:
image
image

I believe this is because some bookkeeping was allowing some state variables to get passed to plant that were actually invalid, yet the final heat transfer reported by the chiller looked all great.  And also improved is that the condenser demand is lower, presumably because the outlet temperature is getting reported to a better value now. :+1:

I am locally running regressions on ~200 files that contain any one of these chillers, and all but these three are running perfectly with no diffs or issues whatsoever, so I feel pretty confident that this is just hitting a corner case which is now being fixed.

I anticipate wrapping this up here, though it is a possibility that this merges with these diffs to allow the rest of the refactor to go in cleaner.  First things first is to see if CI backs up my local diff assessment.


Myoldmopar:
Yep, that's *exactly* what I was expecting for diffs. I think I'll just keep chugging along and we'll see if the diffs change after this point.

Myoldmopar:
Continuing to refactor out the module level variables.  This is turning into a beast.  Nearly every time I change an individual module variable away, more diffs pop up.  Here is one example where I only changed out the evap outlet temp module variable away:

image

The chiller cooling rate changes, but only momentarily.  A module variable was being used without proper initialization, only for a very brief time.  In fact, looking at the timestep level output reveals how quick it was:

image

The unit stays running instead of turning off for a single timestep, helping make the case that this was because of a bad (missing) initialization.

This is a significant change, but in the right direction.  Continuing.....

Myoldmopar:
@mitchute I think with this last commit, this is ready for review.  This is a *massive* overhaul of the chiller code, however, so reviewing it is going to rely mostly on the CI results.  I expect this last commit (1b81bb9) to be clean -- in the sense that there will be almost 200 bnd diffs, some 14 big math diffs, and a few random other things.  I'll take some time once CI is done to summarize the results for you so you don't have to slog through it.

Myoldmopar:
# Review Info

## High level features
 - 4 chiller models converted to plant component:
   - electric chiller
   - engine driven chiller
   - gas turbine chiller
   - constant cop chiller
 - Chiller base class fixed up nice and some code hoisted up to the base class level where practical
   - In an ideal world, some time would be spent reorganizing the code further and collapsing more logic into the base class -- and pulling in other chillers to use this as well, but that isn't within scope and could be a deep rabbit hole
 - Net of almost 2000 lines removed, woot!
 - Runtime results to be summarized here; I wouldn't be surprised if it was a little improved, but I'm not expecting *massive* improvement yet (see API phase 2)

## Diff Results
 - 197 files showing bnd diffs, as expected
 - 8 files with EIO diffs:
   - All of them are tiny warmup convergence changes, except 5ZoneIceStorage, which is showing some component sizing differences, but they are tiny too (chiller sizing from 56939.64304 W to 56939.64180 W)
 - 9 files are showing some form of mathdiffs - 6 small and 3 big

### 5ZoneDetailedIceStorage
Meter diffs are on the order of 3W.  Looking at CSV diffs, chiller operation shows huge diffs in the diff file, but the actual operation is showing the same effect I noted previously: something was shared at the module level and being used "uninitialized" (it was initialized to a meaningful floating point value, but it was being carried over from a previous iteration or previous chiller instance).  Here is the chiller heat transfer rate showing the small blips as the chiller comes on:
image

### 5ZoneDetailedIceStorage2
Same meter diff, same math diff, just slightly different profile due to the different input file configuration:
image

### 5ZoneDetailedIceStorageCubicLinear

Exact same behavior and meter diffs as the detailed ice storage files above.

### 5ZoneDetailedIceStorageSimpleCtrl

This file looks similar, where there is a significant chiller heat transfer change at just a couple select times.  To see the effect of this change, here are the peak diffs of some other cooling outputs in this file, and note that these peak diffs occur at the exact same time as the chiller showing diffs:

image

For almost the whole run, the diffs are incredibly tiny, but at that single timestep where the chiller has diffs, the ice storage temperature is off by less than a degree and the ice storage rate is less than a Watt.  After this momentary diff, the values agree to very high precision once again.

### 5ZoneIceStorage

Looks the same as the others.  Here's a different look at the diff.  This is the absolute diff in chiller COP over the course of both design days and two two-week run periods:

image

The same momentary blip as the chiller comes on a couple times and then everything is a near-perfect match.  This story is starting to get old...

### CompSetPtControl

Well, there is definitely something going on here:
image

Little chiller (in green) matches really well.  Big chiller (in black) does not.  In this branch, the chiller attempts to come on for a few time steps.  This needs investigation.

### HeatRecoveryElectricChiller

These absolute diffs are also showing as blips when looking at them over the whole simulation run:

image

Looking at the values overlaid however reveals this file is running fine:

image

There is, once again, something momentary happening, but it's cleared away and running well.

### PlantApplicationsGuide_Example1

Max diff is on the order of 1e-7 W, nothing to see here.

### TermReheatScheduledPump

This also has a momentary diff, but even it is tiny:

image

## Summary

I need to figure out what happened on the comp set point control file.  Other than that one file, I believe all these diffs are fine.

mitchute:
@Myoldmopar thanks for the summary. The results are looking good. Let me know once you're done investigating the comp set point control file.

Myoldmopar:
OK, I dug through the depths and found out some overall negative, but for the sake of this PR positive, news.  The electric chiller model is not behaving well under a variety of operating conditions.  When the chiller has no condenser flow from the solver, but there is evap flow, the calc routine returns early but the module level variables for evap outlet temp and others were not reinitialized.  So the chiller update routine would then run and incorrectly update the outlet temperatures.  This led to invalid conditions.  On many time steps.  Notice the time steps shown here where the chiller has a condenser delta T but the condenser flow rate is zero:

image

This branch here has similar issues but they are being manifested at different time steps, and so the results look all wonky.

In summary, the broken electric chiller mechanics should not hold this up any longer.  This should go in, and I'll open up a new ticket to investigate the poorly behaved electric chiller soon, separate from all these other changes.

@mjwitte you have any issue?  If not, @mitchute feel free to give it a final look over and let me know if you find anything.

mjwitte:
@Myoldmopar Chiller:Electric and I are not on speaking terms at the moment, so maybe this explains part of the rift.

Myoldmopar:
I'm going to merge this in tonight unless there is objection.

mitchute:
@Myoldmopar looks good. Merging.

PR #7639, opened 2019-12-04
Absorption Chiller Plant Component Refactor

mitchute:
### Pull Request Author

### Reviewer


mitchute:
Ready for review.

src/EnergyPlus/ChillerAbsorption.hh
```diff
@@ -54,51 +54,54 @@
...
-    extern Real64 QEvaporator;         // W - rate of heat transfer to the evaporator coil
-    extern Real64 EvaporatorEnergy;    // J - heat transfer to the evaporator coil
-    extern Real64 QCondenser;          // W - rate of heat transfer to the condenser coil
-    extern Real64 CondenserEnergy;     // J - heat transfer to the condenser coil
+    extern int numBlastAbsorbers; // number of Absorption Chillers specified in input
 
     extern Array1D_bool CheckEquipName;
 
-    extern bool GetInput; // When TRUE, calls subroutine to read input file
+    extern bool getInput; // When TRUE, calls subroutine to read input file
```
Myoldmopar, line 70: Note for later, a lot of times these getInput flags aren't actually needed to be defined and exposed via the header file, it can be defined at the top of the cc file only, so just watch out for that on later refactors.  Note there are occasions where this isn't actually true, so I'm just throwing this out there.

src/EnergyPlus/PlantLoopEquip.cc
```diff
@@ -345,32 +344,9 @@ namespace PlantLoopEquip {
...
-                    sim_component.MinLoad = MinLoad;
-                    sim_component.OptLoad = OptLoad;
-                    sim_component.CompNum = EquipNum;
-                    sim_component.TempDesCondIn = TempCondInDesign;
-                }
-                if (GetCompSizFac) {
-                    sim_component.SizFac = SizingFac;
-                }
+
+                dynamic_cast<ChillerAbsorption::BLASTAbsorberSpecs*> (sim_component.compPtr)->EquipFlowCtrl = EquipFlowCtrl;
```
Myoldmopar: It might be better to just set this inside the function instead.  There are examples of this but I can't think of one at the moment.  Basically, in the oneTimeInit function, or the simulate function, get the flow control value off the plant structure:

```
this->EquipFlowCtrl = DataPlant::PlantLoop(componentLocation.LoopNum).LoopSide(...blah...    .EquipFlowCtrl;
```

This avoids the dynamic cast.

tst/EnergyPlus/unit/Fixtures/EnergyPlusFixture.cc
```diff
@@ -313,6 +310,7 @@ void EnergyPlusFixture::clear_all_states()
     BoilerSteam::clear_state();
     BranchInputManager::clear_state();
     CoolingPanelSimple::clear_state();
+    ChillerAbsorption::clear_state();
```
Myoldmopar, line 313: Always a great sign!

Myoldmopar:
@mitchute could you comment on my two notes above.  If you can address those, and pull develop in, it would be great to get one more clean build and then this can go in.

src/EnergyPlus/ChillerAbsorption.hh
mitchute: @Myoldmopar thanks. This is externed because FaultsManager uses it.

src/EnergyPlus/PlantLoopEquip.cc
mitchute: @Myoldmopar fixed, and develop is pulled in.

Myoldmopar:
Refactors for absorption chiller, indirect absorption chiller, swimming pool, and central GSHP were all merged down locally and tested at once.  The only diff was in the indirect, where the missing reset on the variable was causing an expected diff.  All four PRs will merge at once here to start clearing the path for CI.

PR #7541, opened 2019-10-02
Water Thermal Tanks Plant Component Refactor

mitchute:
Pull request overview
---------------------
Refactor water thermal tanks to use plant component.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #7647, opened 2019-12-06
Add tolerance for high humidity OA control

### LINKED ISSUE #7644: OA controller high humidity control needs a tolerance
Issue overview
--------------
While hunting down diffs due to surface reordering, test file UnitarySystem_FurnaceWithDXSystemRHControl.idf was showing big diffs in heating coil energy use for the winter design day. Tracked it down to the OA controller which was randomly activating the high humidity control economizer (which in this file sets the OA fraction to 0.9).

The problem is here in MixedAir::CalcOAEconomizer:
```
                    if (!this->ModifyDuringHighOAMoisture && Node(this->NodeNumofHumidistatZone).HumRat <= this->OAHumRat) {
                        HighHumidityOperationFlag = false;
                    } else {
                        HighHumidityOperationFlag = true;
                    }
```

For the winter design day, all of the humidity ratios are initialized to the outdoor humidity ratio, and because the internal gains are off for the winter day, so no matter what the economizer does the humidity ratio doesn't change. But this test is randomly turning up false which opens the OA damper. Adding  a small tolerance here corrects the problem.

This raises another question - the high humidity control does not check any of the other economizer limits (such as min OA dry bulb, etc.). Should it? *No. The I/O Ref states that high humidity control takes precedence. It's not really an economizer function, so economizer limits don't apply.*

### Details
 - Platform Win10 64
 - Version of EnergyPlus 9.2.0-d978a3721a
 - Reference #7575 

### Checklist

mjwitte:
Pull request overview
---------------------
 - Fixes #7644 
 - The high humidity control option in Controller:OutdoorAir would somewhat randomly bring in excess outdoor air when the indoor and outdoor humidity ratios were essentially equal. Adding a small tolerance makes the control more consistent.
 - Diffs expected for UnitarySystem_FurnaceWithDXSystemRHControl.idf and others with this control.

### Explanation of Diffs ###
These test files have OA controller "High Humidity Control" = Yes.
FurnaceWithDXSystemRHControl, HeatPumpAirToAirWithRHControl, HPAirToAir_wSolarCollectorHWCoil, HP_wICSSolarCollector, HVACStandAloneERV_Economizer, UnitarySystem_FurnaceWithDXSystemRHControl, and UnitarySystem_WaterCoils_wMultiSpeedFan. Six of these are showing small or large eso diffs, as expected.

Looking at UnitarySystem_FurnaceWithDXSystemRHControl, for example. The following chart shows before and after, zone humidity ratio, OA humidity ratio, and "Air System Outdoor Air High Humidity Control Status". See how the high humidity control was kicking on randomly on the winter design day (left end of chart). After the fix, it stays off.

image
image

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/DataHVACGlobals.hh
+++ b/src/EnergyPlus/DataHVACGlobals.hh
@@ -75,6 +75,7 @@ namespace DataHVACGlobals {
 
     // MODULE PARAMETER DEFINITIONS:
 
+    Real64 const SmallHumRatDiff(1.0E-7);
     extern Real64 const SmallTempDiff;
     extern Real64 const SmallMassFlow;
     extern Real64 const VerySmallMassFlow;
--- a/src/EnergyPlus/MixedAir.cc
+++ b/src/EnergyPlus/MixedAir.cc
@@ -4593,7 +4593,8 @@ namespace MixedAir {
                 if (ZoneSysMoistureDemand(this->HumidistatZoneNum).TotalOutputRequired < 0.0) {
                     //     IF OAController is not allowed to modify air flow during high outdoor humrat condition, then disable modified air flow
                     //     if indoor humrat is less than or equal to outdoor humrat
-                    if (!this->ModifyDuringHighOAMoisture && Node(this->NodeNumofHumidistatZone).HumRat <= this->OAHumRat) {
+                    if (!this->ModifyDuringHighOAMoisture &&
+                        (Node(this->NodeNumofHumidistatZone).HumRat - this->OAHumRat) <= DataHVACGlobals::SmallHumRatDiff) {
                         HighHumidityOperationFlag = false;
                     } else {
                         HighHumidityOperationFlag = true;
--- a/tst/EnergyPlus/unit/MixedAir.unit.cc
+++ b/tst/EnergyPlus/unit/MixedAir.unit.cc
@@ -1522,6 +1522,177 @@ TEST_F(EnergyPlusFixture, MixedAir_MissingHIghRHControlInputTest)
     EXPECT_TRUE(compare_err_stream(error_string, true));
 }
 
+TEST_F(EnergyPlusFixture, MixedAir_HIghRHControlTest)
+{
+    std::string const idf_objects = delimited_string({
+        "  OutdoorAir:Node,",
+        "    Outside Air Inlet Node 1; !- Name",
+
+        "  Schedule:Constant,",
+        "    Min OA Sched,             !- Name",
+        "     Any Number,              !- Schedule Type Limits Name",
+        "     0.5;                     !- Value",
+
+        "  ScheduleTypeLimits,",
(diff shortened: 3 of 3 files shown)


src/EnergyPlus/DataHVACGlobals.hh
```diff
@@ -75,6 +75,7 @@ namespace DataHVACGlobals {
 
     // MODULE PARAMETER DEFINITIONS:
 
+    Real64 const SmallHumRatDiff(1.0E-7);
```
mjwitte, line 78: Using new approach for global constants - just declare it right in the header. Better for compiler optimization instead of using `extern`. Tempted to change all of these here, but don't want to muddy the waters.

src/EnergyPlus/MixedAir.cc
```diff
@@ -4593,7 +4593,8 @@ namespace MixedAir {
                 if (ZoneSysMoistureDemand(this->HumidistatZoneNum).TotalOutputRequired < 0.0) {
                     //     IF OAController is not allowed to modify air flow during high outdoor humrat condition, then disable modified air flow
                     //     if indoor humrat is less than or equal to outdoor humrat
-                    if (!this->ModifyDuringHighOAMoisture && Node(this->NodeNumofHumidistatZone).HumRat <= this->OAHumRat) {
+                    if (!this->ModifyDuringHighOAMoisture &&
+                        (Node(this->NodeNumofHumidistatZone).HumRat - this->OAHumRat) <= DataHVACGlobals::SmallHumRatDiff) {
```
mjwitte, line 4597: Require delta humrat to be > `SmallHumRatDiff` to activate high humidity flow rate.

tst/EnergyPlus/unit/MixedAir.unit.cc
```diff
@@ -1522,6 +1522,177 @@ TEST_F(EnergyPlusFixture, MixedAir_MissingHIghRHControlInputTest)
...
+    Real64 OASignal = 0.0;
+    bool HighHumidityOperationFlag = false;
+    bool firstHVACIteration = false;
+    OAController(ControllerNum).MixSetTemp = 15.0;
+    OAController(ControllerNum).RetTemp = 20.0;
+    OAController(ControllerNum).InletTemp = 25.0;
+    OAController(ControllerNum).MixMassFlow = 2.5;
+    OAController(ControllerNum).MaxOAMassFlowRate = 2.5;
+
+    // Case 1 - OA humrat = zone humrat - no high humidity operation
```
mjwitte, line 1667: New unit test with 4 cases.

mitchute:
@mjwitte this looks good. It looks like CI is having issues again, so I'll merge this once that has a chance to get worked out.

mitchute:
Merging. Some of the expected-diff files have tanks, so make sure to pull develop into #7541 before reviewing.

PR #7650, opened 2019-12-09
Change version to 9.3

Myoldmopar:
Some of this is already done, with transition rules and support files already added to the repo.  As usual, CI will reveal which version IDs didn't get changed automatically, so I anticipate a follow up commit to finalize this before going on.


Myoldmopar:
`RefBldgMediumOfficeNew2004_Chicago_epJSON` has ERR diffs, which makes sense as my version change script is not setup to change the version of epJSON files!

3 unit tests also show version problems.  I'll fix these up and push again.

Myoldmopar:
This is looking fine, EIO diffs due to the version change, but no ERR diffs so it got all the input files now. I'm going to merge this in to get it out of the way for CI.

PR #7675, opened 2020-01-11
Fix unit test build problem on windows debug and cleanup ZoneTempPredictorCorrector unit tests

mjwitte:
Pull request overview
---------------------
 - Fixes unit test build problem on windows debug builds (VS2017[15.9.18] and VS2019[16.4.2]) introduced with #7668, and other cleanup.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [x] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #7677, opened 2020-01-12
Remove objexx gio part 2

lefticus:
Pull request overview
---------------------

AirLoopNetworkBalanceManager.cc
HeatBalanceManager.cc

have had their EIO output updated to use the new OutputFiles with libfmt.

* Documentation updates provided
* Proper rounding to match old Objexx RoundSigDigits
* 3 digit exponents provided to match Objexx RoundSigDigits


lefticus:
@Myoldmopar I just kicked the Mac build to make it rebuild. I cannot fathom why it would have so many test failures when everything else is perfect

Myoldmopar:
Yeah we had to restart the Mac this afternoon, it should be fine now.

mjwitte:
@lefticus FYI, this branch won't build unit tests locally on Windows. Same problem as #7675 showing up in many unit test files on this branch.
[unittestbuildlog.txt](https://github.com/NREL/EnergyPlus/files/4059609/unittestbuildlog.txt)

lefticus:
I'll pull from develop and make sure it's coming out clean after that.

lefticus:
@mjwitte @Myoldmopar do either of you have any insight here? I had all tests passing with no diffs on Windows last night. After merging the fixes for Windows from develop I'm now getting failures only on Windows.

Is the Windows build box known to be misbehaving?

Myoldmopar:
It wasn't known to be misbehaving, but every once in a while it does.  I wouldn't re-run CI at this point though because a branch merged into develop a few minutes ago that will then trigger lots of unrelated diffs here.  If you are completely done with this branch and ready for review, we'll review it based on the other platforms, test it locally, and if we need to we could pull develop in one more time just to confirm.

Myoldmopar:
Oh, and @lefticus, do you need any other info from @mjwitte about the unit test issue?  Last time the refactor branch got in without us realizing that it wouldn't build locally on Windows in debug mode.

lefticus:
I don't think I need anything else from @mjwitte I'd been relying on the dashboard to tell me if something was amiss, but I'll make sure I can run my own Windows builds as well from now on.

From my perspective this branch is done, and I can start working on the next set of outputs, trying to keep the merge diffs small.

mjwitte:
@lefticus This branch is not ready to go. Units tests do not build in debug config on Windows. See https://github.com/NREL/EnergyPlus/pull/7677#issuecomment-574200062 The build output file there shows which files are failing. Even when CI is running normally, it doesn't do a debug config on the windows platform, so you won't see it without doing a local build. We need to consider adding a debug build on CI. @Myoldmopar

lefticus:
@mjwitte clearly I misunderstood, I thought the point of your patches pushed to develop was to get these builds working on Windows Debug.

I'll fix it and get the PR retested

mjwitte:
@lefticus Thanks. Sorry for the confusion. The patch just fixed gio part 1. Part 2 has now touched more unit test source files. I think all you need to do is delete the `using namespace ObjexxFCL;` line in the affected files. And match the include style formatting if need be. The other changes I made over there were extra cleanup just because I was in the file and saw that.

third_party/fmt-6.1.2/include/fmt/core.h
```diff
@@ -233,7 +233,7 @@ FMT_API void assert_fail(const char* file, int line, const char* message);
 #    define FMT_ASSERT(condition, message) \
       ((condition)                         \
            ? void()                        \
-           : fmt::internal::assert_fail(__FILE__, __LINE__, (message)))
+           : ::fmt::internal::assert_fail(__FILE__, __LINE__, (message)))
```
mjwitte, line 236: @lefticus Thanks for the proper fix. Confirmed this branch now builds locally for debug config on windows.

lefticus:
@Myoldmopar your call - do I update from develop again to deal with the new diffs?

mjwitte:
@lefticus IMHO, yes, please, update from develop one more time.

Myoldmopar:
Yeah, a pull from develop wouldn't hurt

Myoldmopar:
This does look good to me.  I think I'll go ahead and drop this in as it only has the one tiny EIO diff so shouldn't interfere heavily with other builds.

lefticus:
And that tiny EIO diff was necessary to produce further ones. So great!

Issue #7680, opened 2020-01-14
Electric Chiller Forgets to Update Sometimes


Myoldmopar:
Issue overview
--------------
Possible related or duplicates:
 - #7134 
 - #6404

While refactoring the electric chiller for PlantComponent in #7637, a bunch of time was ~~spent~~ wasted debugging stray diffs in CompSetPtControl.idf.  As it turns out, even in develop that chiller model is misbehaving.  Primarily due to forgetting to update a number of what used to be module-level shared variables.  Overall, now that the chiller is refactored and cleaned up, someone needs to take a full pass to clean up all return paths to ensure variables are initialized/cleared wherever needed.  This will indeed cause diffs, so this needs to be done in a dedicated branch, not in conjunction with larger work.

This can be seen by just looking at the CompSetPtControl.idf outputs, especially how the condenser flow and delta T do not agree for some timesteps.

(FYI @mjwitte @mitchute)

rraustad:
I'm not positive on this but I think when Chiller:Electric:EIR was developed it was a mimic of Chiller:Electric. So this same symptom should be looked for in EIR chiller example files.

mitchute:
@myoldmopar I'm leaving one module-level variable on the reformulated EIR chiller refactor to be addressed here. See #7683. This looks to be related.

PR #7683, opened 2020-01-15
Chiller Electric Reformulated EIR Plant Component Refactor

mitchute:
### Pull Request Author

### Reviewer


Myoldmopar:
Yep, back to purely text-diffs.  Feel free to commit the CondOutletTemp change again now so CI will demonstrate the diffs and we can clearly point the finger at that one variable change.  Then we'll all feel better understanding that it's just a bug-fix.

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -229,7 +225,7 @@ namespace ChillerReformulatedEIR {
                                                 this->CondOutletNodeNum,
                                                 this->QCondenser,
                                                 this->CondInletTemp,
-                                                RefEIRReport(this->myIdx).CondOutletTemp,
+                                                this->CondOutletTemp,
```
Myoldmopar: It looks like there are a couple things going on that could cause the diff.  Is it due to the removal of the RefEIRReport instance?  Or the module level variable?  Are you able to split out which is the root cause?  It could make it even easier to digest the diff and verify we understand the implications.  I feel like in the end it's going to be found that the root problem was that a temp was not properly updated on every `return` path, but I'd like to see if we can nail down the exact line or code path that is causing it.  It may even help resolve #7680

Myoldmopar:
Deleted the results to re-run Windows on that last commit.

mitchute:
The most recent commits are comment- or style-only changes. The results from the most recent CI runs should be valid.

Myoldmopar:
Thanks @mitchute.  I am only like 99.8% confident in the automagic of clang-format, so I'll let CI have a pass over it to be sure.  Then I'll merge this in.

Myoldmopar:
OK, @nrel-bot is taking forever, but this is good enough for me.  The last commit was cleanup only so I really just needed to see one platform.  Anyway, merging this in; thanks for all the effort on the plant refactor, @mitchute !

PR #7679, opened 2020-01-13
Chiller Electric EIR Plant Component Refactor

mitchute:
### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #7684, opened 2020-01-16
Cleanup some lingering issues in PlantLoopEquip

Myoldmopar:
In preparation of the strip down of plant loop equip when all the plant components are done, we need to make sure each call is identical.  They are nearly done, but there were (and still are at the time of this writing) a few things to clean up.

These fixes should be absolutely no diffs or issues.


src/EnergyPlus/PlantLoopEquip.cc
```diff
@@ -699,10 +668,12 @@ namespace PlantLoopEquip {
             // for heat recovery plant interactions.
 
             if (EquipTypeNum == TypeOf_Generator_FCExhaust) {
+                // TODO: Need to set this during GetInput, not here
```
Myoldmopar: @mitchute would you want to take a crack at cleaning this up?  I'm really just trying to make every single component uniform here before we strip out this entire IF block structure.  I need these dynamic casts gone.  Can these TypeOf just be set on the components themselves during their GetInput routines?  If not there, then could they be set once at the call to the factory in PlantManager?  I see these two components share the same data structure but they have different EquipTypeNum so I know it may be a little tricky, but it seems the plant solver definitely already knows the difference here.

mitchute: You mean like the one I added over on the ElectricEIR refactor: https://github.com/NREL/EnergyPlus/pull/7679/files#diff-ae587fad7fd53d7ab16d3d93e273bcefR323
😄 

I'll take a crack at it.

Myoldmopar: :smile: well the EquipCtrlType is pretty straightforward -- once the component calls ScanPlantLoopsForObject that assignment can be made from knowledge of the loop/side/br/comp right there.  In this case, it's a little less obvious, but hopefully still pretty easy.

Myoldmopar:
Alright I cleaned up the last few issues, if CI is clean here this can get merged in.  With this in place, the refactor will be down to just 3 chillers, of which one @mitchute has done I think and I have started on a second.  So close!

src/EnergyPlus/IceThermalStorage.cc
```diff
@@ -178,13 +179,25 @@ namespace IceThermalStorage {
...
 
-    void SimpleIceStorageData::simulate(const PlantLocation &EP_UNUSED(calledFromLocation),
+    void SimpleIceStorageData::simulate(const PlantLocation &calledFromLocation,
                                         bool EP_UNUSED(FirstHVACIteration),
                                         Real64 &EP_UNUSED(CurLoad),
                                         bool RunFlag)
     {
         std::string const RoutineName("SimpleIceStorageData::simulate");
 
+        // this was happening in PlantLoopEquip before
```
mitchute, line 189: @Myoldmopar the failures are happening here after being called from SimulateWaterHeaterStandAlone since PlantLoop has not been allocated yet.

Myoldmopar: OK.  I knew that was too easy to just work right away.  Actually as I was doing this, I was noticing that RunFlag wasn't even used, so I'm not so sure this code is necessary.  I'll do further testing this morning and get it working.  Thanks!

Myoldmopar:
Alright, I cleaned up a few things that weren't using runflag anyway.  I am pretty hopeful now.

Myoldmopar:
Looking very green, with just Windows to go.  @mitchute if Windows is happy and you don't see any issues with this clean up, feel free to merge at will.

Myoldmopar:
Windows is very happy as expected, @mitchute whenever you see fit, merge away.

mitchute:
Looks good. Merging.

PR #7689, opened 2020-01-16
Remove 9-3 idd install from idd/CMakeLists.txt

Myoldmopar:
It is a generated file, we don't install it from this folder.  We install it from the src/EnergyPlus CMake rules.


Myoldmopar:
With this fix I was able to build package again, I'm going to go ahead and merge it.

PR #7695, opened 2020-01-17
Refactor Globals

mitchute:
Initial workspace to test moving global vars to a single location to support the upcoming reset-state work.


mitchute:
@Myoldmopar @brianlball per discussions today, I made a design change and moved the clear_state to the header. We should think about whether this is the right path.

nrel-bot-2b:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

src/EnergyPlus/AirLoopHVACDOAS.hh
```diff
@@ -184,23 +185,23 @@ namespace AirLoopHVACDOAS {
...
+        void SimAirLoopHVACDOAS(EnergyPlusData &state, bool const firstHVACIteration, int &CompIndex);
 
-        void initAirLoopDOAS(bool const FirstHVACIteration);
+        void initAirLoopDOAS(EnergyPlusData &state, bool const FirstHVACIteration);
 
-        void CalcAirLoopDOAS(bool const FirstHVACIteration);
+        void CalcAirLoopDOAS(EnergyPlusData &state, bool const FirstHVACIteration);
 
-        void SizingAirLoopDOAS();
+        void SizingAirLoopDOAS(EnergyPlusData &state);
```
brianlball, line 196: Add EnergyPlusData &state to all methods that need the formerly global variables.

tst/EnergyPlus/unit/Fixtures/EnergyPlusFixture.hh
```diff
@@ -254,6 +257,9 @@ protected:
                      std::vector<Real64> const &numbers,
                      std::vector<bool> const &numbers_blank);
 
+public:
+    EnergyPlusData state;
```
brianlball, line 265: state data is defined here for GTests

tst/EnergyPlus/unit/Fixtures/EnergyPlusFixture.cc
```diff
@@ -80,6 +81,7 @@ void EnergyPlusFixture::SetUpTestCase()
 
 void EnergyPlusFixture::SetUp()
 {
+    EnergyPlus::clearThisState(state);
     EnergyPlus::clearAllStates();
```
brianlball, line 87: have clearThisState(state) along side clearAllStates() until all methods are moved over

src/EnergyPlus/main.cc
```diff
@@ -56,6 +58,9 @@ int main(int argc, const char *argv[])
     // https://github.com/NREL/EnergyPlus/wiki/Debugging-Tips
     unsigned int fp_control_state = _controlfp(_EM_INEXACT, _MCW_EM);
 #endif
-    ProcessArgs(argc, argv);
-    EnergyPlusPgm();
+
+    EnergyPlusData state;
```
brianlball, line 62: Declare _EnergyPlusData state_ here at top for EnergyPlus.exe

src/EnergyPlus/test_ep_as_library.cc
```diff
@@ -69,7 +70,8 @@ int main(int argc, char *argv[])
         std::cout << "Call this with a path to run EnergyPlus as the only argument" << std::endl;
         return EXIT_FAILURE;
     } else {
-        status = RunEnergyPlus(argv[1]);
+        EnergyPlusData state;
```
brianlball, line 73: Declare EnergyPlusData state here for the test

src/EnergyPlus/api/runtime.cc
```diff
@@ -65,7 +68,9 @@ int energyplus(int argc, const char *argv[]) {
 //    argv[5] = "-i";
 //    argv[6] = epcomp->iddPath.c_str();
 //    argv[7] = epcomp->idfInputPath.c_str();
-    return runEnergyPlusAsLibrary(argc, argv);
+
+    EnergyPlusData state;
```
brianlball, line 72: declare EnergyPlusData state here for API call

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,195 @@
...
+
+#ifndef Globals_hh_INCLUDED
+#define Globals_hh_INCLUDED
+
+// EnergyPlus Headers
+#include <EnergyPlus/EnergyPlus.hh>
+
+#include <string>
+
+struct BaseGlobalStruct
```
brianlball, line 56: Base Class for struct

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,195 @@
...
+
+//struct OutputReportTabular : BaseGlobalStruct
+//{
+//    //int MaxHeaderLength;
+//
+//    void clear_state() override {
+//    }
+//};
+
+struct DataGlobal : BaseGlobalStruct
```
brianlball, line 69: struct for DataGlobal old global variables

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,195 @@
...
+    // MODULE PARAMETER DEFINITIONS
+    static constexpr int EndZoneSizingCalc = 4;
+
+    void clear_state() override {
+        AnnualSimulation = false;
+        DayOfSimChr = "0";
+    }
+};
+
+struct ExteriorEnergyUseGlobals : BaseGlobalStruct
```
brianlball: struct for ExteriorEnergyUseGlobals old global variables

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,195 @@
...
+{
+    struct ExteriorLightUsage
```
brianlball, line 88: moved this struct to here from https://github.com/NREL/EnergyPlus/blob/global/src/EnergyPlus/ExteriorEnergyUse.hh#L101-L123

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,195 @@
...
+    int NumLocalPipes = 0;
+    bool GetPipeInputFlag = true;
+
+    void clear_state() override {
+        NumLocalPipes = 0;
+        GetPipeInputFlag = true;
+    }
+};
+
+struct EnergyPlusData : BaseGlobalStruct
```
brianlball, line 164: The main data struct that contains/declares all the above structs

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -15411,10 +15412,11 @@ namespace OutputReportTabular {
             Lights(iLight).SumTimeNotZeroCons = 0.;
             Lights(iLight).SumConsumption = 0.;
         }
-        for (iLight = 1; iLight <= NumExteriorLights; ++iLight) {
```
brianlball, line 15493: this was moved to the clear_state() in EnergyPlusData.hh and then removed since it followed an ExteriorLights.deallocate()

src/EnergyPlus/DemandManager.cc
```diff
@@ -837,7 +838,7 @@ namespace DemandManager {
                     DemandMgr(MgrNum).Load.allocate(DemandMgr(MgrNum).NumOfLoads);
 
                     for (LoadNum = 1; LoadNum <= DemandMgr(MgrNum).NumOfLoads; ++LoadNum) {
-                        LoadPtr = UtilityRoutines::FindItemInList(AlphArray(LoadNum + 4), ExteriorLights);
+                        LoadPtr = UtilityRoutines::FindItemInList(AlphArray(LoadNum + 4), state.exteriorEnergyUse.ExteriorLights);
```
brianlball, line 840: need to add state.exteriorEnergyUse. since we are no longer using the old global structs/variable.

src/EnergyPlus/DemandManager.cc
```diff
@@ -666,7 +667,7 @@ namespace DemandManager {
         using DataHeatBalance::ZoneElectricObjects;
         using DataZoneControls::TempControlledZone;
         using DataZoneControls::TStatObjects;
-        using ExteriorEnergyUse::ExteriorLights;
+        //!$using ExteriorEnergyUse::ExteriorLights;
```
brianlball: remove since we are passing in the state struct instead of extern global variables,

brianlball (commented):
initial moving of global variables to struct to pass around.  design is still fluid if needed but need to lock down soon so we can start pushing it through after the release.

@lefticus thanks for discussing this with us earlier.  
@mbadams5  would like your comments as well

src/EnergyPlus/Coils/CoilCoolingDXCurveFitPerformance.cc
```diff
@@ -359,9 +360,10 @@ void CoilCoolingDXCurveFitPerformance::calculate(CoilCoolingDXCurveFitOperatingM
 
 }
 
-void CoilCoolingDXCurveFitPerformance::calcStandardRatings(int supplyFanIndex, int const supplyFanType, std::string const &supplyFanName, int condInletNodeIndex, OutputFiles &outputFiles) {
-
+void CoilCoolingDXCurveFitPerformance::calcStandardRatings(EnergyPlusData &state, int supplyFanIndex, int const supplyFanType, std::string const &supplyFanName, int condInletNodeIndex, OutputFiles &outputFiles) {
 
+    using General::SolveRoot;
+    using TempSolveRoot::SolveRoot;
```
lefticus, line 366: This `using` declaration seems unnecessary as you are explicitly scoping the call to `TempSolveRoot` later?

src/EnergyPlus/Data/EnergyPlusData.hh
lefticus: @brianlball why is there a base class? In what contexts is the base class used?

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -0,0 +1,184 @@
...
+    int NumExteriorLights = 0; // Number of Exterior Light Inputs
+    int NumExteriorEqs = 0;    // Number of Exterior Equipment Inputs
+
+    // Object Data
+    Array1D<ExteriorLightUsage> ExteriorLights;        // Structure for Exterior Light reporting
+
+    void clear_state() override {
+        NumExteriorLights = 0;
+        NumExteriorEqs = 0;
+        ExteriorLights.deallocate();
```
lefticus, line 121: What happens if `deallocate` is never called? Is that a memory leak or does `Array1D` do the right thing when the ExteriorLights are destroyed?

src/EnergyPlus/DataGlobals.hh
```diff
@@ -65,7 +66,7 @@ namespace DataGlobals {
 
     extern bool runReadVars;
     extern bool DDOnlySimulation;
-    extern bool AnnualSimulation;
+    //!$extern bool AnnualSimulation;
```
lefticus: Are are these commented out instead of just removed if they are unused?

brianlball: just left over from prototyping.  they will get removed

src/EnergyPlus/Coils/CoilCoolingDXCurveFitPerformance.cc
brianlball: yes, this is temporary and will get cleaned up once we convert all of General and can remove TempSolveRoot

src/EnergyPlus/Data/EnergyPlusData.hh
brianlball: i think its a placeholder for future needs that arrive during fleshing this out, but youre right, its not really doing anything right now is it?

lefticus:
Comments:

 * This is 99% necessary plumbing for removing global state from EnergyPlus to get it to the point of being able to support saving state, rewinding, multi-threaded and many other desirable features.
 * Given this, there is very little to actually review for most of the changes.

## Issues I'd Like To See Addressed

Most of the code is just plumbing the state to where it needs to end up, and during this process you don't know exactly which sub objects need to be plumbed. 

However, we want to reduce who has access to which parts of the global state as much as possible. This will help us identify what parts are actually used by which functions.

Therefore:

- [ ]  Pass only the sub object necessary when you can. Example, GetExteriorEnergyUseInput only uses the `exteriorEnergyUse` sub object. And doesn't forward `state` to any other functions. https://github.com/NREL/EnergyPlus/blob/global/src/EnergyPlus/ExteriorEnergyUse.cc#L166 
    * Suggestion is that the signature for `GetExteriorEnergyUseInput` should be changed from `EnergyPlusData &` to `ExteriorEnergyUseData &`
    * Plus other similar cases, when you know you can reduce the type passed to only what's necessary
- [ ] Pass by `const &` when possible
    * Currently all values are pass by `&`, and none by `const &`
    * Are there any functions which only observe the data in the state object and not mutate it? Modify these functions to take the state by `const &` when possible.
- [x] Merge `OutputFiles` and `EnergyPlusData` *where possible*.
    * Like the above, if the function only requires OutputFiles and doesn't use global state, then continue to only pass the OutputFiles object to it
    * But in the cases where both are needed, just merge it up into EnergyPlusData

## Questions

- [ ] What is `General.hh` and why does it include `EnergyPlusData.hh`?
- [ ] Why are you `using TempSolveRoot::SolveRoot;` then further in the code calling `TempSolveRoot::SolveRoot()` explicitly? Is the goal to get overload resolution with the related `General::SolveRoot()`?

## General Comments

 - [x] Tabs are sneaking into the source code! Be careful with your editor settings!
 - [ ] At some point the formatting will want to be normalized back to what the clang-format files have defined.
    Example:
    ```cpp
    func(thing1,
         thing2,
         thing3);
    ```
    has become
    ```cpp
    func(state, thing1,
         thing2,
         thing3);
    ```
- [x] In many places code that is now not used has just been commented out. If the code is no longer necessary, delete it.

Myoldmopar:
@lefticus thank you very much for your support and your review.  I'll try to address a few things:

> Pass only the sub object necessary when you can.

Absolutely.  We can include this in the current version of this branch as a prototype of limiting access to different parts of the state.

> Pass by const & when possible

Yep, another one that will be easy and a precedent should be set.  We should simply assume const until such point in the call stack that we need to modify it, and only be non-const from that point upward.

> Merge OutputFiles and EnergyPlusData where possible.

I will very much enjoy seeing the point where OutputFiles just becomes another "part" of the this new "state" of EnergyPlus.  I don't think we will address it in this current branch.  For this branch, we aren't adding any more new stuff to this extracted state, just going to clean up what we've already done.  As you are still making your output file changes, I'd like to just leave things as-is here except what we need to do, and then we can make a single effort to get OutputFiles onto the main state in one branch.  After that you can proceed to do output files and we will be able to proceed on our stuff and it shouldn't conflict too much.

> What is General.hh and why does it include EnergyPlusData.hh?

That file is an artifact of the Fortran conversion and the contents of that file could almost certainly be put into better locations.  But I'm not sure it's a problem either.

> TempSolveRoot::SolveRoot

The SolveRoot stuff is a bit hairy so I think anything in this area should just be considered a stop-gap solution that will be remedied once we have things fully fleshed out.

> (Tabs and formatting)

Uh, good call.  I wonder why the CI scripts are not finding the style problems.

> If the code is no longer necessary, delete it.

Yup, agreed, we can go ahead and delete all the commented members now that we feel good about proceeding down this path.

src/EnergyPlus/Data/EnergyPlusData.hh
Myoldmopar: I'm not sure I can answer this properly, but I'll try.

When EnergyPlus runs normally, the clear_state function is never called.  The program ends and all the resources are presumably all destroyed and the program is done.

The reasons you would want to call clear_state are:
1. You are running the unit test suite and you want to clear the state before running another unit test, or
2. You are using EnergyPlus as a library and you want to clear the state between simulations.

In either case, you want to reset the state back to the initialized state, which would primarily include zeroing variables and calling deallocate on everything on the heap, so that the next run will be able to allocate it again.

Based on those use cases, can you give me a bit more info on your concern about deallocate never being called?

Myoldmopar:
Uhh, CI looks exceedingly happy.  :green_apple: :green_book: :green_circle: :green_heart: 

I think Windows took a nap, I'll try to wake him up soon, but anyway, this looks really really good.  I like the latest changes where you took the advice of @lefticus and started passing in only portions of the state variable, not the whole thing.  That's good stuff.

kbenne:
I am not an official reviewer, but I wanted to chime in and say that I strongly support this. This opens the door for new features that could be very exciting, including:

- Save, reset, rewind state. Think about being able to stop, rewind, and replay parts of the simulation. Realtime optimization, MPC, Digital twin all come onto the table.
- Multiple, parallel simulations.
- Nicer code organization once you swallow the change of adding the argument to all of the functions. It is temporary pain for significant future reward.

Myoldmopar:
OK, this is very happy.  Thanks for those who have reviewed this.  Anyone have final thoughts?  I'll likely merge this in the morning.

src/EnergyPlus/Data/EnergyPlusData.hh
mitchute: I put that in, and my intent was to force all of the data classes to implement a ```clear_state```. There are still a handful of places in EnergyPlus where ```clear_state```s are not implemented, and I thought this may be a clean way to enforce their implementation as we proceed forward with this transition. I suppose there could be other methods we need to enforce the implementation of, too. I'm definitely open to other ideas on how to do that.

src/EnergyPlus/Data/EnergyPlusData.hh
mitchute: Building on my previous comment about ```clear_state```s. Here the ```clear_state``` is implemented but is missing most of the members in the data class. It would be great if we could automatically enforce that all variables get added here. I'm unsure how that would happen.

mitchute:
I added a couple of review comments [here](https://github.com/NREL/EnergyPlus/pull/7695/files#r417251815) and [here](https://github.com/NREL/EnergyPlus/pull/7695/files#r417253885). FYI @lefticus @Myoldmopar 

Other than that, "I'm @mitchute and I approve these *changes*."

src/EnergyPlus/Data/EnergyPlusData.hh
Myoldmopar: Yes, the idea is just to create an abstract interface for each state so we could make sure they have clear_states.  And yes, we could/will add more as we go along (`set_state`, anyone...?)

src/EnergyPlus/Data/EnergyPlusData.hh
Myoldmopar: I'm not sure I'm seeing that @mitchute.  The struct is there with members, but the only variables that hold state are the two integers and the array of lights, and those three are all reset here.  I could be wrong, let me know if you see something I'm missing.  Also, yes, having a way to ensure everything is cleared would be great, but even if we can't do that automatically, just having the code organized in these compact state structs should make it much easier to identify variables that aren't reset.

Myoldmopar:
Top notch everyone.  I'm merging this.

PR #7701, opened 2020-01-22
Variable Refrigerant Flow model COP results inaccurate when performance curve minimum does not match min PLR

### LINKED ISSUE #7628: VRF model does not correctly calculate COP when cycling below min PLR
Issue overview
--------------
Recent testing has shown that the COP calculation is incorrect when operating below min PLR.

In function CalcVRFCondenser the total cooling/heating capacity is adjusted by operating PLR. Additionally, if the system is cycling below Min PLR there is a calculation for cycling ratio which is not included here. This means the COP calculation is incorrect below min PLR.

       VRF(VRFCond).TotalCoolingCapacity = TotalCondCoolingCapacity * CoolingPLR;
       VRF(VRFCond).TotalHeatingCapacity = TotalCondHeatingCapacity * HeatingPLR;

As a test, these lines were moved below the calculation for CyclingRatio and added to the equation. Further investigation showed that the local variables CoolingPLR and HeatingPLR also needed to be updated when VRF(VRFCond).VRFCondPLR is capped at MinPLR.

        if (VRF(VRFCond).MinPLR > 0.0) {
            CyclingRatio = min(1.0, VRF(VRFCond).VRFCondPLR / VRF(VRFCond).MinPLR);
            if (VRF(VRFCond).VRFCondPLR < VRF(VRFCond).MinPLR && VRF(VRFCond).VRFCondPLR > 0.0) {
                VRF(VRFCond).VRFCondPLR = VRF(VRFCond).MinPLR;
                if (CoolingPLR > 0.0) CoolingPLR = VRF(VRFCond).MinPLR; // these also need updating
                if (HeatingPLR > 0.0) HeatingPLR = VRF(VRFCond).MinPLR;
            }
        }
        VRF(VRFCond).VRFCondCyclingRatio = CyclingRatio; // report variable for cycling rate
        VRF(VRFCond).TotalCoolingCapacity = TotalCondCoolingCapacity * CoolingPLR * CyclingRatio;
        VRF(VRFCond).TotalHeatingCapacity = TotalCondHeatingCapacity * HeatingPLR * CyclingRatio;

Also noticed in this testing was that the calculations for heat recovery are above these lines of code. This means the performance of heat recovery will also be affected since the PLR used in the performance curves has not been capped at the MinPLR yet. I.e., going in to this section of code.

        VRF(VRFCond).VRFCondPLR = max(CoolingPLR, HeatingPLR);

        HRHeatRequestFlag = any(TerminalUnitList(TUListNum).HRHeatRequest);
        HRCoolRequestFlag = any(TerminalUnitList(TUListNum).HRCoolRequest);
        HREIRFTConst = 1.0;
        Real64 HREIRAdjustment = 1.0;

        if (!DoingSizing && !WarmupFlag) {
            if (HRHeatRequestFlag && HRCoolRequestFlag) {
                a calculation below here uses VRF(VRFCond).VRFCondPLR

Since VRF(VRFCond).VRFCondPLR is used inside the heat recovery code, it is necessary to cap the operating PLR above the heat recovery code (shown here) and use this local variable inside the heat recovery code where performance curves are evaluated.

    VRF(VRFCond).VRFCondPLR = max(CoolingPLR, HeatingPLR);
    Real64 tmpVRFCondPLR = 0.0;
    if (CoolingPLR > 0.0 || HeatingPLR > 0.0) tmpVRFCondPLR = max(VRF(VRFCond).MinPLR, VRF(VRFCond).VRFCondPLR);

Defect file result:

image

### Details

### Checklist
 - [x] Defect file added ([DevSupport\7628](https://github.com/NREL/EnergyPlusDevSupport/tree/master/DefectFiles/7000s/7628))

### LINKED ISSUE #7682: VRF does not show rated COP when using unity performance curves
Issue overview
--------------
User questions COP results when performance curves are changed to equal 1. See UnMetHours post https://unmethours.com/question/42281/vrf-with-constant-curves-has-cop-10/?answer=42284#post-id-42284

Results show varying COP while it seems COP should be constant with these inputs.

And with unity part load fraction correlation curves the PLR and RTF should be equal. Maybe the issue with PLR != RTF is the min PLR input. Below min PLR there is a cycling ratio that is used to calculate RTF and above min PLR the VRF system runs the entire time step where PLR = 1 and RTF should also = 1. Something looks strange with PLR and RTF.

VRF_COP

### Details

### Checklist
 - [x] Defect file added [here](https://github.com/NREL/EnergyPlusDevSupport/blob/master/DefectFiles/7000s/7682/DOAToVRF_ConstantCurves.idf)

rraustad:
Pull request overview
---------------------
 - Fixes #7682 
 - Fixes #7628
 - COP calculation inaccurate when user input for EIRfPLR curve minimum value of X does not match VRF minimum PLR
 - Added warning for EIRfPLR curve minimum value of PLR is less than the VRF minimum PLR
 - Added checks when Cooling Capacity Modifier curve (function of temperature) output is negative and throws recurring warning error message
 - Added checks when Cooling Energy Input Ratio Modifier curve (function of temperature) output is negative and throws recurring warning error message
 - Corrected VRF min PLR limits inputs for three test files (DOASDXCOIL_wADPBFMethod*.idf)
- Added warning when the Condenser Type is WaterCooled and Heating Performance Curve Outdoor Temperature Type is WetBulbTemperature (see issue discussion). Only allow DryBulbTemperature type (EnergyPlus automatically resets to DryBulbTemperature in GetVRFInputData and throws warning message).
- [ ] Or might have to change code if outdoor WB can be used for some performance curves while water temp is used for others. Adding a check in Init seems easiest.

### Pull Request Author

### Reviewer


rraustad:
I know there is an issue here (other than the warning), just can't find the old idf to verify these changes correct the problem.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -1956,6 +2000,44 @@ namespace HVACVariableRefrigerantFlow {
...
+            }
+            if (VRF(VRFNum).HeatEIRFPLR1 > 0) {
+                CurveManager::GetCurveMinMaxValues(VRF(VRFNum).HeatEIRFPLR1, minEIRfLowPLRXInput, maxEIRfLowPLRXInput);
+                if (minEIRfLowPLRXInput > VRF(VRFNum).MinPLR) {
+                    ShowWarningError(RoutineName + cCurrentModuleObject + "=\"" + VRF(VRFNum).Name + "\", invalid");
+                    ShowContinueError("..." + cAlphaFieldNames(20) + " = " + cAlphaArgs(20) + " has out of range value.");
+                    ShowContinueError("...Curve minimum value of X = " + TrimSigDigits(minEIRfLowPLRXInput, 3) +
+                        " must be <= Minimum Heat Pump Part-Load Ratio = " + TrimSigDigits(VRF(VRFNum).MinPLR, 3) + ".");
+                    ErrorsFound = true;
+                }
```
Nigusse, line 2049: @rraustad do we need fatal out? should we provide warning only?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -1956,6 +2000,44 @@ namespace HVACVariableRefrigerantFlow {
...
+            Real64 maxEIRfLowPLRXInput = 0.0;
+
+            if (VRF(VRFNum).CoolEIRFPLR1 > 0) {
+                CurveManager::GetCurveMinMaxValues(VRF(VRFNum).CoolEIRFPLR1, minEIRfLowPLRXInput, maxEIRfLowPLRXInput);
+                if (minEIRfLowPLRXInput > VRF(VRFNum).MinPLR) {
+                    ShowWarningError(RoutineName + cCurrentModuleObject + "=\"" + VRF(VRFNum).Name + "\", invalid");
+                    ShowContinueError("..." + cAlphaFieldNames(9) + " = " + cAlphaArgs(9) + " has out of range value.");
+                    ShowContinueError("...Curve minimum value of X = " + TrimSigDigits(minEIRfLowPLRXInput, 3) +
+                                      " must be <= Minimum Heat Pump Part-Load Ratio = " + TrimSigDigits(VRF(VRFNum).MinPLR, 3) + ".");
+                    ErrorsFound = true;
```
Nigusse, line 2030: @rraustad do we need fatal out? should we provide warning only?

Nigusse:
Diffs are anticipated in the following four test suite idfs:
- _DOASDXCOIL_wADPBFMethod
- DOASDXCOIL_wADPBFMethod
- DOASDXCOIL_wADPBFMethod_NoReturnPath
- HVACTemplate-5ZoneVRF

The diffs in first three files are due to cooling and heating EIRFPLR curve minimum PLR change from 0.5 to 0.25. Similarly, the this curve output lower limits was adjusted to match the new min PLR value. The diff in the last idf is due to the "Heating Performance Curve Outdoor Temperature Type" change from "WetBulbTemperature" to "DryBulbTemperature" for water source VRF systems. Now EnergyPlus internally changes "Heating Performance Curve Outdoor Temperature Type" value from 
"WetBulbTemperature" to "DryBulbTemperature" if the CondenserType is WaterCooled.

I think, this fix needs to be incorporated to ExpandObject as well. @mjwitte for your attention.

mjwitte:
@Nigusse Is this ready for final review?

Nigusse:
@mjwitte Yes.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -577,10 +577,17 @@ namespace HVACVariableRefrigerantFlow {
                 OutdoorWetBulb = OutWetBulbTemp;
             }
         } else {
-            OutdoorDryBulb = OutDryBulbTemp;
-            OutdoorHumRat = OutHumRat;
-            OutdoorPressure = OutBaroPress;
-            OutdoorWetBulb = OutWetBulbTemp;
+            if (VRF(VRFCond).CondenserType == DataHVACGlobals::WaterCooled) {
+                OutdoorDryBulb = DataLoopNode::Node(VRF(VRFCond).CondenserNodeNum).Temp;
+                OutdoorHumRat = DataLoopNode::Node(VRF(VRFCond).CondenserNodeNum).HumRat;
```
mjwitte: Hmm, If this is water-cooled, then this is a water node, so HumRat will be zero or whatever the initial value is on all nodes. Seems better not to set this. Is it getting used somewhere for the watercooled case?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -593,6 +600,7 @@ namespace HVACVariableRefrigerantFlow {
             CondInletHumRat = PsyWFnTdbTwbPb(CondInletTemp, OutdoorWetBulb, OutdoorPressure);
         } else if (VRF(VRFCond).CondenserType == DataHVACGlobals::WaterCooled) {
             CondInletTemp = OutdoorDryBulb; // node inlet temp from above
+            OutdoorWetBulb = CondInletTemp; // for watercooled
```
mjwitte, line 596: In this case, there's a real value to set, but why set OutdoorWetBulb when it's not relevant for WaterCooled? Is this getting used somewhere for the watercooled case?

mjwitte (commented):
Ran before and after with the 7628 and 7682 defect files (from EnergyPlusDevSupport) and I get identical eso and err results?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Nigusse: Yes, this value is used somewhere down a few lines of code. The performance curve is evaluated depending on the HeatingPerformanceOATType curve but not on the condenser type. see this code below:
```         {
                auto const SELECT_CASE_var(VRF(VRFCond).HeatingPerformanceOATType);
                if (SELECT_CASE_var == DataHVACGlobals::DryBulbIndicator) {
                    TotHeatCapTempModFac = CurveValue(VRF(VRFCond).HeatCapFT, InletAirDryBulbC, CondInletTemp);
                    TotHeatEIRTempModFac = CurveValue(VRF(VRFCond).HeatEIRFT, InletAirDryBulbC, CondInletTemp);
                } else if (SELECT_CASE_var == DataHVACGlobals::WetBulbIndicator) {
                    TotHeatCapTempModFac = CurveValue(VRF(VRFCond).HeatCapFT, InletAirDryBulbC, OutdoorWetBulb);
                    TotHeatEIRTempModFac = CurveValue(VRF(VRFCond).HeatEIRFT, InletAirDryBulbC, OutdoorWetBulb);
                } else {
                    TotHeatCapTempModFac = 1.0;
                    TotHeatEIRTempModFac = 1.0;
                }
            } ```

Comment (rraustad) [reply src/EnergyPlus/HVACVariableRefrigerantFlow.cc:596]: I think that was the whole point of this issue, whenf the user entered WetBulbIndicator then the user noticed the outdoor WB was used instead of the inlet water temp. So either we didn't allow WetBulbIndicator with water cooled or we fixed it to give the correct results regardless.

File: src/EnergyPlus/HVACVariableRefrigerantFlow.cc (hunk 3 above)
Comment (Nigusse) [reply src/EnergyPlus/HVACVariableRefrigerantFlow.cc]: I do not know what value to set to. You are right, we should not set to any value. May be I should not have added the check for water cooled in the first place since the condenser inlet node number is zero. Because, that code section is executed only if the condenser node number is zero. I think, I have to revert this change.

**Comment by Nigusse** [comment]
Added a defect file with WaterCooled condenser type

**Comment by Nigusse** [comment]
@mjwitte @rraustad I did not find any diffs either with minPLR issue in the defect file.

File: src/EnergyPlus/HVACVariableRefrigerantFlow.cc (hunk 3 above)
Comment (Nigusse) [reply src/EnergyPlus/HVACVariableRefrigerantFlow.cc]: Reverted this part of the fix. No need to check for WaterCooled condenser type check if the condenser inlet node number is zero.

**Comment by mjwitte** [comment]
@Nigusse You said "Added a defect file with WaterCooled condenser type" - where will I find that?

**Comment by Nigusse** [comment]
@mjwitte it should be here "*\DefectFiles\7000s\7628".

**Comment by mjwitte** [comment]
Ok - got it. So, I see a change in the reported VRF Heat Pump Heating COP, but the cooling COP is the same?

**Comment by Nigusse** [comment]
Based on the original defect file (DOAToVRF_ConstantCurves.idf) I found no COP value diffs between develop and this branch. Also comfirmed the hand calculated values agree with the reported COP. Then, I moved to the other issues. See my plots below for COP comparison. 

 
COP_Comparison

**Comment by Nigusse** [comment]
The new defect file I uploaded (HVACTemplate-5ZoneVRF_Expanded.idf) was meant to verify the new warning message added when the Condenser Type is WaterCooled and Heating Performance Curve Outdoor Temperature Type is WetBulbTemperature. @mjwitte which defect file was showing heating COP diffs? I can look at it?

**Comment by Nigusse** [comment]
I also run defect file (7628 - VariableRefrigerantFlow_5Zone_COPIssue.idf) but did not find any diffs. See heating and cooling COP comparison. They are identical.  The hand calculated value also agree within error margin of 2.0E-14.

COP_Comparison_DefectFile2

**Comment by Nigusse** [comment]
@mjwitte I just pushed the branch after merging in develop.

**Comment by mjwitte** [comment]
@Nigusse Sorry - heating COP diffs were in the new file, 7628/HVACTemplate-5ZoneVRF_Expanded.idf. And is has the new warning:
```
   ** Severe  ** AirConditioner:VariableRefrigerantFlow = VRF SYS 1 WATER SOURCE VRF HEAT PUMP
   **   ~~~   ** Condenser Type = WATERCOOLED
   **   ~~~   ** Illegal Heating Performance Curve Outdoor Temperature Type input for this object = WETBULBTEMPERATURE
   **   ~~~   ** ... input must be DRYBULBTEMPERATURE when Condenser Type is WaterCooled.
   **   ~~~   ** ... Heating Performance Curve Outdoor Temperature Type will be reset to DRYBULBTEMPERATURE and simulation continues.
```
So it makes sense that the heating COPs would change. I didn't realize until now that the cooling side does not have a field for "Performance Curve Outdoor Temperature Type" so nothing changes on the cooling side.

**Comment by mjwitte** [comment]
Using example file DOASDXCOIL_wADPBFMethod.idf (original version), I can see the new warning about min curve values.
The modifications to DOASDXCOIL_wADPBFMethod (and the other changed idfs) result in big diffs in power consumption - I assume that's expected.
HVACTemplate-5ZoneVRF.idf is now throwing the warning about using the wrong curve temperature. I'll push up a quick fix to expandobjects here, merge in develop and hopefully we can call it a week here.

**Comment by mjwitte** [comment]
After fixing expandobjects to use the correct heating curve variable type for aircooled vs watercooled, the new warning is gone. There are still warnings like this:
```
   ** Warning ** GetVRFInput: AirConditioner:VariableRefrigerantFlow="VRF SYS 1 WATER SOURCE VRF HEAT PUMP" curve values
   **   ~~~   ** ... Cooling Capacity Ratio Modifier Function of Low Temperature Curve Name = VRF SYS 1 WATER SOURCE VRFCOOLCAPFT output is not equal to 1.0 (+ or - 10%) at rated conditions.
   **   ~~~   ** ... Curve output at rated conditions = 1.136
```
But those have been there a long time, and VariableRefrigerantFlow_5Zone has similar warnings. Fixing those curves can be for another day.

Myoldmopar:
@rraustad @Nigusse @mjwitte this looks pretty close, if not ready.  I see a few files with diffs that have been there for quite a while on this branch, and the EDD changes are unrelated.  The one comment that sticks out to me as unresolved is:

> The modifications to DOASDXCOIL_wADPBFMethod (and the other changed idfs) result in big diffs in power consumption - I assume that's expected.

If we can wrap up that comment, is there anything else on this one?

mjwitte (approved):

PR #7702, opened 2020-01-22
Addressed CoolingTower:VariableSpeed control problems

### LINKED ISSUE #7464: CoolingTower:VariableSpeed control problems
Issue overview
--------------
User file with CoolingTower:VariableSpeed not delivering setpoint even though the load on the tower is well below capacity. The tower fan is running at full flow, but the outlet temp is well above setpoint. Not sure if it's a design input issue or a control problem in the tower algorithm. Problems shows up on 08/12 and other days.

The tower is also putting out lots of warnings, most of which say they are expected or may not affect performance. So, why issue the warnings then?
```
   ** Warning ** CoolingTower:VariableSpeed "COOLING TOWER VARIABLE SPEED 1" - Water flow rate ratio is outside model boundaries at 0.14.
   **   ~~~   **  ...Valid range = 0.75 to 1.25. Occurrence info = RUN PERIOD 1, 08/01 00:00 - 00:10
   **   ~~~   **  ...Water flow rate ratio passed to the model = 0.75000
   **   ~~~   **  ...Water flow rate ratios outside model boundaries may not adversely affect tower performance.
   ** Warning ** CoolingTower:VariableSpeed "COOLING TOWER VARIABLE SPEED 1" - Inlet air wet-bulb temperature is outside model boundaries at 26.76.
   **   ~~~   **  ...Valid range = -1.00 to 26.67. Occurrence info = RUN PERIOD 1, 08/01 14:00 - 14:10
   **   ~~~   **  ...Inlet air wet-bulb temperature passed to the model = 26.666700
   **   ~~~   **  ...Wet-bulb temperatures outside model boundaries may not adversely affect tower performance.
   ** Warning ** CoolingTower:VariableSpeed "COOLING TOWER VARIABLE SPEED 1" - Cooling tower air flow rate ratio calculation failed 
   **   ~~~   ** ...with conditions as Twb = 26.67, Trange =  6.47, Tapproach =  2.54, and water flow rate ratio =  0.75
   **   ~~~   ** ...a solution could not be found within the valid range of air flow rate ratios
   **   ~~~   **  ...Valid air flow rate ratio range =  0.20 to 1.0.
   ** Warning ** CoolingTower:VariableSpeed "COOLING TOWER VARIABLE SPEED 1" - Tower range temperature is outside model boundaries at 12.50.
   **   ~~~   **  ...Valid range = 1.11 to 11.11. Occurrence info = RUN PERIOD 1, 08/12 03:40 - 03:50
   **   ~~~   **  ...Tower range temperature passed to the model = 12.49717
   **   ~~~   **  ...Range temperatures outside model boundaries may not adversely affect tower performance.
   **   ~~~   **  ...This is not an unexpected occurrence when simulating actual conditions.
   ** Warning ** CoolingTower:VariableSpeed "COOLING TOWER VARIABLE SPEED 1" - Tower approach temperature is outside model boundaries at 11.12.
   **   ~~~   **  ...Valid range = 1.11 to 11.11. Occurrence info = RUN PERIOD 1, 08/12 03:40 - 03:50
   **   ~~~   **  ...Tower approach temperature passed to the model = 11.11685
   **   ~~~   **  ...Approach temperatures outside model boundaries may not adversely affect tower performance.
   **   ~~~   **  ...This is not an unexpected occurrence when simulating actual conditions.
```

### Details
 - Version of EnergyPlus 9.2.0-6150bb56e0
 - Unmethours [39364](https://unmethours.com/question/39364/cooling-tower-out-of-model-boundaries-e)

### Checklist
 - [x] Defect file attached, using USA_TX_Austin-Mueller.Muni.AP.722540_TMY3 weather
[in-moreoutput-2xTower.idf.txt](https://github.com/NREL/EnergyPlus/files/3521721/in-moreoutput-2xTower.idf.txt)

Nigusse:
Pull request overview
---------------------
 - Fixes #7464
 - This fix increased the upper bound of the solution domain of variable speed cooling tower range from 11.1 to 22.22 deltaC (the same as the YorkCalc model). Otherwise, the root solver will not be able to find a solution if the maximum cooling tower range is limited to 11.1. Increasing the maximum value does not mean that the actual tower range determined by the solver will be that high. In fact for this defect file, the actual range  determined for the annual simulation was always below 10.0 deltaC. Some of the 12 test suite files which has variable speed cooling tower could show diffs. 

### Pull Request Author

### Reviewer
 - [ ] I~f new idf included, locally check the err file and other outputs~


Nigusse:
Figure below shows variable speed (VS) cooling tower entering and leaving water temperature, cooling tower entering air wetbulb temperature and tower range (Tr) of develop code. The tower water outlet node setpoint temperature is 30C.  The VS cooling tower is autosized and uses a sizing factor of 2.0. However, in several occasions in January and August months the cooling tower leaving temperature exceeds the setpoint temperature of 30C by more than10 deltaC. Observations:

 - sometimes the cooling tower is not providing enough cooling even if the tower fan is running at full flow (as shown below) and the cooling tower leaving water temperature is much higher than the setpoint (30C)
 - the tower entering water temperature is building up for days.  This happens during heating and cooling seasons
- note that tower range (Tr)[detlaC] maximum is capped at 11.11 (VS tower default curve maximum value)


Develop

rraustad:
Compare the tower outlet water temperature to the tower entering air wet-bulb temperature. The outlet water temperature cannot go below the tower entering air wet-bulb temperature.

Nigusse:
See the updated plot. Now I uploaded the plots that shows the entering water and air wetbulb temperature.

Nigusse:
@rraustad This problem exits even if the tower entering air wetbulb temperature is well below the tower outlet setpoint temperature.

rraustad:
OK, something looks wrong. Add tower air flow and fan power to see when the tower is active.

Nigusse:
Figure below of this fix demonstrates that the VS cooling tower always meets the tower leaving water setpoint temperature. Summary of this fix:

- now the cooling tower provides enough cooling all the time (as shown in figure below), as the result the cooling tower leaving water temperature is always at or below the desired setpoint of 30C. This is true for annual run as well.
- the tower entering water temperature is modest, i.e., not as high as the defect code simply because the cooling tower provdes the required cooling.
- note that cooling tower range (Tr) never reaches the default performance curve maximum value of 11.11 detlaC. Annual run check also shows that the actual tower range (Tr) was always less than 7.3 [detlaC] for this defect file.
 

Branch#7464

rraustad:
Yes, this looks better. Why is the tower outlet water temperature (blue line) always below the tower water inlet temperature (yellow line) in winter? Isn't the tower off?

Nigusse:
See below the cooling tower fan electric power, fan part load ratio and air flow ratio for develop and this branch (the fix). Note that develop shows that the cooling tower does not meet the setpoint even if the fan is running at full flow. 

Branch_VSCTFan

Nigusse:
@rraustad A boiler and the cooling tower are placed in series on the supply branch. The boiler is designed to maintain supply temperature of 20C. I suspect, may be the cooling tower is running in free cooling mode (with the fan off).

rraustad:
If the tower has water flow and no fan power during this time then that would explain it.

rraustad:
Cooling towers do have bypass valves. I wonder if this would be a feature users would need for more accurate modeling?

Nigusse:
@mjwitte @rraustad @Myoldmopar One of the two problems Mike indicated in issue #7464 was "putting out lots of warnings, most of which say they are expected or may not affect performance". 

One way to address these warning issues is suppress them unless the user requires them by using the **DisplayExtraWarnings** option. These warning error messages are about these four variables:

 - Inlet air wet-bulb temperature (checked against towers.MinInletAirWBTemp and towers.MaxInletAirWBTemp)
 - Tower Approach temperatures (checked against towers.MinApproachTemp and towers.MaxApproachTemp)
 - Tower range temperature (checked against towers.MinRangeTemp and towers.MaxRangeTemp)
 - Water flow rate ratio (checked against towers.MinWaterFlowRatio and towers.MaxWaterFlowRatio)

User has no control over these variables.  Therefore, I suggest that we suppress them using the **DisplayExtraWarnings** option.  What do you think?  Any other alternative approach?

mjwitte:
@Nigusse I'm ok with using DisplayExtraWarnings for these. Or maybe warn once and use recurring warnings? The user does have access to these inputs if using one of the userdefined performance options, but not with the built-in options. If I'm following correctly, these warnings indicate that the tower is operating outside the stated validity range of the coefficients? Which is something we don't do with most components. 

So, DisaplyExtraWarnings *and* make it recurring as well so it doesn't generate a huge err file.

mjwitte:
@Nigusse So, what did you conclude about the tower overcooling? FluidBypass is a capacity control option for the singlespeed and twospeed towers. So, I guess it's not an option for this tower. But this tower does have Cell options, so in theory if you take the defect file and increase the number of cells with MinimalCell control, it should never go below setpoint.

Nigusse:
@mjwitte currently all four variables have recurring warning message but **Water flow rate ratios** and **Inlet air wet-bulb temperature* report five instances before switching to recurring warning reporting mode. I can change only the two variables to switch to recurring warning reporting mode after allowing to report only once, instead of 5 times. You are correct, these warning indicate that the tower is operating outside the stated validity range.

mjwitte:
5 instances is OK, I didn't realize they were already set up that way.

Nigusse:
@mjwitte the tower did not go below it's owns set point, instead it is going below the boiler set point because of the free cooling as it does not have the bypass option. Also in this defect file the boiler is placed before the tower in series on the same branch. I do not know if increasing the cell number makes any difference but I did not check, I will run the defect file with higher number of cells and will let you know my findings.

mjwitte:
I'm confused, the tower is in series before the boiler and the tower has a lower setpoint than the boiler? Seems like the two should never fight with each other. Does the tower not have a way to simply pass through the water if it's not being given a load?

Nigusse:
Yes, the boiler and the tower are in series. The boiler set point is 20C and the cooling tower set point is 30C. The boiler controls the loop lower temperature limit and the cooling tower controls the loops upper temperature limit. This loop is connected to hundreds of water source heat pumps.

Nigusse:
I also verified that the cooling tower outlet temperature drops slightly below the set point of 20C when the boiler is ON because of the free cooling provided by the tower. I set the **Fraction of Tower Capacity in Free Convection Regime** input field to zero to confirm this. See below the updated plot:

Branch#7464NoFreeCooling

Nigusse:
Also increasing the number cells from 1 to 4 did not make any difference on the impact of free cooling when the boiler is ON.

mjwitte:
@Nigusse Your plots aren't displaying here for some reason. If the cooling tower set point is 30C, then it shouldn't be putting out water at 20C. It seems wrong to me that it would be doing any free cooling here. What would the other cooling towers do in this instance? Perhaps the plant should be set up so the tower and boiler are in parallel so there would only be flow in one or the other but never both?

Nigusse:
@mjwitte I agree it is preferable to have a parallel configuration of these two equipment. Here we are dealing with a defect file that has boiler and cooling tower in series. Thus the boiler heats the water that goes through it and maintains loop temp at 20C, then that water has to go through the tower and the free cooling operation is unavoidable for this configuration. But E+ allows such configuration. What we can do is to discourage users from modeling such configuration with the VS cooling tower, I think.

Nigusse:
Uploaded the plot again:

Branch#7464NoFreeCooling

mjwitte:
My browsers think github image pages are insecure so I'm not seeing any at the moment. Doesn't matter really. 

This tower needs a bypass control option (new issue). And the docs for the current tower need to explain this behavior if they don't already (I/O Ref). I'll stop harping on this.

Nigusse:
Diffs are expected in this branch. The defect file has demonstrated the VS cooling tower fix does better control of meeting the tower cooling setpoint. The tower fan runs at part load, or even cycle to meet the required load and hence the tower fan electric consumption can change significantly. Examining the ASHRAE9012016_Hospital_Denver test suite idf shows big meter absolute diffs but the percent (relative) diffs are small (in the order of 0.02% and -0.01% for annual metered facility electric and gas energy consumption, respectively).

nrel-bot-2b:

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -5299,7 +5300,7 @@ namespace CondenserLoopTowers {
         Par(4) = Twb;                // inlet air wet-bulb temperature [C]
         Real64 Tr;                   // range temperature which results in an energy balance
         auto f = std::bind(&CoolingTower::residualTr, this, std::placeholders::_1, std::placeholders::_2);
-        General::SolveRoot(Acc, MaxIte, SolFla, Tr, f, 0.001, towers(this->VSTower).MaxRangeTemp, Par);
+        General::SolveRoot(Acc, MaxIte, SolFla, Tr, f, 0.001, VSTowerMaxRangeTemp, Par);
```
mjwitte, line 5303: @Nigusse Is this overriding the range for all tower types?
Rather than overriding this here, why not set `towers(this->VSTower).MaxRangeTemp` up in `GetTowerInput`, [here](https://github.com/NREL/EnergyPlus/blob/c9d665d7d94dda642053ae68f14c734b0a502bbd/src/EnergyPlus/CondenserLoopTowers.cc#L1164).

doc/input-output-reference/src/overview/group-plant-condenser-control.tex
```diff
@@ -299,6 +299,10 @@ \subsection{PlantEquipmentOperation:ComponentSetpoint}\label{plantequipmentopera
 
 In addition to load range based control on the plant loop, sequencing the plant components based on the outlet temperature of individual equipment is allowed. This scheme is common to many present-day Energy Management Systems sequencing. In this scheme, the sequencing is done based on the order specified in the control object described below.
 
+However, when sequencing a boiler and a cooling tower to control the heating and cooling set point temperatures of a condenser loop, it is recommended that the boiler and the cooling tower placed in parallel configuration. The series configuration of these two equipment, though possible to model them, may result in simultaneous heating and cooling operation when the boiler is ON and the cooling tower operates in free cooling mode, i.e.,
```
mjwitte: Minor "boiler and the cooling tower *be* placed in *a* parallel configuration."

src/EnergyPlus/CondenserLoopTowers.cc
Nigusse: Yes, the 22.2222 is used for all variable speed cooling towers. I can set it but the default value is used elsewhere to check if the actual tower range is outside the default range. I am concerned to change any curve values that are related the data used to generate the coefficients. And one reason that I chose  22.2222 is that it is the same as the VS York CT so that it does not affect the York Model; otherwise, I experimented with a value higher than that. If you still insist, I will make the changes per your suggestion.

doc/input-output-reference/src/overview/group-plant-condenser-control.tex
Nigusse: Done!

src/EnergyPlus/CondenserLoopTowers.cc
Nigusse: @mjwitte Below is alternative suggestion for your consideration. First, this change impacts the toolcool VS cooling tower model only; second, this model will be able to check and throw warning if the requested tower range is outside the VS cooltool existing (default) tower range maximum limit.

```
        VSTowerMaxRangeTemp = towers(this->VSTower).MaxRangeTemp;
        if (this->TowerModelType == CoolToolsXFModel) {
            VSTowerMaxRangeTemp = 22.2222; // this relaxes the solution domain
        }
```

mjwitte: Nah, let's leave it like it is. Please push up the doc changes and we'll call this complete.

mjwitte (approved):
Tested with the defect file before and after (with Austin TX weather).

The tower still generates an excessive amount of warnings, but at least it's putting out the requested water temperature now.

src/EnergyPlus/CondenserLoopTowers.cc
Nigusse: Done. Thanks!

PR #7703, opened 2020-01-23
Collapse PlantLoopEquip nested IF block

Myoldmopar:
If CI agrees with me, simply delete 551 lines, with no diffs, and hope for a minimal performance bump.


mitchute:
@Myoldmopar this looks good! Is it ready to go?

Myoldmopar:
Yeah it’s fine, I was definitely withholding optimism about seeing performance gains on CI. Those files have a lot of other stuff going on besides plant, and since they don't run annual you don't see the benefit of the gains happening throughout the year. All that said, I was seeing a few percent improved consistently on local tests running annually. 

Mostly there are no diffs so this is a clean change. Merge away.

PR #7704, opened 2020-01-23
DX Cooling Coil Refactor

Myoldmopar:
This pull request introduces a new object: `Coil:Cooling:DX`, along with supporting "child" objects.

In the longer term, this coil will replace the following currently implemented coils:

- Coil:Cooling:DX:SingleSpeed
- Coil:Cooling:DX:TwoSpeed
- Coil:Cooling:DX:VariableSpeed
- Coil:Cooling:DX:MultiSpeed
- Coil:Cooling:DX:TwoStageWithHumidityControl

However, in the current pull request, the old coils will be left in place.  There is a very large number of parent objects which support these cooling coils in varying levels.  It is not practical to replace all these coils and also change every parent object - all at the same time.

For this pull request, the unitary system will gain support for this new coil, and our suite of unitary example files will be updated to use the new coil.  The transition tools will _not_ automatically convert user IDFs to the new coil.  This will allow us to exercise the coil in many ways before release, without just throwing it out there for all unknown corner cases.  We will create a custom transition binary setup to allow power users to modify their IDFs to the new coil if desired. (It won't change version number so that their IDF will continue to be a valid 9.3 IDF.)

I'll come up with a custom task list here:
 - Verify IDD is completely settled down -- this is the primary need for this PR before IO freeze.

We'll see what CI says in this initial commit.  I am hopeful it is generally clean, but some unitary diffs will be present as we work through those.  I won't be surprised if a unitary failure shows up initially.


Myoldmopar:
Closing this for today as we are expecting a lot of commits to be pushed and I don't want CI trying to run all of them...

Issue #7590, opened 2019-10-30
EIR WWHP Electric Input to Output Ratio Modifier Function of Part Load Ratio Curve Name is ignored


mjwitte:
Issue overview
--------------
For both HeatPump:WaterToWater:EIR:Heating (and Cooling)  the"Electric Input to Output Ratio Modifier Function of Part Load Ratio Curve Name" is ignored. The "Electric Input to Output Ratio Modifier Function of Temperature Curve Name" get used for both the EIR fT and EIRfPLR curves.

The offending pasteo in the code is [here](https://github.com/NREL/EnergyPlus/blob/0e6e9c08a048eb9bb04387a61384bd842816c52f/src/EnergyPlus/WaterToWaterHeatPumpEIR.cc#L1016-L1028).

And @mitchute the offending code is still [here](https://github.com/NREL/EnergyPlus/pull/7489/files#diff-d5cec73fe6ecb0d05b1d13a456017641R1065) in #7489 

### Workaround
With this bug, the power consumption is being computes as 
(loadSideHeatTransfer  / referenceCOP) * EIRfT(loadSideOutletTemp, sourceSideInletTemp) * **EIRfT**(partLoadRatio). 

I suppose if one were to make a table object with appropriate output values for inputs in the range of 0 to 1, this might work, as long as the temperature variables are never expected to operate near freezing.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus v9.2.0 (and 9.1)
 - Unmethours [40787](https://unmethours.com/question/40787/heatpumpwatertowatereir-models-curves-report-wrong-values/)or helpdesk ticket number 15300

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mitchute:
Addressed by #7714. Closing.

PR #7714, opened 2020-01-24
Fix HeatPump:PlantLoop:EIR:Heating/Cooling PLR Curve Name

mitchute:
This PR addresses #7590. The EIR-as-a-function-of-PLR curve was incorrectly referencing the EIR-fn-temperature curve name.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/PlantLoopHeatPumpEIR.cc
+++ b/src/EnergyPlus/PlantLoopHeatPumpEIR.cc
@@ -1064,7 +1064,7 @@ namespace EIRPlantLoopHeatPumps {
                                         "; entered curve name: " + capEIRtName.get<std::string>());
                         errorsFound = true;
                     }
-                    auto &capEIRplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_temperature_curve_name");
+                    auto &capEIRplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_part_load_ratio_curve_name");
                     thisPLHP.powerRatioFuncPLRCurveIndex = CurveManager::GetCurveIndex(UtilityRoutines::MakeUPPERCase(capEIRplrName));
                     if (thisPLHP.capFuncTempCurveIndex == 0) {
                         ShowSevereError("Invalid curve name for EIR PLHP (name=" + thisPLHP.name +
--- a/src/EnergyPlus/PlantLoopHeatPumpEIR.cc
+++ b/src/EnergyPlus/PlantLoopHeatPumpEIR.cc
@@ -1057,18 +1057,18 @@ namespace EIRPlantLoopHeatPumps {
                                         "; entered curve name: " + capFtName.get<std::string>());
                         errorsFound = true;
                     }
-                    auto &capEIRtName = fields.at("electric_input_to_output_ratio_modifier_function_of_temperature_curve_name");
-                    thisPLHP.powerRatioFuncTempCurveIndex = CurveManager::GetCurveIndex(UtilityRoutines::MakeUPPERCase(capEIRtName));
+                    auto &eirFtName = fields.at("electric_input_to_output_ratio_modifier_function_of_temperature_curve_name");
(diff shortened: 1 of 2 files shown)


mitchute:
It looks like there are only two test files that use these objects.... and the EIRfnTemp and EIRfnPLR curves are exactly the same. So, I'm not expecting any diffs.

mjwitte:
"the EIRfnTemp and EIRfnPLR curves are exactly the same" that seems odd - isn't one bivariate and the other univariate?

mitchute:
@mjwitte yes, that's correct. However, the coefficients are such that they produce the same result.

EIRfnTemp: https://github.com/NREL/EnergyPlus/blob/develop/testfiles/PlantLoopHeatPump_EIR_WaterSource.idf#L290

EIRfnPLR: https://github.com/NREL/EnergyPlus/blob/develop/testfiles/PlantLoopHeatPump_EIR_WaterSource.idf#L308

Or maybe I'm just missing something.

src/EnergyPlus/PlantLoopHeatPumpEIR.cc
```diff
@@ -1064,7 +1064,7 @@ namespace EIRPlantLoopHeatPumps {
                                         "; entered curve name: " + capEIRtName.get<std::string>());
                         errorsFound = true;
                     }
-                    auto &capEIRplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_temperature_curve_name");
+                    auto &capEIRplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_part_load_ratio_curve_name");
```
jmarrec: Good catch. I'd take the opportunity to. rename this variable and the ones next to it as appropriate to be correct by themselves.
This `capEIRplrName` should be `eirFplrName` or something. Above `capEIRtName` should be `eirFtName`. ("Cap" shouldn't be part of the name, since it's not a capacity modifier curve).

jmarrec (approved):
Looks good to me, except minor cosmetic changes that should take a minute to address (these are local scoped variables)

src/EnergyPlus/PlantLoopHeatPumpEIR.cc
mitchute: This has been addressed in the latest commit.

jmarrec: Great. I guess a unit test is needed too.

mitchute: Ready to go here.

src/EnergyPlus/PlantLoopHeatPumpEIR.cc
```diff
@@ -1057,18 +1057,18 @@ namespace EIRPlantLoopHeatPumps {
...
                         errorsFound = true;
                     }
-                    auto &capEIRplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_temperature_curve_name");
-                    thisPLHP.powerRatioFuncPLRCurveIndex = CurveManager::GetCurveIndex(UtilityRoutines::MakeUPPERCase(capEIRplrName));
+                    auto &eirFplrName = fields.at("electric_input_to_output_ratio_modifier_function_of_part_load_ratio_curve_name");
+                    thisPLHP.powerRatioFuncPLRCurveIndex = CurveManager::GetCurveIndex(UtilityRoutines::MakeUPPERCase(eirFplrName));
                     if (thisPLHP.capFuncTempCurveIndex == 0) {
                         ShowSevereError("Invalid curve name for EIR PLHP (name=" + thisPLHP.name +
-                                        "; entered curve name: " + capEIRplrName.get<std::string>());
+                                        "; entered curve name: " + eirFplrName.get<std::string>());
```
Myoldmopar, line 1071: Good fixes here, thanks @mitchute

Myoldmopar (approved):
Good to go, thanks @mitchute

PR #7715, opened 2020-01-24
Fix up a couple paper cuts

Myoldmopar:
A couple tiny things are causing the dashboard to be more yellow than usual.
1. An uninitialized variable warning is appearing on Mac.  I believe the variable is actually initialized, the compiler just doesn't see that the uninitialized paths are not actually going to continue.  Anyway, I initialized it to zero, I think it will be perfectly fine.
2. The integration coverage dipped very slightly below the warn limit.  This is because over the past couple months, we've been refactoring plant components and reduced the number of lines in each object usually quite significantly.  Since these components are covered by varying amounts of unit tests, we ended up dropping the number of covered lines.  I just dropped the limit very slightly so it won't warn anymore.

CI should come back clean with no diffs due to the initialized variable thing and all green on integration coverage as well. -- except for the ever-present timeout file, that is.  If that's all good, this can drop in.


mitchute:
Thanks for the fix, @Myoldmopar. Merging.

PR #7716, opened 2020-01-27
Remove objexx gio  3

lefticus:
Pull request overview
---------------------
Further work on removal of gio for eio files.

This one has 2 eio diffs:

1) one particular value is printed as 1.0000E-05 instead of 10.0000E-006. This seems to be a bug in the original Objexx code as it is not consistent with any other output
2) There is a rounding up occurring in one particular case which did not occur in the original code. I have double checked the value and it should be round up. This again seems to be a bug in the original code.


tst/EnergyPlus/unit/OutputFiles.unit.cc
```diff
@@ -0,0 +1,144 @@
...
+    EXPECT_EQ(print_to_string("{:20.8T}", 0.0), "       0.0000000    ");
+    EXPECT_EQ(print_to_string("{:20.8T}", 2.13608134), "       2.1360813    ");
+    EXPECT_EQ(print_to_string("{:20.8T}", 213608134.0), "      213608134.    ");
+    EXPECT_EQ(print_to_string("{:20.8T}", 213608139.6), "      213608140.    ");
+    EXPECT_EQ(print_to_string("{:20.8T}", 0.213608134), "      0.21360813    ");
+    EXPECT_EQ(print_to_string("{:13.6T}", 0.803434E+09), " 0.803434E+09");
+    //    EXPECT_EQ(print_to_string("{:20.8T}", -0.23111252), "     -0.23111252    ");
+    //    EXPECT_EQ(print_to_string("{:20.8T}", -0.23111252), "     -0.23111252    ");
+}
+} // namespace EnergyPlus
```
Myoldmopar, line 144: Fantastic!

src/EnergyPlus/GroundTemperatureModeling/FiniteDifferenceGroundTemperatureModel.hh
```diff
@@ -156,11 +155,11 @@ public:
...
     void doStartOfTimeStepInits();
 
-    Real64 getGroundTemp();
+    Real64 getGroundTemp() override;
 
-    Real64 getGroundTempAtTimeInSeconds(Real64 const depth, Real64 const timeInSecondsOfSim);
+    Real64 getGroundTempAtTimeInSeconds(Real64 const depth, Real64 const timeInSecondsOfSim) override;
 
-    Real64 getGroundTempAtTimeInMonths(Real64 const depth, int const monthOfSim);
+    Real64 getGroundTempAtTimeInMonths(Real64 const depth, int const monthOfSim) override;
```
Myoldmopar, line 162: Nice misc cleanups along the way.

Myoldmopar (approved):
I have no issues here, this is looking great.

src/EnergyPlus/GroundTemperatureModeling/FiniteDifferenceGroundTemperatureModel.hh
lefticus: I'm trying really hard to limit my changes to things that only directly affect the output, but figured I needed to do that when I saw it.

Myoldmopar:
This is all clean.  The CI changes are exceedingly reasonable.  Thanks.  Merging.

PR #7731, opened 2020-01-30
Correction of Zero Solar Load in the Zone Component Load Summary Report

### LINKED ISSUE #7608: Zone component load summary reports zero solar load when used with SizingPeriod:WeatherFileDays
Issue overview
--------------
User file shows zero for Fenestration Solar in the Zone Component Load Summary report when used with SizingPeriod:WeatherFileDays. Switching to regular design days results in Fenestration Solar loads on the cooling day.

### Details
 - Version of EnergyPlus v9.2.0
 - Helpdesk ticket number 15302.

### Workaround
Use SizingPeriod:DesignDay instead of SizingPeriod:WeatherFileDays to generate the component loads report. Revert back to SizingPeriod:WeatherFileDays for final results.

### Checklist
 - [ ] Defect file added to EnergyPlusDevSupport\DefectFiles, run with Chicago weather

RKStrand:
Pull request overview
---------------------
 - Fixes #7608 
 - Corrects the zero solar load as well as other zero loads that show up in the delayed column.  The problem was that decay curves were not being calculated when the user selected design days from the weather file period.  This was due to the fact that the delay curves were based on factors that were only calculated on the first day of the simulation.  This changes things so that those factors are calculated on every day and thus will be available for whatever day is selected as the design day.  Values are now reported as non-zero when appropriate.

### Pull Request Author: Rick Strand

### Reviewer


JasonGlazer:
@RKStrand I am not sure I agree with your solution since the calculation of the decay curves is time consuming, recalculating them each day could substantially increase the calculation time for a simulation using them. Have you looked at the performance impact?

RKStrand:
@JasonGlazer I'll agree that the solution does not seem ideal but I wasn't sure there was another option.  When one is looking at a design day, it's pretty obvious that it's just the first day of the solution.  When the user asks for a design day from the weather file though, one doesn't know beforehand which day of the year that will be.  Or is there a way to figure that out before doing the simulation for that weather environment.  That's where I got stuck and felt like the only way to guarantee that it would do this was doing the calculation every day.  Maybe I missed something in the logic though and I'm happy to alter this if I misunderstood something.  I have not looked at the performance impact but when I looked at the code I thought that the number of calculations wasn't that large.  But perhaps I misunderstood that also?

JasonGlazer:
There are some nested loops of calculations so that is why a believe they are expensive. 

Maybe make the logic different if the user is using design days vs the weather year?

Even if using a weather day, if it is not setting the peak cooling or peak heating that day, then these calculations are not needed.

RKStrand:
@JasonGlazer So, I have altered the logic of when to calculate the pulse.  Now, it only calculates it on the first day of the simulation (as before) OR when the user selects to run the weather file as the sizing period.  If the user has selected the weather file as the sizing period, then it will still calculate these every day because it seems to me that it does not calculate which day of the year should be the design day a priori but rather on the fly.  Is that a correct understanding of the code or am I missing something?

From what I have seen, it seems to indicate that it doesn't know whether or not a particular day is going to be the design day yielding maximum loads until it has actually done the day.  So, doesn't that mean it has to calculate these pulse factors every day just in case a particular day is the actual day for sizing?

With the change I am proposing (haven't put it on GitHub yet but will soon), it only does the calculations every day when the user wants to use the entire year as the design day.  By doing this, they are already willing to wait longer for results.  So, I feel like a slight performance penalty is probably not a huge deal here, and I'm not sure there is any way around that.

What are your thoughts on this strategy?

JasonGlazer:
Rick, that sounds reasonable to me.

RKStrand:
Thanks, @JasonGlazer.  I am currently making sure that energyplus_test runs with the modified subroutine.  Once it does, I will put the new stuff in GitHub and probably also merge in develop.  Hopefully that will happen before lunch today.

JasonGlazer:
@RKStrand any idea what the diffs are about from the W64 CI machine? I would not think it would impact convergence since it is a side calc.

mjwitte:
@JasonGlazer Those diffs are likely due to develop moving before Win64 CI got around to running this (2 days after the other platforms). I would ignore those. If you want to confirm, you could run one of the affected files locally with develop and this branch.

RKStrand:
@mjwitte Thanks for the clarification, Mike.  Hopefully that explains the diffs.

JasonGlazer (approved):

JasonGlazer:
Code review, reran unit tests locally, reran original issue. Looks good. The CI results no longer present so pushed merged version up to trigger again. At this point, I would say it is good to go unless performance is somehow impacted or CI is not green.

src/EnergyPlus/SizingManager.cc
```diff
@@ -844,6 +833,31 @@ namespace SizingManager {
...
+        
+        if ( (isPulseZoneSizing) && (!WarmupFlag) && (HourOfDay == HourDayToPulse) && (TimeStep == TimeStepToPulse) &&
+             ((KindOfSim == ksRunPeriodDesign) || (DayOfSim == 1)) ) {
+            return true;
+        } else {
+            return false;
+        }
+        
+    }
+
```
Myoldmopar, line 860: Wrapping this in a small, well-defined function is good.  And giving those magic numbers meaningful names is also a good thing.

tst/EnergyPlus/unit/SizingManager.unit.cc
```diff
@@ -390,3 +390,98 @@ TEST_F(EnergyPlusFixture, SizingManager_DOASControlStrategyDefaultSpecificationT
     ASSERT_EQ(21.1, ZoneSizingInput(1).DOASLowSetpoint);
     ASSERT_EQ(23.9, ZoneSizingInput(1).DOASHighSetpoint);
 }
+
+TEST_F(EnergyPlusFixture, SizingManager_CalcdoLoadComponentPulseNowTest)
```
Myoldmopar, line 394: Very nice unit test.  :+1:

Myoldmopar (approved):
All good from me.  I concur with @JasonGlazer that if CI comes back clean, this can merge.

Myoldmopar:
EDD diffs are unrelated.  I'm ready for this to go in.  Thanks @RKStrand for the changes and @JasonGlazer for the review.

PR #7755, opened 2020-02-10
Fix Missed PsyCpAirFnWTb Calls

mitchute:
PR fixes a few missed calls and documentation that were missed in #7479. 

### Pull Request Author

### Reviewer


mitchute:
@Nigusse @Myoldmopar The ```PsyCpAirFnWTdb``` and ```_fast``` functions still exist in [Pychrometrics.hh](https://github.com/NREL/EnergyPlus/blob/b1c11ff5a75e8e34a6e667b25b8d350f4f1cddae/src/EnergyPlus/Psychrometrics.hh#L379). The ```_fast``` function can be deleted, IMO, since it is not used anywhere. However, the other is only used for comparison against ```PsyCpAirFnW``` in the unit tests. What do you think about moving that function to the unit test file so that there's no confusion about which function to use?

mitchute:
As expected from #7479, there are a few small math diffs due to the method change. Once we resolve what to do with the old ```PsyCpAirFnWTdb``` function, this should be ready.

mitchute:
Ready for review.

doc/module-developer/src/energyplus-services/psychrometric-services.tex
```diff
@@ -44,7 +44,7 @@ \subsection{PsyRhoAirFnPbTdbW (Pb,Tdb,W,calledfrom)}\label{psyrhoairfnpbtdbw-pbt
 
 Returns the density of air in kilograms per cubic meter as a function of barometric pressure {[}Pb{]} (in Pascals), dry bulb temperature {[}Tdb{]} (in Celsius), and humidity ratio {[}W{]} (kilograms of water per kilogram of dry air).
 
-\subsection{PsyCpAirFnWTdb (W,Tdb,calledfrom)}\label{psycpairfnwtdb-wtdbcalledfrom}
+\subsection{PsyCpAirFnW (W,calledfrom)}\label{psycpairfnw-wcalledfrom}
 
 Returns the specific heat of air in Joules per kilogram degree Celsius as a function of humidity ratio {[}W{]} (kilograms of water per kilogram of dry air) and dry bulb temperature {[}Tdb{]} (Celsius).
```
Nigusse, line 49: This needs removing  the text "dry bulb temperature {[}Tdb{]} (Celsius)"

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -3360,7 +3360,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_GetInput)
     SimUnitarySystem(UnitarySystem(1).Name, FirstHVACIteration, UnitarySystem(1).ControlZoneNum, ZoneEquipList(1).EquipIndex(1), _, _, _, _, true);
 
     ZoneTemp = Node(ControlZoneNum).Temp;
-    CpAir = PsyCpAirFnWTdb(Node(InletNode).HumRat, Node(InletNode).Temp);
+    CpAir = PsyCpAirFnW(Node(InletNode).HumRat, Node(InletNode).Temp);
```
Nigusse: Need to remove "Node(InletNode).Temp" argument.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -3396,7 +3396,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_GetInput)
...
```
Nigusse: Same as above.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -4176,7 +4176,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_VarSpeedCoils)
...
```
Nigusse: Same as above.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -4645,7 +4645,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_VarSpeedCoils_CyclingFan)
...
```
Nigusse: Same as above.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -8754,7 +8754,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_ASHRAEModel_WaterCoils)
...
```
Nigusse: Same as above. There are a few places that need removing the temp argument from the PsyCpAirFnW() function.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -4208,7 +4208,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_VarSpeedCoils)
...
```
Nigusse: Same as above.

tst/EnergyPlus/unit/HVACUnitarySystem.unit.cc
```diff
@@ -4606,7 +4606,7 @@ TEST_F(EnergyPlusFixture, UnitarySystem_VarSpeedCoils_CyclingFan)
...
```
Nigusse: Same as above.

rraustad:
The unit test HVACUnitarySystem.unit.cc is no longer used and should be deleted from the repository. I forgot to do this during refactoring.

mitchute:
Thanks @rraustad. I'll get it taken care of.

mitchute:
@Nigusse Thanks for reviewing this. Feel free to merge it if CI looks good when it finishes.

Nigusse:
@mitchute I looked at the results and I see a few small diffs that look OK, but there are four test files that failed which should not. These four files have ems program. I think these failed test files need to be resolved before merging.

mitchute:
@Nigusse I see two files with EDD diffs, but those are purely text diffs due to changing the EMS method's signature. However, I don't see anything else. Which files are you referring to?

Nigusse:
@mitchute I checked the latest runs, all is good excepts the warnings, which are text diffs as you pointed out rightly. My earlier comment must have been based on previous runs. This is good to merge!

Nigusse:
@mitchute I will merge this branch shortly.

mjwitte:
Hmm, just catching up here. Did we add a transition rule for the old function name to new name or is EMS just mapping the old one behind the scenes?

Myoldmopar:
I don't believe a transition rule was written, so I think the Erl processor must just be mapping to the new function behind the scenes.  I'm not saying we shouldn't fix it, but I believe that's what it is doing.

mitchute:
Missed that too... I'll get that ready.

PR #7757, opened 2020-02-11
Fix Fluid Cooler Hard Crash

### LINKED ISSUE #7533: Fluid Cooler Crash
Issue overview
--------------
A user file is crashing in the fluid cooler sizing code because the plant sizing index is zero.  The fluid cooler is not autosized at all, and the file shouldn't need a sizing:plant object.  Adding a dummy sizing:plant object for that loop allows the file to avoid a hard crash (encounters a fatal error which is a different problem...for now).

Should probably address #5353 while at it. (Which is a follow on from a similar bug fix #5305).

Once fixed, the code should properly handle a fluid cooler file without sizing objects in the file.  As a second verification, you could hardsize a fluid cooler example file and remove the plant sizing objects and see if it does the same thing.

### Details
 - Platform: Original user is Windows, I believe, but I reproduced the issue on Linux
 - Version of EnergyPlus: file is 9.1, but I transitioned it to 9.2 and it had the same behavior

### Checklist
 - [ ] Defect file added: [FluidCoolerCrash_9.1.idf.txt](https://github.com/NREL/EnergyPlus/files/3672493/FluidCoolerCrash_9.1.idf.txt)
 - [ ] Ticket added to Pivotal for defect: #168846714

dareumnam:
Pull request overview
---------------------
 - Fixes #7533 
 - Defect user input file:
[FluidCoolerCrash_9.2.idf.txt](https://github.com/NREL/EnergyPlus/files/4183767/FluidCoolerCrash_9.2.idf.txt)

 - The hard crash with the user file is caused by an IF statement with error messages that call Plant size data, since there is no PlntSizData array when the plant sizing index is zero.
   - Line 1192 (https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/FluidCoolers.cc#L1192)
   - And line 1194 (https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/FluidCoolers.cc#L1194)
 - In the case of the user file,
   - Five IF statements in the FluidCoolerspecs::size() function before L1141 are not executed since the condition is not fulfilled. 
(https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/FluidCoolers.cc#L1141)
   - IF Statement of L1169 is executed since SolFla == -2 in this case.  
(https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/FluidCoolers.cc#L1169)
 - Once I fixed the code to show that message only when the plant sizing index is more than zero.  
 - Add a unit test that size(() function executes without any crash

### Pull Request Author
 - [X] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [X] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [X] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [X] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [X] If any defect files are updated to a more recent version, upload new versions here or on DevSupport

### Reviewer


mitchute:
I went ahead and pulled develop in one more time and ran all but one of the example files which contain FluidCoolers; all show no diffs. I also ran the unit tests and they all pass. The FluidCooler defect file referenced in #7533 (added to DevSupport [here](https://github.com/NREL/EnergyPlusDevSupport/blob/4d84c4e1b9cb48b5fee4c7d9abfc6bc219fa08f3/DefectFiles/7000s/7533/FluidCooler.idf)) fails in develop and now runs here to completion; however, it shows severe errors coming from ```SizeFluidCooler```. I'll look at that some more tomorrow after CI has had a chance to run.

mitchute:
CI seems to be happy with this. The core of the changes is just protecting some sizing variables from being accessed if the sizing routines fail. 

There appears to be some code duplication going on in the sizing routine, for example, in the ```this->HighSpeedFluidCoolerUAWasAutoSized && DataPlant::PlantFirstSizesOkayToFinalize``` IF block just above the proposed changes, [here](https://github.com/dareumnam/EnergyPlus/blob/df9357cf5eb8a47012121a8b77153e072ee1c1a7/src/EnergyPlus/FluidCoolers.cc#L1060-L1095). I'm wondering if these also need similar protections? @Myoldmopar ? Any thoughts on this before I merge it?

Myoldmopar:
Don't hold this up for that @mitchute.  Good catch on the duplicated code, but sizing is on the slate for major work soon, so no need to dwell on that.  I'll just merge this now.  Thanks @dareumnam for the fix and @mitchute for the reviewing.

PR #7778, opened 2020-02-14
Fix formatting warnings and also formatting checker

Myoldmopar:
Formatting warnings were just because tabs were found in a unit test file.  These are fixed here.

I found that the format checker was reporting out messages to CI, but using "filename" instead of just "file".  I fixed that here as well, and now it should properly point developers to the exact line that was problematic.

Reviewer: If you want to run this locally, just make sure you have clang-format installed, and from the root of the repository, run `./scripts/dev/verify_formatting.sh`.  It shouldn't report anything.  Then add a tab character to a src or tst file and re-run.  It should report a JSON warning about the character.

### DIFF
--- a/scripts/dev/verify_file_formatting.sh
+++ b/scripts/dev/verify_file_formatting.sh
@@ -13,7 +13,7 @@ grep -q -P "^ *\t *" "$1"
 
 if [ $? -eq 0 ]
 then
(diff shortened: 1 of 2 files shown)


mitchute:
@Myoldmopar fix verified on Linux. Merging.

PR #7805, opened 2020-02-23
Std function callbacks

Myoldmopar:
Add an alternative signature for registering runtime callbacks in C++ using `std::function`s.


Myoldmopar:
We should pull develop in here as well.  I still don't get it.  CI _did_ run this branch, and completed it.  In any case, it was showing diffs by the time everything ran, so it needs develop updated.

Myoldmopar:
I'll pull develop now and get CI to re-run.

Myoldmopar:
All happy here, merging.

PR #7812, opened 2020-02-26
"Urgent" IDD fix for min-fields on new Plugin Instance object

Myoldmopar:
Pull request overview
---------------------
Min fields is 5, but there are only 4 fields.  IDF-Editor complains when trying to open any IDF based on this version.  This must go in.

@mbadams5 is there a spot to hook in a check for this when generating the schema?  I can make a standalone test, but if I could just have the schema generator fail if it finds this condition, that would be great.

### DIFF
--- a/scripts/dev/generate_epJSON_schema/idd_parser.py
+++ b/scripts/dev/generate_epJSON_schema/idd_parser.py
@@ -145,6 +145,10 @@ def parse_idd(data):
             if 'memo' in obj_data:
                 root['properties'][obj_name]['memo'] = obj_data.pop('memo')
             if 'min_fields' in obj_data:
+                num_fields_with_name = len(obj_data['properties']) + 1
+                if int(obj_data['min_fields']) > num_fields_with_name:
+                    if 'extensions' not in obj_data['properties']:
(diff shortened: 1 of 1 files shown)


mbadams5:
@Myoldmopar Maybe, I'll have to look through the code to see if there is a good spot or not.

Myoldmopar:
@mbadams5 I think I found a good spot, and added in a check for min-fields > number of fields.  I had to protect against extensible objects, but it seems to work well.  If you run it as-is, it passes, but if you go into the IDD and put the plugin object back to min-fields 5, it throws a runtime error when trying to generate the schema.  Runtime errors are what are thrown in other sections of that idd parser, so I figured that was suitable.  Let me know if you like these changes.  And if you do, feel free to merge this if you want.

mitchute:
I verified the fix and test. Calling ```generate_epJSON_schema.py``` now fails when ```\min-fields``` is greater than ```num_fields_with_name```.

@mbadams5 any final thoughts on this before it merges?

mbadams5:
@mitchute I looked over it earlier this morning, but before CI results came back so I didn't merge then. It looks good to me.

mitchute:
The only real warning is Github asking to add labels to the PR. That is done and this looks good. Merging.

PR #7822, opened 2020-02-27
Fixes and Documentation for Performance Overrides and Reporting

JasonGlazer:
Some tasks were identified prior to the I/O freeze  for #7761 that are being completed in the pull request. This includes some code fixes but primarily is the documentation related to that work.

### Pull Request Author

### Reviewer


JasonGlazer:
Dr. Latex (@Myoldmopar) could you look at the tables I just added? They have a lot of columns and could probably be made to render better.

@mjwitte could you look over the IOref docs for the PerformancePrecisionTradeoffs object that is new and let me know what you think?

JasonGlazer:
@mjwitte I ended up making updates to the "classic" EP-Launch for this update, could you please do some checking when you test the next installer?

JasonGlazer:
Fixed the unit test that was failing but I did not see any issue with HospitalLowEnergy when run locally. Maybe it is just a timeout since it runs so slowly.

Myoldmopar:
I pushed a change that you can evaluate @JasonGlazer.  It uses footnotes for long column names to help fit very wide tables onto a normal page.  If you like it, you can adapt it and use it as you need.  If not, just discard those changes.  Here's one table I modified in this branch now:

image

JasonGlazer:
@Myoldmopar Is there a way to indicate where the wrapping should occur? I think that might be better than using abbreviations.

Myoldmopar:
With a tabularx environment, you can use a `\newline` to force a break, I'm not sure of a similar thing in the regular tabular environment.  I think no matter how you break it, it's going to be a little bit off.  It's one of those things where if you play nice with LaTeX, it will make beautiful documents.  Once you start manipulating things in custom ways, it's a constant battle.  My suggestion would be to take a second pass at maybe presenting the data in a different way that doesn't require this much header text.  If you can't come up with anything, I would try to create shorthands, like abbreviations, or at least less text.  If you still don't like that, then we would need to consider switching to tabularx or something which could be problematic if some devs don't have that package installed (I can't recall if it is in the minimal LaTeX distro).

JasonGlazer:
Hopefully the tables are good enough.

JasonGlazer:
@Myoldmopar and @mjwitte I would guess these failed tests don't have anything to do with my code but just that it has already fallen behind develop. Should I remerge develop or would you rather not have me put more onto the CI machines?

Myoldmopar:
@JasonGlazer The convection unit test is unrelated, and I'm hoping I have a fix in a different branch that will merge soon.

However, the two performance log unit tests are popping up in a bunch of branches spuriously.  It looks to me like it could be due to one of two things.
1. The most likely culprit is that you are reading/writing to the same perflog csv file name in each unit test.  Because the unit tests run in parallel, the timing of the unit tests is likely hitting the files at different times, and sometimes there is data in there already.  I would suggest changing the names of the files used to be something different between the two.  And maybe something that is obviously made up like "unit_test_perflog_file_1.csv" and "...flog_file_2.csv".  
2. The other thing that could be doing it is the use of statics in there.  I see how you are handling it and I don't think that's the root cause of this, but it could still be problematic later.  The ideal design here would be that the performance log would be managed by its own class, with member methods like `addHeader(...)`, and `addBody(...)`, and `writeContents(...)`.  Then you wouldn't have to have a single method with all this responsibility.  At a minimum, you need to make sure you are calling to clear those static variables from a clear_state call.  Or, you could move them to module level variables and reset them from a clear_state call.

I would advise you to make the changes to the unit test file names, and make sure you are calling to clear those variables during the clear_state process.  Then pull in develop one more time, CI should be happier, and we'll get this merged.

doc/output-details-and-examples/src/output-files/eplusout-perflogcsv.tex
```diff
@@ -0,0 +1,69 @@
...
+\end{itemize}
+
+When using this file it is good to set PerformancePrecisionTradeoffs to default values the first time to establish a baseline then change one input at a time and rerunning to understand the impact of that input. The total runtime decrease 
+and the impact on energy and the ocscillation variables should be considered prior to committing to a specific PerformancePrecisionTradeoffs scheme.
+
+An example is shown below:
+
+\begin{lstlisting}
+
+\end{lstlisting}
```
Myoldmopar: Did you mean to put something in this block?

doc/input-output-reference/src/overview/group-simulation-parameters.tex
```diff
@@ -823,7 +823,132 @@ \subsubsection{Inputs}\label{inputs-15-014}
...
+7  & TRUE  & ScriptF    & NORMAL & 6 & 1 & FALSE & 0.30 & 161.89 \tabularnewline
+8  & FALSE & CarrollMRT & NORMAL & 6 & 1 & FALSE & 0.30 & 171.00    \tabularnewline
+9  & FALSE & CarrollMRT & MODE01 & 1 & 1 & FALSE & 0.30 & 54.19  \tabularnewline
+10 & FALSE & CarrollMRT & MODE02 & 1 & 1 & FALSE & 0.30 & 53.85  \tabularnewline
+11 & FALSE & CarrollMRT & MODE03 & 1 & 1 & FALSE & 0.30 & 51.21  \tabularnewline
+12 & FALSE & CarrollMRT & MODE04 & 1 & 1 & TRUE  & 0.30 & 51.14  \tabularnewline
+13 & FALSE & CarrollMRT & MODE05 & 1 & 1 & TRUE  & 1.00 & 32.76  \tabularnewline
+\bottomrule
+\end{longtable}
+}
```
Myoldmopar, line 872: The tables look good in the CI built IORef.

doc/output-details-and-examples/src/output-files/eplusout-perflogcsv.tex
JasonGlazer: Good question

JasonGlazer: No need for this, I put examples in the IOref docs

JasonGlazer:
@Myoldmopar I think I addressed everything you brought up.

Myoldmopar:
Excellent!  I'll take a final look and we'll get this in.

doc/output-details-and-examples/src/output-files/eplusout-perflogcsv.tex
```diff
@@ -62,8 +62,3 @@ \section{eplusout\_perflog.csv}
 When using this file it is good to set PerformancePrecisionTradeoffs to default values the first time to establish a baseline then change one input at a time and rerunning to understand the impact of that input. The total runtime decrease 
 and the impact on energy and the ocscillation variables should be considered prior to committing to a specific PerformancePrecisionTradeoffs scheme.
 
-An example is shown below:
```
Myoldmopar: :+1:

src/EnergyPlus/StateManagement.cc
```diff
@@ -404,6 +404,7 @@ void EnergyPlus::clearAllStates()
     UnitHeater::clear_state();
     UnitVentilator::clear_state();
     UserDefinedComponents::clear_state();
+    UtilityRoutines::clear_state();
```
Myoldmopar, line 407: Very nice.  There was a time when `clear_states` were being added almost daily, but now it's a very rare thing for a namespace/module to _not_ have one.  This is good.

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -99,6 +99,14 @@ namespace UtilityRoutines {
     bool outputErrorHeader(true);
     ObjexxFCL::gio::Fmt fmtLD("*");
     ObjexxFCL::gio::Fmt fmtA("(A)");
+    std::string appendPerfLog_headerRow("");
```
Myoldmopar, line 102: I see that these are only in the cc file, not the header.  That is a very good thing.  There is no reason the implementation of this needs to be exposed to any other file, and the clear_state function clears these.  :+1:

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -444,32 +452,29 @@ namespace UtilityRoutines {
...
     // J.Glazer February 2020
     {
-        static std::string headerRow = "";
-        static std::string valuesRow = "";
-
         // the following was added for unit testing to clear the static strings
         if (colHeader == "RESET" && colValue == "RESET") {
-            headerRow = "";
-            valuesRow = "";
+            appendPerfLog_headerRow = "";
```
Myoldmopar, line 457: I think a more ideal thing would be to just have a function called `clearPerfLogHeaders()` rather than relying on special header names to clear these.  But that is OK for right now.  Next time you work on this area, please clean this out.

tst/EnergyPlus/unit/UtilityRoutines.unit.cc
```diff
@@ -179,7 +179,7 @@ TEST_F(EnergyPlusFixture, UtilityRoutines_appendPerfLog2)
     // make sure the static variables are cleared
     UtilityRoutines::appendPerfLog("RESET", "RESET");
 
-    DataStringGlobals::outputPerfLogFileName = "eplusout_perflog.csv";
+    DataStringGlobals::outputPerfLogFileName = "eplusout_2_perflog.csv";
```
Myoldmopar, line 182: Nice.

Myoldmopar (approved):
OK, I think I'll just build this locally and test it quickly instead of waiting on CI to do it all.  Save the CI cycles for other PRs.

Thanks for cleaning all this up, this is much nicer.

Myoldmopar:
The only failure is the `AirTerminalSingleDuctMixer_SimPTAC_ATMInletSide` unit test failure that has suddenly popped up, not just in this branch, but others also.  This looks good to go now, I'm not going to wait on other CI machines to run.  Merging.

PR #7829, opened 2020-03-02
Fix climatic data summary table output

### LINKED ISSUE #7102: Change climatic data summary string searches to avoid potential encoding differences
Issue overview
--------------
As noted [here in #6192](https://github.com/NREL/EnergyPlus/pull/6192#issuecomment-439079747) the string searches for some of the Climatic Data Summary fields have embedded special characters which fail to match if the weather stat file does not have the same encoding as EnergyPlus (which is now UTF-8). Note the bad encoding of the degree symbol for latitude and longitude. Also all the lines related to Köppen classification and ASHRAE Handbook CDD/HDD are missing with the Western (Windows 1252) encoded stats file.

It can't find the Köppen classification because it is searching the Western (Windows 1252) line using an UTF-8 encoded string. We'd need to either re-write the specific parts to be encoding agnostic, read the file stats file as UTF-8, or convert the encoding of that specific input string to UTF-8.

Possible approach (MJW):Would it work to search just parts of the string, leaving out the offending characters, and then output the full utf-8 string, not what's in the stat file? In fact, I'd prefer to see the line show up in the table with a blank value even if the data isn't found in the stat file.

### Details
 - Ref #6192 

### Checklist

mjwitte:
Pull request overview
---------------------
 - Fixes #7102 
 - Change search strings for climatic data summary report to avoid potential encoding differences which resulted in missing entries from the report.

Initial code from @EnergyArchmage.

### Pull Request Author

### Reviewer


nrel-bot-2:
@mjwitte it has been 9 days since this pull request was last updated.

Myoldmopar:
@mjwitte I spent a few minutes in here trying to get this ready to merge.  The code looked fine, but I thought a unit test would be great.  In order to make the unit test, I pulled the section of code that parses the stat file line into a function `parseStatLine`.  I made the `int const` line type variables into an `enum class` in the header.  I then made a unit test that grabs a couple example lines from the fixes.

Local testing shows everything is working OK.  If something comes up when CI runs, we can back out these changes, but I think they'll be good.  If CI is happy when it is done, I'm ready to merge this.

Myoldmopar:
Hey hey!  The find_byref alert is working a champ!

I'll whitelist that function because it is working properly as designed.

Myoldmopar:
Alright did a little extra there.
1. The find_byref_bool override was reporting with filenames using the "filename" key, when CI was expecting to find the "file" key.  This was causing the files to show up as (Unknown file) on the CI results page.  I fixed the key.
2. That script was also reporting out just the source filename, like OutputProcessor.cc, instead of the path in the repository, like src/EnergyPlus/OutputProcessor.cc.  So even if CI had been reporting the filename, it would have been a broken link because it would have tried to point to OutputProcessor.cc at the root of the repo.  I modified it so it reports the proper relative path.
3. I then whitelisted the byref bool variables in this new function so they won't be reported.  Custom check should come out clean now.

The audit diffs look...interesting.  Perhaps a commit in one of the many PRs already merged this morning caused that.  I pulled in develop and we'll see this afternoon.

Myoldmopar (approved):

mjwitte:
@Myoldmopar The audit diffs are from this PR.
```
- numTableEntry=         871
+ numTableEntry=         881
```
More table rows now with the fix. TableDiff doesn't even report when it encounters tables of different size (maybe we could add that to the CI summary?)
I've tested this locally with two existing stat files, then UTF-8ed a stat file and re-ran, and everything is beau-ti-ful . . .  in its own wayyy, like a starry sum-mer night or a fully populated climate summaryyyyyy. CI looks clean, merging. BTW - we may want to check the stat files in the repo and make sure they're fully UTF-8.

PR #7835, opened 2020-03-03
Fixup Generator Expression Usage

Myoldmopar:
@RKStrand your path issue was reproduced on another dev machine.

The root of the issue is that I am using a generator expression to get the path to the E+ binary so I can drop the configured API files next to the binary, and then quoting it, which passed in whitespace escape slashes as a literal path to the underlying cmake script.  Who knew generator expressions would have gotchas?  (@kbenne that's who)

@matthew-larson I believe this addresses the issue you saw.  Feel free to pull this branch and verify if you want.


matthew-larson:
@Myoldmopar I did have to unquote line 779 as well for the source directory path to come out correctly, otherwise it was sticking another forward slash in the space in the file path.

Myoldmopar:
@matthew-larson can you confirm the last change is what you had to do?  I'm ready to merge this, but honestly, on my Mac, even if I put white space in my build path, I don't see an issue even in develop, so I can't confirm.

mitchute:
FWIW: I checked this out, merge in develop locally, and ran ```ZoneCoupledKivaConvection.idf``` on Mac and Linux and am not seeing any of these diffs.

matthew-larson:
@Myoldmopar these are the changes I made, works on my end.

Myoldmopar:
@mitchute thanks for confirming the Kiva file thing.  No idea what's up with that at the moment.

@matthew-larson thanks for confirming!  I'll merge this in now.

PR #7836, opened 2020-03-04
Follow-up to API and Plugin work

### LINKED ISSUE #7803: Follow up to API effort
Issue overview
--------------
Followup to #7776. See PR discussion for more information. 
 
 - [x] Find all TODOs in src/EnergyPlus/api
 - [x] Fix actuators and sensors for all real/bool/int types
 - [x] Documentation
 - [x] API documentation -- not sure of location yet
 - [x] Gracefully handle running a python case in a dev build without python linked
 - [x] Set PYTHONHOME in the plugin manager: [doc](https://docs.python.org/3/c-api/init.html#c.Py_SetPythonHome)
 - [x] Include API header files in <eplusinstall>/include
 - [x] Add note in readme about API being a preliminary thing and subject to small breaking changes initially
 - [x] Add "linked-with-python-version" output to EnergyPlus `--help` or maybe `--version` message
 - [x] Use Doxygen to build API docs?
 - [x] Reorder source files in `src/EnergyPlus/CMakeLists.txt`
 - [x] Check other actuator types (logical, int)
 - [x] Fatal out if plugin is found in input during an API based run
 - [x] Add Python license to install
 - [x] Change EnergyPlusAPI.hh to just .h

Myoldmopar:
Pull request overview
---------------------
 - Fixes #7803 
 - Assortment of follow-up fixes to the python plugin and C/Python API work


nrel-bot-2:
@Myoldmopar it has been 7 days since this pull request was last updated.

Myoldmopar:
Prior to cleaning up user defined Python plugin code:
```
    Start 488: integration.PythonPluginUserDefined5ZoneAirCooled
8/8 Test #488: integration.PythonPluginUserDefined5ZoneAirCooled .....   Passed   28.09 sec
```

After [cleaning up](https://github.com/NREL/EnergyPlus/pull/7836/commits/3533037b3e128d029294f35c90ff50595546002c) user defined Python plugin code:
```
    Start 488: integration.PythonPluginUserDefined5ZoneAirCooled
8/8 Test #488: integration.PythonPluginUserDefined5ZoneAirCooled .....   Passed    2.67 sec
```

The only change I made was that the scripts get the handles one time, rather than every time the plugin is called.  This avoids about 10 trillion string lookups :laughing: 

And hey hey, beyond 10x! :racing_car: :racing_car: :racing_car:

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -229,34 +247,71 @@ int getActuatorHandle(const char* componentType, const char* controlType, const
...
-        *theActuator.IntValue = (int)std::lround(value);
+    if (handle >= 1 && handle <= (int)EnergyPlus::DataRuntimeLanguage::EMSActuatorAvailable.size()) {
+        auto & theActuator(EnergyPlus::DataRuntimeLanguage::EMSActuatorAvailable(handle));
+        if (theActuator.RealValue) {
+            *theActuator.RealValue = value;
+        } else if (theActuator.IntValue) {
+            *theActuator.IntValue = (int)std::lround(value);
+        } else {
+            // follow protocol from EMS manager, where 1.0 is true, 0.0 is false, and anything else is also false
+            *theActuator.LogValue = value < 0.99999 && value < 1.00001; // allow small tolerance while passing between languages and types
```
rraustad: LogValue is bool. If value < 0.99999 it should be false, here it's set to true. I think you meant value > 0.99999 ?

Myoldmopar: Outstanding find @rraustad, thanks!

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -316,60 +392,120 @@ int getPluginTrendVariableHandle(const char* name) {
...
     if (handle < 0 || handle > EnergyPlus::PluginManagement::pluginManager->maxTrendVariableIndex) {
-        throw std::runtime_error("Plugin Trend Variable index out of range");
+        // need to fatal out once the plugin is done
+        // throw an error, set the fatal flag, and then return
+        EnergyPlus::ShowSevereError("Data Exchange API: Problem -- index error in getPluginTrendVariableValue; received handle: " + std::to_string(handle));
+        EnergyPlus::ShowContinueError("The getPluginTrendVariableValue function will return 0 to allow the plugin to finish, then EnergyPlus will abort");
+        EnergyPlus::PluginManagement::shouldIssueFatalAfterPluginCompletes = true;
+        return 0;
     }
     if (timeIndex < 2 || timeIndex > ((int)EnergyPlus::PluginManagement::pluginManager->getTrendVariableHistorySize(handle) - 1)) {
```
rraustad: So I can't get the "value" of the trend variable for timeIndex = 1 from getPluginTrendVariable"Value" function? Should use timeIndex < 1 here?

Myoldmopar: The trend variable is just a wrapper over a regular plugin variable.  The plugin variable must be first declared and assigned, then the trend is used to track the value of it.  Thus the plugin should already have access to the underlying variable, and would just grab the current value of that variable directly rather than doing a trend lookup.  I guess I could just make this go down to one though, it doesn't hurt anything.

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -279,16 +334,28 @@ int getInternalVariableHandle(const char* type, const char* key) {
 }
 
 Real64 getInternalVariableValue(int handle) {
-    if (handle == 0) {
-        return 1;
+    if (handle >= 1 && handle <= (int)EnergyPlus::DataRuntimeLanguage::EMSInternalVarsAvailable.size()) {
```
rraustad, line 342: At line 402 below you use size-1 (in the if test), here just size. I didn't check if these arrays are 1-based index verses 0-based index, just pointing this out.

Myoldmopar: Any of the plugin manager related things will be zero based, while any of the existing structures (EMS variables, actuators, sensors, meters) will be one based.  This gets messy, but I think I'm tracking it, and I have plans to increase the unit testing to further improve robustness.

rraustad: This does look OK. handle vs timeIndex at line 402.

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -316,60 +392,120 @@ int getPluginTrendVariableHandle(const char* name) {
...
     if (handle < 0 || handle > EnergyPlus::PluginManagement::pluginManager->maxTrendVariableIndex) {
-        throw std::runtime_error("Plugin Trend Variable index out of range");
+        // need to fatal out once the plugin is done
+        // throw an error, set the fatal flag, and then return
+        EnergyPlus::ShowSevereError("Data Exchange API: Problem -- index error in getPluginTrendVariableMin; received handle: " + std::to_string(handle));
+        EnergyPlus::ShowContinueError("The getPluginTrendVariableMin function will return 0 to allow the plugin to finish, then EnergyPlus will abort");
+        EnergyPlus::PluginManagement::shouldIssueFatalAfterPluginCompletes = true;
+        return 0;
     }
     if (count < 2 || count > ((int)EnergyPlus::PluginManagement::pluginManager->getTrendVariableHistorySize(handle) - 1)) {
```
rraustad: Same here, what if the min value is in timeIndex ==1, use count < 1.

rraustad: Oh, I see, the array size should be >1 but you'll still get the min value of all elements.

Myoldmopar: I hear you on the trend index, I'll allow a lookup of just the first value.

Myoldmopar: Yeah, on the trend functions that look over a list, I think they should require at least two.  Getting the min/max/average/direction/sum of a single value seems dumb.  But looking up the value at the current time spot seems good.  Is it OK just just modify the "getValue" function to allow 1 and require the others to be >= 2?

rraustad: Yes, I agree.

rraustad: I assume that anytime these trend arrays are allocated the size is > 1. I hope that's always true.

Myoldmopar: Huh.  I think I see a problem.  I appreciate your comments on this.

This shouldn't be comparing `count > size - 1`, this should just be comparing `count > size`.

Imagine if I create a trend variable with 3 history terms.  Then I call to get the min over each time count:
 - With count == 1, this will abort, because why would you call `min` for just one value.
 - With count == 2, this will calculate the min over the first and second values
 - With count == 3, this will abort, because it would have compared `count > size - 1`, which is `3 > 2`, and thought it was out of range.

I'll go through and clean these up!

.gitignore
```diff
@@ -26,6 +26,9 @@ doc/title.tex
 # ignore Energy+.schema.epJSON while it is auto-generated from IDD
 Energy+.schema.epJSON.in
 
+# ignore the doxygen built API docs
+src/EnergyPlus/api/_build
```
Myoldmopar, line 30: Doxygen can be run from the src/EnergyPlus/api directory, and it will put the docs inside _build.  At some point, I imagine the docs will be produced by CI and included in the package, but for now, nope.

cmake/Install.cmake
```diff
@@ -290,6 +290,9 @@ set( RULES_XLS Rules9-2-0-to-9-3-0.md )
 install(FILES "${PROJECT_SOURCE_DIR}/release/Bugreprt.txt" DESTINATION "./")
 install(FILES "${PROJECT_SOURCE_DIR}/release/favicon.png" DESTINATION "./")
 install(FILES "${PROJECT_SOURCE_DIR}/release/readme.html" DESTINATION "./")
+if (LINK_WITH_PYTHON)
+  install(FILES "${PROJECT_SOURCE_DIR}/release/PythonLicense.txt" DESTINATION "./")
+endif()
```
Myoldmopar, line 295: Per our legal team, it is sufficient to just include the python license text in the E+ root install if we are packaging up Python binary bits and standard library source bits.

doc/ems-application-guide/src/application-guide-for-ems.tex
```diff
@@ -4,4 +4,6 @@ \chapter{Introduction}\label{introduction}
...
 
+In addition to the built-in EMS language, version 9.3 of EnergyPlus added the ability for users to create custom Python scripts (plugins) that can be used to extend the immense capabilities already added by normal EMS.  For information on the use of Python Plugins, check the Python Plugins section of the InputOutputReference document.
```
Myoldmopar, line 7: A future revision of the EMSApplicationGuide will contain an overhaul pointing to the Python stuff.  For now I think it's nice to mention this as an alternative, not the primary.  Issues found during this release cycle will cause the code to be more solidified for the next release.

doc/input-output-reference/src/overview/api-usage.tex
```diff
@@ -0,0 +1,300 @@
+
+\section{Using EnergyPlus as a Library}\label{sec:api-usage}
```
Myoldmopar, line 2: This new file documents how to use EnergyPlus as a library.  This does fit in the context of an I/O ref, because this is a whole new way to do I/O with EnergyPlus.  There can always be more documentation, but I think this is pretty well fleshed-out.  It's definitely enough for a developer to get an API called EnergyPlus working.

doc/input-output-reference/src/overview/group-python-plugins.tex
```diff
@@ -0,0 +1,299 @@
+\section{Group -- Python Plugins / Python EMS}\label{group-plugins}
```
Myoldmopar, line 1: This new file documents how to use the new EMS system -- Python Plugins.  The plugins make heavy use of the API functionality, so this section links back up to there as needed.  Eventually this could be moved to the EMSApplicationGuide, but this application also comes with many new input objects, so breaking it up will need to be done carefully.  For now this should be a good solid base that users can learn a lot, and any issues found can be used to inform a cleanup on this for the next release.

release/readme.html
```diff
@@ -32,8 +37,8 @@
             <div class="col-md-8 text-center">
                 <h1><img src="favicon.png" alt="EnergyPlus Logo" width="98" height="98"> EnergyPlus</h1>
                 <p>
-                    Version 9.2.0<br>
-                    30 September 2019<br>
+                    Version 9.3.0<br>
```
Myoldmopar, line 40: Went ahead and fully updated the README for this release, since I was making changes anyway.  Feel free to pop open this html in the browser and have a look to review it.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -256,6 +256,18 @@ SET( SRC
   GroundTemperatureModeling/FiniteDifferenceGroundTemperatureModel.hh
   GroundTemperatureModeling/GroundTemperatureModelManager.cc
   GroundTemperatureModeling/GroundTemperatureModelManager.hh
+  GroundTemperatureModeling/KusudaAchenbachGroundTemperatureModel.cc
```
Myoldmopar, line 259: Just rearranged the source here to get things back to alphabeticalish.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -732,19 +732,23 @@ if(LINK_WITH_PYTHON)
...
+SET( API_HEADERS
   api/func.h
-  api/func.cc
   api/runtime.h
-  api/runtime.cc
   api/datatransfer.h
-  api/datatransfer.cc
   api/EnergyPlusPgm.hh
+  api/EnergyPlusAPI.h
+  api/TypeDefs.h
```
Myoldmopar, line 741: Broke headers into their own list for easier access below, and added the base API.h and TypeDefs.h files since they will need to be included in user-created API scripts.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -833,10 +837,18 @@ if(LINK_WITH_PYTHON)
...
   if(LINK_WITH_PYTHON)
+    # we'll want to grab the standard lib for python plugins
     install(DIRECTORY "$<TARGET_FILE_DIR:energyplus>/python_standard_lib/"
             DESTINATION "./python_standard_lib")
   endif()
+  # we'll want to always provide the C API headers
+  foreach(HEADER ${API_HEADERS})
+    install(FILES "${CMAKE_CURRENT_SOURCE_DIR}/${HEADER}"
+            DESTINATION "./include/EnergyPlus/api")
+  endforeach()
```
Myoldmopar, line 850: The API headers are now available in the E+ package in an ./include/ directory, to allow user scripts to compile against them.

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -129,7 +130,7 @@ namespace CommandLineInterface {
         // Define options
         ezOptionParser opt;
 
-        opt.overview = VerString;
+        opt.overview = VerString + "\nPythonLinkage: " + PluginManagement::pythonStringForUsage();
```
Myoldmopar, line 133: The command line now reports the python version/description.

src/EnergyPlus/CommandLineInterface.hh
```diff
@@ -48,7 +48,7 @@
 #ifndef CommandLineInterface_hh_INCLUDED
 #define CommandLineInterface_hh_INCLUDED
 
-#include <EnergyPlus/api/EnergyPlusAPI.hh>
+#include <EnergyPlus/api/EnergyPlusAPI.h>
```
Myoldmopar, line 51: Renamed as this header is a C header, not C++.

src/EnergyPlus/EMSManager.cc
```diff
@@ -310,9 +310,11 @@ namespace EMSManager {
...
         // also call plugins and callbacks here for convenience
         bool anyPluginsOrCallbacksRan = false;
-        PluginManagement::runAnyRegisteredCallbacks(iCalledFrom, anyPluginsOrCallbacksRan);
-        if (anyPluginsOrCallbacksRan) {
-            anyProgramRan = true;
+        if (iCalledFrom != DataGlobals::emsCallFromUserDefinedComponentModel) { // don't run user-defined component plugins this way
+            PluginManagement::runAnyRegisteredCallbacks(iCalledFrom, anyPluginsOrCallbacksRan);
+            if (anyPluginsOrCallbacksRan) {
+                anyProgramRan = true;
+            }
```
Myoldmopar, line 317: User-defined plugins are called directly from the user defined component manager in an active manner.  This was letting user-defined plugins get run multiple times.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -401,6 +401,7 @@ namespace OutputProcessor {
         EnergyMeters.deallocate();
         EndUseCategory.deallocate();
         UniqueMeterNames.clear();
+        apiVarRequests.clear();
```
Myoldmopar, line 404: Requested variables are now cleared after each simulation -- users will need to re-request them for subsequent runs in the same memory space.  This is documented clearly in the API docs.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -6126,7 +6127,9 @@ void SetupOutputVariable(std::string const &VariableName,           // String Na
         thisIVar.storeType = VariableType;
         thisIVar.VarName = KeyedValue + ':' + VarName;
         thisIVar.VarNameOnly = VarName;
+        thisIVar.VarNameOnlyUC = UtilityRoutines::MakeUPPERCase(VarName);
         thisIVar.VarNameUC = UtilityRoutines::MakeUPPERCase(thisIVar.VarName);
+        thisIVar.KeyNameOnlyUC = UtilityRoutines::MakeUPPERCase(KeyedValue);
```
Myoldmopar, line 6132: The integer output variable type did not have these two fields, so I added them.  Now I can match against integer output variables nicely.

src/EnergyPlus/PluginManager.cc
```diff
@@ -81,6 +81,7 @@ namespace PluginManagement {
 
     // some flags
     bool fullyReady = false;
+    bool shouldIssueFatalAfterPluginCompletes = false;
```
Myoldmopar, line 84: This is a new flag that is used to very nicely exit EnergyPlus when an error occurs during a Python Plugin.  If the plugin calls an API method with an erroneous index, for example, E+ should abort.  However, if E+ aborts, it will fail in the midst of a wild call stack because the C API is being called from Python which is being called from EnergyPlus(C++).  Instead of aborting right in the API, the API methods now issues a descriptive severe error, sets this flag, and then returns functional values in those cases.  Then the plugin can continue running, Python can clear the call stack, and once it is back in EnergyPlus land, the program can throw a nice fatal error and everything ends gracefully.
:heart:

src/EnergyPlus/PluginManager.cc
```diff
@@ -404,6 +417,7 @@ namespace PluginManagement {
         FileSystem::makeNativePath(pathToPythonPackages);
         auto a = Py_DecodeLocale(pathToPythonPackages.c_str(), nullptr);
         Py_SetPath(a);
+        Py_SetPythonHome(a);
```
Myoldmopar, line 420: This should clean up the warnings coming out on Mac's during E+ runs.

src/EnergyPlus/PluginManager.cc
```diff
@@ -520,19 +534,19 @@ namespace PluginManagement {
...
-                    PluginManager::addGlobalVariable(fields.at("variable_name_3"));
+                    this->addGlobalVariable(fields.at("variable_name_3"));
                 }
                 if (fields.find("variable_name_4") != fields.end()) {
-                    PluginManager::addGlobalVariable(fields.at("variable_name_4"));
+                    this->addGlobalVariable(fields.at("variable_name_4"));
                 }
                 if (fields.find("variable_name_5") != fields.end()) {
-                    PluginManager::addGlobalVariable(fields.at("variable_name_5"));
+                    this->addGlobalVariable(fields.at("variable_name_5"));
```
Myoldmopar, line 549: These need to reference the current instance being constructed, not call out to a global hanging function.

src/EnergyPlus/PluginManager.cc
```diff
@@ -567,6 +581,7 @@ namespace PluginManagement {
                 int variableIndex = EnergyPlus::PluginManagement::PluginManager::getGlobalVariableHandle(variableName);
                 int numValues = fields.at("number_of_timesteps_to_be_logged");
                 trends.emplace_back(thisObjectName, numValues, variableIndex);
+                this->maxTrendVariableIndex++;
```
Myoldmopar, line 584: A useful integer to avoid having to calculate the size() everytime.

src/EnergyPlus/PluginManager.cc
```diff
@@ -897,6 +912,9 @@ namespace PluginManagement {
                                        ", make sure it returns an integer exit code, either zero (success) or one (failure)");
         }
         Py_DECREF(pFunctionResponse); // PyObject_CallFunction returns new reference, decrement
+        if (EnergyPlus::PluginManagement::shouldIssueFatalAfterPluginCompletes) {
+            EnergyPlus::ShowFatalError("API problems encountered while running plugin cause program termination.");
+        }
```
Myoldmopar, line 917: This is where EnergyPlus will abort if it finds that an error occurred while a plugin was running.

src/EnergyPlus/PluginManager.cc
```diff
@@ -920,10 +938,11 @@ namespace PluginManagement {
 #endif
 
 #if LINK_WITH_PYTHON == 1
-    void PluginManager::addGlobalVariable(const std::string &name) {
+    void PluginManager:: addGlobalVariable(const std::string &name) {
```
Myoldmopar, line 941: Why is this space here?  If I have to make a commit I'll clean this up.

src/EnergyPlus/PluginManager.cc
```diff
@@ -1124,5 +1143,31 @@ namespace PluginManagement {
     void PluginManager::runSingleUserDefinedPlugin(int EP_UNUSED(index)) {}
 #endif
 
+#if LINK_WITH_PYTHON == 1
+    bool PluginManager::anyUnexpectedPluginObjects() {
```
Myoldmopar, line 1147: Provides a way to let SimulationManager cleanly abort if plugin objects are found during an API run.  Calling through the API and running with Python Plugins is not allowed in the same run.  Nor should it be necessary.  Maybe sometime in the future we can allow it, but I'm not sure it's justified.  Anyway, this is documented in the documentation.

src/EnergyPlus/SimulationManager.cc
```diff
@@ -368,6 +368,12 @@ namespace SimulationManager {
         if (!eplusRunningViaAPI) {
             EnergyPlus::PluginManagement::pluginManager =
                 std::unique_ptr<EnergyPlus::PluginManagement::PluginManager>(new EnergyPlus::PluginManagement::PluginManager);
+        } else {
+            // if we ARE running via API, we should warn if any plugin objects are found and fail rather than running silently without them
+            bool invalidPluginObjects = EnergyPlus::PluginManagement::PluginManager::anyUnexpectedPluginObjects();
+            if (invalidPluginObjects) {
+                ShowFatalError("Invalid Python Plugin object encounter causes program termination");
```
Myoldmopar, line 375: Called function will issue severes, this just fatals after they are done.

src/EnergyPlus/UserDefinedComponents.cc
```diff
@@ -165,6 +165,8 @@ namespace UserDefinedComponents {
         if (thisLoop > 0) {
             if (this->Loop(thisLoop).ErlInitProgramMngr > 0) {
                 EMSManager::ManageEMS(DataGlobals::emsCallFromUserDefinedComponentModel, anyEMSRan, this->Loop(thisLoop).ErlInitProgramMngr);
+            } else if (this->Loop(thisLoop).initPluginLocation > -1) {
+                EnergyPlus::PluginManagement::pluginManager->runSingleUserDefinedPlugin(this->Loop(thisLoop).initPluginLocation);
```
Myoldmopar, line 169: The init function was not being called for a user defined plant component in plugin mode.

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -120,37 +120,82 @@ int apiDataFullyReady() {
...
     // should be called before energyplus is run!
+    // note that the variable request array is cleared during clear_state, so if you run multiple E+ runs, these must be requested again each time.
     EnergyPlus::OutputProcessor::APIOutputVariableRequest request;
     request.varName = type;
     request.varKey = key;
     EnergyPlus::OutputProcessor::apiVarRequests.push_back(request);
 }
 
 int getVariableHandle(const char* type, const char* key) {
+    // Variables are accessed through a single integer ID, but there are multiple internal types: real and integer.
```
Myoldmopar, line 131: Immense documentation and error handling added to all these API files.  I'm not going through everything in comments here, but the API tests and PythonPlugin tests are all passing beautifully, and they exercise these quite heavily.

testfiles/PythonPluginCustomSchedule.py
```diff
@@ -9,10 +9,13 @@ def actuate(self, x):
 
     def on_begin_timestep_before_predictor(self) -> int:
         if 'handles_done' not in self.data:
-            self.data['actuator_heating'] = self.api.exchange.get_actuator_handle(
-                "HTGSETP_SCH", "Schedule:Constant", "Schedule Value"
-            )
-            self.data['handles_done'] = True
+            if self.api.exchange.api_data_fully_ready():
```
Myoldmopar, line 12: This just shows the user how they can check whether the API data is initialized before trying to get handles to variables.

testfiles/PythonPluginDiscreteAirSystemSizes.py
```diff
@@ -17,7 +17,7 @@ def on_end_of_system_sizing(self):
                 "Intermediate Air System Main Supply Volume Flow Rate", "PSZ-AC_%s:%s" % (zone_num, zone_num)
             )
             actuator_handle = self.api.exchange.get_actuator_handle(
-                "PSZ-AC_%s:%s" % (zone_num, zone_num), "Sizing:System", "Main Supply Volume Flow Rate"
+                "Sizing:System", "Main Supply Volume Flow Rate", "PSZ-AC_%s:%s" % (zone_num, zone_num)
```
Myoldmopar, line 20: This was just backwards after the actuator function cleanup done earlier.

testfiles/PythonPluginUserDefined5ZoneAirCooled.py
```diff
@@ -9,6 +10,14 @@ def __init__(self):
 
         # get functional api glycol instance
         self.glycol = self.api.functional.glycol(u"water")
+        self.need_to_get_handles = True
```
Myoldmopar, line 13: Storing the handles and looking them up once, along with getting the file to run better, led to a 20x improvement in runtime for this new file :rofl: :rofl:.

tst/EnergyPlus/api/TestFunctional.c
```diff
@@ -141,7 +142,7 @@ int main() {
     Real64 cp = psyCpAirFnW(0.009);
     printf("C API Test: Expected Cp ~ 1007 J/kgK; Calculated: %8.4f\n", cp);
     Real64 energy = psyHfgAirFnWTdb(24);
-    printf("C API Test: Calculated energy?: %8.4f\n", energy);
+    printf("C API Test: Calculated energy: %8.4f\n", energy);
```
Myoldmopar, line 145: energy?

Myoldmopar (commented):
OK, whoever wants to look this over is more than welcome to do so.  

@rraustad thanks for your comments already, that was very helpful in getting some things tidied up this morning.  I feel much better about the trend functions after your insights provided some great improvements.

@mitchute @mjwitte and others, let me know if you are planning on reviewing this.  We need to get things tightened up today, and hopefully make a release candidate by tonight.  If you are planning on reviewing it, but cannot until tomorrow, let me know and we'll try to hold the door open.

release/readme.html
```diff
@@ -110,40 +128,71 @@ <h5 class="card-title">Yes... and no.</h5>
...
+                        <li class="list-group-item transition-object-name">HVACTemplate:System:VRF</li>
+                        <li class="list-group-item transition-object-name">HVACTemplate:Plant:Boiler</li>
+                        <li class="list-group-item transition-object-name">LifeCycleCost:UsePriceEscalation</li>
+                        <li class="list-group-item transition-object-name">LifeCycleCost:UseAdjustment</li>
+                        <li class="list-group-item transition-object-name">OtherEquipment</li>
+                        <li class="list-group-item transition-object-name">Output:Table:SummaryReports</li>
+                        <li class="list-group-item transition-object-name">ShadowCalculation</li>
+                        <li class="list-group-item transition-object-name">WaterHeater:Mixed</li>
+                        <li class="list-group-item transition-object-name">WaterHeater:Stratified</li>
+                        <li class="list-group-item transition-object-name">ZoneHVAC:HybridUnitaryHVAC</li>
```
mjwitte, line 188: Some of these show up in the transition utility, because they need to be modified when transitioning other objects, but the objects themselves didn't actually change for this release. Delete these: ZoneHVAC:EquipmentList, AirLoopHVAC:ZoneSplitter, AirLoopHVAC:SupplyPlenum, AirLoopHVAC, RoomAir:Node:AirflowNetwork:HVACEquipment (ugh - this never got mentioned in the transition rules doc), and AirflowNetwork:Distribution:Node, AirflowNetwork:Distribution:Linkage.

release/readme.html
```diff
@@ -205,119 +254,192 @@ <h2><a id="Weather_Data_63"></a>Weather Data</h2>
...
 
-    <h2><a id="Highlights_of_this_release"></a>Highlights of this release (V9.2.0)</h2>
+    <h2>Important Changes (OK, back to this library thing...)</h2>
+    <p>There are three changes in this release of EnergyPlus which rely on different software dependencies.</p>
+    <dl>
+        <dt>C/Python API</dt>
+        <dd>
+            A significant change to this release is the ability to call EnergyPlus as a library.  Up until now, the way
+            users interacted with EnergyPlus has always been to setup an input file, and call the EnergyPlus(.exe) binary to run
+            the file, then gather results as needed from output files.  While this is still available, a new approach has been
```
mjwitte: There's also the existing FMU external interface methods. Not sure if that's relevant here for what you're getting at.

release/readme.html
```diff
@@ -328,18 +450,14 @@ <h2><a id="Platforms_90"></a>Platforms</h2>
...
         known (and even more unknown) probably remain. Specifics on the remaining known problems can be found on the
-        GitHub
-        repository <a href="https://github.com/NREL/EnergyPlus/issues">issues page</a>. If you are interested in what
-        has
-        been
-        addressed, the issues resolved in this release are installed similarly in the <a
-                href="changelog.html">changelog</a>.
+        GitHub repository <a href="https://github.com/NREL/EnergyPlus/issues">issues page</a>. If you are interested in
+        what has been addressed, the issues resolved in this release are installed similarly in the
+        <a href="changelog.html">changelog</a>.
```
mjwitte, line 465: Seems like the changelog should be more prominent?

release/readme.html
```diff
@@ -205,119 +254,192 @@ <h2><a id="Weather_Data_63"></a>Weather Data</h2>
...
-            <td><a href="https://github.com/NREL/EnergyPlus/pull/7217">7217</a></td>
-            <td>Increases Schedule:Compact IDD fields</td>
+            <td><a href="https://github.com/NREL/EnergyPlus/pull/7562">7562</a></td>
+            <td>Add Three Alternative Models for Clear Sky Emissivity Calculation</td>
         </tr>
         <tr>
-            <td><a href="https://github.com/NREL/EnergyPlus/pull/7230">7230</a></td>
-            <td>Model DOAS Supplying Air to Inlets of Multiple AHUs</td>
+            <td><a href="https://github.com/NREL/EnergyPlus/pull/7603">7603</a></td>
+            <td>Input file conversion utility</td>
```
mjwitte: I changed the title of this PR to "Input file conversion utility IDF->epJSON and reverse".

mjwitte (commented):
@Myoldmopar Reviewed the readme. Nice new look!

release/readme.html
Myoldmopar: Thanks, removed them locally!

release/readme.html
Myoldmopar: Well, technically, we also had a way for users to call E+ as a library, but it was clunky, not well tested, and not well documented.  I can add a little nod to external interface though.

.gitignore
mitchute: Built the doxygen doc locally, and boy, do they look nice! 🏅

Myoldmopar:
Note the CustomSchedule diffs are expected.  Once I had fixed up the API function calls to error handle better, I realized this plugin was being called before the data was fully ready and so it was running in invalid conditions.  This is now [fixed](https://github.com/NREL/EnergyPlus/pull/7836/files#diff-c6b0641efd62325c8db456704603374cR12) and the file runs better.

release/readme.html
Myoldmopar: Updated in this section.

release/readme.html
Myoldmopar: Made a standalone line in the highlight list of changes to inform the user of the changelog.

mitchute:
I didn't dig through every line of change here, but I did test and build packages on all three platforms. Linux and Mac look great, and I was able to run the PythonPluginCustomOutputVariable.idf file successfully. Windows fails to build, but that's just my VM being dumb (```'operator =' is ambiguous``` in PluginManager.cc, IdfParser.cc and InputProcessor.cc due to calls to nlohmann::json stuff). If any of you Windows users know anything about this, let me know.

Looks like the 🏁 is close, @Myoldmopar !

Myoldmopar:
[:one:](https://github.com/Myoldmopar/PythonEMSPrototype/tree/f3ad02b7258fa0c1b7ca927ad883c8a705611655)

Myoldmopar:
@mitchute merge at will once CI confirms I didn't break anything.  If Windows runs on 9b77e41 [above](https://github.com/NREL/EnergyPlus/pull/7836#commitcomment-37998022), that would be sufficient testing, as the final commit after that was purely readme text changes.  Windows may skip that one entirely and jump to the last commit though.

Thanks for the reviews @mjwitte @rraustad @mitchute !

src/EnergyPlus/api/datatransfer.cc
rraustad: No need to do anything now but after thinking about this I think value > 0.99999 is sufficient. If value is ever > 1.00001 then that's a mistake and no need to waste the clock.

mitchute:
3, 2, 1... 🚀

PR #7840, opened 2020-03-04
Eliminating Reference variables

Myoldmopar:
This is a small refactoring effort that eliminates the `Reference` wrapper from almost all uses in EnergyPlus.  By the time it is done I think it will eliminate all of them.

Quick history: In Fortran, output variable addresses, EMS variable addresses, and a few other things were passed to the appropriate managers and stored as Fortran `POINTER`s.  It was kept this way when the code was converted to C++, however due to the nuances with Fortran `POINTER`s, it wasn't possible to do a quick and direct map to C++ pointers.   Instead, the Objexx library contained `Reference` classes that were used to wrap around the arguments and store the references.  It worked perfectly, but added an unnecessary layer of complexity over the variables.

This PR does not touch the lifetime/ownership/storage of any objects in the code.  The pointers are used as a simple address holder for objects that do have program lifetime.  And using references this way means maximizing the runtime efficiency.

The code in this PR simply changes the signature of a few structs to store addresses instead of `Reference` class instances.  Then instead of the unusual `>>=` expression, it just stores the address directly.

On some preliminary testing, I was able to see 2-4% runtime performance for annual runs.  The impact should vary based on the output impact of each specific run.  This amount isn't huge, but for a tiny effort like this, that also makes the code much cleaner and should have zero diffs whatsoever, it's a high value.

Let's see if CI agrees with me on all accounts, then I'll press forward on the last few `Reference`s and document my changes.


Myoldmopar:
Running some more tests, I'm seeing about 3% runtime improvement on a release build when running an output heavy example file, running develop and this branch each 5 times and averaging the real time.  On Mac.

Myoldmopar:
Added a random little bonus, clearing out the old "lastComponentSimulated" argument in the loop solver.  I had refactored this away back when I reorganized plant last year, but that particular argument got left behind.  From very quick testing, files still appear to be running fine, but I didn't run regressions.  We'll see...

Myoldmopar:
Hmmmm, ZoneCoupledKivaConvection has now thrown stray diffs on the first CI build and the third CI build...but not in the in-between one.  I wonder what's going on there.

Myoldmopar:
I will be curious to see what CI does with this last commit.  I ended up doing a little more surgery than I intended, completely removing the global RVariable and IVariable Reference variables that were in there.  Those were actually more dangerous and confusing than I thought because they routinely called `new` and then the ownership of this heap was handed over to a different structure instance, without any warning.  I ended up just *not* using a pointer in that spot and instead used a member instance directly.  Running 30 or so test files showed no diffs, so we'll see.  If it is totally happy then I am done with this effort, the only remaining Reference variables are in the results schema which I'm not touching for now.

Myoldmopar:
Well that kinda answers it.  So far so good.

mitchute:
Anything else @Myoldmopar before this goes in?

Myoldmopar:
The one CI unit test failure is _nothing_ to do with this work, it's fine.  The performance result is back in the green, and locally I am still seeing around 2-3% runtime improvement.  I would like to do some more refactoring on here, but I don't think I will.  This is good to go anytime.

PR #7813, opened 2020-02-26
Reduce use of Array1S objects

jasondegraw:
Pull request overview
---------------------
Reduces the use of the `Array1S` object, generally replacing the `Array1S` with a reference to an `Array1D` object. This PR mainly removes non-slicing uses of the object, leaving in place 2D to 1D slicing and other similar uses. The changes here should result in no diffs. This is part of the larger effort to migrate to `std::vector`.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [x] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Check any new function arguments for performance impacts

PR #7841, opened 2020-03-04
Reduce Array1A and Array1 usage

jasondegraw:
Pull request overview
---------------------
This PR is a follow up to #7813. It reduces the use of `Array1A` and `Array1` objects in preparation for replacing these arrays. Generally speaking, an `Array1A` is replaced with an `Array1D&` and `Array1` with `Array1D`. However, in the case of `Array1A`, there is generally a call to the `dim` method which we will not have with the replacement array. The sequence is pretty commonly this: an array is declared, allocated, and used in a function call.

```
Array1D<Real64> this_array;
int with_this_size = 42;
...
this_array.allocate(with_this_size);
...
DoSomethingWith(this_array, with_this_size);
```

The function typically looks something like this:

```
void DoSomethingWith(Array1A<Real64> an_array, int a_size)
{
  ...
  an_array.dim(a_size);
  ...
}
```

One option would just be to delete the `dim` call and be done with it, but it is making sure that the arrays are dimensioned correctly for use in the function. So rather than lose that, I defined a macro:

```
#define EP_SIZE_CHECK(array, min_size) assert(array.size() >= min_size)
```

and then the function modified to be:

```
void DoSomethingWith(Array1D<Real64> &an_array, int a_size)
{
  ...
  EP_SIZE_CHECK(an_array, a_size);
  ...
}
```

I had to modify the macro so it's a little uglier now. I hope we can move away from code that relies upon passing array sizes around and I am not proposing that the macro as a general solution. It should only be used in code that has been transitioned from the old arrays to the new arrays.

The branch is not quite ready to go, but it should be ready to go in a day or two. But I would like to get some review of this before too much longer, so I'm putting it in as a draft PR. I also wanted to try out a draft PR. Is this macro a good idea? Should we just delete the `dim` calls?

### Pull Request Author

### Reviewer


Myoldmopar:
@jasondegraw I appreciate all this effort to get rid of the array objects.  I can understand why the dim call was in there, and I think your macro approach is good, but I'll need to spend a little more time one day early this week digging in to understand the implications fully.

jasondegraw:
@Myoldmopar OK. There have been some build warnings cropping up in unexpected places. It looks like the `dim` call was all that was keeping a few arguments from being unused. I guess that's not alarming, but I'm not sure. Looks like there's one more place to apply `EP_UNUSED`.

Myoldmopar:
The fixture diff is one of the those gremlins that occasionally pops up, not related to this work.  @jasondegraw assuming we get some clean results by pulling develop in one more time today, are you ready to call this done from your end?  I haven't gone through the final code, but I don't want to again until you are ready to deliver it.

jasondegraw:
@Myoldmopar Yes, it should be ready to go. It took a couple tries to get all the unused variable warnings to go away, but other than that cropping up again somehow I'm not aware of anything that could be a problem. Merging in develop can be an adventure, if anything trips that up let me know and I can help out.

Myoldmopar:
The four diffs are from #7701 getting merged in.  I'm reviewing this now but it looks good to go from CI's view.

Myoldmopar (approved):
I'm really glad to see the usage of array1 and array1a getting cleaned out as a part of this larger work.  The use of a macro was a good idea for replacing dim, and you seem to have found a huge amount of args that should've been passed by reference.  I'm not sure if that will make a big difference in performance because I'm not sure what Objexx was doing in all these variations, but it will be critical to have that normalized once we are passing vectors around.  :+1:

Myoldmopar:
Calling it.  Looks good to me.  Merging.  Thanks @jasondegraw

PR #7851, opened 2020-03-11
Update to GCC 7.5

Myoldmopar:
The Linux CI machines updated and GCC got pushed from 7.4 to 7.5.  This updates our scripts to point to 7.5


Myoldmopar:
Looking fine, not waiting any longer.  Once this is pulled into people's branches, things should go smooth again.

PR #7854, opened 2020-03-12
Fix Generator:MicroCHP allocation problem

### LINKED ISSUE #7853: Generator:MicroCHP bug results in only the last one being operable
Issue overview
--------------
User file with 2 Generator:MicroCHP objects with FollowThermal only ran one generator, the other never ran.
Some debugging revealed the problem [here](https://github.com/NREL/EnergyPlus/blob/7e32c807a764b2b6f55fc573e75f97ad8b132997/src/EnergyPlus/GeneratorDynamicsManager.cc#L165-L167), the deallocate obliterates any prior generators, so only the last one remains in the `GeneratorDynamics` array. Looks like this was originally programmed to expect a loop inside this function, but now is called multiple times, so the elseif block should be deleted here? Does this function get called for any other generator type?

I don't see any possible workaround except to use a single Generator:MicroCHP.

### Details
 - Version of EnergyPlus 9.3.0-IDDFreeze
 - Helpdesk ticket number 15401

### Checklist
 - [x] Defect file added to EnergyPlusDevSupport\DefectFiles

mjwitte:
Pull request overview
---------------------
 - Fixes #7853
 - Fix allocation bug that prevents more than one Generator:MicroCHP from operating.

### Review Notes
 - First commit adds unit test, EnergyPlusFixture.MicroCHPTest_InitGeneratorDynamics, which [fails](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-345f829b7991a5150495189943b1c0d28dc46c81-x86_64-Linux-Ubuntu-18.04-gcc-7.5-UnitTestsCoverage-Debug.html).
 - Next commit adds the fix, new unit test add [10% coverage to GeneratorDynamics.cc](http://energyplus.s3-website-us-east-1.amazonaws.com/coverage/2020-03-12-EnergyPlus-06f0e37bea-x86_64-gcc-7.5UTCD/EnergyPlus/GeneratorDynamicsManager.cc.gcov.html) which was previously 0%.
 - No diffs expected in test suite.
 - Defect file now shows both generators operating, before fix, only one operated.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/GeneratorDynamicsManager.cc
+++ b/src/EnergyPlus/GeneratorDynamicsManager.cc
@@ -81,8 +81,6 @@ namespace GeneratorDynamicsManager {
     // MODULE INFORMATION:
     //       AUTHOR        B. Griffith
     //       DATE WRITTEN   July 2006
-    //       MODIFIED       na
-    //       RE-ENGINEERED  na
 
     // PURPOSE OF THIS MODULE:
     // collect routines for managing generator states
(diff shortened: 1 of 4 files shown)


mjwitte:
New unit test fails as expected before fixing the bug.

src/EnergyPlus/GeneratorDynamicsManager.cc
```diff
@@ -137,39 +108,16 @@ namespace GeneratorDynamicsManager {
...
-        // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
 
         // get the number of generators that might use this module
         NumGensWDynamics = MicroCHPElectricGenerator::NumMicroCHPs; // TODO  + NumFuelCellCGenerators
 
         if (!allocated(GeneratorDynamics)) {
             GeneratorDynamics.allocate(NumGensWDynamics);
-        } else {
-            GeneratorDynamics.deallocate();
-            GeneratorDynamics.allocate(NumGensWDynamics);
```
mjwitte, line 167: This is the fix. This function is called multiple times, so it should only allocate `GeneratorDynamics` once. These lines were clearing the data for prior generators, so only the last generator would be populated.

Myoldmopar (approved):
Great fix, and great unit test.  I'm running locally, but this should be an easy one to drop in right away.

Myoldmopar:
CI is also very happy, and yes, great work on the improved unit test coverage!

Myoldmopar:
Ran the defect file locally with develop and this branch.  In develop, I see the first generator is not running, but in this branch I see both running.  This is good to go.  Merging.

PR #7850, opened 2020-03-11
Resolve terminal unit sensible output calculation Approach II

Nigusse:
Pull request overview
---------------------
- Fixes #7728
- Changed sensible output provided calculation procedure for all single duct air terminal types.
- Diffs are expected due to this change
- This is pull request is based on approach II
- Approach I is described in PR  #7842. But personally I propose to review and accept approach II (this pull request).

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on 

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally

PR #7861, opened 2020-03-16
Address total/sensible/latent calculations issues for DX cooling coils and others

### LINKED ISSUE #7728: Resolve terminal unit sensible output calculation method differences
Issue overview
--------------
The old direct air terminal unit used a slightly different method to calculate `SysOutputProvided `. Special if block added for `SingleDuctConstVolNoReheat` to minimize diffs in #7448. See code here in [ZoneAirLoopEquipmentManager](https://github.com/NREL/EnergyPlus/blob/af62e24d8e77a64dd3fbd652b80819e6154696a1/src/EnergyPlus/ZoneAirLoopEquipmentManager.cc#L819-L832).

### Details
 - Version of EnergyPlus 9.3.0'ish

### Checklist

### LINKED ISSUE #7855: Address total/sensible/latent calculations issues for DX cooling coils and others
Issue overview
--------------
The total/sensible/latent calculations in DX cooling coils needs to be revised to use most appropriate equation in calculating sensible and latent cooling rates. Currently, the DX cooling coils sensible cooling rate calculation is based on minimum humidity ratio of coils inlet and outlet node, when it should have been average of them. Using minimum humidity ratio sometimes may lead to incorrect sensible/latent split. Currently, only VRF DX cooling coil uses average Hum Rat. This issue will also review if there other places in the code that may need similar fix. This is not as such a defect but it is the moist appropriate approach and leads tractable sensible/latent split of the total cooling rate.

### Details
 - Version 9.2 EnergyPlus 
 -  see comments in pull request https://github.com/NREL/EnergyPlus/pull/7850 

### Checklist

Nigusse:
Pull request overview
---------------------
 - Fixes #7855 
 - Fixes #7728 
 - This code change modifies the sensible/latent output calculations of the single-speed, multi-speed, two-speed, VRF and multi-mode DX cooling coils. The VRF dx coils already had this approach but now they were modified to use the new function.
 - Subsequently the total/sensible/latent output new function were applied the following parent objects: Furnace, MultiSpeedHeatPumps, UnitarySystem, FanCoilUnits, StandAloneERV, and FourPipeInductionUnit.
- Diffs are expected due to this fix.

### Diffs Summary 
*For files with big monthly meter diffs in annual simulation*
| File | Delta Electricity  | Delta Cooling Electricity | Delta Gas | Base Electricity [GJ] | Base Gas [GJ] | 
|:---|:---:|:---:|:---:|:---:|:---:|
| FurnaceWithDXSystemRHcontrol| 1.5% | 3.1% | 8.5%  | 260.6 | 118.9 |
| HeatPumpAirToAirWithRHcontrol| 0.1% | 0.25% | 0.5%  | 254.5 | 104.7 |
| Supermarket| 0.8% | 9.0% | 37.2%  | 2814.6 | 79.1 |
| UnitarySystem_TwoStageDXWithHumidityControl| 0.04% | 0.8% | 0.0%  | 173.6 | 53.1 | 
| _SmallOffice_Dulles | tiny | tiny | 0.1% | 

For the Supermarket, the increased cooling electricity and gas use are due to better humidity control with this branch vs develop. On some days, the RH spikes high in early evening with this branch, because the zone switches to a heating requirement and the system controls are not set to control RH during heating.

image

And here's the temperature profiles.
image

FurnaceWithDXSystemRHcontrol for some reason, seems to fall into one of two modes when first entering setback. Sometimes it falls into dehumidification with reheat and lands at the 18C heating setpoint with RH approx 50%. But at other times it will land between the heating and cooling setpoints and the RH is in the 30s. On 08/05, for example, develop lands in the deadband, but this branch is dehumidifying. But on 07/23 it's the opposite (develop is dehumidifying and this branch is not). That will be a followup investigation, see #8290 .

### Pull Request Author

### Reviewer


Nigusse:
This is initial fix addressing the total/sensible/latent output calculations for the following DX coils:
- single speed DX cooling coils
- multimode DX cooling coils
- VRF DX cooling coils (this coils has already been updated but needed to use new function)
- multi-speed DX Cooling coils

The next step is to identify parent objects that can be impacted by this change. But this could be a long list of parent objects. Furnace, MSHeatPump, UnitarySystem

Nigusse:
Diffs are expected for this fix. Below are the three changes that are expcted to impact results:
- the sensible/latent output calculation for DX cooling coils. This fix guarantees the sensible/latent split sums to the total cooling output. A new function was added to the general routine to calculate the total/sensible/latent outputs given flow rate and state points of the inlet and outlet nodes:
   ` CalcTotalSensibleLatentOutput(m_dot, T2, W2, T1, W1, total, sensible, latent)`
- propagated the sensible/latent output calculation to several parent objects. But this change had slightly undesirable impact on the zone air temperature and predicted load that necessitated to look how the predictor calculates the sensible load, which lead to the next modification.
- modified the sensible load calculation in the predictor CalcZoneSums() function. This change improved the slight disparity observed between zone air setpoint and zone air temperature due to the previous two fixes. Now, overall the zone thermostat and zone air temperature agreement slightly improved in selected test files investigated.

Nigusse:
Sample diffs based on test suite file *UnitarySystem_HeatPumpAuto.idf*. Zone air temperature comparison. This change impacts cooling operation more than it does heating. Where develop code meets the zone air thermostat, this branch also does the same. But in some cases this branch also meets the thermostat perfectly where the develop does not. 

ZoneAirTemperature_UnitarySystem


Zone air temperature and diffs table depicts the improvement better.

ZoneAirTemperatureTable_UnitarySystem

Nigusse:
Predicted load and diffs between develop and this branch for *UnitarySystem_HeatPumpAuto.idf* test file.

PredictedLoad_UnitarySystem

Nigusse:
sensible and latent output diffs between develop and this branch for UnitarySystem_HeatPumpAuto.idf test file.

SensibleCoolingRate_UnitarySystem


LatentCoolingRate_UnitarySystem

Nigusse:
Added three modified test suite files as defect file used for this comparison. These three test files shown similar diffs and trend. I do not see any benefit of posting similar results.

Nigusse:
HVAC Electricity use diffs between develop and this branch for UnitarySystem_HeatPumpAuto.idf test file.

HVACElectricity_UnitarySystem

Nigusse:
@rraustad @mjwitte @Myoldmopar This issue started correcting calculation for total/sensible/latent split of DX cooling coils. Then moved to apply the same approach to similar calculation in parent objects. So far parent objects Furnace, HVACMultiSpeedHeatPump, FanCoilUnit, HVACStandAloneERV, and UnitarySystem were updated. Furthermore, testing impacts of the change necessitated a similar change in predictor/corrector and applied. Based on a detailed test in four selected files the branch sometimes show an improvement based on how tightly the zone thermostat temperature set-point was maintained; otherwise, it is similar to the develop. Review of the rest of the code show there are more 18 modules that need similar code changes in multiple places. This issue is going to require more effort than I anticipated.

rraustad:
This is called scope creep. Do the best you can and open new issues to finish the job.

mjwitte:
@Nigusse Let's hold on adding more modules until everyone agrees with the approach. I haven't had a chance to look at any of this yet.

Nigusse:
Zone Air Temperature for 5ZoneAirCooled.idf test file. This the 5th test suite file. Again the space air temperature shows tighter and improved control during cooling hours.

ZoneAirTemperature_5ZoneAirCooled

ZoneAirTemperature_5ZoneAirCooled2

See the highlighted zone air temperature table for SPACE1-1 for detailed comparison. Where there is high cooling requirements, the zone air temperature are controlled better.

ZoneAirTemperatureTable_5ZoneAirCooled

Nigusse:
Airloop Main Cooling Coil total and sensible cooling rate comparison between develop and this branch for test file 5ZoneAirCooled.idf:

TotalSensibleCoolingRate_5ZoneAirCooled

Nigusse:
Facility and HVAC electric energy consumption comparison between develop and this branch for test file 5ZoneAirCooled.idf:

FacilityHVACElectricity_5ZoneAirCooled

Nigusse:
> @Nigusse Let's hold on adding more modules until everyone agrees with the approach. I haven't had a chance to look at any of this yet.

@mjwitte Further changes are on hold until we discuss the approach. These sample results are based on fixes that are already completed.

Nigusse:
@Myoldmopar @mjwitte @rraustad This issue may require changes that spans all over the place. For now I limited the fix to handful places only. I have also documented the changes made and impacts based on a limited tests in this pull request and in the header. I added comments that reflect scope of the issue and Mike also requested to hold on any further changes until the apprach is agreed on. I suggest that we need to discuss this issue via email or any other means at your convenience.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2177,4 +2177,47 @@ void TestReturnAirPathIntegrity(bool &ErrFound, Array2S_int ValRetAPaths)
     }
 }
 
+void CalcTotalSensibleLatentOutput(Real64 const MassFlow,  // air mass flow rate, {kg/s}
```
mjwitte: Good to have a common function for this to standardize calculations.

src/EnergyPlus/HVACMultiSpeedHeatPump.cc
```diff
@@ -3668,12 +3670,10 @@ namespace HVACMultiSpeedHeatPump {
...
 
-        // calculate sensible load met using delta enthalpy at a constant (minimum) humidity ratio)
-        MinHumRat = Node(MSHeatPump(MSHeatPumpNum).NodeNumOfControlledZone).HumRat;
-        if (Node(OutletNode).Temp < Node(MSHeatPump(MSHeatPumpNum).NodeNumOfControlledZone).Temp) MinHumRat = Node(OutletNode).HumRat;
-        LoadMet = AirMassFlow * (PsyHFnTdbW(Node(OutletNode).Temp, MinHumRat) -
-                                 PsyHFnTdbW(Node(MSHeatPump(MSHeatPumpNum).NodeNumOfControlledZone).Temp, MinHumRat)) -
-                  MSHeatPump(MSHeatPumpNum).LoadLoss;
+        // calculate sensible load met 
+        Real64 SensibleOutput(0.0);  // sensible output rate
+        SensibleOutput = AirMassFlow * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(Node(OutletNode).Temp, Node(OutletNode).HumRat, Node(MSHeatPump(MSHeatPumpNum).NodeNumOfControlledZone).Temp, Node(MSHeatPump(MSHeatPumpNum).NodeNumOfControlledZone).HumRat);
```
mjwitte: Seems like there should also be a function for CalcSensibleOutput to standardize this. And CalcTotalSensibleLatentOutput could call that so everyone's using the same code to calculate sensible.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -9289,11 +9279,14 @@ namespace UnitarySystems {
         //  IF (.NOT. FirstHVACIteration .AND. AirLoopPass .EQ. 1 .AND. AirflowNetworkFanActivated) THEN
         if (!FirstHVACIteration && AirflowNetwork::AirflowNetworkFanActivated) {
             Real64 DeltaMassRate = 0.0;
+            Real64 SensibleOutput = 0.0;  // sensible output rate, {W}
+            Real64 LatentOutput = 0.0;    // latent output rate, {W}
+            Real64 TotalOutput = 0.0;     // total output rate, {W} 
+            Real64 SensibleOutputDelta = 0.0;  // delta sensible output rate, {W}
+            Real64 LatentOutputDelta = 0.0;    // delta latent output rate, {W}
+            Real64 TotalOutputDelta = 0.0;     // delta total output rate, {W}
```
mjwitte: There are several places like this where a bunch of local variables have to be declared just to call the new function. Wondering if there's a way to avoid this.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -9302,19 +9295,14 @@ namespace UnitarySystems {
...
-                                                  Psychrometrics::PsyHFnTdbW(DataLoopNode::Node(ZoneInNode).Temp, MinHumRat)) +
-                                  DeltaMassRate * (Psychrometrics::PsyHFnTdbW(DataLoopNode::Node(this->AirOutNode).Temp, MinHumRat) -
-                                                   Psychrometrics::PsyHFnTdbW(DataLoopNode::Node(this->NodeNumOfControlledZone).Temp, MinHumRat));
+            CalcTotalSensibleLatentOutput(MassFlowRate, DataLoopNode::Node(this->AirOutNode).Temp, DataLoopNode::Node(this->AirOutNode).HumRat, DataLoopNode::Node(ZoneInNode).Temp, DataLoopNode::Node(ZoneInNode).HumRat, TotalOutput, SensibleOutput, LatentOutput);
+            CalcTotalSensibleLatentOutput(DeltaMassRate, DataLoopNode::Node(this->AirOutNode).Temp, DataLoopNode::Node(this->AirOutNode).HumRat, DataLoopNode::Node(this->NodeNumOfControlledZone).Temp, DataLoopNode::Node(this->NodeNumOfControlledZone).HumRat, TotalOutputDelta, SensibleOutputDelta, LatentOutputDelta);
+            this->m_SenLoadLoss = SensibleOutput - SensibleOutputDelta;
```
mjwitte: minus? not plus?  Same question below.

mjwitte (commented):
In general this is moving in a good direction to use a common function for this. What's the reference to show that this method is more correct than the current method?

src/EnergyPlus/UnitarySystem.cc
Nigusse: I will correct the incorrect sign.

src/EnergyPlus/UnitarySystem.cc
Nigusse: I did not see any way to avid it but certainly I will look into it again.

src/EnergyPlus/HVACMultiSpeedHeatPump.cc
Nigusse: I can add a separate new function it is efficient to use the Psychrometrics function "PsyDeltaHSenFnTdb2W2Tdb1W1" directly?

src/EnergyPlus/GeneralRoutines.cc
Nigusse: Currently in most places the sensible heat transfer between two state points (e.g. between heating and cooling coil inlet and outlet nodes, between air terminal units outlet and zone air node) use minimum humidity ratio to calculate sensible heat transfer. The minimum humidity ratio approach does not lead to sum of sensible and latent components equal to the total calculated from enthalpy difference unless the humidity ratio is the same for state 1 and state 2. I have verified the sensible and latent splits derived from enthalpy difference that are proposed here as a fix do sum to the total And there are more one way how you can represent the splits that sum to the total. I have used this approach to address similar problem in VRF DX cooling coil for issue #6950.

src/EnergyPlus/HVACMultiSpeedHeatPump.cc
mjwitte: Oh, sorry, I didn't recognize this function was the same one used in `CalcTotalSensibleLatentOutput`. OK, I see. 

Makes me wonder if `CalcTotalSensibleLatentOutput` should be in Psychrometrics instead of general?

Nigusse: Originally, I added the function into DXCoils but then moved to General as I expended its use to other modules. I am happy to move it to Psychrometrics if there is benefit.

Nigusse:
The new DX cooling coil model [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/Coils/CoilCoolingDX.cc#L541-L547) needs similar change if you guys are comparing results with existing dx cooling coil models.

Nigusse:
> In general this is moving in a good direction to use a common function for this. What's the reference to show that this method is more correct than the current method?

Attached document shows one way of doing the correct sensible and latent split that sum up to the total calculated from moist air enthalpy difference. 
[TotalandSensibleCalculationUsingEnthalpyOfAir.pdf](https://github.com/NREL/EnergyPlus/files/4507124/TotalandSensibleCalculationUsingEnthalpyOfAir.pdf)

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2154,4 +2154,47 @@ void TestReturnAirPathIntegrity(OutputFiles &outputFiles, bool &ErrFound, Array2
...
+                                   Real64 const TDB2,      // dry-bulb temperature at state 2 {C}
+                                   Real64 const dW2,       // humidity ratio at  at state 2
```
lgu1234: @Nigusse Argument name consistency. If TDB2 is used, W2 should be used, instead of dW2. Same as dW1. In general, dW may be understood as W difference between two pints.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2154,4 +2154,47 @@ void TestReturnAirPathIntegrity(OutputFiles &outputFiles, bool &ErrFound, Array2
...
+
+    using Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1;
+    using Psychrometrics::PsyHFnTdbW;
+
+    TotalOutput = 0.0;
+    LatentOutput = 0.0;
+    SensibleOutput = 0.0;
+    if (MassFlow > 0.0) {
+        TotalOutput = MassFlow * (Psychrometrics::PsyHFnTdbW(TDB2, dW2) - Psychrometrics::PsyHFnTdbW(TDB1, dW1)); // total addition/removal rate, {W};
+        SensibleOutput = MassFlow * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(TDB2, dW2, TDB1, dW1); // sensible addition/removal rate, {W};
```
lgu1234: @Nigusse There are many ways to split total. Since this function uses average W to calculate Cp, it is better to provide another function use min or inlet W to calculate sensible with a whole set of functions. Or, an optional argument may be used as
CalcTotalSensibleLatentOutput(....,SplitFlag)
SplitFlag = 1, average W
=2, Inlet W
= 3, min W

mjwitte: @lgu1234 ? We should use only one method. We should use one function and call it from everywhere.

lgu1234: @mjwitte I agree that it is better to have a single method. I also prefer to use the same method to process air movement, such as infiltration, mixing and cross mixing. The existing cooling coil calculation uses the min W to calculate the sensible. It is a good method and needs to keep. For the coiling coil, the min W occurs at the supply side, as inlet. This is the preferred method by using the source side condition theoretically.

However, the min W could occur at either source side or receiving side for infiltration, due to air movement from outdoor to indoor. Shall we use source side W, instead of min W? in addition, when we deal with mixing objects, the average W is preferred, because sensible calculations from source side and receiving side will be equal, without energy leak.

The above is my thought. That is the reason why I propose 3 methods, to allow case dependent.

rraustad: The specific heat of air changes with humidity ratio. For a cooling coil that removes moisture the specific heat used in the calculation of energy transfer, at least as far as I know, is not the entering or leaving air condition. It's the average (or the integral). For infiltration or mixing I would think the same logic applies. If there are references that say otherwise we need to review those to make sure the energy calculation is correct. It doesn't matter what calculation we use now, right or wrong, those need to be updated so we stop going in circles with this type of change. It might take more than 1 method, I'm not sure. But if 1 method can cover all cases then we should use only 1.

src/EnergyPlus/ZoneTempPredictorCorrector.cc
```diff
@@ -6416,8 +6417,9 @@ namespace ZoneTempPredictorCorrector {
                 //  how can we tell?  predict step must be lagged ?  correct step, systems have run.
                 auto const &node(Node(zec.InletNode(NodeNum)));
                 NodeTemp = node.Temp;
+                NodeHumRat = node.HumRat;
                 MassFlowRate = node.MassFlowRate;
-                CpAir = PsyCpAirFnW(ZoneAirHumRat(ZoneNum));
+                CpAir = PsyCpAirFnW(0.5 * (NodeHumRat + ZoneAirHumRat(ZoneNum)));
```
EnergyArchmage: Maybe the averaging of w is too simplistic here and there would need to be mass-flow-weighting across all the inlet nodes.  In the context of more than one inlet node, the average humidity ratio may need to account for all the inlet nodes.  One low flow inlet may be doing little to change humidity ratio compared to another high flow inlet.  Consider the aggregate/balance latent load across multiple inlets.

Nigusse: I agree, zone entering humidity ratio should be mass flow rate weighted average across all inlet nodes of a zone. Will look into it and make appropriate changes. @EnergyArchmage  Thanks.

mjwitte: I understand the intent here of including all of the other nodes, but that's going to make it impossible to match what happens here with the HVAC system output controls that use a similar calculation to figure out how much flow and/or what temperature to supply to meet the zone load. The system only knows about it's supply node, not any others.

mjwitte: This is inside a loop over inlets (or ADUs, depending on which loop) so there is already some weighting by flow as `SumSysMCp` and `SumSysMCpT` accumulate within these loops.

Looking further up in this function, infiltration and other flows come in from `MCPI, MCPV` etc. These are calculated in `ZoneAirLoopEquipmentManager::CalcAirFlowSimple` And these are not consistent. Infiltration and ventilation use the source cp, mixing uses the average, refrigeration door mixing uses both zones (somewhow).

For the zone balance, using the source humrat for Cp seems reasonable, because you're looking at the energy to move the source air to the zone condition. Granted it should be the zone condition after correction, but we don't have the luxury of knowing that.

nrel-bot-2:

Nigusse:
@mjwitte summary of changes:
1. CalcTotalSensibleLatentOutput() function is modified to use minimum HumRat, instead of average HumRat. This is expected to produce diffs in VRF air terminal unit
2. Modified several unit test that failed during local test run. Some are reverted to match develop, some required a small tweaks to make them pass and three new unit tests added for this issue were also modified based on minimum HumRat calculation
3. Reverted Hg calculation in one place in CalcMultiSpeedDXCoilCooling() 
4. Pull request #7850 for issue #7728 is merged into this branch and uses minimum HumRat, instead of average HumRat. Develop uses average HumRat for single-duct constant volume no reheat air terminal output calculation.

Nigusse:
@mjwitte I see several example files with FanCoilUnit are failing and it appears it is caused by how the minimum humidity ratio is set. In FCU *LoadMet* calculation *Develop* doesn't compare the two HumRat values to determine the minimum HumRat, rather assumes either the zone HumRat or FCU inlet HumRat as the minimum. This is different from this branch that makes the determination by comparing the two HumRat values. I have reverted this code to match results with develop. May be this will be dealt separately.

mjwitte:
@Nigusse Don't undo anything yet. Let's just see where the chips fall when using the same function everywhere.

Nigusse:
@mjwitte Okay I am holding off from making any changes. 

Below are highlights of the possible causes for big diffs observed:

(1) 13 example files with FanCoilUnit (out of 15) are failing and suspect the diffs are caused by how the minimum humidity ratio is set.  FCU *LoadMet* calculation in *Develop* doesn't compare the two HumRat values to determine the minimum HumRat, rather assumes either the zone HumRat or FCU inlet as the minimum HumRat. This is different from the fix in this branch. 

(2) CpAir value used to calculate SensOutputProvided in *AirTerminal:SingleDuct:ConstantVolume:NoReheat.* 
[Develop](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/SingleDuct.cc#L4772-L4773) uses average HumRat to calculate SensOutputProvided while this fix [branch](https://github.com/NREL/EnergyPlus/blob/171788635_Issue7855/src/EnergyPlus/Psychrometrics.hh#L1302-L1304) uses minimum HumRat . This change was introduced from a pull request #7850 merged into this branch.

(3) Difference in minimum HumRat value setting approach used in UnitarySys::calculateCapacity() function for *LoadMet* calculation.

(4) 12 example files with Big MTR diffs and 11 of them are supermarket example files. Note that all the 12 IDFs have *AirTerminal:SingleDuct:ConstantVolume:NoReheat* object, which is impacted by the change item 2 above.

(5) 8 out of 10 VRF example files show diffs. This is expected because this fix modifies the function used to calculate the VRF DX cooling coil cooling rate (changed from average HumRat to minimum HumRat). 

(6) I think, all other example files big diffs are primarily caused by changes in item 2 above and change item 3 could also cause diffs in a few of those example files.

src/EnergyPlus/DXCoils.cc
```diff
@@ -12127,6 +12126,8 @@ namespace DXCoils {
                     OutletAirDryBulbTemp = HSOutletAirDryBulbTemp;
                 } else {
                     if (FanOpMode == ContFanCycCoil) {
+                        // average latent heat of vaporization, {J/kg} that maintains correct sensible/latent split;
+                        MinAirHumRat = min(InletAirHumRat, SpeedRatio * HSOutletAirHumRat + (1.0 - SpeedRatio) * LSOutletAirHumRat);
                         Hfg = PsyHfgAirFnWTdb(MinAirHumRat, HSOutletAirDryBulbTemp * SpeedRatio + (1.0 - SpeedRatio) * LSOutletAirDryBulbTemp);
```
mjwitte, line 12131: MinAirHumRat is only used here now. So it can be declared here instead of way up top. But also, it's unused in `PsyHfgAirFnWTdb`, so may as well set it to zero.
But why isn't this one changed to use `PsyHfgAvgFnTdb2Tdb1` like line 12368 below?

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2139,4 +2139,106 @@ void TestReturnAirPathIntegrity(EnergyPlusData &state, IOFiles &ioFiles, bool &E
...
+    // Q_sensible = m_dot * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(TDB2, W2, TDB1, W1);
+    // or Q_sensible = m_dot * cp_moistair_average * (TDB2 - TDB1)
+    //    cp_moistair_average = Psychrometrics::PsyCpAirFnW(0.5 * (W2 + W1));
+    //    cp_moistair_average = cp_dryair + 0.5 * (W2 + W1) * cp_watervapor
+    // Q_latent = Q_total - Q_latent;
+    // or Q_latent = m_dot * hg_average * (W2 - W1);
+    //    hg_average = Psychrometrics::PsyHfgAvgFnTdb2Tdb1(TDB2, TDB1);
+
+    using Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1;
+    using Psychrometrics::PsyHFnTdbW;
```
mjwitte: No `using` statements in the new functions.

src/EnergyPlus/Psychrometrics.hh
```diff
@@ -1292,14 +1292,25 @@ namespace Psychrometrics {
     }
 
     inline Real64 PsyDeltaHSenFnTdb2W2Tdb1W1(Real64 const TDB2, // dry-bulb temperature at state 2 {C}
-                                             Real64 const dW2,  // humidity ratio at  at state 2
-                                             Real64 const TDB1, // dry-bulb temperature at  at state 1 {C}
-                                             Real64 const dW1   // humidity ratio  at state 1
+                                             Real64 const dW2,  // humidity ratio at state 2
```
mjwitte: Here and elsewhere, these arguments are misleading. They are not a delta. They should be `W1`, `W2`.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2139,4 +2139,106 @@ void TestReturnAirPathIntegrity(EnergyPlusData &state, IOFiles &ioFiles, bool &E
...
+
+    TotalOutput = 0.0;
+    LatentOutput = 0.0;
+    SensibleOutput = 0.0;
+    if (MassFlow > 0.0) {
+        TotalOutput = MassFlow * (Psychrometrics::PsyHFnTdbW(TDBEquip, dWEquip) -
+                                  Psychrometrics::PsyHFnTdbW(TDBZone, dWZone)); // total addition/removal rate, {W};
+        SensibleOutput =
+            MassFlow * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(TDBEquip, dWEquip, TDBZone, dWZone); // sensible addition/removal rate, {W};
```
mjwitte: `CalcZoneSensibleLatentOutput` is using `PsyDeltaHSenFnTdb2W2Tdb1W1` but `CalcZoneSensibleOutput` is using `PsyDeltaHSenFnTdbEquipTdbWZone`. This one should be using `PsyDeltaHSenFnTdbEquipTdbWZone`. And I'm not sure the `TotalOutput` here is correct. DeltaEnthalpy applies to conditions across a coil, not necessarily the same for air supplied to a zone

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2450,8 +2450,8 @@ namespace FanCoilUnits {
                             Calc4PipeFanCoil(state, FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOut);
                         }
                     }
-                    QUnitOut = AirMassFlow *
-                               (PsyHFnTdbW(Node(OutletNode).Temp, Node(InletNode).HumRat) - PsyHFnTdbW(Node(InletNode).Temp, Node(InletNode).HumRat));
+                    QUnitOut = AirMassFlow * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(
+                                                 Node(OutletNode).Temp, Node(InletNode).HumRat, Node(InletNode).Temp, Node(InletNode).HumRat);
```
mjwitte: I was expecting this to be one of the new `Calc...` functions, namely CalcZoneSensibleOutput`. Here and many other places in FanCoilUnits, HVACMultiSpeedHeatPump, HVACSingleDuct, and UnitarySystem.

mjwitte (commented):
@Nigusse FYI, I pushed up some formatting changes yesterday.
This is getting closer to the goal. Some more comments/questions below.

src/EnergyPlus/DXCoils.cc
Nigusse: Ok. I will move MinAirHumRat declaration here and set it to zero.

Nigusse: I think I missed to update this when I was reverting the code change. Now using PsyHfgAirFnWTdb in both places.

src/EnergyPlus/GeneralRoutines.cc
Nigusse: OK. Don't need use statements here.

src/EnergyPlus/Psychrometrics.hh
Nigusse: Ok. renamed the variables.

src/EnergyPlus/GeneralRoutines.cc
Nigusse: My concern for NOT using  PsyDeltaHSenFnTdbEquipTdbWZone() in CalcZoneSensibleLatentOutput() is that the later function's use varies a lot. Sometimes it is used to calculate total/sensible/latent outputs and other times it is used to calculate sensible/latent outputs only.

src/EnergyPlus/FanCoilUnits.cc
Nigusse: I used CalcZoneSensibleOutput in "Corrector" only. I'm not sure if we want to use in other places without investigating its implication.

Nigusse: I think the FanCoil unit can use this new function (CalcZoneSensibleOutput) but other places have to be investigated case by case.

Nigusse: Updated FanCoilUnit to use the new function CalcZoneSensibleOutput() in several places.

src/EnergyPlus/GeneralRoutines.cc
Nigusse: PsyDeltaHSenFnTdbEquipTdbWZone() uses the zone HumRat to calculate sensible heat transfer rate and this may not be the right approach for CalcZoneSensibleLatentOutput() since we are providing supply and zone HumRat values.  If we decide to use PsyDeltaHSenFnTdbEquipTdbWZone() in CalcZoneSensibleLatentOutput() then the supply HumRat argument will not be used and the zone HumRat may not be the minimum. As the result, the latent component can be incorrect.

mjwitte: @Nigusse For a given supply and zone condition, `CalcZoneSensibleLatentOutput` and `CalcZoneSensibleOutput` must return the same sensible load. The intent of these functions is to calculate the load impact on the zone from an air stream entering the zone, and if I recall correctly, we decided that the zone humrat was the appropriate one to use (even if it is not the minimum). Perhaps there are more use cases than just coil (change across the coil) and zone?

Nigusse: @mjwitte Okay I will make the changes in CalcZoneSensibleLatentOutput() function as we did in the CalcZoneSensibleOutput() function. It is good to document why we are using ZoneHumRat in certain places and MinHumRat in other places.

src/EnergyPlus/Psychrometrics.hh
mjwitte: @Nigusse Did you push up your latest changes? These are still named `dW`. Also, these special Psych functions need some comments to explain how they are derived. It appears that these are based on `PsyHFnTdbW` and then algebraically simplified to the delta H case to reduce computation time. Some comments referencing the base function name would be helpful, otherwise, it's not clear where these equations and constants come from. A comment to the effect of this is equivalent to `PsyHFnTdbW(TDB2, min(W1, W2)) - PsyHFnTdbW(TDB1, min(W1,W2))` would be adequate. Here and in the other similar functions. Also with a brief comment explaining the use case for each one (delta across a coil, supply into a zone, etc.).

Nigusse: I definitely renamed these variables elsewhere but may be I missed some of them.  I will renam the remaining variables and add comments how the two functions were derived and where the constant coefficients came from.

Nigusse: Done! I will test the unit test suite locally and push the branch shortly.

Nigusse:
@mjwitte the unit test UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTest failed locally.

Nigusse:
I fixed the unit test by modifying the speed ratio expected outcome. This unit test came from latest merge from develop. Now the unit test suite is happy and I will push the branch shortly.

mjwitte:
@Nigusse We're getting close. I ran annual regressions on representative files (e.g. just one Supermarket file) that were showing big diffs on CI. 4 of these had big monthly meter diffs. The annual diffs are summarized at the top of this PR. Can you explain the big change in energy use for FurnaceWithDXSystemRHcontrol and Supermarket?

Also, somewhere along the way for this, you had plots that showed that the thermostat control for some files was more precise - is that still the case after all of the changes that have been made?

@Myoldmopar This PR introduces some new functions to standardize the sensible/latent load split for 2 cases: across a coil, and supply entering a zone. Please review the new functions.

Nigusse:
The thermostat control improvements are not applicable as we are now using the minimum HumRat across all.

Nigusse:
@mjwitte We reverted some of changes that impacted thermostat control. The changes that reverted were to use an average CpAir (based on average HumRat) instead of CpAir calculated at minimum HumRat, and similar change in corrector. Those tighter thermostat control observed were primarily because of these two changes.

Nigusse:
@mjwitte Updated thermostats control results check. Thermostat control tightened in UnitarySystem_HeatPumpAuto example file but not in 5ZoneAirCooled. Note that in the latest regression ran these two example files have passed the CI tests.

See sample results for UnitarySystem_HeatPumpAuto.idf. This is consistent with previous results:

7855_UnitarySystem_heatPumpAuto_ZoneAirTemperature

See sample results for 5ZoneAirCooled.idf.  Results for this example file is identical with develop and I do not see any diffs in other variables as well.

7855_5ZoneAirCooled_ZoneAirTemperature

src/EnergyPlus/DXCoils.cc
```diff
@@ -783,17 +782,14 @@ namespace DXCoils {
...
-                    MinAirHumRat = min(DXCoil(DXCoilNum).InletAirHumRat, DXCoil(DXCoilNum).OutletAirHumRat);
-                    DXCoil(DXCoilNum).SensCoolingEnergyRate = AirMassFlow * (PsyHFnTdbW(DXCoil(DXCoilNum).InletAirTemp, MinAirHumRat) -
-                                                                             PsyHFnTdbW(DXCoil(DXCoilNum).OutletAirTemp, MinAirHumRat));
-                    //  Don't let sensible capacity be greater than total capacity
-                    if (DXCoil(DXCoilNum).SensCoolingEnergyRate > DXCoil(DXCoilNum).TotalCoolingEnergyRate) {
-                        DXCoil(DXCoilNum).SensCoolingEnergyRate = DXCoil(DXCoilNum).TotalCoolingEnergyRate;
-                    }
-
-                    DXCoil(DXCoilNum).LatCoolingEnergyRate = DXCoil(DXCoilNum).TotalCoolingEnergyRate - DXCoil(DXCoilNum).SensCoolingEnergyRate;
+                    CalcComponentSensibleLatentOutput(AirMassFlow,
```
Myoldmopar: Calculating this in one function certainly helps keep things unified.

src/EnergyPlus/DXCoils.cc
```diff
@@ -12104,21 +12088,35 @@ namespace DXCoils {
...
+                                                  InletAirDryBulbTemp,
+                                                  InletAirHumRat,
+                                                  HSOutletAirDryBulbTemp,
+                                                  HSOutletAirHumRat,
+                                                  SensibleOutputHS,
+                                                  LatentOutputHS,
+                                                  TotalOutputHS);
+                DXCoil(DXCoilNum).TotalCoolingEnergyRate = TotalOutputHS * SpeedRatio + TotalOutputLS * (1.0 - SpeedRatio);
+                DXCoil(DXCoilNum).SensCoolingEnergyRate = SensibleOutputHS * SpeedRatio + SensibleOutputLS * (1.0 - SpeedRatio);
+                DXCoil(DXCoilNum).LatCoolingEnergyRate = LatentOutputHS * SpeedRatio + LatentOutputLS * (1.0 - SpeedRatio);
```
Myoldmopar, line 12117: OK, so you calculate high and low speed outputs and then interpolate.  No problem.

src/EnergyPlus/FanCoilUnits.cc
```diff
@@ -2451,8 +2451,8 @@ namespace FanCoilUnits {
                             Calc4PipeFanCoil(state, FanCoilNum, ControlledZoneNum, FirstHVACIteration, QUnitOut);
                         }
                     }
-                    QUnitOut = AirMassFlow *
-                               (PsyHFnTdbW(Node(OutletNode).Temp, Node(InletNode).HumRat) - PsyHFnTdbW(Node(InletNode).Temp, Node(InletNode).HumRat));
+                    CalcZoneSensibleOutput(AirMassFlow, Node(OutletNode).Temp, Node(InletNode).Temp, Node(InletNode).HumRat, QUnitOut);
```
Myoldmopar, line 2455: OK, an accompanying calculation function for zones instead of components.

src/EnergyPlus/GeneralRoutines.cc
```diff
@@ -2136,4 +2136,94 @@ void TestReturnAirPathIntegrity(EnergyPlusData &state, IOFiles &ioFiles, bool &E
...
+
+    TotalOutput = 0.0;
+    LatentOutput = 0.0;
+    SensibleOutput = 0.0;
+    if (MassFlow > 0.0) {
+        TotalOutput = MassFlow * (Psychrometrics::PsyHFnTdbW(TDB2, W2) - Psychrometrics::PsyHFnTdbW(TDB1, W1)); // total addition/removal rate, {W};
+        SensibleOutput = MassFlow * Psychrometrics::PsyDeltaHSenFnTdb2W2Tdb1W1(TDB2, W2, TDB1, W1); // sensible addition/removal rate, {W};
+        LatentOutput = TotalOutput - SensibleOutput;                                                // latent addition/removal rate, {W}
+    }
+}
```
Myoldmopar, line 2170: Protecting the psychrometric function calls with a mass flow check is a good move here.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -8960,7 +8960,8 @@ namespace HVACVariableRefrigerantFlow {
             TempIn = DataLoopNode::Node(VRFTU(VRFTUNum).VRFTUInletNodeNum).Temp;
         }
         // latent heat vaporization/condensation used in moist air psychometrics
-        Real64 const H2OHtOfVap = PsyHfgAvgFnTdb2Tdb1(TempOut, TempIn);
+        // Real64 const H2OHtOfVap = PsyHfgAvgFnTdb2Tdb1(TempOut, TempIn);
```
Myoldmopar: You can just remove this line if any other changes are needed.

src/EnergyPlus/Psychrometrics.hh
```diff
@@ -1292,23 +1292,42 @@ namespace Psychrometrics {
...
+        // returns sensible enthalpy difference of moist air going from state 1 to state 2 (e.g across coils)
+        // using the minimum humidity ratio state points 1 and 2. This enthalpy difference multiplied by
+        // supply air mass flow rate yields sensible heat transfer rate across coils in Watts
+
+        // the following two functions for calculating enthalpy difference are equivalent:
+        // PsyDeltaHSenFnTdb2W2Tdb1W1() = PsyHFnTdbW(TDB2, min(W1, W2)) - PsyHFnTdbW(TDB1, min(W1,W2))
+        // PsyDeltaHSenFnTdb2W2Tdb1W1() function was derived by simplyfying the above expression
+        // The constant coefficients came from the equation for moist air enthalpy, PsyHFnTdbW()
+
+        Real64 Wmin = min(W1, W2);
```
Myoldmopar: I'd mark it as const just for good measure; it won't make a difference in runtime.

src/EnergyPlus/Psychrometrics.hh
```diff
@@ -1292,23 +1292,42 @@ namespace Psychrometrics {
...
+        // returns sensible enthalpy difference between equipment supply air and zone air evaluated
+        // using the humidity ratio at zone air node. This enthalpy difference multiplied by supply
+        // air mass flow rate yields the sensible heat transfer rate in Watts.
+
+        // the following two functions for calculating enthalpy difference are equivalent:
+        // PsyDeltaHSenFnTdbEquipTdbWZone() = PsyHFnTdbW(TDB2, WZone) - PsyHFnTdbW(TDB1, WZone)
+        // PsyDeltaHSenFnTdbEquipTdbWZone() function was derived by simplyfying the expression above
+        // The constant coefficients come from the equation for moist air enthalpy, PsyHFnTdbW()
+
+        return (1.00484e3 + max(1.0e-5, WZone) * 1.85895e3) * (TDBEquip - TDBZone);
```
Myoldmopar: Can you rewire `PsyDeltaHSenFnTdb2W2Tdb1W1` to call this function?  That way we don't have a copy of these coefficients lying around to maintain?

Myoldmopar (approved):
This seems quite nice from a code standpoint.  I realize the bigger review issue here is the diffs, but the code itself seems fine.  I made a few comments, and the only one that I think could really necessitate another commit is calling the second Psy function from the first.  They are both marked inline, which I know does not guarantee them to be inline, but for the one-liner function I think it could be.  And if you made that change, we wouldn't have a second set of those coefficients lying around.  I'll mark it approved because even that is probably not enough to worry about more effort on here.  If other commits are made, consider changing that too though.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Nigusse: Done!

src/EnergyPlus/Psychrometrics.hh
Nigusse: Done.

src/EnergyPlus/Psychrometrics.hh
Nigusse: Done. I will run the unit tests suite and push the branch shortly.

Nigusse:
@Myoldmopar Merged in develop, addressed review comments, run unit test suite locally and all passed.

mjwitte:
@Nigusse @Myoldmopar I made three sets of changes to this code:
1. fix lingering dWs in header for the new functions - changed them to W
2. clarify usage of the merged PsyDeltaHSenFnTdb2Tdb1W function - since it now gets called with minW for components and zoneW for zones. Renamed it, renamed arguments, and added comments.
3. use zone humrat for cpAir as appropriate - there are several places that use cpAir to make adjustments or calculate the heating/reheat coil load. It was using cpAir of the furnace inlet node - changed those to use the zone node - didn't seem to have any impact.

At this point, the results are the same as before (but I did introduce a build warning).

mjwitte:
@Nigusse @Myoldmopar I found one source of the big diffs in FurnaceWithDXSystemRHcontrol and Supermarket. See #8286 there were a few places in the furnace logic that resulted in cooling coil operation at times when the zone was in the deadband, but the system humidified more than the moisture load to dehumidifying setpoint with the coils off. See more details in #8286. I ran a local annual test suite on select files with this change and the big annual diffs are gone. See additional columns in the table at the top of this PR.  So, ideally, #8286 should get a review and drop in first, but I don't know if we have time for that.

Nigusse:
@mjwitte @Myoldmopar Looking at the humidity control plots (top of this PR) for supermarket example file, it appears this branch sometimes shows tighter Humidstat control (meeting target 50%RH) more often than develop code does. I think this a good sign that it is behaving better. I can spend sometime tomorrow to check on this example file how it behaves between this branch and PR #8286. Looking at the PR #8286 the 13 of the 15 idfs showing diffs are supermarket related. Investigating this example file is helpful in making the call in approving the fix.

mjwitte (approved):
@Nigusse I'm satisfied that these changes are valid. This will merge today. Thanks for the (many) iterations on this.

Nigusse:
@mjwitte Thank you.

PR #7869, opened 2020-03-18
Require OpenGL by default for builds

nealkruis:
Pull request overview
---------------------
OpenGL will be used by default on all machines.

Small diffs possible for ShadingTestGPU.idf on Linux CI when it uses PixelCounting instead of PolygonClipping. This will also trigger running some unit tests on Linux CI that weren't previously executed.

### Pull Request Author

### Reviewer


Myoldmopar:
This is looking very good so far.  I assume you are seeing the build warning @nealkruis.

nealkruis:
@Myoldmopar I'm on it.

Myoldmopar:
Unit test failure on Windows is not your issue, but the Linux Integration Debug build does show a failing file: integration.SolarShadingTestGPU.  Please address asap or let me know if I need to do anything.

third_party/penumbra/src/gl/context.cpp
```diff
@@ -285,9 +285,11 @@ float Context::setScene(mat4x4 sunView, const SurfaceBuffer &surfaceBuffer, bool
   near_ = -MAX_FLOAT;
   far_ = MAX_FLOAT;
 
-  for (GLuint i = surfaceBuffer.begin * model.vertexSize;
-     i < surfaceBuffer.begin * model.vertexSize + surfaceBuffer.count * model.vertexSize;
-     i += model.vertexSize) {
+  // If surface buffer has not been set use entire model instead.
+  GLuint beg = surfaceBuffer ? surfaceBuffer->begin * model.vertexSize : 0;
```
Myoldmopar, line 304: ~~I'm sure there is some sentinel signal coming through to indicate that surfaceBuffer won't be accessed (`modelSet`?), but seeing a pointer with a default value of `nullptr` being accessed without a check makes me feel things.~~

.decent_ci-MacOS.yaml
```diff
@@ -4,5 +4,5 @@ compilers:
      - IFW
      - TGZ
     release_build_cmake_extra_flags: -DBUILD_DOCS:BOOL=ON
-    cmake_extra_flags: -DBUILD_FORTRAN=ON -DBUILD_PACKAGE:BOOL=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=ON -DREGRESSION_BASELINE_PATH:PATH=$REGRESSION_BASELINE -DREGRESSION_SCRIPT_PATH:PATH=$REGRESSION_DIR -DREGRESSION_BASELINE_SHA:STRING=$REGRESSION_BASELINE_SHA -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF -DOPENGL_REQUIRED:BOOL=ON -DLINK_WITH_PYTHON=ON
+    cmake_extra_flags: -DBUILD_FORTRAN=ON -DBUILD_PACKAGE:BOOL=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=ON -DREGRESSION_BASELINE_PATH:PATH=$REGRESSION_BASELINE -DREGRESSION_SCRIPT_PATH:PATH=$REGRESSION_DIR -DREGRESSION_BASELINE_SHA:STRING=$REGRESSION_BASELINE_SHA -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF -DLINK_WITH_PYTHON=ON
```
Myoldmopar, line 7: Seeing the number of required configs go down makes me very happy.

third_party/penumbra/src/gl/context.cpp
mbadams5: @Myoldmopar It looks like to me it is doing a pointer check, just by using a ternary instead of an if statement.

Myoldmopar: Sorry, I saw the begin operator and didn't see right before it.  Thanks!

Myoldmopar:
@nealkruis @mbadams5 let me know if you've found anything.  I can chat today, though there is one long meeting right in the middle of the day.

Myoldmopar:
So deep inside penumbra, you are temporarily disabling floating point exception checks while accessing the GPU scene or view creation/clearing, because something inside there is tripping up one of the floating point environment exception flags?  And you also avoid calculating it if the pixel area is zero (which is the parallel to surface instance, right?).

I'm good with this.  CI is good with this.  Merging.

PR #7872, opened 2020-03-19
Fix heat emission HVAC relief and rejection output:variable reported values

xuanluo113:
Pull request overview
---------------------

This is to fix the wrong hourly reporting of the HVAC relief and rejection reported values from output:variable. The HVAC relief and rejection emission are calculated at HVAC time step at the reporting stage. However,  `SysTotalHVACReliefHeatLoss` and `SysTotalHVACRejectHeatLoss` calculated at the reporting functions are not always synced to the hourly/timestep variable values. 

The annual tabular report is showing the right value, as it's aggregating the calculated values of `SysTotalHVACReliefHeatLoss` and `SysTotalHVACRejectHeatLoss` on site, while the hourly reported variables are missing some HVAC steps. The monthly values are automatically aggregated by time step variables, so they are also wrong. 

The fix is to move the calculation to somewhere at `SimHVAC`. An example file of this problem and fixed results are attached.

[in.idf.txt](https://github.com/NREL/EnergyPlus/files/4355657/in.idf.txt)
[eplusout-detailed.txt](https://github.com/NREL/EnergyPlus/files/4355662/eplusout-detailed.txt)

### Pull Request Author

### Reviewer


xuanluo113:
@Myoldmopar @mjwitte Not sure if it's still possible to get this bug fixing in for 9.3. I just noticed this yesterday and contexts are explained in the pull request overview.
This may cause regression diff for monthly heat emission report.

Myoldmopar:
@xuanluo113 this is causing a file to hard crash due to an array index out of bounds (which CI reported 4 days ago).  I'm pushing this until the next release.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4790,6 +4790,45 @@ namespace OutputReportTabular {
         // The routine generates an annual table with the following columns which correspond to
         // the output variables and data structures shown.
 
+        // Using/Aliasing
```
Myoldmopar, line 4793: So the Gather function is still called from wherever it was previously called from.  This simply extracts out a portion of it into a separate Calc function which is called from the HVACManager, as well as from inside here.  OK.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -4894,17 +4919,17 @@ namespace OutputReportTabular {
...
         for (iChiller = 1; iChiller <= NumGTChillers; ++iChiller) {
             if (GTChiller(iChiller).CondenserType != WaterCooled) {
-                SysTotalHVACRejectHeatLoss += ElectricChiller(iChiller).CondenserEnergy;
+                SysTotalHVACRejectHeatLoss += GTChiller(iChiller).CondenserEnergy;
             }
         }
         for (iChiller = 1; iChiller <= NumConstCOPChillers; ++iChiller) {
             if (ConstCOPChiller(iChiller).CondenserType != WaterCooled) {
-                SysTotalHVACRejectHeatLoss += ElectricChiller(iChiller).CondenserEnergy;
+                SysTotalHVACRejectHeatLoss += ConstCOPChiller(iChiller).CondenserEnergy;
```
Myoldmopar, line 4932: Yeah these are good fixes.

src/EnergyPlus/SystemReports.cc
```diff
@@ -2285,18 +2285,6 @@ namespace SystemReports {
                                     PrimaryAirSystem(SysIndex).Name);
             }
         }
-        SetupOutputVariable("Air System Relief Air Total Heat Loss Energy",
```
Myoldmopar, line 2288: And these were setup here but that moved, which is why we have a bunch of rdd diffs.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -413,7 +413,7 @@ namespace HeatBalanceSurfaceManager {
                     Surface(SurfNum).WindDir = Node(Surface(SurfNum).LinkedOutAirNode).OutAirWindDir;
                 }
 
-                if (Surface(SurfNum).HasSurroundingSurfProperties) {
+                if (InitSurfaceHeatBalancefirstTime && Surface(SurfNum).HasSurroundingSurfProperties) {
```
Myoldmopar, line 416: This looks fine.  This would just be to avoid wasting runtime, right?

Myoldmopar (commented):
The code changes look reasonable.  Unit test?

Myoldmopar:
CI is generally really happy.  One file with big math diffs, presumably the only file that reports out the affected output variables.  I will build and test locally before proceeding any further.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
rraustad: I actually don't like seeing things like this because each time through must evaluate 2 expressions. I'd rather see:

    if (InitSurfaceHeatBalancefirstTime) {
        if (Surface(SurfNum).HasSurroundingSurfProperties) {
            ... do something
            InitSurfaceHeatBalancefirstTime = false;
        } else {  ... and sometimes
            InitSurfaceHeatBalancefirstTime = false;
        }
    }

rraustad: Wait a minute. Does this need to be InitSurfaceHeatBalancefirstTime(SurfNum) ??

Myoldmopar: @rraustad you're looking for bonus points.  

From a mental perspective, it does look good your way because you don't have to mentally process two different things as you are understanding the code.  From a code perspective however, as long as we are using the built-in boolean operators, the operator will "short-circuit", and not actually evaluate the second one if it doesn't need to.  In this case, if `InitSurfaceHeatBalancefirstTime` is false, it won't attempt to evaluate the second one because it knows it doesn't need to.  

There are more complex situations where you might have to write your own operators.  Perhaps if you wanted to compare two class instances in a very custom way, you would write your own operator.  In this case, whether it short-circuits or not is up to your own implementation.

Myoldmopar: > Wait a minute. Does this need to be InitSurfaceHeatBalancefirstTime(SurfNum) ??

I thought so too!  But I don't think it does.  It is right inside of a tight SurfNum loop, so it looks to be valid for all surfaces, one time through.

rraustad: Good explanation on the short-circuit. Compilers are smarter than I am. Maybe I need to rethink my strategy.

Myoldmopar: I had to take a small detour just for fun.  This demonstrates a small example of a custom operator.  Let's say you have a Person class that defines a bunch of attributes, but you only care about whether they are awesome.  Your custom && operator can just compare that one thing, and you can do short-circuiting pretty easily.  OK I'm heading back to work slash tornado preparation.

```
#include <iostream>

struct Person {
    bool awesome = false;
    bool operator&&(const Person &x) const
    {
        // try to short-circuit
        if (!this->awesome) {
            std::cout << "SHORT CIRCUITING\n";
            return false;
        }
        // only check the second arg if necessary
        std::cout << "COULD NOT SHORT CIRCUIT\n";
        return x.awesome;
    }
};

int main() {
    Person awesomePerson {true};
    Person notAwesomePerson {false};
    std::cout << "Comparing awesome person first -- should need to check both\n";
    if (awesomePerson && notAwesomePerson) {}
    std::cout << "Comparing not-awesome person first -- should short circuit\n";
    if (notAwesomePerson && awesomePerson) {}
}

// running this results in:
// Comparing awesome person first -- should need to check both
// COULD NOT SHORT CIRCUIT
// Comparing not-awesome person first -- should short circuit
// SHORT CIRCUITING
```

xuanluo113: Not sure I quite understand it. It should only be calculated once, but I guess not necessarily here. It only has to come after surface geometry calculation after `ViewFactorSkyIR` and `ViewFactorSkyGround` are calculated if not by user input. Perhaps at the end of  GetInputData?

Myoldmopar: It's just that we are trained to be on alert for certain patterns. When checking a global flag (`InitSurfaceHeatBalancefirstTime`) and then a "per-surface" condition (`Surf(SurfNum).a`), together, it is a clue that this loop could potentially only be run for one individual surface.  In this case, the loop is tightly packed into this single function, so I think it is fine.  But we're always watching for that sort of thing.

rraustad: @Myoldmopar is awesomePerson! But Person has a custom operator, still I think you proved your point.

rraustad: The true test is for the x10 team where the E+ function with highest call count is tested with this hierarchy. A single argument if test with nested if vs multiple argument if test to see impact of compiler short-circuiting (with first argument as most likely false). Thanks for the home schooling.

Myoldmopar:
Built and tested locally.  I see in develop that `HVAC System Total Heat Rejection Energy` is coming up always zero in the provided test file.  I see in this branch it is getting values.  Seems like this is good to go.

@rraustad @mjwitte anything before merge?

rraustad:
I don't see anything else that looks suspect. As long as the summation variables (those  using +=) give the correct result then this is good.

PR #7874, opened 2020-03-20
Fix optimum start to not override thermostat post-occupancy

### LINKED ISSUE #7866: Optimum start overriding unoccupied thermostat setpoints to occupied setpoints post-occupancy
Issue overview
--------------
This issue occurs when using the AvailabilityManager:OptimumStart object with a 24/7 applicability schedule.  When the optimum start is enabled pre-occupancy and overrides the thermostat setpoint to the occupied setpoint during the optimum start period, the setpoint remains at the occupied setpoint for the remainder of the day, when it should be resetting back to the unoccupied setpoints post-occupancy

The screenshot below shows the baseline thermostat schedule without optimum start on the left and the thermostat schedule with optimum start on the right.  The values in red show the optimum start thermostat schedules, which are remaining at the occupied setpoints until the first hour of the following day.

image

### Details
 - EnergyPlus v9.2

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #7866 
 - This pull request fixes an issue with the AvailabilityManager:OptimumStart object.  When using this object, it was observed that the post-occupancy thermostat setpoint was being overridden to the occupied setpoint post-occupancy when the setpoint should go back to it's original setback temperature.  The OptStartFlag is what controls the thermostat setpoint in the ZoneTempPredictorCorrector CalcZoneAirTempSetPoints function [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L4008).  During a separate merge (commit 832cb162f513fea6607a2b24f9c32cf6e936b0d8), the OptStartFlag was moved from being reset for each iteration to only being reset at the beginning of the day.  This was resulting in the thermostat setpoint being set to the occupied setpoint for the remainder of the day during optimum start days.  I moved the OptStartFlag to be reset for each iteration to prevent the thermostat setpoint from being set to the occupied setpoint following the fan start time and the rest of the day.  The screenshots below show the thermostat setpoints before the fix and after where you can see the optimum start turning on the system at 6:00 and setting back the temperature correctly at 19:00 after the fix.

Before Fix:
image

After Fix:
image

### Pull Request Author

### Reviewer


mitchute:
@matthew-larson do you have an example file for this? If possible, I suggest we modify the ```OptimalStart_RefBldgLargeOfficeNew2004_Chicago.idf``` so this behavior is clearly shown. Right now I'm not able to get that file to reproduce this error. All I'm able to see is that the thermostat setpoint in the modified code is being actuated one timestep before it's happening in develop.

Var_4
Var_5

matthew-larson:
@mitchute I do have an example file but I like the idea of modifying the OptimalStart_RefBldgLargeOfficeNew2004_Chicago.idf, that shouldn't be too difficult.

matthew-larson:
@mitchute In reviewing the example file and effects of the pull request, it appears to be showing the fixed behavior.  In the develop branch, the heating thermostat setpoint stays at 21C for the remainder of the day eventhough the heating thermostat schedule changes to 10C at 10pm.  In the fix branch, the heating setpoint drops down to 10C at 10pm as specified by the heating thermostat schedule as it should.  The system never runs past 10pm so the results are not actually effected by this change.  Below is a screenshot of a chart showing this behavior, and I've also attached the output spreadsheet showing the 2 models.
image
[eplusout.xlsx](https://github.com/NREL/EnergyPlus/files/4659388/eplusout.xlsx)

mitchute:
@matthew-larson thanks, now I see the changes. I propose you add the zone thermostat heating and cooling setpoint temperature outputs to the idf file, and pull develop in one more time. This should be able to go right in after that.

mitchute:
I'm not going to sweat the CI failures; develop moved before it could pick it up. The only real diffs are the AUD and MTD diffs on ```OptimalStart_RefBldgLargeOfficeNew2004_Chicago``` which are expected. Merging.

PR #7884, opened 2020-03-26
Assign Fan AirLoopNum even if fan is not yet activated

### LINKED ISSUE #7871: AirflowNetwork model terminates in Debug Mode when setting up report outputs
Issue overview
--------------
This issue was discovered when running an AirflowNetwork model in Debug mode.  The simulation terminates [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L8456) due to FanNum going out of range (equaling 2 when there is only 1 fan).  It appears that DisSysCompCVFData(FanNum).AirLoopNum is not being set prior to the ReportAirflowNetwork being called, resulting in the DisSysCompCVFData(FanNum).AirLoopNum being equal to 0.  The DisSysCompCVFData(FanNum).AirLoopNum is set in the ValidateDistributionSystem function [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L9898) but this function is being passed over when the fan is not operating during the first iteration, as noted [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/AirflowNetworkBalanceManager.cc#L497) by the AirflowNetworkFanActivated variable.  The AirLoopNum needs to be set before the ReportAirflowNetwork function whether the fan is operating or not.

### Details
 - EnergyPlus v9.2 (Debug mode)

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #7871 
 - This pull request moves the call to ValidateDistributionSystem to occur during the first iteration instead of the first time the fan is operating.  The ValidateDistributionSystem function validates the HVAC AirflowNetwork objects and sets the AirLoopNum for all the HVAC components.  I don't see any reason why this can't be called during the first iteration instead of waiting until the fan operates, which is what is causing the crash when running an AirflowNetwork model in Debug mode.

### Pull Request Author

### Reviewer


lgu1234:
@jasondegraw I will review it.

matthew-larson:
I'm keeping this as a draft as this still needs some work.

jasondegraw:
@matthew-larson That's fine. Mistakes have been made in the past, so just making sure this gets reviewed properly.

lgu1234:
@nealkruis @matthew-larson I tried to test the issue and was unable to locate a test file. Is it possible to upload the test file?

matthew-larson:
@lgu1234 We do have an internal model that we have been using and once we feel like this is a little more ready for external review, we'll upload a test file that can be used.

lgu1234:
@matthew-larson Thanks.

nrel-bot-2:
@matthew-larson @lgentile it has been 28 days since this pull request was last updated.

matthew-larson:
@lgu1234 Now that the CI checks out, I think this is ready for additional review.  Here is the file I used for debugging and testing.  It should fail in the develop branch when running in debug mode.
[afn_fan_airloopnum.txt](https://github.com/NREL/EnergyPlus/files/4657890/afn_fan_airloopnum.txt)

lgu1234:
@matthew-larson Thansk for letting me know. When I tried to run the test file, I got the error following messages:

   ** Severe  ** ProcessScheduleInput: Schedule:File="EXTERIORLIGHTS SCHEDULE", File Name="gains.csv" not found.
   **   ~~~   ** Certain run environments require a full path to be included with the file name in the input field.
   **   ~~~   ** Try again with putting full path and file name in the field.
   **  Fatal  ** ProcessScheduleInput: Preceding Errors cause termination.
   ...Summary of Errors that led to program termination:
   ..... Reference severe error count=23
   ..... Last severe error=ProcessScheduleInput: Schedule:File="EXTERIORLIGHTS SCHEDULE", File Name="gains.csv" not found.

Please upload gains.csv.

matthew-larson:
Sorry about that, forgot about the schedule:file objects.  Here is the gains file, just change the extension to csv.  Let me know if you have any other issues.
[gains.txt](https://github.com/NREL/EnergyPlus/files/4663173/gains.txt)

lgu1234:
@matthew-larson I tested the test file. The debug crash occurred when AirLoopNum is not assigned to Fan. Then I checked your addition as a new function AssignFanAirLoopNum(). I think the function is not necessary. First, this function needs to be called once, not every FirstHVACIteration = true. Second, AirLoopNum assignment is available in the function ValidateDistributionSystem, but called later.

I tried a different way to fix this issue. The revision happens in two places.

Start from line Line 8450 in the ReportAirflowNetwork function:

Current code
        for (AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
            if (DisSysNumOfCVFs == 0) continue;
            for (FanNum = 1; FanNum <= DisSysNumOfCVFs; ++FanNum) {
                if (DisSysCompCVFData(FanNum).AirLoopNum == AirLoopNum) break;
            }
My revision
        for (AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
            if (DisSysNumOfCVFs == 0) continue;
            bool FanAirLoopNumFound = false;
            for (FanNum = 1; FanNum <= DisSysNumOfCVFs; ++FanNum) {
                if (DisSysCompCVFData(FanNum).AirLoopNum == AirLoopNum) {
                    FanAirLoopNumFound = true;
                    break;
                }
            }
            if (!FanAirLoopNumFound) continue;

Start from line Line 8677 in the ReportAirflowNetwork function:

Current code
        for (AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
            if (DisSysNumOfCVFs == 0) continue;
            for (FanNum = 1; FanNum <= DisSysNumOfCVFs; ++FanNum) {
                if (DisSysCompCVFData(FanNum).AirLoopNum == AirLoopNum) break;
            }

My revision

        for (AirLoopNum = 1; AirLoopNum <= NumPrimaryAirSys; ++AirLoopNum) {
            if (DisSysNumOfCVFs == 0) continue;
            bool FanAirLoopNumFound = false;
            for (FanNum = 1; FanNum <= DisSysNumOfCVFs; ++FanNum) {
                if (DisSysCompCVFData(FanNum).AirLoopNum == AirLoopNum) {
                    FanAirLoopNumFound = true;
                    break;
                }
            }
            if (!FanAirLoopNumFound) continue;

The test file runs complete using the debug version.

No differences were found in all examples files using AirflowNetowrk objects.

matthew-larson:
@lgu1234 Thank you for testing the bug and the feedback.  You are correct in that this function should only be called once, I can make that fix.  Digging through the ValidateDistributionSystem function, I'm just wondering if some of the code that doesn't rely on the fan being activated, can be pulled out to a separate function.  The issue comes up when the ValidateDistributionSystem isn't called before the ReportAirflowNetwork function, so pulling out the fan air loop assignments made sense to me, but maybe it should include more than just the fan air loop assignments, like the nodes and other components.  The PrimaryAirSystem seems like the only array that causes a failure without the AirflowNetworkFanActivated set to true.

lgu1234:
@matthew-larson You may try to pull sections of PrimaryAirSystem out from ValidateDistributionSystem.

matthew-larson:
Thanks @lgu1234, I'll go through this exercise and update the pull request.

matthew-larson:
I took a look at pulling out sections from the ValidateDistributionSystem but it ended up getting much more complicated and beyond the scope of this bug.  I updated the function call to use `BeginEnvrnFlag` as suggested but we decided to leave the rest of the ValidateDistributionSystem function as-is pending refactoring if deemed necessary in future releases.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -491,6 +491,9 @@ namespace AirflowNetworkBalanceManager {
         // VAV terminal set only
         if (present(FirstHVACIteration) && FirstHVACIteration) VAVTerminalRatio = 0.0;
 
+        // Set AirLoop Number for fans
+        if (BeginEnvrnFlag) AssignFanAirLoopNum();
```
lgu1234: @matthew-larson BeginEnvrnFlag does not guarantee this function is called only once. It is better to have a local static variable to ensure this function is called once.

Myoldmopar: No.  Do not use a static variable, as they cause haywire when trying to reset the simulation during unit testing and other applications.  But as @lgu1234 said, this approach will call it potentially many times.  You should add a member variable to an AFN manager or appropriate location.  Or at worst, add a global variable that can be set to true once called, then easily reset during the clear_state call.

nealkruis: @lgu1234 Using local static variables is not very good practice. It isn't thread safe, and it is not good object oriented design practice. The more we can move EnergyPlus code into separate functions that are called at distinct points in the code the better (rather than relying on various flags to be set with global context). Here's some [discussion on Stack Overflow](https://stackoverflow.com/questions/15252842/is-a-bad-practice-to-declare-static-variables-into-functions-member-functions) about this.

I understand that some of these practices were inherited from limitations in Fortran, but--in my opinion--we should avoid perpetuating them if possible.

That said, there maybe no other way to prevent this from being called multiple times without a much larger refactor. This function is not going to be performance critical, so I don't see much of an issue either way.

lgu1234: @matthew-larson If a static variable is not a good practice, @Myoldmopar 's suggestion is a good way to go.

matthew-larson: Thanks everyone for the discussion, that makes sense, I'll look at adding a member variable.

lgu1234:
@matthew-larson I tested your change and found that the program was crashed in Debug mode. I also tried different ways to enhance it. Unfortunately, no successful yet.

matthew-larson:
@lgu1234 I just merged develop and fixed the conflicts so everything is up to-date.  Everything seems to be running well for me now.

lgu1234:
@matthew-larson I am going to try it again after your fix and let you know what happens.

lgu1234:
@matthew-larson I tested the program. The debug version ran the test file to completion. It is a good fix. I also tested the unit test. It worked as expected. I approve the fix.

nealkruis:
@Myoldmopar any other review needed on this before this drops in?

Myoldmopar:
Not really, it looks good to me, but just to be sure let me build and run it with latest develop pulled in.  Please hold...

Myoldmopar:
Ran locally with latest develop and it's all good.  Merging.

PR #7900, opened 2020-04-02
Reduce build time for Fortran

Myoldmopar:
- Reduce the number of transition binaries that are built by default.  Building the last 5 by default now.  That takes us back to the 8.8 to 8.9 transition version which should cover IDFs from 2017.
- Add CMake flag to allow building all of them to enable updating files from prior to 2017.
- Turn on the CMake flag for CI, since we want to at least build (and soon exercise) those binaries all the way back to 1.0 on CI.


src/Transition/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ set(LIB_SRC
 )
 add_library( TransitionLib STATIC ${LIB_SRC} )
 
+set(BUILD_OLD_TRANSITIONS OFF CACHE BOOL "Only turn this on if you really want to build all the old transition tools")
```
Myoldmopar, line 27: Add CMake cached variable that allows building ALL the transition binaries.  It is OFF by default because surely 99% of dev work will not need to build back farther than the 2017 default oldest version.  To build all of them, just flip this switch to ON and you're all set.

src/Transition/CMakeLists.txt
```diff
@@ -24,6 +24,9 @@ set(LIB_SRC
...
+
+if (BUILD_OLD_TRANSITIONS)
```
Myoldmopar, line 29: Only include this first chunk if the flag is ON.

src/Transition/CMakeLists.txt
```diff
@@ -55,6 +58,7 @@ list( APPEND VERSIONS 8_4_0 )
 list( APPEND VERSIONS 8_5_0 )
 list( APPEND VERSIONS 8_6_0 )
 list( APPEND VERSIONS 8_7_0 )
+endif()
 list( APPEND VERSIONS 8_8_0 )
```
Myoldmopar, line 62: Always include from this version on.  We may adjust this again later.

src/Transition/CMakeLists.txt
```diff
@@ -136,42 +140,19 @@ foreach( i RANGE 1 ${end} )
...
-install(FILES "InputRulesFiles/Rules7-2-0-to-8-0-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-0-0-to-8-1-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-1-0-to-8-2-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-2-0-to-8-3-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-3-0-to-8-4-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-4-0-to-8-5-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-5-0-to-8-6-0.xls" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-6-0-to-8-7-0.md" DESTINATION "PreProcess/IDFVersionUpdater")
-install(FILES "InputRulesFiles/Rules8-7-0-to-8-8-0.md" DESTINATION "PreProcess/IDFVersionUpdater")
 install(FILES "InputRulesFiles/Rules8-8-0-to-8-9-0.md" DESTINATION "PreProcess/IDFVersionUpdater")
```
Myoldmopar: We had arbitrarily picked 7.2 as the original "oldest" version to package up with E+.  I think it's suitable to move on from there.  I am including the last 5 versions here, but I will caution that I am considering making the Transition a fully built standalone version for the releases.  My idea would be to include the latest few versions in E+ as here, but then make an additional CI config for each platform that will build transition with all the versions and post them as release assets.  So don't get too hung up on the absolute number here.

rraustad:
An obvious question is "how much time does this save"?

Myoldmopar:
For Visual Studio users, maybe not much.  But for makefile based builds, once Fortran is turned on, it can cause a lot of rebuild time, and I mean, how many times does anyone run `Transition-V1-0-0-to-V1-0-1` 🤣

mitchute:
One convenience request I have related to this is if we could add a "fortran" build target that builds all of the other Fortran projects that are turned on? Similar to what is done with docs.

Myoldmopar:
@mitchute that would be a nice thing, though we'll have to consider the balance of effort as we try to also push to get rid of Fortran stuff.  I think it's probably worth it, but I'm not volunteering to do it, either.

Myoldmopar:
Anyone have issues/concerns with this?

src/Transition/CMakeLists.txt
mjwitte: @Myoldmopar My only concern here (minor, but likely to come up at some point) is that we have all of the transitions up to 7.1 zipped up in the helpdesk download area. So, once this becomes part of a release (I'm assuming we'll leave this out of the pending bugfix release) there will be a gap. So we need to be sure we either extend those or have the new Transition package ready at the same time.

Myoldmopar: That's a really great point.  I'm going to take out the install changes.  That way the only thing this PR does is modify the build, and primarily just for us make-file-based builds.  Thanks.

Myoldmopar:
Alright, reverted out the install changes @mjwitte.  This is now purely a change to avoid building all the old Transition binaries unless you really want to.  CI is set to build them all to keep us from accidentally breaking an old transition build without noticing.  As long as CI comes back happy, this should be good.  Let me know if you have any other concerns.

Myoldmopar:
@mjwitte you feel OK with this set of changes now?  Doesn't affect the install at all, just helps us makefile-based builders by skipping all the really old transitions unless we really want them.

Myoldmopar:
I'll take your thumbs up as approval :rofl:.  Merging.  Thanks!

PR #7921, opened 2020-04-09
Add a CreateIDFUsingNewRule template, bump to 9.4.0 and add Transition files for that

jmarrec:
Pull request overview
---------------------

* Add a CreateIDFUsingNewRule template, bump to 9.4.0 and add Transition files for that
* Automatically add `Version` object if missing in `EnergyPlusFixture::process_idf` to avoid constant warnings
* Remove `Version` in all tests where it doesn't make sense (all but the IdfParser / InputProcessor)
* Add a .git-blame-ignore-revs file
    * Whenever you use "git blame", you see a ton of lines blames because of that stuff while it's actually irrelevant to what you're usually trying to get at. This is opt-in, and isn't used by github currently. Locally you can do $ git config blame.ignoreRevsFile .git-blame-ignore-revs to register it.

### Pull Request Author

### Reviewer


tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -833,7 +868,10 @@ TEST_F(InputProcessorFixture, parse_idf_extensible_blank_extensibles)
                          {"temperature_convergence_tolerance_value", 0.4000},
                          {"solar_distribution", "FullExterior"},
                          {"maximum_number_of_warmup_days", 25},
-                         {"minimum_number_of_warmup_days", 6}}}}}};
+                         {"minimum_number_of_warmup_days", 6}}}}},
+                     {"Version",
+                      {{"",
+                        {{"version_identifier", "9.4"}}}}}};
```
Myoldmopar: I don't mind the removal of version strings from unit tests where they aren't needed.  But what happened to make these need to be here?  The change_version script won't find these, so that'll either need to be modified or we'll just need to update these manually.

rraustad:
Err file for RefBldgMediumOfficeNew2004_Chicago_epJSON

    +   ** Warning ** Version: in IDF="9.3" not the same as expected="9.4"
    +   ** Warning ** Version: in IDF="9.3" not the same as expected="9.4"

tst/EnergyPlus/unit/InputProcessor.unit.cc
jmarrec: Same reason as why `Building` and `GlobalGeometryRules` are added to all tests there: because if not present, they are automatically added by `process_idf`. So by adding them explicitly, you can better control the way the roundtrip is done. 

You bring a good point though: I will bump to 9.5 locally and see if there's a need to update these or not (it may throw a warning for expected version...), in which case I'll use `DataStringGlobals::MatchVersion` to avoid the problem.

**Edit**: After bumping to 9.5, I did use DataStringGlobals

mbadams5: `Building` and `GlobalGeometryRules` are automatically added if they are not present because they are the only two required objects in all of E+. They must be present for an IDF to be valid. `Version` is not automatically added by `process_idf` but it does check to see if version numbers match during `process_idf`. Why does `Version` need to be added here? I personally don't mind since it makes it more explicit and unit tests will fail if they are not updated properly if something changes.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -577,6 +606,12 @@ TEST_F(InputProcessorFixture, parse_bad_utf_8_json_3)
           "\"starting_vertex_position\":\"UpperLeftCorner\","
           "\"vertex_entry_direction\":\"Counterclockwise\""
         "}"
+      "},"
+      "\"Version\":{"
+        "\"\":{"
+          "\"idf_order\":0,"
```
mbadams5, line 612: You shouldn't need this anymore and if you do for the test to pass then I consider that a bug.

tst/EnergyPlus/unit/InputProcessor.unit.cc
jmarrec: Because I made `process_idf` add Version automatically in this PR. The reasoning is that otherwise you automatically get a warning:

```
** Warning** Version: missing in IDF, processing for EnergyPlus version=\"" + MatchVersion + "\""
```

which is annoying when you are calling `compare_err_stream` later on, and the major reason why a bunch of unit tests had an explicit `Version` object in there, which required us to bump the version in these units tests when bumping the `cmake/Version.cmake`.

I'm not sure I'm really clearly stating the above, hopefully I am?

tst/EnergyPlus/unit/InputProcessor.unit.cc
jmarrec: https://github.com/NREL/EnergyPlus/blob/b86e836101bd67a64d95b8b889ff508fa2f8251f/tst/EnergyPlus/unit/InputProcessor.unit.cc#L610-L618

jmarrec: `process_idf` will add the Version object, so I do need to add it to the `expected` much like GlobalGeometryRules. Why is that a bug?

mbadams5: @jmarrec I mean specifically `"\"idf_order\":0,"` isn't needed. I know it was still in the GlobalGeometryRules and Building objects that are added automatically, but I forgot to remove those once those fields were no longer necessary. If `process_idf` requires `"idf_order":0` still after you remove it from that function, that is what I consider a bug.

jmarrec: Ah ok thanks for the clarification. I copied the idf_order over without second thought here

jmarrec: @mbadams5 I removed the `idf_order` for Building, GlobalGeometry, and Version in this test, and it now fails:

```
[ RUN      ] InputProcessorFixture.parse_bad_utf_8_json_3
/home/julien/Software/Others/EnergyPlus/tst/EnergyPlus/unit/InputProcessor.unit.cc:614: Failure
      Expected: expected
      Which is: "{\"Building\":{\"\xEF\xBF\xBD\xEF\xBF\xBD\xEF\xBF\xBD\":{\"idf_max_extensible_fields\":0,\"idf_max_fields\":8,}},\"GlobalGeometryRules\":{\"\":{\"coordinate_system\":\"Relative\",\"daylighting_reference_point_coordinate_system\":\"Relative\",\"rectangular_surface_coordinate_system\":\"Relative\",\"starting_vertex_position\":\"UpperLeftCorner\",\"vertex_entry_direction\":\"Counterclockwise\"}},\"Version\":{\"\":{\"version_identifier\":\"9.4\"}}}"
To be equal to: input_file
      Which is: "{\"Building\":{\"\xEF\xBF\xBD\xEF\xBF\xBD\xEF\xBF\xBD\":{\"idf_max_extensible_fields\":0,\"idf_max_fields\":8,\"idf_order\":1}},\"GlobalGeometryRules\":{\"\":{\"coordinate_system\":\"Relative\",\"daylighting_reference_point_coordinate_system\":\"Relative\",\"idf_order\":0,\"rectangular_surface_coordinate_system\":\"Relative\",\"starting_vertex_position\":\"UpperLeftCorner\",\"vertex_entry_direction\":\"Counterclockwise\"}},\"Version\":{\"\":{\"idf_order\":0,\"version_identifier\":\"9.4\"}}}"
```

Building has idf_order = 1, which is added by the parser itself. The other two are getting it from the `process_idf` here: https://github.com/NREL/EnergyPlus/blob/b86e836101bd67a64d95b8b889ff508fa2f8251f/tst/EnergyPlus/unit/Fixtures/EnergyPlusFixture.cc#L276-L300

jmarrec:
Rebase onto develop resolving conflicts and force pushed. Update your refs as appropriate.

Myoldmopar:
> force pushed

Unless you are talking about a force push involving midi-chlorians, please don't force push to branches in our repo.  (You are welcome to do whatever you want in your fork of course.  And also I should be clear I'm happy to have you push branches directly to this repo too, as long as they aren't forced.)

I pulled develop into this branch and ran unit tests.  I fixed the two files that had wrong versions.  I pushed that up and I'll let CI have a final pass at it.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,451 @@
+# This file includes the list of commits to exclude from `git blame`.
```
Myoldmopar, line 1: Well I wasn't expecting to see this file.  But I'm not opposed to it.  It looks like an interesting addition.  I see this [feature](https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt#L83-L84) was added to Git on April 20, and I see that Github isn't actually using it yet, but that some local clients are using it.  I think it will definitely give a better look at the actual blame and changes, for which I am in support.

cmake/Version.cmake
```diff
@@ -1,10 +1,10 @@
 
 set( CMAKE_VERSION_MAJOR 9 )
-set( CMAKE_VERSION_MINOR 3 )
+set( CMAKE_VERSION_MINOR 4 )
 set( CMAKE_VERSION_PATCH 0 )
 
 
-set( PREV_RELEASE_SHA "921312f" )
+set( PREV_RELEASE_SHA "baff089" )
```
Myoldmopar, line 7: :+1:

Myoldmopar (commented):
I can't see the whole diff here, so I'll be relying on CI and local testing.  I'll locally check out some of the primary changed transition files to inspect.

Myoldmopar:
9 additional ERR diffs due to new IDFs since this was merged.  I'll fix those now.

jmarrec:
> > force pushed
> 
> Unless you are talking about a force push involving midi-chlorians, please don't force push to branches in our repo. (You are welcome to do whatever you want in your fork of course. And also I should be clear I'm happy to have you push branches directly to this repo too, as long as they aren't forced.)
> 
> I pulled develop into this branch and ran unit tests. I fixed the two files that had wrong versions. I pushed that up and I'll let CI have a final pass at it.

Noted that is your preference, I'll merge develop in instead.

 (It is really a matter of preferences whether you'd rather have a clean history (rebasing & force-pushed) or prefer having multiple "merge develop in PR" commits. For example I recently made a PR to CMake and they aggressively make you rebase and rewrite history.)

Myoldmopar:
That looks really good now.  The fact that there are no ERR diffs is encouraging.  I will wait until tonight to merge this so that it doesn't throw diffs all evening.  Then over the weekend I can pull develop into any branches necessary and let them re-run.

Myoldmopar:
And let's go to version 9.4.  Thanks @jmarrec.

PR #7927, opened 2020-04-14
API FollowUp and Testing

### LINKED ISSUE #7888: Problems linking to API
Issue overview
--------------
While developing some automated package testing code, I encountered a problem while exercising the workflow of linking to the API.  Note: Python Plugins workflows work just fine, and properly exercise the APIs.  However, when trying to link to the API from Python is causing a problem.  

As a bit of a backstory, when I build the EnergyPlusAPI target, I dynamically link to the Python shared library, which is brought into install tree and included in the package.  I run the `install_name_tool` on the EnergyPlusAPI target by using GetPrerequisites on the EnergyPlus(.exe) binary and finding the API target inside there. 

 - [ ] Which means step 1 is: Just execute on the API target directly, EnergyPlus itself should not be dependent on the Python lib anyway.  This needs cleanup.

When I run `install_name_tool`, I am changing the dependency on /some/path/to/PythonLibNameHere to `@executable_path/PythonLibNameHere`, and then copying PythonLibNameHere to the install tree.  After installing EnergyPlus, during a Python Plugin workflow, the executable being run is energyplus(.exe), and so the @executable_path/PythonLibNameHere works just fine.  However, trying to link a Python script to the API is problematic, because the executable_path is now the Python interpreter somewhere on the system.

An underlying problem here is that API workflows DO NOT depend on the EnergyPlus python lib packaged up.  Calling the EnergyPlusAPI from some other Python should NOT end up causing a dependence on that Python library.  

- [ ] So this is the second thing.  We need to figure out how to do this:

 - One option is to break EnergyPlusAPI into a couple pieces.  One part that does NOT depend on Python, and can be built apart from the Python lib dependency, and the other part that DOES depend on Python.  The EnergyPlus(.exe) binary will depend on both, since it will call both things.  But calling the Python EnergyPlus API wrapper would never hit the library that depends on Python, so no problem.  
 - Another option is to `dlopen` inside EnergyPlus to the Python dynamic library.  This way we wouldn't have an explicitly declared dependency that would fail on calling the API.  The `dlopen` calls would only be made if we aren't running in API/Library mode.  This is actually a pretty clean approach.
 - Another option is to figure out how to statically link in Python on all platforms.  This would result in the most ideal package since it would not include a Python dynamic library at all.

In order to hit this error, I was installing EnergyPlus into a blank machine image, then creating a small Python script that instantiated the API and tried to make function calls.  During API instantiation, the Python wrapper goes out and loads in the C API functions, which are in the EnergyPlusAPI target.  Unfortunately this depends on the Python library, which is how the following errors are manifested:

```
OSError: dlopen(/tmp/EnergyPlus-9.3.0-baff08990c-Darwin-x86_64/libenergyplusapi.dylib, 6): Library not loaded: @executable_path/Python
  Referenced from: /tmp/EnergyPlus-9.3.0-baff08990c-Darwin-x86_64/libenergyplusapi.dylib
```

On Windows, I am providing the API DLL of course, ~~but not the .lib file~~ AND the .lib file.  I am now linking to the C API in both eager and lazy modes.  Using Python works great as long as you add the E+ install folder to the PATH (and set PYTHONPATH to the E+ install).

Linux is working happily, but it could just be luck -- I'm really not sure yet.  In general, we just need to tighten this up substantially.

A nice status summary as of 3/30/20 is available here: https://github.com/Myoldmopar/EPTravisTester/pull/4#issuecomment-606210378

Useful links:
 - https://wincent.com/wiki/%40executable_path%2C_%40load_path_and_%40rpath
 - https://gregoryszorc.com/blog/2018/12/18/distributing-standalone-python-applications/
 - https://medium.com/@donblas/fun-with-rpath-otool-and-install-name-tool-e3e41ae86172

Myoldmopar:
Pull request overview
---------------------
 - Fixes #7888 

## Detailed Changes Include:
 - Fleshed out unit tests for API code, only running on Mac/Linux
 - Fixed issues with getting variables
 - Fixed issues with actuator searching
 - Add `get_construction_handle` method to Python API
 - Create a Python DLL wrapper so that the E+ DLL does not have a hardwired dependency on the Python DLL itself
 - Eliminate throws because that's not valid on declspec'd functions, add error flag supporting functions to API
 - Fix `listAllCSVData` function and usage to avoid memory leak
 - Adds Python plugin integration into the plant user defined op scheme
 - Fixup API docs
 - RANDOM: Changed Github auth type so that it won't throw warnings during package builds


src/EnergyPlus/api/datatransfer.cc
```diff
@@ -105,8 +104,13 @@ const char * listAllAPIDataCSV() {
         output.append(variable.VarNameOnly).append(",");
         output.append(variable.KeyNameOnlyUC).append("\n");
     }
-    // add output vars and meters
-    return output.c_str();
+    // note that we cannot just return a c_str to the local string, as the string will be destructed upon leaving
+    // this function, and undefined behavior will occur.
+    // instead make a small copy -- I realize it could cause its own issues, but I would not expect users to be making
+    // many many calls to this.  And if they did, just free the memory and be done.
+    char *p = new char[std::strlen(output.c_str())];
```
mbadams5, line 113: @Myoldmopar This is going to be ripe for creating a memory leak. The caller needs to call `delete` on this allocated memory. Does Python take ownership of the returned `const char *`? I know in your unit tests there is a memory leak when implicitly converting to `std::string`.

tst/EnergyPlus/unit/api/datatransfer.unit.cc
```diff
@@ -0,0 +1,608 @@
...
+};
+
+TEST_F(DataExchangeAPIUnitTestFixture, DataTransfer_TestListAllDataInCSV)
+{
+
+    std::string const idf_objects = delimited_string({"Version, 9.3.0;"});
+    ASSERT_TRUE(process_idf(idf_objects, false)); // this had to be here or I was getting a strange segfault during a JSON string dtor
+
+    // first off, the function should return, even if there isn't anything meaningful in it (it will have headers)
+    std::string csvDataEmpty = listAllAPIDataCSV();
```
mbadams5: @Myoldmopar This is a memory leak. You implicitly convert to `std::string` but it does not take ownership of the returned `const char *`. You no longer have a pointer to that allocated memory and cannot `delete` it. That said, it isn't probably a huge deal for these unit tests but shows how easy it is to leak memory.

Myoldmopar:
I'm wondering if anyone could offer any thoughts on this.  Mac/Linux are very happy (**100%!**) , but Windows is failing.  I can reproduce the failure locally, but I'm baffled tonight as to what is going on.

A caveat: I know I need to clean up the declspec/throw mismatch, but I don't believe that is affecting this function since it doesn't attempt a throw anyway.

Now, for the actual problem.  The unit tests are failing all over the place.  I can debug the `DataTransfer_TestGetVariableHandlesRealTypes` unit test and step along and see it breaking.  The unit test is really simple, so I can step through the process:

1. When the unit test starts, I can inspect the underlying value of the output variable array: `OutputProcessor::RVariableTypes.data_`, and it is properly null.
2. Two output variables are "requested" in my unit test fixture, and then they are properly set up with `OutputProcessor::SetupOutputVariable` calls.
3. Back in the unit test, I can again check that `RVariableTypes.data_` variable, and now it shows up with a meaningful pointer.  All is well!
4. In the next call, my unit test calls into the datatransfer.cc function `getVariableHandle`, and as soon as the code goes into that frame, the value of `RVariableTypes.data_` goes BACK TO NULL.  
5. That function will fail to find a match because for some reason the output processor struct appears empty, so it returns -1.
6. But then as soon as it goes back into my unit test code, the value of the value pointer goes BACK TO THE RIGHT VALUE.

This same code works perfectly fine on the other platforms.  I will continue to investigate, and I know there are still a few things to tidy up (a stray debug print statement coming out on stdout, and the invalid throws, ...), but I would love if someone had any thoughts on this odd behavior.  It is as if the OutputProcessor namespace is being masked temporarily when we step into this other code.

Perhaps @mbadams5 @lefticus @kbenne @nealkruis @mitchute @jasondegraw ?

lefticus:
> 
>     4. In the next call, my unit test calls into the datatransfer.cc function `getVariableHandle`, and as soon as the code goes into that frame, the value of `RVariableTypes.data_` goes BACK TO NULL.
> 
>     5. That function will fail to find a match because for some reason the output processor struct appears empty, so it returns -1.
> 
>     6. But then as soon as it goes back into my unit test code, the value of the value pointer goes BACK TO THE RIGHT VALUE.
> 
> 
> This same code works perfectly fine on the other platforms. I will continue to investigate, and I know there are still a few things to tidy up (a stray debug print statement coming out on stdout, and the invalid throws, ...), but I would love if someone had any thoughts on this odd behavior. It is as if the OutputProcessor namespace is being masked temporarily when we step into this other code.
> 

I'm just now joining this, but the only logical explanation is that you are dealing with two different `RVariableTypes.data_` objects.

How is this object shared between the API and the main runtime? Is it a DLL layer? Should there be an `__dllexport()` enabled for windows?

Myoldmopar:
Let me try to add a little more detail here.

We have a bunch of EnergyPlus source, let's call it "OutputProcessor.cc and Friends".  This is used to build the static library libenergyplus as an intermediate step in the build process.

We then have a handful of additional files of source that are used to build an API, let's call it 
"DataTransfer.cc and Friends".  This is used to build the shared library libenergyplusapi.  This shared library is distributed with the install package.  The source in this group has a plentiful amount of `__declspec()` decorated functions.

We then have a main function that is used to build the final energyplus executable, which links to and depends on the libenergyplusapi shared library.

As for testing, we have a ton of unit test source code, let's call it "DataTransfer.unit.cc and Friends", which is used to build the energyplus_tests gtest-backed executable.  

Now, in my mind, the unit test executable had to be linked statically to the libenergyplus static library because we didn't expose all the guts of EnergyPlus over an API on the shared library.  However, when I look at the unit test CMake rules, I see `set(test_dependencies energyplusapi)` used which indicates it is linking to the shared library.  There may be some CMake magic going on because I don't see it declared as a shared link anywhere.  So I'm kinda assuming it is linking to the static lib still.  Obviously there is still a little uncertainty here.

But here's the deal.  If we are linking the unit test code to the static library as I assume, then the API code (DataTransfer and Friends) shouldn't be accessible....at all.  It's not built into the static library, it's only built as the API layer on top of the shared library.  If my assumption is right, then how is the unit test code finding the DataTransfer and Friends code?

If there is something going on with how the linker is finding the functions, perhaps it then loads a new image of the lib at runtime and that's why we see two different representations of the underlying EnergyPlus static lib code (OutputProcessor and Friends).  And that's why when I switch from one function to another, it is as if the shadow version suddenly pops up.

OK, not sure if that's actually going to help, but maybe therapeutically I'll benefit from writing it all out as I currently understand it.  Time to dig back in.

lefticus:
The `set(test_dependencies energyplusapi)` may only be setting a build order dependency, not an actual linkage? But that is easy to inspect if you pull up the target in VS and look at the linker flags.

I want to ask if any of these objects are global statics, but that's a silly question given the history of E+. The real question is, is `RVariableTypes.data_` or anything it relies on a global static?

I ask because it took me on the order of weeks of full time work to sort out the singletons shared between DLL's in OpenStudio.

 * Linux: worked fine
 * Windows: each DLL got its own Logger singleton
 * MacOS: worked fine, but double free on shutdown

And everything your telling me says your dealing with a similar situation. It's not necessarily the function (`__dllexport`) that needs to be exported, but the actual object? (again this only applies if it's a `static`).

This is the ancient and forgotten result of that history: https://github.com/NREL/OpenStudio/blob/develop/src/utilities/core/Logger.hpp#L140-L142

I don't really recall what all needed to happen, but could probably remember if I needed to.

Myoldmopar:
@lefticus you could totally be right.  I'm not able to say yet.  After a morning session with @mbadams5 I'm still not sold on exactly what is happening between the static lib, shared lib, and unit test binary.  Feels like there may be some CMake voodoo.

One final update: up until now, the unit tests have been put in to test the internal guts of EnergyPlus, including a major amount of function calls in EnergyPlus, and of which none of them are exposed over the shared lib API.  So the unit tests should/must be linking to the static library, otherwise it wouldn't have any way to get into the guts of the shared library and access things that aren't exported.  (optional prepositional phrase: ...on Windows.)

What I've done with this branch is I've added unit test code that is calling code that is only built in the shared library.  Yet I'm still adding it to the existing unit test executable.  So for one thing, I don't exactly know where the linker is getting everything, and for another thing, I'm not so sure I should be doing this anyway.  It's like the very hard link to the static library is working fine, and then when it gets to the calls to the shared library, it panics because it doesn't know where to find it, somehow magically finds it and instantiates it, but it creates its own underlying global state where addresses are different and down is up and Nibor leads the bizarro titans. :shrug: 

For now, I've decided that these new "unit" tests that exercise the API should be kept separate.  I already created a small C program that exercises a few API functions in a totally separate executable environment, nothing to do with GTest.  I will likely be converting these API "unit" tests over to that C API test program.  That separate binary is linked directly to the shared library, and eliminates this weird cross talk going on with the GTest built unit tester.  I think this path will actually make for a better precedent.  When "unit testing" we should be testing things internal to E+, but when wanting to test a function exposed over the API, it should be added to the separate API test.

Note: We could possibly add a second unit test binary called `energyplus_api_tests.exe` that is built using GTest, but *definitely* links to the shared library.  I dunno.  Not worrying about it anymore for now.

Thanks for the time those of you have spent trying to figure this one out.

src/EnergyPlus/api/datatransfer.h
```diff
@@ -391,6 +391,13 @@ ENERGYPLUSLIB_API Real64 currentTime();
 /// \see apiDataFullyReady
 ENERGYPLUSLIB_API int minutes();
 
+/// \brief Returns the duration of the current zone simulation time step, in fractional hours
+/// \details The zone time step will vary throughout the simulation as needed to maintain convergence while being cautious about program runtime.
+///          This function returns the current value of the time step
+/// \remark The behavior of this function is not well-defined until the `apiDataFullyReady` function returns true.
+/// \see apiDataFullyReady
+ENERGYPLUSLIB_API Real64 zoneTimeStep();
```
rraustad, line 399: Not sure of the intent here but TimeStepZone is fixed throughout the simulation (contrary to the comments here). It's TimeStepSys that varies and SysTimeElapsed keeps track of the sub zone time step to the point where SysTimeElapsed = TimeStepZone.

image

rraustad: OK, so the next API function, systemTimeStep, returns the system time step so it's just the comments that are wrong above. And maybe a user would need to know the progression of the simulation through the sub zone time step, e.g., ENERGYPLUSLIB_API Real64 systemTimeStepElapsed();

src/EnergyPlus/PluginManager.cc
```diff
@@ -55,41 +61,355 @@
...
+        // When calling E+ as a library, the plugin system is not available
+        // So when calling for plugins, the user must be calling as energyplus.exe
+        // In this case, the current executable path will be the E+ install root
+        // Which is where we find the Python Wrapper Shared Library
+        if (wrapperDLLHandle) return; // already found
+        std::string programPath = FileSystem::getProgramPath();
+        std::string programDir = FileSystem::getParentDirectoryPath(programPath);
+        std::string sanitizedProgramDir = PluginManager::sanitizedPath(programDir);
+        // set the path to the python library
+#ifdef _WIN32
```
rraustad, line 134: Don't need to protect _WIN64 ?

src/EnergyPlus/api/datatransfer.h
mitchute: > Not sure of the intent here but TimeStepZone is fixed throughout the simulation (contrary to the comments here). It's TimeStepSys that varies and SysTimeElapsed keeps track of the sub zone time step to the point where SysTimeElapsed = TimeStepZone.
> 
> image

Thanks @rraustad. The intent was to give the user access to the same variables as shown in the EMS App Guide, Table 2.2. I can update the comments to note the fixed timestep.

src/EnergyPlus/api/datatransfer.py
```diff
@@ -669,3 +736,19 @@ def current_environment_num(self) -> int:
         :return: The current environment number.
         """
         return self.api.currentEnvironmentNum()
+
+    def actual_time(self) -> RealEP:
```
mitchute: Blah. Return value type hint should be ```int```. Same on the next one. Let me know if you want me to fix that.

src/EnergyPlus/PluginManager.cc
Myoldmopar: @rraustad good eye, but _WIN32 is just a broad preprocessor definition that applies to multiple Windows platforms.  From the [definition](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019): 

> _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.

So it will be defined for our x86 or x64 builds.

src/EnergyPlus/api/datatransfer.py
Myoldmopar: I'll fix it now, I'm doing some other work here anyway.

.gitignore
```diff
@@ -27,7 +27,8 @@ doc/title.tex
 Energy+.schema.epJSON.in
 
 # ignore the doxygen built API docs
-src/EnergyPlus/api/_build
+src/EnergyPlus/api/docs/c/_build
+src/EnergyPlus/api/docs/python/_build
```
Myoldmopar, line 31: Adding some paths to ignore when doing api documentation builds.

CMakeLists.txt
```diff
@@ -21,6 +21,11 @@ include(cmake/Version.cmake)
 
 set( OPENGL_REQUIRED ON CACHE BOOL "Require OpenGL in order to build" )
 if(OPENGL_REQUIRED)
+  # On Linux specifically, OpenGL is complaining about being able to find OpenGL using both the
+  # exported module target or the library variables.  If it has to choose, and doesn't have a user-preference,
+  # a warning is shown on every configure step.  This next line just sets the preference so the warning won't show up.
+  # See this for more info: https://cmake.org/cmake/help/v3.10/module/FindOpenGL.html#linux-specific
+  set(OpenGL_GL_PREFERENCE GLVND)
```
Myoldmopar, line 28: I kept getting a warning on Linux, this just sets a value for the default GL preference, and should only cause the warning to go away.  See the comment for more info if needed @nealkruis @mbadams5

CMakeLists.txt
```diff
@@ -34,6 +39,7 @@ option( BUILD_PACKAGE "Build package" OFF )
 option( BUILD_TESTING "Build testing targets" OFF )
 option( BUILD_FORTRAN "Build Fortran stuff" OFF )
 option( BUILD_DOCS "Build LaTeX-pdf documentation" OFF )
+option( LINK_WITH_PYTHON "Link with Python Library for Python Plugin Builds" OFF )
```
Myoldmopar, line 42: Moved this to the top level so it can be accessed by other subdirectories as needed.

CMakeLists.txt
```diff
@@ -285,3 +291,17 @@ mark_as_advanced(FMT_INSTALL)
...
+mark_as_advanced(GLFW_USE_OSMESA)
+mark_as_advanced(GLFW_USE_WAYLAND)
+mark_as_advanced(GLFW_VULKAN_STATIC)
+mark_as_advanced(BUILD_PENUMBRA_TESTING)
+mark_as_advanced(PENUMBRA_COVERAGE)
+mark_as_advanced(PENUMBRA_STATIC_LIB)
+mark_as_advanced(BUILD_GMOCK)
+mark_as_advanced(BUILD_GTEST)
+mark_as_advanced(gmock_build_tests)
+mark_as_advanced(BUILD_WCE_TESTING)
```
Myoldmopar, line 307: Just cleaning up the list of cmake variables.

cmake/PythonSetupAPIinBuild.cmake
```diff
@@ -17,7 +17,6 @@ configure_file( "${API_SOURCE_DIR}/common.py" "${API_TARGET_DIR}/common.py" )
 configure_file( "${API_SOURCE_DIR}/datatransfer.py" "${API_TARGET_DIR}/datatransfer.py" )
 configure_file( "${API_SOURCE_DIR}/api.py" "${API_TARGET_DIR}/api.py" )
 configure_file( "${API_SOURCE_DIR}/func.py" "${API_TARGET_DIR}/func.py" )
-configure_file( "${API_SOURCE_DIR}/datatransfer.py" "${API_TARGET_DIR}/datatransfer.py" )
```
Myoldmopar, line 20: This line was duplicated, removing.

scripts/dev/find_byref_bool_override.py
```diff
@@ -43,6 +43,11 @@
                          'src/EnergyPlus/test_ep_as_library.cc',
                          'EnergyPlusPgm.cc']
 
+EXPECT_MISSING_NAMESPACE = [
+    'src/EnergyPlus/PythonLibWrapper.cc',
+    'src/EnergyPlus/PythonLibWrapper.hh'
+]
```
Myoldmopar, line 49: Telling this script to ignore the missing namespace on functions when parsing the code looking for suspicious byref bool overrides.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -727,8 +726,15 @@ endif()
 if (WIN32)
   target_link_libraries( energypluslib Shlwapi )
 endif()
+
+# we will build the intermediate wrapper DLL that hard links to Python DLL here
```
Myoldmopar: Here is where we build the intermediate library.  Here's the build process:

- Build EnergyPlus static library which houses all the internal guts of EnergyPlus
- Build PythonWrapper which hard-links to the Python DLL and exposes functionality
- Build EnergyPlus shared library which does not hard link to the PythonWrapper, but rather delay-links to it at runtime, only if needed.  This shared library wrapper just takes the static lib and adds some API exposed functions.
- Build EnergyPlus executable which hard links to the EnergyPlus shared library.  This executable just contains basically a main function.
- If doing testing, the GTest unit test executable is built and linked to the EnergyPlus library -- which I assume to be the static library, though there seems to be stuff going on by CMake.

(If you are looking at this section as soon as I write this comment, you'll see I am actually linking the energypluslib to the pythonwrapper in CMake, which is counter to what I wrote above.  I am about to push a commit that changes the hard link and makes the python wrapper a simple build dependency for energyplusapi.)

src/EnergyPlus/CMakeLists.txt
```diff
@@ -777,11 +783,11 @@ set_target_properties(energyplusapi PROPERTIES VERSION ${ENERGYPLUS_VERSION})
 
 # bring in the Python API layer right next to the executable
 add_custom_command(
-        TARGET energyplus
+        TARGET energyplusapi
```
Myoldmopar, line 792: Setting up the API wrapper is a closer match for the shared library not the executable.  Doesn't really change anything but makes it more appropriate.

src/EnergyPlus/DisplayRoutines.cc
```diff
@@ -93,6 +93,8 @@ void DisplayString(std::string const &String) // String to be displayed
 
     // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
     // na
+    if (DataGlobals::fMessagePtr) DataGlobals::fMessagePtr(String);
+    if (DataGlobals::messageCallback) DataGlobals::messageCallback(String.c_str());
```
Myoldmopar, line 97: `DisplayString()` was not using the callbacks, this just reports those strings out to the callbacks when needed.

src/EnergyPlus/Plant/EquipAndOperations.hh
```diff
@@ -141,6 +141,8 @@ namespace DataPlant {
         int ReferenceNodeNumber;        // DELTA CTRL ONLY--for calculation of delta Temp
         int ErlSimProgramMngr;          // EMS:ProgramManager to always run when this model is called
         int ErlInitProgramMngr;         // EMS:ProgramManager to run when this model is initialized and setup
+        int initPluginLocation;         // If Python Plugins are used to init this, this defines the location in the plugin structure
+        int simPluginLocation;          // If Python Plugins are used to simulate this, this defines the location in the plugin structure
```
Myoldmopar, line 145: User defined plant op schemes did not know about Python plugins, now they do.

src/EnergyPlus/PluginManager.cc
```diff
@@ -45,6 +45,12 @@
 // OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 // POSSIBILITY OF SUCH DAMAGE.
 
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <dlfcn.h>
+#endif
```
Myoldmopar, line 52: This allows accessing an external shared library's exported symbols at runtime.  On Windows, `_WIN32` will be defined for both 32 or 64 bit builds, and results in importing symbols from `windows.h`.  Otherwise it gets `dlfcn.h`.

src/EnergyPlus/PluginManager.cc
```diff
@@ -55,41 +61,355 @@
...
-// so if we are building a Debug build of the C++ code, we need
-// to undefine _DEBUG during the #include command for Python.h.
-// Otherwise it will fail
-#undef _DEBUG
-#include <Python.h>
-#define _DEBUG
-#else
-#include <Python.h>
-#endif
-#endif
```
Myoldmopar, line 70: We are in PluginManager, which is part of the EnergyPlus shared library.  The primary design change in this PR results in an EnergyPlus shared library that does not have a build time dependency on the Python DLL.  So we no longer import the Python header or hard link to the Python shared library functions.

src/EnergyPlus/PluginManager.cc
```diff
@@ -55,41 +61,355 @@
...
     std::vector<std::string> globalVariableNames;
     std::vector<Real64> globalVariableValues;
 
     // some flags
     bool fullyReady = false;
-    bool shouldIssueFatalAfterPluginCompletes = false;
+    bool apiErrorFlag = false;
+
+    // function pointers into the PythonWrapper DLL
+    void *wrapperDLLHandle = nullptr;
```
Myoldmopar, line 79: These declarations store the handles to the new PythonWrapper DLL and the functions inside it.

src/EnergyPlus/PluginManager.cc
```diff
@@ -55,41 +61,355 @@
...
+#elif __linux__
+        std::string pythonLibPath = programDir + DataStringGlobals::pathChar + "libpythonwrapper.so";
+#else // Apple
+        std::string pythonLibPath = programDir + DataStringGlobals::pathChar + "libpythonwrapper.dylib";
+#endif
+        std::string pythonWrapper = PluginManager::sanitizedPath(pythonLibPath);
+        bool dll_errors = false;
+        char *err = nullptr;
+        // now actually open the library and assign the functions
+#ifdef _WIN32
```
Myoldmopar, line 145: On Windows we'll be using LoadLibrary and GetProcAddress, while on Unix we'll be using dlopen and dlsym.  But in both cases we get things from strings at runtime.

src/EnergyPlus/PluginManager.cc
```diff
@@ -110,17 +430,24 @@ namespace PluginManagement {
     }
 
 #if LINK_WITH_PYTHON == 1
-    std::string pythonStringForUsage() {
-        std::string sVersion = Py_GetVersion();
+    std::string pythonStringForUsage()
+    {
+        if (DataGlobals::eplusRunningViaAPI) {
+            return "Python Version not accessible during API calls";
```
Myoldmopar, line 436: Return early to avoid attempting to make a Python function call during API runs.

src/EnergyPlus/PluginManager.cc
```diff
@@ -411,23 +738,26 @@ namespace PluginManagement {
...
         // make this relative to the binary
-        std::string pathToPythonPackages = sanitizedProgramDir + "/python_standard_lib";
+        std::string pathToPythonPackages = sanitizedProgramDir + DataStringGlobals::pathChar + "python_standard_lib";
         FileSystem::makeNativePath(pathToPythonPackages);
-        auto a = Py_DecodeLocale(pathToPythonPackages.c_str(), nullptr);
-        Py_SetPath(a);
-        Py_SetPythonHome(a);
+        wchar_t *a = (*EP_Py_DecodeLocale)(pathToPythonPackages.c_str(), nullptr);
+        (*EP_Py_SetPath)(a);
+        (*EP_Py_SetPythonHome)(a);
```
Myoldmopar, line 750: These calls demonstrate the new way of making calls.  
- In the Python DLL itself, and the header, there are symbols and macros exposed like: `Py_SetPath`.
- In our PythonWrapper, we directly call those symbols and macros, but expose them as plain functions like: `EP_Py_SetPath`
- Then in this code, we have a reference to that `EP_Py_SetPath` function as a pointer
- To call it, we just deference the pointer and make the call, which is what the asterisk in the parentheses does.

src/EnergyPlus/PluginManager.cc
```diff
@@ -596,6 +926,18 @@ namespace PluginManagement {
 #endif
     }
 
+    PluginManager::~PluginManager()
+    {
+#if LINK_WITH_PYTHON
+        //(*EP_Py_FinalizeEx)();
```
Myoldmopar, line 932: Calling finalize was causing some big problems.  But after thinking this through, we shouldn't need to do that.  The only time Python is initialized is during PythonPlugin simulations.  Those simulations are only available when executing EnergyPlus as an executable.  When calling EnergyPlus this way, the entire thread and all contents are deleted when EnergyPlus finishes.  No need to finalize, because we won't be initializing again here.

src/EnergyPlus/PythonLibWrapper.cc
```diff
@@ -0,0 +1,97 @@
...
+  #define _DEBUG
+#else
+#include <Python.h>
+#endif
+
+#if _WIN32 || _MSC_VER
+#define PYTHONWRAP_API __declspec(dllexport)
+#else
+#define PYTHONWRAP_API
+#endif
```
Myoldmopar, line 64: This file is what is built to become the PythonWrapper.  The purpose of this file is to:
 - Include the Python header, link to the Python DLL, and make calls into the DLL as needed
 - Expose functions to be called by the EnergyPlus lib at runtime as needed

The main reason we need this wrapper is because a bunch of these Python calls made by this code are actually Macros that do a ton of logic.  We cannot access these macros dynamically at runtime, or EnergyPlus would be able to directly delay link to the Python DLL.  Instead, we need to create a file of pure function symbols that we can call from EnergyPlus.

src/EnergyPlus/api/datatransfer.h
```diff
@@ -451,6 +464,20 @@ ENERGYPLUSLIB_API int currentEnvironmentNum();
...
+/// \see apiDataFullyReady
+/// \see kindOfSim
+ENERGYPLUSLIB_API int actualTime();
+
+/// \brief Returns a simple sum of the date/time function
+/// \details Could be used in random seeding.
+/// \remark The behavior of this function is not well-defined until the `apiDataFullyReady` function returns true.
+/// \see apiDataFullyReady
+/// \see kindOfSim
+ENERGYPLUSLIB_API int actualDateTime();
```
Myoldmopar, line 479: @mitchute identified a couple EMS functions missing.  Thanks!

src/EnergyPlus/api/datatransfer.h
```diff
@@ -81,6 +81,15 @@ ENERGYPLUSLIB_API const char * listAllAPIDataCSV();
 ///          Do not call for variable, meter, actuator, or any other internal exchange data prior to this returning true.
 /// \return Returns 0 (success) once the data is ready, otherwise returns 1.
 ENERGYPLUSLIB_API int apiDataFullyReady();
+/// \brief Provides a user-facing check on the API error flag
+/// \details Some API functions return a value of 0, which could potentially indicate an error, or an actual 0 value.
+///          This function provides a way to disambiguate the response value.
+/// \return Returns 0 (success) if the error flag is not activated, or 1 if the error flag is true.
+ENERGYPLUSLIB_API int apiErrorFlag();
```
Myoldmopar, line 88: Functions do not attempt to throw anymore.  For the few that did attempt a throw, they return zero, and the client is then directed to check the error flag status to disambiguate between real zero outputs and error values.  This is a very common approach.

src/EnergyPlus/api/datatransfer.py
```diff
@@ -303,17 +339,39 @@ def get_internal_variable_handle(self, variable_type: Union[str, bytes], variabl
...
         function to then get the value of the variable.
 
         :param variable_handle: An integer returned from the `get_internal_variable_handle` function.
-        :return: Floating point representation of the internal variable value
+        :return: Floating point representation of the internal variable value.  Returns zero if the handle is invalid.
+                 Use the api_error_flag function to disambiguate between valid zero returns and error states.
         """
         return self.api.getInternalVariableValue(variable_handle)
 
+    def get_construction_handle(self, var_name: Union[str, bytes]) -> int:
```
Myoldmopar, line 354: @mitchute found this function was exposed over C, but not wrapped in Python.

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -1,11 +1,3 @@
-
-# this is repeated from the core E+ build, which is unfortunate
-# we need to either:
-# - figure out how to build the unit tests without the Python dependency, which would mean putting a conditional DEFINE in the PluginManager.hh code, or
-# - just get this once at highest level configure time and save it for everywhere
-find_package(PythonLibs 3 REQUIRED)
```
Myoldmopar, line 6: We won't be trying to link to Python on all cases here, so this is moved to a lower block.

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -212,18 +203,27 @@ set( test_src
...
+
+if(LINK_WITH_PYTHON)
+  if (CMAKE_HOST_UNIX)
+    find_package(PythonLibs 3 REQUIRED)
+    include_directories(${PYTHON_INCLUDE_DIRS})
+    list(APPEND test_src api/datatransfer.unit.cc)
+    list(APPEND test_src PluginManager.unit.cc)
+    list(APPEND test_dependencies ${PYTHON_LIBRARIES} )
+  endif()
+endif()
```
Myoldmopar, line 219: There is some voodoo going on, and I'm not investigating further for now.  On Linux/Mac, the API code is being unit tested thoroughly.  On Windows it doesn't work.  And being very frank, I'm not quite sure how it is actually working on Mac/Linux.  In any case, I'm limiting it to just Mac/Linux and we'll figure it out later.  The most likely scenario is that we don't try to test any API code from the unit test binary, and instead we do the API calls from the separate API test scripts/programs.

tst/EnergyPlus/unit/api/datatransfer.unit.cc
```diff
@@ -0,0 +1,640 @@
...
+#include <EnergyPlus/OutputFiles.hh>
+#include <EnergyPlus/OutputProcessor.hh>
+#include <EnergyPlus/PluginManager.hh>
+#include <EnergyPlus/api/datatransfer.h>
+
+#include "../Fixtures/EnergyPlusFixture.hh"
+
+using namespace EnergyPlus;
+
+class DataExchangeAPIUnitTestFixture : public EnergyPlusFixture
```
Myoldmopar, line 67: This file contains a very large set of unit tests for the data exchange API code.  I fixed many things in this branch while working on these unit tests and getting them running.  Weird how testing found problems.  Weird.

Anyway, this unit test fixture defines a pretty big chunk of stuff to help make testing easier, like utilities to setup variables, actuators, and do a bunch of stuff properly - and carefully.

Myoldmopar (commented):
Just did a big walkthrough of the code changes.  I am about to push a minor tweak to get things working fully on Mac.  I will tag a new test build so that I can run my installer sanity tests against the CI builds.  If everything is happy I'll report the results and I think this will be ready to go.

This is incredibly more robust than the original 9.3 code, and for those few people who may have already started playing with the plugins or API, this will provide a much better experience.  These changes won't affect the other users.

Myoldmopar:
OK, largeOff runs fine for me now locally with that initialization fixed (FYI @mitchute)

Myoldmopar:
OK, I think this is done.  Linking up to the API is now an immensely better experience for the users.  Using my sanity tester for the installers, on clean machines, I was able to diagnose a bunch of issues, on all three platforms, that were remedied by this branch.  The install packages which were built from this branch overnight last night now run perfectly, without any copying binaries, setting extra paths, or other nonsense.

Is there anything specific that anyone would like to see demonstrated/tested before this goes in?  There are only a few things left to drop in prior to a bug fix release, and this is a big one.  

Oh, and I've also updated the API docs: https://nrel.github.io/EnergyPlus/api/index.html

Myoldmopar:
OK, after some investigation of possible Python EMS diffs that were resolved without needing to make any E+ source changes, I feel even more confident this branch is ready.  I'm going to merge this in.  Whew.

PR #7940, opened 2020-04-21
Move debug file to {fmt}

lefticus:
Pull request overview
---------------------

Move dbg from GIO to {fmt}

Expected changes:

Some debug tables were previously hitting format size limits and outputting `**********`, with the new formatting routines, the number is output instead

Note: dbg is not tested at all with the current tests.

**However, I did run diffs for both the DebuggingOutput and Developer options for EnergyPlus locally**

### DIFF
--- a/src/EnergyPlus/DataGlobals.cc
+++ b/src/EnergyPlus/DataGlobals.cc
@@ -202,7 +202,6 @@ namespace DataGlobals {
     int OutputStandardError(0);                      // Unit number for the standard error output file
     std::ostream *err_stream(nullptr);               // Internal stream used for err output (used for performance)
     int StdOutputRecordCount(0);                     // Count of Standard output records
-    int OutputFileDebug(0);                          // Unit number for debug outputs
     int OutputFilePerfLog(0);                        // Unit number for performance log outputs
     int OutputFileShadingFrac(0);                    // Unit number for shading output
     int StdMeterRecordCount(0);                      // Count of Meter output records
@@ -307,7 +306,7 @@ namespace DataGlobals {
         OutputFiles::getSingleton().eso.close();
         OutputStandardError = 0;
         StdOutputRecordCount = 0;
-        OutputFileDebug = 0;
+        OutputFiles::getSingleton().debug.close();
         OutputFiles::getSingleton().zsz.close();
         OutputFiles::getSingleton().ssz.close();
         OutputFiles::getSingleton().mtr.close();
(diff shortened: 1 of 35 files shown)


lefticus:
@kbenne I cannot reproduce the Mac failure on Linux. Is this an expected/known/transient failure?

Myoldmopar:
The Mac does have a couple test cases that fail periodically.  Almost always guaranteed to be something not being reset in the clear_state calls.  Based on the CI results from prior to your last push and after your last push I would not worry about it.  I don't think it's anything you did.

src/EnergyPlus/DataGlobals.cc
```diff
@@ -307,7 +306,7 @@ namespace DataGlobals {
         OutputFiles::getSingleton().eso.close();
         OutputStandardError = 0;
         StdOutputRecordCount = 0;
-        OutputFileDebug = 0;
+        OutputFiles::getSingleton().debug.close();
```
Myoldmopar, line 309: :+1: :+1: :+1:

src/EnergyPlus/DataRuntimeLanguage.cc
```diff
@@ -108,7 +108,7 @@ namespace DataRuntimeLanguage {
     int const OperatorGreaterThan(12);    // >
     int const OperatorRaiseToPower(13);   // ^
     int const OperatorLogicalAND(14);     // &&
-    int const OperatiorLogicalOR(15);     // ||
+    int const OperatorLogicalOR(15);     // ||
```
Myoldmopar, line 111: I dunno.  I kinda liked Operatior.  And it's been in there for 11 years.  OK fine.

src/EnergyPlus/RuntimeLanguageProcessor.cc
```diff
@@ -1360,314 +1368,65 @@ namespace RuntimeLanguageProcessor {
                     Token(NumTokens).Type = TokenOperator;
                 }
 
-                // First check for two character operators:  == <> <= >=
+                // parse an operator if found,
```
Myoldmopar, line 1371: So now you're just doing some extra stuff :upside_down_face: :+1:

Myoldmopar (approved):
No problems here, I think this is good to go in.  Another file down!

Myoldmopar:
Pulled in latest develop locally just to make sure no other develop branch movement caused issues.  It built just fine and ran tests just fine.  This is good to go in.  Thanks @lefticus for taking down another file.

src/EnergyPlus/RuntimeLanguageProcessor.cc
lefticus: Yeah, sorry I couldn't bring myself to change the same code in 50 places when one simple function meant I only had to change it in one place.

Issue #7586, opened 2019-10-30
Any EMS in combination with an exterior window screen throws a severe error - should extend the EMS actuator to include this and other options


mjwitte:
Issue overview
--------------
When EMS is setting up actuators for Window Shading Control, if the file has any WindowShadingControls that apply shading type ExteriorScreen, these severe errors will be thrown:
```
   ** Severe  ** Missing shade or blind layer in window construction name = 'WINDOW_WITH_SCREEN', surface name = 'PERIMETER_TOP_ZN_3_WALL_NORTH_WINDOW'.
   **   ~~~   ** ...'Control Status' or 'Slat Angle' EMS Actuator cannot be set for a construction that does not have a shade or a blind layer.
   **   ~~~   ** ...Add shade or blind layer to this construction in order to be able to apply EMS Actuator.
```

1. This happens if the idf has *any* EMS, even if there are no actuators directed at a window with an exterior screen. The error is thrown in `EMSManager::SetupWindowShadingControlActuators`.

2a. One option is to simply skip over windows that have an exterior screen and leave things as-is.

2b. A better option would be to set up the actuator for windows with an exterior screen. Looking at `SolarShading::WindowShadingManager` it appears that an EMS actuator for exterior screen would work just fine. Just need to update the documentation to list the `SurfaceWindow(n).ShadingFlag` control value: "4: exterior screen is on".

3. There are additonal values for `SurfaceWindow(n).ShadingFlag` in the code that are not listed in the EMS documentation. It appears that these would work just fine with an EMS actuator, so add this list to the EMS documentation:
```
        //  10: window has an interior shade that is off but may be
        //       triggered on later to control daylight glare
        //  20: window has switchable glazing that is unswitched but may be switched later
        //       to control daylight glare or daylight illuminance
        //  30: window has exterior shade that is off but may be triggered on later
        //       to control daylight glare
        //  60: window has an interior blind that is off but may be
        //       triggered on later to control daylight glare
        //  70: window has an exterior blind that is off but may be
        //       triggered on later to control daylight glare
        //  80: window has a between-glass shade that is off but may be
        //       triggered on later to control daylight glare
        //  90: window has a between-glass blind that is off but may be
        //       triggered on later to control daylight glare
```

### Workaround
These errors can be ignored.

### Details
Some additional details for this issue (if relevant):
 - Version of EnergyPlus 9.2.0
 - PNNL e-mail, Oct 18, 2019

### Checklist
[...]

Issue #7809, opened 2020-02-25
Floating-point error when summing Daylighting Zone Fractions


zach-b:
We are running into floating-point rounding errors when setting the fraction of a zone controlled by the daylighting reference.

Although the decimal representation of these fractions given in the IDF sums to exactly 1, it seems that the floating point representation in EnergyPlus is > 1, which causes a Fatal Error.

Would it be possible to tolerate values slightly larger than 1 to accomodate for this case ? Because we are not sure how to proceed when filing the IDF otherwise.

### Details

```
   ** Severe  ** GetDaylightingControls: Fraction of Zone controlled by the Daylighting reference points is > 1.0.
   **   ~~~   ** ..discovered in "Daylighting:Controls" for Zone="Z-F0ON27MII", trying to control 1.00 of the zone.
``` 

Sample Daylighting:Controls that causes EnergyPlus to fail:
```
Daylighting:Controls,
    z-f0on27mii,                   ! Name
    z-f0on27mii,                   ! Zone Name
    splitflux,                     ! Daylighting Method
    on,                            ! Availability Schedule Name
    continuous,                    ! Lighting Control Type
    0.0,                           ! Minimum Input Power Fraction for Continuous or ContinuousOff Dimming Control
    0.0,                           ! Minimum Light Output Fraction for Continuous or ContinuousOff Dimming Control
    1,                             ! Number of Stepped Control Steps
    1.0,                           ! Probability Lighting will be Reset When Needed in Manual Stepped Control
    ,                              ! Glare Calculation Daylighting Reference Point Name
    0.0,                           ! Glare Calculation Azimuth Angle of View Direction Clockwise from Zone y-Axis
    22.0,                          ! Maximum Allowable Discomfort Glare Index
    ,                              ! DElight Gridding Resolution
    z-f0on27mii|jo_ecl_circu+0,    ! Daylighting Reference Point 0 Name
    0.1053,                        ! Fraction of Zone Controlled by Reference Point 0
    200.0,                         ! Illuminance Setpoint at Reference Point 0
    z-f0on27mii|jo_ecl_circu+1,    ! Daylighting Reference Point 1 Name
    0.0936,                        ! Fraction of Zone Controlled by Reference Point 1
    200.0,                         ! Illuminance Setpoint at Reference Point 1
    z-f0on27mii|jo_ecl_circu+2,    ! Daylighting Reference Point 2 Name
    0.1213,                        ! Fraction of Zone Controlled by Reference Point 2
[...]

PR #7941, opened 2020-04-21
Allow ExteriorScreen used in EMS Window Shading Control

lgu1234:
Pull request overview
---------------------
The fix uses Option 2b for #7586. ExteriorScreen in WindowShadingControl is allowed for EMS Window Shading Control.

An additional test file was uploaded.

### Pull Request Author

### Reviewer


tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -1558,5 +1571,39 @@ TEST_F(EnergyPlusFixture, EMSManager_TestOANodeAsActuators)
...
+
+    DataSurfaces::WindowShadingControl(1).ShadingType = 0;
+    DataSurfaces::WindowShadingControl(2).ShadingType = DataSurfaces::WSC_ST_ExteriorScreen;
+
+    DataSurfaces::TotSurfaces = 2;
+
+    SetupWindowShadingControlActuators();
+
+    EXPECT_FALSE(DataSurfaces::SurfaceWindow(2).ShadingFlagEMSOn);
+    EXPECT_EQ(DataSurfaces::SurfaceWindow(2).ShadingFlagEMSValue,0);
```
Myoldmopar: I'm not sure these lines actually reflect the test.  These values on the window data structure are default initialized to false and zero, regardless of whether an EMS actuator was actually set up.  I think what you'd want to do is make sure an EMS actuator was instantiated in the actuator list with these names and control types.  I'm not sure if there is an example of that in the unit tests already or if there is a helper function in the EMS code that would make it easy, but if you need assistance let me know.

tst/EnergyPlus/unit/EMSManager.unit.cc
```diff
@@ -1558,5 +1571,39 @@ TEST_F(EnergyPlusFixture, EMSManager_TestOANodeAsActuators)
     EXPECT_TRUE(Node(1).IsLocalNode);
     EXPECT_FALSE(Node(2).IsLocalNode);
     EXPECT_TRUE(Node(3).IsLocalNode);
-
+}
+TEST_F(EnergyPlusFixture, EMSManager_TestWindowShadingControlExteriorScreenOption2b)
```
Myoldmopar: This unit test is set up nicely.  I think it needs a change at the end, but it is succinct and clearly set up.  When you make the change to the test expectations at the bottom, please also rename the test (no need for the 2b anymore)

src/EnergyPlus/EMSManager.cc
```diff
@@ -1849,6 +1850,13 @@ namespace EMSManager {
                                      SurfaceWindow(loopSurfNum).SlatAngThisTSDegEMSon,
                                      SurfaceWindow(loopSurfNum).SlatAngThisTSDegEMSValue);
                 }
+            } else if (WindowShadingControl(Surface(loopSurfNum).WindowShadingControlPtr).ShadingType == WSC_ST_ExteriorScreen) {
```
Myoldmopar, line 1855: Works for me.

doc/ems-application-guide/src/ems-actuators/thermal-envelope.tex
```diff
@@ -19,6 +19,8 @@ \subsection{Window Shading Control}\label{window-shading-control}
   2.0:~~~~~~ Glazing is switched to a darker state (switchable glazings only).
 \item
   3.0: ~~~~~ Exterior shade is on.
+\item
+  4.0: ~~~~~ Exterior screen is on.
```
Myoldmopar, line 23: You had mentioned there were additional values that could be added to the documentation.  Did you intend on adding more here?  I'm not saying you necessarily should but just wrapping it all up.

Myoldmopar (changes requested):
Overall this is a great small change.  I am requesting a change to the unit test checks, but the code itself looks good.

Myoldmopar:
I pulled in develop locally and ran the new defect file.  With develop I get the EMS error, with this branch the file runs successfully.  

I also resolved the minor conflict.  I will push up the changes now.

Myoldmopar:
OK, @lgu1234 if you can fix up the changes I requested, this will be ready to go.

tst/EnergyPlus/unit/EMSManager.unit.cc
lgu1234: @Myoldmopar You are right. The unit test ensure that EMS actuator is instantiated. I have a test file available in DevelopmentSupport\EnergyPlusDevSupport\DefectFiles\7000s\7586 as 7586b.idf to check if the EMS actuator is working or not.

tst/EnergyPlus/unit/EMSManager.unit.cc
lgu1234: @Myoldmopar I am going to make the change by removing "2b".

doc/ems-application-guide/src/ems-actuators/thermal-envelope.tex
lgu1234: @Myoldmopar Although @mjwitte suggested to add more values, the key choices for this field are limited, and I have not found a good way to allow more choices. You may merge this one first and don't close the issue, so that I may add more choices later as a separate PR.

tst/EnergyPlus/unit/EMSManager.unit.cc
lgu1234: @Myoldmopar I added a section in the unit test to ensure that the EMS actuator overrides the ShadingFlag value.

lgu1234:
@Myoldmopar It is ready to review again.

Myoldmopar:
Yep, this looks good now, thanks @lgu1234.  I'll let CI wrap up a pass with your changes and this is likely ready to go in.

Myoldmopar (approved):
Thanks for making the fixes @lgu1234 .  I think this will go in once CI is done.

lgu1234:
@Myoldmopar I tried to remove an unused variable, then compiled unit test. Here is an error:

35>C:\Users\Gu\Documents\GitHub\Bugs\7586\EnergyPlus\tst\EnergyPlus\unit\DaylightingManager.unit.cc(1575): error C2100: illegal indirection

    std::array<Real64, 10> fractions({0.1053, 0.0936, 0.1213, 0.1018, 0.0893, 0.0842, 0.0882, 0.1026, 0.1134, 0.1003});

I also check the code in develop and no difference. I am not sure how to fix it.

Myoldmopar:
Illegal indirection is raised when you try to dereference a non-pointer.

```
int i;  // create a regular integer
int *j;  // create a pointer integer
*j = 1;  // the memory j points to now holds a value of 1
*i = 1;  // uh oh cannot dereference a non-pointer -- this will cause C2100
```

I don't see that anywhere in the line you showed.  Can you try a rebuild just to make sure the compiler didn't get confused?  Or even a computer reboot, right @rraustad? :upside_down_face:

lgu1234:
@Myoldmopar The unit test is a new one from 
#7809 as DaylightingManager_GetInputDaylightingControls_RoundingError. I upload my fix. Hope CI will not complain.

Myoldmopar:
@lgu1234 I have that same fix in a separate branch - https://github.com/NREL/EnergyPlus/pull/7991

It shouldn't have caused a C2100 though.  It makes no sense.  That's why I was saying it would be worth you rebuilding and/or restarting your machine.  @rraustad has occasionally run into strange situations where Visual Studio is throwing warnings or doing funky stuff and a fresh start helped.

Myoldmopar:
@jmarrec do you have any thoughts on why that line would be throwing a false C2100?  That's in the unit test you just added, but nothing looks wrong to me.  And with VS2017 it should be easily able to support the syntax there...right?

Myoldmopar:
@lgu1234 OK, so CI just failed to run the unit tests on a separate branch.  I guess something did break.  @mjwitte or @rraustad are you seeing a Windows compiler error in the unit tests in develop now?

rraustad:
Pulled latest develop and compiled with python including unit tests with no warnings.

Myoldmopar:
Thanks @rraustad.  It's working with Visual Studio 2019, but for some reason, Visual Studio 2017 isn't liking the `std::array`.  I was able to reproduce it in my VM and also confirm that changing to `std::vector` compiles.  I've pushed a branch and will get it merged in ASAP.  Windows results were just starting to look good, but now they'll be a bit ugly until this gets sorted over the weekend.

Myoldmopar:
@lgu1234 this got conflicted due to the fmtLD line change in both develop and your branch.  I'll pull the latest here and make sure my resolution worked OK, then get this in.

lgu1234:
@Myoldmopar Thanks.

Myoldmopar:
That is as nice and green a set of CI results as we've seen in some time.  And Windows build and test time less than an hour...

Merging.

PR #7912, opened 2020-04-07
Rezero sizing arrays to correct DOAS loads in Loads Component Summary reports

rraustad:
Pull request overview
---------------------
 - Fixes #7906
 - Some sizing arrays were not re-zero'd during sizing for load component reporting

**NOTE: ENHANCEMENTS MUST FOLLOW A SUBMISSION PROCESS INCLUDING A FEATURE PROPOSAL AND DESIGN DOCUMENT PRIOR TO SUBMITTING CODE**

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [x] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [x] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [x] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [x] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] ~If any defect files are updated to a more recent version, upload new versions here or on DevSupport~
 - [ ] ~If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label~
 - [ ] ~If structural output changes, add to output rules file and add OutputChange label~
 - [ ] ~If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies~

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #7947, opened 2020-04-24
Refactor RezeroZoneSizingArrays

rraustad:
Pull request overview
---------------------
 - Followup to #7912
 - Create method on struct ZoneSizingData to zero sizing arrays and condense ZoneEquipmentManager::RezeroZoneSizingArrays to a few lines.
 - Add method to allocate and initialize sizing arrays.

### Pull Request Author
 - [ ] ~If any diffs are expected, author must demonstrate they are justified using plots and descriptions~
 - [ ] ~If changes fix a defect, the fix should be demonstrated in plots and descriptions~
 - [ ] ~If any defect files are updated to a more recent version, upload new versions here or on DevSupport~

### Reviewer


rraustad:
@Myoldmopar I don't understand this compiler error.

``` 
2>C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus/SimAirServingZones.hh(193,26): error C2065: 'OutputFiles': undeclared identifier (compiling source file C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus\DataSizing.cc)
2>C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus/SimAirServingZones.hh(193,39): error C2065: 'outputFiles': undeclared identifier (compiling source file C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus\DataSizing.cc)
2>C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus/SimAirServingZones.hh(193,62): error C2062: type 'const int' unexpected (compiling source file C:\Users\RRaustad\Documents\MyProjects\EnergyPlus\src\EnergyPlus\DataSizing.cc)
```

rraustad:
Linux CI came back OK so maybe it's on my end?

rraustad:
OK now I see these errors on the CI but don't know how to fix.

src/EnergyPlus/DataSizing.cc
```diff
@@ -46,8 +46,10 @@
 // POSSIBILITY OF SUCH DAMAGE.
 
 // EnergyPlus Headers
+#include <EnergyPlus/DataEnvironment.hh>
 #include <EnergyPlus/DataPrecisionGlobals.hh>
 #include <EnergyPlus/DataSizing.hh>
+#include <EnergyPlus/SimAirServingZones.hh>
```
mjwitte: @rraustad I'm guessing this is the culprit, because SimAirServingZones.cc includes DataSizing.hh. You could pass in `NumOfTimeStepInDay` as an argument. That would be cleaner anyway. And I don't see why `DataEnvironment` was added here.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -1202,7 +1202,9 @@ namespace ZoneEquipmentManager {
                 CalcZoneSizing(DesDayNum, CtrlZoneNum).HeatLoadSeq.allocate(NumOfTimeStepInDay);
                 CalcZoneSizing(DesDayNum, CtrlZoneNum).CoolLoadSeq.allocate(NumOfTimeStepInDay);
                 CalcZoneSizing(DesDayNum, CtrlZoneNum).HeatZoneTempSeq.allocate(NumOfTimeStepInDay);
+                CalcZoneSizing(DesDayNum, CtrlZoneNum).DesHeatSetPtSeq.allocate(NumOfTimeStepInDay);
```
rraustad: New method required uniformity in allocation of arrays for ZoneSizing, CalcZoneSizing, FinalZoneSizing and CalcFinalZoneSizing.

rraustad:
@Myoldmopar @mjwitte this should address the clutter in RezeroZoneSizingArrays. CI looks clean.

src/EnergyPlus/ZoneEquipmentManager.cc
rraustad: Seems like another method could be used here to allocate all arrays.

rraustad:
Since I have a warning to correct I was looking for other things that could get an update. Here we loop through the number of time steps in a day. Why not just use this for example and get rid of the loop on NumOfTimeStepInDay:

    FinalZoneSizing(CtrlZoneNum).HeatFlowSeq = CalcFinalZoneSizing(CtrlZoneNum).HeatFlowSeq;

instead of this at line 2332 of ZoneEquipmentManger

                for (CtrlZoneNum = 1; CtrlZoneNum <= NumOfZones; ++CtrlZoneNum) {
                    if (!ZoneEquipConfig(CtrlZoneNum).IsControlled) continue;
                    for (TimeStepIndex = 1; TimeStepIndex <= NumOfTimeStepInDay; ++TimeStepIndex) {
                        FinalZoneSizing(CtrlZoneNum).HeatFlowSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).HeatFlowSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatLoadSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).HeatLoadSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolFlowSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).CoolFlowSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolLoadSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).CoolLoadSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatZoneTempSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).HeatZoneTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatOutTempSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).HeatOutTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatZoneRetTempSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).HeatZoneRetTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatZoneHumRatSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).HeatZoneHumRatSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).HeatOutHumRatSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).HeatOutHumRatSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolZoneTempSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).CoolZoneTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolOutTempSeq(TimeStepIndex) = CalcFinalZoneSizing(CtrlZoneNum).CoolOutTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolZoneRetTempSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).CoolZoneRetTempSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolZoneHumRatSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).CoolZoneHumRatSeq(TimeStepIndex);
                        FinalZoneSizing(CtrlZoneNum).CoolOutHumRatSeq(TimeStepIndex) =
                            CalcFinalZoneSizing(CtrlZoneNum).CoolOutHumRatSeq(TimeStepIndex);
                    }
                }

rraustad:
There is a little hit on performance with these changes. I guess from pulling these functions into RAM? Any way to avoid that performance hit?

    This commit      Baseline          % Improvement
    322011091591   321964085111       -0.0146 %

Myoldmopar:
You can try marking the functions as inline.  I don't think marking a function inline will explicitly always make a function inline, but I think it can be used to hint the compiler to try inlining on that function.  @mbadams5 @amirroth would have more insight there.

amirroth:
Mass inlining only makes sense if the body of the function takes **at most** a few instructions more than setting up the arguments, calling, and popping the results of the function. Otherwise, you get code bloat and inlining backfires.

rraustad:
What bothers me with this performance hit is that these new methods are only used a few times during sizing to initially set up the arrays and then to re-zero the arrays IF pulse heat loads are simulated (depends on table reports selected). That should be an insignificant impact on overall simulation speed. To see even a -0.01% hit on performance just bothers me given where in the simulation these changes were made (i.e., called a few times during sizing). So now what if similar changes as these were made to a high call count function used during the simulation? The performance hit might be significant. I'm just trying to think this through.

rraustad:
Does anyone see a trend here?

Previous commit:

    This commit      Baseline          % Improvement
    322011091591   321964085111       -0.0146 %

With the last change:

    322265244811   321964085111       -0.0935%

rraustad:
These results are really surprising. A factor of 10 (ish). I thought that this last commit would recover the performance loss from previous commits since the compiler would be optimizing the array copy instead of a using a loop. It did just the opposite. I'm now skeptical of the performance speed results.

rraustad:
I will suggest that the last commit be reverted to see if the performance impact reduces back to where it was.

rraustad:
The last change did have a big performance impact.

    Before array copy    Baseline          % Improvement
    322011091591   321964085111       -0.0146 %

After array copy:

    322265244811   321964085111       -0.0935 %

Reverting array copy

    322024794565   321964085111    -0.01886 %

image

rraustad:
This just doesn't make sense to me. The last change was an array copy at "EndZoneSizingCalc". So it's done once, or twice for a pulse sim. This is something to consider for the 10x team.

@Myoldmopar stopping here. This can be merged at will.

Myoldmopar:
@rraustad let me try one thing here just to test.  When we are accessing data on Array1D objects, the Objexx code has to calculate the offset to find the actual location in the underlying data storage.  We also have the ability to directly access the underlying data storage by using the square bracket (vector) form.  This should reduce the number of computations because it won't be constantly having to calculate the offset every time we call in.  I want to see how much of a gain we get from this, if any.

amirroth:
A performance of difference of 0.01% is noise and not worth your time. Honestly. Drop this and move on to other work. If this pops up as a huge problem in the future you can revisit.

rraustad:
There's a lot of eyes on this one ... and pointed to several interesting coding issues. @Myoldmopar that unit test runs fine here individually and as a whole.

image

Myoldmopar:
@rraustad reverted my commit and pulled in develop, assuming CI is still happy I'll merge this today.  Thanks!

amirroth:
Did you try std::fill for these and it was slower than hand-rolled loops?

Myoldmopar:
OK, I made a couple changes @rraustad and @amirroth.  In the reset to zero function I used std::fill.  In the allocate function I used an alternate function in Objexx that accepts an initializer value so it doesn't have to go out and loop over all the arrays again (at least in the E+ side of things).  I'm not sure it will make a difference performance wise, or if we will see that reflected in our performance "metric".  But the full unit test suite did pass, so things are likely at least still functioning properly.  I'll let CI have a full run over it, including the debug builds, before moving forward.

Myoldmopar:
Looking good so far.  Assuming the debug build works OK, I'll clean out the unused argument to get Linux all green.  Mac warning is being resolved in a separate PR.

rraustad:
That appears to be a better way to zero arrays.

rraustad:
Did I do something wrong? Looking for a change in clock-on-the-wall-time.

    TEST_F(EnergyPlusFixture, ZeroArraysUnitTest_stdFill) {
        Array1D<Real64> myArray;
        myArray.dimension(1000);
        for (int i = 1; i < 1000000; ++i) {
            std::fill(myArray.begin(), myArray.end(), 0.0);
        }
    }
    TEST_F(EnergyPlusFixture, ZeroArraysUnitTest_hardLoop) {
        Array1D<Real64> myArray;
        myArray.allocate(1000);
        for (int i = 1; i < 1000000; ++i) {
            for (int z = 1; z <= 1000; ++z) {
                myArray(z) = 0.0;
            }
        }
    }

image

rraustad:
I am just blown away that the brute force method is faster than the library.

rraustad:
Here's the timing if looping 50,000 times on the array reset. ??????

image

rraustad:
So I guess we need to be counting clock cycles ... not instructions executed and not time.

Myoldmopar:
As @amirroth was saying last time, valgrind does not necessarily represent runtime.  Once the code drops into kernel space, the instructions are no longer counted, so things like IO and heap allocation are not represented well there.  It's what we've got for now though.  We can certainly start discussions about a better suite that actually runs E+ in a release build and reports simulation time, but that computer has got to be exceedingly protected from any possible interruption like antivirus or backup or other traffic to ensure reliable results.

rraustad:
Good exercise though, even if it didn't prove execution speed

Myoldmopar:
Indeed, and I'll wait on Windows to run on it, just to make sure it works fine, then I'll clean up that unused argument and merge.  Thanks for this fun exercise.

mbadams5:
@rraustad Did you build in Release or Debug?
Also writing performance testing code can be tricky. I think it is much greater than 0% chance the compiler could optimize away the outer loop in your hardLoop test since it isn't doing any actual work.

rraustad:
It's release ... and thanks for asking, this is a head scratcher on all fronts. Which of these changes are faster? I know, it's hard to tell.

amirroth:
A couple of things on both this test fixture and the actual code:

- @mbadams5 is right, the compiler could be optimizing the hand-rolled test fixture "away" because nothing is reading the array afterwards. You could add a loop to both fixtures that sums the contents of the array and prints it out to make sure.

- Even if the compiler is not optimizing the hand-rolled code away entirely, it could be optimizing it in a way that is similar to the way std::fill optimizes (e.g., with vectors), except without the overhead of the function call (I guess it could also be inlining std::fill here, but I doubt it). The hand-rolled code could be legitimately faster in this case. And this could very well be happening in the actual code too. Loops that just zero a bunch of arrays are obviously easy to optimize. Maybe I just made a bad performance suggestion, not really looking at context. Sometimes the knee-jerk reaction to use a library is wrong, at least from a performance standpoint.

- The heap is in user space and and so are 99.999% of the code within malloc and 100% of the code within free so Valgrind can see them. [Ed: the 0.001% of malloc that isn't in user-space is the call to sbreak to map new virtual memory pages.] The original point of Valgrind (and still its main application I believe) is tracking heap usage to identify memory leaks, double frees, and overwrites. The problem with Valgrind in this case (and also in general) is that pure instruction count obviously matters for performance, but it is far from the only thing that matters. Not all instructions are created equal performance-wise. Cache behavior matters on both the code and data side. Wonky, data-dependent conditionals (I hate using this term because all conditionals are data dependent at some level) matter a lot because they lead to branch misprediction penalties. Time (user, kernel, wall-clock) is the only thing that will capture these effects and their ultimate impact on performance. [Ed: there are Valgrind features that will count cache hits and misses also using a shadow cache but those are really slow. Intel processors also have performance counters, but those are used to unpack and understand performance (I think VTune uses them for example), not to measure it.]

Myoldmopar:
I went ahead and pushed up the argument warning cleanup.  I won't wait on CI to run completely on that one.  I'll make sure Linux approves of it to verify the argument change was functional.  However, I will still wait until Windows reports on the above commit with my `fill` changes before merging.

Myoldmopar:
OK, with the last commit that Windows ran, the integration tests all passed and the regressions all had warnings that were just due to this branch being behind develop.  I'm calling it and merging this in.  Thanks all.

PR #7948, opened 2020-04-24
Addressed HeatExchanger:AirToAir:SensibleAndLatent negative effectiveness and sizing

### LINKED ISSUE #7901: HeatExchanger:AirToAir:SensibleAndLatent negative effectiveness and sizing
Issue overview
--------------
1. HeatExchanger:AirToAir:SensibleAndLatent is allowing a negative latent effectiveness to produce very cold (<-100C) outlet node temperatures. At a minimum, effectiveness should be clipped at zero, and maybe with a one-time warning that there's a problem.

2. HeatExchanger:AirToAir:SensibleAndLatent "Nominal Supply Air Flow Rate" (for "Air-to-Air HX 15") is autosizing to 0.48682 while the max OA flow rate is autosizing to 4.88937 m3/s which contributes to the problem. Seems the HX flow should match the max OA flow rate?
*Correction: The system is a makeup air system (DOAS) and the Sizing:System Type of Load to Size On=Sensible. It should be VentilationRequirement. Then everything would be sizing to the minOA flow as it should be.*

Background
-------------
User file is failing with
```
   ** Severe  ** PsyRhoAirFnPbTdbW: RhoAir (Density of Air) is calculated <= 0 [-7.76103].
   **   ~~~   ** pb =[98390.00], tdb=[-317.32], w=[1.0000000E-005].
   **   ~~~   **  Routine=Unknown, Environment=RUN PERIOD 1, at Simulation time=08/04 07:20 - 07:25
   **  Fatal  ** Program terminates due to preceding condition.
```

Trapping this in the debugger revealed the failure occurs in `CalcMoreNodeInfo` for `Node(670)` which is "Node 393, !- Exhaust Air Outlet Node Name" in HeatExchanger:AirToAir:SensibleAndLatent "Air-to-Air HX 15'.

Trapping for negative outlet temp in `MixedAir::CalcAirToAirGenericHeatExch` revealed that  the latent effectiveness is -22 (yes, negative) which results in a wildly negative exhaust air leaving enthalpy.

This is partly driven by the actual flow rate being 10x the nominal flow rate for the heat exchanger. For some reason, the HX flow rate is autosizing to 0.48682 while the max OA flow rate is autosizing to 4.88937 m3/s. 

Looking at the HX cooling effectiveness inputs explains why the latent effectiveness is going negative:

0.76, !- Sensible Effectiveness at 100% Cooling Air Flow {dimensionless}
0, !- Latent Effectiveness at 100% Cooling Air Flow {dimensionless}
0.76, !- Sensible Effectiveness at 75% Cooling Air Flow {dimensionless}
0.73, !- Latent Effectiveness at 75% Cooling Air Flow {dimensionless}

The last field should be zero (like all of the other HXs in this idf).

### Details
 - Platform Win64
 - Version of EnergyPlus v9.3.0
 - Helpdesk ticket number 15440

### Checklist
 - [x] Defect file added EnergyPlusDevSupport\DefectFiles, run with Toronto weather.

Nigusse:
Pull request overview
---------------------
 - Fixes #7901
 - Fixed Heat Exchanger nominal volume flow rate sizing issue. Now the nominal air flow rate is sized to the maximum or the minimum outdoor design air flow rate depending on the OA controller specification. Now the defect file runs successfully.
 - Provided warning when there is incorrect user-inputs for latent effectiveness at 100% and 75% rated flow rates. Warning message is thrown when only one of them is zero.  

### Pull Request Author

### Reviewer


Nigusse:
Sample warning error message:

```
   ** Warning ** HeatExchanger:AirToAir:SensibleAndLatent "AIR-TO-AIR HX 15"
   **   ~~~   ** ...Latent Effectiveness at 100% Cooling Air Flow = 0.00
   **   ~~~   ** ...Latent Effectiveness at 75% Cooling Air Flow = 0.73
   **   ~~~   ** ...Latent cooling effectiveness at 75% and 100% rated flow can be 0.0 but not one of them only.
```

Nigusse:
Sample results of HX nominal supply air flow rate sizing:

Before the fix:
```
 Component Sizing Information, Controller:OutdoorAir, CONTROLLER OUTDOOR AIR 15, Maximum Outdoor Air Flow Rate [m3/s], 4.88937
 Component Sizing Information, HeatExchanger:AirToAir:SensibleAndLatent, AIR-TO-AIR HX 15, Design Size Nominal Supply Air Flow Rate [m3/s], 0.48682
```

After the fix:

```
 Component Sizing Information, Controller:OutdoorAir, CONTROLLER OUTDOOR AIR 15, Maximum Outdoor Air Flow Rate [m3/s], 4.88937
 Component Sizing Information, HeatExchanger:AirToAir:SensibleAndLatent, AIR-TO-AIR HX 15, Design Size Nominal Supply Air Flow Rate [m3/s], 4.88937
```

Nigusse:
Three test suite files (ASHRAE9012016_HotelLarge_Denver.idf, DOAS_wNeutralSupplyAir_wFanCoilUnits.idf, and DOAToUnitVentilator.idf) have failed. This fix is expected to show diffs because of sizing results change. See below eio output for ASHRAE9012016_HotelLarge_Denver.idf test file:

```
  Component Sizing Information, Controller:OutdoorAir, VAV WITH REHEAT_OA_CONTROLLER, Maximum Outdoor Air Flow Rate [m3/s], 26.10562
  Component Sizing Information, Controller:OutdoorAir, VAV WITH REHEAT_OA_CONTROLLER, Minimum Outdoor Air Flow Rate [m3/s], 11.78914
- Component Sizing Information, HeatExchanger:AirToAir:SensibleAndLatent, VAV WITH REHEAT HEAT RECOVERY, Design Size Nominal Supply Air Flow Rate [m3/s], 11.78914
+ Component Sizing Information, HeatExchanger:AirToAir:SensibleAndLatent, VAV WITH REHEAT HEAT RECOVERY, Design Size Nominal Supply Air Flow Rate [m3/s], 26.10562
```

Nigusse:
Five test suite files are expected to show diffs because of changing the nominal supply air flow rate change from minimum to maximum design outdoor air flow rate for "HeatExchanger:AirToAir:SensibleAndLatent".

mjwitte:
Hmmm, so that does raise the question of which is the correct airflow rate to size the HX. In some applications, the min OA flow might be correct, because anything > min OA could mean economizer operation so the HX would be off/bypassed. But for a DOAS (which I think was the defect file application), with variable flow, Max OA flow makes more sense. Can we detect this and make the right choice, or does the user need an input to control?

Nigusse:
The test suite file DOAToPTAC.idf has HR HeatExchanger on the OA system. This is DOAS, so the maximum and minimum are the same. See sizing results below: 

```
Component Sizing Information, Controller:OutdoorAir, DOAS OA CONTROLLER, Maximum Outdoor Air Flow Rate [m3/s], 0.26908
 Component Sizing Information, Controller:OutdoorAir, DOAS OA CONTROLLER, Minimum Outdoor Air Flow Rate [m3/s], 0.26908
 Component Sizing Information, HeatExchanger:AirToAir:SensibleAndLatent, DOAS HEAT RECOVERY, Design Size Nominal Supply Air Flow Rate [m3/s], 0.26908

```

Nigusse:
I am trying to think what kind of design condition limit the HR-HX flow to the minimum other than DOAS, in which case the minimum and the maximum design outdoor flows are the same?

Nigusse:
One scenario I can think of is when the Economizer Control Type in controlled OA controller is set to *NoEconomizer*. If this is the case, then I will investigate to mine that information to set a flag that can be used during sizing such the heat exchanger is sized to minimum flow. Is there any other scenario where the HX needs to be sized to minimum flow?

mjwitte:
But a variable flow DOAS could have different min/max OA flows. I don't think we have any example files like that, but it can be done. It may work to check the economizer and HX lockout settings.

Nigusse:
OK. Below are the four cases the HX nominal flow is sized to Minimum OA Floe If there is outdoor air controller in the OA system:

- Economizer Type is *NoEconmizer*
- Economizer Control Action Type is set to *MinimumFlowWithBypass*
- When the Lockout Type is *LockoutWithHeating* or *LockoutWithCompressor* 
- When the Lockout Type is *NoLockout* and Heat Recovery Bypass Control Type is *BypassWhenOAFlowGreaterThanMinimum*

Nigusse:
@mjwitte I think, when the Economizer Type not NoEconomizer but the Heat Recovery Bypass Control Type is BypassWhenWithinEconomizerLimits, then the HX should be sized to Minimum Flow?

src/EnergyPlus/DataSizing.cc
```diff
@@ -313,6 +313,7 @@ namespace DataSizing {
     int DataDesicDehumNum(0);                       // index to desiccant dehumidifier
     bool DataDesicRegCoil(false);                   // TRUE if heating coil desiccant regeneration coil
     bool HRFlowSizingFlag(false);                   // True, if it is a heat recovery heat exchanger flow sizing
+    bool SizeHRHXtoMinFlow(false);                  // true, if the heat recovery heat exchanger nominal flow is sized to min OA flow
```
Nigusse: New flag variable is used to decide the of use minimum or maximum OA flow for sizing Heat Recovery HX.

src/EnergyPlus/HeatRecovery.cc
```diff
@@ -529,6 +532,12 @@ namespace HeatRecovery {
                                  "\" latent heating effectiveness at 75% rated flow is less than at 100% rated flow.");
                 ShowContinueError("Latent heating effectiveness at 75% rated flow is usually greater than at 100% rated flow.");
             }
+            if ((ExchCond(ExchNum).HeatEffectLatent100 == 0.0 && ExchCond(ExchNum).HeatEffectLatent75 > 0.0) || (ExchCond(ExchNum).HeatEffectLatent100 > 0.0 && ExchCond(ExchNum).HeatEffectLatent75 == 0.0)) {
+                ShowWarningError( cCurrentModuleObject + " \"" + ExchCond( ExchNum ).Name + "\"" );
+                ShowContinueError("..." + cNumericFieldNames(3)  + " = " + RoundSigDigits(rNumericArgs(3), 2));
+                ShowContinueError("..." + cNumericFieldNames(5)  + " = " + RoundSigDigits(rNumericArgs(5), 2));
+                ShowContinueError("...Latent heating effectiveness at 75% and 100% rated flow can be 0.0 but not one of them only.");
+            }
```
Nigusse: Throws warning when only one of the heating latent effectiveness (at 100% or 75% of rated flow) user-input value is zero. But allows both to be zero.

src/EnergyPlus/HeatRecovery.cc
```diff
@@ -543,6 +552,13 @@ namespace HeatRecovery {
                                  "\" latent cooling effectiveness at 75% rated flow is less than at 100% rated flow.");
                 ShowContinueError("Latent cooling effectiveness at 75% rated flow is usually greater than at 100% rated flow.");
             }
+            if ((ExchCond(ExchNum).CoolEffectLatent100 == 0.0 && ExchCond(ExchNum).CoolEffectLatent75 > 0.0) || (ExchCond(ExchNum).CoolEffectLatent100 > 0.0 && ExchCond(ExchNum).CoolEffectLatent75 == 0.0)) {
+                ShowWarningError( cCurrentModuleObject + " \"" + ExchCond( ExchNum ).Name + "\"" );
+                ShowContinueError("..." + cNumericFieldNames(7) + " = " + RoundSigDigits(rNumericArgs(7), 2));
+                ShowContinueError("..." + cNumericFieldNames(9) + " = " + RoundSigDigits(rNumericArgs(9), 2));
+                ShowContinueError("...Latent cooling effectiveness at 75% and 100% rated flow can be 0.0 but not one of them only.");
+            }
+
```
Nigusse: Same as the above but this is for cooling.

src/EnergyPlus/HeatRecovery.cc
```diff
@@ -5386,6 +5403,28 @@ namespace HeatRecovery {
...
+                    thisOAController.Lockout == MixedAir::LockoutWithCompressorPossible ||
+                    (thisOAController.Lockout == MixedAir::NoLockoutPossible &&
+                     thisOAController.HeatRecoveryBypassControlType == DataHVACGlobals::BypassWhenWithinEconomizerLimits)) {
+                    HXNominalFlowSizingFlag = true;
+                }
+            }
+        }
+        return HXNominalFlowSizingFlag;
+    }
 } // namespace HeatRecovery
```
Nigusse: Added new function that set the value of SizeHRHXtoMinFlow flag depending on the OA controller inputs. Various combination of economizer inputs such economizer control type, bypass control, and lockout status determine the flag value.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -2630,26 +2630,26 @@ namespace ReportSizingManager {
...
+                                    AutosizeDes = FinalSysSizing(CurSysNum).DesHeatVolFlow;
+                                } else if (SELECT_CASE_var == Other) {
+                                    AutosizeDes = FinalSysSizing(CurSysNum).DesMainVolFlow;
+                                } else {
+                                    AutosizeDes = FinalSysSizing(CurSysNum).DesMainVolFlow;
+                                }
+                            }
+                            if (SizeHRHXtoMinFlow) {
+                                if (FinalSysSizing(CurSysNum).DesOutAirVolFlow > 0.0) {
+                                    AutosizeDes = FinalSysSizing(CurSysNum).DesOutAirVolFlow;
```
Nigusse: The HR heat exchanger nominal flow rate is autosized to maximum OA flow if the SizeHRHXtoMinFlow is false and the heat exchanger is on OA system.

Nigusse: This change does not affect HR heat exchanger nominal flow sizing if the HR unit is on air loop main branch.

Nigusse:
Current fix shows when the economizer is NoLockout and the Heat Recovery Bypass Control Type is *BypassWhenOAFlowGreaterThanMinimum*, the HR is sized at max flow but when the Heat Recovery Bypass Control Type is *BypassWhenWithinEconomizerLimits*, the HR is sized to min flow. I think this logic should be reversed. @mjwitte what do you think?

Nigusse:
Below is the logic that sets a flag used for sizing heat recovery heat exchanger nominal flow to minimum OA value:

```
        if (thisOAController.Econo == MixedAir::NoEconomizer || thisOAController.EconBypass ||
            thisOAController.Lockout == MixedAir::LockoutWithHeatingPossible ||
            thisOAController.Lockout == MixedAir::LockoutWithCompressorPossible ||
            (thisOAController.Lockout == MixedAir::NoLockoutPossible &&
             thisOAController.HeatRecoveryBypassControlType == DataHVACGlobals::BypassWhenOAFlowGreaterThanMinimum)) {
             HXNominalFlowSizingFlag = true;}
```

Nigusse:
This pull request is ready for review.

src/EnergyPlus/HeatRecovery.cc
mjwitte: It's possible that a user might want one of these to be zero without the other, so I'm not sure this it the most general warning. The real problem is if the calculated effectiveness (either sensible or latent) goes negative. Would it be better to check for effectiveness <0 during the simulation, clip it to zero and throw a one-time warning there that lists the pertinent inputs (like you have here) and then says 
"... effectiveness is less than zero. Check the following inputs. Effectiveness reset to zero and the simulation continues." ?

Protecting against negative effectiveness is crucial, but I don't see a code change to add that.

src/EnergyPlus/DataSizing.hh
```diff
@@ -314,6 +314,7 @@ namespace DataSizing {
     extern int DataDesicDehumNum;                     // index to desiccant dehumidifier
     extern bool DataDesicRegCoil;                     // TRUE if heating coil desiccant regeneration coil
     extern bool HRFlowSizingFlag;                     // True, if it is a heat recovery heat exchanger flow sizing
+    extern bool SizeHRHXtoMinFlow;                    // true, if the heat recovery heat exchanger nominal flow is sized to min OA flow
```
mjwitte: I'm not a fan of adding yet another DataSizing flag, but I'll hold that thought for now.

src/EnergyPlus/ReportSizingManager.cc
```diff
@@ -2630,26 +2630,26 @@ namespace ReportSizingManager {
                 } else if (SizingType == SystemAirflowSizing) {
                     if (HRFlowSizingFlag) { // HX sizing
                         if (CurOASysNum) {
-                            if (FinalSysSizing(CurSysNum).DesOutAirVolFlow > 0.0) {
```
mjwitte, line 2634: Nit-picking, but this could simply be a one-line change here.
`if(SizeHRHXtoMinFlow && FinalSysSizing(CurSysNum).DesOutAirVolFlow > 0.0) {`

src/EnergyPlus/ReportSizingManager.cc
mjwitte: This isn't really sizing to max OA flow when SizeHRHXtoMinFlow is false. It's assuming that the max OA flow rate will match the CurDuctType flow or more likely `DesMainFlow`. 
Is there a FinalSysSizing variable for max OA flow?
Is `CurDutcType` even applicable and current when `CurOASysNum` is >0? I know this if else block is present in the current code, but I'm suspicious that it ever lands anywhere but the final else. @rraustad ?

src/EnergyPlus/HeatRecovery.cc
```diff
@@ -5386,6 +5403,28 @@ namespace HeatRecovery {
         }
     }
 
+    bool GetHeatRecoveryHXNominalFlowSizingFlag(int const OASysNum) 
+    {
+        bool HXNominalFlowSizingFlag;
```
mjwitte: The name of this function and variable is somewhat ambiguous - does "nominal" mean MinOA? At a minimum this needs a comment to explain what this function is doing. I'd probably rename the function to "GetHeatRecoveryHXMinFlowSizingFlag"?

src/EnergyPlus/HeatRecovery.cc
mjwitte: If don't see a check on the HX field "Economizer Lockout". If this is set to No, then HXNominalFlowSizingFlag = false because the HX will always be operating and economizer settings don't matter.

mjwitte: If `thisOAController.Econo == MixedAir::NoEconomizer`, I don't think it will ever send a bypass signal, so I think this should be the first test and if it's true, set the function to false and skip the rest.
Why do the Lockout options apply here?

mjwitte (commented):
Some initial code review comments/questions.

src/EnergyPlus/ReportSizingManager.cc
Nigusse: OK. will do the change. Done.

src/EnergyPlus/HeatRecovery.cc
Nigusse: Agreed, will do the change.

Nigusse: done.

src/EnergyPlus/DataSizing.hh
Nigusse: I know it complicates but I do not if there is any better was to do when you have all kinds control flavors.

Nigusse:
Only OA controller object has max OA flow rate but it gets from the same set of assignments that we use in sizing. See below:

```
                    if (SELECT_CASE_var == ControllerOutsideAir) {

                        CheckSysSizing(CurrentModuleObject, this->Name);

                        {
                            auto const SELECT_CASE_var1(CurDuctType);
                            if (SELECT_CASE_var1 == Main) {
                                this->MaxOA = FinalSysSizing(CurSysNum).DesMainVolFlow;
                            } else if (SELECT_CASE_var1 == Cooling) {
                                this->MaxOA = FinalSysSizing(CurSysNum).DesCoolVolFlow;
                            } else if (SELECT_CASE_var1 == Heating) {
                                this->MaxOA = FinalSysSizing(CurSysNum).DesHeatVolFlow;
                            } else if (SELECT_CASE_var1 == Other) {
                                this->MaxOA = FinalSysSizing(CurSysNum).DesMainVolFlow;
                            } else {
                                this->MaxOA = FinalSysSizing(CurSysNum).DesMainVolFlow;
                            }
                        }

                    } else if (SELECT_CASE_var == ControllerStandAloneERV) {
```

src/EnergyPlus/HeatRecovery.cc
Nigusse: Renamed the local variable to HXMinFlowSizingFlag and re-arranged the logic.

Nigusse:
Below is code snippet that sets the flag used to size the HX to minimum flow. The first two and the fourth cases, I think, are OK. However, the third option is debatable but I made this logic based on the documentation. Especially the *LockoutWithCompressorPossible* option the economizer is locked out during heating and cooling operation, so I thought the heat exchange should be sized to the minimum but in the case of *LockoutWithHeatingPossible* the economizer is locked during heating only. I am OK if you want to remove the later. Any ways, I am happy to be corrected if any of the logic are not applicable.    
```

        if (thisOASys.OAControllerIndex > 0) {
            auto &thisOAController(MixedAir::OAController(thisOASys.OAControllerIndex));
            if (thisOAController.ControllerType_Num == MixedAir::ControllerOutsideAir) {
                if (thisOAController.Econo == MixedAir::NoEconomizer) {
                    HXMinFlowSizingFlag = true;
                } else if (thisOAController.EconBypass) {
                    HXMinFlowSizingFlag = true;
                } else if (thisOAController.Lockout == MixedAir::LockoutWithHeatingPossible ||
                           thisOAController.Lockout == MixedAir::LockoutWithCompressorPossible) {
                    HXMinFlowSizingFlag = true;
                } else if (thisOAController.Lockout == MixedAir::NoLockoutPossible &&
                           thisOAController.HeatRecoveryBypassControlType == DataHVACGlobals::BypassWhenOAFlowGreaterThanMinimum) {
                    HXMinFlowSizingFlag = true;
                }
            }
        }
```

mjwitte:
```
 if (thisOAController.Econo == MixedAir::NoEconomizer) {
                    HXMinFlowSizingFlag = false;
```
Because the HX will never be bypassed if there is no economizer. I guess for all of these we need to look at the OA controller logic and see under what conditions is can set the HX bypass flag to true.

Nigusse:
@mjwitte regarding providing warning when the effectiveness values goes negative, the code [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HeatRecovery.cc#L2195-L2196) never lets the effectiveness to drop below zero. It looks negative effectiveness value is not used in the calculation. Do you still want to see the warning message added?

Nigusse:
My understanding has always been when NoEconomizer is the Economizer Type, then the OA system provides just ventilation air requirement, i.e. the minimum OA amount. Note that this logic applies only when the HX is on Outdoor Air System. It is for this reason that I set the HXMinFlowSizingFlag  = true for this case. 

If the HX is on the main branch, then the HX is sized whatever flow that applicable to the main branch. All these new logic added apply only when the HX is on Outdoor Air System.

mjwitte:
RE: negative effectiveness, when I posted the issue, I saw a negative effectiveness in the debugger which was causing the bad results. Not sure how it got there. If you can't see where that was happening I'll try to reproduce it.

RE: NoEconomizer. That's true for a recirculating system with some fraction of outdoor air. But it's not true for a DOAS with variable flow. Maybe we need some other way to figure out the intent. OR maybe we just ditch all of this guessing the intent and add a new field somewhere in Sizing:System or in the HX object to declare the sizing intent? I'd rather not do that, but that may be the most straightforward approach.

Nigusse:
Re: Negative Effectiveness: I removed the previous warning message based on user input and replaced it with warning message provided only when the actual effectiveness is less than zero. Now a onetime warning will be thrown when the actual effectiveness is negative for sensible and latent calculations in heating and cooling mode. I agree it is worth to throw warning if it some how results in negative effectiveness.

Nigusse:
@mjwitte Addressed most of your review change request except some of the economizer setting impact on the logic of the HX sizing.

Nigusse:
@mjwitte Rolled back the heat recovery HX flow sizing changes and modified the unit tests accordingly.

src/EnergyPlus/HeatRecovery.cc
```diff
@@ -2190,11 +2192,52 @@ namespace HeatRecovery {
...
+                        ShowContinueError("...Sensible Effectiveness at 100% Cooling Air Flow = " +
+                                          RoundSigDigits(ExchCond(ExNum).CoolEffectSensible100, 2));
+                        ShowContinueError("...Sensible Effectiveness at 75% Cooling Air Flow = " +
+                                          RoundSigDigits(ExchCond(ExNum).CoolEffectSensible75, 2));
+                        ShowContinueError("...Sensible effectiveness reset to zero and the simulation continues.");
+                    }
+                    onetimewarningsensibleEff = true;
+                }
+            }
+            bool static onetimewarninglatentEff = false;
```
mjwitte: Rather than a static bool here, I'd add a couple of bools to the `ExchCond` struct. That way if more than one HX exhibits the problem, they can all be reported.

mjwitte (commented):
@Nigusse Tested this with the original defect file. The new error message shows up (good), but the file still fails with a fatal error. Running this in the debugger reveals that effectiveness is calculated again in lines 2308-2323 and it's coming up negative again here. So it needs to be trapped here as well.

src/EnergyPlus/HeatRecovery.cc
Nigusse: OK. I will address your change request. @mjwitte Thanks.

Nigusse: Also adding the following HX Air Volume Flow ratio as part of the error message. The flow ratio used to calculate the effectiveness.

```cpp 
ShowContinueError("...Heat Exchanger Air Volume Flow Ratio = " + RoundSigDigits(HXAirVolFlowRatio, 2));```

mjwitte (approved):
Original defect file now runs to completion, showing the new warning. Nice thorough unit tests. All good.

PR #7956, opened 2020-04-29
Remove FlowLock From Pond GHE

### LINKED ISSUE #7954: Pond Ground Heat Exchanger Updating Temperature History Based on Plant FlowLock Status
The pond ground heat exchanger updates its temperature history term based partly on the plant's FlowLock status. See here: https://github.com/NREL/EnergyPlus/blob/9e88f630aeacbd4d030305988ac2d6d071005d80/src/EnergyPlus/PondGroundHeatExchanger.cc#L447-L451

The current conditions could be met multiple times per timestep, regardless of whether or not time has advanced. This should be updated to check whether time has advanced, and then update the temperature.

### Details

### Checklist

mitchute:
This is a minor part of the refactoring effort to eliminate the usage of FlowLock checks and logic in individual plant components. This logic should be handled by the Plant Manager. PR also addresses #7954.

Diffs are expected.

### Pull Request Author

### Reviewer


mitchute:
@Myoldmopar can you take a look at this when you get a second? I'm inclined to say that this is ready for review even though I'm not totally satisfied with the results. There may be a larger underlying problem here that I'm not seeing.

They are using the system time step in the RK integration scheme, here: https://github.com/NREL/EnergyPlus/blob/af2cd0be0de754f8de91ae0c5f2f7af8e042e363/src/EnergyPlus/PondGroundHeatExchanger.cc#L509-L522

However, they were updating the temperature history term based on checking FlowLock status. So I eliminated the FlowLock check and am instead checking to see if time has advanced. If you look at the pond temperature results for the Design Day only simulations, things don't look so good. See the next plot.

DD_Pond_Temps

But if you look at it over the course of a year, I think it looks much better.

Annual_Pond_Temps

Before, we had daily temperature swings in the pond temperature of around 8-10 C which seems unrealistic. The updated annual results look closer to reality. However, if you zoom in on the first 2000 timesteps (at a 15 minute timestep), we can see that the bulk temperature doesn't really start doing anything until about 600 timesteps in, or about 6 days.

Var_0

Let me know if you see anything here.

mitchute:
This is (hopefully) ready for review now. I've updated the code so the temperature history term updates properly at the start of the timestep. The following plot is the winter and summer design day performance compared to develop.

DD_Pond_Temps

This is causing big diffs in the annual results, but that's expected due to the bug this fixes.

Annual_Pond_Temps

The next two figures are the first six days of the simulation and six days in July. The temperature oscillation amplitude is bigger than before which I'm not totally comfortable with, but at least the temperature is getting updated at the right time so I assume the rest of the model is functioning as designed.

6_day_Pond_Temps
July_Pond_Temps

src/EnergyPlus/PondGroundHeatExchanger.cc
```diff
@@ -443,16 +441,17 @@ namespace PondGroundHeatExchanger {
             this->MyFlag = false;
         }
 
-        // check if we are in very first call for this zone time step
-        if (DataGlobals::BeginTimeStepFlag && FirstHVACIteration &&
-            DataPlant::PlantLoop(this->LoopNum).LoopSide(this->LoopSideNum).FlowLock == 1) { // DSU
+        if (FirstHVACIteration && !DataHVACGlobals::ShortenTimeStepSys && this->firstTimeThrough) {
```
Myoldmopar, line 443: This makes good sense to me.  I believe first HVAC will be true on the first attempt by the HVAC system and also when the HVAC system attempts to downstep.  This catches the first HVAC iteration on only the initial case.  And it only catches it once by using the first time through flag.  This means this should catch only the first HVAC system time step, so it only advances the past temperatures once we have moved forward in time.

src/EnergyPlus/PondGroundHeatExchanger.cc
```diff
@@ -443,16 +441,17 @@ namespace PondGroundHeatExchanger {
...
             // update past temperature
             this->PastBulkTemperature = this->BulkTemperature;
+            this->firstTimeThrough = false;
+        } else {
+            this->firstTimeThrough = true;
```
Myoldmopar, line 448: I'm a little unsure about resetting this on a pure `else` here.  I'm afraid that fact that the plant iterates can cause this to be invalid.  The very first HVAC iteration on a new system time step will cause the IF side to be hit, which will advance time by moving temperatures, and it will set the firstTimeThrough flag to false.  Let's say plant iterates a second time (it will), the firstTimeThrough flag will be false, so the IF side won't be hit.  Instead the ELSE side will be hit and the firstTimeThrough flag will be set back to true.  Now, let's say plant iterates a third time, it will then pass the IF side again, and advance time, which isn't right.

I think instead of an ELSE block here, you might need an `ELSE IF NOT FIRSTHVAC`, that way it will only be reset once the HVAC side has advanced away from FirstHVAC.  Then the IF side won't get hit again until it starts the next FirstHVAC.

src/EnergyPlus/PondGroundHeatExchanger.cc
```diff
@@ -498,12 +497,12 @@ namespace PondGroundHeatExchanger {
 
         Real64 PondMass = this->Depth * this->Area *
                           FluidProperties::GetDensityGlycol(
-                              fluidNameWater, max(this->PondTemp, DataPrecisionGlobals::constant_zero), this->WaterIndex, RoutineName);
+                              "WATER", max(this->PondTemp, DataPrecisionGlobals::constant_zero), this->WaterIndex, RoutineName);
```
Myoldmopar: Hard-wiring the water string literal here seems like a step backwards.  I like getting rid of the static string, but maybe we can do something that achieves all the goals:
 - Avoid using the string literal everywhere
 - Avoid the small computational hit associated with the static const string
 - Avoid the IDE warning about the global-level static constructor
 - Avoid the use of `auto` if reasonable.

@lefticus @mbadams5 @amirroth what is the absolute best path forward for constant strings with function scope?  I'm sure there's a constexpr solution, but I'd just like to figure out the best option once and for all so we can apply it throughout the code.  While we do like `auto` for creating references, for non-reference declarations, we would like to use explicit types, if it is reasonable.

src/EnergyPlus/PondGroundHeatExchanger.cc
```diff
@@ -808,7 +807,7 @@ namespace PondGroundHeatExchanger {
         if (massFlowRate == 0.0) {
             CalcEffectiveness = 1.0;
         } else {
-            NTU = DataGlobals::Pi * TubeInDiameter * this->CircuitLength * NumCircuits / (TotalResistance * massFlowRate * SpecificHeat);
+            NTU = DataGlobals::Pi * TubeInDiameter * this->CircuitLength * this->NumCircuits / (TotalResistance * massFlowRate * SpecificHeat);
```
Myoldmopar, line 806: I appreciate you adding `this` to make member variables obvious.

Myoldmopar (commented):
Thanks for this @mitchute.  I think there may need to be a little tweak on the time advancing part, or we should at least have a discussion on it.  Also I want to take this moment to resolve the local scope string declaration best-practice.

src/EnergyPlus/PondGroundHeatExchanger.cc
rraustad: So this only saves the PastBulkTemperature at the zone time step yet the pond HX is simulated at the system time step. Makes me think of the zone heat balance history terms that are adjusted when the system time step is reduced.

Myoldmopar: @rraustad the idea is to advance time once per system time step.  That is confirmed with the time step used in the update equation [here](https://github.com/NREL/EnergyPlus/blob/c6c1fcb9418c83bd273d0adf567a1f524fee3db0/src/EnergyPlus/PondGroundHeatExchanger.cc#L520).  The IF side of this should only be hit one time per system time step, right at the beginning.  This will ensure that the temperatures are shifted from the last time step once, right then.  It was wrong before, and I think this new way is closer, but I'm having trouble keeping all the flags straight, so I may need to run an experiment to convince myself.  

@mitchute we may need to run an experiment where we add a report variable to the pond heat exchanger that reports how many times it thinks it updated the temperatures per system time step.  The value can be updated in the context of this logic and then if we report it out at the detailed interval, it *should* be exactly 1 every system time step.

mitchute: @myoldmopar I added your suggested logic on the ```ELSE``` statement and things went all sorts of haywire. I'll add this report variable with the current logic and report back.

rraustad: A perfect test to see if this is working correctly.

mitchute: @rraustad @Myoldmopar I added a report variable as you suggested which incremented by 1 upon entering the ```IF``` section noted here. As is, the ```IF``` block is being hit multiple times per iteration, as you expected. Changing ```ELSE``` to ``` ELSE IF NOT FIRSTHVAC``` causes the temperature to be updated exactly 1 time per system timestep. The results looked bad when compared to the develop, but I think this gives us some confidence that at least this piece is working as expected.

Var_42
Var_0
Var_0

Myoldmopar: Thanks for the update on the report variable, I was glad to see it going up by exactly 1 every single system timestep in detailed output.  That was very reassuring.  I think this is an exceedingly important change.  The pond was artificially advancing the temperatures through time way, way, more than it should've been.  This new pattern of checking for a new time step can be applied throughout the hvac and plant world as a reliable way to move through time.  I don't like these diffs, but as far as I can tell, this is much, much more correct than before.

Myoldmopar:
I pulled develop in locally and ran regressions on the big list of diffs from CI.
- HeatPumpWaterToAirWithRHControl: Very small warmup convergence EIO diffs
- PondGroundHeatExchanger: Very small warmup convergence EIO diffs
- VSHeatPumpWaterToAirWithRHControl: Very small warmup convergence and then also a warning in the ERR file about the plant storing heat
- Table diffs in these three files that show peak electricity usage changes on the order of 100W, which is about 1% on the relative change side

But the math diffs are kinda weird.

With Heat Pump Water to Air RH Control, the zone temperature just randomly decides to go up in the evening?
image

I haven't been able to readily see what is causing this spike.  I have verified the only diffs in the two branches are the flowlock changes.  I was expecting some big diffs, but this spike is pretty wild.  I don't know if I'm comfortable dropping this into the bug fix release without some more investigation.  @mitchute what are your thoughts?

mitchute:
Yeah, that's weird. I'm ok holding off until we figure it out.

rraustad:
17:00 is usually when HVAC turns off. Is the system time step reducing here?

mitchute:
@Myoldmopar I've made a few more cleanups here, but the strange behavior you noted is still there. If you roll back the three lines in question it all goes away, at least when compared to develop. I think we've gone through the exercise of determining that what we've fixed by removing FlowLock is actually the correct path. And this isn't a complicated model, so right now I'm inclined to say that the problem is outside of this model.

Myoldmopar:
I wasn't expecting this to have trouble on Windows.  I will try to pull develop back in one more time just for fun.

src/EnergyPlus/PondGroundHeatExchanger.cc
lefticus: @Myoldmopar I just saw this message.

The "correct" answer is `static constexpr std::string_view` for a compile-time known string that you want to give a name to. This is only available in C++17.

The next best option is `static constexpr fmt::string_view`. Simply because we have access to it. I think I'm going to move to this for my format strings soon.

`static const std::string` is the wrong answer for anything that's a "small" string (15 characters or less). This is going to add overhead in accessing the string object and entering the scope the variable is in (this overhead is checking to see if the object has been initialized or not).

For longer strings `static const std::string` is OK, as it would save a dynamic allocation on each function call / loop / whatever.

As an aside, in my personal code I've relying on using `auto` more and more for anything that's returned from a function. And the reasoning is simple: I want to avoid conversions. This is true regardless of if it is a `&` or a value type. Both can accidentally invoke implicit conversions, and even integral conversions (like `size_t` -> `int`) can add overhead.

If we had `-Wconversion` enabled then we would at least be aware of these conversions when they happen if we stick with explicit types.

(source: 3 years of teaching Object Lifetime classes and answering all of the questions students ask about all the things.)

Myoldmopar:
I figured out the Windows problem.  Two variables in the pond GHE file were converted to constexpr values, but they were left in the header as plain const floating points.  I'll push up a fix momentarily.

Myoldmopar:
However, I wasn't sure why CI was missing those error messages, causing me to have to build locally on my VM.  Once I built on my VM, I added the problematic output lines to the CI unit test and found they were indeed sneaking under the radar.  I fixed CI to capture these lines properly, and I'll have CI re-run the Windows build here and see if it gives the errors.  Then I'll push the fix and we'll get this merged.  (https://github.com/Myoldmopar/decent_ci/pull/39)

Myoldmopar:
OK, Decent CI fix was merged, and I deleted the Windows CI results here so Windows will rebuild.  I am hopeful it reports error messages then I'll push the fix for this branch, get clean results, and then merge this in.

Myoldmopar:
Yay!  It worked!  OK, I'll fix the Windows build error and CI should report "clean"

Myoldmopar:
Well, close.  Mac reported that SmallNum was actually not used at all, so I'm fixing that now.

Myoldmopar:
OK this is all good.  Merging.  Thanks @mitchute

PR #7962, opened 2020-05-01
Clean a few things in the plugin manager

Myoldmopar:
Pull request overview
---------------------
Unit tests were failing when run serially because a function pointer was not reset to null.  Cleaned up a few other things along the way.

### DIFF
--- a/src/EnergyPlus/PluginManager.cc
+++ b/src/EnergyPlus/PluginManager.cc
@@ -724,9 +724,41 @@ namespace PluginManagement {
         globalVariableNames.clear();
         globalVariableValues.clear();
         plugins.clear();
-        pluginManager.reset(); // delete the current plugin manager instance, which was created in simulation manager, this clean up Python
         PluginManagement::fullyReady = false;
         PluginManagement::apiErrorFlag = false;
+        PluginManager * p = PluginManagement::pluginManager.release();
+        delete p;
+        wrapperDLLHandle = nullptr;
+        EP_Py_SetPath = nullptr;
+        EP_Py_GetVersion = nullptr;
+        EP_Py_DecodeLocale = nullptr;
+        EP_Py_InitializeEx = nullptr;
+        EP_PyRun_SimpleString = nullptr;
+        EP_Py_FinalizeEx = nullptr;
+        EP_PyErr_Fetch = nullptr;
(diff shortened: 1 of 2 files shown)


src/EnergyPlus/PluginManager.cc
```diff
@@ -724,9 +724,41 @@ namespace PluginManagement {
         globalVariableNames.clear();
         globalVariableValues.clear();
         plugins.clear();
-        pluginManager.reset(); // delete the current plugin manager instance, which was created in simulation manager, this clean up Python
         PluginManagement::fullyReady = false;
         PluginManagement::apiErrorFlag = false;
+        PluginManager * p = PluginManagement::pluginManager.release();
```
Myoldmopar, line 729: Release the unique pointer ownership and then call delete on the pointer itself.

src/EnergyPlus/PluginManager.cc
```diff
@@ -724,9 +724,41 @@ namespace PluginManagement {
...
+        delete p;
+        wrapperDLLHandle = nullptr;
```
Myoldmopar, line 731: Reset all the DLL and function pointers to null.

src/EnergyPlus/PluginManager.cc
```diff
@@ -929,11 +961,11 @@ namespace PluginManagement {
     PluginManager::~PluginManager()
     {
 #if LINK_WITH_PYTHON
-        //(*EP_Py_FinalizeEx)();
+        if (EP_Py_FinalizeEx) (*EP_Py_FinalizeEx)();
```
Myoldmopar, line 964: Only call these functions if the DLL has been opened.

src/EnergyPlus/PluginManager.hh
```diff
@@ -196,38 +196,6 @@ namespace PluginManagement {
     extern bool fullyReady;
     extern bool apiErrorFlag;
 
-    // lazy pointers into the Python DLL
```
Myoldmopar, line 199: These never needed to be extern-d in the header.

mjwitte (approved):
I don't pretend to understand all of the changes here, but it fixes the problem - all unit tests run locally. Not waiting for anymore CI reports.

PR #7965, opened 2020-05-01
Global exterior energy

brianlball:
convert ExteriorEnergyUse over to /Data/EnergyPlusData


src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -109,16 +132,38 @@ struct ExteriorEnergyUseData : BaseGlobalStruct
         }
     };
 
+    struct ExteriorEquipmentUsage
```
Myoldmopar: So this is the big question, I think.  What do we do with structs?   Ahead of this refactor, it is pretty simple, structs and function declarations go in the matching header with the code.  But with this refactor, we have a choice.  We can either move the structs out to the data folder or we can leave it in the original header and forward declare the EnergyPlus data struct instance.  The main pro/con I see is that if we start pulling headers into this file, the file is going to quickly grow to a very large size.  Like very.  I don't really see any problems with forward declaring the data struct so that we can leave the headers in the original location.  Another perk about this approach is that the files changed on Github will go down too, so it will be easier to review.  I'm certainly open for discussion, but that's the way I'm leaning right now.  Let me know if you want me to take a quick pass here and apply that.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -85,6 +85,29 @@ struct DataGlobal : BaseGlobalStruct
 
 struct ExteriorEnergyUseData : BaseGlobalStruct
 {
+    // MODULE PARAMETER DEFINITIONS:
+    int const ElecUse = 1;         // Electricity
```
Myoldmopar: Whenever possible, when we have a group of `int const` in EnergyPlus code, these should be converted over to an `enum` or `enum class`.  They aren't actually 'data', they are basically just another 'type'.  As I mentioned in the previous comment, I am leaning toward leaving structs in their original header locations.  And these converted `int const` into `enums` should be left in there as well.  In fact, in a bunch of cases, you'll probably find they don't need to be in the header at all, and can just be put at the top of the cc file as a nice bonus.

(after further review below, I see the fuel type is used in the unit test, so the enum will need to stay in the header, no problem.)

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -109,16 +132,38 @@ struct ExteriorEnergyUseData : BaseGlobalStruct
...
+        {
+        }
+    };
+
     int NumExteriorLights = 0; // Number of Exterior Light Inputs
     int NumExteriorEqs = 0;    // Number of Exterior Equipment Inputs
 
     // Object Data
     Array1D<ExteriorLightUsage> ExteriorLights;        // Structure for Exterior Light reporting
+    Array1D<ExteriorEquipmentUsage> ExteriorEquipment; // Structure for Exterior Equipment Reporting
```
Myoldmopar: :+1:

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -109,16 +132,38 @@ struct ExteriorEnergyUseData : BaseGlobalStruct
...
 
-    void clear_state() override {
+    void clear_state() {
         NumExteriorLights = 0;
         NumExteriorEqs = 0;
         ExteriorLights.deallocate();
+        ExteriorEquipment.deallocate();
+        UniqueExteriorEquipNames.clear();
+        GetExteriorEnergyInputFlag = true;
```
Myoldmopar: :+1:

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -3294,14 +3294,14 @@ namespace InternalHeatGains {
                         ZoneOtherEq(Loop).OtherEquipFuelType = noOtherFuelTypeZero;
                         FuelTypeString = AlphaName(2);
                     } else {
-                        ExteriorEnergyUse::ValidateFuelType(ZoneOtherEq(Loop).OtherEquipFuelType,
+                        ExteriorEnergyUse::ValidateFuelType(state.exteriorEnergyUse, ZoneOtherEq(Loop).OtherEquipFuelType,
```
Myoldmopar: This is really nice.  I can see a not-so-distant future where this line changes from:

```
ExteriorEnergyUse::ValidateFuelType(state.exteriorEnergyUse, ...);
```
to
```
state.exteriorEnergyUse.validateFuelType(...);
```

src/EnergyPlus/StateManagement.cc
```diff
@@ -313,7 +313,7 @@ void EnergyPlus::clearAllStates()
     EMSManager::clear_state();
     EvaporativeCoolers::clear_state();
     EvaporativeFluidCoolers::clear_state();
-    ExteriorEnergyUse::clear_state();
+    //ExteriorEnergyUse::clear_state();
```
Myoldmopar, line 316: So over time, the StateManagement file will become moot as well.  All the clear_states will go through the data manager.  This is going to be great.

tst/EnergyPlus/unit/InternalHeatGains.unit.cc
```diff
@@ -130,7 +130,7 @@ TEST_F(EnergyPlusFixture, InternalHeatGains_OtherEquipment_CheckFuelType)
         if (equip.Name == "OTHEREQ1") {
             ASSERT_EQ(equip.OtherEquipFuelType, 0);
         } else if (equip.Name == "OTHEREQ2") {
-            ASSERT_EQ(equip.OtherEquipFuelType, ExteriorEnergyUse::PropaneUse);
+            ASSERT_EQ(equip.OtherEquipFuelType, state.exteriorEnergyUse.PropaneUse);
```
Myoldmopar: Ahh, and here is one reason why we won't be able to put that enum into the cc file, because the unit test needs to test against it.  That's fine, the enum can go in the header, but in general, it's nice to move things out of the header that don't need to be there.

Myoldmopar (commented):
This is a very exciting PR.  It's nice to see another chunk getting refactored and yet not having to make changes in hundreds of files.  My main review comment is that I think we want to leave the structs defined in their original location, not brought into the data file.  In that original header, you'll likely need to forward declare the EnergyPlusData struct or whatever struct you use in the function declarations.

src/EnergyPlus/ExteriorEnergyUse.hh
```diff
@@ -90,40 +72,21 @@ namespace ExteriorEnergyUse {
 
     // Clears the global data in ExteriorEnergyUse.
     // Needed for unit tests, should not be normally called.
-    void clear_state();
+    //void clear_state();
```
Myoldmopar: Oh, and go ahead and delete these lines, not just comment.

src/EnergyPlus/BaseboardElectric.hh
```diff
@@ -57,6 +57,8 @@
 
 namespace EnergyPlus {
 
+    struct EnergyPlusData;
```
Myoldmopar, line 60: So the main struct will live inside the EnergyPlus namespace.  Anywhere that we are using a forward declaration of this struct we need to put it inside the EnergyPlus namespace, not before it, and not inside another nested namespace.  This goes with any other lower level structs you may use as well.

src/EnergyPlus/BaseboardElectric.hh
```diff
@@ -125,7 +127,7 @@ namespace BaseboardElectric {
 
     void clear_state();
 
-    void SimElectricBaseboard(EnergyPlusData &state, std::string const &EquipName, int const ActualZoneNum, int const ControlledZoneNum, Real64 &PowerMet, int &CompIndex);
+    void SimElectricBaseboard(EnergyPlus::EnergyPlusData &state, std::string const &EquipName, int const ActualZoneNum, int const ControlledZoneNum, Real64 &PowerMet, int &CompIndex);
```
Myoldmopar, line 130: Accompanying that change, whenever we declare it as an argument type, it needs to be declared properly with `EnergyPlus::EnergyPlusData`.  This goes with the other lower-level types you may use: `EnergyPlus::ExteriorEnergyUseData`.

src/EnergyPlus/BaseboardRadiator.hh
```diff
@@ -185,12 +183,6 @@ namespace BaseboardRadiator {
 
 } // namespace BaseboardRadiator
 
-//+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-//******************************************************************************************************
-//+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-//******************************************************************************************************
-//+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```
Myoldmopar, line 192: Feel free to clean up stuff like this as you refactor.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -66,119 +68,95 @@ struct BaseGlobalStruct
...
+        // MODULE PARAMETER DEFINITIONS
+        static constexpr int EndZoneSizingCalc = 4;
+
+        void clear_state() override {
+            AnnualSimulation = false;
+            DayOfSimChr = "0";
+        }
+    };
+
+    struct ExteriorEnergyUseData : BaseGlobalStruct {
```
Myoldmopar, line 87: These structs should remain very lean, only variables.  No nested structures, no int const values, just the actual variables storing data.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -66,119 +68,95 @@ struct BaseGlobalStruct
...
+            ExteriorEquipment.deallocate();
+            UniqueExteriorEquipNames.clear();
+            GetExteriorEnergyInputFlag = true;
+        }
+    };
+
+    struct FansData : BaseGlobalStruct {
+        // constants
+        static constexpr int ExhaustFanCoupledToAvailManagers = 150;
+        static constexpr int ExhaustFanDecoupledFromAvailManagers = 151;
```
Myoldmopar, line 110: I didn't remove these, but I would suggest you doing it.  I made an enum class from the constant integers in the exterior energy use namespace.  The enum class should live back in the original header file.  You'll need to go in and clean up the usage of those variables but it is generally easy.  If you don't feel like tackling that during any of these refactors, it is OK, but don't bring them into this namespace.  They are really just constants, not variables.

src/EnergyPlus/DataGlobals.hh
```diff
@@ -54,7 +54,6 @@
 
 // EnergyPlus Headers
 #include <EnergyPlus/EnergyPlus.hh>
-#include <EnergyPlus/Data/EnergyPlusData.hh>
```
Myoldmopar, line 57: This was causing a circular inclusion.  And it's not needed.  If, for some reason, DataGlobals needs something from the Data.hh file, it can forward declare it.

src/EnergyPlus/ExteriorEnergyUse.hh
```diff
@@ -53,77 +53,85 @@
...
-
-    // DERIVED TYPE DEFINITIONS:
+struct EnergyPlusData;
+struct ExteriorEnergyUseData;
 
-    // MODULE VARIABLE DECLARATIONS:
+namespace ExteriorEnergyUse {
 
-    // SUBROUTINE SPECIFICATIONS FOR MODULE <module_name>
+    enum class ExteriorFuelUsage {
```
Myoldmopar, line 65: Here is where I converted all those `int const` into an enumeration.  As I said, this isn't part of the actual scope here, so don't sweat every one of these, but in either case, leave them declared in the original header.

src/EnergyPlus/GroundTemperatureModeling/GroundTemperatureModelManager.hh
```diff
@@ -61,6 +61,8 @@
 
 namespace EnergyPlus {
 
+    struct DataGlobal;
```
Myoldmopar, line 64: I had to go around a number of places and add this forward declaration.  It had previously worked because these files all included `DataGlobals.hh`, which then included the main data file.  When I took that out, these did not compile anymore.  This new pattern works fine though.

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -3291,7 +3290,7 @@ namespace InternalHeatGains {
 
                     std::string FuelTypeString("");
                     if (AlphaName(2) == "NONE") {
-                        ZoneOtherEq(Loop).OtherEquipFuelType = noOtherFuelTypeZero;
+                        ZoneOtherEq(Loop).OtherEquipFuelType = ExteriorEnergyUse::ExteriorFuelUsage::Unknown;
```
Myoldmopar, line 3293: Minor changes related to converted the ints to an enumeration.

src/EnergyPlus/api/runtime.cc
```diff
@@ -54,7 +54,7 @@
 #include <EnergyPlus/UtilityRoutines.hh>
 
 void cClearAllStates() {
-    EnergyPlusData state;   //THIS IS TEMPORARY
+    EnergyPlus::EnergyPlusData state;   //THIS IS TEMPORARY
```
Myoldmopar, line 57: This definitely needs to be temporary.  I will do some thinking about this.  I think the main point is that now when someone wants to run an EnergyPlus simulation as a library, they need to own the state.

Myoldmopar (commented):
Let me know if you want to chat about my changes or my comments here anytime.

mitchute:
CI is as good as it's going to get here. Let's call this one done. Merging.

PR #7959, opened 2020-04-30
Clean up compiler warnings from GCC 9.3

Myoldmopar:
Pull request overview
---------------------
I installed Ubuntu 20.04 today and GCC 9.3 came with it.  Building EnergyPlus revealed some warnings that need to be tidied.  The primary things are:

- All -ffor-scope flags need to be eliminated, it is deprecated and is now warning about the flag being there.
- There are some deprecated copy warnings coming out that are emitted if the user has a user defined copy constructor without a user defined assignment constructor, vice versa, or one of the user defined constructors have been made private so it appears to be missing.  For our case, I think it is fine to just mute the warning.
- There is a sign comparison warning from assertions inside ObjexxFCL Array1D.

PR #7967, opened 2020-05-05
Mute unrecognized warnings so we can still emit them on newer compilers

Myoldmopar:
Followup to #7959 where the result ended up causing lots of warnings on some machines.  With this change, new unknown warnings do not cause problems with older compilers, so that we can still emit them with newer compilers.  This flag has an interesting behavior however, in that if any other problem arises during building, the muted warning should be called out, just in case it was hiding a problem.


mitchute:
I'm still having issues over here on this.

```
warning: unknown warning option '-Wno-deprecated-copy'; did you mean '-Wno-deprecated'? [-Wunknown-warning-option]
warning: unknown warning option '-Wno-unknown-warning'; did you mean '-Wno-unknown-argument'? [-Wunknown-warning-option]
```

mitchute:
Works on my machine 😄 

Should we wait for CI to finish or just merge it?

Myoldmopar:
I want to just merge it, but let's let CI wrap up on it, then merge away!  You don't have to wait on Windows though, this is just in the GCC/Clang block.

mitchute:
Looks like something happened to the Mac CI. It is worth investigating before this goes in?

Myoldmopar:
No, it seems Mac went off the deep end, nothing to do with this branch.  Merge away.

PR #7968, opened 2020-05-05
Global Boilers and Chillers

mitchute:
### Pull Request Author

### Reviewer


Myoldmopar:
I didn't wait on the build to finish before pushing, but I think that takes care of it.  The includes went circular.  DataPlant.hh was including the plant topology, and deep down in the plant topology classes, it included the big Data.hh file.  I moved those to forward declarations and that shouldn't happen anymore for plant components now.

Myoldmopar:
There we go, that's nice and green

mitchute:
The windows unit test failure passes on my VM. I won't worry about it any further at this point.

mjwitte:
@mitchute Yeah, these have been showing up randomly on the Windows CI lately.
 `unknown file: error: C++ exception with description "InitializeMeters: Could not open file eplusout.mtd for output (write)." thrown in the test body.
`

Myoldmopar:
I think once we get the mtd file converted over to the new OutputFiles structure, these will go away.

src/EnergyPlus/BoilerSteam.hh
```diff
@@ -141,19 +142,14 @@ namespace BoilerSteam {
 
         void getDesignCapacities(const PlantLocation &EP_UNUSED(calledFromLocation), Real64 &MaxLoad, Real64 &MinLoad, Real64 &OptLoad) override;
 
-        void getSizingFactor(Real64 &SizFac) override;
+        void getSizingFactor(Real64 &sizFac) override;
 
         void onInitLoopEquip(EnergyPlusData &EP_UNUSED(state), const PlantLocation &EP_UNUSED(calledFromLocation)) override;
 
-        static PlantComponent *factory(std::string const &objectName);
+        static PlantComponent *factory(BoilerSteamData &boilers, std::string const &objectName);
```
mitchute, line 149: Boiler and chiller data classes now being passed into the factory and GetInput methods.

src/EnergyPlus/Boilers.cc
```diff
@@ -233,75 +201,76 @@ namespace Boilers {
             // ErrorsFound will be set to True if problem was found, left untouched otherwise
             GlobalNames::VerifyUniqueBoilerName(
                 DataIPShortCuts::cCurrentModuleObject, DataIPShortCuts::cAlphaArgs(1), ErrorsFound, DataIPShortCuts::cCurrentModuleObject + " Name");
-            Boiler(BoilerNum).Name = DataIPShortCuts::cAlphaArgs(1);
-            Boiler(BoilerNum).TypeNum = DataPlant::TypeOf_Boiler_Simple;
+            auto &thisBoiler = boilers.Boiler(BoilerNum);
```
mitchute, line 204: Adding the state class reference to the front of everything was starting to get cumbersome to look at. ```auto &``` added where possible.

src/EnergyPlus/Boilers.hh
```diff
@@ -52,38 +52,26 @@
...
-    // Boiler flow modes
-    extern int const FlowModeNotSet;
-    extern int const ConstantFlow;
-    extern int const NotModulated;
-    extern int const LeavingSetPointModulated;
-
-    // MODULE VARIABLE DECLARATIONS:
-    extern int NumBoilers;          // Number of boilers
-    extern bool GetBoilerInputFlag; // Boiler input flag, false if input is processed
+    enum class TempMode
```
mitchute, line 69: New ```enum``` to get rid of some of these ```extern```s. The other boilers don't appear to need anything like this, so I left it in the header file.

src/EnergyPlus/ChillerAbsorption.cc
```diff
@@ -105,40 +105,22 @@ namespace ChillerAbsorption {
...
-    int const FlowModeNotSet(200);
-    int const ConstantFlow(201);
-    int const NotModulated(202);
-    int const LeavingSetPointModulated(203);
+    int constexpr waterIndex(1);
+    constexpr auto calcChillerAbsorption("CALC Chiller:Absorption ");
+    constexpr auto moduleObjectType("Chiller:Absorption");
 
-    int const waterIndex(1);
+    const char * fluidNameWater = "WATER";
```
mitchute, line 112: Better than string literals all over the place? 🤷‍♂️ 

The following was suggested,

```constexpr char * fluidName = "Water";``` 

but it throws warnings on my machine.

```warning: ISO C++11 does not allow conversion from string literal to 'char *const' [-Wwritable-strings]```

File: src/EnergyPlus/ChillerAbsorption.cc [hunk 5]
```diff
@@ -255,20 +237,20 @@ namespace ChillerAbsorption {
...
             ShowSevereError("No " + DataIPShortCuts::cCurrentModuleObject + " equipment specified in input file");
             // See if load distribution manager has already gotten the input
             ErrorsFound = true;
         }
 
-        if (allocated(BLASTAbsorber)) return;
+        if (allocated(chillers.BLASTAbsorber)) return;
 
-        BLASTAbsorber.allocate(numBlastAbsorbers);
+        chillers.BLASTAbsorber.allocate(chillers.numBlastAbsorbers);
```
mitchute: Missed ```auto &``` here. Add it?

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -377,7 +352,7 @@ namespace ChillerReformulatedEIR {
                                                DataIPShortCuts::cAlphaArgs(7),
                                                "Chilled Water Nodes");
 
-            ElecReformEIRChiller(EIRChillerNum).CondenserType = WaterCooled;
+            thisChiller.CondenserType = DataPlant::CondenserType::WATERCOOLED;
```
mitchute, line 354: ```CondenserType``` and ```FlowMode``` ints were duplicated across multiple components (and there may still be others out there doing the same). These were moved to an ```enum``` in ```DataPlant```.

src/EnergyPlus/CostEstimateManager.cc
```diff
@@ -89,32 +89,12 @@ namespace CostEstimateManager {
...
-
-    // MODULE VARIABLE DECLARATIONS:
-    // na
-
-    // SUBROUTINE SPECIFICATIONS FOR MODULE
-
-    // Functions
-
-    void SimCostEstimate()
+    void SimCostEstimate(EnergyPlusData &state)
```
mitchute, line 97: A few new ```&state``` references in places.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -68,6 +79,107 @@ namespace EnergyPlus {
 //    }
 //};
 
+    struct BoilersData : BaseGlobalStruct {
```
mitchute, line 82: Most of these boilers and chiller data classes only contain three members: a ```getInput``` flag, an integer ```numObjects``` member, and a 1D array of objects. These should probably get consolidated later.

src/EnergyPlus/Plant/DataPlant.hh
```diff
@@ -261,6 +261,23 @@ namespace DataPlant {
     extern int PlantManageSubIterations; // tracks plant iterations to characterize solver
     extern int PlantManageHalfLoopCalls; // tracks number of half loop calls
 
+    // Enum classes
```
mitchute, line 264: New ```enum```s I mentioned earlier.

src/EnergyPlus/PlantChillers.cc
```diff
@@ -108,47 +108,7 @@ namespace PlantChillers {
...
-        GetEngineDrivenInput = true;
-        GetElectricInput = true;
-        GetGasTurbineInput = true;
-        GetConstCOPInput = true;
-        ElectricChiller.deallocate();
-        EngineDrivenChiller.deallocate();
-        GTChiller.deallocate();
-        ConstCOPChiller.deallocate();
-    }
+    Real64 constexpr KJtoJ(1000.0); // convert Kjoules to joules
```
mitchute, line 111: ```constexpr``` for numerical constants. There's probably something similar for the string constants.

src/EnergyPlus/StateManagement.cc
```diff
@@ -250,16 +247,8 @@ void EnergyPlus::clearAllStates()
     AirflowNetworkBalanceManager::clear_state();
     BaseboardElectric::clear_state();
     BaseboardRadiator::clear_state();
-    Boilers::clear_state();
```
mitchute, line 253: All of this now happens over in the state data class.

mitchute (commented):
Another step towards controlling the global state of E+. A nice pattern is beginning to emerge here.

src/EnergyPlus/ChillerAbsorption.cc
Myoldmopar: No, the allocate call just returns a reference to the original object, no need to capture it (in our use cases, anyway).

src/EnergyPlus/Data/EnergyPlusData.hh
Myoldmopar: This is going to prove to be exceedingly common as we move through plant components.

src/EnergyPlus/BoilerSteam.hh
Myoldmopar: That's interesting.  I didn't think of that side-effect of this.  We'll be passing a reference to the state around to all the factories.  It makes me wonder if the factories will eventually become part of the state data classes.  Pondering, pondering...

src/EnergyPlus/Boilers.hh
Myoldmopar: :+1: :+1:

src/EnergyPlus/ChillerAbsorption.cc
Myoldmopar: I think it's better than string literals.  Using a pointer to a constant character makes good sense until we come up with anything better.  The warning is because a string literal is by definition a constant character string.  It cannot be changed.  The way you've declared it here, it's all fine, you have a pointer to a const char (`const char *`).  The constexpr way was trying to make it a constant pointer to a non-constant char (`char * const`), which won't work because the string literal _is constant_.

src/EnergyPlus/ChillerAbsorption.cc
```diff
@@ -255,20 +237,20 @@ namespace ChillerAbsorption {
...
 
-        if (numBlastAbsorbers <= 0) {
+        if (chillers.numBlastAbsorbers <= 0) {
             ShowSevereError("No " + DataIPShortCuts::cCurrentModuleObject + " equipment specified in input file");
             // See if load distribution manager has already gotten the input
             ErrorsFound = true;
         }
 
-        if (allocated(BLASTAbsorber)) return;
+        if (allocated(chillers.BLASTAbsorber)) return;
```
Myoldmopar: You are welcome to change variable names like `BLASTAbsorber` if you want.  You can ping me or others if you want an idea, but in this case, I think it would be just more clear to just say `chillers.absorptionChillers` or something.  No big deal for now though, carry on!

tst/EnergyPlus/unit/ChillerAbsorption.unit.cc
```diff
@@ -1784,42 +1783,43 @@ TEST_F(EnergyPlusFixture, ChillerAbsorption_Calc)
     Real64 AbsChillEvapLoad;
     bool AbsChillRunFlag = true;
     // check chiller inputs
-    EXPECT_EQ(BLASTAbsorber(AbsChillNum).NomCap, 100000.0);
-    EXPECT_EQ(BLASTAbsorber(AbsChillNum).FlowMode, ChillerAbsorption::LeavingSetPointModulated);
+    auto thisChiller = &state.dataChillerAbsorbers.BLASTAbsorber(AbsChillNum);
```
Myoldmopar: This is creating a pointer.  While I'm not afraid of pointers, I think it would be good to avoid this pattern too broadly.  I would suggest making the subtle change:

```
auto thisChiller = &state.dataChillerAbsorbers.BLASTAbsorber(AbsChillNum);
          to
auto & thisChiller = state.dataChillerAbsorbers.BLASTAbsorber(AbsChillNum);
```

This way you are creating a reference instead of a pointer.

tst/EnergyPlus/unit/ChillerConstantCOP.unit.cc
```diff
@@ -108,27 +108,27 @@ TEST_F(EnergyPlusFixture, ChillerConstantCOP_WaterCooled_Autosize)
...
-    ConstCOPChillerSpecs::getInput();
+    ConstCOPChillerSpecs::getInput(state.dataPlantChillers);
 
     DataPlant::PlantLoop(1).Name = "ChilledWaterLoop";
     DataPlant::PlantLoop(1).FluidName = "ChilledWater";
     DataPlant::PlantLoop(1).FluidIndex = 1;
     DataPlant::PlantLoop(1).PlantSizNum = 1;
     DataPlant::PlantLoop(1).FluidName = "WATER";
-    DataPlant::PlantLoop(1).LoopSide(1).Branch(1).Comp(1).Name = ConstCOPChiller(1).Name;
+    DataPlant::PlantLoop(1).LoopSide(1).Branch(1).Comp(1).Name = state.dataPlantChillers.ConstCOPChiller(1).Name;
```
Myoldmopar: Creating another reference to this chiller is a great way to clean up this test like you did with the absorber.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -140,21 +252,73 @@ namespace EnergyPlus {
         }
     };
 
+    struct PlantChillersData : BaseGlobalStruct {
```
Myoldmopar, line 255: Yes.  This class should be split into four different ones.  Sorta.  Each one will again only have a num integer, getinput bool, and an array of objects.  I think I see where this is going for reals.

Myoldmopar (changes requested):
This is just a great effort.  The only thing that I really want to fix up before merging is the pointer usage in the unit test code.  It will be a very quick change to just make it a reference instead.  It could be nice to propagate that same change elsewhere in the unit tests, but that doesn't have to be here.  I may make those changes though.

src/EnergyPlus/ChillerAbsorption.cc
```diff
@@ -120,7 +120,7 @@ namespace ChillerAbsorption {
             chillers.getInput = false;
         }
         // Now look for this particular object
-        for (auto &thisAbs : chillers.BLASTAbsorber) {
+        for (auto &thisAbs : chillers.absorptionChillers) {
```
Myoldmopar, line 123: Yeah, that's nicer :smile:

tst/EnergyPlus/unit/ChillerAbsorption.unit.cc
```diff
@@ -1783,43 +1783,43 @@ TEST_F(EnergyPlusFixture, ChillerAbsorption_Calc)
     Real64 AbsChillEvapLoad;
     bool AbsChillRunFlag = true;
     // check chiller inputs
-    auto thisChiller = &state.dataChillerAbsorbers.BLASTAbsorber(AbsChillNum);
-    EXPECT_EQ(thisChiller->NomCap, 100000.0);
-    EXPECT_EQ(thisChiller->FlowMode, DataPlant::FlowMode::LEAVINGSETPOINTMODULATED);
+    auto &thisChiller = state.dataChillerAbsorbers.absorptionChillers(AbsChillNum);
```
Myoldmopar, line 1786: Yes.  Thanks!

Myoldmopar (approved):
All good here.  Thanks for the quick fixes.

Myoldmopar:
@mitchute I then #7972 will merge in shortly, then I'll pull develop into this branch locally and run regressions.  Assuming all is well I think this will go in directly.

Myoldmopar:
Pulled develop in and ran regressions overnight.  All good here.  Merging.

PR #7972, opened 2020-05-06
Convert outputreports from gio

lefticus:
Pull request overview
---------------------

Convert all output files generated from OutputReports.cc from gio to {fmt}.

This is a refactoring change. It contains a considerable reworking of OutputReports.cc to reduce duplicated code.

Expected change:
 
 * All WRL output files have changed due to a bug fix in the naming of surfaces because of a reused variable that wasn't properly updated.

See: https://github.com/NREL/EnergyPlus/pull/7972/files#r423837883 and https://github.com/NREL/EnergyPlus/pull/7972/files#r423837070

Notes:

When using the "ensure_open" function (which makes sure that a globally accessible file is open and returns a reference to it, otherwise throwing an error). The parameter is name of the calling function so that we can keep parity with the originally produced errors

https://github.com/NREL/EnergyPlus/pull/7972/files#diff-2c8fa705f6ae2ba41ebc0fba4edc3b14R58-R64


src/EnergyPlus/ChillerElectricEIR.cc
```diff
@@ -714,12 +710,8 @@ namespace ChillerElectricEIR {
                     ShowContinueError("Energy input ratio as a function of part-load ratio curve shows negative values.");
                     ShowContinueError("EIR as a function of PLR curve output at various part-load ratios shown below:");
                     ShowContinueError("PLR          =    0.00   0.10   0.20   0.30   0.40   0.50   0.60   0.70   0.80   0.90   1.00");
-                    ObjexxFCL::gio::write(StringVar, "'Curve Output = '");
-                    static ObjexxFCL::gio::Fmt fmtF72("((F7.2),$)");
-                    for (int CurveValPtr = 1; CurveValPtr <= 11; ++CurveValPtr) {
-                        ObjexxFCL::gio::write(StringVar, fmtF72) << CurveValArray(CurveValPtr);
-                    }
-                    ObjexxFCL::gio::write(StringVar);
```
Myoldmopar: Where is this writing?  Standard out?

src/EnergyPlus/BranchInputManager.cc
```diff
@@ -2968,8 +2959,7 @@ namespace BranchInputManager {
         print(outputFiles.bnd, "{}\n", "! ===============================================================");
         static constexpr auto Format_700("! <#Branch Lists>,<Number of Branch Lists>");
         print(outputFiles.bnd, "{}\n", Format_700);
-        ObjexxFCL::gio::write(ChrOut, fmtLD) << NumOfBranchLists;
-        print(outputFiles.bnd, "{}\n", " #Branch Lists," + stripped(ChrOut));
+        print(outputFiles.bnd, " #Branch Lists,{}\n", NumOfBranchLists);
```
Myoldmopar: These are just so much cleaner.

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -564,43 +557,28 @@ namespace CommandLineInterface {
...
-            if (!FileExists) {
+            if (!fileExists(inputWeatherFileName)) {
                 DisplayString("ERROR: Could not find weather file: " + getAbsolutePath(inputWeatherFileName) + ".");
                 DisplayString(errorFollowUp);
                 exit(EXIT_FAILURE);
             }
         }
 
-        OutputFiles::getSingleton().debug.ensure_open();
+        OutputFiles::getSingleton().debug.ensure_open("OpenOutputFiles");
```
Myoldmopar, line 574: Can you clarify this action?  It looks like it is making sure the "OpenOutputFiles" file is open.

src/EnergyPlus/OutputFiles.hh
```diff
@@ -117,6 +128,14 @@ public:
     OutputFile bnd{"eplusout.bnd"};
 
     OutputFile debug{"eplusout.dbg"};
+
+    OutputFile dfs{"eplusout.dfs"};
+
+    OutputFileName sln{"eplusout.sln"};
+    OutputFileName dxf{"eplusout.dxf"};
+    OutputFileName sci{"eplusout.sci"};
+    OutputFileName wrl{"eplusout.wrl"};
```
Myoldmopar, line 137: More files converted, w00t!

Myoldmopar (commented):
This is another really exciting set of IO changes, @lefticus.  I know it's still a draft PR, so I'm not doing a big review, just nosing around.

src/EnergyPlus/ChillerElectricEIR.cc
lefticus: Yes, that's apparently a way to write to stdout. I had to dig through the source to figure that out.

src/EnergyPlus/CommandLineInterface.cc
lefticus: It's opening the `debug` file and using the string `"OpenOutputFiles"` in the error report if something goes wrong - it's the name of the calling function.

lefticus:
@Myoldmopar this is going to have an expected ~35 WRL diffs because I fixed a bug in the VRML output.

It's also a considerable rewrite of the OutputReports.cc file, which I apologize for. But there was so much duplicated code I felt it necessary to clean it up as much as possible.

After #7971 is done I'll pull the remaining changes into here and we should be able to focus on just the WRL diffs.

After this PR is merged I think I'll have the output portion of GIO completed in 2 more PRs.

...

Then we have to address the *input* portions of GIO.

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -1457,19 +1456,8 @@ namespace ChillerReformulatedEIR {
...
-                    for (int CurveValPtr = 1; CurveValPtr <= 11; ++CurveValPtr) {
-                        ObjexxFCL::gio::write(StringVar, "(F7.2,$)") << CondTempArray(CurveValPtr);
-                    }
-                    ObjexxFCL::gio::write(StringVar);
-                    ShowContinueError(StringVar);
-                    ObjexxFCL::gio::write(StringVar, "'Curve Output = '");
-                    for (int CurveValPtr = 1; CurveValPtr <= 11; ++CurveValPtr) {
-                        ObjexxFCL::gio::write(StringVar, "(F7.2,$)") << CurveValArray(CurveValPtr);
-                    }
-                    ObjexxFCL::gio::write(StringVar);
```
lefticus: This is definitely a bug on my part. I missed this. It needs to be written to stdout. @Myoldmopar I'll address this before I take the PR out of draft mode.

src/EnergyPlus/ChillerElectricEIR.cc
lefticus: Sorry yes, that's a bug introduced. I'll fix it.

src/EnergyPlus/ChillerReformulatedEIR.cc
mjwitte: @lefticus @Myoldmopar No, this isn't writing to stdout. This is writing formatted values to `StringVar` and then `StringVar` is the error string in line 1472 `ShowContinueError(StringVar);`. So, I think you've covered this with the new lines 1459 and 1460.  I doubt this code ever gets exercised in the standard test suite, we need to create a file that causes this error to fire to see if the new code is working.

src/EnergyPlus/OutputReports.cc
```diff
@@ -2575,134 +1977,79 @@ void VRMLOut(std::string &PolygonAction, std::string &ColorScheme)
...
         // still have to do shading surfaces for zone
         colorindex = 4;
-        for (surf = 1; surf <= TotSurfaces; ++surf) {
+        for (int surf = 1; surf <= TotSurfaces; ++surf) {
             //      !if (surface(surf)%heattranssurf) CYCLE ! Shading with a construction is allowed to be HT surf for daylighting shelves
             if (Surface(surf).Class != SurfaceClass_Shading) continue;
             if (Surface(surf).ZoneName != Zone(zones).Name) continue;
             if (Surface(surf).Sides == 0) continue;
-            ObjexxFCL::gio::write(unit, Format_710) << "# " + Surface(surf).ZoneName + ':' + Surface(surf).Name;
-            ObjexxFCL::gio::write(unit, Format_801) << colorstring(colorindex) << "Surf" + csurfnumber;
```
lefticus, line 2645: Inside this loop in the previous code the value for `csurfnumber` was never being updated. Instead the last value from https://github.com/NREL/EnergyPlus/pull/7972/files#diff-2d563c83691f73f081f932ae94ff8e32L2578 was being used.

src/EnergyPlus/OutputReports.cc
```diff
@@ -2575,134 +1977,79 @@ void VRMLOut(std::string &PolygonAction, std::string &ColorScheme)
...
-            for (vert = 1; vert <= Surface(surf).Sides; ++vert) {
-                ObjexxFCL::gio::write(unit, Format_802) << Surface(surf).Vertex(vert).x << Surface(surf).Vertex(vert).y << Surface(surf).Vertex(vert).z;
+            print(wrlfile, "# {}:{}\n", Surface(surf).ZoneName, Surface(surf).Name);
+            print(wrlfile, Format_801, colorstring(colorindex), "Surf", surf);
```
lefticus, line 2020: Here the bug in https://github.com/NREL/EnergyPlus/pull/7972/files#r423837070 is fixed because the variable `csurfnumber` was removed from the top level of the function, so it was impossible to accidentally reuse the value here.

lefticus:
@mjwitte @Myoldmopar thank you for merging the last PR, and sorry that this one was based on the last one. My intention was to keep them completely separate so they could be evaluated in parallel.

This code should be much more straightforward review. Every WRL file has changed (see review comments above) but there should be no other changes.

Myoldmopar:
The changes here are large, but contained to only a small number of files.  I didn't reconcile every single changed line, but the fact that CI looks so good is very encouraging.  I now understand the argument to `ensure_open`, and I feel like I'm resolved on the stdout writing changes.  @mjwitte I recall you creating a file that demonstrated the issue, but I don't see that in this conversation.  Was I imagining that?  And either way, do you feel these changes match original behavior now?  If so this should be able to go in.

src/EnergyPlus/CommandLineInterface.cc
Myoldmopar: Thanks for the clarification

Myoldmopar (approved):

lefticus:
@Myoldmopar Any changes that were to STDIO (or not as the case may be) should be limited to the last PR, you just happened to notice them on this PR because of the shared history.

mjwitte (approved):
@Myoldmopar Correct, the test file was for #7971. As for WRL diffs, they look fine. The WRL renders the same before and after (my viewer doesn't show the surface names, anyway). We may want to remove `  Output:Surfaces:Drawing,VRML;` from most of these test files and keep it just in 5ZoneAirCooled. Just wasted cycles to produce and diff the same file over and over.

Myoldmopar:
Thanks for the review @mjwitte.  I agree we should clean those out, at a later time of course...I'm hearing a percussive sound like a can rolling down the road.

OK, all good here, merging.  Thanks @lefticus.

Myoldmopar:
@lefticus did you mean to restore this branch?

lefticus:
hmm, no.

PR #7983, opened 2020-05-11
Fix identity matrix for diffuse solar exchange factors between zones 

xuanluo113:
Pull request overview
---------------------
This is to fix the diagonal values of the matrix computing solar exchange factors between zones. The diagonal values of the matrix are set to ones at each time step and are used to calculate  zone short-wave flux density [QS, QSLights](https://github.com/NREL/EnergyPlus/blob/f33887fecc7983e8c120726bc6a27eaf0cee9792/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L2612), and [QSDifSol](https://github.com/NREL/EnergyPlus/blob/f33887fecc7983e8c120726bc6a27eaf0cee9792/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L3310). However, during kickoff simulation and sizing, the values are set to [zeros](https://github.com/NREL/EnergyPlus/blob/d7e981526ad2e221ede42812bb8e2d13da268273/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4009), which I suppose they should not, as the zone short-wave flux density values are set to zero during kicking off. And changing this line is causing major diffs in this branch.

### Pull Request Author

### Reviewer


Myoldmopar:
This is a pretty significant amount of diffs.  I understand the change, but I need some testing before we can just merge this.  First, you should come up with a unit test to ensure that the matrix is always reset to the identity whenever it needs to be.  Second, some plots showing the impact on simulation results.  It could be that this change simply triggers HVAC to be on and off at different times, but overall energy use isn't affected.  That should be demonstrated through some plots.  There are 7 files with big ESO diffs, and while you don't need to show plots for every file, you should show enough to confirm the change.

xuanluo113:
@Myoldmopar The unit test is added. 

Comparing all the regression diffs in ASHRAE_LargeOffice during the design day simulations:
1. The eio diff is due to the different warmup days required (as the load change during Kickoff periods)
2. The err diff and html diff are due to a change of HVAC warning condition in Cooling tower convergence calculation.

```    
**   ~~~   **  ...Valid range = 1.11 to 11.11. Occurrence info = DENVER-AURORA-BUCKLEY.AFB_CO_USA ANN CLG .4% CONDNS DB=>MWB, 07/21 02:30 - 02:38
-   **   ~~~   **  ...Tower range temperature passed to the model = 1.05610
+   **   ~~~   **  ...Tower range temperature passed to the model = 1.05611
```
Due to the 0.00001 diff in the invalid temp range, the plant loop errors occur more time than the dev branch during warmup and are causing different behavior during sizing as the figure below shows.
Screen Shot 2020-05-23 at 11 54 24 AM

3. This is causing the data center zone temperature difference at the same hours.
Screen Shot 2020-05-27 at 10 34 58 AM

4. This is not causing noticeable diffs in energy meters and zone temperatures other than the data center zone.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4002,7 +4002,7 @@ namespace HeatBalanceSurfaceManager {
         }
 
         RecDifShortFromZ = false;
-        FractDifShortZtoZ = 0.0;
+        FractDifShortZtoZ.to_identity();
```
Myoldmopar, line 4005: This change is very succinct, and makes good sense.

tst/EnergyPlus/unit/HeatBalanceSurfaceManager.unit.cc
```diff
@@ -2561,4 +2561,61 @@ TEST_F(EnergyPlusFixture, HeatBalanceSurfaceManager_OutsideSurfHeatBalanceWhenRa
...
+
+    DataGlobals::KickOffSimulation = true;
+    ComputeDifSolExcZonesWIZWindows(DataGlobals::NumOfZones);
+
+    EXPECT_EQ(1, DataHeatBalSurface::FractDifShortZtoZ(1, 1));
+    EXPECT_EQ(1, DataHeatBalSurface::FractDifShortZtoZ(2, 2));
+    EXPECT_FALSE(DataHeatBalSurface::RecDifShortFromZ(1));
+    EXPECT_FALSE(DataHeatBalSurface::RecDifShortFromZ(2));
+
+}
```
Myoldmopar, line 2619: This is a really great unit test.  Just enough setup to demonstrate the change.

Myoldmopar (approved):
The code change here is great, and the unit test is as well.  I am still nervous about how widespread the diffs are.  I will need to run a little bit locally to convince myself but then I bet this can merge in.

Myoldmopar:
I ran a sample of 50 files locally for design days and found the incident rate of big diffs was actually pretty low.  I then went back and ran some specific files that were showing big diffs on CI.  The diffs seemed small.  I re-ran those files with annual runs and found the MTR diffs were technically out of range, but actually small.  I think this is ready to go.  Anyone have any concerns before merging?  (Timer set...)

Myoldmopar:
*Timer goes off*

OK, merging.

PR #7985, opened 2020-05-11
Revert "Reduce build time for Fortran"

Myoldmopar:
Ugh, this is causing an issue on the Windows build where it's not packaging up the transition binaries.  I'm not taking the time to dig in and fix it, I'm just reverting this for now.


Myoldmopar:
Windows is happy, I'm calling this done.  Forget I ever tried the change...

PR #7991, opened 2020-05-14
Small fixes

Myoldmopar:
Pull request overview
---------------------
This makes two changes:
- Our Windows CI resource is struggling heavily to keep up with demand.  We will have that remedied some time soon, but until then, I propose we turn off regressions on Windows.  Windows CI will still do the full build, and run all unit and integration tests, but it won't have to build/run a baseline branch, cutting the CI time in half roughly.
- Turning on Mac CI revealed an unused variable that isn't showing up on other platforms.  This removes that unused variable so Mac will be all green again.


Myoldmopar:
Oh look a super all green Mac CI response, lovely!  The next step is to make sure Windows is OK with my change.

mitchute:
Are you expecting a Windows test here?

Myoldmopar:
Yeah, I am modifying the Windows CI config so I definitely want to wait for it.  And wow is it slow.  Assuming this goes smoothly, this will be a big step in making Windows more responsive.

Myoldmopar:
And since I had time to investigate while Windows was taking so long, I took a look at the CI test runtime on Windows.  Here is everything over 4 minutes:

| EVERYTHING OVER 4 MINUTES LONG                                  |         |
|-----------------------------------------------------------------|---------|
| integration.SolarShadingTest_ExternalFraction                   | 1593.46 |
| integration.LgOffVAVusingBasement                               | 492.444 |
| integration.ASHRAE9012016_ApartmentHighRise_Denver              | 346.574 |
| integration.ASHRAE9012016_Hospital_Denver                       | 345.373 |
| integration.ASHRAE9012016_OfficeLarge_Denver                    | 343.399 |
| integration.ASHRAE9012016_OutPatientHealthCare_Denver           | 342.225 |
| integration.ASHRAE9012016_SchoolPrimary_Denver                  | 330.139 |
| integration.ASHRAE9012016_SchoolSecondary_Denver                | 329.324 |
| integration.AbsorptionChiller_Macro                             | 305.435 |
| integration._ExternalInterface-functionalmockupunit-to-schedule | 303.424 |
| integration._ExternalInterface-functionalmockupunit-to-actuator | 303.111 |
| integration.Exercise1D-Solution                                 | 302.864 |
| integration.Exercise1C-Solution                                 | 302.841 |
| integration.Exercise2                                           | 302.222 |
| integration._ExternalInterface-functionalmockupunit-to-variable | 300.846 |
| btwxt:TwoDFixture.construct_from_gridded_data                   | 290.345 |
| regression.5ZoneWaterCooled_HighRHControl                       | 240.196 |

Most of these are expected, but there are a few outliers.  
- It's really amazing that solar shading external fraction is taking so long.  That's almost 30 minutes to run the external fraction solar shading test...in a release build.  Someone needs to investigate.  @xuanluo113 ?
- The `btwxt` unit test runs for almost 5 minutes.  This seems pretty high for just a unit test.  @nealkruis ?
- The high RH control regression test runs for 4 minutes.  That's 4 minutes of just comparing results.  Is that file reporting at a detailed level and with a zillion output variables?  @Myoldmopar ?

Myoldmopar:
> * The high RH control regression test runs for 4 minutes.  That's 4 minutes of just comparing results.  Is that file reporting at a detailed level and with a zillion output variables?  @Myoldmopar ?

That file has 28 output variables listed, all of which are hourly.  Almost all of them have unique keys given, and the ones that don't should not expand into a great number of instances (zones, coils, chillers).  It doesn't really make sense that this regression would take so long unless something else is going on with the machine.

nealkruis:
Yeah, something is wrong there. Maybe an optimization flag in MSVC? On my Mac all Btwxt unit tests run in 0.19 seconds. I'll look into this on my Windows VM.

Myoldmopar:
Thanks @nealkruis.  I'm going to take the WIndows machine down and restart it, and make sure nothing is running on the machine (backups, etc.)

mjwitte:
@Myoldmopar SolarShadingTest_ExternalFraction has 102 Schedule:File schedules which is known to be slow (re-reading and re-parsing the same file multiple times). Does the schedule manager refactor touch the file reading for that? If not, that's a worthwhile parallel task. That said, this file just ran on my Windows machine in just under 6 minutes.

Myoldmopar:
Yes, the schedule manager branch has a rewrite of that code that drastically improves the runtime of that file.  I could potentially wrap that into a standalone piece as a step toward getting schedule manager work going again.  

I also just rebooted the Windows box, hopefully that alone will help clean up some of this slowness.

xuanluo113:
@Myoldmopar Yes, as Mike said, this test is to import the external schedule column by column, instead of all at once. This should be improved if we don't close and reopen the file every iteration of the loop.

Myoldmopar:
So I just realized I needed to make another tweak to get Windows CI to work.  I had disabled `skip_regressions` in the CI yaml file, which is all good, in that it won't waste time prebuilding the baseline branch.  However, I had left the `ENABLE_REGRESSION_TESTS` and related flags which Decent CI would be passing into CMake.  This would mean a bunch of `regression.*` tests would be included in the CTest listing, however they would totally fail because the baseline wouldn't have been built.  I just cleaned out all the regression variables as well here.  I am really considering just dropping this in right away.  The Windows CI bot is completely overloaded and we need to make this change to temporarily alleviate the situation.  Mac and Linux will still be running regressions.  Unless anyone objects, I'm going to merge this shortly and if I missed something and Windows CI starts doing anything funny I'll be happy to fix it after the fact.

nealkruis:
@Myoldmopar there's no issue on my Windows VM build of Btwxt. I wonder if it has something to do with the fact that it's the first test in the list? Maybe something is holding up ctest or google test (like the gtest_discover_tests)?

PR #7992, opened 2020-05-15
Use std::vector instead of std::array

Myoldmopar:
Recent unit test change introduced a std::array usage that Visual Studio 2017 for some reason struggles with.  In this case it was easy enough to convert it to a vector.  I'll get it merged in and we should be all clean again sometime over the weekend.

### DIFF
--- a/tst/EnergyPlus/unit/DaylightingManager.unit.cc
+++ b/tst/EnergyPlus/unit/DaylightingManager.unit.cc
@@ -1572,7 +1572,7 @@ TEST_F(EnergyPlusFixture, DaylightingManager_GetInputDaylightingControls_Roundin
 
     EXPECT_EQ(10, DataDaylighting::ZoneDaylight(1).TotalDaylRefPoints);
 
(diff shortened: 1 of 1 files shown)


Myoldmopar:
@lgu1234 can you pull this branch and make sure it builds and runs for you?  If so I'll get this dropped in right away and start cleaning up other branches.

Myoldmopar:
@Nigusse if you are also seeing the C2100 error and can verify this branch fixes it, I'll use that as confirmation as well.

lgu1234:
@Myoldmopar I tested and it worked. Thanks for your quick response. I am using VS2017.

Myoldmopar:
@lgu1234 thanks!  I'll merge this in so other branches pick this back up.

jmarrec:
@Myoldmopar I didn't expect MSVC to have a problem with this and throw a C2100, sorry.
It is confused by the fact that I use `frac({...})` and thinks it needs to call a constructor which doesn't exist. 

(std::array would be better here (ideally would be `constexpr` in the future) as it'll be stack-allocated, but given that it's only 10 doubles, vector is plenty fine and won't make a difference)

Thanks for fixing it.

Issue #7622, opened 2019-11-16
Standardize Output Variable and Meter Fuel Types


mjwitte:
**This is part 4 of 4: Part 1 = #7532, Part 2 = #7523 , and Part 3 is #7621.**

Issue overview
--------------
This is the final piece of standardizing fuel type on both input and output. Issues #5941 and #7621 dealt with the input side, with the goal of no diffs. This issue takes the plunge and modifies output variable names and meter names to the final standard list:

**Input choice, Meter Resource and Output Variable, Table Headings** (*updated 03 Aug 2020 MJW*)
Electricity, Electricity, Electricity
NaturalGas, NaturalGas, Natural Gas
Propane, Propane, Propane
Diesel, Diesel, Diesel
Gasoline, Gasoline, Gasoline
FuelOilNo1, FuelOilNo1, Fuel Oil No 1
FuelOilNo2, FuelOilNo2, Fuel Oil No 2
Coal, Coal, Coal
OtherFuel1, OtherFuel1, Other Fuel 1
OtherFuel2, OtherFuel2, Other Fuel 2
Water, Water, Water

Note that Meter Resource = Input Choice when this is all done. That may allow some simplification in how the objects use fuel types.

This will impact:

1. Meter resource types for Gas will become NaturalGas.

2. Output variable names will change for Electricity, NaturalGas, FuelOilNo1, FuelOilNo2, OtherFuel1, and OtherFuel2 ( more less?). This will required potentially many lines in the transition Report Variables 9-2-0 to 9-3-0.csv - script this? Or see modify transition to replace on substrings - maybe that's better.

3. Electricity output variables are currently different from other fuels. "Electric Power" and "Electric Energy" vs others that are "Gas Rate" and "Gas Energy". So, if we go with always matching the long version of the input fuel type, then these would become "Electricity Rate" and "Electricity Energy" - a little less natural, but makes things consistent across all fuel types. ~Yes? No? Some other approach?~ *updated 03 Aug 2020 MJW - Yes, change to Electricity Rate and Electricity Energy.*

4. For the most part table output column headings are using the desired strings, but some column headings may be affected and some table row headings may change.

5. There will be diffs everywhere:  mtd, eso, mtr, table - even though it's more work, it may be best to stage this in 2 parts: change the meters first (there should be no eso diffs) then change the output variables (there should be no mtr diffs). Or make a special version of mathdiff that can map old to new variable names?

6. More???

### Details
Some additional details for this issue (if relevant):
[...]

PR #8002, opened 2020-05-19
Added new functions to validate fuel types and set output strings

### LINKED ISSUE #7621: Add new function to validate fuel types and set output strings
**This is part 3 of 4: Part 1 = #7532, Part 2 = #7523 , and Part 4 is #7622.**

Issue overview
--------------
The goal of this issue to to remove all if blocks for fuel type in the input processing functions and replace them with a standard function call.

This issue should only standardize the fuel type choice input processing and introduce *no output changes* - that can wait for #?????.

### Background ###
#7523 standardized all fuel type inputs to a single list with no synonyms. Synonyms were removed from all (most?) input processing functions. #7523 introduces no output changes.

Some input processing using the functions `DataGlobalConstants::AssignResourceTypeNum` and `DataGlobalConstants::GetResourceTypeChar` to validate fuel types and set the standard strings for output variables and meter resource types. But many input processing functions still have their own if blocks to test for "NaturalGas" "Electricity" etc. This issue is to use a standard fuel type function everywhere as appropriate. 

### More Details ###
1. Given that fuel types are a subset of resource types, `AssignResourceTypeNum` could be modified to take an argument to only allow fuels (no other resource types) or it may be best to have a new function(s) just for fuel type inputs.

2. In general, objects with a fuel type choice need to validate the input choice and assign strings to use for output variables and the correct resource type to use for meters.

3. `AssignResourceTypeNum` and `GetResourceTypeChar` may be ripe for some refactoring.

4. For this first pass with no output diffs, the mappings are shown below. One might be tempted to say we don't need three strings. That's true for some fuel types, like Diesel, but FuelOilNo1 will ultimately map to "Fuel Oil No 1" for output variables and table headings in the last phase of this. For now some of the old names are kept to produce no diffs.

**Input choice, Meter Resource, Output Variable**
Electricity, Electricity, Electric
NaturalGas, Gas, Gas
Propane, Propane, Propane
Diesel, Diesel, Diesel
Gasoline, Gasoline, Gasoline
FuelOilNo1, FuelOil#1, FuelOil#1
FuelOilNo2, FuelOil#2, FuelOil#2
Coal, Coal, Coal
OtherFuel1, OtherFuel1, OtherFuel1
OtherFuel2, OtherFuel2, OtherFuel2
Water, Water, Water

4. For every type except NaturalGas, the meter resource type is identical to the input fuel type. And ultimately, NaturalGas will be the same, also. So perhaps the input fuel type can simply be passed directly to the `ResourceTypeKey` in `SetupOutputVariable`.

5. Other related functions are "GetStandardMeterResourceType" which is called from during SetupOutputVariable processing. So, this function could stay as-is and take care of the one oddball meter resource "Gas" which ultimately will move to "NaturalGas". This function should be evaluated for refactoring as well. Even though the input choices should be fixed in the idd/schema, this function should probably remain to trap any oddball nonstandard fuel types that might get added. And it can clean up any upper/lower case inconsistencies.

6. There may be some subtleties here that gum up the works for the no-diff case. For example electricity use is typically "Electric Power" and "Electric Energy" in output variables. While most (all?) other fuels are "Gas Rate", "Gas Energy".

Lots of scattered info here - sorry.

### Details
 - Followup for #5941.

### Checklist

dareumnam:
Pull request overview
---------------------
 - Fixes #7621
 - Removed if blocks for fuel type in the input processing functions and replaced them with standard functions calls 
 - The three standard functions were added in UtilityRoutines.cc & UtilityRoutines.hh
 - `bool ValidateFuelType(std::string const &FuelTypeInput,
                          std::string &FuelTypeOutput,
                          bool &FuelTypeErrorsFound)`
Replace if blocks with **fuel type strings** in:

| Object | File | Fuel Type in the IDD |
| --- | --- | --- |
| Chiller:EngineDriven | PlantChillers.cc | \required-field |
| Chiller:CombustionTurbine | PlantChillers.cc | \default was set |
| ChillerHeater:Absorption:DirectFired | ChillerGasAbsorption.cc | \default was set |
| Generator:InternalCombustionEngine | ICEngineElectricGenerator.cc | \default was set |
| Generator:CombustionTurbine | CTElectricGenerator.cc | \default was set  |
| Generator:MicroTurbine | MicroturbineElectricGenerator.cc | \default was set |

 - `bool ValidateFuelTypeWithFuelTypeNum(std::string const &FuelTypeInput,
                                         int &FuelTypeNum,
                                         bool &FuelTypeErrorsFound)`
Replace if blocks with **fuel type numbers** in :

| Object | File | Fuel Type in the IDD |
| --- | --- | --- |
| AirConditioner:VariableRefrigerantFlow | HVACVariableRefrigerantFlow.cc | \default was set |
| Coil:Cooling:DX:MultiSpeed | DXCoils.cc | \required-field |
| Coil:Heating:DX:MultiSpeed | DXCoils.cc | \required-field |

   - `bool ValidateFuelTypeWithAssignResourceTypeNum(std::string const &FuelTypeInput,
                                                   std::string &FuelTypeOutput,
                                                   int &FuelTypeNum,
                                                   bool &FuelTypeErrorsFound)`
Replace if blocks with **fuel type strings** and `DataGlobalConstants::AssignResourceTypeNum()` in:

| Object | File | Fuel Type in the IDD |
| --- | --- | --- |
| Boiler:HotWater | Boilers.cc | \required-field |
| Boiler:Steam | BoilerSteam.cc | \required-field |

 - Add unit tests
   - BoilerHotWater.unit.cc 
   - BoilerSteam.unit.cc
   - ChillerGasAbsorption.unit.cc
   - DXCoils.unit.cc 
   - HVACVariableRefrigerantFlow.unit.cc
   - PlantChillers.unit.cc
   - CTElectricGenerator.unit.cc (Added to \tst\EnergyPlus\unit\CMakeLists.txt)
   - ICEngineElectricGenerator.unit.cc (Added to \tst\EnergyPlus\unit\CMakeLists.txt)
   - MicroturbineElectricGenerator.unit.cc (Added to \tst\EnergyPlus\unit\CMakeLists.txt)

### Pull Request Author

### Reviewer


src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +658,146 @@ namespace DataGlobalConstants {
         return ResourceTypeChar;
     }
 
+    void ValidateFuelType(std::string FuelTypeInput)
```
mitchute: @Myoldmopar @mjwitte is DataGlobalConstants the right place for this?

dareumnam:
@mitchute This is ready for review now.

src/EnergyPlus/BoilerSteam.cc
```diff
@@ -195,59 +195,20 @@ namespace BoilerSteam {
...
-                    ShowContinueError("Invalid " + DataIPShortCuts::cAlphaFieldNames(2) + '=' + DataIPShortCuts::cAlphaArgs(2));
-
-                    // Set to Electric to avoid errors when setting up output variables
-                    thisBoiler.BoilerFuelTypeForOutputVariable = "Electric";
-                    ErrorsFound = true;
-                }
+            // Validate fuel type input
+            DataGlobalConstants::FuelTypeInput = DataIPShortCuts::cAlphaArgs(2);
+            DataGlobalConstants::ValidateFuelTypeWithAssignResourceTypeNum(DataGlobalConstants::FuelTypeInput);
+            if (DataGlobalConstants::FuelTypeErrorsFound) {
```
rraustad: Where is DataGlobalConstants::FuelTypeErrorsFound set back to false for the next boiler? If the first boiler has an incorrect fuel type then won't this message print for all boilers?

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +658,146 @@ namespace DataGlobalConstants {
...
+        // FUNCTION INFORMATION:
+        //       AUTHOR         Dareum Nam
+        //       DATE WRITTEN   May 2020
+        //       MODIFIED       na
+        //       RE-ENGINEERED  na
+
+        // PURPOSE OF THIS FUNCTION:
+        // Validates fuel types and sets output strings
+
+        auto const SELECT_CASE_var(FuelTypeInput);
```
rraustad: Would a switch statement be more efficient for these functions?

src/EnergyPlus/DataGlobalConstants.hh
```diff
@@ -242,6 +242,11 @@ namespace DataGlobalConstants {
     extern int const iEvapCoolerInDirectRDDSpecial;
     extern int const iEvapCoolerDirectResearchSpecial;
 
+    extern std::string FuelTypeInput;
+    extern std::string FuelType;
+    extern int FuelTypeNum;
+    extern bool FuelTypeErrorsFound;
```
Myoldmopar: I do not understand the need for these global variables.  That new function accepts a `FuelTypeInput` argument.  What is this global variable `FuelTypeInput` for?  I think you need to add two arguments to the function to return back the FuelTypeNum and FuelType values.  No need to add global variables.

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +658,146 @@ namespace DataGlobalConstants {
...
+        } else if (SELECT_CASE_var == "OTHERFUEL1") {
+            FuelType = "OtherFuel1";
+            FuelTypeNum = DataGlobalConstants::AssignResourceTypeNum("OTHERFUEL1");
+
+        } else if (SELECT_CASE_var == "OTHERFUEL2") {
+            FuelType = "OtherFuel2";
+            FuelTypeNum = DataGlobalConstants::AssignResourceTypeNum("OTHERFUEL2");
+
+        } else {
+            FuelTypeErrorsFound = true;
```
Myoldmopar: Instead of having a global variable for fuel type errors found, I would, for now:
- Make this function a `bool` instead of a `void`
- Return true if the function successfully matches a fuel type, and return false if not -- OR return true if errors were found and false if it succeeds, but just be clear about it.

Then the calling function can handle things appropriately.

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +658,146 @@ namespace DataGlobalConstants {
...
+        } else if (UtilityRoutines::SameString(FuelTypeInput, "OtherFuel1")) {
+            FuelTypeNum = 8; // FuelTypeOtherFuel1
+        } else if (UtilityRoutines::SameString(FuelTypeInput, "OtherFuel2")) {
+            FuelTypeNum = 9; // FuelTypeOtherFuel2
+        } else {
+            FuelTypeErrorsFound = true;
+        }
+    }
+
+    void ValidateFuelTypeWithAssignResourceTypeNum(std::string FuelTypeInput)
```
Myoldmopar: As I mentioned in another comment, you need to pass in and out the `FuelType` string and the `FuelTypeNum` integer.  This input string also needs to be a constant reference so that it isn't copied when this function is called.  The new signature should be:

```
bool ValidateFuelTypeWithAssignResourceTypeNum(
    std::string const & FuelTypeInput,
    std::string & FuelType,
    int & FuelTypeNum
);
```

src/EnergyPlus/DataGlobalConstants.cc
Myoldmopar: Can't switch on a string, unfortunately.  In the functions that convert integer to the appropriate string, you can switch on the integers.

src/EnergyPlus/DataGlobalConstants.cc
Myoldmopar: I'm not sure.  Maybe UtilityRoutines.cc?

tst/EnergyPlus/unit/ICEngineElectricGenerator.unit.cc
```diff
@@ -0,0 +1,138 @@
...
+        "  0,                       !- Minimum Value of x",
+        "  1;                       !- Maximum Value of x",
+    });
+
+    ASSERT_TRUE(process_idf(idf_objects));
+
+    GetICEngineGeneratorInput();
+
+    EXPECT_EQ(DataGlobalConstants::FuelType, "Diesel");
+    EXPECT_FALSE(DataGlobalConstants::FuelTypeErrorsFound);
```
Myoldmopar: Since we are taking the global variables back out, you won't be able to check this.  And that's OK.  In the GetInput routine, the code that is performing input validation should check the value returned from the fuel validation function, and decide what to do.  If an error occurs, the GetInput routine should wrap up and throw a fatal error.  In that case, you can test successful fuel entries just by calling GetInput on good values.  And then you can test bad fuel entries by using EXPECT_THROW.  You can find examples if you search for that text in the unit test code.

Myoldmopar (changes requested):
The biggest change needed here is to remove the global variables.  Those can be eliminated by just passing extra variables into and out of the new worker functions, and having the function return true/false for success or failure.  Then the calling code needs only to check the return value and decide whether to gracefully throw a fatal or continue on.  Unit tests will need to be updated accordingly.

Other than that, the fuel function code seems fine, I'll give it a final pass once the global variables are taken care of.

src/EnergyPlus/DataGlobalConstants.cc
rraustad: Be careful, if there are other errors in GetInput, then this function could over-ride ErrorsFound to false.

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +659,112 @@ namespace DataGlobalConstants {
...
+            return OTHERFUEL2;
+    }
+
+    void ValidateFuelType(std::string FuelTypeInput)
+    {
+        // FUNCTION INFORMATION:
+        //       AUTHOR         Dareum Nam
+        //       DATE WRITTEN   May 2020
+        //       MODIFIED       na
+        //       RE-ENGINEERED  na
```
rraustad: Please delete these comment lines with "na" in a future commit. I don't mind the name/date comments (could be 1 line though) but other disagree. So at the very least get rid of the na lines. Keep new functions as slim as possible.

rraustad:
@dareumnam nice job so far!

src/EnergyPlus/DataGlobalConstants.cc
rraustad: I think this warrants another function argument for &ErrorsFound. Where an error is passed back only if true. @jmarrec updated this a while back. See GetFanInletNode as an example.

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +659,112 @@ namespace DataGlobalConstants {
...
+        FUELOILNO2,
+        PROPANE,
+        OTHERFUEL1,
+        OTHERFUEL2
+    };
+
+    FuelTypeNumber ConvertFueltypeString(std::string FuelTypeInput)
+    {
+        if (FuelTypeInput == "ELECTRICITY")
+            return ELECTRICITY;
```
rraustad: This is so much better.

src/EnergyPlus/DataGlobalConstants.cc
```diff
@@ -652,6 +659,112 @@ namespace DataGlobalConstants {
...
+        // Validates fuel types and sets output strings
+
+        FuelTypeNumber Fuel = ConvertFueltypeString(FuelTypeInput);
+
+        switch (Fuel) {
+        case ELECTRICITY:
+            FuelType = "Electric";
+            FuelTypeNumforAssignResource = DataGlobalConstants::AssignResourceTypeNum("ELECTRICITY");
+            FuelTypeNum = 1; // FuelTypeElectricity in DXCoils.cc and HVACVariableRefrigerantFlow.cc
+            break;
```
rraustad: So is setting 3 vars efficient? Is FuelTypeNum only used in DXCoils and HVACVariableRefrigeratntFlow? Seems like you need only 1 int here. Is FuelTypeNumforAssignResource and FuelTypeNum the same value? Seems likely.

dareumnam: FuelTypeNum is used in DXcoils and HVACVariantRefrigerantFlow, and FuelTypeNumforAssignResouce is used in Boilers and Boilersteam with AssignResourceTypeNum() function (https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/DataGlobalConstants.cc#L325-L490), but these two are not the same int value for each fuel type, so once I separated these two int vars. I'll figure out if these can be combined while removing the global vars. Thank you for your insight.

src/EnergyPlus/BoilerSteam.cc
dareumnam: Thank you for pointing out. I added lines to set the error flag back.

src/EnergyPlus/DataGlobalConstants.cc
dareumnam: I'll move the function to UtilityRoutines.cc!

src/EnergyPlus/DataGlobalConstants.cc
dareumnam: Thanks for your insights. Since C++ can't switch on a string, once I used enum to convert strings to integer and make one standard function instead of the three functions. But I think I might need to return back to if-else blocks while removing global vars.

mitchute:
@Myoldmopar @rraustad It looks like Dareum got this wrapped up and addressed your comments. Do you need any other changes here before this goes in? I'm merging in develop now to let CI take one more pass.

Myoldmopar (approved):
It's so nice having all this functionality in one location.  The HVACVRF file changes are difficult to mine through because the formatting did so much, but I think that's OK.  As long as CI is happy with it, I trust it didn't do any damage in there.

src/EnergyPlus/UtilityRoutines.hh
```diff
@@ -573,6 +573,12 @@ namespace UtilityRoutines {
 
     inline bool exists(const std::string& filename);
 
+    bool ValidateFuelType(std::string const &FuelTypeInput, std::string &FuelTypeOutput, bool &FuelTypeErrorsFound);
+
+    bool ValidateFuelTypeWithFuelTypeNum(std::string const &FuelTypeInput, int &FuelTypeNum, bool &FuelTypeErrorsFound);
+
+    bool ValidateFuelTypeWithAssignResourceTypeNum(std::string const &FuelTypeInput, std::string &FuelTypeOutput, int &FuelTypeNum, bool &FuelTypeErrorsFound);
+
```
mjwitte, line 581: @dareumnam I know this it merged already, but I'm wondering why there are three separate functions here? Seems like these could be collapsed into one function, or at least build off each other so that there's only one place with ifs on the various fuel type strings.

Also, there is still a similar function in [`ExteriorEnergyUse::ValidateFuelType`](https://github.com/NREL/EnergyPlus/blob/8e084f1905da7c4c88baba8b67fcdf3bdd066dab/src/EnergyPlus/ExteriorEnergyUse.cc#L449-L454).

dareumnam: @mjwitte Thanks for your comment. This was to remain each object as original as possible, particularly setup output variables and error messages. As you comment, they can be collapsed into a single function if we want to do that, but it needs every object to get fixed. I can do that if it is needed.

mjwitte: @dareumnam I would like to finish addressing #7622 before the I/O freeze on Aug 19. The changes for Fuel Oil are waiting for final review in #8134. Then we can build on that and finish up all of the fuel/meter naming work.

dareumnam: @mjwitte I see. Then I'll wrap up #7622 right after #8134 gets merged. Thanks for letting me know.

mjwitte: @dareumnam Also, note a slight change from the original plan. For output variables we will now use the condensed PascalCase names, not the expanded names. For example "Boiler NaturalGas Energy". This is different from the original plan that was laid out in #7622. The only place expanded names will be used is for table headings (Natural Gas).

PR #8005, opened 2020-05-19
Prelude to #7388 - dependency bump for nlohmann json and valijson

jmarrec:
Pull request overview
---------------------

Prelude to #7388 - dependency bump for nlohmann json and valijson

* valijson bumped to current master (https://github.com/tristanpenman/valijson/commit/92d369e91605c73c32587fb777738f17d3e23405), I could have bumped only to v0.2 I suppose, but reviewing the 5 commits to master snce v0.2 I see no reason no to use recent ones
* While I was at it (and early on) I also bumped nlohmann's json from 3.7.0 to 3.7.3
* Create a patch file for both, and filled the third_party/README.md with info
* Update the InputValidation to match the new output and updated on test that checks for a RunPeriod with no name since it is affected.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
[...]

PR #8006, opened 2020-05-19
#7388 -  Non UTF-8 characters throwing false blank name error

### LINKED ISSUE #7388: Non UTF-8 characters throwing false blank name error
Issue overview
--------------
Non-UTF-8 characters causing false blank name error. The defect file from #7048 fails in v9.2.0-26151c5477 with this error:
```
   ** Severe  ** <root>[Construction] - Object contains properties that could not be validated using 'properties' or 'additionalProperties' constraints
   ** Severe  ** <root>[Construction] - Object name is required and cannot be blank or whitespace
   **  Fatal  ** Errors occurred on processing input file. Preceding condition(s) cause termination.
```
@Nigusse [traced the problem](https://github.com/NREL/EnergyPlus/issues/7048#issuecomment-504454488) to this object:
```
  Construction,
    solaio pavimento(1°piano)/2°piano,  !- Name
```

@mbadams5 [commented](https://github.com/NREL/EnergyPlus/issues/7048#issuecomment-505223437): ... if the characters are UTF-8, it should just work. If they are not UTF-8 then it should still not be a blank name.

Using a text editor to convert the file to UTF-8 eliminates the blank name error. So, it appears that the original file is not UTF-8.

### Details
 - Version of EnergyPlus v9.2.0-26151c5477

### Checklist
Note this file has been modified slightly to avoid the original problems in #7048. The ...UTF-8 file runs to completion, the other one fails with blank name.

jmarrec:
Pull request overview
---------------------
 - Fixes #7388 
 - Non UTF-8 characters throwing false blank name error

This PR includes #8005 which deals with the third party dependency bumps only.
A test was added.

The last commit is optional, and allows latin1 characters to go through. Basically RE2 (the regex engine used) is strict about UTF8 unless you tell it to parse as latin1.
valijson official uses std::regex, which is fine about non utf8 chars.

I'm not sure it is better to allow these non strictly-utf8 names through than throw a specific warning about encoding. The specific defect file on E+ DevSupport will run fine with the last commit though.

### Pull Request Author

### Reviewer


mitchute:
@Myoldmopar @mbadams5 if you can, feel free to weigh in on the optional commit to allow latin1 characters. https://github.com/NREL/EnergyPlus/pull/8006/commits/3b74470783f89fc91d5acb7590ae20f6bdbcccea I'm not going to be able to say either way on that one.

I'm gonna do some testing locally.

mbadams5:
I don't think we should allow Latin1 characters in validation. The JSON library can only handle UTF-8 characters, and while it can ignore or replace non-UTF-8 characters, I'd rather fail sooner. I think it is appropriate for validation to throw an error if it cannot parse as UTF-8 (with an appropriate error message).

Myoldmopar:
> I'd rather fail sooner.

I'm all in favor of this.  If I understand correctly, we need to revert 3b7447078.  I can do that, pull in latest develop, and push.

Myoldmopar:
Pulled in latest develop, reverted latin1 commit, and built.  Unit tests all pass.  I think this will be good to go if CI agrees.

src/EnergyPlus/InputProcessing/InputValidation.cc
```diff
@@ -100,10 +104,12 @@ bool Validation::validate(json const &parsed_input)
                     context += *it;
 
                 errors_.emplace_back(context + " - " + error.description);
-                if (max_context == 2 &&
-                    error.description ==
-                        "Object contains a property that could not be validated using 'properties' or 'additionalProperties' constraints: ''.") {
-                    errors_.emplace_back(context + " - Object name is required and cannot be blank or whitespace");
+                if (max_context == 2) {
+                    if (error.description == nameError) { // C++20: should mark this [[likely]]
```
Myoldmopar, line 108: In this case, since we are doing string comparisons, and also this is just in the input processing part of the run, and also this is not expected to be a common occurrence, I'm not sure we'll need to add the `[[likely]]` declaration.  But in any case, the error handling change is good.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -623,6 +623,39 @@ TEST_F(InputProcessorFixture, parse_bad_utf_8_json_3)
...
 }
 
+TEST_F(InputProcessorFixture, parse_latin1_json)
+{
+
+    // Test for #7388 -  Non UTF-8 characters throwing false blank name error
+
+    std::string const idf(delimited_string({
+        "  Construction,",
+        "    \x31\xB0\x70\x69\x61\x6E\x6F,  !- Name",
```
Myoldmopar, line 633: Did someone let Elon Musk name this construction object?

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -623,6 +623,39 @@ TEST_F(InputProcessorFixture, parse_bad_utf_8_json_3)
...
+    EXPECT_EQ("<root>[Construction] - Object contains a property that could not be validated using 'properties' or 'additionalProperties' constraints: '1\xB0piano'.", errors[0]);
+    EXPECT_EQ("<root>[Construction] - Object name is required and cannot be blank or whitespace, and must be UTF-8 encoded", errors[1]);
+
+
+    json &epJSON = getEpJSON();
+
+    auto it = epJSON.find("Construction");
+    ASSERT_NE(it, epJSON.end());
+    auto iit = it->begin();
+    EXPECT_EQ("1\xB0piano", iit.key());
```
Myoldmopar, line 655: :+1:

Myoldmopar (approved):

Myoldmopar:
Good to go.  Thanks @jmarrec for the fix and @mbadams5 for chiming in.

tst/EnergyPlus/unit/InputProcessor.unit.cc
jmarrec: :) I'm just faking a latin1 encoding here. In the Defect file, the only "problem" is the use of `°`  and the fact that the file is latin-1 encoded (well, a derivative, it's probably a Windows-1252 if I were to guess)

PR #7462, opened 2019-08-17
Schedules which change pattern during Daylight Savings Time cause incorrect results at hour 23

rraustad:
Pull request overview
---------------------
Fixes #5155. When a schedule is used when DST is active, the program will look at hour 1 of the schedule when time of day = 24. This change looks to the next day's schedule for hour 1.

### Work Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] At least one of the following appropriate labels must be added to this PR to be consumed into the changelog:
   - Defect: This pull request repairs a github defect issue.  The github issue should be referenced in the PR description
   - Refactoring: This pull request includes code changes that don't change the functionality of the program, just perform refactoring
   - NewFeature: This pull request includes code to add a new feature to EnergyPlus
   - Performance: This pull request includes code changes that are directed at improving the runtime performance of EnergyPlus
   - DoNoPublish: This pull request includes changes that shouldn't be included in the changelog

### Review Checklist
This will not be exhaustively relevant to every PR.
 - [ ] Functional code review (it has to work!)
 - [ ] If defect, results of running current develop vs this branch should exhibit the fix
 - [ ] CI status: all green or justified
 - [ ] Performance: CI Linux results include performance check
 - [ ] Unit Test(s)
 - C++ checks:
   - [ ] Argument types
   - [ ] If any virtual classes, ensure virtual destructor included, other things
 - IDD changes:
   - [ ] Verify naming conventions and styles, memos and notes and defaults
   - [ ] Open windows IDF Editor with modified IDD to check for errors
   - [ ] If transition, add to input rules file for interfaces
   - [ ] If transition, add transition source
   - [ ] If transition, update idfs
 - [ ] If new idf included, locally check the err file and other outputs
 - [ ] Required documentation updates?
 - [ ] ExpandObjects changes?
 - [ ] If output changes, including tabular output structure, add to output rules file for interfaces

PR #8024, opened 2020-05-29
Fix #8022 - ScheduleManager::GetCurrentScheduleValue and LookUpScheduleValue return different results

### LINKED ISSUE #8022: ScheduleManager::GetCurrentScheduleValue and LookUpScheduleValue return different results
Issue overview
--------------

In the course of investigating #7814, I found out that there could be some potential problems in the way `ScheduleManager` does its stuff. #7814 contains a discussion on the subject, but the key point is that `GetCurrentScheduleValue` appears to return different value than `LookUpScheduleValue` which is perhaps surprising, so we should ensure that everything is correct as is or fix it.

I'm taking out key points on the #7814 thread for convenience, but be sure to check the original discussion also:

ScheduleManager.cc has three routines with similar blocks of code.
These two are **exactly** the same (minus whitespace changes)

* [ScheduleManager::ReportScheduleValues](https://github.com/NREL/EnergyPlus/blob/36db22b9af45b33371b30b85de916659c091c7c8/src/EnergyPlus/ScheduleManager.cc#L4952-L4985) (called from `HeatBalanceManager::ReportHeatBalance`)
* [ScheduleManager::UpdateScheduleValues](https://github.com/NREL/EnergyPlus/blob/36db22b9af45b33371b30b85de916659c091c7c8/src/EnergyPlus/ScheduleManager.cc#L2805-L2839) (called from `WeatherManager::SetCurrentWeather`)

This one is much different: [ScheduleManager::LookUpScheduleValues](https://github.com/NREL/EnergyPlus/blob/36db22b9af45b33371b30b85de916659c091c7c8/src/EnergyPlus/ScheduleManager.cc#L2912-L2975). This one can also be cleaned up as it does stuff it doesn't have to.

The difference appears to be the determination of the week/day schedule pointers, in one case (LookUp) it's done with `HourOfDay` then after determination is done `DSTIndicator` is added, for the other it's directly done with `HourOfDay + DSTIndicator`

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): current develop https://github.com/NREL/EnergyPlus/commit/404930d1730dee0df0ca94f672e19113ff2b8d5a

### Checklist

jmarrec:
Pull request overview
---------------------
 - Fixes #8022 - ScheduleManager::GetCurrentScheduleValue and LookUpScheduleValue return different results

### Pull Request Author

### Reviewer


src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2743,12 +2743,13 @@ namespace ScheduleManager {
...
         }
 
+        // Checking if valid index is passed is necessary
         if (ScheduleIndex == -1) {
             return 1.0;
         } else if (ScheduleIndex == 0) {
             return 0.0;
         } else if (!Schedule(ScheduleIndex).EMSActuatedOn) {
-            return Schedule(ScheduleIndex).CurrentValue;
+            return Schedule(ScheduleIndex).CurrentValue;  // This block probably unecessary, UpdateScheduleValues already does it
```
jmarrec, line 2752: TODO

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2792,59 +2793,26 @@ namespace ScheduleManager {
...
-                } else {
-                    WeekSchedulePointer = Schedule(ScheduleIndex).WeekSchedulePointer(DayOfYear_Schedule + 1);
-                    if (DayOfWeek <= 7 && HolidayIndex > 0) {
-                        DaySchedulePointer = WeekSchedule(WeekSchedulePointer).DaySchedulePointer(7 + HolidayIndex);
-                    } else {
-                        DaySchedulePointer = WeekSchedule(WeekSchedulePointer).DaySchedulePointer(DayOfWeek);
-                    }
-                    Schedule(ScheduleIndex).CurrentValue = DaySchedule(DaySchedulePointer).TSValue(NumOfTimeStepInHour, WhichHour - 24);
-                }
+                Schedule(ScheduleIndex).CurrentValue = LookUpScheduleValue(ScheduleIndex, DataGlobals::HourOfDay, DataGlobals::TimeStep);
```
jmarrec, line 2805: `UpdateScheduleValues` now calls `LookUpScheduleValues` with current day and timestep, to avoid duplicating code.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2792,59 +2793,26 @@ namespace ScheduleManager {
...
             }
         }
     }
 
     Real64 LookUpScheduleValue(int const ScheduleIndex,
-                               int const ThisHour,    // Negative => unspecified
+                               int const ThisHour,
```
jmarrec, line 2811: This is never called without specifying ThisHour, so make it required.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2880,101 +2848,53 @@ namespace ScheduleManager {
...
-        int DaySchedulePointer;
-        int WhichHour;
-        int WhichTimeStep;
 
-        if (!ScheduleInputProcessed) {
-            ProcessScheduleInput(OutputFiles::getSingleton());
-            ScheduleInputProcessed = true;
+
+        if (ThisHour > 24) {
+            ShowFatalError("LookUpScheduleValue called with thisHour=" + fmt::to_string(ThisHour));
```
jmarrec, line 2852: Can be removed probably...

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2880,101 +2848,53 @@ namespace ScheduleManager {
...
+        // Hourly Value
+        int thisHour = ThisHour + DSTIndicator;
+        int thisDayOfYear = DataEnvironment::DayOfYear_Schedule;
+        int thisDayOfWeek = DataEnvironment::DayOfWeek;
+        int thisHolidayIndex = DataEnvironment::HolidayIndex;
+        if (thisHour > 24) { // In case HourOfDay is 24 and DSTIndicator is 1, you're actually the next day
+            thisDayOfYear += 1;
+            thisHour -= 24;
+            thisDayOfWeek = DataEnvironment::DayOfWeekTomorrow;
+            thisHolidayIndex = DataEnvironment::HolidayIndexTomorrow;
```
jmarrec, line 2877: This is new by the way. But I think it makes sense... If you're looking at tomorrow, then update the DayOfWeek and the HolidayIndex accordingly.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2880,101 +2848,53 @@ namespace ScheduleManager {
...
-                        DaySchedulePointer = WeekSchedule(WeekSchedulePointer).DaySchedulePointer(DayOfWeekTomorrow);
-                    }
-                    WhichHour -= 24;
-                }
-            } else {
-                // Determine which Week Schedule is used
-                //  Cant use stored day of year because of leap year inconsistency
-                WeekSchedulePointer = Schedule(ScheduleIndex).WeekSchedulePointer(DayOfYear_Schedule);
+        int WeekSchedulePointer = Schedule(ScheduleIndex).WeekSchedulePointer(thisDayOfYear);
+        int DaySchedulePointer;
```
jmarrec, line 2881: After that code is much dryer, but it should be functionally equivalent to the old one (aside from DayOfWeek/HolidayIndex being updated to tomorrow)

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -2880,101 +2848,53 @@ namespace ScheduleManager {
...
-            } else {
-                if (WhichHour <= 24) {
-                    LookUpScheduleValue = DaySchedule(DaySchedulePointer).TSValue(NumOfTimeStepInHour, WhichHour);
-                } else {
-                    LookUpScheduleValue = DaySchedule(DaySchedulePointer).TSValue(NumOfTimeStepInHour, WhichHour - 24);
-                }
-            }
+        // TODO: the (thisDayOfWeek < 7) looks fishy and maybe uncessary... how about if there are more than 7 holidays?
+        // It should use 7 + thisHolidayIndex in that case but it won't
+        if (thisDayOfWeek <= 7 && thisHolidayIndex > 0) {
```
jmarrec, line 2885: TODO

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -4949,40 +4865,9 @@ namespace ScheduleManager {
...
-                    } else {
-                        DaySchedulePointer = WeekSchedule(WeekSchedulePointer).DaySchedulePointer(DayOfWeek);
-                    }
-                    Schedule(ScheduleIndex).CurrentValue = DaySchedule(DaySchedulePointer).TSValue(NumOfTimeStepInHour, WhichHour - 24);
-                }
-            }
-        }
+        // TODO: Is this needed?
+        // Why is it doing exactly the same as UpdateScheduleValues?
+        UpdateScheduleValues();
```
jmarrec, line 4868: TODO. I noticed it does exactly the same as UpdateScheduleValues, so replace it with a call to UpdateScheduleValues. But is it really necessary?

tst/EnergyPlus/unit/ScheduleManager.unit.cc
```diff
@@ -750,3 +751,56 @@ TEST_F(EnergyPlusFixture, ScheduleFileColumnSeparator)
...
+    EXPECT_EQ(1.0, ScheduleManager::Schedule(1).CurrentValue);
+    EXPECT_EQ(1.0, ScheduleManager::GetCurrentScheduleValue(1));
+
+    DataEnvironment::DSTIndicator = 1; // DST IS ON
+    ScheduleManager::UpdateScheduleValues();
+    // Since DST is on, you're actually on the next day, on 6/1 at 1:00
+    // so it **should** return 3.0
+    EXPECT_EQ(3.0, ScheduleManager::Schedule(1).CurrentValue);
+    EXPECT_EQ(3.0, ScheduleManager::GetCurrentScheduleValue(1));
+    EXPECT_EQ(3.0, ScheduleManager::LookUpScheduleValue(1, DataGlobals::HourOfDay, DataGlobals::TimeStep));
```
jmarrec, line 805: Main test. 803-805 is what I think it should be doing.

jmarrec:
I clearly need to look into it some more given the integration test results on linux.

jmarrec:
The failing unit and integration tests are due to the SolarShading module.

`SolarShading::SkyDifSolarShading` calls `SkyDifSolarShading::SHADOW(0,0)`, on this line:

https://github.com/NREL/EnergyPlus/blob/f4f554d594ae993ebbf2567fc68450b8fda67983/src/EnergyPlus/SolarShading.cc#L10591

so with hour = 0... which will make this routine crash since TSValue is 1-indexed, unless you either:

* Do like the prevous ScheduleManager::LookUpScheduleValue was doing here, if Hour = 0, then Hour = 24: https://github.com/NREL/EnergyPlus/blob/f4f554d594ae993ebbf2567fc68450b8fda67983/src/EnergyPlus/ScheduleManager.cc#L2929-L2931
* Or call SHADOW(24,0) directly instead.

Myoldmopar:
This has a minor build warning.  I'll pull develop and fix that real quick and we'll see what CI says.

Myoldmopar:
CI is showing big diffs on the leap year annual file.  I'm wondering exactly what change caused the diffs.  The tomorrow-stuff?

mjwitte:
@Myoldmopar Make sure this doesn't undo https://github.com/NREL/EnergyPlus/pull/7462?

Myoldmopar:
Good point.  But the unit test that @JasonGlazer added is still passing with these changes in place, and the only change is that the DayOfWeekTomorrow needed to be assigned in @JasonGlazer's unit test.   So it _should_ be OK.......but I can verify.

Myoldmopar:
Also, I don't intend on reviewing/testing further until the leap year annual diffs are resolved/justified.

JasonGlazer:
@Myoldmopar what unit test are you talking about? Did you mean @jmarrec instead of me in your comment above?

Myoldmopar:
@JasonGlazer sorry for the confusion.  @mjwitte brought up PR #7462 in the comment above and I saw your name on that PR.  But looking closer, I see that while you were the one who merged it, those changes were actually committed by @rraustad.

jmarrec:
I've been trying to pinpoint the difference in the `_5ZoneAirCooled_LeapYear_annual.idf` for hours now. I've tried a few things where I couldn't find any differences. I looked at the variable `('CENTRAL BOILER', 'Boiler Heating Rate', 'W')` and finally got some sense of when it is happening, and it seems to be periodic:

image

If I filter on any diff that's larger than 20, I get something like this;

image


**So it seems that it happens at 6:00 (or once at 7:00) AM every monday during DST**

I think it is because of the fact that I update to DayOfWeekTomorrow now (and HolidayIndexTomorrow) as I highlighted here: https://github.com/NREL/EnergyPlus/pull/8024#discussion_r432595764

The fact that it happens at 6:00 and not at midnight is probably just some kind of delay (thermal mass?)


**Edit: Confirmed**

If I comment out the lines in https://github.com/NREL/EnergyPlus/pull/8024#discussion_r432595764, now I get 0 difference.

jmarrec:
Another test I ran is to create a Schedule:Compact that just ramps up, and get the resulting Output:Variable Schedule Value.

Something like:

```
Output:Variable,RampingUp,Schedule Value,Hourly;
Output:Variable,RampingUp,Schedule Value,Timestep;

  Schedule:Compact,
    RampingUp,                 !- Name
    Any Number,                !- Schedule Type Limits Name
    Through: 1/1,
    For: AllDays,
     Until: 1:00, 1,
     Until: 2:00, 2,
     Until: 3:00, 3,
     Until: 4:00, 4,
     Until: 5:00, 5,
     Until: 6:00, 6,
     Until: 7:00, 7,
     Until: 8:00, 8,
     Until: 9:00, 9,
     Until: 10:00, 10,
     Until: 11:00, 11,
     Until: 12:00, 12,
     Until: 13:00, 13,
     Until: 14:00, 14,
     Until: 15:00, 15,
     Until: 16:00, 16,
     Until: 17:00, 17,
     Until: 18:00, 18,
     Until: 19:00, 19,
     Until: 20:00, 20,
     Until: 21:00, 21,
     Until: 22:00, 22,
     Until: 23:00, 23,
     Until: 24:00, 24,
    Through: 1/2,
    For: AllDays,
     Until: 1:00, 25,
     Until: 2:00, 26,
     [...]
    Through: 12/31,
    For: AllDays,
     Until: 1:00, 8761,
     Until: 2:00, 8762,
     Until: 3:00, 8763,
     Until: 4:00, 8764,
     Until: 5:00, 8765,
     Until: 6:00, 8766,
     Until: 7:00, 8767,
     Until: 8:00, 8768,
     Until: 9:00, 8769,
     Until: 10:00, 8770,
     Until: 11:00, 8771,
     Until: 12:00, 8772,
     Until: 13:00, 8773,
     Until: 14:00, 8774,
     Until: 15:00, 8775,
     Until: 16:00, 8776,
     Until: 17:00, 8777,
     Until: 18:00, 8778,
     Until: 19:00, 8779,
     Until: 20:00, 8780,
     Until: 21:00, 8781,
     Until: 22:00, 8782,
     Until: 23:00, 8783,
     Until: 24:00, 8784;
```

image

src/EnergyPlus/ScheduleManager.cc
jmarrec:
```
            thisDayOfWeek = DataEnvironment::DayOfWeekTomorrow;
            thisHolidayIndex = DataEnvironment::HolidayIndexTomorrow;
```

This is the cause of the difference in the `_5ZoneAirCooled_LeapYear_annual.idf`

rraustad:
So then the schedule value at 6:00 AM is now correct? In the diff file _5ZoneAirCooled_LeapYear_annual.idf what are the boiler SPs at 5:45, 6:00 and 6:15 and how do those numbers compare to the SP in the csv results?

jmarrec:
I think it comes from this pattern in the schedules:

FanAvailShcedule and a few other like the Htg-SetP-Sch
```
  Schedule:Compact,
    ReheatCoilAvailSched,    !- Name
    Through: 3/31,           !- Field 1
    Until: 24:00,1.0,        !- Field 3
    Through: 9/30,           !- Field 5
    For: WeekDays,           !- Field 6
    Until: 6:00,0.0,         !- Field 7
    Until: 20:00,1.0,        !- Field 9
    Until: 24:00,0.0,        !- Field 11
    For: SummerDesignDay WinterDesignDay, !- Field 13
    Until: 24:00,1.0,        !- Field 14
    For: AllOtherDays,       !- Field 16
    Until: 24:00,0.0,        !- Field 17
    Through: 12/31,          !- Field 19
    For: AllDays,            !- Field 20
    Until: 24:00,1.0;        !- Field 21

  Schedule:Compact,
    Htg-SetP-Sch,            !- Name
    Temperature,             !- Schedule Type Limits Name
    For: SummerDesignDay,    !- Field 2
    Until: 24:00,16.7,       !- Field 3
    For: WinterDesignDay,    !- Field 5
    Until: 24:00,22.2,       !- Field 6
    For: WeekDays,           !- Field 8
    Until: 6:00,16.7,        !- Field 9
    Until: 20:00,22.2,       !- Field 11
    Until: 24:00,16.7,       !- Field 13
    For: WeekEnds Holiday,   !- Field 15
    Until: 24:00,16.7,       !- Field 16
    For: AllOtherDays,       !- Field 18
    Until: 24:00,16.7;       !- Field 19

  Schedule:Compact,
    Min OA Sched,            !- Name
    For: Weekdays,           !- Field 2
    Until: 6:00,0.02,        !- Field 3
    Until: 20:00,1.0,        !- Field 5
    Until: 24:00,0.02,       !- Field 7
    For: AllOtherDays,       !- Field 9
    Until: 24:00,0.02;       !- Field 10

  Schedule:Compact,
    FanAvailSched,           !- Name
    Through: 3/31,           !- Field 1
    Until: 24:00,1.0,        !- Field 3
    Through: 9/30,           !- Field 5
    For: WeekDays,           !- Field 6
    Until: 6:00,0.0,         !- Field 7
    Until: 20:00,1.0,        !- Field 9
    Until: 24:00,0.0,        !- Field 11
    For: SummerDesignDay WinterDesignDay, !- Field 13
    Until: 24:00,1.0,        !- Field 14
    For: AllOtherDays,       !- Field 16
    Until: 24:00,0.0,        !- Field 17
    Through: 12/31,          !- Field 19
    For: AllDays,            !- Field 20
    Until: 24:00,1.0;        !- Field 21

```

jmarrec:
The Schedules are exactly the same if you look at the E+ Output:Variable Schedule Value, (eg: `('REHEATCOILAVAILSCHED', 'Schedule Value', '')`), what's different is the way the value to be used is used internally by E+ which isn't facilitating things in order to demonstrate whether the DayOfWeekTomorrow change is good or not (I think the lines I added make sense, but that's not saying there isn't something else that's weird for eg in WeatherManager...)

rraustad:
So if you look at 'Htg-SetP-Sch', 'Schedule Value' before and after this fix, they are different, and with this branch are correct?

jmarrec:
Thanks for your help @rraustad !

I think I found something here. I've made modifications to the Schedule:Compacts, so that they are different between 0:00 and 6:00 on weekdays versus weekends.

```

  Schedule:Compact,
    Htg-SetP-Sch,            !- Name
    Temperature,             !- Schedule Type Limits Name
    For: SummerDesignDay,    !- Field 2
    Until: 24:00,16.7,       !- Field 3
    For: WinterDesignDay,    !- Field 5
    Until: 24:00,22.2,       !- Field 6
    For: WeekDays,           !- Field 8
    Until: 6:00,16.7,        !- Field 9
    Until: 20:00,22.2,       !- Field 11
    Until: 24:00,16.7,       !- Field 13
    For: WeekEnds Holiday,   !- Field 15
    Until: 24:00,0,          !- Field 16
    For: AllOtherDays,       !- Field 18
    Until: 24:00,0;          !- Field 19
```

Notice that on Weekends, it's now 0, and look at the results now. As you can see with the fix, it correctly applies the weekend condition. Without the fix (current develop (Original)), it doesn't increment the DayOfWeek and therefore picks up 16.7 instead of 0!

image

This might look weird a bit here since I have used my own time index with start of interval conditions `sch_val_tz.index = pd.date_range(start='2016-01-01 00:00', freq='15min', periods=8784*4, tz='America/New_York')` ('as you may know, Python is REALLY not a fan of the "24:00" time...)

 I will do the same with the E+ Time info.

jmarrec:
here the raw output from the SQL, using this query:


```
SELECT Month,Day,Hour,Minute,Dst,Interval,DayType,Value From ReportVariableWithTime
WHERE KeyValue='HTG-SETP-SCH'
AND Month=3
AND ((Day=18 AND (Hour=22 OR Hour=23 OR Hour=24))
    OR (Day=19 AND (Hour=0 OR Hour=1)))
AND ReportingFrequency='Zone Timestep'
```

### Before:

image


### After:
image


(the DayType reported int the SQL assumes no DST FYI...)

rraustad:
In the After SQL query, why does row 6 -9 show 0? Should be 16.7, right? Are those figures switched (i.e., before is really after)?

jmarrec:
@rraustad This is the way E+ reports time really: it reports Day/Hour as if DST wasn't applied.

Let me illustrate this

## 1. Query the SQL ReportVariableWithTime

image

## 2. Add a column with the localized timestamp

The DST dates match, as I've modified the example file DST:

```
  RunPeriodControl:DaylightSavingTime,
    2nd Sunday in March,                    !- Start Date
    1st Sunday in November;                    !- End Date
```

As you can see, the localize timestamp has a start-of-interval convention:

image

## Refilter like the SQL query above

Notice that the 'DST' column is indeed 1, and the localized timestamp does show UTC-4 (DST) instead of UTC-5 (EST)

Screenshot from 2020-06-26 09-31-05


----


Is the above clearing any doubt you have? Otherwise let me know please.

rraustad:
Yeah, I forgot about the shift due to DST.

Myoldmopar:
@rraustad if you feel satisfied with this, feel free to merge.

rraustad:
With @jmarrec brilliant test with a schedule that increments throughout the year then I think this is complete. However, I have one last question. In the figure from June 25, 2020 there are 2 lines, ORIGINAL - RAMPINGUP Sch Value and New - RAMPINGUP Sch Value. Why isn't the original line showing problems?

rraustad:
I guess it could be the resolution of that figure, where problems with a schedule off by 1 can't be seen on a scale of 9000.

jmarrec:
I think scale is the problem there. We're talking about an offset of 1/8784 really.i guess it was a bad idea in the first place to draw this graph, but I didn't know where to look initially so as with any data exploration sometimes you ask the wrong questions.
If you really want me too I can redo that part next week, but keep in mind that it is for sure more processed than the raw sql query anyways.
(As part of fixing 8111 I will make the ramp up thing an actual unit test too FYI.)

rraustad:
@jmarrec I don't want to make more work than necessary to address the issue at hand. You did, however, intrigue me with a tricked out unit test. I will let you decide whether or not that unit test is necessary and consider this issue closed. Thanks for your effort and insight into your debugging techniques.

rraustad:
Re: Debugging techniques .. if you would have subtracted New from Original and plotted that on the Y2 axis I'll bet it would have shown where the diffs were. I do that often but didn't think of it here.

PR #8027, opened 2020-06-01
Fix latex issues via value added scripts

### LINKED ISSUE #8026: Misc. LaTeX Errors Not Caught in Normal Workflow
Issue overview
--------------
There are quite a few LaTeX errors in the LaTeX build logs that have not yet been fixed. Several are a bit cryptic to find and pull out of the build logs. Even though LaTeX calls these errors, the docs build anyway: sometimes the LaTeX processor is able to muddle through and do the right thing and at other times, there are actual glitches in the final PDF. These issues include duplicate labels tags, improper equation formatting, bad latex tags, improper citation references, etc.

### Checklist

michael-okeefe:
Pull request overview
---------------------
 - Fixes #8026 

### Purpose and Approach

Fixes Misc. LaTeX Errors Not Caught in Normal Workflow by adding a python-based LaTeX log parsing capability in Python that summarizes the relevant issues and warnings in the LaTeX log files by file and line number and also adds a JSON summary of the specific issues that can be further processed and/or displayed by the CI. This pull request also then fixes the majority of these issues across all the docs. These issues include package-based issues and warnings (for e.g., issues reported by \siunits) as well as issues and warnings from both the TeX and LaTeX processing pipeline.

The only new code added is the `doc/tools/parse_latex_log.py` file. That file uses no external dependencies beyond the python standard library. Fixes to the docs were found and fixed using the errors and warnings reported by the LaTeX log parser. The hope is that this script will make LaTeX errors and warnings more visible and actionable for future EnergyPlus documentation authors.

### Pull Request Author
 - [X] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer


doc/acknowledgments/src/notices-and-acknowledgments/other-software-included-or-referenced.tex
```diff
@@ -128,7 +128,7 @@ \subsection{MODELISAR Notice}\label{modelisar-notice}
 
 This file is licensed by the copyright holders under the BSD License\footnote{\url{http://www.opensource.org/licenses/bsd-license.html\textgreater}}:
 
-\begin{center}\rule{0.5\linewidth}{\linethickness}\end{center}
+\begin{center}\rule{0.5\linewidth}{0.4pt}\end{center}
```
mitchute, line 131: What's the purpose of this change?

doc/engineering-reference/src/climate-sky-and-solar-shading-calculations/climate-calculations.tex
```diff
@@ -111,10 +111,10 @@ \subsection{Sky Radiation Modeling}\label{sky-radiation-modeling}
 \noindent Example:
 \newline
 
-Clear sky ($N=0$), $T_{db} = 20 + 273.15 = \SI{293.15}{\kelivin}$, $T_{dp} = 10 + 273.15 = \SI{283.15}{\kelvin}$
```
mitchute, line 114: Ah ```kelivin```, almost Kevin's favorite number. https://www.urbandictionary.com/define.php?term=Keleven

doc/engineering-reference/src/climate-sky-and-solar-shading-calculations/climate-calculations.tex
```diff
@@ -111,10 +111,10 @@ \subsection{Sky Radiation Modeling}\label{sky-radiation-modeling}
...
+Clear sky ($N=0$), $T_{db} = 20 + 273.15 = \SI{293.15}{\kelvin}$, $T_{dp} = 10 + 273.15 = \SI{283.15}{\kelvin}$
 \newline
 
-$\epsilon_{sky} = \left(0.787 + 0.764  \ln\left(283.15/273\right)\right)  \left(1 + 0.0224  N - 0.0035  N^2 + 0.00028  N^3) = 0.815$
+$\epsilon_{sky} = \left(0.787 + 0.764  \ln\left(283.15/273\right)\right)  \left(1 + 0.0224  N - 0.0035  N^2 + 0.00028  N^3 \right) = 0.815$
```
mitchute, line 117: Nice to see these balanced parentheses being found.

doc/engineering-reference/src/daylighting-and-window-calculations/time-step-daylighting-calculation.tex
```diff
@@ -196,23 +196,23 @@ \subsection{Interior Illuminance}\label{interior-illuminance}
 The total illuminance at a reference point from all of the exterior windows in a zone is
 
 \begin{equation}
-  {I_{tot}}({i_L}) = \sum\limits_{\scriptstyle{\rm{windows}}\atop\scriptstyle{\rm{in~zone}}} {{I_{win}}({i_s},{i_L})}
+  {I_{tot}}({i_L}) = \sum\limits_{\genfrac{}{}{0pt}{2}{\rm{windows}}{\rm{in~zone}}} {{I_{win}}({i_s},{i_L})}
```
mitchute, line 199: Much better.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-001/coils.tex
```diff
@@ -1611,38 +1611,38 @@ \subsubsection{Special Calculations for Coil:Cooling:DX:TwoStageWithHumidityCont
 
 1. No latent control
 
-   Perform normal operation with SHR_{Normal}
+   Perform normal operation with SHR\(_{Normal}\)
```
mitchute, line 1614: The output of these changes look identical on my machine. What's the purpose?

doc/input-output-reference/src/hvac-template-objects/group-hvac-templates.tex
```diff
@@ -7557,7 +7557,7 @@ \subsubsection{Inputs}\label{inputs-25-000}
 
 The name of this object.
 
-\paragraph{Field: Chiller Object Type}\label{field-chiller-object-type}
```
mitchute, line 7560: Excellent. No need for unnecessary label references.

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
```diff
@@ -863,7 +863,7 @@ \subsubsection{Inputs}\label{inputs-8-011}
 
 \textbf{Idso} -- the clear sky emissivity are calculated using Idso model.
 
-\textbf{BerdahlMartin} -- the clear sky emissivity are calculated using Berdahl & Martin model.
+\textbf{BerdahlMartin} -- the clear sky emissivity are calculated using Berdahl \& Martin model.
```
mitchute, line 866: Good catch.

doc/input-output-reference/src/overview/group-simulation-parameters.tex
```diff
@@ -963,7 +963,7 @@ \subsection{PerformancePrecisionTradeoffs}\label{performanceprecisiontradeoffs}
 The Part Load Ratio (PLR) for a single speed coil or a multiple speed coil at speed 1 is calculated using the equation below
 
 \begin{equation}
-{\left {PLR}\right} = \frac{SystemLoad-OutputOff}{OutputFull - OutputOff}
+PLR = \frac{SystemLoad-OutputOff}{OutputFull - OutputOff}
```
mitchute, line 966: The output of this change looks identical. Regardless, this is a good cleanup.

mitchute (commented):
This is a nice set of changes that should help enable us to clean up documentation issues. 👍  👍

doc/acknowledgments/src/notices-and-acknowledgments/other-software-included-or-referenced.tex
michael-okeefe: This is a fix for a warning generated by LaTeX. `\linethickness` is not a valid option for the second argument to `\rule` but the LaTeX compiler typically muddles through. Just fixing in the name of removing warnings from the LaTeX compiler.

doc/engineering-reference/src/simulation-models-encyclopedic-reference-001/coils.tex
michael-okeefe: Another warning and, again, the LaTeX compiler is able to guess in this instance and muddle through. What's wrong is the `_{blah}` stuff is not valid outside of math mode. The `\(` and `\)` puts us in math mode (similar to `$`...`$` but allows checking for balancing).

doc/input-output-reference/src/overview/group-simulation-parameters.tex
michael-okeefe: Yeah, this is yet another instance where the LaTeX compiler issues a warning because the syntax is incorrect. If we really wanted to wrap PLR in "{}", we need to wrap it in `\left\{` and `\right\}`. However, I was assuming the wrapping of {} wasn't intentional. Cleaning up LaTeX issues where the compiler muddles through (so output doesn't change) but where the syntax as written is incorrect does really help with importing the LaTeX for the web-based documentation. Thanks!

Myoldmopar:
This looks great!  CI caught the warnings and parsed them out nicely.  However, we need to remove the other warnings that were already present before the CI improvements.  If you can clean those up, I think we'll be good to go.

Myoldmopar:
And to be clear, there are a couple options:
 - Make a commit that cleans out just the extra `LATEXWARNING::` stuff that is showing up, but leaves the LaTeX "mistakes" in place so that we demonstrate perfect CI behavior.  Then follow it up with a commit that gets the LaTeX code looking all clean.  Or...
 - Just do both of those at once and get a clean result.

I don't care either way.  Once both of those steps are done, however, mark this as ready for review and we'll get it in.

Thanks @nealkruis @michael-okeefe

nealkruis:
We can either pipe stdout into a txt file, or have an option to silence the python script output.

If I'm understanding things correctly, that would mean that we would see the three tests fail, but there wouldn't be any output below them. This would be followed by the new failure messages from parsing the *_errors.json files.

We could change it so that executing the python script is just another custom target or custom command in CMake instead of a CTest. That way you wouldn't see the extra failed tests.

Either way, once we get that sorted out, we'll clean up the actual LaTeX warnings/errors so this can go in cleanly.

Myoldmopar:
> We could change it so that executing the python script is just another custom target or custom command in CMake instead of a CTest. That way you wouldn't see the extra failed tests.

Yep, this would be ideal.  Please make it so, but leave the LaTeX warnings in place, and we'll make sure it all looks good, then do one clean commit and merge.

Thanks!

nealkruis:
Results are in! It definitely prevents other tests from being run successfully if python sends a bad return code.

@michael-okeefe let's set it up to always return zero so it doesn't prevent other tests from running.

@Myoldmopar since these aren't really Test Results, would it make sense to have the CI put them somewhere else (e.g., in "Build Messages" or in a separate section just for documentation)?

michael-okeefe:
I just modified the script so that it never returns a non-zero exit code and pushed that up (it was already on-par with the latest from develop). Shall I also start cleaning up the remaining LaTeX issues that were identified?

Myoldmopar:
This looks fantastic.  Once you address those LaTeX issues and we get a clean build we'll merge this in.  Thanks for the refinements here, @nealkruis @michael-okeefe !

Myoldmopar:
Oh, and about your comment on the build results/test results.  Unfortunately, since this is run as a post-processing ctest, I think it's too late for CI to grab these as "build results".  But they do show up fine as parsed test results.

CMakeLists.txt
```diff
@@ -103,8 +103,8 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "MinSizeRel" "RelWithDebInfo")
 endif()
 
-# we are making *a Python 3 Interpreter* a required dependency, so find it here
-find_package(PythonInterp 3 REQUIRED)
+# we are making *a Python 3.6 Interpreter* a required dependency, so find it here
+find_package(PythonInterp 3.6 REQUIRED)
```
Myoldmopar, line 107: Sometime in the future we need to move away from this PythonInterp as it was [deprecated](https://cmake.org/cmake/help/v3.12/module/FindPythonInterp.html) in 2018 in favor of `find_package(Python3)`, but first we need to move our minimum cmake up.

michael-okeefe:
> This looks fantastic. Once you address those LaTeX issues and we get a clean build we'll merge this in. Thanks for the refinements here, @nealkruis @michael-okeefe !

Sounds great. I just started on addressing the last of the LaTeX issues. Thanks!

CMakeLists.txt
```diff
@@ -337,6 +337,7 @@ if( BUILD_PACKAGE )
 endif()
 
 if (BUILD_DOCS)
+  set(DOCS_TESTING ${BUILD_TESTING} CACHE BOOL "" FORCE)
```
Myoldmopar: OK, so as soon as you enable both BUILD_DOCS and BUILD_TESTING, this will be forced to ON.  No problem.  Since this is automatically triggered to ON/OFF, I wonder if we should either mark it advanced, or just take it out of the cache entirely and treat it like a local cmake variable.

doc/CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ cmake_minimum_required(VERSION 3.5.1)
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
+if(NOT EXISTS ${PYTHON_EXECUTABLE})
+  # we are making *a Python 3.6 Interpreter* a required dependency, so find it here
+  find_package(PythonInterp 3.6 REQUIRED)
+endif()
```
Myoldmopar, line 10: It would be cool to just reuse the previously found Python, but I know that may be an annoying thing to manage, so it's OK.

doc/cmake/doc-targets.cmake
```diff
@@ -11,5 +11,14 @@ macro( CREATE_DOC_TARGET SOURCE_FILENAME OUTPUT_FILENAME )
...
+
   set_target_properties(zPDF_${OUTPUT_FILENAME} PROPERTIES FOLDER Documentation)
+
+  if (DOCS_TESTING)
+    message("Checking errors for ${OUTPUT_FILENAME}")
+    add_custom_command(TARGET zPDF_${OUTPUT_FILENAME}
+      POST_BUILD
+      COMMAND ${PYTHON_EXECUTABLE} "${PROJECT_SOURCE_DIR}/tools/parse_latex_log.py" "${PROJECT_SOURCE_DIR}/${SOURCE_FILENAME}/${SOURCE_FILENAME}.log" "${PROJECT_SOURCE_DIR}/${SOURCE_FILENAME}" "${PROJECT_BINARY_DIR}/${OUTPUT_FILENAME}_errors.json"
+      )
+  endif()
```
Myoldmopar, line 22: OK, so if this is just a POST_BUILD on the PDF building portion, then the file should be available once "make" is complete.  In this case, I think I actually could have CI find the error files early enough to catch these as build errors, not test failures.  Let me know if I am correct in the build step and I can make an alteration to CI.

doc/tools/parse_latex_log.py
```diff
@@ -0,0 +1,552 @@
...
+                'external-interfaces-application-guide',
+                'getting-started',
+                'input-output-reference',
+                'interface-developer',
+                'module-developer',
+                'output-details-and-examples',
+                'plant-application-guide',
+                'tips-and-tricks-using-energyplus',
+                'using-energyplus-for-compliance',
+                ]
```
Myoldmopar: Is this list really there just for testing?  Or is this something we have to maintain?

Myoldmopar (approved):
This looks pretty great as-is, though a couple thoughts came to mind that could be considered:
 - Do we want to make DOCS_TESTING just an internal variable, nothing exposed to the cache?
 - If the doc test is run as a post build step, I would be happy to tell CI to search for the `errors.json` files after the build rather than after the test.

doc/tools/parse_latex_log.py
michael-okeefe: Yeah, I was debating about this and ended up leaving it in as I found it useful but you are right -- it could become a maintenance issue if anyone else were to depend on it. On a second look, I'd vote to just remove it.

Myoldmopar: :+1:

michael-okeefe: I've fixed this and the remaining LaTeX issues that were flagged. Just pushed that up.

I also added a check in the python script to delete the target JSON file if it exists before writing as a persistent JSON error file would be confusing for those working off of their local laptops and actively fixing issues.

Myoldmopar:
This looks great!  I'm not sure why Mac hasn't reported yet, but Mac wasn't involved, so I'm OK with not waiting on Mac for this.  The only comment I had unresolved was whether we should just eliminate the DOCS_TESTING cache variable since it is controlled via the two cache variables BUILD_DOCS and BUILD_TESTING.  This doesn't have to be held up for that, but I'd like to know what you think of that change at least.

nealkruis:
We'll still need DOCS_TESTING for the documentation CMake project (if it's created independently). I'll make it an internal variable in the EnergyPlus CMake project.

Myoldmopar:
I don't see how this could've broken that one failing Windows file.  I'm feeling OK with just letting this merge as-is.  I've got a [branch of Decent CI](https://github.com/Myoldmopar/decent_ci/pull/42) that moves the "test errors" into "build errors" which is much more appropriate here. I won't worry about merging that immediately but I'll try some time soon.

PR #8028, opened 2020-06-01
Protect ScanPlantLoopsForObject call with member var flag

Myoldmopar:
Pull request overview
---------------------
Fixes a problem with the water use connection code, where it would call ScanPlantLoopsForObject many, many, many times.  Now it is just once per object instance.  CI should see zero diffs or issues.

Runtime won't be affected for most files, but some larger files that include plant connected water use objects will see a big impact.  Consider `HospitalBaselineReheatReportEMS.idf`:
- Before: 05min 23.67sec
- After: 04min 16.67sec
- Impact: 20.7% runtime reduction

FYI @amirroth @hongtz68 @xuanluo113


src/EnergyPlus/WaterUse.cc
```diff
@@ -992,7 +992,7 @@ namespace WaterUse {
             this->setupMyOutputVars = false;
         }
 
-        if (allocated(DataPlant::PlantLoop) && !this->StandAlone) {
+        if (this->plantScanFlag && allocated(DataPlant::PlantLoop) && !this->StandAlone) {
```
Myoldmopar, line 995: I verified this loop was getting hit often by adding a print statement and running an annual simulation.  The print output came out zillions of times (approximately).  I simply added this flag so that this call will only be made once per instance of a water use connection.

src/EnergyPlus/WaterUse.cc
```diff
@@ -1009,6 +1009,7 @@ namespace WaterUse {
             if (errFlag) {
                 ShowFatalError("InitConnections: Program terminated due to previous condition(s).");
             }
+            this->plantScanFlag = false;
```
Myoldmopar, line 1012: And of course reset it to false so it won't be hit again.

src/EnergyPlus/WaterUse.hh
```diff
@@ -221,7 +222,7 @@ namespace WaterUse {
               ColdSupplyTemp(0.0), ColdTemp(0.0), HotTemp(0.0), DrainTemp(0.0), RecoveryTemp(0.0), ReturnTemp(0.0), WasteTemp(0.0), TempError(0.0),
               MainsVolume(0.0), TankVolume(0.0), ColdVolume(0.0), HotVolume(0.0), TotalVolume(0.0), Power(0.0), Energy(0.0), NumWaterEquipment(0),
               MaxIterationsErrorIndex(0), PlantLoopNum(0), PlantLoopSide(0), PlantLoopBranchNum(0), PlantLoopCompNum(0), MyEnvrnFlag(true),
-              setupMyOutputVars(true)
+              setupMyOutputVars(true), plantScanFlag(true)
```
Myoldmopar, line 225: The flag needed to be added as a member variable.  I could have potentially hijacked the `setupMyOutputVars` flag, but then it would get sketchy as that flag would be doing more than just setting up output variables.  Keeping them separate is good for this use case.

Myoldmopar (commented):
I'm trying to think of a good unit test for this change.  Really, a test would just ensure that the Scan function wasn't being called twice.  We aren't really set up for that type of test here, I mean in my Jasmine JavaScript testing, I'd just:

```
describe("ScanPlantLoops called by WaterUse", function() {
  beforeEach(function() {
    spyOn(plantUtilities, 'scanPlantLoops');
  });
  it("tracks the number of times scanPlant was called", function() {
    expect(plantUtilities.scanPlantLoops.calls.count()).toEqual(0);
    initWaterUse();
    initWaterUse();
    expect(plantUtilities.scanPlantLoops.calls.count()).toEqual(2);
  });
});
```

-ish.  For now, I am open to suggestion, or I can defer a unit test until a later PR where I refactor this scanPlantLoops pattern to a single location...or something.

xuanluo113:
@Myoldmopar I have less knowledge of this function, but based I what I read just now, I think it's trying to find the `LoopNum`, `BranchNum`, etc, given the name string of a plant component. These numbers should be settled once scanned, and should not change during run time. So for each component, their `LoopNum` could be initialized with 0 or a null pointer, and set once by calling this function. Is that right? 

So I just added an assertion of `LoopNum = 0 or null` inside this function and ran the medium office, and found PERHAPS [here](https://github.com/NREL/EnergyPlus/blob/e858cb2b02e0e681bc2492fbe3938c1e5d03a0d1/src/EnergyPlus/WaterThermalTanks.cc#L5587) it's also called more than it wants. I added some prints at the end of the `ScanPlantLoopsForObject` function and these lines are replicating millions of times trying to rematch the water tank component but getting the same results:

```
CompName: CORE_BOTTOM WATER EQUIPMENT, LoopNum: 1, LoopSideNum: 1, BranchNum: 2, CompNum: 1
CompName: CORE_MID WATER EQUIPMENT, LoopNum: 1, LoopSideNum: 1, BranchNum: 3, CompNum: 1
CompName: CORE_TOP WATER EQUIPMENT, LoopNum: 1, LoopSideNum: 1, BranchNum: 4, CompNum: 1
```

Not sure if my assumption is right, and currently I cannot think of any unit test-like way other than this assertion right now. May need more time investigating this.

Myoldmopar:
@xuanluo113 which file did you run to get the repeated outputs in the water tank?  I've run a few tank files and can't reproduce.

Myoldmopar:
I was able to find a file that had it repeating, but it's not actually repeating too much.  `5ZoneVAV-ChilledWaterStorage-Stratified` shows the code calling scan plant loops a couple dozen times, which is definitely too many, but I'm not sure it's the biggest offender.  It looks like the reason it keeps doing it is because that block of code is doing a couple different things.  It is calling scan plant loops to find the location, and also calling interconnect plant loops.  This is stuff that should be done just one time per connection.  But then it is also doing sizing.  Because the sizing is being done in that same block, the flag is only set to false once plant sizing is locked in.  This should be split into two different blocks.  One block to do the one-time topology stuff, then a second block to do the sizing multiple times.  I will try to get a quick fix in on that, but I'm not going to hold this up if it gets more complex.  Currently this PR fixes what seems to be the most egregious scan plant loops offender.

xuanluo113:
> @xuanluo113 which file did you run to get the repeated outputs in the water tank? I've run a few tank files and can't reproduce.

I was using RefBldgMediumOffice.

Myoldmopar:
Thanks @xuanluo113 I was able to protect the scan plant loops calls from being hit excessively in the tanks by adding a second flag to the tank structure.  It could certainly be cleaned up further, but with this commit I am seeing no diffs in RefBldgMediumOffice, so it should be sufficient for now.

Reviewers: I don't think I'll let the scope of this creep further.  If CI is happy here, then this can go in.

xuanluo113 (approved):

Myoldmopar:
Looks good to me :smile: Merging.

PR #7980, opened 2020-05-08
Fix Controller:WaterCoil chilled water temperature and humidity control

mjwitte:
Pull request overview
---------------------
 - Fixes #7653
 - Refine the humidity control reset for Controller:WaterCoil with control variable TemperatureAndHumidityRatio.
 - Reinstate (and fix) the Rootfinder `CheckIncrementRoundOff` function which checks to see if the next X increment is larger than the X tolerance and ends iterations if not.
 - Expand Rootfinder `CheckIncrementRoundOff` function to check for change in Y value < tolerance.
 - Remove call to `UpdateController` when `CheckBracketRoundOff` returns true.

Diffs - CI Tests
-------------
*Italics indicate potential problem files*
| File | T&HumRat? | eso/mtr Diffs |Error Diffs | 
|:---|:---:|:---:|:---|
| 5ZoneCoolingPanelBaseboard  | Y | Big | |
| 5ZoneCoolingPanelBaseboardAuto | Y | Big | |
| 5ZoneCoolingPanelBaseboardTotalLoad | Y | Big | |
| 5ZoneCoolingPanelBaseboardVarOff | Y | Big | |
| 5ZoneElectricBaseboard | Y | Big | |
| 5ZoneWaterCooled_Baseboard | Y | Big | |
| 5ZoneWaterCooled_BaseboardScalableSizing | Y | Big | |
| 5ZoneWaterCooled_GasFiredSteamHumidifier | Y | Big | |
| 5ZoneWaterCooled_HighRHControl | Y | Big | |
| 5ZoneWaterCooled_MultizoneAverageRHControl | Y | Big | |
| 5ZoneWaterCooled_MultizoneMinMaxRHControl | Y | Big | |
| DOAS_wNeutralSupplyAir_wFanCoilUnits | Y | Big | |
| EMPD5ZoneWaterCooled_HighRHControl | Y | Big | |
| HeatPumpWaterHeater | Y | Big | |
| *HeatPumpWaterHeaterStratified* | Y | Big | New warning - SolveAirLoopControllers: Maximum iterations (50) exceeded | 
| VSHeatPumpWaterHeater | Y | Big | |
| DOASDualDuctSchool | N | Small | |
| HospitalBaseline | Y | Small | Old warnings gone! - SolveAirLoopControllers: Maximum iterations (50) exceeded |
| HospitalBaselineReheatReportEMS | Y | Small | Old warnings gone! - SolveAirLoopControllers: Maximum iterations (50) exceeded |
[...]

PR #8046, opened 2020-06-04
Quick attempts to fix integration failures

Myoldmopar:
In a recent commit, I found not one, but two files timing out in the linux integration debug build.  These days about half the time the files all pass, but that's not nearly enough.  We need all these files to pass fully so that something real doesn't slip through the cracks.

HospitalLowEnergy definitely needs some love, the HVAC system has lots of iteration problems, etc.  But in the meantime, I just put the timestep from 10 minutes to 15 minutes, and as expected, it cut the runtime down, and it should start passing here.  The file has "big" diffs, but the meter diffs were actually on the order of small fractions of a percent change.  If someone wants to really dig in to that file and get it working right at the 10 minute timestep level, that's fine, but in the meantime, we need to get it at least running consistently on CI.  Otherwise we might as well just not test it.

A python plugin file timed out, and with a quick cleanup, I found an initialization flag was never being set in the Python code, and so the code was looking for handles to actuators and variables every - single - time it ran.  Many, many, string searches going on.  Once I changed that the file runtime went from 42 seconds to about 2.75 seconds, on a local release build.

I'm open to discussion on the HospitalLowEnergy change, but if no one has the time to dig in to the shenanigans in that file, we should at least just use this slight change to the time step to get it running reliably.

### DIFF
--- a/testfiles/PythonPluginPlantOperation_largeOff.idf
+++ b/testfiles/PythonPluginPlantOperation_largeOff.idf
@@ -7708,116 +7708,30 @@
     Boiler:HotWater,         !- Equipment 1 Object Type
     HeatSys1 Boiler;         !- Equipment 1 Name
 
-!  EnergyManagementSystem:ProgramCallingManager,
-!    Heating_dispatch,        !- Name
-!    UserDefinedComponentModel,  !- EnergyPlus Model Calling Point
-!    Heating_dispatch_Values; !- Program Name 1
-
   PythonPlugin:Instance,
     Heating_dispatch,                  !- Name
     Yes,                                    !- Run During Warmup Days
     PythonPluginPlantOperation_largeOff,    !- Python Module Name
-    Heating_dispatch_Values;           !- Plugin Class Name
-
-!  EnergyManagementSystem:Program,
-!    Heating_dispatch_Values, !- Name
-!    IF HeatSys1_LoopDmnd > 0.0,  !- Program Line 1
-!    SET Boiler_Disptch = HeatSys1_LoopDmnd,  !- Program Line 2
-!    ELSE,                    !- <none>
-!    SET Boiler_Disptch = 0.0,!- <none>
-!    ENDIF;                   !- <none>
-
-!  EnergyManagementSystem:InternalVariable,
-!    HeatSys1_LoopDmnd,       !- Name
-!    HeatSys1 Operation Scheme,  !- Internal Data Index Key Name
-!    Supply Side Current Demand Rate;  !- Internal Data Type
-
-!  EnergyManagementSystem:Actuator,
-!    Boiler_Disptch,          !- Name
-!    HeatSys1 Operation Scheme:HeatSys1 Boiler,  !- Actuated Component Unique Name
-!    Plant Equipment Operation,  !- Actuated Component Type
-!    Distributed Load Rate;   !- Actuated Component Control Type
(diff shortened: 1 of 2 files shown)


mjwitte:
@Myoldmopar I have no problem with the timestep change for HospitalLowEnergy. #7980 appears to improve that file somewhat, showing an integration coverage runtime of 4300s. So, the combination of this change and that may put is well into the safe zone. What I find puzzling is that HospitalLowEnergy isn't really super-slow in a release build. I wonder if there's a certain aspect that slows down more than others in debug builds?

Myoldmopar:
Thanks for the comments @mjwitte, that is good to know.

I'll do a quick dive into the diffs.

- EIO Diffs
  - CTF values are obviously different for the different timestep.
  - Zone sizing values changed somewhat to account for the different peak moments:
    - \- Zone Sizing Information, FLOOR 1 CAFE, Cooling, 35767.77606
    - \+ Zone Sizing Information, FLOOR 1 CAFE, Cooling, 35162.07367
    - \- Zone Sizing Information, FLOOR 1 CAFE, Heating, 1071.72886,
    - \+ Zone Sizing Information, FLOOR 1 CAFE, Heating, 1091.77578
  - Some minor component sizing changes:
    - \- Component Sizing Information, COIL:COOLING:WATERTOAIRHEATPUMP:EQUATIONFIT, --, Design Size Rated Total Cooling Capacity [W], 4681.67973
    - \+ Component Sizing Information, COIL:COOLING:WATERTOAIRHEATPUMP:EQUATIONFIT, --, Design Size Rated Total Cooling Capacity [W], 4682.95429
    - \- Component Sizing Information, COIL:COOLING:WATERTOAIRHEATPUMP:EQUATIONFIT, --, Design Size Rated Water Flow Rate [m3/s], 2.24994E-004
    - \+ Component Sizing Information, COIL:COOLING:WATERTOAIRHEATPUMP:EQUATIONFIT, --, Design Size Rated Water Flow Rate [m3/s], 2.25055E-004
  - Warmup convergence diffs, but nothing apparently impactful
- ERR Diffs
  - Timing and number of warnings is scrambled, but nothing seems wrong
- MTR Big Diffs
  - The absolute values seem high at first, but then looking at them in relative terms they are reasonably small:
image
- ESO Diffs
  - One of the biggest offenders appears to be: `FLOOR 1 LOUNGE HEAT PUMP INLET NODE:System Node Standard Density Volume Flow Rate [m3/s](Hourly)`.  The diff plots seem to reveal continual diffs, but plotting the two results together show it is behaving quite similarly:
image

This all seems quite within reason for a time step change.  If anyone has issues, I'm happy to discuss further, and if not, then can go in and start getting integration back to 100% consistently in develop.

Myoldmopar:
HospitalLowEnergy is now at 3595.64, down from hovering around the 4800 CI limit.

mjwitte (approved):
@Myoldmopar Nice. PythonPluginPlantOperation_largeOff.idf shows 168s in the integration coverage test instead of >4000s. I'm good with these changes. Merge will occur in 30 minutes - speak now or forever hold your peace. This is an instantaneous launch window.

Myoldmopar:
While there is some weather in the area, it will not affect today's launch.  First stage propulsion reports GO.

image

mjwitte:
5 . . . 4 . . . 3 . . . 2 . . . 1 . . . 0 . . . we have liftoff to merge . . .

Issue #6976, opened 2018-09-20
Clean up daylighting output reporting for more than two reference points


JasonGlazer:
While working on #6889 found that in the DaylightingManager ::CalcDayltgCoefficients() routine, saw that some output reports seems to be tied to the number of daylighting reference points. These need to be cleaned up after the release goes out since they impact output report format. This code is used when writing to OutputFileInits and to OutputFileDFS.

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

PR #8063, opened 2020-06-08
Fix DaylightManager Unit Test on Some Machines

mitchute:
Fixes failed unit tests on my machine caused by #6976. CI didn't catch this, so we should probably figure that out at some point.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/DaylightingManager.cc
+++ b/src/EnergyPlus/DaylightingManager.cc
@@ -172,6 +172,8 @@ namespace DaylightingManager {
     bool DayltgInteriorMapIllum_FirstTimeFlag(true);
     bool ReportIllumMap_firstTime(true);
     bool SQFirstTime(true);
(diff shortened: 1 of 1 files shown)


Myoldmopar:
@mitchute the reason CI didn't find this is because on CI, the tests are run as `ctest -j N`, so each test has its own memory and these statics are initialized "properly" each time.  When you run them serially using `energyplus_tests`, the memory space remains the same, so these statics stay "as they were".  We need to do a complete overhaul of the code and identify all non-const static local variables and remove them immediately.  Or soon, anyway. :smile:  That will take care of so many of these problems.

mitchute:
> We need to do a complete overhaul of the code and identify all non-const static local variables and remove them immediately.

Or we could just wait until they fail on my machine. 😄

Myoldmopar:
2 statics gone...woot!  Thanks @mitchute.

PR #8069, opened 2020-06-10
Plugin Manager Cleanups

Myoldmopar:
While testing out the recent fix for Mac symlink runs, I found a couple issues that need to be addressed:
- Transition wouldn't work because the IDD numbering was off
- The 9.3 IDD hadn't been added either
- I encountered a problem running a plugin file and I think I found why, but I need to test.
- Did some general cleanups in the plugin manager as well

At some point I'll kick off a CI package build from this branch so I can test out CI built packages.


cmake/Install.cmake
```diff
@@ -300,8 +300,8 @@ endif()
 set(CPACK_RESOURCE_FILE_README "${PROJECT_SOURCE_DIR}/release/readme.html")
 
 install(FILES "${PROJECT_SOURCE_DIR}/bin/CurveFitTools/IceStorageCurveFitTool.xlsm" DESTINATION "PreProcess/HVACCurveFitTool/")
-install(FILES "${PROJECT_SOURCE_DIR}/idd/V9-2-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/")
-install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/" RENAME "V9-3-0-Energy+.idd" )
+install(FILES "${PROJECT_SOURCE_DIR}/idd/V9-3-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/")
+install(FILES "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater/" RENAME "V9-4-0-Energy+.idd" )
```
Myoldmopar, line 304: These were missed in the version update PR (#7921)

idd/CMakeLists.txt
```diff
@@ -12,3 +12,4 @@ install( FILES "V8-9-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater" )
 install( FILES "V9-0-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater" )
 install( FILES "V9-1-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater" )
 install( FILES "V9-2-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater" )
+install( FILES "V9-3-0-Energy+.idd" DESTINATION "PreProcess/IDFVersionUpdater" )
```
Myoldmopar, line 15: And this was missed in #7921

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -381,7 +381,7 @@ namespace HeatBalanceManager {
         ManageEMS(emsCallFromEndZoneTimestepAfterZoneReporting, anyRan); // EMS calling point
 
         UpdateEMSTrendVariables();
-        EnergyPlus::PluginManagement::pluginManager->updatePluginValues();
+        EnergyPlus::PluginManagement::PluginManager::updatePluginValues();
```
Myoldmopar, line 384: Went through and made some things static that weren't needed as member methods.

src/EnergyPlus/PluginManager.cc
```diff
@@ -777,7 +777,7 @@ namespace PluginManagement {
     {
 #if LINK_WITH_PYTHON == 1
         // we'll need the program directory for a few things so get it once here at the top and sanitize it
-        std::string programPath = FileSystem::getProgramPath();
+        std::string programPath = FileSystem::getAbsolutePath(FileSystem::getProgramPath());
```
Myoldmopar, line 780: This is necessary if you are running a python plugin file from a symlinked energyplus and you are using python plugins that live right next to the E+ binary itself.  Kind of a corner case, but this should address it.

src/EnergyPlus/PluginManager.hh
```diff
@@ -153,16 +153,16 @@ namespace PluginManagement {
...
+        static Real64 getTrendVariableValue(int handle, int timeIndex);
         static size_t getTrendVariableHistorySize(int handle);
-        Real64 getTrendVariableAverage(int handle, int count);
-        Real64 getTrendVariableMin(int handle, int count);
-        Real64 getTrendVariableMax(int handle, int count);
-        Real64 getTrendVariableSum(int handle, int count);
-        Real64 getTrendVariableDirection(int handle, int count);
+        static Real64 getTrendVariableAverage(int handle, int count);
+        static Real64 getTrendVariableMin(int handle, int count);
+ static        Real64 getTrendVariableMax(int handle, int count);
```
Myoldmopar: I'm not sure why this wasn't indented...

Myoldmopar:
It actually looks like all three of my sanity tests were failing purely because of the transition problem.  I'll make a package build from this branch as-is and re-run them, then decide if I need to do anything else.

Myoldmopar:
I got them all working [really nicely](https://travis-ci.com/github/Myoldmopar/EPTravisTester/builds/170865301).  However, I had to skip symlink testing on Windows.  I didn't realize symlinks were such a special privilege on Windows that programs had to be granted the permission to make them.  I don't think the user account on the Travis instance has that ability, so the symlink runs just quietly fail with a file-not-found error.  I am not going to sweat it, symlink usage on Windows is a different beast.  I think we are good to go here with these cleanups.  Anyone can speak up and request changes or offer thoughts if they want, otherwise this will drop later.

Myoldmopar:
I pulled in develop and fixed the indentation problem.  I anticipate merging this in later today unless anyone objects.

Myoldmopar:
Cannot imagine how the indentation change would've broken that one Windows file.  Calling this done and merging.

PR #8073, opened 2020-06-10
Prevent Kiva exposed perimeter warning

shorowit:
Pull request overview
---------------------
IDF files using Kiva can potentially generate warnings like this:
```
   ** Warning ** SurfaceProperty:ExposedFoundationPerimeter: SLAB, Total Exposed Perimeter is greater than the perimeter of SLAB
   **   ~~~   ** SLAB perimeter = 51, SurfaceProperty:ExposedFoundationPerimeter exposed perimeter = 51
   **   ~~~   ** Total Exposed Perimeter will be set equal to SLAB perimeter
```
In the IDF file provided below, the perimeter of both the `SurfaceProperty:ExposedFoundationPerimeter` and `BuildingSurface:Detailed` objects are exactly identical (51.2064 m). Presumably the issue is caused by floating point precision when calculating the perimeter of the `BuildingSurface:Detailed` object from its vertices. This PR adds tolerance to the warning.

Test file:
[in.idf.txt](https://github.com/NREL/EnergyPlus/files/4761347/in.idf.txt)

### Pull Request Author

### Reviewer


nealkruis:
This looks good, but you might consider using [`Kiva::isGreaterThan`](https://github.com/NREL/EnergyPlus/blob/develop/third_party/kiva-ep/src/libkiva/Functions.cpp#L32) (or something similar in EnergyPlus--@Myoldmopar do you know of anything?) for a consistent approach to handling floating point comparisons.

Myoldmopar:
> -@Myoldmopar do you know of anything?)

I would at a minimum suggest declaring the applied tolerance as a const float in the function, or some convenient location, and check if the value is within the tolerance both above as well as below.  I assume there are a couple functions like that already in EnergyPlus, but I don't know of any off the top of my head.  Plant has an [integer based one](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/PlantUtilities.hh#L176) that just checks a range, not exactly what we're looking for here.

shorowit:
Ok, I added a const for the tolerance. Note that we're not trying to ensure the value is 1 (within some tolerance) but rather that it does not _exceed_ 1 (including a tolerance).

Myoldmopar:
Change makes sense, I'm not sure why CI had trouble.  Do you anticipate appending a unit test to this fix?

shorowit:
I wasn't planning on creating a unit test given it's a minor warning bugfix. There is an IDF in the linked GitHub issue that can be used to demonstrate the fix.

Myoldmopar (approved):

Myoldmopar:
This is totally fine.  I verified the fix locally.  I'm not sure why CI is struggling with this minimal commit.  I'm hesitant to merge.  I'll just push one more commit merging develop in and we'll see if it stalls again.

Myoldmopar:
All good, merging.  Thanks @shorowit

PR #8074, opened 2020-06-11
Fix Unit Tests to Run Serially

mitchute:
Fixes some unit tests that don't pass when tests executed serially.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/DataWater.cc
+++ b/src/EnergyPlus/DataWater.cc
@@ -122,6 +122,11 @@ namespace DataWater {
     bool WaterSystemGetInputCalled(false); // set true once input data gotten.
     bool AnyIrrigationInModel(false);      // control flag set true if irrigation input for ecoroof DJS PSU Dec 2006
 
+    void clear_state()
+    {
+        WaterSystemGetInputCalled = false;
+    }
+
     // Object Data
     SiteRainFallDataStruct
         RainFall; // type of rainfall modeling | design annual rain | rain sched id | nominal annual rain | current rate | current amount
--- a/src/EnergyPlus/DataWater.cc
+++ b/src/EnergyPlus/DataWater.cc
@@ -117,15 +117,24 @@ namespace DataWater {
     int NumRainCollectors(0);    // number of rainfall collectors in model
     int NumGroundWaterWells(0);  // number of
     int NumSiteRainFall(0);
-    int NumIrrigation(0);                  // DJS PSU Dec 2006 number of irrigation descriptions (1 allowed)
     bool AnyWaterSystemsInModel(false);    // control flag set true if any water systems
     bool WaterSystemGetInputCalled(false); // set true once input data gotten.
     bool AnyIrrigationInModel(false);      // control flag set true if irrigation input for ecoroof DJS PSU Dec 2006
 
     void clear_state()
     {
+        NumWaterStorageTanks = 0;
+        NumRainCollectors = 0;
+        NumGroundWaterWells = 0;
+        NumSiteRainFall = 0;
+        AnyWaterSystemsInModel = false;
         WaterSystemGetInputCalled = false;
+        AnyIrrigationInModel = false;
         RainFall = {};
+        Irrigation = {};
+        WaterStorage.deallocate();
+        RainCollector.deallocate();
+        GroundwaterWell.deallocate();
     }
 
     // Object Data
--- a/src/EnergyPlus/DataWater.cc
+++ b/src/EnergyPlus/DataWater.cc
@@ -125,6 +125,7 @@ namespace DataWater {
     void clear_state()
     {
         WaterSystemGetInputCalled = false;
+        RainFall = {};
     }
 
     // Object Data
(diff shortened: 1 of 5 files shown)


src/EnergyPlus/DataWater.cc
```diff
@@ -122,6 +122,11 @@ namespace DataWater {
     bool WaterSystemGetInputCalled(false); // set true once input data gotten.
     bool AnyIrrigationInModel(false);      // control flag set true if irrigation input for ecoroof DJS PSU Dec 2006
 
+    void clear_state()
+    {
+        WaterSystemGetInputCalled = false;
```
mjwitte, line 131: This is a good start, but the clear_state should reset all of the non-const variables above.

mjwitte (commented):
Testing here. Confirmed that develop fails. 

We should be catching these during review (there *is* a checklist item for "Run Unit Test(s) locally". Maybe we should add "serially".) It can take a while for them all to run, but I exclude the property data tests 'cause they're *reallly* slow.
`energyplus_tests.exe --gtest_filter=-*PropertiesRefData*`

mjwitte:
Also, running all tests on this branch, I got this array bounds failure:
```
EnergyPlusFixture.WindowEquivalentLayer_VBMaximizeBeamSolar
Assertion failed: contains( i ), file C:\MJW Git\EnergyPlus\third_party\ObjexxFCL\src\ObjexxFCL/Array1.hh, line 1172
```

mitchute:
@mjwitte @Myoldmopar this now passes locally in debug and release modes, but I see yours is still failing. I'll take a look. And yes, we should be adding all of the non-const variables to clear_states. The globals refactor work will eventually get this done, but I'll add the ones in the files affected here.

E+ Dev Team: if anyone sees non-consts that should be added to clear_states, feel free to add them. Some namespaces (like in this fix) didn't even have a clear_state, so if you find that feel free to add one. Last year I was playing around with the clang AST matcher tools and found something like 1200 ```static``` variables, and 13,500 variables with global scope. The summary is here, though, it's quickly becoming out of date: https://docs.google.com/spreadsheets/d/1wSEx6BbjdW3IhxOqb_SLOg29W04QY9Lf99eh5RH0GBc/edit?usp=sharing As you work, feel free to eliminate as many of these as you can.

mitchute:
> Hmm, will that do the same as `RainFaill.deallocate();` as other `clear_states` do? I was expecting to see `NumWaterStorageTanks = 0;` etc. here as well.

```RainFall``` is a just an struct instance, not an array or vector, so we can't use ```deallocate```. The ```MyStructInst = {};``` seems to be an accepted C++ 11 way of resetting a class instance (https://stackoverflow.com/a/15183438)

mitchute:
> Also, running all tests on this branch, I got this array bounds failure:
> 
> ```
> EnergyPlusFixture.WindowEquivalentLayer_VBMaximizeBeamSolar
> Assertion failed: contains( i ), file C:\MJW Git\EnergyPlus\third_party\ObjexxFCL\src\ObjexxFCL/Array1.hh, line 1172
> ```

@mjwitte did you run this with the latest commit (c0f4b78b25cf4be2a274c88dec632960d9ea2a3c). I was hitting that error also before this latest commit.

mjwitte:
Hmm, maybe not. I'll pull again and re-run.
Regarding `.deallocate()` I was moving too quickly.  The `Array1D`s below the struct should be deallocated, like `WaterStorage` etc.?

mitchute:
Yes, they should. I just wanted to make sure to identify the ones that were causing failures before doing "extra".

mjwitte:
Ok, sorry to jump the gun on reviewing. Holler when you're done.

mjwitte:
@Myoldmopar If a PR is labeled `DoNotPublish` and `Defect` will it show up in the changelog?

Myoldmopar:
> @Myoldmopar If a PR is labeled `DoNotPublish` and `Defect` will it show up in the changelog?

It will [not](https://github.com/NREL/EnergyPlus/blob/develop/doc/tools/create_changelog.py#L127-L130).  Only NewFeature, Defect, Performance, and Refactoring are dropped in the changelog.  The purpose of DoNotPublish is just to explicitly enforce that all PRs should be labeled with "something".

mjwitte:
I'm confused, it looks like a PR with both labels `Defect` and `DoNotPublish` (like this one) *will* show up in the changelog. So, if you don't want something in the changelog, it should just have a single `DoNotPublish` label.

Myoldmopar:
Great catch!  Yes, I was forgetting about PRs with multiple labels.  In a DoNotPublish case, you shouldn't put on another label, or it will be published.  I can change that behavior so that DoNotPublish overrides other PRs for the changelog.  I'll make that change soon, that way PRs can still use the other labels to reflect their intent.

mitchute:
I'm done here unless something else breaks. Feel free to review/test.

mjwitte (approved):
All good here. Will merge once CI completes.

mjwitte:
So, do we attribute this [Mac failure](https://nrel.github.io/EnergyPlusBuildResults/EnergyPlus-174b39b1710514eff5c03139c6be129e615ac415-x86_64-MacOS-10.13-clang.html) to Mac randomness, or does it need further investigation? This seems to be a different failure that what we've been getting from Mac CI recently.

mitchute:
I just ran the tests 3 times locally (on my Mac) and didn't see any troubles. I suppose I could clean up a few more variables and see is CI still complains. Thoughts?

Myoldmopar:
I just found that that same error was in other branches, so I think this is OK...

PR #8076, opened 2020-06-12
Global CondenserLoopTowers

mitchute:
### Pull Request Author

### Reviewer


mitchute:
Ready for review.

mitchute:
Minor diffs expected due to this commit (2374a10054250648a09b38618f9750858a794b67) moving one last module-level variable over to the object's member variable.

src/EnergyPlus/ChillerIndirectAbsorption.cc
```diff
@@ -106,7 +106,7 @@ namespace ChillerIndirectAbsorption {
     // OTHER NOTES:
     // Manufacturers performance data can be used to generate the coefficients for the model.
 
-    auto constexpr calcChillerAbsorptionIndirect("CALC Chiller:Absorption:Indirect ");
+    const char * calcChillerAbsorptionIndirect("CALC Chiller:Absorption:Indirect ");
```
mitchute, line 109: I think this is preferred, though it doesn't always work.

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -116,12 +116,6 @@ namespace ChillerReformulatedEIR {
     // 1. Hydeman, M., P. Sreedharan, N. Webb, and S. Blanc. 2002. "Development and Testing of a Reformulated
     //    Regression-Based Electric Chiller Model". ASHRAE Transactions, HI-02-18-2, Vol 108, Part 2, pp. 1118-1127.
 
-    // chiller part load curve types
-    int const PLR_LeavingCondenserWaterTemperature(1); // Type 1_LeavingCondenserWaterTemperature
-    int const PLR_Lift(2);                             // Type 2_Lift
```
mitchute, line 121: I missed moving this to an ```enum``` for the last globals refactor. It's been moved now.

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -116,12 +116,6 @@ namespace ChillerReformulatedEIR {
...
-
-    Real64 condOutletTempMod(0.0);
```
mitchute, line 123: One last module-level variable removed. Now the objects are using their own member variable, as they should. This causes a few diffs as indicated in a previous comment.

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -105,59 +107,20 @@ namespace CondenserLoopTowers {
...
-    int const EvapLossByUserFactor(80);
-    int const EvapLossByMoistTheory(81);
-
-    int const BlowdownByConcentration(90);
-    int const BlowdownBySchedule(91);
-
     std::string const cCoolingTower_SingleSpeed("CoolingTower:SingleSpeed");
     std::string const cCoolingTower_TwoSpeed("CoolingTower:TwoSpeed");
     std::string const cCoolingTower_VariableSpeed("CoolingTower:VariableSpeed");
     std::string const cCoolingTower_VariableSpeedMerkel("CoolingTower:VariableSpeed:Merkel");
```
mitchute, line 112: It's complaining when I try to move this to ```const char *```. For example:

> ShowSevereError(cCoolingTower_SingleSpeed + " \\"" + this->Name + "\\". Free convection air flow rate must be less than the design air flow rate.")

```Invalid operands to binary expression ('const char *' and 'const char [3]')```

Any suggestions?

File: src/EnergyPlus/CondenserLoopTowers.hh [hunk 5]
```diff
@@ -216,7 +254,7 @@ namespace CondenserLoopTowers {
 
         // From module level variables, apparently the module AirFlowRateRatio was used slightly different from the struct's AirFlowRatio variable
         //  so removing this caused diffs that I did not spend time investigating...they might be fine diffs, check some time later
-        Real64 __AirFlowRateRatio;
+        Real64 airFlowRateRatio;
```
mitchute, line 259: Renamed some variables. CLion is complaining because leading underscores are supposed to indicate a protected identifier.

mitchute (commented):
Walkthrough of changes.

tst/EnergyPlus/unit/ChillerElectric.unit.cc
```diff
@@ -317,8 +317,6 @@ TEST_F(EnergyPlusFixture, ChillerElectric_WaterCooled_Simulate)
     Real64 curLoad = -10000.0;
     bool runFlag = true;
 
-    EnergyPlusData state;
```
mitchute, line 320: ```state``` is a member variable of the test fixture. This shouldn't be redeclared as a local variable here.

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -3800,10 +3792,6 @@ TEST_F(EnergyPlusFixture, OutputReportTabular_GatherHeatEmissionReport)
     EXPECT_EQ(2 * reliefEnergy * DataGlobals::convertJtoGJ, BuildingPreDefRep.emiHVACRelief);
     EXPECT_EQ(condenserReject + coilReject, DataHeatBalance::SysTotalHVACRejectHeatLoss);
     EXPECT_EQ(2 * condenserReject * DataGlobals::convertJtoGJ + coilReject * DataGlobals::convertJtoGJ, BuildingPreDefRep.emiHVACReject);
-
```
mitchute, line 3803: No need to call ```clear_state()``` explicitly after unit tests. This happens automatically by the test fixture before each test.

src/EnergyPlus/CondenserLoopTowers.cc
Myoldmopar: So the reason it doesn't work is because the arguments are fundamentally pointers, and there is no addition for pointers.  You have a couple options.  You can make them const char and then when they are used in error message expressions later, you can create a new string from the char array:  `std::string(cCoolingTowerString) + std::string(something)`.  There is surely a penalty associated with that, but if it is only for error reporting then that is acceptable.

Myoldmopar:
CI results look reasonable.  The numeric changes are minimal and the ERR diffs show varying warnings, some up some down.  And moving that module level variable is bound to cause some diffs, it's actually really great that they are just small diffs.  I'll pull this and take a look at the code changes one more time.

src/EnergyPlus/ChillerIndirectAbsorption.cc
Myoldmopar: I think it works as long as you don't end up trying to add another char * to the array.  As long as you are using it as a standalone variable, or passed into another function, or temporarily cast it to a string then it's fine.  But it should be noted that this could quietly cause a performance degradation if it is frequently passed into a function that expects a string argument.  In that case the string constructor will be called to create a new temporary based on this char array every time the function is called.

src/EnergyPlus/ChillerReformulatedEIR.cc
Myoldmopar: :+1:

src/EnergyPlus/ChillerReformulatedEIR.cc
Myoldmopar: :+1:

src/EnergyPlus/ChillerReformulatedEIR.cc
```diff
@@ -1446,13 +1440,13 @@ namespace ChillerReformulatedEIR {
...
                                                this->ChillerEIRFTYTempMax);
-            if (this->PartLoadCurveType == PLR_LeavingCondenserWaterTemperature) {
+            if (this->PartLoadCurveType == PLR::LeavingCondenserWaterTemperature) {
                 CurveManager::GetCurveMinMaxValues(this->ChillerEIRFPLRIndex,
                                                    this->ChillerEIRFPLRTempMin,
                                                    this->ChillerEIRFPLRTempMax,
                                                    this->ChillerEIRFPLRPLRMin,
                                                    this->ChillerEIRFPLRPLRMax);
-            } else if (this->PartLoadCurveType == PLR_Lift) {
+            } else if (this->PartLoadCurveType == PLR::Lift) {
```
Myoldmopar, line 1452: Just so nice.

src/EnergyPlus/CondenserLoopTowers.hh
```diff
@@ -48,32 +48,70 @@
...
+        Unassigned,
+        FanCycling,
+        FluidBypass
+    };
+
+    enum class CellCtrl {
+        Unassigned,
+        MinCell,
+        MaxCell
+    };
```
Myoldmopar, line 108: :relaxed:

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
Myoldmopar: :+1: :+1:

Myoldmopar (approved):
Very nice changes.  Looks good.

Myoldmopar:
Built and ran locally.  All good.  Merging.  Thanks @mitchute

PR #8086, opened 2020-06-17
Suggested hooks for Spawn

kbenne:
@Myoldmopar I've been working to distill down the hooks that are needed to make spawn work, and I think this is pretty close. It probably isn't ready for merge yet, but I would like to work in the direction of merging this into develop.

There are two main additions.

1. A new callback `externalHVACManager` which as the name suggests, simply circumvents the normal HVACManager routine and instead allows Spawn to hook in and do its thing.

2. A stop simulation api. Sometimes spawn might want to "shut it down" before the run period is over.


src/EnergyPlus/api/runtime.cc
```diff
@@ -73,6 +73,10 @@ int energyplus(int argc, const char *argv[]) {
     return runEnergyPlusAsLibrary(state, argc, argv);
 }
 
+void stopSimulation() {
+  EnergyPlus::DataGlobals::stopSimulation = true;
```
Myoldmopar, line 89: It would be nice to make this stopSimulation a function call so that clients don't have to mine their way into the DataGlobals variable.

src/EnergyPlus/HeatBalanceAirManager.cc
```diff
@@ -4217,7 +4217,11 @@ namespace HeatBalanceAirManager {
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
         // na
 
-        ManageHVAC(state, OutputFiles::getSingleton());
+        if(DataGlobals::externalHVACManager) {
```
Myoldmopar, line 4217: That is indeed a minimal set of changes.

Myoldmopar (commented):
No significant comments now, this is a surprisingly small set of changes to bypass EnergyPlus HVAC.  I think once the stopSimulation is changed from direct variable access to an API function, this could start proceeding toward merge.  I would hope others will comment on this.

Myoldmopar:
I think we should add an API function to register the HVAC function, too.  Rather than just assigning it directly from outside the library.

src/EnergyPlus/HVACManager.cc
```diff
@@ -442,6 +442,7 @@ namespace HVACManager {
 
         if (UseZoneTimeStepHistory) PreviousTimeStep = TimeStepZone;
         for (SysTimestepLoop = 1; SysTimestepLoop <= NumOfSysTimeSteps; ++SysTimestepLoop) {
+            if (DataGlobals::stopSimulation) break;
```
rraustad, line 454: Not sure how fast the break should be, and this is a good spot for the master break, but there is another place in code where the entire HVAC system is simulated in a [while loop](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HVACManager.cc#L907-L914). Some complex systems could take some time to complete if not quickly converged. Just a suggestion.

src/EnergyPlus/api/runtime.cc
kbenne: https://github.com/NREL/EnergyPlus/blob/APIs-and-spawn2/src/EnergyPlus/api/runtime.h#L121 it looks like the function needs some documentation though. I'll add something.

kbenne: Something other than the stopSimulation() function? I figured they would use that. The DataGlobals::stopSimulation was not intended to be used publicly.

src/EnergyPlus/HeatBalanceAirManager.cc
kbenne: Oh yeah!

src/EnergyPlus/HVACManager.cc
kbenne: Agreed. I will add a break down in the HVAC manager. See I wasn't concerned about that because my use case is circumventing EnergyPlus's HVAC manager and replacing with Modelica. For completion of the feature I think your idea is correct.

kbenne:
> I think we should add an API function to register the HVAC function, too. Rather than just assigning it directly from outside the library.

Yes. Fine by me. I'll add a function to register the callback. Does this belong in runtime.h or what did you have in mind?

kbenne:
Thanks for looking @rraustad and @Myoldmopar

src/EnergyPlus/HVACManager.cc
kbenne: I added this https://github.com/NREL/EnergyPlus/pull/8086/commits/52e942a8ef8ee2c79c6821856f3c620ae3cdbecf

kbenne:
@Myoldmopar I added a function to register the external HVAC manager https://github.com/NREL/EnergyPlus/commit/11b501579bde86ccda273fd61b175a64b1512228

kbenne:
@Myoldmopar it shows pending but the dashboard is showing up to date and passing results for all checks. I propose to close this PR and open a new one against the exact same branch and commit. I think we might see green instead of pending here if I do that. Before I do that I want to make sure there is no further discussion on this PR.

Myoldmopar:
This looks _really_ good.  There are only a few things I can think of that we could do to wrap this up.

1. Do you think you could expand on the documentation of the HVAC registration function in the API to explain what an overriding function would need to do?  Maybe not gory details, but maybe a list of possible sensor/actuator combinations that would allow a client to do the HVAC overriding, even in an overly simple fashion.

2. I think we should go ahead and add this to the Python API as well, it's just a matter of adding it to runtime.py.  I can do that if you don't have any concerns with it.

3. I think we should probably test this, in an overly simple fashion.  Some lines should be added to TestRuntime.c and TestRuntime.py, or if you think it should be kept totally separate that's fine too.

I'm happy to tackle 2, and if you can add just a few extra lines of documentation with 1, I'll be happy to do 3.

kbenne:
Thanks @Myoldmopar, that sounds like a great plan.

Myoldmopar:
OK, so I took care of 2 and the documentation clearly states how this new function is primarily a placeholder for future applications.  I do like that I was able to get the simulation running with external HVAC hooked up *within* our own repo and CI test environment.  And hey, CI is *super* happy with the commit here.  I am inclined to merge.  Anyone have any issue?  Speak up otherwise this will merge in the morning.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -746,10 +746,19 @@ SET( API_CORE_SRC
...
   api/EnergyPlusPgm.cc)
 
+option( BUILD_STATIC_ENERGYPLUS_API "Build a static library instead of a shared library for the EnergyPlus API" OFF )
+mark_as_advanced( BUILD_STATIC_ENERGYPLUS_API )
+if( BUILD_STATIC_ENERGYPLUS_API )
+	set(API_LIBRARY_TYPE STATIC)
+	ADD_DEFINITIONS(-DENERGYPLUS_STATIC_API)
+else()
+	set(API_LIBRARY_TYPE SHARED)
+endif()
```
Myoldmopar, line 756: OK, so now if someone wants to develop tooling around E+ and wants to compile E+ statically, it's just a matter of setting this flag.

src/EnergyPlus/HeatBalanceAirManager.cc
```diff
@@ -4223,6 +4231,16 @@ namespace HeatBalanceAirManager {
 
     // END Algorithm Section of the Module
 
+    void initializeForExternalHVACManager(EnergyPlusData &state) {
```
Myoldmopar, line 4234: Eventually this function will be adapted into a whole suite of functions that perform initializations and cleanups and what-not in order to get our built-in HVAC calculations working well with an external HVAC manager.

src/EnergyPlus/api/runtime.cc
Myoldmopar: Yeah I dunno what I missed when I commented this.

Myoldmopar (approved):

PR #8090, opened 2020-06-17
Finish removing gio::write()

lefticus:
Pull request overview
---------------------

Refactoring to remove gio::write() calls from EnergyPlus

This PR removes the last of the calls to gio::write, gio::write itself and associated gio::Write support types.

## Important Notes

 * Most files affected by this change are debug and test files that are rarely used and haven't been touched (according to the git logs) since the conversion to C++
 * Some have no method in the source *at all* for actually opening an output file, which means they would have defaulted to Unit=0, which is stdout. This behavior has been maintained where possible
 * Removing these calls to functions that never do anything could result in performance increases (smaller binary, saved function calls), but they were maintained
 * A considerable amount of code has been commented out for at least 6 years. That code has not been modified
 * gio::write has been completely removed so that there's no temptation to use it again

## Testing

Even though most of this output is to debug and rarely used files a best-effort was made to ensure that there were no to minimal output file changes.

 * A modified version of EnergyPlusRegressionTool was used to check additional files
 * Debug output was enabled (either temporarily hard coded or via environment variables) where possible, and tested
 * NOTE: some debug output modes did not work at all before these changes because of unopened files. That is still the case (execution order has debug output before output file is opened)
 * At least one bug was fixed where CSV header rows did not have commas separating columns
 * Another bug was addressed where the first field always had a `,` after it instead of the currently selected separator

***My work on this conversion process leads me to believe that TAB and TXT output files are not tested at all.***


Myoldmopar:
That unit test failure has popped up in other branches too, it's not specific to this one.  CI results look great.  I'm looking over the changes now.

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -10029,34 +10026,21 @@ namespace DaylightingManager {
...
+                outputFile.ensure_open("ReportIllumMap");
+                return outputFile;
+            };
             if (MapColSep == CharTab) {
-                {
-                    IOFlags flags;
-                    flags.ACTION("readwrite");
-                    flags.STATUS("UNKNOWN");
-                    ObjexxFCL::gio::open(IllumMap(MapNum).UnitNo, outputFiles.outputMapTabFileName + MapNoString, flags);
-                    if (flags.err()) goto Label901;
```
Myoldmopar, line 10040: :smile:

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -10029,34 +10026,21 @@ namespace DaylightingManager {
...
-                }
+                if (!openMapFile(outputFiles.outputMapTabFileName).good()) return;
                 //				CommaDelimited = false; //Unused Set but never used
             } else if (MapColSep == CharComma) {
-                {
-                    IOFlags flags;
-                    flags.ACTION("readwrite");
-                    flags.STATUS("UNKNOWN");
-                    ObjexxFCL::gio::open(IllumMap(MapNum).UnitNo, outputFiles.outputMapCsvFileName + MapNoString, flags);
-                    if (flags.err()) goto Label902;
```
Myoldmopar, line 10049: :smile:

src/EnergyPlus/DaylightingManager.cc
```diff
@@ -10029,34 +10026,21 @@ namespace DaylightingManager {
...
-                }
+                if (!openMapFile(outputFiles.outputMapCsvFileName).good()) return;
                 //				CommaDelimited = true; //Unused Set but never used
             } else {
-                {
-                    IOFlags flags;
-                    flags.ACTION("readwrite");
-                    flags.STATUS("UNKNOWN");
-                    ObjexxFCL::gio::open(IllumMap(MapNum).UnitNo, outputFiles.outputMapTxtFileName + MapNoString, flags);
-                    if (flags.err()) goto Label903;
```
Myoldmopar, line 10058: :smile:

src/EnergyPlus/OutputFiles.hh
```diff
@@ -142,6 +151,30 @@ private:
...
+    OutputFile &operator*()
+    {
+        return *ptr();
+    }
+
+    OutputFile *operator->()
+    {
+        return ptr();
+    }
+};
```
Myoldmopar, line 176: Interesting.  What brought about the need for this?

Myoldmopar (approved):
Fantastic progress.

Myoldmopar:
I'm _curious_ why you had to add the shared file handle, but I have no problem with it.  This is good to go.  Thanks @lefticus .

lefticus:
> I'm _curious_ why you had to add the shared file handle, but I have no problem with it. This is good to go. Thanks @lefticus .

Because Objexx `Array` types require copyable types. So I needed to wrap my non-copyable `OutputFile` into something that could be copied and have shared references, just for the sake of the implementation.

I thought about moving that code to `std::vector` but decided it was out of scope for this particular effort.

PR #8094, opened 2020-06-19
8093 error callback

### LINKED ISSUE #8093: Error Callback
The error function callback provided by EnergyPlus https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/DataGlobals.hh#L296 does not inform the severity of the error. Consider adding an argument to indicate error level.

kbenne:
Fixes #8093


nrel-bot-3:

kbenne:
@Myoldmopar this is additive on top of https://github.com/NREL/EnergyPlus/pull/8086

src/EnergyPlus/DataGlobals.cc
```diff
@@ -251,9 +251,9 @@ namespace DataGlobals {
     int Progress(0); // current progress (0-100)
     void (*fProgressPtr)(int const);
     void (*fMessagePtr)(std::string const &);
-    void (*progressCallback)(int const);
-    void (*messageCallback)(const char * message);
-    void (*errorCallback)(const char * errorMessage);
+    std::function<void(int const)> progressCallback;
+    std::function<void(const std::string &)> messageCallback;
+    std::function<void(EnergyPlus::Error e, const std::string &)> errorCallback;
```
Myoldmopar, line 256: I do like moving to using std::function as long as we are inside the C++ boundary.

src/EnergyPlus/api/runtime.cc
```diff
@@ -104,16 +104,23 @@ void registerProgressCallback(EnergyPlusState EP_UNUSED(state), void (*f)(int co
...
+void registerStdOutCallback(EnergyPlusState EP_UNUSED(state), std::function<void (const std::string &)> f) {
     EnergyPlus::DataGlobals::messageCallback = f;
 }
 
+void registerStdOutCallback(EnergyPlusState state, void (*f)(const char * message)) {
+    const auto stdf = [f](const std::string & message) {
+        f(message.c_str());
+    };
+    registerStdOutCallback(state, std::function<void (const std::string &)>(stdf));
+}
```
Myoldmopar, line 96: I feel like really we should invert all the register callbacks to this approach.

src/EnergyPlus/api/TypeDefs.h
```diff
@@ -51,4 +51,18 @@
...
+enum class Error: int {
+  Continue = 0,
+  Info = 1,
+  Warning = 2,
+  Severe = 3,
+  Fatal = 4
+};
+
+} // namespace EnergyPlus
+#endif
```
Myoldmopar, line 66: So if you are accessing this endpoint in C, or Python, you'll just need to hardcode the numbers?  I do like the use of a proper enum class inside C++, but what are your thoughts on balancing that nicety vs having to hardwire numbers elsewhere.  If we went with a plain C enum, then we could access it nicely through the API.

Myoldmopar (commented):
I'd like to chat about the enum mostly, otherwise this is pretty good.

src/EnergyPlus/api/TypeDefs.h
kbenne: Yeah this certainly does prioritize C++, which to me is a feature not a bug but that is subjective. The C api is just an int https://github.com/NREL/EnergyPlus/blob/8093-error-callback/src/EnergyPlus/api/func.h#L94 

A C style enum would be ok, but what would the enum values be named? There are no namespaces so you would probably want something like ENERGYPLUS_ERROR_CONTINUE, or something like that. Perhaps with less uppercase shouting.

@lefticus you got any input?

Myoldmopar:
OK, I got develop pulled in, but I think I might like to move the error callback into the runtime API section where all the other callback registration functions live.  Any issue with that?

Myoldmopar:
OK, reproduced the errors locally and fixed them.  I anticipate this will be ready to go later today after CI verifies.

Myoldmopar:
I'm not going to wait on integration debug to pass, this change is totally unrelated.  Here you go @kbenne, merging in.

PR #8100, opened 2020-06-21
Resilience Metrics Reporting

xuanluo113:
Pull request overview
---------------------
With increasing frequency and severity of extreme weather events (e.g., heat waves), it is crucial to ensure urban buildings and infrastructure are resilient to provide critical services to preserve human life and properties during natural disasters. Building resilience has the opportunity to become an additional value proposition for technologies and systems if it can be reliably quantified, valued, and trusted by stakeholders. Metrics or an assessment of potential vulnerability, likelihood, and consequence for each risk would help to prioritize further consideration of these risk factors of building resilience [1]. There is a need to measure the resilience of buildings to help owners make better decisions and protect their assets, to better assess the built environment resilience of larger geographic units such as communities and cities, and to complement existing assessments of building sustainability. Building performance simulation is a powerful tool to quantify building’s resilience and vulnerability under extreme weather events.

EnergyPlus currently calculates at each time-step a set of variables (e.g., zone/space air dry-bulb temperature, relative humidity, CO2 concentration, illuminance level) that can be further processed to generate reports on resilience metrics for thermal performance, visual performance, and indoor air quality (IAQ). For the spatial resolution, the reporting needs to consider both the entire building for aggregated results and individual spaces for detailed and worst-case results. For the temporal resolution, the reporting needs to consider the aggregated and average results as well as the worst-case results (e.g., maximum or minimum values at a specific time, worst day, worst week) in summer, winter, and the whole year. The resilience metrics target occupants’ health and wellbeing during extreme weather events. Therefore, the metrics involve the aggregations across space (occupants in the space) and time (e.g., X occupant·hours (Y%) with Heat Index between certain range for Zone Z). Post-processing to generate these reports requires EnergyPlus models to produce lots of time-step report variables, which can be time-consuming and can require significant storage. Hence, we propose to add a list of resilience-related tabular reports in EnergyPlus to provide a consistent and convenient path for users of resilience analysis through modeling and simulation.

### Pull Request Author

### Reviewer


src/EnergyPlus/DataHeatBalFanSys.hh
```diff
@@ -232,6 +232,24 @@ namespace DataHeatBalFanSys {
     extern Array1D_int TempControlType;
     extern Array1D_int ComfortControlType;
 
+    extern Array1D<Real64> ZoneHeatIndex;
+    extern Array1D<Real64> ZoneHumidex;
+    extern Array1D_int ZoneNumOcc;
+
+    extern Array1D<std::vector<double>> ZoneHeatIndexHourBins;
```
Myoldmopar: Use `Real64` typedef not `double`

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -213,6 +213,9 @@ namespace OutputReportTabular {
     extern bool displayEconomicResultSummary;
     extern bool displayHeatEmissionsSummary;
     extern bool displayEioSummary;
+    extern bool displayThermalResilienceSummary;
```
Myoldmopar, line 220: Are these reset in any clear_state calls?

Myoldmopar (changes requested):
I built this locally and verified the failing unit test suite immediately.  The unit test fails on an OutputReportTabular unit test, and I see you added some new global variables to that file, and I don't see those being reset in a clear_state.  I may have just missed the reset, in which case the problem is elsewhere, but it certainly seems suspicious.

src/EnergyPlus/OutputReportTabular.hh
xuanluo113: @Myoldmopar I suppose they are reset [here](https://github.com/NREL/EnergyPlus/blob/9805e969689297e1395bc106d9cc9ac4d1cbf0c5/src/EnergyPlus/OutputReportTabular.cc#L594).

Myoldmopar: Got it, I zipped right past that.  OK, the problem is elsewhere.

Myoldmopar:
@xuanluo113 there is definitely something not being reset.  The integration test failure is failing on the TestRuntimeC test.  During this test, EnergyPlus is executed, reset, and then executed again.  When I run the unit test suite locally I see the same thing -- serially the unit test fails.  But when CI runs the unit tests, they don't fail, because it doesn't run them serially.  I'm building a debug build and should be able to find the problem quickly, but if not I'll just need you to debug in and find what isn't being reset here.

xuanluo113:
> @xuanluo113 there is definitely something not being reset. The integration test failure is failing on the TestRuntimeC test. During this test, EnergyPlus is executed, reset, and then executed again. When I run the unit test suite locally I see the same thing -- serially the unit test fails. But when CI runs the unit tests, they don't fail, because it doesn't run them serially. I'm building a debug build and should be able to find the problem quickly, but if not I'll just need you to debug in and find what isn't being reset here.

@Myoldmopar Thanks so much. I didn't get the CI's complaints. My local builds, on both mac and ubuntu, do no reproduce the error so I kind of got stuck. Good to know and I can look at the issue in the reset direction as well.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4876,6 +4889,313 @@ namespace HeatBalanceSurfaceManager {
...
+
+        int HINoBins = 5; // Heat Index range - number of bins
+        int HumidexNoBins = 5; // Humidex range - number of bins
+        int SETNoBins = 4; // SET report column numbers
+        static std::vector<double> lowSETLongestHours(NumOfZones, 0.0);
+        static std::vector<double> highSETLongestHours(NumOfZones, 0.0);
+        static std::vector<int> lowSETLongestStart(NumOfZones, 0.0);
+        static std::vector<int> highSETLongestStart(NumOfZones, 0.0);
+        static bool hasPierceSET = true;
+        static bool oneTimeFlag = true;
```
Myoldmopar: You **cannot** use static flags inside functions.  There's no way to reset this in a clear_state because it is hidden inside a function.  There are these two plus two inside other Report* functions you added here, and they have to be extracted out and become variables on the namespace, and then they need to be cleared in a clear_state call.  This should fix the problem.

xuanluo113: I see. That makes sense. Thanks!

Myoldmopar:
@xuanluo113 this does look a lot better.  Locally running shows no problem.  Once CI starts working on your latest commit I anticipate it will be all green.  There are still a few static variables inside the functions, however.  Are you sure those don't need to be reset in a clear_state as well?

xuanluo113:
@Myoldmopar I thought those were fine as those vectors are [reinitialized](https://github.com/NREL/EnergyPlus/blob/101849dd2a92326eb15a445e4c80a3a3a3dfc8f5/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4951) at the beginning of each simulation. But seems the API test still fails. Will try clearing those.

Myoldmopar:
No, CI hasn't run that latest commit yet.  Those results are for 9805e96.  However, there is now a conflict in the heat balance surface manager.  I suggest you wait until CI reports on 101849d, make sure it is happy, then move forward, addressing the conflict and any other identified issues.

Myoldmopar:
CI looks happy...just one report so far but still it's a good sign!

xuanluo113:
Thanks. Just pushed another commit adding more unit tests and documentation, and merging conflicts.

Myoldmopar:
@xuanluo113 there are two issues going on.  One is that your new test file is failing with an array index out of bounds in a debug build.  Two is that this branch is now causing HospitalLowEnergy to time out again...we had that file running much faster, so I'm not sure if something in your branch is causing a huge slow-down or what is happening.

xuanluo113:
@Myoldmopar Now CI turns green. The `HospitalLowEnergy` timeout was due to all the [warnings](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/OutputReportTabular.cc#L11437) printed when `AllSummary` is declared in the model. When that's declared, the Thermal Resilience reports are required to be included in the tabular reports. But SET-hour report requires the People defined in a zone and Pierce thermal comfort method defined for the People. Similar to the Visual and CO2 reports. Actually all the "587 tests had: ERR diffs." is triggered by this, which does not look good. Apart from (1) do not report resilience tables when `AllSummary` is declared; (2) eliminate/reduce the warnings, do you have any better suggestions?

Myoldmopar:
@xuanluo113 so first, CI is clearly reporting to you that the TeX changes have error messages, and they are pointing you to the exact lines.  There are 3 instances of `&` in your changes that need to be escaped to be valid TeX.  

> The HospitalLowEnergy timeout was due to all the warnings printed when AllSummary is declared in the model.

I don't see how printing a warning could've caused the hospital file to time out, but maybe I'm misunderstanding.

> do you have any better suggestions?

It depends on the definition of `AllSummary`.  Does `AllSummary` mean that **every** single summary must be written?  Or does it mean that **every summary valid for the current simulation** must be written.  I'm inclined to say that you should just write the summaries that **are** valid, and ignore the ones that aren't.  Otherwise we need to modify every IDF to include those other objects, or we need to add a new key `AllSummaryExceptSomeStuff`.  I guess you could add a one time extra warning message that says "Some summaries could not be written because you didn't include necessary objects", but that's still going to add a warning to like every IDF output.

mjwitte:
@xuanluo113 @Myoldmopar I would suggest following the pattern used by the EconomicResultSummary report. It is activated by AllSummary* options, but it does not report if there are no tariffs, no warnings issued. See [code here.](https://github.com/NREL/EnergyPlus/blob/a6234498a7f4244983ac3d778025fa3e41b2c69a/src/EnergyPlus/EconomicTariff.cc#L4242-L4244) Would that approach work here? And this could be noted in the I/O ref output reports description for this report - "... is not produced if there are no ..... objects."

mjwitte:
Or another option is to produce the report without warnings and without content, but that seems silly. If this report is specifically requested, you could throw a warning if the required inputs are not present.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4877,6 +4906,311 @@ namespace HeatBalanceSurfaceManager {
...
+
+        Real64 c1 = -42.379;
+        Real64 c2 = 2.04901523;
+        Real64 c3 = 10.14333127;
+        Real64 c4 = -.22475541;
+        Real64 c5 = -.00683783;
+        Real64 c6 = -.05481717;
+        Real64 c7 = .00122874;
+        Real64 c8 = .00085282;
+        Real64 c9 = -.00000199;
```
mjwitte, line 5026: What are these? Should have some comments explaining what these are and citing a source.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4877,6 +4906,311 @@ namespace HeatBalanceSurfaceManager {
...
+            // Record last time step SET to trace SET unmet duration;
+            for (int iPeople = 1; iPeople <= TotPeople; ++iPeople) {
+                int ZoneNum = People(iPeople).ZonePtr;
+                ZoneNumOcc(ZoneNum) = People(iPeople).NumberOfPeople * GetCurrentScheduleValue(People(iPeople).NumberOfPeoplePtr);
+                ZoneOccPierceSETLastStep(ZoneNum) = ZoneOccPierceSET(ZoneNum);
+                if (ZoneNumOcc(ZoneNum) > 0) {
+                    if (People(iPeople).Pierce) {
+                        ZoneOccPierceSET(ZoneNum) = ThermalComfort::ThermalComfortData(iPeople).PierceSET;
+                    } else {
+                        ZoneOccPierceSET(ZoneNum) = NAN;
```
mjwitte: Do we really want to be setting NANs? This will halt in a debug config (at least the way the windows debug config is currently set up).

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -11198,6 +11250,276 @@ namespace OutputReportTabular {
...
+            WriteReportHeaders(tblTitle, "Entire Facility", OutputProcessor::StoreType::Averaged);
+
+            int columnNum = 5;
+            std::string subTitle = "Heat Index Hours";
+            std::vector<std::string> columnHeadNames = { "Safe (≤ 80°F) [Hours]",
+                                                         "Caution (80, 90°F] [Hours]",
+                                                         "Extreme Caution (90, 103°F] [Hours]",
+                                                         "Danger (103, 125°F] [Hours]",
+                                                         "Extreme Danger (> 125°F) [Hours]" };
+            WriteResilienceBinsTable(columnNum, tblTitle, subTitle, sqlTitle, columnHeadNames, ZoneHeatIndexHourBins);
```
mjwitte: @xuanluo113 This seems like the hard way to build a new table report instead of using the `OutputReportPredined` functions like `newPreDefReport` and `newPreDefColumn`.
Is this hard-wired to always write deg F and other data in deg C below? Output tables should be sensitive to units conversion options specified in OutputControl:Table:Style. @JasonGlazer ?

xuanluo113:
@Myoldmopar @mjwitte Thanks for the clarification. The warnings were quite verbose (perhaps causing the timeout), printing each specific Zone with no People - Pierce model, and with no `Daylighting:Control` when visual resilience is declared.
I would go with Mike's second solution, as I feel it's necessary to give a warning when users specific require a report, and it's okay to be silent otherwise.

Another question I have is, I wanted to [set up the variable of Heat Index](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4942) only when Thermal Resilience Report is declared. But [this flag](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/OutputReportTabular.cc#L2075) is read in `UpdateTabularReports`, which is called 
> at the end of the time step loop and updates the arrays of data that will later be put into the tabular reports.

Seems if only weather file run period is defined, the Zone Heat Index is not showing in the RDD for some reason. Is there any better way to prefetch this flag? Thanks for helping out.

src/EnergyPlus/OutputReportTabular.cc
xuanluo113: I was following the style of [WriteAdaptiveComfortTable](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/OutputReportTabular.cc#L11157) first and was not aware of the `OutputReportPredined` style. I could clean them up to follow the predefined style then.

There's nothing to do with deg F or C. The standard [equations](https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml) and [thresholds](https://www.weather.gov/ama/heatindex) are in deg F, so the table headers are written this way. The calculation results are converted back to [deg C](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4963) at the time step loop.

JasonGlazer: The column headings having ranges of values in F is an unusual case. I would think the SI version of the table the values should be shown in ranges of C instead of F. You might need to be sensitive to the unit system requested by the user and change the values when defining the table.

xuanluo113: @JasonGlazer Thanks for the clarification. I'll convert the headers to deg C thresholds. 
Regarding the predefined table style, if I want to add Zone Min, Max, Average and Sum of a runtime variable to the table end, is it the best practice to calculate these aggregation values in the [FillRemainingPredefinedEntries](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/OutputReportTabular.cc#L6447) function?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: @mjwitte This is to explicitly set the value to some kind of "null" instead of zero during runtime at certain time steps (no occupant). If that is not working, it would work the same to set it to some unreasonable values like "999" or "-999" for temperatures or add more logics with flags like `ZoneOccPierceSETIsNull` and `ZoneOccPierceSETLastTimeStepIsNull`.

Myoldmopar: -999 is fine.  Or just -1 if that can be used as a sentinel?

xuanluo113:
@Myoldmopar @mjwitte @JasonGlazer Seems the CI shows green for the latest commit. A few updates addressing the previous comments.

1. Fixed NaN.
2. Fixed the I/O documentation compiling error.
3. Display warnings only when resilience flags are explicitly declared.
4. Converted the tables to the predefined style.
5. Always calculate the Heat Index and Humidex and set the RDDs. (RDD diffs)

A remaining issue is some extra warnings regarding the report header units, like:
```
+   ** Warning ** Unable to find a unit conversion from Caution (26.7, 32.2°C] [Hours] into IP units
+   **   ~~~   ** Applying default conversion factor of 1.0
```
Is it for the brackets and parentheses used here? Seems [this](https://github.com/NREL/EnergyPlus/blob/2ef4fbe91abb3699c687a1920db0eef78619e379/src/EnergyPlus/OutputReportTabular.cc#L16630) function is trying the evaluate "Hours]" after parsing.

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -2650,6 +2768,75 @@ namespace OutputReportPredefined {
...
         pdchLeedSchStPt11pmWedCnt = newPreDefColumn(pdstLeedSchedSetPts, "Days with Same 11pm Value");
+
+        pdrThermalResilience = newPreDefReport("ThermalResilienceSummary", "ThermR", "Annual Thermal Resilience Summary");
+
+        pdstHIHours = newPreDefSubTable(pdrThermalResilience, "Heat Index Hours");
+        pdchHIHourSafe = newPreDefColumn(pdstHIHours, "Safe (≤ 26.7°C) [Hours]");
+        pdchHIHourCaution = newPreDefColumn(pdstHIHours, "Caution (26.7, 32.2°C] [Hours]");
+        pdchHIHourExtremeCaution = newPreDefColumn(pdstHIHours, "Extreme Caution (32.2, 39.4°C] [Hours]");
+        pdchHIHourDanger = newPreDefColumn(pdstHIHours, "Danger (39.4, 51.7°C] [Hours]");
+        pdchHIHourExtremeDanger = newPreDefColumn(pdstHIHours, "Extreme Danger (> 51.7°C) [Hours]");
```
JasonGlazer: The headers for these columns are only in SI units and won't be getting converted automatically. Maybe the newPreDefColumn needs to have an new optional third argument which is the IP version of the column header.

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -2650,6 +2768,75 @@ namespace OutputReportPredefined {
...
+
+        pdstHIOccuHours = newPreDefSubTable(pdrThermalResilience, "Heat Index OccupantHours");
+        pdchHIOccuHourSafe = newPreDefColumn(pdstHIOccuHours, "Safe (≤ 26.7°C) [OccupantHours]");
+        pdchHIOccuHourCaution = newPreDefColumn(pdstHIOccuHours, "Caution (26.7, 32.2°C] [OccupantHours]");
+        pdchHIOccuHourExtremeCaution = newPreDefColumn(pdstHIOccuHours, "Extreme Caution (32.2, 39.4°C] [OccupantHours]");
+        pdchHIOccuHourDanger = newPreDefColumn(pdstHIOccuHours, "Danger (39.4, 51.7°C] [OccupantHours]");
+        pdchHIOccuHourExtremeDanger = newPreDefColumn(pdstHIOccuHours, "Extreme Danger (> 51.7°C) [OccupantHours]");
```
JasonGlazer: What is difference between Hours and OccupantHours? OccupantHours is a new unit and we try to avoid adding them unless we have a good reason.

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -2650,6 +2768,75 @@ namespace OutputReportPredefined {
...
+        pdstHumidexOccuHours = newPreDefSubTable(pdrThermalResilience, "Humidex OccupantHours");
+        pdchHumidexOccuHourLittle = newPreDefColumn(pdstHumidexOccuHours, "Little to no Discomfort (≤ 29) [OccupantHours]");
+        pdchHumidexOccuHourSome = newPreDefColumn(pdstHumidexOccuHours, "Some Discomfort (29, 40] [OccupantHours]");
+        pdchHumidexOccuHourGreat = newPreDefColumn(pdstHumidexOccuHours, "Great Discomfort; Avoid Exertion (40, 45] [OccupantHours]");
+        pdchHumidexOccuHourDanger = newPreDefColumn(pdstHumidexOccuHours, "Dangerous (45, 50] [OccupantHours]");
+        pdchHumidexOccuHourStroke = newPreDefColumn(pdstHumidexOccuHours, "Heat Stroke Quite Possible (> 50) [OccupantHours]");
+
+        pdstHeatingSETHours = newPreDefSubTable(pdrThermalResilience, "Heating SET Hours");
+        pdchHeatingSETHours = newPreDefColumn(pdstHeatingSETHours, "SET ≤ 12.2°C Hours (°C)");
+        pdchHeatingSETOccuHours = newPreDefColumn(pdstHeatingSETHours, "SET ≤ 12.2°C OccupantHours (°C)");
```
JasonGlazer, line 2856: I assume these are also hours, please add [hr]. Actually in general, all the "[Hours]" should just be "[hr]"

xuanluo113: Okay. Thanks.

src/EnergyPlus/OutputReportPredefined.cc
xuanluo113: Occupant * hours. So the unit should be [hrs] I guess. I'll change them to be consistent.

JasonGlazer:
@xuanluo113 as far as performance, you could change the calculations for the heat index and humidity index so they only occur when someone requests the variables or the tabular report that shows them.

xuanluo113:
@JasonGlazer I wanted to [set up the variable of Heat Index](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4942) only when Thermal Resilience Report is declared. But [this flag](https://github.com/NREL/EnergyPlus/blob/1426e755443ea4f8fe19497d4c7be554ac27a048/src/EnergyPlus/OutputReportTabular.cc#L2075) is read in `UpdateTabularReports`, which is called 
> at the end of the time step loop and updates the arrays of data that will later be put into the tabular reports.

Seems if only weather file run period is defined, the Zone Heat Index is not showing in the RDD for some reason. Do you know if there's a better way to prefetch this flag? Thanks for helping out.

JasonGlazer:
The output variable setup should always be done so that it shows up in the RDD but if the output variable is not requested and the table is not requested, you can skip the computation of the metric.

xuanluo113:
@JasonGlazer That makes sense. Thank you!

xuanluo113:
@JasonGlazer @Myoldmopar The remaining issue was fixed and the branch is up-to-date.

doc/engineering-reference/CMakeLists.txt
```diff
@@ -148,6 +148,8 @@ SET( INCLUDED_TEX
   ${PROJECT_SOURCE_DIR}/engineering-reference/src/special-modules-reporting/environmental-impacts.tex
   ${PROJECT_SOURCE_DIR}/engineering-reference/src/special-modules-reporting/zone-component-loads-summary.tex
   ${PROJECT_SOURCE_DIR}/engineering-reference/src/special-modules-reporting/output-table-annual.tex
+  ${PROJECT_SOURCE_DIR}/engineering-reference/src/special-modules-reporting/heat-emissions.tex
+  ${PROJECT_SOURCE_DIR}/engineering-reference/src/special-modules-reporting/resilience-metrics.tex
```
Nigusse, line 152: I see you have included *resilience-metrics* documentation but I don't see the *Resilience Metrics* section when I compile the engineering-reference project.  @xuanluo113 did you compile and confirmed that the *Resilience Metrics*  section is in the pdf?

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -4945,6 +4945,8 @@ \subsection{Zone Thermal Output(s)}\label{zone-thermal-outputs-1}
 HVAC,Average,Zone Air Temperature [C]
 Zone,Average,Zone Mean Radiant Temperature [C]
 Zone,Average,Zone Operative Temperature [C]
+Zone,Average,Zone Heat Index [C]
+Zone,Average,Zone Humidity Index [C]
```
Nigusse, line 4949: OK. I see you have added these two zone level temporal variables.

doc/input-output-reference/src/standard-output-reports/output-table-summaryreports.tex
```diff
@@ -429,6 +429,17 @@ \subsubsection{Predefined Annual Summary Reports}\label{predefined-annual-summar
...
+
+The CO2 Resilience Summary provides the indoor CO2 level unmet hours and occupant$·$hours summary report for resilience. This includes the CO2 Level Hours and OccupantHours reports. To activate the $CO_2$ concentration calculation in EnergyPlus, the \textbf{ZoneAirContaminantBalance} object needs to be specified and with the field \textbf{Carbon Dioxide Concentration} set to Yes. Users can define a schedule of outdoor air $CO_2$ concentration in the field \textbf{Outdoor Carbon Dioxide Schedule Name}. $CO_2$ generation rate at the zone level can be specified using the \textbf{ZoneContaminantSourceAndSink:CarbonDioxide} object.
+
+\paragraph{Visual Resilience Summary}\label{VisualResilienceSummary}
+
+The Visual Resilience Summary provides the indoor illuminance level unmet hours and occupant$·$hours summary report for resilience. This includes the Illuminance Level Hours and OccupantHours reports. To activate the indoor illuminance calculation in EnergyPlus, users need to define the \textbf{Daylighting:Controls} and the \textbf{Daylighting:ReferencePoint} objects, even if no daylighting controls are actually implemented in the building simulation model.
```
Nigusse, line 442: Compiled the input-out-put reference documentation and confirmed that this new section is included.

src/EnergyPlus/DataHeatBalFanSys.cc
```diff
@@ -240,6 +240,23 @@ namespace DataHeatBalFanSys {
...
+    Array1D<std::vector<Real64>> ZoneCO2LevelHourBins;
+    Array1D<std::vector<Real64>> ZoneCO2LevelOccuHourBins;
+    Array1D<std::vector<Real64>> ZoneLightingLevelHourBins;
+    Array1D<std::vector<Real64>> ZoneLightingLevelOccuHourBins;
+
+    Array1D<Real64> ZoneOccPierceSET;
+    Array1D<Real64> ZoneOccPierceSETLastStep;
+    Array1D<std::vector<Real64>> ZoneLowSETHours; // LowSETHour, LowSETOccHour, LowSetLongestDur, LowSetLongestTimeStamp
+    Array1D<std::vector<Real64>> ZoneHighSETHours; //HighSETHour, HighSETOccHour, HighSetLongestDur, HighSetLongestTimeStamp
+
```
Nigusse, line 259: All real vectors are now defined as *Real64* as suggested in earlier review comment.

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -2650,6 +2768,75 @@ namespace OutputReportPredefined {
...
+        pdchHumidexHourGreat = newPreDefColumn(pdstHumidexHours, "Great Discomfort; Avoid Exertion (> 40, ≤ 45) [hr]");
+        pdchHumidexHourDanger = newPreDefColumn(pdstHumidexHours, "Dangerous (> 45, ≤ 50) [hr]");
+        pdchHumidexHourStroke = newPreDefColumn(pdstHumidexHours, "Heat Stroke Quite Possible (> 50) [hr]");
+
+        pdstHumidexOccuHours = newPreDefSubTable(pdrThermalResilience, "Humidex OccupantHours");
+        pdchHumidexOccuHourLittle = newPreDefColumn(pdstHumidexOccuHours, "Little to no Discomfort (≤ 29) [hr]");
+        pdchHumidexOccuHourSome = newPreDefColumn(pdstHumidexOccuHours, "Some Discomfort (> 29, ≤ 40) [hr]");
+        pdchHumidexOccuHourGreat = newPreDefColumn(pdstHumidexOccuHours, "Great Discomfort; Avoid Exertion (> 40, ≤ 45) [hr]");
+        pdchHumidexOccuHourDanger = newPreDefColumn(pdstHumidexOccuHours, "Dangerous (> 45, ≤ 50) [hr]");
+        pdchHumidexOccuHourStroke = newPreDefColumn(pdstHumidexOccuHours, "Heat Stroke Quite Possible (> 50) [hr]");
```
Nigusse, line 2852: Now the heading now replaced with [hr] instead of [hours] per previous review comments. And temperatures in summary tables heading are in Celsius scale.

src/EnergyPlus/OutputReportPredefined.cc
Nigusse: Resilience-Metrics table summary outputs format all look fine. Below is a ample for quick look.

SampleResilienceTableSummary

src/EnergyPlus/OutputReportPredefined.cc
```diff
@@ -2650,6 +2768,75 @@ namespace OutputReportPredefined {
...
+        pdchIllumHourDim = newPreDefColumn(pdstIllumHours, "Dim (> 100, <= 300 lux) [hr]");
+        pdchIllumHourAdequate = newPreDefColumn(pdstIllumHours, "Adequate (> 300, <= 500 lux) [hr]");
+        pdchIllumHourBright = newPreDefColumn(pdstIllumHours, "Bright (>500 lux) [hr]");
+
+        pdstIllumOccuHours = newPreDefSubTable(pdrVisualResilience, "Illuminance Level OccupantHours");
+        pdchIllumOccuHourDark = newPreDefColumn(pdstIllumOccuHours, "A Bit Dark (<= 100 lux) [hr]");
+        pdchIllumOccuHourDim = newPreDefColumn(pdstIllumOccuHours, "Dim (> 100, <= 300 lux) [hr]");
+        pdchIllumOccuHourAdequate = newPreDefColumn(pdstIllumOccuHours, "Adequate (> 300, <= 500 lux) [hr]");
+        pdchIllumOccuHourBright = newPreDefColumn(pdstIllumOccuHours, "Bright (>500 lux) [hr]");
+
```
Nigusse, line 2891: Resilience-Metrics IAQ and Illuminance table summary outputs format look fine. Below is a ample for quick look.

SampleResilienceIAQIlluminanceSummary

tst/EnergyPlus/unit/HeatBalanceSurfaceManager.unit.cc
```diff
@@ -2936,7 +3087,5 @@ TEST_F(EnergyPlusFixture, HeatBalanceSurfaceManager_TestInitHBInterzoneWindow)
 
     EXPECT_TRUE(allocated(DataHeatBalSurface::FractDifShortZtoZ));
     EXPECT_EQ(1, DataHeatBalSurface::FractDifShortZtoZ(1, 1));
-
 }
-
```
Nigusse: Merged in develop and resolved the conflict.

Nigusse: Built and ran the branch with develop pulled in and all unit tests passed.

src/EnergyPlus/OutputReportTabular.hh
```diff
@@ -820,6 +825,20 @@ namespace OutputReportTabular {
...
+    void WriteVisualResilienceTables();
+
+    void WriteResilienceBinsTable(int const columnNum,
+                                  std::vector<int> const &columnHead,
+                                  Array1D<std::vector<double>> const &ZoneBins);
+
+    void WriteSETHoursTable(int const columnNum,
+                            std::vector<std::string> const &columnHead,
+                            Array1D<std::vector<double>> const &ZoneBins);
+
```
Nigusse, line 841: @xuanluo113 Please replace double with Real64 per previous review comment from @Myoldmopar.

Nigusse (commented):
The code change and unit test look good. Pulled in develop, built and ran the branch.  All unit test passed. I have a couple comments that @xuanluo113 needs to address: 
-  Complied the Eng/Ref documentation but I could not see/find the Resilience Metric Sub-section in the pdf file.
- a couple of variable declaration still use *double* instead of *Real64*. See review comments. 

The updated branch got pushed. In the meantime I will test the new example file.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
Nigusse: The RDD file shows these two zone level report variables:

```
Output:Variable,*,Zone Heat Index,hourly; !- Zone Average [C]
Output:Variable,*,Zone Humidity Index,hourly; !- Zone Average []
```

Nigusse:
The following three tabular summary reports will not be generated unless specified in *Output:Table:SummaryReports* object:

```
- Annual Thermal Resilience Summary
- Annual CO2 Resilience Summary
- Annual Visual Resilience Summary
```

Confirmed by commenting out and running the in the example file provided. 
```
  Output:Table:SummaryReports,
    AnnualBuildingUtilityPerformanceSummary,  !- Report 1 Name
    DemandEndUseComponentsSummary;            !- Report 2 Name
!    ThermalResilienceSummary,                 !- Report 3 Name
!    CO2ResilienceSummary,                     !- Report 4 Name
!    VisualResilienceSummary;                  !- Report 5 Name

```

Myoldmopar:
It's disappointing that the `double` declarations weren't cleaned up.  @Nigusse thanks for taking a look at this.  It looks like it is good outside of the `double` declarations, and the documentation isn't showing up because the new TeX file was never included in the table of contents.  It should've been listed [here](https://github.com/NREL/EnergyPlus/blob/6806643a4f2a1bd773f1aab83ba3d3e2306cfc3d/doc/engineering-reference/engineering-reference.tex#L306).

If this continues to come back clean then I won't waste more CI cycles right now.  Instead I'll open an issue that @xuanluo113 will need to address by cleaning up the double declarations and ensuring the new documentation shows up in the engineering reference.

Myoldmopar:
Built and tested locally with develop; all is well.  CI diffs are as expected.  I'm going to merge this and open up a follow-up issue.

PR #8109, opened 2020-06-25
Allow EMS to actuate mass flow and supply temperature for Ideal Loads when the zone load is zero

lgu1234:
Pull request overview
---------------------
Fixes part of #5788 

EMS output variables of mass flow rate and supply air temperature as EMS actuators are reported inconsistently with system values during PurchaseAir system off. The fix ensure: EMS mass flow rate = 0, and EMS supply air temperature = PurchaseAir SupplyTemp.

Based on current calling sequence in HVACManager, when EMS is called at the last iteration, ZoneEquipment, a parent of IdealLoads, is possible to get bypassed, so that EMS overrides in IdealLoads module only is not enough to provide correct outputs due to bypass of ZoneEquipment calling. RuntimeLanguageProcessor module is revised slightly to ensure correct EMS outputs are reported during system off.  

An additional test file is provided as 5788-1.idf, because current 2 test files are crashed in Debug Mode.

### Pull Request Author

### Reviewer


src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -179,6 +181,8 @@ namespace PurchasedAirManager {
     int NumPurchAir;
     int NumPlenumArrays;
     bool GetPurchAirInputFlag(true);
+    Array1D_bool EMSMdotSetupFlag;
+    Array1D_bool EMSTsupSetupFlag;
```
Myoldmopar: In this branch you added two new variables to the purchased air data struct:
 - int EMSMassFlowRateOutVarNum;
 - int EMSSupplyTempOutVarNum;     

Why not just add two more:
- bool EMSMdotSetupFlag;
- bool EMSTsupSetupFlag;

That way you don't have to have these extra arrays hanging out and allocate them separately and what not.

Myoldmopar (commented):
I have reservations about these changes.  Adding a bypass flag to override the user's overridden value seems...questionable.  I understand it may make the situation work, but I'm not sure we want to be in the business of doing that.  If users are going to take on EMS to manipulate the program, I don't want to stand in the way...and if we expose actuators that allow them to get into serious trouble, then maybe we shouldn't expose those.  I may just need more time to digest this.  @EnergyArchmage do you have any thoughts on this bypass flag?

src/EnergyPlus/PurchasedAirManager.cc
EnergyArchmage: seems to me that the better fix is to not skip EMS overrides because of schedule or zero load.

lgu1234: @Myoldmopar @EnergyArchmage If EMS is not skipped, the calling structure needs to be revised. Here is a section code in HVACManager:

        ManageEMS(emsCallFromAfterHVACManagers, anyEMSRan); // calling point
        ManageEMS(emsCallFromHVACIterationLoop, anyEMSRan); // calling point id

        // first explicitly call each system type with FirstHVACIteration,

        // Manages the various component simulations
        SimSelectedEquipment(state, SimAirLoopsFlag,
                             SimZoneEquipmentFlag,
                             SimNonZoneEquipmentFlag,
                             SimPlantLoopsFlag,
                             SimElecCircuitsFlag,
                             FirstHVACIteration,
                             SimWithPlantFlowUnlocked);

The change is to combine all bool arguments as one. If any one is true, all zone equipment, airloop and plant loop have to called to update node and EMS override values, regardless of convergence status in one of loop. Unfortunately, it will slow down. This isssu is not a bug and becomes a new feature.

mjwitte: I agree, based on the discussions in https://github.com/NREL/EnergyPlus/issues/5788 if the user set EMS flow >0, it should happen.

mjwitte: @lgu1234 I don't think you need to mess with those flags. Just changing the logic within the ideal loads function should be adequate. Other equipment with EMS overrides follow the same pattern, I think.

lgu1234: @mjwitte If I don't skip EMS override, there are no changes in EMS outputs. Anyway, I an going to make changes in the ideal loads module only based on comments.

mjwitte: No I'm confused. My expectation for this was to simply move the existing EMSoverride code to a place where it won't be skipped if there is no load. Maybe that isn't easy given how the function is structured, but it should be possible. Perhaps there needs to be a duplicate EMSOverride check is the system falls into the "off" block?

lgu1234: @mjwitte Let me explain what I found during debug in HVACManager

The EMS is called first 
ManageEMS(emsCallFromHVACIterationLoop, anyEMSRan); // calling point id

Then SimSelectedEquipment is called with a set of bool arguments

    SimSelectedEquipment(state, SimAirLoopsFlag, SimZoneEquipmentFlag,.....)

If revision occurs in IdealLoads module only as a part of ZoneEquipment. When SimZoneEquipmentFlag (an argument of SimSelectedEquipment) is false, ZoneEquipment execution is bypassed. That means any EMS overrides in IdealLoads will be bypassed. The override values from EMSManager will remain as is. 

What I have done is that I introduce bypassfactor to skip EMSManager override and keep override values in IdealLoads as final values.

mjwitte: We don't worry about that for any other components. If `SimZoneEquipmentFlag` is false, then nothing should be changing at that point in the zone equipment. The equipment should always get called at least once.

lgu1234:
Result differences occur in 2 example files as expected, because node values at EMS actuators are overridden by EMS during system off.

nrel-bot-3:
@lgu1234 @lgentile it has been 28 days since this pull request was last updated.

Myoldmopar:
Pushed up with develop to get some fresh results.

mjwitte (commented):
@lgu1234 This cleans up the reported values, but it doesn't fix the original problem in #5788. If the unit is scheduled off, then this code is relevant. But if the unit is scheduled on, the EMS overrides should be applied even if the unit thinks there is no cooling or heating load. It think the mass flow override should happen here, before `if (SupplyMassFlowRate > 0.0)`.

lgu1234:
@mjwitte Here is my understanding. A section of code is listed below between line 2581 and 2591 in PurchasedAirManager module:

            // EMS override point  Purch air supply temp and humidty ratio ..... but only if unit is on, SupplyMassFlowRate>0.0
            if ((PurchAir(PurchAirNum).EMSOverrideSupplyTempOn) && (SupplyMassFlowRate > 0.0)) {
                SupplyTemp = PurchAir(PurchAirNum).EMSValueSupplyTemp;
            }
            if ((PurchAir(PurchAirNum).EMSOverrideSupplyHumRatOn) && (SupplyMassFlowRate > 0.0)) {
                SupplyHumRat = PurchAir(PurchAirNum).EMSValueSupplyHumRat;
            }

            if (SupplyMassFlowRate > 0.0) {
                PurchAir(PurchAirNum).FinalMixedAirTemp = MixedAirTemp;
                PurchAir(PurchAirNum).FinalMixedAirHumRat = MixedAirHumRat;
.....

Do you think I need to insert a section of code under "if (SupplyMassFlowRate > 0.0)"

            if (SupplyMassFlowRate > 0.0) {
                if (PurchAir(PurchAirNum).EMSOverrideMdotOn)) {
                    SupplyMassFlowRate = PurchAir(PurchAirNum).EMSValueMassFlowRate;
                 }
                PurchAir(PurchAirNum).FinalMixedAirTemp = MixedAirTemp;
                PurchAir(PurchAirNum).FinalMixedAirHumRat = MixedAirHumRat;

Please let me know what you think.

lgu1234:
@mjwitte Or, I can make changes as below. The added section is highlighted.

Capture1

mjwitte:
@lgu1234 My point is that `SupplyMassFlowRate > 0.0` should *not* be the condition for applying the EMS overrides. It should only be a test on the PurchAir schedule.

mjwitte:
Actually, I think it you just remove `&& (SupplyMassFlowRate > 0.0)` from all of the EMS override `if`s, it will work as expected, because I think they are already inside the large `if(UnitOn)` block. Based on a quick look on github, so I may be missing a few.

lgu1234:
@mjwitte I changes two if statement by removing" && (SupplyMassFlowRaete > 0.0)" for EMS mass flow rate override and ran 4 test files before and after the change:

EMSConstantVolumePurchasedAir.idf
      This is an example file with differences

The rest of 3 files are for test. 

Differences:
5788-1.idf
gebtyp_B-MJW-ChangeCallingPoint2.idf

No differences
VAVplusIdealLoads-FixSaturationWarnings2.idf

Enclosed file contains differences.
[eprun_DIFF.CSV.TXT](https://github.com/NREL/EnergyPlus/files/5219421/eprun_DIFF.CSV.TXT)

mitchute:
@mjwitte @Myoldmopar @EnergyArchmage 

Any more comments on this before it goes in?

mitchute:
@lgu1234 I've plotted the time series data on EMSConstantVolumePurchasedAir and the results are looking strange. Any thoughts?

Var_20

Var_23

Var_58

mjwitte:
@mitchute Possibly, I haven't reviewed the most recent changes.

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2713,6 +2726,13 @@ namespace PurchasedAirManager {
                 Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
             }
 
+            if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
+                PurchAir(PurchAirNum).EMSValueMassFlowRate = 0.0;
```
EnergyArchmage: this looks odd, only EMS programs should be setting such a value

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2689,6 +2695,13 @@ namespace PurchasedAirManager {
                 if (Contaminant.GenericContamSimulation) {
                     Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
                 }
+                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
+                    PurchAir(PurchAirNum).EMSValueMassFlowRate = 0.0;
```
EnergyArchmage: this looks odd, EMS programs would be setting the value here.

src/EnergyPlus/PurchasedAirManager.cc
lgu1234: @EnergyArchmage What do you suggest? My intent was to set the value=0, since no cooling and heating.

src/EnergyPlus/PurchasedAirManager.cc
lgu1234: @EnergyArchmage What do you suggest? My intent was to set the value=0, since no cooling and heating.

EnergyArchmage: The intent with EMS is let the user's program decide what to do.  This will override whatever the user may be trying to do. The flow should be set to whatever value the user has set in EMSValueMassFlowRate regardless of zone loads.

lgu1234: @EnergyArchmage Are you OK with the following code with zero SupplyMassFlowRate?

                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
                    SupplyMassFlowRate = PurchAir(PurchAirNum).EMSValueMassFlowRate;
                }

EnergyArchmage:
Yes, that is what I would expect. Two places.

On September 18, 2020 11:07:53 AM EDT, Lixing Gu <notifications@github.com> wrote:
>@lgu1234 commented on this pull request.
>
>
>
>> @@ -2689,6 +2695,13 @@ namespace PurchasedAirManager {
>                 if (Contaminant.GenericContamSimulation) {
>               Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
>                 }
>+                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
>+                    PurchAir(PurchAirNum).EMSValueMassFlowRate = 0.0;
>
>@EnergyArchmage Are you OK with the following code with zero
>SupplyMassFlowRate?
>
>                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
>       SupplyMassFlowRate = PurchAir(PurchAirNum).EMSValueMassFlowRate;
>                }
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/NREL/EnergyPlus/pull/8109#discussion_r491013939

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

src/EnergyPlus/PurchasedAirManager.cc
mjwitte: @EnergyArchmage A related question is: If the unit is scheduled off, should the EMSOverride be accepted or ignored?

EnergyArchmage:
Re availability schedule, it could go either way. I would lean towards allowing the EMS override to dominate but an EMS user could override the schedule as well so it is not limiting either way.

On September 18, 2020 12:02:27 PM EDT, "Michael J. Witte" <notifications@github.com> wrote:
>@mjwitte commented on this pull request.
>
>
>
>> @@ -2689,6 +2695,13 @@ namespace PurchasedAirManager {
>                 if (Contaminant.GenericContamSimulation) {
>               Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
>                 }
>+                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
>+                    PurchAir(PurchAirNum).EMSValueMassFlowRate = 0.0;
>
>@EnergyArchmage A related question is: If the unit is scheduled off,
>should the EMSOverride be accepted or ignored?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/NREL/EnergyPlus/pull/8109#discussion_r491047021

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mitchute:
It looks like this is about ready, assuming CI doesn't complain.

Here's some results from the original defect file:

Var_0

Var_3

The EMS variables are being reset when there's no flow, which is what we're looking for. EMSConstantVolumePurchasedAir still looks like it does in the figures above but I'm going to assume that the reason for those diffs is that EMS is now behaving as it should.

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2689,6 +2695,13 @@ namespace PurchasedAirManager {
                 if (Contaminant.GenericContamSimulation) {
                     Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
                 }
+                if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
+                    SupplyMassFlowRate = PurchAir(PurchAirNum).EMSValueMassFlowRate;
+                    OAMassFlowRate = 0.0;
+                }
+                if (PurchAir(PurchAirNum).EMSOverrideSupplyTempOn) {
+                    SupplyTemp = PurchAir(PurchAirNum).EMSValueSupplyTemp;
+                }
```
mjwitte: Why add these here? The prior EMS mass flow overrides (lines 2086, 2203 and 2454) have already done their job. At this point, if `SupplyMassFlowRate = 0.0` then it should stay that way. 

Same for the supply temp and humrat which are already overridden at lines 2582 and 2585 - but those lines need to have the ` && (SupplyMassFlowRate > 0.0)` removed.

And in all places where the  `&& (SupplyMassFlowRate > 0.0)` is removed, the comment lines above the if need to be updated.

All of these are inside `UnitOn` which is set by the availability schedule and availability managers. That's ok, but the EMS actuator documentation should be updated to clarify this.

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2713,6 +2726,13 @@ namespace PurchasedAirManager {
                 Node(InNodeNum).GenContam = Node(ZoneNodeNum).GenContam;
             }
 
+            if (PurchAir(PurchAirNum).EMSOverrideMdotOn) {
+                SupplyMassFlowRate = PurchAir(PurchAirNum).EMSValueMassFlowRate;
+            }
+            if (PurchAir(PurchAirNum).EMSOverrideSupplyTempOn) {
+                SupplyTemp = PurchAir(PurchAirNum).EMSValueSupplyTemp;
+            }
+
```
mjwitte: Same comment here - these lines should be removed.

mjwitte (changes requested):
@lgu1234 This is close, but still not quite right.

src/EnergyPlus/PurchasedAirManager.cc
lgu1234: Removed && (SupplyMassFlowRate > 0.0) at lines 2582 and 2585.
Removed lines between 2698 and 2704
I am going to add description in EMS documentation.

src/EnergyPlus/PurchasedAirManager.cc
lgu1234: Removed

mjwitte:
@mitchute The example files EMSConstantVolumePurchasedAir and PythonPluginConstantVolumePurchasedAir don't make much sense now with this change. I'm working on revisions to these.

mjwitte:
@lgu1234 @mitchute 
Here's the before/after zone temps for EMSConstantVolumePurchasedAir.
image
image
image

mjwitte (approved):
@lgu1234 @mitchute I'm satisfied with these changes now. We'll see if you all agree.

lgu1234:
@mjwitte I am happy with your satisfaction. Thanks.

Myoldmopar:
I support this, and thanks for fixing up the IDFs.  I'll go ahead and merge this in.

PR #8121, opened 2020-06-29
Add new performance precision tradeoffs modes, functionality, and reports, TeX document updates

jcyuan2020:
Added New PerformancePrecisionTradeoffs modes and expanded functionality for the PerformancePrecisionTradeoffs module: 
 - Added two new PerformancePrecisionTradeoffs modes;
 - Expand functionality (include one more manually override variable) for the Advanced mode;
 - Added two new _perflog.csv output variables;
 - Updated the PerformancePrecisionTradeoffs modes in the Input-Output reference TeX document;
 - Fixed some typesetting problems in the original TeX document.

### Pull Request Author

### Reviewer
 - [x] If feature, test running new feature, try creative ways to break it
 - [x] Verify IDF naming conventions and styles, memos and notes and defaults


src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -6554,6 +6557,8 @@ namespace HeatBalanceSurfaceManager {
                 }
             } // ...end of loop to check for convergence
 
+            MaxAllowedDelTemp = DataConvergParams::MaxAllowedDelTempOverrideValue;
```
mjwitte: @jcyuan2020 Let's not reset this every time `CalcHeatBalanceInsideSurf2` is called. In my other branch, I'm going to move `MaxAllowedDelTemp` up into `DataConvergParams` so performance override can act on it directly.  This will create a conflict depending on which PR merges first.

jcyuan2020: @mjwitte Thank you for the comments! We should definitely synchronize on this. I had originally considered a couple of options to make the variable `MaxAllowedDelTemp` global. One is to move it up to `EnergyPlus::DataConvergParams` as you indicated. Anyway, this called for a need to `#include <EnergyPlus/DataConvergParams.hh>` in `HeatBalanceSurfaceManager.cc` (as the current addition did in Line 130 in HeatBalanceSurfaceManager):
```
#include <EnergyPlus/DataConvergParams.hh>
```

The other neater way is to move `MaxAllowedDelTemp` into `EnergyPlus::DataHeatBalance`. This way could be much neater because: 1) it does not need the extra `#include <EnergyPlus/DataConvergParams.hh>` to be aded into `HeatBalanceSurfaceManager.cc`; 2) also, the `EnergyPlus/DataHeatBalance.hh` header file is already included in present `SimulationManager.cc`, so the Override mode can operate on this variable directly in `SimulationManager.cc` if `MaxAllowedDelTem` is in the `DataHeatBalance` namespace; 3) further, by putting `MaxAllowedDelTemp` into `EnergyPlus::DataHeatBalance`, it is actually then dealt by the same way as the `MaxAllowedDelTempCondFD` variable, which was already in `DataHeatBalance` prior to any changes (See the following code in Line 6562-6565 in `HeatBalanceSurfaceManager.cc`): 
```
if (!DataHeatBalance::AnyCondFD) {
                if (MaxDelTemp <= MaxAllowedDelTemp) Converged = true;
            } else {
                if (MaxDelTemp <= MaxAllowedDelTempCondFD) Converged = true;
```
What do you think? If you think the latter way can also work in your branch, we can synchronize this part by trying this option out as well.

mjwitte: OK, I'll put it in DataHeatBalance.

jcyuan2020: @mjwitte  The latter method discussed above is now implemented and committed. New commits were pushed, with old test unit fixed as well.

jcyuan2020:
The few failed tests above are related to the .eio change in the PerformancePrecisionTradeoffs section brought by this pull request. They have nothing with code errors and a new format adoption should be able solve the problem. 

Here is the related section in the eio.diff: 
```
-! <Performance Precision Tradeoffs>, Use Coil Direct Simulation, Zone Radiant Exchange Algorithm, Override Mode, Number of Timestep In Hour, Force Euler Method, Minimum Number of Warmup Days, Force Suppress All Begin Environment Resets, MaxZoneTempDiff
- Performance Precision Tradeoffs, No, ScriptF, Normal, 6, No, 1, No, 0.300
+! <Performance Precision Tradeoffs>, Use Coil Direct Simulation, Zone Radiant Exchange Algorithm, Override Mode, Number of Timestep In Hour, Force Euler Method, Minimum Number of Warmup Days, Force Suppress All Begin Environment Resets, Minimum System Timestep, MaxZoneTempDiff, MaxAllowedDelTemp
+ Performance Precision Tradeoffs, No, ScriptF, Normal, 6, No, 1, No, 1.0, 0.300, 2.0000E-003
```

JasonGlazer:
@jcyuan2020 is this ready for final review and merging?

jcyuan2020:
> 
> 
> @jcyuan2020 is this ready for final review and merging?

@JasonGlazer Yes, it is.

JasonGlazer:
Everything looks ready to merge but I still have to update the Transition program because the field value meanings have changed.

JasonGlazer:
@jcyuan2020 I just merged the additional PerformancePrecision modes pull request into EnergyPlus. Thanks for your work.

jcyuan2020:
@JasonGlazer Thank you!

PR #8130, opened 2020-07-02
Solar shading test cleanup

mjwitte:
Pull request overview
---------------------
 - Remove all differences between SolarShadingTest.idf and SolarShadingTestGPU.idf except the ShadowCalculation object.
 - Delete all SolarShadingTest* rvi files.

### Pull Request Author

### Reviewer


Myoldmopar:
It will be interesting to see what CI shows, since the CSV files will be all sorts of different.

mjwitte:
For SolarShadingTest it will choke and show no eso diffs, because the new variable is timestep and everything else is hourly or daily, So, the diff file will say
`Time series in <...\SolarShadingTest\eplusout.csv> and <...\SolarShadingTest\eplusout.csv> do not match` We may want to change that in both files to show everything at timestep?

testfiles/SolarShadingTestGPU.idf
```diff
@@ -2211,7 +2211,7 @@
     Fraction,                !- Schedule Type Limits Name
     Through: 12/31,          !- Field 1
     For: AllDays,            !- Field 2
-    Until: 24:00,0.0;!0.2;        !- Field 3
```
mjwitte, line 2214: @mbadams5 Looks like this schedule has no effect with PixelCounting. Comparing the GPU file with the schedule as zero and 0.2 produces no change. The PolygonClipping file does show a difference with the schedule change.

Myoldmopar:
Shading transmittance diffs due to the schedule change, some renumbering/reordering, and difference in audit due to report variable request differences.  Custom check failed in develop, and is now fixed there, so that's nothing to do with this.  I think this is good to go.  Should I dig deeper before merging @mbadams5 @mjwitte ?

mjwitte:
@Myoldmopar You saw diffs from the schedule change? I did not locally on windows. Oh, wait, that's because CI reverts to PolygonClipping - no GPU. I'm fine with this as-is, unless you think we want to change the reporting frequencies to all be timestep.

Issue #5846, opened 2016-08-31
OA controller economizer schedule and min OA flow autosizing


mjwitte:
## Issue overview
1. In Controller:OutdoorAir, in PR #5794, it was originally proposed that the "Time of Day Economizer Control Schedule Name" be changed to allow any economizer type to be schedule off.  Currently, this schedule only adds TOD economizer operation on top of any other specified economizer type.  Given that several example files have a base economizer plus TOD operation, it is now proposed that a new economizer availability schedule field be added to allow the base economizer to be scheduled off as desired.
2. In Controller:OutdoorAir, the "Minimum Outdoor Air Flow Rate" field can be autosized.  It will autosize to whatever OA flow rate flows down from the relevant Sizing:System and/or Sizing:Zone objects.  The typical default usage is to set up a DesignSpecification:OutdoorAir object for each zone type with some flow/person and flow/area specified, using the "Sum" method and then reference this in Sizing:Zone.  This will autosize the "Minimum Outdoor Air Flow Rate" using the design occupancy and the zone area.  If the user then specifies a Controller:MechanicalVentilation object with DCV or an IAQ method, the autosized Minimum Outdoor Air Flow Rate value would then silently (without warning) prevent any OA flow rate less than the area flow plus per-person flow at design occupancy, thus limiting the effectiveness of the DCV control (likely preventing any DCV operation at all).  Options for improving this:
   - Check the final value of Minimum Outdoor Air Flow Rate (autosized or user-specified) against the Controller:MechanicalVentilation inputs and issue a warning if the Minimum Outdoor Air Flow Rate is too high.  But what is too high?
*New warning added in #8136*
   - Add more options to DesignSpecification:OutdoorAir or to allow different specs for the minimum OA flow rate in this situation.
   - Add more columns to the outdoor air summary table report to indicate how much OA flow would have been required by DCV vs "standard" OA flow vs economizer operation vs final. 
*Note that this will happen as part of the OA Summary Report enhancements in #7629.*
   - Other thoughts?
### Details

Some additional details for this issue (if relevant):
- Version of EnergyPlus (v8.6.0)
### Checklist

Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
- [ ] Defect file added (list location of defect file here)
- [ ] Ticket added to Pivotal for defect (development team task)
- [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

lymereJ:
[...]

PR #8136, opened 2020-07-07
Provide a recursive warning when Minimum Outdoor Air Flow Rate Overrides the Controller:MechanicalVentilation OA request

lgu1234:
Pull request overview
---------------------
Fixes one part of #5846.

 This fix address one of comments to provide a recursive warning when OutAirMinFrac > MechVentOutsideAirMinFrac :

> 2.1 Check the final value of Minimum Outdoor Air Flow Rate (autosized or user-specified) against the Controller:MechanicalVentilation OA flow request, and issue a warning if the Minimum Outdoor Air Flow Rate is greater.

At every timestep, if the Controller:MechanicalVentilation request is > 0 but it is less than the product of Minimum Outdoor Air Flow Rate and the current value of Minimum Outdoor Air Schedule, a warning will be issued. This condition typically indicates unintended inputs causing DCV or VRP outdoor air controls to be overridden. It can be caused by setting Minimum Outdoor Air Flow Rate = autosize.

### Pull Request Author

### Reviewer
 - [ ] I~f new idf included, locally check the err file and other outputs~


lgu1234:
"356 tests had: ERR diffs" are from the added recursive warning, as expected.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3674,6 +3674,24 @@ namespace MixedAir {
         }
         this->MechVentOAMassFlowRequest = MechVentOAMassFlow;
         //****** use greater of Mechanical Ventilation Outside Air fraction and OutAirMinFrac
+        if (OutAirMinFrac > MechVentOutsideAirMinFrac) {
+            if (!WarmupFlag) {
+                if (CountMechVentFrac == 0) {
+                    ++CountMechVentFrac;
```
mjwitte: `this->CountMechVentFrac` for clarity.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3674,6 +3674,24 @@ namespace MixedAir {
...
+                    ShowWarningError("Min OA fraction > Mechanical vetilation min OA fraction for OA controller=" + this->Name + ", Min OA fraction is used");
+                    ShowWarningError("Therefore, Mechanical vetilation is not effective as desired.");
+                    ShowContinueErrorTimeStamp("Min OA fraction = " + RoundSigDigits(OutAirMinFrac, 4) +
+                                               ", Min Mech Vent OA fraction = " + RoundSigDigits(MechVentOutsideAirMinFrac, 4));
+                } else {
+                    ++CountMechVentFrac;
```
mjwitte: No need to continue incrementing `CountMechVentFrac` here.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3674,6 +3674,24 @@ namespace MixedAir {
...
         this->MechVentOAMassFlowRequest = MechVentOAMassFlow;
         //****** use greater of Mechanical Ventilation Outside Air fraction and OutAirMinFrac
+        if (OutAirMinFrac > MechVentOutsideAirMinFrac) {
+            if (!WarmupFlag) {
+                if (CountMechVentFrac == 0) {
+                    ++CountMechVentFrac;
+                    ShowWarningError("Min OA fraction > Mechanical vetilation min OA fraction for OA controller=" + this->Name + ", Min OA fraction is used");
+                    ShowWarningError("Therefore, Mechanical vetilation is not effective as desired.");
+                    ShowContinueErrorTimeStamp("Min OA fraction = " + RoundSigDigits(OutAirMinFrac, 4) +
+                                               ", Min Mech Vent OA fraction = " + RoundSigDigits(MechVentOutsideAirMinFrac, 4));
```
mjwitte: How about this error text:
```
                    ShowWarningError(RoutineName+"Minimum OA fraction > Mechanical Ventilation Controller request for Controller:OutdoorAir=" + this->Name + ", Min OA fraction is used.");
                    ShowContinueError("This may be overriding desired ventilation controls. Check inputs for Minimum Outdoor Air Flow Rate, Minimum Outdoor Air Schedule Name and Controller:MechanicalVentilation");
                    ShowContinueErrorTimeStamp("Minimum OA fraction = " + RoundSigDigits(OutAirMinFrac, 4) +", Mech Vent OA fraction = " + RoundSigDigits(MechVentOutsideAirMinFrac, 4));
```

src/EnergyPlus/MixedAir.cc
```diff
@@ -3674,6 +3674,24 @@ namespace MixedAir {
...
+                } else {
+                    ++CountMechVentFrac;
+                    ShowRecurringWarningErrorAtEnd(this->Name +
+                                                       "\": Min OA fraction > Mechanical vetilation min OA fraction, continues...",
```
mjwitte: vetilation min OA fraction --> ventilaton OA fraction

tst/EnergyPlus/unit/MixedAir.unit.cc
```diff
@@ -5670,6 +5674,20 @@ TEST_F(EnergyPlusFixture, CO2ControlDesignOARateTest)
     EXPECT_NEAR(0.003183055786, OAController(1).OAMassFlow, 0.00001);
     EXPECT_NEAR(0.001560321463, OAController(1).MinOAFracLimit, 0.00001);
 
+    // #5846
+    OAController(1).MinOAMassFlowRate = 0.05;
```
mjwitte, line 5678: Nice concise addition to existing unit test.

mjwitte (commented):
@lgu1234 Merged in develop, tested locally. Works as advertised. A few cleanup comments.

src/EnergyPlus/MixedAir.cc
lgu1234: @mjwitte Corrected.

src/EnergyPlus/MixedAir.cc
lgu1234: @mjwitte A good catch, Removed.

src/EnergyPlus/MixedAir.cc
lgu1234: @mjwitte Your suggestion is much better than mine. Use yours. Thanks,

src/EnergyPlus/MixedAir.cc
lgu1234: @mjwitte Corrected.

mjwitte:
@lgu1234 Pushed a tiny cleanup to the recurring message. Once CI finished, this should drop in.

src/EnergyPlus/MixedAir.cc
```diff
@@ -3674,6 +3674,25 @@ namespace MixedAir {
         }
         this->MechVentOAMassFlowRequest = MechVentOAMassFlow;
         //****** use greater of Mechanical Ventilation Outside Air fraction and OutAirMinFrac
+        if (OutAirMinFrac > MechVentOutsideAirMinFrac) {
```
mjwitte: @lgu1234 Sorry - I didn't look closely at the err diffs before. The new warning is tripping even for files that don't have a Controller:MechanicalVentilation object. 
This should be `if ((MechVentOutsideAirMinFrac > 0.0) && (OutAirMinFrac > MechVentOutsideAirMinFrac))`. That will also exclude times when the Controller:MechanicalVentilation object is scheduled off.

lgu1234: @mjwitte Thanks for a good catch. I am going to make the change tomorrow.

mjwitte:
@lgu1234 This reveals that 3 test files incorrectly have the Controller:OutdoorAir Minimum Outdoor Air Flow Rate set to autosize in combination with a Controller:MechanicalVentilation object using DCV or VRP control. These files also have a Minimum Outdoor Air Schedule Name specified which is intended to set OA to zero during unoccupied hours. This schedule should be in the Controller:MechanicalVentilation object, not in the Controller:OutdoorAir object. Or perhaps the schedule shouldn't be used at all with DCV or VRP? (i.e. Controller:MechanicalVentilation should always be on, which is what these files currently do.)

This raise the question - should we trap these combinations of inputs during input processing and issue warnings about potential conflicts? But there may be times when this is done correctly, so perhaps the new warning should mention these things? Or we should add some language to the I/O Reference if it's not already there? @rraustad @Myoldmopar ?

lgu1234:
@mjwitte The latest upload decreases the number of files with err diffs from 356 to 3 due to your catch. Results makes more sense. Thanks again.

rraustad:
As long as the correct combinations are known, then yes I agree to trap/warn in GetInput. The new warning does point to the issue. I am seeing things like this in GUI interfaces where the GUI writes out certain objects independent of other objects so some help in that area would pay off. If the min OA schedule is present and shouldn't be then it could be disregarded, or I guess allowed to be used if present (the user may have wanted more OA for some reason) with a warning so users would notice the issue. Not quite sure how to determine the "correct" combinations.

mjwitte:
@rraustad Good points. The new warning points to those specific fields, so this is good except I think we need to fix the three test files that are throwing these warnings. They probably aren't doing much DCV/VRP variation as-is.

mjwitte:
@lgu1234 I think those three files should use 0.0 instead of "autosize" in the Minimum Outdoor Air Flow Rate field, and the Minimum Outdoor Air Schedule should be blank. And that means the MinOA_MotorizedDamper_Sched schedule definition can also be removed.

lgu1234:
@mjwitte I am going to change 3 example files to remove recursive warnings. At the same time, I will take a look at IO Ref to see what I add some tips.

mjwitte:
@Myoldmopar EnergyPlusFixture.AirTerminalSingleDuctMixer_FCU_NightCycleTest fails here on MacOS-10.13; EnergyPlusFixture.PackagedTerminalHP_VSCoils_Sizing fails on MacOS-10.15. Looking forward to hearing what, if any, tests fail on your 10.14. I don't think these failures are due to the changes in this branch, but want to be sure before merging.

Myoldmopar:
Building now...

Myoldmopar:
Fresh release build on Mac, ran all unit tests and zero failures...carry on.

Honestly, a bunch of these random state issues could be eliminated with my clear state branch....

mjwitte (approved):
Thanks @Myoldmopar for the unit test check. This is good to go in.

PR #8143, opened 2020-07-09
Added two weather variables for outputs and calculate sky emissivity and temperature using timestep interpolated inputs

### LINKED ISSUE #7957: Add more options to Site weather variable outputs
Issue overview

Please provide options to allow more fields in the Weather file to be listed as output options in the .rdd file, such as the “Total Cloud Cover” and the “Opaque Cloud Cover”. If successfully implemented, there should be something that looks like the following in the .rdd file after a successful run: 
Output:Variable,*,Site Total Cloud Cover,hourly; !- Zone Average [tenths of sky]
Output:Variable,*,Site Opaque Cloud Cover,hourly; !- Zone Average [tenths of sky]
The issue was discovered during performing ASHRAE Std 140 weather file testing. 

### Details
 - Platform (Windows 10, 64-bit)
 - Version of EnergyPlus 9.3.0

### Checklist

### LINKED ISSUE #8030: Sky emissivity and sky temperature calculation should use timestep-wise inputs instead of hourly ones
Issue overview
--------------
In current calculation, the sky emissivity (and thus sky temperature) takes in a few input parameters such as opaque sky cover, dry bulb temperature, dewpoint temperature, and relative humidity. The calculation was done at hourly level and uses hourly values for input parameters. The sky emissivity and sky temperature are then interpolated after an hourly calculated value is obtained. A more natural way would be use native interpolated input parameters (opaque sky cover, dry bulb temperature, and dew point temperature, and relative humidity) to compute the timestep-wise sky emissivity and sky temperature. A test case was conducted and the two methods can generally yield a difference of 0.15°C to 1.6°C in the computed sky temperatures. 

### Details
 - EnergyPlus 9.3.0

### Checklist

jcyuan2020:
Pull request overview
---------------------
 - Fixed #7957: Added two weather variables from weather file---Total Sky Cover and Opaque Sky Cover---to the output list as possible output variables for users. Now the user can select those two variables to output. 
 - Fixed #8030: A related issue with sky emissivity and sky temperature calculation, which takes a few hourly weather parameters such as sky cover, Dry Bulb Temperature, Dew-point Temperature, and Relative Humidity as inputs and then do the timestep interpolations on esky and sky temperature results. This issue addresses to issue to use timestep interpolated input parameters (sky cover, DBT, DPT, and RH) as inputs to calculate esky and sky temperatures. 

Expected impact

There could be some minor difference brought in due to minor sky temperature changes in the regression run, albeit most can be explained. One of the examples shows a large difference in the table reports, due to the reports sometime picks a maximum value of load to report. In one of the examples, the table reports shows absolute discrepancy in peak sensible heating load, due to two very close values at two times (Nov 12 06:02 vs. Feb 25 06:02, see attached figures). For the reported line is the maximum zone total sensible load value, and the difference between those two are minimal (about 0.1-0.2 W out of ~3500W, or <0.01% relative diff). The original version and the new version have minor differences in exactly which time stamp has the peak total zone sensible load, and in term caused a very large difference in the sub-component comparison (the totals are close, but the detailed components are quite different--and they “staggered” in a sense). The line by line comparison of the two versions are confirmed to report very minimal difference when compared at the same time stamps. The total site and source energy usages calculated by these two version also so minimal difference. 

image
image

vs. 

image
image

### Pull Request Author

### Reviewer


jcyuan2020:
There is one test unit failure reported--EnergyPlusFixture.Test_UnitaryHybridAirConditioner_Unittest (not ours and not related to this push or related issues). However, I debugged this UnitaryHybrid on my PC and the test passed fine without problems. 

The others three regression fails should be the RDD differences if related to the pushed code. It is expected that the code will added two lines in the RDD file: 
```
Zone,Average,Site Beam Solar Radiation Luminous Efficacy [lum/W]
 Zone,Average,Site Daylighting Model Sky Clearness []
 Zone,Average,Site Daylighting Model Sky Brightness []
+Zone,Average,Site Total Sky Cover []
+Zone,Average,Site Opaque Sky Cover []
 Zone,Average,Site Daylight Saving Time Status []
 Zone,Average,Site Day Type Index []
 Zone,Average,Site Mains Water Temperature [C]
```

mjwitte:
@jcyuan2020 Please add some comparison plots here showing the change in sky temperature.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -3657,16 +3683,44 @@ namespace WeatherManager {
...
+                    Real64 InterpRelHumDeci = TomorrowOutRelHum(TS, Hour) * 0.01; 
+                    Real64 InterpIRHoriz = TomorrowHorizIRSky(TS, Hour); 
+
+                    // Sky emissivity now takes interpolated timestep inputs rather than interpolated calcation esky results
+                    ESky = CalcSkyEmissivity(Environment(Envrn).SkyTempModel, InterpOpaqueSkyCover, InterpDryBulb, InterpDewPoint, InterpRelHumDeci);
+                    if (!Environment(Envrn).UseWeatherFileHorizontalIR || InterpIRHoriz >= 9999.0) {
+                        TomorrowHorizIRSky(TS, Hour) = ESky * Sigma * pow_4(InterpDryBulb + TKelvin);
+                    } else {
+                        TomorrowHorizIRSky(TS, Hour) = InterpIRHoriz;
+                    }
```
mjwitte: @jcyuan2020 Can the caclulation of `ESky` be moved inside the if block?
Is `InterpIRHoriz` reliably >=9999 after interpolation?  Let's say, for example, IRHoriz is 100 at hour 1, then 9999 (missing) at hour 2. What is the interpolated value at 1:30? Is it 9999, or something like 5149? Actually, I can see above that it will not be.

jcyuan2020:
@mjwitte  Sky Temperature Comparisons example plots: 

image

image

src/EnergyPlus/WeatherManager.cc
```diff
@@ -3657,16 +3683,44 @@ namespace WeatherManager {
...
+                    TomorrowTotalSkyCover(TS, Hour) = LastHrTotalSkyCover * WtPrevHour + Wthr.TotalSkyCover(Hour) * WtNow;
+                    TomorrowOpaqueSkyCover(TS, Hour) = LastHrOpaqueSkyCover * WtPrevHour + Wthr.OpaqueSkyCover(Hour) * WtNow;
+
+                    Real64 InterpOpaqueSkyCover = TomorrowOpaqueSkyCover(TS, Hour); 
+                    Real64 InterpDryBulb = TomorrowOutDryBulbTemp(TS, Hour); 
+                    Real64 InterpDewPoint = TomorrowOutDewPointTemp(TS, Hour); 
+                    Real64 InterpRelHumDeci = TomorrowOutRelHum(TS, Hour) * 0.01; 
+                    Real64 InterpIRHoriz = TomorrowHorizIRSky(TS, Hour); 
+
+                    // Sky emissivity now takes interpolated timestep inputs rather than interpolated calcation esky results
```
mjwitte, line 3277: This section duplicates calculations done earlier in lines 3516:3535. Are those lines still necessary?

jcyuan2020: Lines 3516:3535 is exactly the original way of calculating hourly values of HorizIRSky and SkyTemp and then interpolate here. Theoretically, yes, with the new method implemented, the lines from 3516 and 3535 can be removed.

src/EnergyPlus/WeatherManager.cc
jcyuan2020: The` Esky` location is the original code structure and logic. But after an examining context, it is possible to have it get into the if block. Only it is a little weird that is another similar if block a few lined down and they are in sync. If the if criterion of these two "ifs" are synchronously changed together in later revision, there seems to be no problem.

jcyuan2020: The `InterpIRHoriz` is a little bit trickier. There are a few issue here. It is used both in the "if" criterion to determine if the weather file contains a valid value (9999); but it also play a role in the body of if-else.
In a sense it might be a rare occurrence that in a weather file it has valid value in one hour then it does not the other. But in that case, the original method would be messed up as well, as it still will interpolate (actually interpolated before here) `IRHoriz` as output.
Ideally if everything here is timestep-wise would be the best. We can also use the hourly `IRHoriz` for the "if" criterion but use the interpolated `IRHoriz` as output (seems this will aligned with the original method's path) for the occurrence of `IRHoriz` in the weather file is invalid (>=9999).

src/EnergyPlus/WeatherManager.cc
jcyuan2020: Lines 3516:3535 is exactly the original way of calculating hourly values of HorizIRSky and SkyTemp and then interpolate here. Theoretically, yes, with the new method implemented, the lines from 3516 and 3535 can be removed.

jcyuan2020: @mjwitte The code has been revised based on the few discussions above. The lines 3516:3535 were kept with some modifications though, since they served as a section of calculating the hourly values and part of them involved in an interpolated value (`HorizIR`, and then for Sky Temp) in one of the `if` branches.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -3490,6 +3511,9 @@ namespace WeatherManager {
                     TomorrowWindSpeed(CurTimeStep, Hour) = WindSpeed;
                     TomorrowWindDir(CurTimeStep, Hour) = WindDir;
                     TomorrowLiquidPrecip(CurTimeStep, Hour) = LiquidPrecip;
+                    TomorrowTotalSkyCover(CurTimeStep, Hour) = TotalSkyCover;
+                    TomorrowOpaqueSkyCover(CurTimeStep, Hour) = OpaqueSkyCover;
+                    HrIRHoriz_Weather(Hour) = IRHoriz;
```
mjwitte: @jcyuan2020 This section still seems unnecessary. It seems everything here from line 3516:3537 should be replaced with `TomorrowHorizIRSky(CurTimeStep, Hour) = IRHoriz;`
`TomorrowSkyTemp` can be filled in later. It's a derived value, not a value incoming from the epw. This could also use a comment like // Missing values of IRHoriz are dealt with later during interpolation

src/EnergyPlus/WeatherManager.cc
mjwitte: Then here, check for `TomorrowHorzIRSky(1, Hour) >= 9999` ?

src/EnergyPlus/WeatherManager.cc
jcyuan2020: @mjwitte Agree

src/EnergyPlus/WeatherManager.cc
jcyuan2020: @mjwitte Agree, except that I think here we should check `TomorrowHorizIRSky(TS, Hour) >= 9999` instead, since `TomorrowHorizIRSky(1, Hour)` would have been override when doing `(<2,3,4...>, Hour)`. This would be well with your suggestion on doing the direct copy for weather on `TomorrowHorizIRSky(CurTimeStep, Hour) = IRHoriz;` on Line 3516.

mjwitte: @jcyuan2020 I see your point, but I don't think you'll find all of the TS values are filled in. The loop on `CurTimeStep <= NumIntervalsPerHour` is based on the incoming epw data timestep. For hourly weather files, `NumIntervalsPerHour` is 1.

mjwitte: So, perhaps a local bool at the top of the hour loop to check if HorizIRSky was missing for this hour?

jcyuan2020: @mjwitte I see. Would a bool flag have the same effect of Line 3516  `HrIRHoriz_Weather(Hour) = IRHoriz;` and then later check on Lines 3703 and 3714 Hr  `HrIRHoriz_Weather(Hour) >= 9999.0`. This was the mechanism added to the most recent revision (about 10 days ago) to address how to check if hourly IRHoriz data is missing for the hour.

mjwitte: Actually, the bool isn't necessary. The TS=1 values are saved in `Wthr`, so just test here on `Wthr.HorizIRSky(Hour)`.

jcyuan2020: @mjwitte I see. This will then invoke again the requirement for keeping Lines 3516-3535. Also this is required due to the `if` condition in Line 3602 `NumOfTimeStepInHour > 1`--if Lines 3516-3535 are removed, then under the situation of using hourly timestep (when `NumOfTimeStepInHour ==1`), the HorizIR and SkyTemp would not be filled and updated correctly since this whole block of interpolation from Line 3602 to Line 3745 would not be executed when `NumOfTimeStepInHour` is equal to 1.

src/EnergyPlus/WeatherManager.cc
jcyuan2020: An update: due to the `if` condition statement on Line 3602 (`NumOfTimeStepInHour > 1`, Lines 3516-3535 should still be kept. Otherwise, for the situation of using hourly timestep (when `NumOfTimeStepInHour ==1`), the `HorizIR` and `SkyTemp` would not be filled and updated correctly since this whole block of interpolation at Lines 3602-3745 would not run when `NumOfTimeStepInHour == 1`.

src/EnergyPlus/WeatherManager.cc
mjwitte: Good point. OK, so I guess there's no way around having both sections. 
If it makes sense to move this to a small function to avoid code duplication, please do that. 
Should the code at 3156 include an `if (NumIntervalsPerHour > 1 ||  NumOfTimeStepInHour == 1)`?
I'll stop commenting here and let you work out a final solution (which may look much like what you had before, sorry).

jcyuan2020: @mjwitte Thank you for the comments and they helped me to understand the problems much better. The code has been revised per the recent discussions above.

jcyuan2020:
Seems that the most recent merge of develop branch affected the test unit compilation. Now I need to take a further look to fix it.

mjwitte:
@Myoldmopar @nmerket I just tagged you to review this. Code-wise it's ready to go.
@jcyuan2020 I just realized that we need to add the new output variables to the I/O Reference.

jcyuan2020:
@mjwitte Will take care of the doc change.

mjwitte:
@Myoldmopar @nmerket Now this is ready for review.

mitchute:
@jcyuan2020 can you pull develop in here again?

jcyuan2020:
> 
> 
> @jcyuan2020 can you pull develop in here again?

@mitchute Just pushed a new commit in. Thanks!

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
```diff
@@ -2006,69 +2010,75 @@ \subsubsection{Outputs}\label{outputs-4-008}
...
+\subsubsection{Site Direct Solar Radiation Rate per Area {[}W/m2{]}}\label{site-direct-solar-radiation-rate-per-area-wm2}
 
-\paragraph{Site Diffuse Solar Radiation Rate per Area {[}W/m2{]}}\label{site-diffuse-solar-radiation-rate-per-area-wm2}
+Site Direct Solar Radiation Rate per Area is amount of solar radiation in \si{W/m^2} received within a \ang{5.7} field of view centered on the sun. This is also known as Beam Solar.
 
-Diffuse solar is the amount of solar radiation in W/m\(^{2}\) received from the sky (excluding the
+\subsubsection{Site Total Sky Cover {[]}}\label{site-total-sky-cover}
 
-solar disk) on a horizontal surface.
+This is the total sky cover in tenths of sky. The value ranges from 0 (clear) to 10 (full cover).
```
mitchute, line 2067: "In tenths of sky"? With values from 0-10? Is this standard practice for this? I'd assume a "fractional" value should range from 0-1. My testing shows that this is incrementing in intervals of eighths, not tenths. Not a super important point, but just noting it.

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
```diff
@@ -2006,69 +2010,75 @@ \subsubsection{Outputs}\label{outputs-4-008}
 
 Note that these data values may be interpolated from ``hour'' points (ref: Weather Data Hourly Interpolation). Most of the data values represent the ``average'' over the reporting resolution period.
 
-\paragraph{Site Outdoor Air Drybulb Temperature {[}C{]}}\label{site-outdoor-air-drybulb-temperature-c}
+\subsubsection{Site Outdoor Air Drybulb Temperature {[}C{]}}\label{site-outdoor-air-drybulb-temperature-c}
```
mitchute, line 2013: Good cleanups here

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
```diff
@@ -1732,7 +1732,7 @@ \subsection{Site:WaterMainsTemperature}\label{sitewatermainstemperature}
 
 Alternatively, the Site:WaterMainsTemperature object can read values from a schedule. This is useful for measured data or when water comes from a source other than buried pipes, e.g., a river or lake.
 
-If there is no Site:WaterMainsTemperature object in the input file, a default constant value of 10 C is assumed.
+If there is no Site:WaterMainsTemperature object in the input file, a default constant value of \SI{10}{\degreeCelsius} is assumed.
```
mitchute, line 1735: ```\SI``` FTW.

mitchute (commented):
One question regarding the output, but no problems otherwise. Please pull develop in again and we'll let CI take one more pass.

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
jcyuan2020: Yes, it is a standard practice in the weather file to use "tenths of sky" as the unit. It is value from 0 to 10 from the weather file source. It does not have to be a whole integer number though (e.g 5.5 should be fine as well).

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
jcyuan2020: Thanks!

doc/input-output-reference/src/overview/group-location-climate-weather-file-access.tex
mitchute: OK... I still think this is a confusing point for the decerning reader who really looks at it. I guess we'll just roll with it. 🙈 🙉

mitchute:
There was a unit test failure on Mac, but it hasn't been present in previous commits. I ran tests on my local Mac in release and debug and the tests all passed. I deleted the CI results anyway so we'll see. Assuming that comes back OK, we'll merge this.

jcyuan2020:
> 
> 
> There was a unit test failure on Mac, but it hasn't been present in previous commits. I ran tests on my local Mac in release and debug and the tests all passed. I deleted the CI results anyway so we'll see. Assuming that comes back OK, we'll merge this.

Is the failed one `UnitaryHybridAirConditioner_Unittest`? It had been showing up a few times before. I had debugged this test on my PC a few times and the test passed fine without problems: 
https://github.com/NREL/EnergyPlus/pull/8143#issuecomment-656821757

mitchute:
I can't remember which one now, and now I've deleted the results... so I guess we'll never know.

mitchute:
@Myoldmopar found the failed test: AirTerminalSingleDuctMixer_SimFCU_ATMInletSideTest

jcyuan2020:
> 
> 
> @Myoldmopar found the failed test: AirTerminalSingleDuctMixer_SimFCU_ATMInletSideTest

I am taking a look.

jcyuan2020:
> 
> 
> > @Myoldmopar found the failed test: AirTerminalSingleDuctMixer_SimFCU_ATMInletSideTest
> 
> I am taking a look.

This unit test passed fine on my PC.

mitchute:
It looks like @nrel-bot-3 is already out for happy hour, so I ran this locally on a different Mac in debug and release and didn't see any test failures. 

@Myoldmopar merge at will.

Myoldmopar:
@mjwitte or @mitchute, have you rigored over the diffs in this branch?  The unit test issue appears to be fine, so it's just the diffs that concern me.  I see a little bit in the top comment, but with this many files, I'm wondering if it is worth more time.

jcyuan2020:
> 
> 
> @mjwitte or @mitchute, have you rigored over the diffs in this branch? The unit test issue appears to be fine, so it's just the diffs that concern me. I see a little bit in the top comment, but with this many files, I'm wondering if it is worth more time.

@Myoldmopar Some information about the diffs. May not be a full explanation of all the diffs, but see if it helps:

For the large number (700+) RDD and AUD differences, they are pretty much all associated with added two variables to the .rdd files. For example, in the .rdd file, the diff would take the form of: 
```
+Output:Variable,*,Site Total Sky Cover,hourly; !- Zone Average []
+Output:Variable,*,Site Opaque Sky Cover,hourly; !- Zone Average []
```
And in the .audit file, the diff could in general take the form of: 
```
- NumOfRVariable=        1043
- NumOfRVariable(Total)=        1043
+ NumOfRVariable=        1045
+ NumOfRVariable(Total)=        1045
```
In the .audit files for different cases, the NumOfRVariable value would be different. But the revised code would always has two more (expected) NumOfRVariable than the original code, due to the added two output variables. 

The above would pretty much eliminate the 700+ of .rdd and .aud diffs. The what are left are a couple cases of big table diffs. In the top of the PR message, an example was given to explain one of such cases after a careful examination. These diffs might take slightly different forms in different cases. But the example was trying to give an ideal of how some of these diffs could be explained.

Myoldmopar:
Right, the audit and rdd and edd files are inconsequential diffs.  I'm more concerned about the 3 files showing big math diffs, 14 files showing small math diffs, and 16 files with EIO diffs.

mitchute:
This looks like it's ready to me. I've looked at RadLoHydrHeatCoolAuto and _5ZoneAirCooled_annual. Both have ESO big diffs. _5ZoneAirCooled_annual has big table diffs. The results look similar for both, but some results for _5ZoneAirCooled_annual are given below.

Var_3
Var_9
Var_15

Zooming in,

Var_3
Var_9
Var_15

the changes are not significant. Should we dig in farther?

Myoldmopar:
Please don't dwell.  It looks fine, just go.

jcyuan2020:
Let me share some comparisons as well (not completely done; I will continue to post). I have been looking at the results for the math diffs and the large diff files. The small math diffs are typically small, for example, in the following cases: 
```
regression.ASHRAE9012016_Hospital_Denver
Max absolute diff: 0.5049591064453125, field: Electricity:Facility [J](Monthly), time: July, relative: 5.0074176298116025e-12			
Max relative diff: 1.0564203386897138e-11, field: NaturalGas:Facility [J](Monthly), time: July, absolute: 0.22652435302734375			
regression.VSHeatPumpWaterToAirWithRHControl
Max absolute diff: 7.840488251531497e-06, field: POND HEAT EXCHANGER:Pond Heat Exchanger Heat Transfer Rate [W](TimeStep), time:  01/21  02:00:00, relative: 4.475925163596036e-10			
Max relative diff: 9.203639466287549e-10, field: SPACE5-1:Zone Air System Sensible Cooling Rate [W](TimeStep), time:  07/21  02:00:00, absolute: 8.731149137020111e-10				
regression.AirflowNetwork_AtticDuct (may need some attention)
Max absolute diff: 0.0005988178544473755, field: ATTIC THERMAL ZONE:AFN Zone Duct Radiation Heat Gain Rate [W](Hourly), time:  07/21  14:00:00, relative: 0.0001403210620482697			
Max relative diff: 252.81092448956315, field: TRUSSES:Surface Inside Face Conduction Heat Transfer Rate per Area [W/m2](Hourly), time:  01/21  17:00:00, absolute: 9.239074311532629e-07		
```
The third one had a very large rel diff because the base is very small (actually close the zero), so small errors would be magnified in the relative percent scale.

jcyuan2020:
In general, here we do expect some small diffs in the heat/cool load or energy since the interpolation would cause some minor change in the sky temperature, as we shown in previous the comment here: 
https://github.com/NREL/EnergyPlus/pull/8143#issuecomment-656893977
image
image

One point is that even though the the differences shown are discernible, they are still very small compared to the difference introduced by using different simulation programs. See the following graph, which, is an example from the ASHRAE 140 project when it compares the sky temperatures obtained on the same weather date file by using different simulation programs (where series `A` to `H` stand for different Energy Simulation programs; and `G` is the one for EnergyPlus). 
image
The inter-simulation-program difference would be much larger compared to the difference shown in this PR where only interpolation method was slightly changed.

dareumnam:
I also built this branch and compared three testfiles's csv outputs with plotting earlier this morning. Even the outputs that shows big absolute diffs look very similar, RadLoHydrHeatCoolAuto, for example. diffs were occurred each zone's Zone Radiant Heating/Cooling Energy, and plotting looks like below.

image
image

image
image

image
image

mitchute:
Where's the dead horse emoji? 💀  We're done here. Merging. Thanks, @jcyuan2020 and all for the reviews.

jcyuan2020:
Thank you all!

PR #8149, opened 2020-07-10
Use correct IDD versions when configuring IDDs in build tree

Myoldmopar:
Another thing got missed when we did the version number updating.  The IDDs that are configured and placed in the build tree were still old numbers.  @mjwitte maybe you noticed this at some point?  Not sure how many others frequently build Fortran, so this may not have been noticed.  Anyway, it's just updating the version numbers in a few cmake commands.

To verify, build with `BUILD_FORTRAN` on and check the Products folder:
- Baseline branch: 
  - You will not find a V9-4-0-Energy+.idd like you would expect
  - You will find a V9-3-0-Energy+.idd file that actually says 9.4 in the header
  - You will find a missing Report Variables 9.3-9.4 file
- Fix branch:
  - You will find a proper V9-4 IDD with 9.4 in the header
  - You will find a proper V9-3 IDD with 9.3 in the header
  - You will find the correct report vars file


mjwitte (approved):
Tested locally. Good to go.

PR #8157, opened 2020-07-15
WeatherManager clean up

nmerket:
Pull request overview
---------------------
This is a clean up pass on WeatherManager.cc and WeatherManager.hh to clean up comments, extern variables that aren't used, using statments, IDE warnings (Clang-Tidy), static variables in functions (by moving to namespace and adding to clear_state where applicable). 

### Pull Request Author
 - [x] ~~If any diffs are expected, author must demonstrate they are justified using plots and descriptions~~
 - [x] ~~If changes fix a defect, the fix should be demonstrated in plots and descriptions~~
 - [x] ~~If any defect files are updated to a more recent version, upload new versions here or on DevSupport~~
 - [x] ~~If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label~~
 - [x] ~~If structural output changes, add to output rules file and add OutputChange label~~
 - [x] ~~If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies~~

### Reviewer


src/EnergyPlus/WeatherManager.cc
```diff
@@ -112,103 +113,31 @@ namespace WeatherManager {
...
     int const GregorianToJulian(1); // JGDate argument for Gregorian to Julian Date conversion
     int const JulianToGregorian(2); // JGDate argument for Julian to Gregorian Date conversion
 
     Real64 const Sigma(5.6697e-8);    // Stefan-Boltzmann constant
-    Real64 const TKelvin(KelvinConv); // conversion from Kelvin to Celsius
 
     static std::string const BlankString;
-    Array1D_string const DaysOfWeek(7, {"SUNDAY", "MONDAY", "TUESDAY", "WEDNESDAY", "THURSDAY", "FRIDAY", "SATURDAY"});
-    std::map<std::string, WeekDay> weekDayLookUp{{"SUNDAY", WeekDay::Sunday},
+    Array1D_string const DaysOfWeek(7, {"SUNDAY", "MONDAY", "TUESDAY", "WEDNESDAY", "THURSDAY", "FRIDAY", "SATURDAY"});  // NOLINT(cert-err58-cpp)
```
nmerket: Clang tidy was complaining about these lines because they could be run before main or something, see [slack discussion](https://energyplusteam.slack.com/archives/C0ET7FAG3/p1594318250057900).

src/EnergyPlus/WeatherManager.cc
```diff
@@ -257,56 +184,55 @@ namespace WeatherManager {
     bool IDFDaylightSaving(false);      // True if a DaylightSaving Time Period is input (IDF files)
     bool DaylightSavingIsActive(false); // True if a DaylightSavingPeriod should be used for Environment
     bool WFAllowsLeapYears(false);      // True if the Weather File (WF) header has "Yes" for Leap Years
-    int WFLeapYearInd(0);               // Indicator for current Weather file "Leap Year", used in DayOfYear calculations and others.
+    // Indicator for current Weather file "Leap Year", used in DayOfYear calculations and others.
```
nmerket: Should have deleted this comment too.
```suggestion
```

src/EnergyPlus/WeatherManager.cc
```diff
@@ -317,82 +243,112 @@ namespace WeatherManager {
...
+    Real64 LastHrLiquidPrecip;
+    Real64 NextHrBeamSolarRad;
+    Real64 NextHrDifSolarRad;
+    Real64 NextHrLiquidPrecip;
+
+    // SetUpDesignDay static vars
+    bool PrintDDHeader;
+
+    // ProcessEPWHeader static vars
+    std::string EPWHeaderTitle;
```
nmerket: Adding variables to namespace that were previously `static` within the function.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -557,6 +507,28 @@ namespace WeatherManager {
...
+        // ReadEPlusWeatherForDay static vars
+        CurDayOfWeek = 1;
+        ReadEPlusWeatherCurTime = 1.0;
+        LastHourSet = false;
+
+        // SetUpDesignDay static vars
+        PrintDDHeader = true;
+
+        // ProcessEPWHeader static vars
+        EPWHeaderTitle = "";
```
nmerket: Resetting previously static variables in `clear_state`.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -2231,32 +2032,32 @@ namespace WeatherManager {
...
             }
 
-            EndYearFlag = false;
-            if (DayOfMonth == EndDayOfMonth(Month)) {
-                EndMonthFlag = true;
-                EndYearFlag = (Month == 12);
+            DataEnvironment::EndYearFlag = false;
+            if (DataEnvironment::DayOfMonth == EndDayOfMonth(DataEnvironment::Month)) {
+                DataEnvironment::EndMonthFlag = true;
+                DataEnvironment::EndYearFlag = (DataEnvironment::Month == 12);
```
nmerket, line 2014: Lots of namespaces where there used to be `using`.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -5725,33 +4993,13 @@ namespace WeatherManager {
...
-        static std::string MonthlyString(",2,Cumulative Days of Simulation[],Month[]  ! When Monthly ");
-        static std::string RunPeriodString(",1,Cumulative Days of Simulation[] ! When Run Period ");
-        static std::string YearlyString(",1,Calendar Year of Simulation[] ! When Annual ");
+        static ObjexxFCL::gio::Fmt const A("(a)");
+        static std::string const EnvironmentString(",5,Environment Title[],Latitude[deg],Longitude[deg],Time Zone[],Elevation[m]");
+        static std::string const TimeStepString(",8,Day of Simulation[],Month[],Day of Month[],DST Indicator[1=yes 0=no],Hour[],StartMinute[],EndMinute[],DayType");
+        static std::string const DailyString(",5,Cumulative Day of Simulation[],Month[],Day of Month[],DST Indicator[1=yes 0=no],DayType  ! When Daily ");
+        static std::string const MonthlyString(",2,Cumulative Days of Simulation[],Month[]  ! When Monthly ");
+        static std::string const RunPeriodString(",1,Cumulative Days of Simulation[] ! When Run Period ");
+        static std::string const YearlyString(",1,Calendar Year of Simulation[] ! When Annual ");
```
nmerket: Made these `const` instead of moving to the namespace because they don't change and aren't needed outside the function.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -10560,7 +9192,7 @@ namespace WeatherManager {
         bool epwFileExists;
         bool epwHasLeapYear(false);
 
-        if (!OADryBulbAverage.OADryBulbWeatherDataProcessed) {
+        if (!this->OADryBulbWeatherDataProcessed) {
```
nmerket, line 8866: This is in a method on the this struct, but there's only ever one of them and it always acts on it. I just made it use the `this` pointer instead of using the only instance of this struct directly.

src/EnergyPlus/WeatherManager.hh
```diff
@@ -67,205 +67,107 @@ namespace EnergyPlus {
...
-
-    extern int const GregorianToJulian; // JGDate argument for Gregorian to Julian Date conversion
-    extern int const JulianToGregorian; // JGDate argument for Julian to Gregorian Date conversion
+    enum class DateType
+    {
+        InvalidDate = -1,
+        MonthDay = 1,
+        NthDayInMonth = 2,
+        LastDayInMonth = 3
+    };
```
nmerket, line 76: Turned these int variables all into `enum`s. It makes the code it affects definitely more readable.

nmerket (commented):
A few notes about what I did here...

Myoldmopar:
This looks good, and CI is *super* happy with the last commit.  It is unfortunate that the PR like, completely, conflicts with the global weather manager refactor.  I will be eating the wasted time over on that side and just re-doing the global refactor change myself.  I think it will be made somewhat easier with your changes, especially with all the enums you added here.

When you feel you are fully complete here, let me know and I'll do final testing/review.

Myoldmopar:
@nmerket what's your feeling on the timing to wrap this up?  There are now at least 2 other fixes that will be affecting weather manager, and I don't want the conflicts to get crazy.

nmerket:
@Myoldmopar Assuming tests come back clean, this is ready to merge as far as I am concerned.

Myoldmopar:
All clean...looks great.  I'll pull develop in to verify the build and also look over the changes, but this seems to be in great shape.

scripts/dev/find_byref_bool_override.py
```diff
@@ -421,7 +421,7 @@
         ],
         # Docstring is explicit
         "ReportWeatherAndTimeInformation": [
-            "PrintEnvrnStamp"
+            "printEnvrnStamp"
```
Myoldmopar, line 424: Just updated this in the approved list to account for case change.

src/EnergyPlus/AirLoopHVACDOAS.cc
```diff
@@ -970,25 +970,25 @@ namespace AirLoopHVACDOAS {
...
                 this->HeatOutTemp = DesDayInput(i).MaxDryBulb;
-                if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType_WetBulb) { // wet bulb temperature
+                if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType::WetBulb) { // wet bulb temperature
                     this->HeatOutHumRat =
                         Psychrometrics::PsyWFnTdbTwbPb(this->HeatOutTemp, DesDayInput(i).HumIndValue, DataEnvironment::StdPressureSeaLevel);
-                } else if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType_DewPoint) { // dewpoint
+                } else if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType::DewPoint) { // dewpoint
                     this->HeatOutHumRat = Psychrometrics::PsyWFnTdpPb(DesDayInput(i).HumIndValue, DataEnvironment::StdPressureSeaLevel);
-                } else if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType_HumRatio) {
+                } else if (DesDayInput(i).HumIndType == WeatherManager::DDHumIndType::HumRatio) {
```
Myoldmopar, line 991: enum classes are always just so nice.

src/EnergyPlus/General.cc
```diff
@@ -2053,7 +2054,7 @@ namespace General {
         }
 
         if (InternalError) {
-            DateType = -1;
+            DateType = WeatherManager::DateType::InvalidDate;
```
Myoldmopar, line 2057: Changes to enum look good.

src/EnergyPlus/General.hh
```diff
@@ -72,6 +72,10 @@ namespace OutputProcessor {
     enum class TimeStepType;
 }
 
+namespace WeatherManager {
+    enum class DateType;
+}
```
Myoldmopar, line 77: I'm guessing this is to avoid a circular issue.  Leads me to think the enum should be pulled into its own little separate namespace.  If there are any changes left to be made, I'd like you to consider that.

src/EnergyPlus/WeatherManager.hh
Myoldmopar: :+1:

src/EnergyPlus/WeatherManager.cc
```diff
@@ -113,102 +113,27 @@ namespace WeatherManager {
...
-
-    Real64 const Sigma(5.6697e-8);    // Stefan-Boltzmann constant
-    Real64 const TKelvin(KelvinConv); // conversion from Kelvin to Celsius
-
-    Array1D_string const DaysOfWeek(7, {"SUNDAY", "MONDAY", "TUESDAY", "WEDNESDAY", "THURSDAY", "FRIDAY", "SATURDAY"});
-    std::map<std::string, WeekDay> weekDayLookUp{{"SUNDAY", WeekDay::Sunday},
+    Real64 const Sigma(5.6697e-8); // Stefan-Boltzmann constant
+
+    Array1D_string const DaysOfWeek(7, {"SUNDAY", "MONDAY", "TUESDAY", "WEDNESDAY", "THURSDAY", "FRIDAY", "SATURDAY"}); // NOLINT(cert-err58-cpp)
+    std::map<std::string, WeekDay> weekDayLookUp{{"SUNDAY", WeekDay::Sunday},                                           // NOLINT(cert-err58-cpp)
```
Myoldmopar, line 120: I'm not sure about muting these warnings.  They are valid, and imply that we are doing something improper.  In this case, I imagine these global variables should really be owned by some class and not just have this lifetime issue.  As we are working toward collecting all the global state and refactoring it, the warnings like this will start going away.  I do like having the cleaner IDE, and I don't think I will request you take out these NOLINT for now, just letting you know they will be cleaned up by better design, in the short term.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -646,27 +566,27 @@ namespace WeatherManager {
                                           DataIPShortCuts::lAlphaFieldBlanks,
                                           DataIPShortCuts::cAlphaFieldNames,
                                           DataIPShortCuts::cNumericFieldNames);
-            underwaterBoundaries.push_back(UnderwaterBoundary());
-            underwaterBoundaries[i - 1].Name = DataIPShortCuts::cAlphaArgs(1);
-            underwaterBoundaries[i - 1].distanceFromLeadingEdge = DataIPShortCuts::rNumericArgs(1);
-            underwaterBoundaries[i - 1].OSCMIndex = UtilityRoutines::FindItemInList(underwaterBoundaries[i - 1].Name, DataSurfaces::OSCM);
-            if (underwaterBoundaries[i - 1].OSCMIndex <= 0) {
+            underwaterBoundaries.emplace_back();
+            auto &underwaterBoundary{underwaterBoundaries[i - 1]};
```
Myoldmopar, line 570: :+1:

src/EnergyPlus/WeatherManager.cc
```diff
@@ -1069,367 +1048,333 @@ namespace WeatherManager {
...
-
-                            bool missingLeap(false); // Defer acting on anything found here until after the other range checks (see below)
-                            if (Environment(Envrn).ActualWeather && !WFAllowsLeapYears) {
-                                for (int year = Environment(Envrn).StartYear; year <= Environment(Envrn).EndYear; year++) {
-                                    if (isLeapYear(year)) {
-                                        ShowSevereError(RoutineName + "Weatherfile does not support leap years but runperiod includes a leap year (" +
-                                                        std::to_string(year) + ")");
-                                        missingLeap = true;
-                                    }
+                    if ((DataGlobals::KindOfSim == DataGlobals::ksRunPeriodWeather) || (DataGlobals::KindOfSim == DataGlobals::ksRunPeriodDesign)) {
```
Myoldmopar, line 1051: Getting rid of the SELECT_CASE_vars is nice.

src/EnergyPlus/WeatherManager.cc
Myoldmopar: :+1:

src/EnergyPlus/WeatherManager.cc
```diff
@@ -4426,44 +3927,46 @@ namespace WeatherManager {
...
-                ConstantHumidityRatio = false;
+        DataGlobals::CurrentTime = 25.0;
+        Real64 HumidityRatio; // Humidity Ratio -- when constant for day
+        bool ConstantHumidityRatio;
 
-            } else {
-                ShowSevereError("SetUpDesignDay: Invalid Humidity Indicator type");
-                ShowContinueError("Occurred in Design Day=" + DesDayInput(EnvrnNum).Title);
-            }
+        switch (DesDayInput(EnvrnNum).HumIndType) {
```
Myoldmopar, line 3934: Switch blocks :+1:

src/EnergyPlus/WeatherManager.cc
```diff
@@ -4991,60 +4432,29 @@ namespace WeatherManager {
...
-
-        X = DayCorrection * DayOfYear; // Convert Julian date (Day of Year) to angle X
+        // Day of Year in Radians (Computed from Input DayOfYear)
+        Real64 X = DayCorrection * DayOfYear; // Convert Julian date (Day of Year) to angle X
 
         // Calculate sines and cosines of X
-        SinX = std::sin(X);
-        CosX = std::cos(X);
+        Real64 SinX = std::sin(X);
+        Real64 CosX = std::cos(X);
```
Myoldmopar, line 4457: Rescoped variables :+1:

src/EnergyPlus/WeatherManager.cc
```diff
@@ -6946,54 +6075,33 @@ namespace WeatherManager {
...
+        DDHumIndModifier.allocate(DataGlobals::NumOfTimeStepInHour, 24, TotDesDays);
         DDHumIndModifier = 0.0;
-        DDBeamSolarValues.allocate(NumOfTimeStepInHour, 24, TotDesDays);
+        DDBeamSolarValues.allocate(DataGlobals::NumOfTimeStepInHour, 24, TotDesDays);
         DDBeamSolarValues = 0.0;
-        DDDiffuseSolarValues.allocate(NumOfTimeStepInHour, 24, TotDesDays);
+        DDDiffuseSolarValues.allocate(DataGlobals::NumOfTimeStepInHour, 24, TotDesDays);
         DDDiffuseSolarValues = 0.0;
-        DDSkyTempScheduleValues.allocate(NumOfTimeStepInHour, 24, TotDesDays);
+        DDSkyTempScheduleValues.allocate(DataGlobals::NumOfTimeStepInHour, 24, TotDesDays);
```
Myoldmopar, line 6104: It definitely feels like these should be a singly allocated struct rather than individual arrays.  There are certainly parts of the code where we iterate on objects of an array in a tight loop, and in those cases, we benefit from having individual arrays of values.  But that is definitely not the case here.  It would be easier to maintain if these were just member variables of a struct which was allocated one time.  I still don't think it's necessary to add more changes to do that here since it appears this is done, but just forward thinking.

src/EnergyPlus/WeatherManager.cc
```diff
@@ -9579,235 +8421,66 @@ namespace WeatherManager {
...
-        //  ENDIF
-        if (Missed.SnowDepth > 0) {
-            if (!MissedHeader) {
-                ShowWarningError(MissString);
-                MissedHeader = true;
-            }
-            ShowMessage(format(msFmt, "\"Snow Depth\"", Missed.SnowDepth));
-        }
+        bool MissedHeader = false;
+        auto missedHeaderCheck{[&](Real64 const value, std::string const &description) {
```
Myoldmopar, line 8434: :+1: :+1:

src/EnergyPlus/WeatherManager.cc
Myoldmopar: OK, this seems fine, I think.  Basically there will only ever be one instance of a AnnualMonthlyDryBulbWeatherData.  And you are calling this method from somewhere and you are just altering the value of the single instance.  But it really doesn't matter, I mean if you had more than one for whatever reason, this would still be the right thing to do.  Right?

Myoldmopar (commented):
This is a really great set of cleanups that will set the weather manager up for further refactoring moving forward.  While there are a few things that could make this even better, I don't think they added up to enough to justify requiring more commits.  That said, please comment back on my notes and we'll resolve everything.  If you do feel like making any of the commented things, that would be cool too, of course.

scripts/dev/find_byref_bool_override.py
nmerket: :+1:

src/EnergyPlus/AirLoopHVACDOAS.cc
nmerket: I know, right?

src/EnergyPlus/General.hh
nmerket: It was. That's probably a good suggestion.

src/EnergyPlus/WeatherManager.cc
nmerket: Sounds good.

src/EnergyPlus/WeatherManager.cc
nmerket: Good point.

src/EnergyPlus/WeatherManager.cc
nmerket: Yes, this would be the right thing to do in that case. Previously it was kind of weird because this method was on the struct/class but instead of operating on the struct it was attached to, operated on the one and only `OADryBulbAverage`. If there ever were more than one, it wouldn't work right.

nmerket:
@Myoldmopar Thanks for the review. I'm going to be out for the rest of the week. I can make the changes next week if needed, or if you want this merged sooner and are happy with it the way it is go ahead and merge it.

Myoldmopar:
The things I noted can be addressed in future work.  This needs to go in so that it doesn't get conflicted with other work.  Thanks @nmerket.

PR #8160, opened 2020-07-17
Fixes to clear_state

Myoldmopar:
Users who are testing out the new EnergyPlus API found that there were problems running files multiple times.  I sorta expected this, given all the static variables that remain inside functions.  So I took the entire test suite and tried to run them twice through the API.  About half ran fine twice, and half failed for a variety of reasons.  Sometimes fatal errors and sometimes seg faults.  Gross.

With this PR, I hope to get every test suite file *running* twice.  The next step will be capturing the output of both runs and checking for diffs.  While I'm sure there will be *some* numeric diffs, I'm actually not as worried about them.  I'm more concerned about table headers, EIO headers, and those sorts of things.  There are a bunch of local static variables that control when to write only headers.

As a part of this work, I spent most of the time searching for static bool variables and pulling them to clear_state functions, adding new clear_state functions where needed, and other general cleanup.  A ton of static bool variables are not actually needed statically, like essentially all the of `static bool ErrorsFound(false)` variables, which just need `static` removed, since they fatal at the end of the function anyway.  Plenty of other ones just needed to be put in a clear_state, and the process was straightforward.

However, there were a couple spots where static variables were being used uninitialized.  It is possible that by pulling them to the namespace, there will be a change in initial value, and then I also re-initialize them in the clear_state, which could cause minor differences.

I ran regressions locally and found really great results.  One file had numeric diffs: EMSThermochromicWindow.  I'm still not sure why this is different, but the ESO diffs are actually **very** small (e-10 on two surface temp variables), tiny EIO warmup diffs, and no meter diffs at all.  I do want to figure out why, but I won't spend too many hours trying to figure out that one if that's the only impact.  More importantly, there are a handful of files with a table diff in the heat rejection annual summary.  It's way off, and I'm trying to track down the culprit.  By debugging through the CalcHeatEmissionReport function, the final sum value for the heat rejection bounces around.  There are 5 files that show this issue.  I ran a quick check to find objects that are common between all the files to try to identify what object could be the problem.  The list includes standard things of course, like Building, BuildingSurface, Curves, Constructions, Materials, Location, Version, Zone, and others, but some likely culprits include:

- Coil:Heating:DX:SingleSpeed
- OutdoorAir:Mixer
- Output:Table:SummaryReports
- OutputControl:Table:Style
- ZoneControl:Thermostat
- ZoneHVAC:EquipmentConnections
- ZoneHVAC:EquipmentList

I have looked over coil related changes and I don't see anything obvious, but I'll keep looking.

As for the progress toward getting all the files running, I checked for updated success rates as I made changes.  Here are the number of files that succeed running twice through the API over "time" (note I was testing 705 total files):

| Current Develop | SQLite disabled | Some clear_state fixes | SQL fixed, more state fixes | Not doing DDOnly | Fixed octree and IP | Complex window fixes | More init fixes | Kiva, unit htr, radiants | More static bools | More static bools |
|-----------------|-----------------|------------------------|-----------------------------|------------------|---------------------|----------------------|-----------------|--------------------------|-------------------|-------------------|
| 365             | 440             | 464                    | 467                         | 476              | 509                 | 515                  | 578             | 613                      | 626               | 638               |

There are also 50 files that are not running because my quick little API script didn't run preprocessors, so those may be success or failure, it is uncertain.  That leaves 17 failing files, although some changes I made just this morning should fix at least two more.

I also checked runtime just to make sure I didn't cause something to run more often, and with a total runtime of the test suite around 1.5 hours, the runtime change was less than 7 seconds.  

I am indeed curious whether CI will back up my regression results.  I will investigate the table issue, and provide a walkthrough of the changes.  Once CI is done I will post any results necessary.

This will be the first step of many:
- Get all test files running serially in the API -- this current PR
- Check for and fix all diffs when running serially in the API -- this could be a big lift
- Clean out any straggling static non-const local variables in the code -- a lot of things can be moved to member variables so that you don't need local arrays of "MyEnvrnFlag", for example.


Myoldmopar:
At this point I'm up to 645 successful runs. 50 files fail each time, which I found is due to the files having hvac template objects, or because they are PythonPlugin files, which doesn't play well with the API.  If this comes back clean, I'm inclined to get it merged in rapidly and start a new branch soon to evaluate and resolve diffs between a first and second API run.

Oh, and I found a file that wasn't needed so I deleted it and the accompanying header.

Oh, and one thing I'm struggling with is a couple of Kiva files that seg fault the second time they are run.  I need to figure out what re-initialization I need to do to get the Kiva manager instance cleared and ready for a new run.  (@nealkruis)

src/EnergyPlus/DaylightingDevices.cc
```diff
@@ -205,6 +205,7 @@ namespace DaylightingDevices {
 
     void clear_state() {
         COSAngle.clear();
+        COSAngle.allocate(NumOfAngles);
```
Myoldmopar, line 208: Needed to reallocate since it was allocated to a specific amount in the declaration.

Myoldmopar (commented):
DataPlantPipingSystems were not needed and as such deleted.  Other than that it's pretty much just more standard initialization fixes.

Myoldmopar:
This is failing unit tests when they are run serially so don't consider merging yet til I figure that out.

Myoldmopar:
@mbadams5 I could use your thoughts on the input processor `clear_state`.  

While working on this, I found many files failing to run twice in the API because of the input processor.  When I commented all the code inside the input processor, the calls ran fine and files ran twice successfully in many cases.  I knew there would be side effects, but things were running smoothly so I deferred addressing the issue.

Now when I run unit tests serially, I see that the error stream is not being cleared in between tests, so at one of the `compare_err_stream` calls, it has a straggling bit of error from the previous test and fails.  I found that if I uncomment the input processor's clear_state code, the unit tests all pass happily.

So I'm a bit stuck...if I leave the input processor's clear state code, EnergyPlus won't run twice in a row in the API.  If I comment the input processor's clear state, the unit tests won't run serially because things aren't cleaned up.  There must be something in the unit test fixture that allows the code to re-run each time.

I can demonstrate on a webmeeting if you'd like to see it in action.

Myoldmopar:
OK, I believe I have a fix in place locally, I'll push it up this afternoon and we'll see.

Myoldmopar:
OK, so this seems to be in really good shape now...I'm only about 100 files into the "run twice" test, but nearly every file is passing.  Thanks to @mbadams5 I have a better understanding of the interaction between the unit test structure and the input processor.  What it really boils down to is that when running for unit tests, we don't want to wipe away the entire inputProcessor object itself, just parts of it.  So there is an instance method on the input processor class that wipes away just those specific things.  However, when we want to run as part of an API, we actually want to recreate the input processor entirely.  So the shared clearAllStates no longer does **anything** with the input processor.  The caller is responsible for either 1) clearing away just specific things about the input processor, or 2) creating an entirely new input processor.  The API and unit test fixture both handle this in their respective ways, the API user or unit test author don't have to do anything extra.  Once the tests pass I'll report back here or I'll push more commits if they are needed.

Myoldmopar:
OK, I re-ran the full serial test and I'm back to 644 test files passing fully while at the same time the unit test suite also passes serially.  The CI results show the one diff in EMSThermochromicWindow plus the spurrious and unrelated GPU file diffs.  I believe this is now completely ready to go, though of course a reviewer can wait on the Linux Integration to finish as needed.  If anyone wants more walkthrough of the changes or anything else, please let me know, otherwise this is ready.

mitchute:
Tests look good and CI is happy. Merging.

PR #8173, opened 2020-07-22
Fix State confusion around API

Myoldmopar:
In the current develop branch, there is some awkwardness around the EnergyPlus "state".  As we pull data from the EnergyPlus global state and start putting more of it onto an actual state data structure, the awkwardness lies in who should own this state object on the API.  It is really quite simple -- any client wanted to execute EnergyPlus should own this state.  In the EXE form, the main function will create a state instance and pass it into the library to run.  In the unit test form, the unit test fixture owns a state instance that then can be passed in and out of the library.  In the API form, we had a hacked fix where an API function could create a local state object and pass that in, but it wasn't really clear where things would get cleared, etc.  It just wasn't great.  This changes it so that the client must own and maintain their instance of this "state".  It's a much more robust design, and although it is a serious breaking change with the API, that is why I left this release cycle available for such major changes.  They will be calmer moving forward.

I'll describe some of the key changes here by showing the difference in making E+ API calls from Python.  When calling EnergyPlus in the current develop branch, you just do essentially the following:
```
from pyenergyplus.api import EnergyPlusAPI
def handler():
    print("New Environment")
api = EnergyPlusAPI()
api.runtime.callback_begin_new_environment(handler)
api.runtime.run_energyplus(sys.argv[1:])
```

With this change, the client code is now:
```
from pyenergyplus.api import EnergyPlusAPI
def handler(_state):
    print("New Environment")
api = EnergyPlusAPI()
state = api.runtime.new_state()
api.runtime.callback_begin_new_environment(state, handler)
api.runtime.run_energyplus(state, sys.argv[1:])
```

The big difference being that the client must call `new_state()` in order to create a new instance of the EnergyPlus state, which they must manage.  Then this state is passed around when function calls are made.  The state object also happens to be called back to the callback functions.  This doesn't have any major value at the moment, but as the API evolves, there may be more things a client could do with that object instance.

Technically I also could've gotten away with not using state as a new argument in the callback registration functions.  Currently those functions just register callbacks in variables held in the EnergyPlus global state, so they don't touch that little state object.  However, as we continue to move global variables into this new state object, at some point we will start storing the callbacks on that state object itself, so we will need it at that point.  This change will be hidden from the users by adding this new argument at this point.  *#futureproof*

This PR should not have any diffs and any CI issues will be considered problematic unless I can't recall something I fixed.  I'll walkthrough the code changes where possible to document special changes.


src/EnergyPlus/EMSManager.cc
```diff
@@ -246,7 +246,8 @@ namespace EMSManager {
 
     // MODULE SUBROUTINES:
 
-    void ManageEMS(int const iCalledFrom,                 // indicates where subroutine was called from, parameters in DataGlobals.
+    void ManageEMS(EnergyPlusData &state,
```
Myoldmopar, line 249: This is the root of most of these changes.  Since we are now passing our callback functions a state instance, then every call to ManageEMS must now have access to the state object.  And ManageEMS is called in many nooks of the code.  So all of these code paths needed the state argument added.  No functional changes for normal EnergyPlus workflows/calculations, just passing state around.

src/EnergyPlus/PluginManager.cc
```diff
@@ -423,11 +424,11 @@ namespace PluginManagement {
         return (int)callbacks.size();
     }
 
-    void runAnyRegisteredCallbacks(int iCalledFrom, bool &anyRan)
+    void runAnyRegisteredCallbacks(EnergyPlusData &state, int iCalledFrom, bool &anyRan)
     {
         if (DataGlobals::KickOffSimulation) return;
         for (auto const &cb : callbacks[iCalledFrom]) {
-            cb();
+            cb((void *) &state);
```
Myoldmopar, line 431: Passing the state to the callback, cast as a void pointer.

src/EnergyPlus/PluginManager.cc
```diff
@@ -402,7 +403,7 @@ namespace PluginManagement {
 #endif
     }
 
-    void registerNewCallback(int iCalledFrom, const std::function<void()> &f)
+    void registerNewCallback(EnergyPlusData &EP_UNUSED(state), int iCalledFrom, const std::function<void(void *)> &f)
```
Myoldmopar, line 406: Right now the callbacks are stored in an EnergyPlus "global" variable, so this state isn't necessary at the moment.  However, at a later date, the callbacks will be stored in this "state" structure, so I just added this arg now to make future changes easier.

Myoldmopar, line 406: Also note here the function form that is passed in as a callback is now: `const std::function<void(void *)> &`, so the callback must be a void function that accepts a void pointer as a single argument.

src/EnergyPlus/StateManagement.cc
```diff
@@ -216,12 +216,11 @@
 #include <EnergyPlus/WaterToAirHeatPumpSimple.hh>
 #include <EnergyPlus/WeatherManager.hh>
 
-void EnergyPlus::clearThisState(EnergyPlusData &state)
+void EnergyPlus::clearAllStates(EnergyPlusData &state)
```
Myoldmopar, line 219: This was a really awkward situation, and it is now much better.  Now the clear all states function must take in a state argument.  The clear all states function will now clear the state of the argument, then it will go through and clear the remaining global states as well.  As the global refactor gets done, the number of "extra" clear_state calls in here will reduce to the point where this is just a call to `state.clear_state()`, and this function can be eliminated.

src/EnergyPlus/api/func.cc
```diff
@@ -89,7 +89,7 @@ Real64 glycolViscosity(Glycol glycol, Real64 temperature) {
 }
 
 Refrigerant refrigerantNew(const char* refrigerantName) {
-    auto refrigerant = new EnergyPlus::FluidProperties::RefrigerantAPI(refrigerantName);
+    auto *refrigerant = new EnergyPlus::FluidProperties::RefrigerantAPI(refrigerantName);
```
Myoldmopar, line 92: @amirroth will love this.  But anyway, Clang Tidy wanted me to use an asterisk here as a way to clarify that I am creating a pointer type object.

src/EnergyPlus/api/runtime.cc
```diff
@@ -53,13 +53,21 @@
 #include <EnergyPlus/StateManagement.hh>
 #include <EnergyPlus/UtilityRoutines.hh>
 
-void cClearAllStates() {
-    EnergyPlus::EnergyPlusData state;   //THIS IS TEMPORARY
-    EnergyPlus::clearThisState(state);
-    EnergyPlus::clearAllStates(state.outputFiles);
+EnergyPlusState stateNew() {
```
Myoldmopar, line 56: The cClearAllStates function is now gone, and it is replaced with a full set of state management functions:
```
stateNew()
stateReset(state)
stateDelete(state)
```

This makes for much more intentional management of the simulation.

src/EnergyPlus/api/runtime.cc
```diff
@@ -83,147 +90,180 @@ void issueText(const char * message) {
     EnergyPlus::ShowContinueError(message);
 }
 
-void registerProgressCallback(void (*f)(int const)) {
+void registerProgressCallback(EnergyPlusState EP_UNUSED(state), void (*f)(int const)) {
+    // auto *this_state = reinterpret_cast<EnergyPlus::EnergyPlusData *>(state);
     EnergyPlus::DataGlobals::progressCallback = f;
```
Myoldmopar, line 95: I did *not* convert over the older progressCallback and messageCallback functions to accept a state object argument.  This would have meant changing DisplayString and other Display* functions to accept a state object argument.  Which means like every single function in the entire codebase would've needed state.  I need to decide the best path here.

src/EnergyPlus/api/runtime.cc
```diff
@@ -83,147 +90,180 @@ void issueText(const char * message) {
...
 }
 
-void registerStdOutCallback(void (*f)(const char * message)) {
+void registerStdOutCallback(EnergyPlusState EP_UNUSED(state), void (*f)(const char * message)) {
     EnergyPlus::DataGlobals::messageCallback = f;
 }
 
-void callbackBeginNewEnvironment(std::function<void ()> f) {
-    EnergyPlus::PluginManagement::registerNewCallback(EnergyPlus::DataGlobals::emsCallFromBeginNewEvironment, f);
+void callbackBeginNewEnvironment(EnergyPlusState state, std::function<void (EnergyPlusState)> const &f) {
```
Myoldmopar, line 102: As mentioned previously, state is now passed into the API when a client is registering a callback, and the callback function itself should now accept a void* argument (the state).

src/EnergyPlus/api/runtime.h
```diff
@@ -51,27 +51,39 @@
...
+ENERGYPLUSLIB_API void callbackInsideSystemIterationLoop(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfZoneTimeStepBeforeZoneReporting(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfZoneTimeStepAfterZoneReporting(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfSystemTimeStepBeforeHVACReporting(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfSystemTimeStepAfterHVACReporting(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfZoneSizing(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfSystemSizing(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackEndOfAfterComponentGetInput(EnergyPlusState state, std::function<void (EnergyPlusState)> f);
+ENERGYPLUSLIB_API void callbackUnitarySystemSizing(EnergyPlusState state, const std::function<void (EnergyPlusState)>& f);
```
Myoldmopar, line 85: The C++ interface for registering functions is also changed to accept state in both the registration functions and the callbacks themselves.

src/EnergyPlus/api/runtime.py
```diff
@@ -31,50 +31,54 @@ def __init__(self, api: cdll):
         self.api.issueText.argtypes = [c_char_p]
         self.api.issueText.restype = c_void_p
         self.py_progress_callback_type = CFUNCTYPE(c_void_p, c_int)
-        self.api.registerProgressCallback.argtypes = [self.py_progress_callback_type]
```
Myoldmopar, line 34: Accompanying Python changes here for different functional forms and state management function changes.

third_party/ObjexxFCL/src/ObjexxFCL/Array1S.hh
```diff
@@ -1111,7 +1111,7 @@ public: // Subscript
 	T const &
 	operator []( size_type const i ) const
 	{
-		assert( i < std::numeric_limits< int >::max() );
+		assert( i < std::numeric_limits< size_type >::max() );
```
Myoldmopar, line 1114: I don't know if anyone else was having a bunch of compiler warnings about this or just me.  `std::numeric_limits<int>::max()` was returning an integer which was being compared to a `size_type`.  I just changed it to look for the max value on size_type.

Myoldmopar (commented):
Walk-through laid out the interesting changes.  Outside of those changes called out, the rest are just passing state around everywhere.

Myoldmopar:
Reviewers: I will make additional documentation changes to accompany these fixes, but I don't want that to hold up this wide-reaching PR.  If CI is happy, consider this done for the moment and I'll be following up with *at least* doc updates, maybe more fixes.

Myoldmopar:
And only diffs in the GPU case, great!

src/EnergyPlus/EMSManager.cc
mitchute: Fewer for @brianlball  and I to do!

src/EnergyPlus/api/runtime.cc
mitchute: For the uninformed (i.e. me), why do we need an explicit ```deleteState()``` method? Doesn't this have to be deleted on the client-side?

mitchute (commented):
I can't say that I fully understand all of the memory management magic you're doing to get data passed back and forth, but the new ```state``` references all look fine. I'll run a few tests, but if CI is green and my tests look OK I'll assume this is ready.

Myoldmopar:
> For the uninformed (i.e. me), why do we need an explicit deleteState() method? Doesn't this have to be deleted on the client-side?

When we are out in the API, we can't just call the C++ constructor for an EnergyPlusData instance.  We need a helper.  So we set up a C method called `stateNew` that will allocate some memory to hold a shiny new instance of an EnergyPlusData.  This is being managed by the C++ side, so then we also need a worker function that will do the casting and deletion of the memory as well.  It is also possible that these "new" and "delete" functions could ultimately do more behind the scenes than just allocate the memory, but for now that's all they do.

mitchute:
> This is being managed by the C++ side

Just making sure I understand things. The actual ```state``` instance is managed by and lives in C++/E+ land, but the thing that gets returned by ```stateNew``` is a... pointer, integer index, ... (?) to that state? ```deleteState``` deletes the actual ```state``` object but the client has to manage the pointer/int-index on its side.

Is that remotely correct?

Myoldmopar:
When you call `stateNew`, EnergyPlus will allocate memory for a new state object and pass that pointer back to the client.  The client can keep the same state for multiple runs by asking E+ to reset the state (along with the remaining global variables) with `stateReset`.  Then when the client is fully done, the client should call `stateDelete` to ask EnergyPlus to release that memory.  At that point, the pointer that was originally given during the call to `stateNew` will point to invalid memory, and if more simulations are to be run, another call to `stateNew` should be made.

mitchute:
I ran tests locally and the CI failures noted above are separate, known issues. Merging.

PR #8175, opened 2020-07-26
Fix the Thermal Comfort CEN 15251 Running Mean Temperature Calculation

### LINKED ISSUE #8174: Zone Thermal Comfort CEN 15251 Adaptive Model Running Average Outdoor Air Temperature Error
Issue overview
--------------
This issue is related to a previous one concerning ASHRAE 55 Adaptive Model (#5381), some similar and different bugs are observed as follows.

Similarity:
1. One more line skipped when reading the weather file;
2. One earlier day used as the starting date;
3. lack of consideration of leap years.

Difference:
1. Wrong equation used for the rest of the days. The current one is `runningAverageCEN = 0.2 * runningAverageCEN + 0.8 * avgDryBulbCEN`, while it should be `runningAverageCEN = 0.8 * runningAverageCEN + 0.2 * avgDryBulbCEN`.

### Details

### Checklist

airallergy:
Pull request overview
---------------------
 - Fixes #8174: to fix bugs in the first-day and following-up calculation of the running mean temperature for the CEN 15251 adaptive model stated in the issue.

 - Calculation validation
**Note:** T_od is calculated directly from the weather file, which is derived from the [ep database](https://www.energyplus.net/weather-location/north_and_central_america_wmo_region_4/CAN/ON/CAN_ON_London.716230_CWEC).
Calculation Validation

- Major changes comparison

1. fix the skipping line number in epw for the initialisation
    before: skip 9 lines; after: skip 8 lines

2. fix the wrong formula for the rest of days
    before: T_rm = 0.2 * T_rm-1 + 0.8 * T_od-1; after: T_rm = 0.8 * T_rm-1 + 0.2 * T_od-1
comparison

- Reference

The running mean temperature equation comes from BS EN 15251:2007, a snippet is given below:
image

### Pull Request Author

### Reviewer


src/EnergyPlus/ThermalComfort.cc
```diff
@@ -243,6 +243,7 @@ namespace ThermalComfort {
     Array1D<Real64> ZoneOccHrs;
     bool useEpwData(false);
     Array1D<Real64> DailyAveOutTemp(30, 0.0);
+    Real64 avgDryBulbCEN(0.0);
```
airallergy: being moved here for the sake of unit tests

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2853,13 +2856,10 @@ namespace ThermalComfort {
 
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
         std::string epwLine;
-        static Real64 avgDryBulbCEN(0.0);
         Real64 dryBulb;
         Real64 tComf;
         Real64 tComfLow;
-        static Real64 runningAverageCEN(0.0);
         Real64 numOccupants;
-        static bool useEpwData(false);
```
airallergy: being made global in the previous pull request concerning ASH55

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2964,7 +2971,7 @@ namespace ThermalComfort {
 
         if (BeginDayFlag && !firstDaySet) {
             // Update the running average, reset the daily avg
-            runningAverageCEN = 0.2 * runningAverageCEN + 0.8 * avgDryBulbCEN;
+            runningAverageCEN = alpha * runningAverageCEN + (1.0 - alpha) * avgDryBulbCEN;
```
airallergy: T_rm = (1 - alpha) * T_od-1 + alpha * T_rm-1, according to the cen standard

src/EnergyPlus/ThermalComfort.cc
Myoldmopar: Adding another global here doesn't seem necessary.  Can you just modify the function so that it either returns a Real64 or pass in a Real64 by reference and check the value after calling the function that way?

src/EnergyPlus/ThermalComfort.cc
Myoldmopar: I just want to make sure I'm clear of the intent here.  So this should've been removed in the previous PR?

src/EnergyPlus/ThermalComfort.cc
Myoldmopar: OK, so this won't make any difference because alpha was already fixed as a constant 0.8, but I like that there are fewer magic numbers here anyway.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2888,14 +2888,21 @@ namespace ThermalComfort {
...
+                // determine number of days in year
+                int DaysInYear;
+                if (DataEnvironment::CurrentYearIsLeapYear) {
+                    DaysInYear = 366;
+                } else {
+                    DaysInYear = 365;
+                }
                 auto epwFile = ioFiles.inputWeatherFileName.open("CalcThermalComfortAdaptiveCEN15251");
-                for (i = 1; i <= 9; ++i) { // Headers
+                for (i = 1; i <= 8; ++i) { // Headers
```
Myoldmopar: Right, it should start taking data on the 9th line.  To make this really clear, you could declare it above:
`int const numHeaderRowsInEPW = 8;`.  Then you don't even need the `// Headers` comment.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2888,14 +2888,21 @@ namespace ThermalComfort {
             const bool epwFileExists = FileSystem::fileExists(ioFiles.inputWeatherFileName.fileName);
             readStat = 0;
             if (epwFileExists) {
+                // determine number of days in year
+                int DaysInYear;
+                if (DataEnvironment::CurrentYearIsLeapYear) {
+                    DaysInYear = 366;
+                } else {
+                    DaysInYear = 365;
+                }
```
Myoldmopar, line 3013: This IF block does seem reasonable.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2888,14 +2888,21 @@ namespace ThermalComfort {
...
                 auto epwFile = ioFiles.inputWeatherFileName.open("CalcThermalComfortAdaptiveCEN15251");
-                for (i = 1; i <= 9; ++i) { // Headers
+                for (i = 1; i <= 8; ++i) { // Headers
                     epwFile.readLine();
                 }
                 jStartDay = DayOfYear - 1;
                 calcStartDay = jStartDay - 7;
                 if (calcStartDay > 0) {
-                    calcStartHr = 24 * (calcStartDay - 1) + 1;
+                    calcStartHr = 24 * calcStartDay + 1;
```
Myoldmopar, line 3022: So I guess the `- 1` wasn't needed because we already offset DayOfYear by 1 a couple lines up?

tst/EnergyPlus/unit/ThermalComfort.unit.cc
```diff
@@ -998,3 +999,25 @@ TEST_F(EnergyPlusFixture, ThermalComfort_CalcThermalComfortPierceSET)
...
+
+TEST_F(EnergyPlusFixture, ThermalComfort_CalcThermalComfortAdaptiveCEN15251)
+{
+    state.files.inputWeatherFileName.fileName = configured_source_directory() + "/tst/EnergyPlus/unit/Resources/ThermalComfortCEN15251Test.epw";
+
+    // test the initialisation
+    DayOfYear = 1;
+    CurrentYearIsLeapYear = false;
+    CalcThermalComfortAdaptiveCEN15251(state.files, true, true, 0.0);
+    EXPECT_NEAR(ThermalComfort::runningAverageCEN, -1.3671408, 0.01);
```
Myoldmopar: Are these output values harvested from some other tool?  I assume they aren't just what the function happened to return, but the one below does have an awful lot of precision which makes me concerned.

Myoldmopar (changes requested):
I'd like the global variable thing taken back out and instead add an argument to the function that returns the variable value.  In general, we are trying to reduce the global state of the program these days.

I think your calculation changes in the function are reasonable, but I'll have to think through the calcStartDay thing to make sure I understand the original intent and effect of the change.

Finally it would be good if you could comment on what you are using as a validation in the unit tests.  The unit test is straightforward enough, I just want to make sure the function is producing what you would expect, hopefully from a third party tool/resource.

Thanks for this contribution, I think with just a bit of discussion and tweaks we can get this in.

src/EnergyPlus/ThermalComfort.cc
airallergy: There used to be one of this variable in both ASH55 and CEN15251 separately, and the one in ASH55 was made global in the precious PR for the interest of unit tests (as I understand). So I think this one is redundant in the CEN15251, since there has been a shared global one.

src/EnergyPlus/ThermalComfort.cc
airallergy: Not exactly, the original formula has incorrectly switched coefficients, where alpha (0.8) should be multiplied by the running mean temperature instead of the dry bulb. This bit is the biggest change in the algorithm in this PR I suppose.

src/EnergyPlus/ThermalComfort.cc
airallergy: This I will do shortly. Do you think I should change this bit for the ASH55 above by the way?

src/EnergyPlus/ThermalComfort.cc
airallergy: I am not sure about the original intent, but I think this could be one explanation.

tst/EnergyPlus/unit/ThermalComfort.unit.cc
airallergy: To be precise, these exact values are indeed from the ep simulation output, but I did validate them in both the excel manual calculation and a python script calculation directly from the epw file. However, rest assured, differences are only a matter of precision, and I suppose I can add a screenshot of the excel one for demonstration. One thing to note, the dry bulb temperature would be slightly different, as briefly explained in [the previous ASH55 PR](https://github.com/NREL/EnergyPlus/pull/6474#issue-168947418), leading to differences in these values that are small but uncertain.

src/EnergyPlus/ThermalComfort.cc
airallergy: I agree with your opinion about the globals, it is simply that I have failed so far to come up a more proper way to do the unit tests. I was worried that modifying inputs and outputs of the function might mess up other parts of codes that I'm unfamiliar with. Let me have some more thinking on this. Also if you have any other ideas, do let me know.

src/EnergyPlus/ThermalComfort.cc
airallergy: I have changed the ASH55 as well for now, let me know if you think it inappropriate for this is a CEN15251 PR.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2629,6 +2631,7 @@ namespace ThermalComfort {
         int j;
         bool weathersimulation;
         Real64 inavgdrybulb;
+        int const numHeaderRowsInEpw = 8;
```
mitchute: There's a method in WeatherManager for this: https://github.com/NREL/EnergyPlus/blob/575a350ad418265382f0dd3f0c06ce038f4f8636/src/EnergyPlus/WeatherManager.cc#L8325

Can we move to using that instead of hardcoding the header offset?

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2873,6 +2874,7 @@ namespace ThermalComfort {
...
```
mitchute, line 2991: Same comment here.

Myoldmopar:
I'd like to see this go in, but I think it's too much to get done by our freeze which is....soon.  I'm changing the milestone for the next release, but if you can bring this up to date with develop and address the review comments in very short order, it could still go in.

nrel-bot-3:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-2b:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@airallergy @lgentile it has been 28 days since this pull request was last updated.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3015,22 +3016,33 @@ namespace ThermalComfort {
...
+                } else {
+                    DaysInYear = 365;
                 }
+
+                // Borrowed from OpenEPlusWeatherFile
+                // state.files.inputWeatherFile.close();
+                // state.files.inputWeatherFile.fileName = state.files.inputWeatherFileName.fileName;
+                // state.files.inputWeatherFile.open();
+                // WeatherManager::SkipEPlusWFHeader(state);
+                WeatherManager::OpenEPlusWeatherFile(state, ErrorsFound, false);
```
airallergy: I am not entirely sure about the relationship between `state.files.inputWeatherFileName` and `state.files.inputWeatherFile`, but in order to use `SkipEPlusWFHeader`, it seems the former needs to be changed to the latter in ASH55 and CEN15251 calculation, and all `epwFile.readLine()` to `state.files.inputWeatherFile.readLine()` accordingly. I am not that certain if this would affect other bits of codes, which I am not familiar with, all tests are passed from my end though.

To use `SkipEPlusWFHeader`, either to directly use `OpenEPlusWeatherFile` with an additional local variable `bool ErrorsFound(false)` (the current version), or to use the four lines from `OpenEPlusWeatherFile` (commented out), which would be preferable?

nrel-bot-3:
@airallergy @lgentile it has been 31 days since this pull request was last updated.

airallergy:
@Myoldmopar, fixed some conflicts and variable name changes emerging after my last commits, seems good to me now, would you like to take a look at this?

airallergy:
Hi @Myoldmopar @mitchute, I think I have made all the requested changes, would you mind taking a look at it, so it could make it in the next release?

mitchute:
It IS tagged for the 9.6 release, so I don't think there should be a problem getting it in. Since we can't push to your fork, we may ask you to merge up with the develop branch one more time, but we should probably wait on that until the I/O freeze release deadline has passed.

airallergy:
@mitchute , sure no problem, just let me know when you see fits. I haven't seen any new conflicts as of yet, so it should be of no efforts.

mitchute:
@airallergy could you resolve this conflict and pull develop in one more time? There was a new unit test added on top of the one you added, so just keep them both to resolve the conflict.

airallergy:
@mitchute done!

src/EnergyPlus/ThermalComfort.cc
mitchute: `inputWeatherFileName` is just the name of the weather file e.g. "foo.epw" for example. `inputWeatherFile` is the actual instance of that file that's been loaded into memory. All of your conversions from `epwFile` to `state.files.inputWeatherFile` is due to removing the local alias here: 

https://github.com/NREL/EnergyPlus/pull/8175/files#diff-d53cae7b21da3b8c1d38b3f3cd81f586119b970ee9d76142def9d6a2cb817411L2795

Regarding your question: you can just use `OpenEPlusWeatherFile` and remove the commented lines.

src/EnergyPlus/ThermalComfort.cc
mitchute: Agreed

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2748,6 +2749,7 @@ namespace ThermalComfort {
         int j;
         bool weathersimulation;
         Real64 inavgdrybulb;
+        bool ErrorsFound(false);
```
mitchute: We should be doing something with this new `ErrorsFound` flag and not just letting it get set to "True" and then doing nothing. Let's add a `ShowFatalError` (protected by an IF block) directly after each one of these calls to `OpenEPlusWeatherFile`.

tst/EnergyPlus/unit/ThermalComfort.unit.cc
```diff
@@ -1601,6 +1602,38 @@ TEST_F(EnergyPlusFixture, ThermalComfort_CalcThermalComfortFanger_Correct_TimeSt
     EXPECT_NEAR(state->dataThermalComforts->ThermalComfortData(1).FangerPPD, 100.0, 0.001);
 }
 
+TEST_F(EnergyPlusFixture, ThermalComfort_CalcThermalComfortAdaptiveCEN15251)
```
mitchute, line 1603: New, compact unit test look OK.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3081,7 +3093,7 @@ namespace ThermalComfort {
         if (state.dataGlobal->BeginDayFlag && !state.dataThermalComforts->firstDaySet) {
             // Update the running average, reset the daily avg
             state.dataThermalComforts->runningAverageCEN =
-                0.2 * state.dataThermalComforts->runningAverageCEN + 0.8 * state.dataThermalComforts->avgDryBulbCEN;
+                alpha * state.dataThermalComforts->runningAverageCEN + (1.0 - alpha) * state.dataThermalComforts->avgDryBulbCEN;
```
mitchute, line 3092: @airallergy  Can you post an excerpt from where you're getting these equations on the PR for later reference?

mitchute (changes requested):
There are a few changes requested here, but nothing major. I'll do some testing before this merges.

src/EnergyPlus/ThermalComfort.cc
airallergy: Thanks for the explanation.

src/EnergyPlus/ThermalComfort.cc
airallergy: Aren't all related error messages handled by `ProcessEPWHeader` which is called by `OpenEPlusWeatherFile` in `WeatherManager.cc`? Or is it that we should have a summary message here, something like `failed reading the epw file {epw file name}, check detailed error messages above(?)`?

mitchute: I just traced it through. You're calling `OpenEPlusWeatherFile` with the `ProcessHeader` arg set to `false`. Which means that you drop straight into `SkipEPlusWFHeader` - this _does not_ pass the `ErrorsFound` flag. So basically, there's no way for `ErrorsFound` to ever be changed from what you set it to.

`OpenEPlusWeatherFile` will fatal out by itself if it can't read the file, and `SkipEPlusWFHeader` will also fatal if there's a problem skipping the header lines. Don't worry about adding anything else since this is the signature we're stuck with on these functions.

airallergy: yes, right, we do not process the header here, I just recalled that thats why I asked about whether to directly use the commented lines previously :) I will just delete the commented lines and do a commit immediately.

mitchute:
OK, I ran some tests and this looks ready. @airallergy it would be good if you could address [this comment](https://github.com/NREL/EnergyPlus/pull/8175#pullrequestreview-456233363), quoted below:

> Finally it would be good if you could comment on what you are using as a validation in the unit tests. The unit test is straightforward enough, I just want to make sure the function is producing what you would expect, hopefully from a third party tool/resource.

Also, pull in develop one more time while you're at it. I'm not sure why it hasn't picked this one up yet but we should pull it up one more time anyway.

airallergy:
@mitchute, I have addressed this comment before, please check the pr overview which includes a screenshot of detailed calculation using excel, and a diagram of existing and expected annual results. Kindly let me know if you think any further information is required.

Will do another pull shortly.

mitchute:
@Myoldmopar any idea why CI isn't picking this up? I think it's ready if we can get a CI run to confirm its status.

Myoldmopar:
It's probably because the PR list is too long and CI doesn't see it.

mitchute:
OK, running tests locally.

mitchute:
Hmm, I'm a little confused here. I ran full regressions locally and get the following results.

Screen Shot 2021-09-01 at 10 48 33 AM

The unsuccessful runs are OK, and the run with big math diffs is expected and OK. However, the runs with big table diffs are showing up with the following:

Screen Shot 2021-09-01 at 10 48 48 AM

Big diffs in the HVAC Rejected Heat?? How does the thermal comfort code affect that??

mitchute:
OK, so I pulled develop in _one more time_, but I have some strange results. I ran the full test suite again and am getting similar, althought not identical results. Some of the files with table diffs match up, but not all of them. Nor do the magnatude of the diffs.

Screen Shot 2021-09-02 at 3 27 18 PM

However, when I run those files with diffs by themselves I don't see any diffs.

Anyone have thoughts here?

rraustad:
Are you using the same weather file used for the full test suite? See ...\testfiles\CMakeLists.txt

mitchute:
I’m using the regression tool in both instances. Except one time I’m running the full suite and the next time I’m just running the ones with diffs. No changes otherwise.

src/EnergyPlus/ThermalComfort.cc
rraustad: `24 * (calcStartDay - 1)` is a conversion to hours. Why exactly was this changed when calcStartDay wasn't changed?

airallergy: Because it was wrong by offsetting one excessive day here. This was also an issue with the ASHRAE comfort calculation function above, which was fixed in #6474.

rraustad: I see, this code is wrapping the epw to 24*calcStartDay as the number of hours last year and +1 is the first hour of this year. If you pulled develop this would not show as a diff. Never mind that right now.

airallergy:
Why did one test seem to fail without running? Do I need to pull from develop once more?

rraustad:
@airallergy that happened to me last week. Click on the Details link. At the top of the next page, click the red button Delete results and force rebuild and then click the accept? button. That way you don't run all the other CI servers.

airallergy:
@rraustad , thanks for point this out, but since I don’t have the writing access to that repo, I have to do a pr for deleting it. Would you minding doing it for me?

rraustad:
Done.

rraustad:
There are some unexplained diffs in this branch that I assume are due to this branch in the forked repo being behind NREL's develop. Maybe the best thing to do here is to pull the latest NREL develop into this branch and push it back up for fresh CI results.

airallergy:
Sure, will do it later today.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -2793,21 +2795,19 @@ namespace ThermalComfort {
                 }
                 state.dataThermalComforts->DailyAveOutTemp = 0.0;
 
-                auto epwFile = state.files.inputWeatherFilePath.open(state, "CalcThermalComfortAdaptiveASH55");
-                for (i = 1; i <= 8; ++i) { // Headers
-                    epwLine = epwFile.readLine().data;
-                }
+                WeatherManager::OpenEPlusWeatherFile(state, ErrorsFound, false);
```
Myoldmopar: Is this change necessary?  It feels like it could have side effects, and is muddying up the changes here.  Maybe I'm misunderstanding why this is necessary.

airallergy: Do you refer to the change in the ASHRAE function while this pr is mainly about CEN15251? Or do you refer to the use of `OpenEPlusWeatherFile` in both functions?

Myoldmopar: I'm wondering why you needed to change from

```
auto epwFile = state.files.inputWeatherFilePath.open(state, "CalcThermalComfortAdaptiveASH55");
```

to

```
WeatherManager::OpenEPlusWeatherFile(state, ErrorsFound, false);
```

I'm not actually saying the previous is correct, but I am wondering if that actually affects your bug fix here.  I'm trying to minimize the number of lines that your PR is changing to make sure we can get it reviewed and merged.  If this is an important change then it would be good to know why.  If it's not an important change, then it could potentially just make it tougher to review if it causes changes in results.

airallergy: Yes, I understand your point and totally agree with you, I was just trying to figure out which part should be altered. So let me put it this way.

These few lines involve skipping the epw header, and I currently changed that for both CEN and ASH. Changing of that for ASH to `OpenEPlusWeatherFile` is **not** essential for the sake of this pr, there was no error as to these specific lines (apart from some variable name changes in develop, but that was handled by pulling from upstream). I just did it to keep some consistency, as I was recommended to use `OpenEPlusWeatherFile` in CEN instead of hard coding the skipping.

In terms of `OpenEPlusWeatherFile`, I didn't find any of its usage outside `WeatherManager.cc`, and there are hard-coded skipping instead. Maybe you want to do a standalone pr for that.

Kindly let me know what you would like me to do here, I am ok with either way.

Myoldmopar: My intent really is to just minimize the changes in this PR so that we can get this in for you.  I hate that this has taken so much effort and the amount of effort in reviewing goes up as the number of lines go up.  (Not always, but I think you get my point.)  

I would cut this down as far as possible just to the bare minimum set of changes so that we can look at the results changes and feel comfortable with them changing.

airallergy: Now no touching the ASH function, no use of  `OpenEPlusWeatherFile`, diffs should be much cleaner now. waiting to see ci runs.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3003,14 +3004,22 @@ namespace ThermalComfort {
...
+                int DaysInYear;
+                if (state.dataEnvrn->CurrentYearIsLeapYear) {
+                    DaysInYear = 366;
+                } else {
+                    DaysInYear = 365;
+                }
+
                 auto epwFile = state.files.inputWeatherFilePath.open(state, "CalcThermalComfortAdaptiveCEN15251");
-                for (i = 1; i <= 9; ++i) { // Headers
+                for (i = 1; i <= numHeaderRowsInEpw; ++i) {
```
Myoldmopar, line 3016: This went from 9 to 8 in this PR.  Makes sense with 8 header rows:

1. LOCATION
2. DESIGN CONDITIONS
3. TYPICAL/EXTREME PERIODS
4. GROUND TEMPERATURES
5. HOLIDAYS/DAYLIGHT SAVINGS
6. COMMENTS 1
7. COMMENTS 2
8. DATA PERIODS

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3003,14 +3004,22 @@ namespace ThermalComfort {
             const bool epwFileExists = FileSystem::fileExists(state.files.inputWeatherFilePath.filePath);
             readStat = 0;
             if (epwFileExists) {
+                // determine number of days in year
+                int DaysInYear;
+                if (state.dataEnvrn->CurrentYearIsLeapYear) {
+                    DaysInYear = 366;
+                } else {
+                    DaysInYear = 365;
```
Myoldmopar, line 3012: This is still a good change.

src/EnergyPlus/ThermalComfort.cc
Myoldmopar: This whole thing appears to be built around an assumption that the weather file has hourly data, which makes me feel funny. This PR isn't changing it, just handling this part better, and that is acceptable here.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3080,7 +3089,7 @@ namespace ThermalComfort {
         if (state.dataGlobal->BeginDayFlag && !state.dataThermalComforts->firstDaySet) {
             // Update the running average, reset the daily avg
             state.dataThermalComforts->runningAverageCEN =
-                0.2 * state.dataThermalComforts->runningAverageCEN + 0.8 * state.dataThermalComforts->avgDryBulbCEN;
+                alpha * state.dataThermalComforts->runningAverageCEN + (1.0 - alpha) * state.dataThermalComforts->avgDryBulbCEN;
             state.dataThermalComforts->avgDryBulbCEN = 0.0;
```
Myoldmopar, line 3093: Thanks for adding the reference information in the PR, this should be totally acceptable.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3031,9 +3040,9 @@ namespace ThermalComfort {
                     }
                 } else { // Do special things for wrapping the epw
                     calcEndDay = jStartDay;
-                    calcStartDay += 365;
+                    calcStartDay += DaysInYear;
```
Myoldmopar, line 3043: Using the new dynamically determined days in year is great.

src/EnergyPlus/ThermalComfort.cc
```diff
@@ -3031,9 +3040,9 @@ namespace ThermalComfort {
...
                     calcEndHr = 24 * calcEndDay;
-                    calcStartHr = 24 * (calcStartDay - 1) + 1;
+                    calcStartHr = 24 * calcStartDay + 1;
```
Myoldmopar, line 3045: I'm willing to accept this as the corrected start by shifting a day.

Myoldmopar (approved):
I'm happy with this as long as CI comes back clean (I believe we'll still have 1 diff expected).

Myoldmopar:
@mitchute you are still showing changes requested.  I believe that's just out of date at this point.  I ran full tests and regressions locally and this is all in good shape to go in, just the one diff file.  CI is looking super happy so far.  I'm inclined to merge but I will wait a little while.  May merge before sleepy time.  Let me know if anyone is waiting on anything else here.

Myoldmopar:
Looking good.  Going ahead and merging.  Thanks all!

PR #8178, opened 2020-07-28
Avoid using DataStringConstants::NL whenever possible and trust OS

lefticus:
- On windows, with a text file, '\n' does the right thing

Very simple PR. Should specifically fix the issues that @mjwitte noticed for error output files.


src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -1844,12 +1844,12 @@ namespace UtilityRoutines {
 
 
         if (UtilityRoutines::outputErrorHeader && err_stream) {
-            *err_stream << "Program Version," + VerString + ',' + IDDVerString + DataStringGlobals::NL;
+            *err_stream << "Program Version," << VerString << ',' << IDDVerString << '\n';
```
Myoldmopar, line 1847: What, [not](https://www.youtube.com/watch?v=GMqQOEZYVJQ) `std::endl`?

third_party/ObjexxFCL/src/ObjexxFCL/char.constants.hh
```diff
@@ -16,7 +16,6 @@
 namespace ObjexxFCL {
 
 extern char const CMA;
-extern char const NL;
```
Myoldmopar, line 19: Nice nice nice.  One less to manage.

Myoldmopar (approved):
I haven't tested this, but the changes look top notch.

src/EnergyPlus/UtilityRoutines.cc
lefticus: https://www.youtube.com/watch?v=GMqQOEZYVJQ

lefticus: Oh I missed your "not" being a link :D

mjwitte (approved):
Tested locally on windows. All good now.

mjwitte:
But I did not run unit tests locally... 
Many unit tests are failing on windows due to line ending differences in error message string compares. Here's one example:
```
[ RUN ] EnergyPlusFixture.HeatBalanceIntRadExchange_CarrollMRT
C:\ci\runs\clone_branch\tst\EnergyPlus\unit\Fixtures\EnergyPlusFixture.cc(204): error: Expected: expected_string
Which is: " ** Severe ** Geometry not compatible with Carroll MRT Zone Radiant Exchange method.\r\n"
To be equal to: stream_str
Which is: " ** Severe ** Geometry not compatible with Carroll MRT Zone Radiant Exchange method.\n"
C:\ci\runs\clone_branch\tst\EnergyPlus\unit\HeatBalanceIntRadExchange.unit.cc(128): error: Value of: compare_err_stream(error_string, true)
Actual: false
Expected: true
```
`expected_string` is constructed with `EnergyPlusFixture::delimited_string` which is using `DataStringGlobals::NL` as the default delimiter.

`DataStringGlobals::NL` is used in main.cc, InputProcessor.cc, EnergyPlusFixture.hh, and IdfParserFixture.hh. And IdfParser.cc is rolling it's own local `NL`.

mbadams5:
@mjwitte All those uses of `NL` can be changed to the new approach. I just did it in various spots to be consistent with the rest of EnergyPlus previously.

mjwitte:
@lefticus Will you be able to address [this](https://github.com/NREL/EnergyPlus/pull/8178#issuecomment-665612993)?

Myoldmopar:
@mjwitte it looks like you are saying there is still a step to be made to fix up the unit tests, is that still the case here?

mjwitte:
@Myoldmopar Yes, this still needs more work. See the lastest [Windows CI results](https://github.com/NREL/EnergyPlus/pull/8178#commitcomment-40984791)

Myoldmopar:
OK, I can tackle that cleanup.  I built and ran it on Linux but the unit tests run fine.  I'll need to build and run on Windows to see it, which isn't a big problem.  I'll try to do that soon.  Unless @lefticus is already making these changes...

Myoldmopar:
OK, on Windows the unit test suite now passes.  I just moved all the failing tests and worker functions to use plain old `\n`.  @mbadams5 if you could bless my changes in f9823b8 that would be great.

mbadams5:
@Myoldmopar The changes seem reasonable and the unit tests seem to pass.

Myoldmopar:
@mjwitte do you want to re-run with my fix?  It looks like it is good to go to me.

lefticus:
Sorry, missed this ongoing conversation @Myoldmopar the changes you made are inline with similar changes I've had to make as I've updated each output file to being in text mode instead of binary mode.

mjwitte:
@Myoldmopar I've re-run with this branch. Output files look good. Converted between idf and epJSON in both directions and the converted files look good. Other than the failed movefile_test on windows this is good to go. I'm assuming that's a random test failure. The test passes locally.

Myoldmopar:
Thanks @lefticus and @mjwitte, merging this in now.

Issue #8186, opened 2020-08-05
ERL trace assertion error


JasonGlazer:
Issue overview
--------------
An assertion error is shown when running debug:

Assertion failed: outputFile.os, file
C:\repos\EnergyPlus01\src\EnergyPlus\IOFiles.hh, line 360

with the defect file.

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
 - Version of EnergyPlus (if using an intermediate build, include SHA)
 - Unmethours link or helpdesk ticket number

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added ( EnergyPlusDevSupport\DefectFiles\8000s\8186\orig_back_930-fix.idf)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

lefticus:
Fixed in #8187

PR #8187, opened 2020-08-05
Output to stdout if no edd output file is requested

lefticus:
No change in tests expected.

 * restores previous behavior before removal of GIO
 * optionally allow any output file to fall back to stdout

Addresses #8186

### DIFF
--- a/src/EnergyPlus/IOFiles.hh
+++ b/src/EnergyPlus/IOFiles.hh
@@ -51,6 +51,7 @@
 #include <fstream>
 #include <fmt/format.h>
 #include <fmt/ostream.h>
+#include <iostream>
 #include <ostream>
 #include <vector>
 #include <cassert>
@@ -124,6 +125,9 @@ private:
 class InputOutputFile
 {
 public:
+    std::string fileName;
+    bool defaultToStdOut = false;
+
     void close();
     void del();
     bool good() const;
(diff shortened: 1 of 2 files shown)


Myoldmopar:
@JasonGlazer since you opened the issue, are you available to test this?

JasonGlazer:
@Myoldmopar Sure, I should have time next week.

Myoldmopar:
I was able to reproduce actually.  With develop:

```
Warming up {1}
energyplus: /eplus/repos/1eplus/src/EnergyPlus/IOFiles.hh:360: void EnergyPlus::print(EnergyPlus::InputOutputFile&, fmt::v6::string_view, const Args& ...) [with Args = {}; fmt::v6::string_view = fmt::v6::basic_string_view<char>]: Assertion `outputFile.os' failed.
Aborted (core dumped)
```

With this branch:

```
Warming up {1}
****  Begin EMS Language Processor Error and Trace Output  *** 
<Erl program name, line #, line text, result, occurrence timing information ... >
**FATAL:Previous EMS error caused program termination.
EnergyPlus Run Time=00hr 00min  4.68sec
Program terminated: EnergyPlus Terminated--Error(s) Detected.
```

Much more graceful exit.  I'm good with this going in.

src/EnergyPlus/IOFiles.hh
```diff
@@ -357,8 +360,20 @@ template <typename... Args> void print(std::ostream &os, fmt::string_view format
...
+            return outputFile.os.get();
+        } else {
+            if (outputFile.defaultToStdOut) {
+                return &std::cout;
+            } else {
+                assert(outputFile.os);
+                return nullptr;
+            }
+        }
+    }();
```
Myoldmopar, line 374: I feel like this is a band-aid that we'll want to rip back off later.  It is probably indicative that the EDD writing code needs to be altered to write to `cout` instead.  But it'll work for now.  Thanks for getting it working.

Myoldmopar:
Unit test suite is running fine on my machine so far, but it's a debug build so it's taking a really long time.  I'm going to go ahead and merge this and move on.  Thanks @lefticus

Issue #8180, opened 2020-07-31
Empty error file with input processor errors


mjwitte:
Issue overview
--------------
While testing another PR, noticed that an idf with an invalid key choice produces an empty error file. Not sure when this started. Sometime after v9.3.0 bugfix release which works fine.

### Details
Some additional details for this issue (if relevant):
 - Platform Windows 10
 - Version of EnergyPlus v9.4.0-https://github.com/NREL/EnergyPlus/commit/e884660c04f1cc6f14ff32f105ee650637069b7f

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Defect file added - https://github.com/NREL/EnergyPlusDevSupport/blob/0738e371061a1135ef5fcf95843c52902c2970af/DefectFiles/8000s/8180/8180-1ZoneUncontrolled-develop.idf
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

jmarrec:
Works fine on **Ubuntu 18.04** with current develop 922f888aef

```
(py38)julien@8180 (master *%=)$ $ep_build_rel/Products/energyplus -D -d out-ori 8180-1ZoneUncontrolled-develop.idf 
EnergyPlus Starting
EnergyPlus, Version 9.4.0-922f888aef, YMD=2020.08.03 14:31
**FATAL:Errors occurred on processing input file. Preceding condition(s) cause termination.
EnergyPlus Run Time=00hr 00min  0.15sec
Program terminated: EnergyPlus Terminated--Error(s) Detected.

(py38)julien@8180 (master *%=)$ cat out-ori/eplusout.err 
Program Version,EnergyPlus, Version 9.4.0-922f888aef, YMD=2020.08.03 14:31,
   ** Severe  ** <root>[SimulationControl][SimulationControl 1][do_zone_sizing_calculation] - "Maybe" - Failed to match against any enum values.
   **  Fatal  ** Errors occurred on processing input file. Preceding condition(s) cause termination.
   ...Summary of Errors that led to program termination:
   ..... Reference severe error count=1
   ..... Last severe error=<root>[SimulationControl][SimulationControl 1][do_zone_sizing_calculation] - "Maybe" - Failed to match against any enum values.
   ************* Warning:  Node connection errors not checked - most system input has not been read (see previous warning).
[...]

PR #8188, opened 2020-08-05
Move initialization of IO singleton to correct place

lefticus:
Addresses #8180

Note that the existence of this singleton is not ideal. It is still here
to avoid updating every place that ShowError* is called from


Myoldmopar:
I agree that it is not ideal, and I look forward to making the big change to all Show* functions to just include the full state once and for all.  That will be a big one. This relocation is fine though, the earlier the better I suppose.

mjwitte (approved):
@lefticus Thanks. Confirmed fix locally.

Issue #7496, opened 2019-09-10
IDF parsing error for WindowShadingControl in v9.0 and above


ayezioro:
Hi,
I opened this [discussion ](https://discourse.ladybug.tools/t/e-9-01-and-above-windowproperty-shadingcontrol-not-working/6684/18)in the Ladybug discourse and now i think i should report here ... and ask for advice.
Checked also this request in unmethours [forum](https://unmethours.com/question/36128/idf-parsing-error-for-windowshadingcontrol-in-v90/). 
From the later i edited the Energy+.idd file in order to add more fields to the WindowShadingControl.
It worked for a while. Now i get to a building where the amount of fields i added is not enough. Of course i can add more, but it can be not enough for other cases.

My question here is twofold:
1. Is this issue, that all Shading Controls are written to the first object in the idf file, something that can yield wrong results?
2. In case the previous answer is negative, it will be possible to solve the issue in the next version of the software?

Thanks!!
-A.

mjwitte:
@ayezioro Trying to piece all of this together. Some questions:

a. Is the "IDF parsing error" in IDF Editor? The only solution for this is to extend the IDD object in your local copy of the IDD.
b. Does the EnergyPlus simulation complete successfully?
c. What software is writing all of the shading controls into a single WindowShadingControl object?  If the shading controls are linked to daylighting and the surfaces are in different zones, then the results will likely be incorrect.

ayezioro:
Hi @mjwitte,
In response to your questions:
1. I see the issue in the IDF editor, though the file was not created directly in this editor. I've already extended the IDD but it is not a stable and always reliable solution. It worked when i was working with, say, 3 floor buildings. When i rise the hight (i.e. 9 floors) it was not enough and the simulation crashed. The thing is you can not know how many lines to extend the IDD.
2. Yes. The simulation ends (for those cases that the IDD can confine the amount of shading controls assigned to one (or a couple) of windows.
3. I'm using Honeybee (a plugin for Grasshopper). @chriswmackey, one of the authors, said something changed in the OS SDK after 2.5, but i'm not sure exactly what is going on. As i understand Honeybee uses this SDK for creating the OSM and IDF files. 
For now the simulations are intended for energy consumption calculations but in the future they can also include daylighting.

I understand this is a messy issue, and probably bad explained. If you need i can provide a link for a sample file.

Thanks!!
-A.

mjwitte:
[...]

PR #8196, opened 2020-08-11
Allow Multiple Scheduled Shades to Reference a Single Window

JasonGlazer:
Pull request overview
---------------------
Allow modeling a window with different shading coefficient based on schedule (such as during the summer and winter). This is based on a requirement in "ANSI/RESNET/ICC 301-2014 Standard for the Calculation and Labeling of the Energy Performance of Low-Rise Residential Buildings" which requires (see page 14) that glazing should have an interior shade coefficient of 0.70 in the summer and 0.85 in the winter. 

Here is the [new feature proposal.](https://github.com/NREL/EnergyPlus/blob/AllowMultipleScheduledShades/design/FY2020/NFP-MultipleScheduledShades.md)

### Pull Request Author

### Reviewer


JasonGlazer:
I just converted this back to a draft PR. I thought of a new test to try (putting in extreme values for the transmittance and reflectance values and seeing if they are appropriately changing the output) and it did not work so this is not ready to review yet.

JasonGlazer:
I think the diffs are just due to develop moving so remerged and hopefully CI will be clean

JasonGlazer:
@Myoldmopar and @mjwitte I think this is ready for review and merge but develop changed during its CI tests so only some are coming back clean. Merging it again when we know develop is stable will probably come back clean. As a reminder this does not change IO (except for added fields for WindowShadingControl to address #7496) so it can probably be put off until more pressing reviews are performed.

JasonGlazer:
@mjwitte and @Myoldmopar looks like this came back clean from the CI. Please review and let me know if you have any comments.

mjwitte:
@JasonGlazer Do you have an idf to use for testing this?

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -4115,7 +4115,9 @@ \subsection{WindowShadingControl}\label{windowpropertyshadingcontrol}
...
+
+To deploy one shading device during one time period and another during another time period for the same window, specify two different WindowShadingControl objects that both reference the same fenestration.  Each of the WindowShadingControl input objects would include on the same window on their lists of fenestration objects, but would reference different schedule input objects. The referenced schedule input objects should have non-zero values specified at different times for each WindowShadingControl. This approach can be used to add shading that is different for different seasons. When more than one WindowShadingControl is used for the same window based on schedule, the type of shading, control type, setpoints, and several other parameters must be the same in the different WindowShadingControl objects, but the shading device and construction can be different.
```
mjwitte, line 4120: If other changes are needed, there's a small typo here: ". . . would include ~on~ the same window on their lists . . . " 
Also, this section is quite long and could use some subheadings to highlight the different points being made.

mjwitte: Why are the same shading control type and setpoints required?

src/EnergyPlus/SolarShading.cc
```diff
@@ -10348,6 +10358,25 @@ namespace SolarShading {
         } // End of surface loop
     }
 
+    int selectActiveWindowShadingControl(int curSurface)
+    {
+        // J. Glazer - Aug 2020 - For a given surface, determine based on the schedules which window shading control object should be active
```
mjwitte, line 10363: If other changes are needed, remove author date portion of comment. (In general, author credits are only via blame now.) Same comment for other new functions.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10348,6 +10358,25 @@ namespace SolarShading {
...
+        int selected = 1; // presume it is the first shading control - even if it is not active it needs to be some shading control which is then turned off in the WindowShadingManager
+        if (Surface(curSurface).windowShadingControlList.size() == 1) {
+            selected = Surface(curSurface).windowShadingControlList.front();
+        } else {
+            for (auto wsc = Surface(curSurface).windowShadingControlList.begin(); wsc != Surface(curSurface).windowShadingControlList.end(); ++wsc) {
```
mjwitte, line 10368: Couldn't this simply be ` for (int wsc : SurfaceTmp(curSurface).windowShadingControlList)` ? Just wondering for my own understanding. Similar for loops also occur elsewhere.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8853,22 +8888,24 @@ namespace SurfaceGeometry {
...
+                                } else {
+                                    MatGapFlow1 = dataConstruction.Construct(ConstrNumSh).LayerPoint(4);
+                                    MatGapFlow2 = dataConstruction.Construct(ConstrNumSh).LayerPoint(6);
+                                }
+                                if (dataMaterial.Material(MatGapFlow1).GasType(1) != 1 || dataMaterial.Material(MatGapFlow2).GasType(1) != 1) {
+                                    ErrorsFound = true;
+                                    ShowSevereError(cCurrentModuleObject + "=\"" + cAlphaArgs(1) +
+                                                    "\", gas type must be air on either side of the shade/blind");
+                                }
+                                break; // only need the first window shading control sicne they should be the same
```
mjwitte, line 8908: sicne --> since

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -11357,6 +11394,8 @@ namespace SurfaceGeometry {
 
             ConstrNewSh = TotConstructs + 1;
             SurfaceTmp(SurfNum).ShadedConstruction = ConstrNewSh;
+            WindowShadingControl(WSCPtr).ShadedConstruction = ConstrNewSh;
```
mjwitte, line 11397: Since the ShadedConstruction number is stored with the WindowShadingControl, that means when the Shading Device Material Name options is used, every window listed in a given WindowShadingControl object must have the same base unshaded construction. Is that checked somewhere? Would it be better to store a `shadedConstructionList` on the surface in order to allow the same shade to be applied to a list of windows with differing base constructions?

Tried a case with a different construction on WindowEast "WindowConstruction 2" uses a different material layer than the other window constructions. Got this error:
```
   ** Severe  **  The glass and gas layers in the shaded and unshaded constructions do not match for window=WINDOWNORTH
   **   ~~~   ** Unshaded construction=WINDOWCONSTRUCTION
   **   ~~~   ** Shaded construction=WINDOWCONSTRUCTION 2:WINDOWCOOLINGSHADE:INT
```
This would be a confusing error to encounter. Testing with 9.4-IOFreeze (with a single WSC object), the file runs fine even though the windows in the list have different constructions. So this needs to be fixed. Test file here: 
[test2b-xtreme-dual-gap-noEMS-differentwindowsonstructions-OneWSC.idf.txt](https://github.com/NREL/EnergyPlus/files/5138872/test2b-xtreme-dual-gap-noEMS-differentwindowsonstructions-OneWSC.idf.txt)

mjwitte (changes requested):
Tested this with 2 of the provided idfs and compared results. Window transmittances are changing as expected. Also compared with two cases that apply shade A for both shading controls, and another applying Shade B in both controls. These agree at the appropriate times with the base files.

Unit tests all pass.

src/EnergyPlus/SolarShading.cc
JasonGlazer: Good suggestion.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
JasonGlazer: I guess just because it wasn't part of the NFP. I think it would require some additional code changes in SolarShading.

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: The error message you are showing is doing the check you are suggesting. 

I think to accomplish what you are suggesting, the user would simply use the construction rather than the material option in WindowShadingControl.

It probably should have been reported as an error message previously but it wasn't checked as thoroughly before. I didn't add that error message but I think code that I added is applying that check more consistently now. I think it is working as intended. Please explain further what changes you want here.

mjwitte: If I understand correctly, previously, each shaded window surface stored it's own shaded and unshaded construction which was set at the time the WindowShadingControl input was processed. So, it the WSC applied a shade material to a list of windows with different base constructions, they all got their own custom shaded construction with that shade layer. Now the shaded construction is being stored on the WSC struct and is getting stepped on by the last unique window construction. Using the shaded construction option in the WSC would require a separate WSC object for every different window type.  All the same checks were there before in `CheckWindowShadingControlFrameDivider`. This breaks files that ran successfully before.

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
mjwitte: This seems overly restrictive. In `WindowShadingManager`, once the `activeWindowShadingControl` is set, all of the logic follow the selected WindowShadingControl. So, I don't see any reason why that couldn't have different rules. I don't think it would even need to be the same shading type.

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: Ok. I see what you are saying.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8468,6 +8468,42 @@ namespace SurfaceGeometry {
...
+                                        "\" has multiple WindowShadingContols that are not similar.");
+                            ShowContinueError("for: \"" + WindowShadingControl(firstWindowShadingControl).Name + " and: " + WindowShadingControl(*wsc).Name); 
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    bool isWindowShadingControlSimilar(int a, int b)
```
JasonGlazer, line 8491: @mjwitte so this is the routine that checks that various windowshadingcontrol object fields match. What do you think should be checked? I was concerned for unintended side affects of allowing these to be different so that is why I made it fairly restrictive, what do you think needs to be the same between window shading controls for a given window?

mjwitte: @JasonGlazer I'm not sure anything needs to be the same except the matching surface, unless there are some precalculations that are done based on the specific shading control type. And, I think we're still restricting the surfaces to all be in the same zone, so that would match also. Along those lines, I suppose if Multiple Surface Control Type = Group, then the list of surfaces should match? Or maybe just a warning that one control may override the other for an entire group?

Myoldmopar:
@JasonGlazer @mjwitte there are a number of comments as I've read through here that look like they still need to be addressed.  There are also a couple conflicting files.  Can you provide an update?  As soon as you are ready for another review, I'm happy to look it over and run some tests locally.

JasonGlazer:
@Myoldmopar one of the comments @mjwitte made about storing the shaded construction on the surface structure has triggered a wide range of other fixes as well related to storm windows with shades that I had overlooked earlier. I am working on it and expect to be done in the next few days. Then I need to remerge with develop which I expect to be complicated due to the number of changes in #8141. This week I hope.

JasonGlazer:
Closing this PR. Work continued under #8269

PR #8202, opened 2020-08-12
Fix AirflowNetworkBalanceManager divide by zero overflow issue causing NaN values in window temperature nan calculations.

### LINKED ISSUE #8200: NaN value for new window surface temperature due to overflow in AirflowNetworkBalanceManager
Issue overview
--------------
The attached EnergyPlus idf file results in a series of EnergyPlus error outputs indicating NaN values occuring during new window surface temperature calculations. The observed output observed is as follows:
`  ** Warning ** Saturation Temperature not converged after 51 iterations (PsyTsatFnPb)
  **  ~~~  ** Routine=Unknown, Environment=RUNPERIOD 1, at Simulation time=03/13 17:20 - 17:30
  **  ~~~  ** Input Pressure= NAN
  **  ~~~  ** Resultant Temperature= NAN
  ** Severe ** Convergence error in SolveForWindowTemperatures for window NORTH WALL WINDOW 1
  **  ~~~  ** Environment=RUNPERIOD 1, at Simulation time=03/13 17:30 - 17:40
  **  ~~~  ** Glazing face index = 1 ; new temperature =NANC ; previous temperature = 17.3239C
  **  ~~~  ** Glazing face index = 2 ; new temperature =NANC ; previous temperature = 17.5398C
  ** Fatal ** Program halted because of convergence error in SolveForWindowTemperatures for window NORTH WALL WINDOW 1`

This error is likely due to a divide by zero error causing a cascade of NaN values. 

### Details
 - Platform: All
 - Version of EnergyPlus: 9.3
[failing.txt](https://github.com/NREL/EnergyPlus/files/5064369/failing.txt)

### Checklist

ajscimone:
Pull request overview
---------------------
 - Fixes #8200
 - This pull request fixes a divide by zero error inside of the CalcAirflowNetworkHeatBalance function of the AirflowNetworkBalanceManager when a memeber of the AirflowNetworkLinkSimu array's Flow attribute is zero. This divide by zero error creates a NaN value, which can cascade through different calculations such as the new window temperature calculation which caused the discovery of this bug. Fixing this bug is a simple check for a zero condition on the attribute of the array element, and arbitrarily setting a value rather than attempting a divide by zero.

### Pull Request Author

### Reviewer


lgu1234:
@ajscimone I tried to repeat error message by running develop-based EnergyPlus using the defect file and Chicago weather, and was unable to get the same message. Please let me know which weather file was used.

ajscimone:
@lgu1234 I used the attached weather file to reproduce this bug (converted to a txt because Github does not support .epw files). I double checked, and this bug still appears for me. I've went ahead and opened up this PR, so if the CI comes back happy, this is all ready for review.
[TMY-Colorad-v5.0.txt](https://github.com/NREL/EnergyPlus/files/5097589/TMY-Colorad-v5.0.txt)

lgu1234:
@ajscimone I ran the defect file using the weather file you provided. The E+ is compiled based on existing develop branch. Unfortunately, I am unable to repeat what you found. Attached is the file for your reference.

one thing needs to be mentioned:

   ** Warning ** CalcDoe2DXCoil: Coil:Cooling:DX:SingleSpeed "HVAC COOLING COIL" - Full load outlet air dry-bulb temperature < 2C. This indicates the possibility of coil frost/freeze. Outlet temperature = 1.41 C.
   **   ~~~   **  ...Occurrence info = RUNPERIOD 1, 01/08 12:13 - 12:15
   **   ~~~   ** ... Possible reasons for low outlet air dry-bulb temperatures are: This DX coil
   **   ~~~   **    1) may have a low inlet air dry-bulb temperature. Inlet air temperature = 16.469 C.
   **   ~~~   **    2) may have a low air flow rate per watt of cooling capacity. Check inputs.
   **   ~~~   **    3) is used as part of a HX assisted cooling coil which uses a high sensible effectiveness. Check inputs.
   ** Warning ** Temperature out of range [-100. to 200.] (PsyPsatFnTemp)
   **   ~~~   **  Routine=PsyTwbFnTdbWPb, Environment=RUNPERIOD 1, at Simulation time=01/08 16:17 - 16:20
   **   ~~~   **  Input Temperature=-129.92

[eplusout.err.txt](https://github.com/NREL/EnergyPlus/files/5098561/eplusout.err.txt)

src/EnergyPlus/General.cc
```diff
@@ -3656,22 +3656,19 @@ namespace General {
         return results;
     }
 
-    Real64 epexp(Real64 x)
+    Real64 epexp(Real64 numerator, Real64 denominator, Real64 overflowReturn, Real64 defaultHigh)
```
jasondegraw: This should be rolled back and the functions separated back out. The first version was intended for cases where the only thing to worry about was whether we could avoid the `exp` call entirely, and the second version is the version with additional safety. This undoes the original intent and imposes all the logic in every case. The logic adds up.

src/EnergyPlus/General.hh
```diff
@@ -399,9 +399,8 @@ namespace General {
 
     std::vector<std::string> splitString(const std::string &string, char delimiter);
 
-    Real64 epexp(Real64 x);
+    Real64 epexp(Real64 numerator, Real64 denominator=1.0, Real64 overflowReturn=0.0, Real64 defaultHigh=std::numeric_limits<float>::max());
```
jasondegraw: This is also not great w.r.t. the original intent and involves the numerical limit even when it's not needed. For me this is another reason to keep both versions of the function.

jasondegraw (changes requested):
As @lgu1234 has been unable to reproduce the issue, it may be premature for me to comment but I'll be hard pressed to find time next week. So here goes. At a minimum, (i) the problem needs to be reproduced or there needs to be some explanation as to why it's not being reproduced and (ii) the combination of functions needs to be undone.

src/EnergyPlus/General.cc
nealkruis: @jasondegraw, the second function (with defaultHigh) is never used within the code (that I could find). I'd propose dropping that from the combined function. The rest of it always happens (or should always happen) anyway outside the function every time it's called to prevent the divide-by-zero, so why not bring it into the function?

There is a possibility of overflow still that we might want to protect against, but the current code isn't doing that (again, as far as I can tell).

nealkruis:
@lgu1234 the bug is reproducible in version 9.3.0 on Windows, but is very sensitive to HVAC iteration calculations. I can't even reproduce it on Mac for the same version, or in the latest develop on any machine. That does not mean that it is resolved, but that this file is no longer reproducing the conditions that trigger the bug. I will keep working with @ajscimone to track down exactly what's going on. We will make sure that we can resolve the issue by cherry-picking our changes to a branch off of 9.3.0.

src/EnergyPlus/General.cc
jasondegraw: @nealkruis Used or not is not a good argument here, the original intent was (and still is) to begin building out functions with better protections and more options in terms of what is calculated, and that as that happened we'd start using these throughout the code. Requiring that every call to `exp`must involve a numerator and a denominator would run rather counter to the original goal. That's why moving the check inside is a bad idea. I'm not disputing that there are cases where there may need to be additional protections, that's all fine. The combined function is not.

nealkruis: Okay, so maybe we'll wrap the call from Airflow Network in another function that checks the denominator and then calls `epexp`. That way you don't miss one of the cases with the external checks.

I still don't really understand the intent of the second form of the function. Why pass in a "defaultHigh"? There is a real limit on the arguments to exp before it returns HUGE_VAL (~700). If you want to avoid an invalid return, shouldn't we just limit the input to be max(x,700)?

Also, is there a reason to use -70 as the lower bound for tiny numbers? or is that a somewhat arbitrary cut off?

BTW, the suggested function did not require a denominator (you could just do `epexp(x)` still), but it would sometimes do an otherwise unnecessary divide by 1.0 (does the cpu even do an operation for divide by 1.0?).

rraustad: If I recall correctly, something just under -70 causes an underflow and over 370 an overflow. Those limits could easily be tested.

nealkruis: Just tested: Underflow doesn't really happen. At some point, C++ just returns zero. Overflow in C++ is actually closer to 700.

jasondegraw: @nealkruis I think it's fine to have a form that has a denominator as an argument, there's precedent for that sort of thing, e.g. `atan2`. But there's still an `atan`, so no, wrapping an additional function around this is not a solution. All I asked was that we keep the two separate functions, one with a single argument and one with the additional arguments. It's really not that hard.

See #7233 and #7414.

lgu1234: As I remembered, the current epexp function deals with underflow only and will not process overflow. Since possible problems in AFN are for underflow, the current function should be enough. There is no need to add more arguments.

In addition, there is no need to remove "if (AirflowNetworkLinkSimu(i).FLOW2 == 0.0) {". If we know the solution at FLOW = 0, it is not needed to call the function epexp for speedup.

mjwitte: @nealkruis Chiming in late in the game, but underflows are a thing, at least on Windows debug builds. I stumble across them frequently in the Tarcog functions. So, let's not remove underflow protection.

nealkruis: @mjwitte My understanding is that `std::exp` handles underflow internally and returns zero.

jasondegraw:
I'm going to say that I agree with @lgu1234 that the changes to how the exponential was being called are not ideal. There may still be an issue here, but I don't believe that this is the fix. With that in mind, I'm going to suggest that we close this PR and revisit this again at a later time.

nealkruis:
@jasondegraw What good does closing this PR do? Please be patient while we continue to investigate this further. In the future, it would probably save a lot of time if we wait for the PR to be completed (and out of draft mode) before others start reviewing.

jasondegraw:
@nealkruis Investigation should have taken place before the PR was posted. If the proper procedure had been followed from the beginning with the report of a potential defect, then your process complaints might carry a little more weight. It looks to me like there was one unprotected call of `epexp`, and that could have been addressed pretty simply.

nealkruis:
@ajscimone and I have looked at this a little closer, and it appears there are four considerations here:

1. Avoiding underflow and overflow
2. Performance
3. DRY code
4. Building a useful set of functions

Here are our findings and recommendations on each of these:

**Avoiding underflow and overflow**

Let's talk about underflow first, since that is the only thing actually being handled in the current code. As far as we can tell, underflow is only an issue on Windows debug builds because of the `\fp:strict` flag. This is a valuable flag to have to make sure we aren't being too sloppy with floating point exceptions in our code. However, the `std::exp` function handles underflow internally. It's only because we have this flag that we see the issue @mjwitte originally reported (#7233). Without the `\fp:strict` flag, the compiler returns `+0` and moves along its happy way. That doesn't mean that we necessarily want to rely on this behavior as there are also performance considerations that may impact what we do.

We aren't currently handling overflow and, to-date, it has not been a problem. Maybe we can continue doing so?

We recommend (depending on what we decide about performance) that we either:
- use pragmas to temporarily disable floating point undeflow exeptions just for these instances of `std::exp` in debug builds on Windows, or
- replace `\fp:strict` with something a little more forgiving.

**Performance**

We benchmarked `std::exp(double)` against `epexp(double)` for inputs on either side of -70 (happy to share the benchmarking code if anyone wants to look at it). For each test we executed each function 10,000,000 times. We then ran the tests 1,000 times to get a distribution of timings. We then compared the minimum time for each test (as that is probably the best representation of an unloaded CPU). These results are from clang on macOS with `-O3` optimization.

| Function                            | std::exp(-71.0) |  epexp(-71.0) |  std::exp(370.0) |  epexp(370.0) | diff(-71.0) | diff(370.0) |
|-------------------------------------|-----------------|---------------|------------------|---------------|-----------|-------------|
| Min. time for 10,000,000 calls (ms) | 2,613           | 2,575         | 2,572            | 2,593         | -1.45%    | 0.82%       |

`epexp` was 1.45% faster for inputs less than -70, and 0.82% slower for inputs greater than -70.

My take away from this, is we should only be setting the lower limit on input if we anticipate more than 65% of the calls being less than -70. But, honestly, of the 1,000 times we ran the test, only 460 actually faster. We feel any difference in performance is in the noise.

Our recommendation would be to drop the input check altogether.

**DRY code**

The current structure is not DRY. While we could address the underlying divide-by-zero bug simply by adding another `if/else` block around the problematic code, that is not actually making the code better and does little to help prevent a similar bug in the future.

Regardless of whether we use `std::exp` or `epexp`, our recommendation would be to wrap the exponential function in a different function that handles the divid-by-zero **in a single location**. It doesn't really matter whether this replaces `epexp` or if we call it `epexp2` or `afn_exp`. We still recommend we use a function that looks something like this:

```
Real64 epexp2(Real64 numerator, Real64 denominator, Real64 defaultReturn=0.0)
{
    if (denominator == 0.0) {
        return defaultReturn;
    } else {
        return std::exp(numerator/denominator);
    }
}
```

If we wanted to make this more general yet (to handle quotient arguemtns in general), we could pass the function (e.g., `std::exp`) as an argument as well. That's probably not necessary, though.

**Having a useful set of functions**

We agree that it is good to build a general set of useful functions with better protections. However, given the discussion above, we would recommend changes to the `epexp` functions:

- The `epexp(double)` function is arguably (see above) unnecessary, and can maybe be removed.
- We see the utility of the `epexp(double, double)` function even though we are not currently handling overflow. We feel that this function could be designed better to remove the underflow protection and either take a default for `defaultHigh=700.0`, or simply hard-code it in the function (we still might need to protect from overflow for any subsequent operations on the return value). Maybe it should also be commented out until it is actually used (if for no other reason, to silence warnings from smart IDEs)?

jasondegraw:
@nealkruis, @ajscimone Responding to your points, and combining 3 and 4:

1. I don't believe that a minor AirflowNetwork issue is the place we should be discussing changing compiler flags for the whole code. To me that's a separate issue. I am not a fan of pragmas, and the team was happy with the solution that came out of the previous issue, so for me this one is settled in favor of the code as it is with the underflow protection.

2. The benchmarks here are just too close to "no change" for me to believe that the difference in performance is sufficient to make it a deciding factor.

3. and 4. Both of these come down to preference, and as @lgu1234 is the author of this code his preference should be given more weight. As I've indicated, I'd be fine with two functions, but Gu has indicated there's no need to remove the zero check, so that's what the way we should go. That puts us back at the code as it is with underflow protection.

I don't really want to expend too much more time on this, this PR has been reviewed and changes have been requested.

nealkruis:
I just pushed some changes that hopefully address the remaining concerns here. @jasondegraw, counter-responses here:

1. Yes, these changes are creeping beyond the original issue a bit, but they aren't unrelated. It sounds like the team didn't have a full picture regarding the previous issue at the time. Yes, the solution "works", but it was really treating a symptom (underflow in an `std::exp` call), when the actual problem is that our Windows debug builds are throwing exceptions on underflows at all.

    I removed the underflow check in `epexp` and instead [modified the floating point exception control-word bit value](https://github.com/NREL/EnergyPlus/pull/8202/files#diff-5112f14a467c1b13cf492f87273ae26fR60) to ignore underflow exceptions altogether. There are several types of floating point exceptions of varying severity, and for our purposes I can't imagine a situation where underflow is ever an issue we care about. The underflow exception is just saying that any values within the [underflow gap](https://en.wikipedia.org/wiki/Arithmetic_underflow#Underflow_gap) will be rounded to zero. A similarly innocuous floating point exception (that we are already ignoring) is "inexact", which is triggered anytime a floating point value can't be exactly represented (e.g., `1.0/10.0`). We have [similar settings on unix builds](https://github.com/NREL/EnergyPlus/blob/a9d719b4118bb141c2b0b04ce4e262f486655c92/src/EnergyPlus/api/EnergyPlusPgm.cc#L242) to ignore both of these types of exceptions, and I will do some follow up on a separate branch to clean these settings up to make them consistent across platforms. [Here's a helpful article](https://en.cppreference.com/w/cpp/numeric/math/math_errhandling) that describes the various types of floating point exceptions.

2. I agree, the performance difference is too in-the-noise to have any weight in decisions here.

3. @lgu1234 and I had a call yesterday to review the proposed changes. We decided that I would push up some of these changes to our proposal so that he can look at them closer. The [new version](https://github.com/NREL/EnergyPlus/pull/8202/files#diff-fb4d548d04a2beef6e2f7f4c4e748a9dR402) of the `epexp` function is set up to handle the divide-by-zero check that was already happening everywhere it was called in the code. This function is `inline` so that there is no functional or performance difference from repeating the zero-check in the code. This function makes the code DRYer and more unit-testable (especially for this particular bug). At @lgu1234's suggestion, I have also removed the default return argument in favor of always returning zero when the denominator is zero.

4. I left a [commented-out version](https://github.com/NREL/EnergyPlus/pull/8202/files#diff-fb4d548d04a2beef6e2f7f4c4e748a9dR412) of the function that handles overflow in case we decide we need to use it someday.

## Reproducing the original issue:

@lgu1234 you can see the branch off of v9.3.0-BugFix that @ajscimone created to demonstrate that the bug is resolved [here](https://github.com/bigladder/EnergyPlus/tree/fix-window-temperature-nan-calculations-demonstration). The code on this branch addresses the issue, but not up-to-date with what was just pushed. If necessary, we can cherry-pick the actual changes back to v9.3.0-BugFix to demonstrate that it still resolves the issue. Again, to reproduce the bug, you must use v9.3.0-BugFix on Windows. It is still not clear why triggering this bug in other builds is so difficult, but I don't think that is critical to resolving this PR (though it is still disconcerting).

Myoldmopar:
@lgu1234 would you mind taking a final review pass on this?

lgu1234:
@Myoldmopar I am going to take a look.

lgu1234:
@Myoldmopar @nealkruis @mjwitte 

The intent to create a function of EpExp is to eliminate underflow crash with std::exp using MSVC Debug version. The recent check-in from @nealkruis in EnergyPluspgm.cc seems to fix this issue (https://github.com/NREL/EnergyPlus/pull/8255)

#ifdef MSVC_DEBUG
    // the following line enables NaN detection in Visual Studio debug builds. See
    // https://github.com/NREL/EnergyPlus/wiki/Debugging-Tips
    int fp_control_state = _controlfp(_EM_INEXACT | _EM_UNDERFLOW, _MCW_EM); // These exceptions are disabled (_EM_INEXACT and _EM_UNDERFLOW will not throw)
#endif

I added 4 lines of code in the beginning of GetAirflowNetworkInput and tested with the Debug version using AirflowNetwork3zVent.idf:

        Real64 xx = std::exp(-100.0);
        Real64 xxx = std::exp(-1000.0);
        Real64 xxxx = std::exp(-10000.0);
        Real64 xxxxx = std::exp(-10000000.0);

The values are obtained during Debug:

xx = 3.7200759760208361e-44
xxx = 0.0
xxxx = 0.0
xxxxx = 0.0

The debug program ran to completion for the test file.  Since _EM_UNDERFLOW fixes underflow, is it possible to eliminate EpExp function and use std::exp instead?

nealkruis:
@lgu1234, yes, the idea here is that we no longer need the version of `epexp` that prevents underflow.

It is still useful to have a function that handles the divide-by-zero in a single place. I am repurposing the `epexp` function name to serve this purpose (though we could call it something else, too).

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -6859,20 +6859,12 @@ namespace AirflowNetworkBalanceManager {
                 }
 
                 if (!dataAirflowNetworkBalanceManager.LoopOnOffFlag(AirflowNetworkLinkageData(i).AirLoopNum) && AirflowNetworkLinkSimu(i).FLOW <= 0.0) {
-                    if (AirflowNetworkLinkSimu(i).FLOW2 > 0.0) {
-                        Ei = General::epexp(-UThermal * DuctSurfArea / (AirflowNetworkLinkSimu(i).FLOW2 * CpAir));
-                    } else {
-                        Ei = 0.0;
-                    }
+                    Ei = General::epexp(-UThermal * DuctSurfArea,  (AirflowNetworkLinkSimu(i).FLOW2 * CpAir));
```
Myoldmopar, line 6862: OK, I'm glad we aren't duplicating the `if` check.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -7158,12 +7142,8 @@ namespace AirflowNetworkBalanceManager {
                     LT = AirflowNetworkLinkageData(i).NodeNums[0];
                     DirSign = -1.0;
                 }
-                if (AirflowNetworkLinkSimu(i).FLOW == 0.0) {
-                    Ei = 0.0;
-                } else {
-                    Ei = General::epexp(-DisSysCompDuctData(TypeNum).UMoisture * DisSysCompDuctData(TypeNum).L *
-                                        DisSysCompDuctData(TypeNum).hydraulicDiameter * Pi / (DirSign * AirflowNetworkLinkSimu(i).FLOW));
-                }
+                Ei = General::epexp(-DisSysCompDuctData(TypeNum).UMoisture * DisSysCompDuctData(TypeNum).L *
```
Myoldmopar, line 7145: And I assume the `epexp` function is now giving back zero to match this behavior.

Myoldmopar (approved):
This looks fine.  I'm building it locally after resolving upstream conflicts.  If it builds and runs fine I'll push and merge it in.

Myoldmopar:
I don't have write access to push merge conflicts back here...so I made a new branch/PR and will merge that instead.

Myoldmopar:
Closed via #8288

PR #7936, opened 2020-04-16
Enable c++17 mode

lefticus:
This is a test to see about building with C++17 mode on the CI and what performance impact (or not) there might be in moving to `std::string_view` for common parameters and values for things like `RoutineName`

**Not ready for merging yet** (probably)

Fix #8103

PR #8211, opened 2020-08-14
Global Pointer-to-Implementation Redesign

mitchute:
### Pull Request Author

### Reviewer


mitchute:
@Myoldmopar I'm still having a problem here: https://github.com/NREL/EnergyPlus/blob/faa61e82b1d6c86e490e865da5f8bf7536a42134/src/EnergyPlus/api/runtime.cc#L66

Once that's resolved, we should get this merged in and work on moving the others over to the new design in smaller chunks.

mitchute:
My local Windows VM gave this the 👍 . Maybe CI needs some attention. In any case, I believe this is ready.

Myoldmopar:
Currently attempting a build on my Windows laptop, we'll see...

Myoldmopar:
Worked fine for me, I'm building a full package just to verify.  I'll restart CI on Windows and see what happens as well.  I suspect this will drop in tomorrow morning.

jasondegraw:
@mitchute @Myoldmopar Perhaps before this is merged there could be some discussion of what the plan is here? This isn't quite a tradition PIMPL design, as it looks rather much more open and isn't offering much in the way of safety. Is the plan to extend this to the rest of the global data? If so, is the plan to add more protection of data?

Myoldmopar:
The overarching plan is to collect the 3700+ global variables in EnergyPlus into a single structure, and carry a lightweight reference to that around the program.  There are multiple reasons why we are pursuing this, including being able to manage/persist/reset the state of the program, as well as being able to capture the global state of the program into a single thread-safe variable for running simulations in parallel.  Originally we were collecting global variables of each namespace into a small struct, and then keeping concrete instances of those classes on this high-level `EnergyPlusData` class.  However, this was leading down a dark road where all CC files depended on the EnergyPlusData.hh file, and that file depended on every other hh file.  It was leading to where changing *any* hh file would have caused a full rebuild, every time.  The dependency had to be broken, and the path chosen was to adapt the concrete instances to be unique pointers to each namespace's small struct collection of "global" variables.  Then the EnergyPlusData.hh file could include forward declarations rather than include all the headers, and we no longer have the dependency problem.  You can call it pimpl or anything else, all I'm trying to do is get rid of the chaos that is 3700+ global variables lying around the program.

jasondegraw:
That makes sense, but I guess I am not a huge fan of using unique_ptr like that. Those could be null and really require a check before dereferencing. In a discussion on another topic entirely, @lefticus pointed me to `not_null` in the Microsoft GSL here:

https://github.com/microsoft/GSL/blob/ef714fa49e43254ae9d448702e17a527e97d42e5/include/gsl/pointers#L63

Any thoughts on going a route like this that would avoid null ever being an issue?

mitchute:
@jasondegraw forgive my ignorance on this, but are you concerned that some developer is going to release these [```unique_ptr```s](https://github.com/NREL/EnergyPlus/pull/8211/files#diff-13e680eee4c56491e6c63cef1ac99c7dR58-R60) somewhere in their code? CI should catch that pretty quickly, I'd guess, since anything that references that global data is going to cause a hard crash. The state manager should be the only place where these are created or deleted, so no one should be messing with them outside of that. Or is there some other case you're thinking about that I'm not seeing?

jasondegraw:
@mitchute Well, dereferencing null pointer is a pretty big deal and it's better to have policies embodied in code rather than by CI. Not everyone pushes to the main repo all the time, and as things become more modular hopefully there will be more people than just the dev team using the code. There are "modern C++" ways of enforcing safety that won't require someone to know that they shouldn't touch something. Maybe this will help: what's the benefit of using `unique_ptr` here? I'm not seeing much, so why not just use a raw pointer? Sorry to be blunt, but raw pointers are considered bad practice these days, so why is this also not bad practice?

Myoldmopar:
I don't want to dismiss any concerns about using the unique pointers, but I think that @mitchute hit it there.  The pointers will only be created/destroyed by the state manager.  No one else should be managing it.  We will certainly be watchful of changes to allocate or clear these objects, but I think in general, developers will quickly get comfortable just using the passed around `state.dataNamespace->chillers(1)` in place of `EnergyPlus::Namespace::chillers(1)`.  And once we put in the effort to get the code adapted to this new form, we'll be back to normal-looking changesets as folks submit branches for defects or new features.

I agree that dereferencing a null pointer is a big deal, but there's really no reason why any one of these pointers would be null, unless someone is doing something really badly in their code changes -- which we cannot control.  As soon as EnergyPlus begins a simulation (or enters through the unit test entry), the state is constructed and these objects are populated.  It feels somehow _less_ dangerous than how we require a component to call GetInput prior to calling calculate.  If they call the calculate routine with an out of range index, it will dereference a null pointer and boom.  I think this state approach hides all the state management in one place - the state manager, making it easier to keep isolated and hopefully clarifying that it is an area of the code that others shouldn't need to play.

All that said, I definitely am open to suggestion.  If you have a design that breaks the dependency problem so the state header file won't depend on the _entire_ header state of EnergyPlus, without using pointers, I'm totally fine with it.

jasondegraw:
Sigh. You sure about that? I'm pretty sure I agreed with the overall premise in dependency management and pointed to a potential solution that involved pointers, so there's that. There are also accessor-based solutions that could accomplish similar things.

mitchute:
@Myoldmopar I took a look at GSL ```not_null``` as @jasondegraw suggested. This looks like a pretty useful feature and it would give us the ability to protect these pointers with a set of possible responses. These are posted below but taken from [here](https://www.bfilipek.com/2017/10/notnull.html).

```
// 1. GSL_TERMINATE_ON_CONTRACT_VIOLATION: 
//       std::terminate will be called (default)
// 2. GSL_THROW_ON_CONTRACT_VIOLATION: 
//       a gsl::fail_fast exception will be thrown
// 3. GSL_UNENFORCED_ON_CONTRACT_VIOLATION: 
//       nothing happens
```

In this case, we'd probably want to go with option 3, but that's just my guess.

The real problem, though, is that E+ is not CXX14 compliant. I pulled the GSL library into E+ and attempted to build anyway, but it failed. If incorporating ```not_null``` is the way we decide to go, we'd need to get #7936 merged first.

Myoldmopar:
@jasondegraw thanks for noting the `not_null` and thanks to @mitchute thank you for looking into it a little further.  It looks like it could be a really great thing as long as there are no runtime checks against null incurred upon every lookup.  We already have a few pointers in the code that are just hanging out everywhere (sqlite, inputprocessor), should we consider moving them to `not_null` at the same time?

Since the feature is apparently C++14, we can't do anything about that now, however it looks like we'll be able to have @lefticus on board and we could leverage that capability to make these sorts of great improvements to not just this part of the code but all sorts of other nooks and crannies.

The pointers used here are not accessible to anything via API.  The API methods do accept an instance of one of these big `Data` structures, but the API calls are simply couriering the instance around.  They can't call into the pointers that live on the struct.  I don't think they represent any more danger than the input processor instance that we carry around the guts of EnergyPlus.  I believe you are correct that we should make it better, but I think given all the circumstances, we should let this go in now and make a full pass over our pointer usage once we get C++14 brought in.

If anyone else has concerns please continue to speak up.  We could also open up a separate issue/document to track these issues so they are not lost once we are ready to go with C++14.

tst/EnergyPlus/unit/BaseboardRadiator.unit.cc
```diff
@@ -387,7 +387,7 @@ TEST_F(EnergyPlusFixture, BaseboardConvWater_SizingTest)
 
     DataZoneEquipment::GetZoneEquipmentData1(state);
     // get electric baseboard inputs
-    BaseboardRadiator::GetBaseboardInput(state.dataBaseboardRadiator);
+    BaseboardRadiator::GetBaseboardInput(*state.dataBaseboardRadiator);
```
Myoldmopar: I'm going to clean up these usages and just pass state around so it will be much cleaner.

Myoldmopar:
OK, got develop pulled in and fixed up some of the rough edges where we were passing child data classes around instead of the reference to the whole state.  I think this is ready to go if CI agrees.

PR #8213, opened 2020-08-16
Finish API/State Changes

Myoldmopar:
Pull request overview
---------------------
This pull request future-proofs the API calls in relation to the new `state` object.  This will break API calls, which is why I am doing it now, in the self-specified grace period for major API changes during this first release cycle after the API was added.

This also cleans up and really improves the API documentation.  I intend on making an "application guide" of some form to help guide users to using the API and Python Plugin system as well, but that can be closer to release time.


Myoldmopar:
Wonderfully clean on CI.  I'll give a walkthrough of the changes I am making here.

cmake/PythonSetupAPIinBuild.cmake
```diff
@@ -21,4 +21,5 @@ configure_file( "${API_SOURCE_DIR}/api.py" "${API_TARGET_DIR}/api.py" )
 configure_file( "${API_SOURCE_DIR}/func.py" "${API_TARGET_DIR}/func.py" )
 configure_file( "${API_SOURCE_DIR}/runtime.py" "${API_TARGET_DIR}/runtime.py" )
 configure_file( "${API_SOURCE_DIR}/plugin.py" "${API_TARGET_DIR}/plugin.py" )
+configure_file( "${API_SOURCE_DIR}/state.py" "${API_TARGET_DIR}/state.py" )
```
Myoldmopar, line 24: The new `state` object really needs to stand alone at the core of the EnergyPlus API, so I moved it to its own file.

src/EnergyPlus/PluginManager.cc
```diff
@@ -105,13 +105,18 @@ namespace PluginManagement {
...
     int (*EP_PyList_Check)(PyObjectWrap) = nullptr;
     unsigned long (*EP_PyList_Size)(PyObjectWrap) = nullptr;
     PyObjectWrap (*EP_PyList_GetItem)(PyObjectWrap, size_t) = nullptr;
     int (*EP_PyUnicode_Check)(PyObjectWrap) = nullptr;
     int (*EP_PyLong_Check)(PyObjectWrap) = nullptr;
     long (*EP_PyLong_AsLong)(PyObjectWrap) = nullptr;
     void (*EP_Py_SetPythonHome)(const wchar_t *) = nullptr;
+    PyObjectWrap (*EP_Py_BuildValue)(const char *) = nullptr;
+    PyObjectWrap (*EP_PyLong_FromVoidPtr)(void *) = nullptr;
+    PyObjectWrap (*EP_PyString_FromString)(const char *) = nullptr;
```
Myoldmopar, line 119: We are now passing a pointer to the current `state` instance into the plugin override functions, so that those functions can then pass that state around as well.  Needed to add a few new functions to call into the Python DLL appropriately.

src/EnergyPlus/PluginManager.cc
```diff
@@ -312,6 +327,21 @@ namespace PluginManagement {
...
+        *(void **)(&EP_PyLong_FromVoidPtr) = (void *)GetProcAddress((HINSTANCE)wrapperDLLHandle, "EP_Wrap_PyLong_FromVoidPtr");
+        if (!EP_PyLong_FromVoidPtr) {
+            ShowSevereError("Could not load Python function: EP_Wrap_PyLong_FromVoidPtr");
+            dll_errors = true;
+        }
+        *(void **)(&EP_PyString_FromString) = (void *)GetProcAddress((HINSTANCE)wrapperDLLHandle, "EP_Wrap_PyString_FromString");
+        if (!EP_PyString_FromString) {
+            ShowSevereError("Could not load Python function: EP_Wrap_PyString_FromString");
+            dll_errors = true;
+        }
```
Myoldmopar, line 344: Windows CI seemed to have no problem with these added calls.

src/EnergyPlus/PluginManager.cc
```diff
@@ -416,6 +452,15 @@ namespace PluginManagement {
...
         checkWrapperDLLFunction(err, "EP_Wrap_Py_SetPythonHome", dll_errors);
+        *(void **)(&EP_Py_BuildValue) = dlsym(wrapperDLLHandle, "EP_Wrap_Py_BuildValue");
+        err = dlerror();
+        checkWrapperDLLFunction(err, "EP_Wrap_Py_BuildValue", dll_errors);
+        *(void **)(&EP_PyLong_FromVoidPtr) = dlsym(wrapperDLLHandle, "EP_Wrap_PyLong_FromVoidPtr");
+        err = dlerror();
+        checkWrapperDLLFunction(err, "EP_Wrap_PyLong_FromVoidPtr", dll_errors);
+        *(void **)(&EP_PyString_FromString) = dlsym(wrapperDLLHandle, "EP_Wrap_PyString_FromString");
+        err = dlerror();
+        checkWrapperDLLFunction(err, "EP_Wrap_PyString_FromString", dll_errors);
```
Myoldmopar, line 463: Linux/Mac also liked this.

src/EnergyPlus/PluginManager.cc
```diff
@@ -454,7 +499,7 @@ namespace PluginManagement {
 #if LINK_WITH_PYTHON == 1
         for (auto &plugin : plugins) {
             if (plugin.runDuringWarmup || !DataGlobals::WarmupFlag) {
-                bool const didOneRun = plugin.run(iCalledFrom);
+                bool const didOneRun = plugin.run(state, iCalledFrom);
```
Myoldmopar, line 502: Since we pass the current `state` to the plugin functions, we have to pass `state` around to a few new places in the code -- not many though.

src/EnergyPlus/PluginManager.cc
```diff
@@ -1226,38 +1271,55 @@ namespace PluginManagement {
                 std::string functionName = (*EP_PyUnicode_AsUTF8)(item);
                 if (functionName == this->sHookBeginNewEnvironment) {
                     this->bHasBeginNewEnvironment = true;
+                    this->pBeginNewEnvironment = (*EP_PyString_FromString)(functionName.c_str());
```
Myoldmopar, line 1274: Store `PyObject` versions of the function names just once to avoid string operations later.

src/EnergyPlus/PluginManager.cc
```diff
@@ -1275,91 +1337,129 @@ namespace PluginManagement {
 #if LINK_WITH_PYTHON == 1
         (*EP_Py_DECREF)(this->pClassInstance);
         (*EP_Py_DECREF)(this->pModule); // PyImport_Import returns a new reference, decrement it
+        if (this->bHasBeginNewEnvironment) (*EP_Py_DECREF)(this->pBeginNewEnvironment);
```
Myoldmopar, line 1340: Decrement the reference count for "new" PyObject instances if they exist.

src/EnergyPlus/PluginManager.cc
```diff
@@ -1275,91 +1337,129 @@ namespace PluginManagement {
...
+        if (!pFunctionName) {
             return false;
         }
 
         // then call the main function
-        PyObjectWrap pFunctionResponse = (*EP_PyObject_CallMethod)(this->pClassInstance, functionName, nullptr);
+        //static const PyObjectWrap oneArgObjFormat = (*EP_Py_BuildValue)("O");
+        PyObjectWrap pStateInstance = (*EP_PyLong_FromVoidPtr)((void*)&state);
+        PyObjectWrap pFunctionResponse = (*EP_PyObject_CallMethod2ObjArg)(this->pClassInstance, pFunctionName, pStateInstance, nullptr);
+        (*EP_Py_DECREF)(pStateInstance);
```
Myoldmopar, line 1462: Call the appropriate Python plugin function and pass state to it.

src/EnergyPlus/PluginManager.hh
```diff
@@ -99,43 +95,65 @@ namespace PluginManagement {
...
+        bool bHasAfterPredictorAfterHVACManagers = false;
+        bool bHasInsideHVACSystemIterationLoop = false;
+        bool bHasEndOfZoneTimestepBeforeZoneReporting = false;
+        bool bHasEndOfZoneTimestepAfterZoneReporting = false;
+        bool bHasEndOfSystemTimestepBeforeHVACReporting = false;
+        bool bHasEndOfSystemTimestepAfterHVACReporting = false;
+        bool bHasEndOfZoneSizing = false;
+        bool bHasEndOfSystemSizing = false;
+        bool bHasAfterComponentInputReadIn = false;
+        bool bHasUserDefinedComponentModel = false;
```
Myoldmopar, line 133: I think I could actually get rid of these booleans now, and just use the PyObject variables to check for not `nullptr`, or _truthy._

src/EnergyPlus/PythonLibWrapper.hh
```diff
@@ -99,6 +101,9 @@ extern "C" {
     PYTHONWRAP_API long EP_Wrap_PyLong_AsLong(PyObject *obj);
     PYTHONWRAP_API void EP_Wrap_Py_SetPath(const wchar_t *);
     PYTHONWRAP_API void EP_Wrap_Py_SetPythonHome(const wchar_t *home);
+    PYTHONWRAP_API PyObject * EP_Wrap_Py_BuildValue(const char *format);
+    PYTHONWRAP_API PyObject * EP_Wrap_PyLong_FromVoidPtr(void *v);
+    PYTHONWRAP_API PyObject * EP_Wrap_PyString_FromString(const char *s);
```
Myoldmopar, line 106: New functions added to the Python shared library intermediary.

src/EnergyPlus/api/api.py
```diff
@@ -67,14 +68,10 @@ def __init__(self, running_as_python_plugin: bool = False):
...
+        self.api.apiVersionFromEPlus.argtypes = [c_void_p]
         self.api.apiVersionFromEPlus.restype = c_char_p
-        api_version_from_ep = float(self.api.apiVersionFromEPlus())
-        api_version_defined_here = float(self.api_version())
-        if api_version_defined_here != api_version_from_ep:
-            raise Exception("API version does not match, this API version: %s; E+ is expecting version: %s" % (
-                api_version_defined_here, api_version_from_ep
-            ))
+        # self.state provides access to the main EnergyPlus state management class, instantiated and ready to go
+        self.state = State(self.api)
```
Myoldmopar: Deferred calling the API version check because it needs `state`, and we can't get a `state` until after the API is created...

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -60,7 +61,7 @@
 #include <EnergyPlus/OutputProcessor.hh>
 #include <EnergyPlus/PluginManager.hh>
 
-char * listAllAPIDataCSV() {
+char * listAllAPIDataCSV(EnergyPlusState) {
```
Myoldmopar, line 64: Added `state` as an argument to all the API functions to future proof them.  Right now most of the functions don't end up using it, but they will once we stop getting everything off the global program state.

src/EnergyPlus/api/plugin.py
```diff
@@ -49,53 +47,53 @@ def _detect_overridden(self) -> List[str]:
                 diff.remove(known_skip)
         return diff
 
-    def on_begin_new_environment(self) -> int:
+    def on_begin_new_environment(self, state) -> int:
```
Myoldmopar, line 50: Added state as an argument to the plugin functions...all user-overridden functions will need to accept this argument, and then they can use it to call back into the API.

src/EnergyPlus/api/plugin_tester.py
```diff
@@ -200,7 +200,8 @@ def plugin_file_tester(file_path):
                         if func in expected_overrides:
                             method_to_call = getattr(plugin_instance, func)
                             try:
-                                response = method_to_call()
+                                dummy_state_arg = 0
+                                response = method_to_call(dummy_state_arg)
```
Myoldmopar, line 204: Modified the plugin tester to call the plugin method with an argument.

testfiles/PythonPluginAirflowNetworkOpeningControlByHumidity.py
```diff
@@ -9,16 +9,18 @@ def __init__(self):
         self.ZoneRH_handle = None
         self.MyOpenFactor_handle = None
 
-    def on_begin_timestep_before_predictor(self) -> int:
+    def on_begin_timestep_before_predictor(self, state) -> int:
```
Myoldmopar, line 12: The bulk of the changes from here on out are simply converting the plugin python files and the API test files to use the new state argument.

Myoldmopar (commented):
Walkthrough complete.

src/EnergyPlus/api/runtime.py
```diff
@@ -118,7 +121,10 @@ def run_energyplus(self, state: c_void_p, command_line_args: List[Union[str, byt
         cli_args = cli_arg_type(*args_with_program_name)
         return self.api.energyplus(state, len(args_with_program_name), cli_args)
 
-    def issue_warning(self, message: Union[str, bytes]) -> None:
+    # def stop_simulation(self, state: c_void_p) -> None:
```
mitchute, line 132: Keep this around?

mitchute (commented):
No problems from me on this. Unit tests pass locally. I'll attempt to run some other tests by calling E+ as a library.

src/EnergyPlus/api/runtime.py
Myoldmopar: I should probably keep it and just uncomment it, and wire it up to the underlying C function.  @kbenne 's PR will have some conflicts once this goes in, and while I'm resolving those, I'll uncomment this.  Good eye, though.

mitchute: There's also a couple of others in ```runtime.cc```. FYI

https://github.com/NREL/EnergyPlus/blob/5bf36f34bad6254a899b749d0c76a9d072c25124/src/EnergyPlus/api/runtime.cc#L269-L275

Myoldmopar: Another keen eye. Those are a little different.  The way the User-Defined components work is by taking the name of a program from the IDF and running it explicitly.  This didn't fit well with the callback approach we have for all the other calling points, so these aren't currently enabled.

We should take time after 9.4 to think about how to do component models really well.  For now these will just sit here as commented placeholders.

Myoldmopar:
OK, with these last commits, I cleaned up the state instance class and variable naming so that it is more clear that the object on the API is just an instance of the `StateManager` API category, not an actual instance of an underlying EnergyPlus `state`.  I also added a function to the runtime API class that will validate the number of arguments in the function you are trying to register.  If you try to register a function without any arguments, you will now get an informative TypeError during registration.  I think this will be ready to go in now.

mitchute:
Thanks for the cleanup, @Myoldmopar. Clarifying that it's a ```StateManager``` and not an actual instance of ```state``` is great, and the new error message is much better. 

I'm posting my library test file here FFR. 

```python
import sys

# insert path to build so Python can find pyenenergyplus
sys.path.insert(0, "cmake-build-release/Products")

from pyenergyplus.api import EnergyPlusAPI

# setup new API and state instances
api = EnergyPlusAPI()
state = api.state_manager.new_state()


# define method to override calling point
def my_before_timestep_before_predictor(state):
    # get handles
    zone_rh_handle = api.exchange.get_variable_handle(state,
                                                      "System Node Relative Humidity",
                                                      "Zone 1 Node")

    open_factor_handle = api.exchange.get_actuator_handle(state,
                                                          "AirFlow Network Window/Door Opening",
                                                          "Venting Opening Factor",
                                                          "Zn001:Wall001:Win001")

    # get value
    zone_rh = api.exchange.get_variable_value(state, zone_rh_handle)

    # calculate
    if zone_rh < 25.0:
        api.exchange.set_actuator_value(state, open_factor_handle, 0.0)
    elif zone_rh > 60.0:
        api.exchange.set_actuator_value(state, open_factor_handle, 1.0)
    else:
        api.exchange.set_actuator_value(state, open_factor_handle, (zone_rh - 25) / (60 - 25))


# override callback point with my method
api.runtime.callback_begin_system_timestep_before_predictor(state, my_before_timestep_before_predictor)

# execute simulation
idf_path = "testfiles/PythonPluginAirflowNetworkOpeningControlByHumidity.idf"
api.runtime.run_energyplus(state, ["-D", idf_path])
```

To run, remove the ```PythonPlugin:Instance``` object from the ```PythonPluginAirflowNetworkOpeningControlByHumidity.idf``` file and update the paths to the build and test file. 

Anything else you want to do before this merges? If not, I'll get it merged in ASAP.

mjwitte:
@Myoldmopar @mitchute Buildling with `LINK_WITH_PYTHON` unchecked fails [here](https://github.com/NREL/EnergyPlus/blob/9123217b2e042a136b6287d5fce1ed5f5e9f86db/src/EnergyPlus/PluginManager.cc#L1492-L1495).

Myoldmopar:
Gah!  I forgot CI doesn't always build with that OFF, only on Packages.  I'll fix it directly, please hold.

Myoldmopar:
Fixed in develop now, thanks for noting it so quickly @mjwitte !

mjwitte:
@Myoldmopar Thanks for the quick fix. Builds nicely here now.

PR #8215, opened 2020-08-18
Fix logic issue in special day calculation

Myoldmopar:
Pull request overview
---------------------
 - Fixes #8214
 - Marking as DoNotPublish since the defect was just introduced in #8157

Issue #8214, opened 2020-08-17
Special Days Failures


mitchute:
Issue overview
--------------
Special days are causing simulation failures. To test, run ```5ZoneVAV-ChilledWaterStorage-Stratified.idf``` with ```USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw```. This fails on current develop but runs with the 9.3.0 bugfix version. Confirmed by myself and *the* @mjwitte .

### Details
Some additional details for this issue (if relevant):
 - Platform (Mac 10.15.4)
 - Version of EnergyPlus (5bf36f34b)
 - Unmethours link or helpdesk ticket number - None

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] ~Defect file added (list location of defect file here)~
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

PR #8217, opened 2020-08-18
Add SpecialDays to Annual Run

mitchute:
Adding special days to an annual run.

### Pull Request Author

### Reviewer

### DIFF
--- a/testfiles/_5ZoneAirCooled_annual.idf
+++ b/testfiles/_5ZoneAirCooled_annual.idf
@@ -112,6 +112,68 @@
     25,                      !- Maximum Number of Warmup Days
     6;                       !- Minimum Number of Warmup Days
 
+! US National Holidays
+
+  RunPeriodControl:SpecialDays,
+    New Years Day,           !- Name
+    January 1,               !- Start Date
+    1,                       !- Duration {days}
+    Holiday;                 !- Special Day Type
+
+  RunPeriodControl:SpecialDays,
+    Veterans Day,            !- Name
+    November 11,             !- Start Date
+    1,                       !- Duration {days}
+    Holiday;                 !- Special Day Type
+
+  RunPeriodControl:SpecialDays,
+    Christmas,               !- Name
+    December 25,             !- Start Date
+    1,                       !- Duration {days}
+    Holiday;                 !- Special Day Type
+
+  RunPeriodControl:SpecialDays,
+    Independence Day,        !- Name
+    July 4,                  !- Start Date
+    1,                       !- Duration {days}
+    Holiday;                 !- Special Day Type
+
+  RunPeriodControl:SpecialDays,
+    MLK Day,                 !- Name
+    3rd Monday in January,   !- Start Date
+    1,                       !- Duration {days}
+    Holiday;                 !- Special Day Type
+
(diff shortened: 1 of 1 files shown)


Myoldmopar:
I've got no problem with this.  Does this suffice for you, @mjwitte?

mjwitte:
@mitchute Since #8215 is not merged yet, I expected this would error out. What's different in the https://github.com/NREL/EnergyPlus/issues/8214 defect file?

Myoldmopar:
Yeah that's a good point @mjwitte.  In my head it was because this was built on the other branch, but it's not.  So I'm not sure.

mitchute:
We should probably look into this a little more. This file was actually fine with special days. It was the stratified chilled water tank file that complained. Do we need to spend a little time investigating why, or maybe just add an annual run with that file?

mjwitte:
For CI efficiency, we'd rather not add another annual run. Let's figure out why this one ran fine and the other one failed. Of course, the special days fix is already merged, so we'll have to test that locally.

mitchute:
@mjwitte @Myoldmopar following up on this.

I've rolled back the code to play with why the ```_5ZoneAirCooled_annual.idf``` file (which *has* special days already) failed to catch this. We can make the file fail like the ```5ZoneVAV-ChilledWaterStorage-Stratified``` file by changing the run period from this:

```
  RunPeriod,
    RUNPERIOD 1,             !- Name
    1,                       !- Begin Month
    1,                       !- Begin Day of Month
    2019,                    !- Begin Year
    ...
    2019,                    !- End Year
    ,                        !- Day of Week for Start Day
    ...
```

to this:

```
  RunPeriod,
    RUNPERIOD 1,             !- Name
    7,                       !- Begin Month
    11,                      !- Begin Day of Month
    ,                        !- Begin Year
    ...
    ,                        !- End Year
    Sunday,                  !- Day of Week for Start Day
    ...
```

Failing to provide a year make the code pick a default year which changes the calculation of the ```MonWeekDay``` array (https://github.com/NREL/EnergyPlus/blob/c3fbac0a78cc1f13110035e495678114a14b6206/src/EnergyPlus/WeatherManager.hh#L333). I don't understand why changing the month or start day changes it...

That doesn't, to me at least, fully explain the root cause of the issue, but I'm not sure this warrants further investigation at this point. I think @nmerket  is going to be spending some time in this part of the code next FY, so maybe he will be adding unit tests to help catch issues like this in the future.

Let me know how you suggest we proceed.

mitchute:
Edit: Sorry, *this* PR is the reason ```_5ZoneAirCooled_annual``` has special days 🤦 

Regardless, even after adding special days, it still fails to catch the issue without the changes to the RunPeriod.

mjwitte:
@mitchute I think this is good to add (as modified) to catch any future problems (even if it's not a complete test).

Myoldmopar:
I second the motion to drop this in, and also to not spend too much time dwelling since it will hopefully have better unit testing added in soon.

Myoldmopar:
Thanks @mitchute, merging.

mjwitte:
Um, I don't think this had the enhanced RunPeriod object yet.

PR #8258, opened 2020-09-07
Correction of Flow Fraction Maximum for Variable Flow Radiant Systems

### LINKED ISSUE #8126: Upper Limit on Flow Fraction for Variable Flow Low Temperature Radiant Systems Not Implemented
Issue overview
--------------
While implementing on-off control for some of the radiant systems in #8044, it was noticed that for variable flow systems the flow fraction was not capped at 1.0 like in the electric system.  This seems to be an error.  If the user sets the maximum flow for this component, the system should not be able to get more than that flow.  By not capping the flow fraction at 1.0, the system continues and it is uncertain whether this should be allowed.  Technically, EnergyPlus could resolve a flow fraction greater than 1.0 by either limiting the flow to the user's maximum OR giving the component more flow if it's available.  It isn't clear what exactly it is doing so this probably needs to be investigated further as fixing this will cause big diffs in the output as was discovered in working on #8044.

### Details
 - Platform (Operating system, version): all
 - Version of EnergyPlus (if using an intermediate build, include SHA): current
 - Unmethours link or helpdesk ticket number: none

### Checklist
 - [x] Defect file added (list location of defect file here): no specific defect file--some of the existing files in the test suite exhibit differences (examples: RadLoTempHydrCtrlOpt, RadLoTempHydrCoolTower, and several others--see results from #8044 

RKStrand:
Pull request overview
---------------------
 - Fixes #8126
 - During recent enhancements to the variable flow low temperature radiant system model, it was noticed that the maximum flow fraction for this system was never limited to 1.0 of the maximum flow.  While it is uncertain at this point whether this was an oversight or a conscious decision that the flow resolver would prevent the maximum flow from being eclipsed, what is clear now is that this was a bug and that the flow resolver in fact did not resolve this properly.  In fact, a look at the results showed that though the flow at the radiant system inlet node was being correctly set to the maximum when the flow was set higher than the maximum, the incorrect flow value that was higher than the maximum based on user input was being passed along and was impacting the supply side inlet node.  This resulted in a flow and energy mismatch between the two sides of the loop that was not being trapped.  This was fixed by not allowing the flow fraction in the variable flow radiant system to not go past the maximum value or limiting the flow fraction to 1.0.  This is similar to the electric radiant system which also limits the total power to the system to the maximum established by user input.  This change to the code now results in a flow rate match between the radiant system demands and the supply side.  As part of this fix, the code in the flow fraction setting algorithm was corrected, a new unit test specifically testing what happens when more flow than the maximum is requested was added, and modification of all of the variable flow radiant system IDF/RVI files to now also output the supply side flow rate were made.

It should be noted that changes to the output/results are anticipated with this change because the plant supply side flow rate is now correct.  Before it was just plain wrong and the energy consumed by the plant has also changed as a result.  The changes are not trivial and will likely impact ALL of the hydronic radiant IDFs (the ones that have variable flow systems).  These changes range from small to large, but the key thing is that the results are now correct whereas before they were in violation of the conservation of mass (and energy) in the plant loop.

### Pull Request Author
 - [upon request] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [x] If any diffs are expected, author must demonstrate they are justified using plots and descriptions (see description above, plots can be generated if needed)
 - [x] If changes fix a defect, the fix should be demonstrated in plots and descriptions (see previous item in this list)
 - [NA] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [NA] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [NA] If structural output changes, add to output rules file and add OutputChange label
 - [NA] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer


testfiles/RadLoHydrHeatCoolAuto.idf
```diff
@@ -1971,47 +1971,49 @@
     ,                        !- Number of Circuits
     ;                        !- Circuit Length {m}
 
-  Output:Variable,*,Zone Mean Air Temperature,timestep;
+  Output:Variable,*,Zone Mean Air Temperature, detailed;
```
mitchute: Are these changes a remnant from debugging?

testfiles/RadLoHydrHeatCoolAutoCondFD.idf
```diff
@@ -2066,45 +2066,47 @@
 
 !-   ===========  ALL OBJECTS IN CLASS: OUTPUT:VARIABLE ===========
 
-  Output:Variable,*,Zone Mean Air Temperature,timestep;
+  Output:Variable,*,Zone Mean Air Temperature, detailed;
```
mitchute: Same here, and for others as well.

tst/EnergyPlus/unit/LowTempRadiantSystem.unit.cc
```diff
@@ -2193,37 +2193,63 @@ TEST_F(LowTempRadiantSystemTest, calculateOperationalFractionTest)
     functionResult = thisRadSys.calculateOperationalFraction(offTemperature, controlTemperature, throttlingRange);
     EXPECT_NEAR(expectedResult, functionResult, 0.001);
 
-    // Test 3a: Temperature Difference is not zero and positive, throttling range is non-zero but less than temperature difference
+    // Test 3a: Temperature Difference is not zero and positive, throttling range is non-zero but greater than temperature difference
```
mitchute, line 2196: Unit tests look good.

src/EnergyPlus/LowTempRadiantSystem.cc
```diff
@@ -5325,7 +5324,8 @@ namespace LowTempRadiantSystem {
         } else if (throttlingRange < 0.001) {
             return 1.0; // Throttling range is essentially zero and there is a temperature difference--turn it full on
         } else {
-            return temperatureDifference/throttlingRange;   // Temperature difference is non-zero and less than the throttling range--calculate the operation fraction
+            // Temperature difference is non-zero and less than the throttling range--calculate the operation fraction, but limit to a maximum of 1.0
```
mitchute, line 5327: Yeah, that's definitely a bug.

mitchute (approved):
This looks good. Unit tests all pass locally. Let me know your thoughts on my comments, but otherwise, this is ready.

testfiles/RadLoHydrHeatCoolAuto.idf
RKStrand: Thanks, @mitchute, for reviewing this PR!
Yeah, the input file changes were made during testing and debugging so that I could see what was going on all the time during the run.  I left them like this because I figured that it couldn't hurt, but the IDFs could also be flipped back to "timestep" if you would prefer that.  If so, let me know and I'll flip them back.  I'm okay with this either way and don't have any strong preference.

testfiles/RadLoHydrHeatCoolAutoCondFD.idf
RKStrand: See previous comment.

testfiles/RadLoHydrHeatCoolAuto.idf
mitchute: I don't have an opinion either way. @Myoldmopar @mjwitte what's your preference?

mjwitte:
I lean slightly towards putting these back to timestep. The newly added output variables can stay.

mitchute:
OK. Go ahead with that last change @RKStrand. We'll get this in after that.

RKStrand:
Ok, sounds good, @mitchute.  I'll switch them all back to timestep.  I'll also make the new output variable timestep also for consistency.  Once that's done, I'll commit and let you know.  Thanks!

RKStrand:
@mitchute Ok, I have changed everything back to "timestep" from "detailed", merged in develop, and made a commit.  Should be ready to go assuming CI comes back the same as it was before.

mitchute:
Thanks, @RKStrand! I'll run tests locally one more time, but I expect this is ready to go.

mitchute:
As expected, lots of diffs in the hydronics files. This is ready, merging.

RKStrand:
Thanks @mitchute!

PR #8266, opened 2020-09-09
Steam conversion factor

### LINKED ISSUE #8265: Site-source conversion factor for steam is wrong
Issue overview
--------------
The default site-source conversion factor for steam in the ABUPS report might be wrong. With all defaults, the value is 0.300 for steam. This seems more like the conversion efficiency of steam. Other sources put the site-source factor at 1.20-1.35.

image

### Details
 - Version of EnergyPlus v9.4 in develop

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #8265 
 - This pull request updates the site-to-source conversion factor variable `sourceFactorSteam` from 0.3 to the correct value of 1.20 in the Site to Source Energy Conversion Factors Table to match the intent of the `Pollution.SteamConvEffic` variable in the PollutionModule.cc file.  The Total and Net Source Energy are calculated from this `sourceFactorSteam` variable.  The `fuelSourceFactor` variable for steam also needs to take into account the `Pollution.SteamConvEffic` variable to calculate the correct Source Energy End Use for District Heating when steam is used.  Below shows the before and after for the relevant source energy use tables using an `Exterior:FuelEquipment` object with constant 1,000W steam use for the January.

Also, Steam is not included as an option in the `FuelFactors` object for the user to define it.

image

Big Table diffs are expected as this is changing the Site to Source Energy Conversion Factors for every html table output file.

### Pull Request Author

### Reviewer


nealkruis:
@JasonGlazer I think you worked on adding these [originally](https://github.com/NREL/EnergyPlusArchive/commit/afc44acee9dc324732cb78fc71eea14cb31d9732#diff-56646bd42b8fbf08715db98bde23ff66R1137)? Do you have any insight here?

If we do make this change, shouldn't we update the other factors here while we're at it?

This might require transition rules (so old files use the same values as the old defaults), and [documentation](https://bigladdersoftware.com/epx/docs/9-3/input-output-reference/input-for-output.html#environmentalimpactfactors) changes.

JasonGlazer:
I don't specifically remember but I would guess that it is based on the assumption of utilization of waste heat that is making electricity from a generator. It is possible that it needs to be that value to compensate for another conversion factor in the chain. Sorry, it was a long time ago.

Before you make such a change, have you made sure that it functions like you would expect when you put in that conversion efficiency in an input file? Have you checked the code and how it is used?

matthew-larson:
The steam factor seems to only be used for calculating the source energy use, but I will do some additional testing along with the CI checks to make sure it checks out with using an input file as well.

matthew-larson:
So it looks like this input (sourceFactorSteam) and the value in the conversion table have no effect on the models.  The steam energy use ends up under the District Heating category and uses the efficiencyDistrictHeating variable, which is equal to 0.3.  Need to do a little more digging on this to determine the best path forward.

image

image

matthew-larson:
Looks like the sourceFactorSteam variable is used directly for the Total and Net Source Energy for the Site and Source Energy table but a factor of 0.3 (no matter the sourceFactorSteam variable value) is used in the Source Energy End Use Components Summary report tables in being bundled with District Heating.  We can probably leave it in the District Heating category if desired but just make sure the correct factor is being applied for steam vs district heating.  The district heating site energy is using the natural gas source factor divided by 0.3 to calculate the source energy.

JasonGlazer (approved):

JasonGlazer:
I did a code review and the changes look fine. @nealkruis  and @mitchute please take a look.

mitchute (approved):
Looks good to me.

Myoldmopar:
Seems reasonable to me as well, merge at will.

PR #8269, opened 2020-09-10
Allow Multiple Scheduled Shades to Reference a Single Window

JasonGlazer:
Pull request overview
---------------------
Allow modeling a window with different shading coefficient based on schedule (such as during the summer and winter). This is based on a requirement in "ANSI/RESNET/ICC 301-2014 Standard for the Calculation and Labeling of the Energy Performance of Low-Rise Residential Buildings" which requires (see page 14) that glazing should have an interior shade coefficient of 0.70 in the summer and 0.85 in the winter. 

Here is the [new feature proposal.](https://github.com/NREL/EnergyPlus/blob/AllowMultipleScheduledShadesB/design/FY2020/NFP-MultipleScheduledShades.md)

This is a continuation of the work under PR #8196 . Merging into develop was difficult due to #8141 so I just moved my code into the latest develop rather than trying to resolve all merge conflicts.

### Pull Request Author

### Reviewer


JasonGlazer:
@mjwitte and @Myoldmopar I think this is ready for another review.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10366,6 +10380,23 @@ namespace SolarShading {
...
 
+    int selectActiveWindowShadingControlIndex(int curSurface)
+    {
+        // For a given surface, determine based on the schedules which index to the window shading control list vector should be active
+        int selected = 0; // presume it is the first shading control - even if it is not active it needs to be some shading control which is then turned off in the WindowShadingManager
+        if (Surface(curSurface).windowShadingControlList.size() > 1) {
+            for (std::size_t listIndex = 0; listIndex < Surface(curSurface).windowShadingControlList.size(); ++listIndex) {
+                int wsc = Surface(curSurface).windowShadingControlList[listIndex];
+                //pick the first WindowShadingControl that has a non-zero schedule value
+                if (ScheduleManager::GetCurrentScheduleValue(WindowShadingControl(wsc).Schedule) != 0.0) {
```
mjwitte: Should this test be `if >0.0`? I/O Ref description for WindowShadingControl Schedule Name says "If schedule value > 0 , shading control is active, ..."

doc/input-output-reference/src/overview/group-thermal-zone-description-geometry.tex
```diff
@@ -4115,7 +4115,9 @@ \subsection{WindowShadingControl}\label{windowpropertyshadingcontrol}
...
+
+To deploy one shading device during one time period and another during another time period for the same window, specify two different WindowShadingControl objects that both reference the same fenestration.  Each of the WindowShadingControl input objects would include the same window on their lists of fenestration objects, but would reference different schedule input objects. The referenced schedule input objects should have non-zero values specified at different times for each WindowShadingControl. This approach can be used to add shading that is different for different seasons. When more than one WindowShadingControl is used for the same window based on schedule, the type of shading, control type, setpoints, and several other parameters must be the same in the different WindowShadingControl objects, but the shading device and construction can be different.
```
mjwitte, line 4120: Do all of these constraints still apply?

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -8616,6 +8614,42 @@ namespace SurfaceGeometry {
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    bool isWindowShadingControlSimilar(int a, int b)
+    {
+        // Compares two window shading controls are the same except for the name, schedule name, construction, and material
```
mjwitte, line 8639: Oh, I see, this is still enforcing all of the same constraints. I guess we can work on relaxing that after v9.4.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -2085,22 +2084,13 @@ namespace SurfaceGeometry {
...
-                if (Surface(SurfNum).ShadedConstruction != 0) SurfWinShadedConstruction(SurfNum) = Surface(SurfNum).ShadedConstruction;
-
-                // no need to set the below -- it is the default
-                // Set variable that indicates if shading device has movable slats
-                //      SurfaceWindow(SurfNum)%MovableSlats = .FALSE.
-
-                // TH 2/9/2010. Fixed for CR 8010 for speed up purpose rather than fixing the problem
-                WinShadingControlPtr = Surface(SurfNum).WindowShadingControlPtr;
                 if (Surface(SurfNum).HasShadeControl) {
+                    WinShadingControlPtr = Surface(SurfNum).activeWindowShadingControl; // use first item since others should be identical
```
mjwitte, line 2088: I think this section needs to cycle through the `windowShadingControlList` in case one of them uses a different type of blind.

mjwitte (commented):
@JasonGlazer Some questions/concerns.

src/EnergyPlus/SolarShading.cc
JasonGlazer: Good catch. I will change

src/EnergyPlus/SurfaceGeometry.cc
JasonGlazer: I'm not sure, the ShadingType and SlatAngleControlForBlinds are two of the items that need to be the same currently in the isWIndowShadingControlSimilar() function. There is only one SurfWinBlindNumber so this may have to stay as the same across WIndowShadingControls.

mjwitte: OK. I suppose `SurfWinBlindNumber` could be changed to `SurfWinActiveBlindNumber` and we could add a `blindNumberList` to `Surface`. But this can wait for a Phase II where we can fully test relaxing all of these matching requirements.

Myoldmopar (approved):
From a code perspective, this is pretty much exactly what I would have expected for the given changes. Lots of going from a single integer value to instead looking for an integer from a list.  I'll build locally.

Myoldmopar:
CI is happy and it builds and run unit tests fine locally.

Is there an example file for this feature?

mjwitte:
There is no example file. Would you prefer a new one or modify an existing one?

Myoldmopar:
Modify an existing one is preferred if it's reasonable.  One less to maintain...

Myoldmopar:
I'm inclined to merge this in sooner rather than later to avoid conflicts popping up in it.  If you feel like there will be an example file shortly then I'll hold off, otherwise I'll take this as-is.

Myoldmopar:
I built locally with develop pulled in again and unit tests all pass.

I then took the PurchAirWindowBlind test file and converted it to be a scheduled shade instead.  I then added a second scheduled shade control.  I do see differences in results, not exactly what I was expecting, but I think it's probably more due to me trying to cobble together a test case in short order.  No errors arose and it all went fine, so I'm going to merge this in.  If a test case can be developed in a standalone PR tomorrow that would be great, but if not the users will have to go by the documentation changes, which are suitable.

PR #8272, opened 2020-09-14
Fix Zone Equipment sensible heat gain summary calculation and reporting

### LINKED ISSUE #6794: Sensible Heat Gain Summary showing positive values in Zone Equipment Cooling column
Issue overview
--------------
The sensible heat gain summary (SensibleHeatGainSummary table report) is intended to show positive values for heating (heat addition) and negative values for cooling (heat removal). But the user's file has positive values in the column for "HVAC Zone Eq & Other Sensible Air Cooling".

A similar results can be seen with an annual run of example file HVACTemplate-5ZoneFanCoil-DOAS.idf which shows a positive value for "HVAC Zone Eq & Other Sensible Air Cooling" for Space5-1.

1. It appears that there is a problem sorting out heating and cooling contributions from zone equipment vs air loop terminal units.
The problem is [here](https://github.com/NREL/EnergyPlus/blob/eb68ccb2bfb3d01807412e33399299da8cdb4b11/src/EnergyPlus/OutputReportTabular.cc#L6407-L6410). The zone equipment contribution is not being separated every time step, but the split is being done at the end of the simulation:
ZoneEqCooling = NetHVACCooling - AirTerminalCooling.  Likewise for heating. So, over the course of the year, the DOAS air terminals are overcooling the zones and the zone equipment is providing heating to offset the overcooling.

2. The Input Output Reference documentation for the "Sensible Heat Gain Summary" report lists the output variables used to generate the report values, The documentation still lists the original "HVAC Input Sensible Air Heating" (and cooling) as columns 1 and 2, but these were later split into four columns, two for zone equipment and 2 for air terminals. The documentation needs to be updated to match.

### Details
 - Version of EnergyPlus v8.9.0
 - Unmethours [27574](https://unmethours.com/question/27574/clarification-needed-report-sensible-heat-gain-summary/)

### Checklist

jcyuan2020:
Pull request overview
---------------------
 - Fixes #6794 
 - This pull request fixed the positive sensible heat request reported and discussed in Issue #6794. In the original code, the calculation of the Zone Equipment Sensible Heat and Cool numbers are calculated after an annual (or the full simulation period) values of Total HVAC sensible heat/cool gain and the ATU (Air Terminal Unit) sensible heat/cool gain were calculated. As was discovered and discussed in the Unmet Hours ticket refereed in Issue #6794, this will cause some heating and cooling rate could be canceled out since it could be common that a Zone Equipment or an ATU to provide heating at some particular time, but then in another time provides cooling at another. A positive cooling number could then show up with the annual difference method applied in the original code. The attached case file in Issue #6794 showed such an example. 
- The heating and cooling of the Zone Equipment heating and cooling should be determined and accumulated at each HVAC system time-step level. The current Pull Request provided such a fix to carry out the Zone Equipment HVAC heating or cooling contributions splitting at the HVAC time step (a.k.a the system time-step level). This change will allow the heating and cooling summation of the Zone HVAC equipment to be calculated correctly. When the fix was applied to the code and the attached example case file for Issue #6794, the Zone Equipment sensible cooling energy report no longer shows any positive value. 

### Pull Request Author

### Reviewer


src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5242,9 +5248,28 @@ namespace OutputReportTabular {
...
+        for (iunit = 1; iunit <= NumAirDistUnits; ++iunit) {
+            // HVAC equipment should already have the multipliers included, no "* mult" needed (assumes autosized or multiplied hard-sized air flow).
+            curZone = AirDistUnit(iunit).ZoneNum;
+            if ((curZone > 0) && (curZone <= NumOfZones)) {
+                ZonePreDefRep(curZone).SHGSAnHvacATUHt += AirDistUnit(iunit).HeatGain;
+                ATUDeltaHeat(curZone) += AirDistUnit(iunit).HeatGain;
+                ZonePreDefRep(curZone).SHGSAnHvacATUCl -= AirDistUnit(iunit).CoolGain;
+                ATUDeltaCool(curZone) -= AirDistUnit(iunit).CoolGain;
+                ATUHeat(curZone) = AirDistUnit(iunit).HeatRate;
+                ATUCool(curZone) = -AirDistUnit(iunit).CoolRate;
```
mjwitte: These need to accumulate `+=` and `-=` because there can be more than one `AirDistUnit` in a zone. This is something I missed back when adding the ability to serve a zone with more than one airloop.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5254,12 +5279,13 @@ namespace OutputReportTabular {
             ZonePreDefRep(iZone).SHGSAnLiteAdd += ZnRpt(iZone).LtsTotGain * mult * timeStepRatio;
             // HVAC Input Sensible Air Heating
             // HVAC Input Sensible Air Cooling
-            if ((ZnAirRpt(iZone).SumMCpDTsystem + ZnAirRpt(iZone).SumNonAirSystem) > 0.0) {
-                ZonePreDefRep(iZone).SHGSAnHvacHt +=
-                    ZnAirRpt(iZone).SumMCpDTsystem * TimeStepSys * SecInHour + ZnAirRpt(iZone).SumNonAirSystem * mult * TimeStepSys * SecInHour;
+            ZoneEqHeatorCool(iZone) =
+                ZnAirRpt(iZone).SumMCpDTsystem * TimeStepSys * SecInHour + ZnAirRpt(iZone).SumNonAirSystem * mult * TimeStepSys * SecInHour 
+                - ATUDeltaHeat(iZone) - ATUDeltaCool(iZone);
```
mjwitte: If `TimeStepSys * SecInHour` is removed from this line, then `ATUHeat` and `ATUCool` [W] can be used here and we don't need the new `ATUDeltaHeat` and `ATUDeltaCool` variables.
Also, `ZoneEqHeatorCool` is just used locally in this loop, so it does not need to be an array.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5254,12 +5279,13 @@ namespace OutputReportTabular {
...
+            if (ZoneEqHeatorCool(iZone) > 0.0) {
+                ZonePreDefRep(iZone).SHGSAnZoneEqHt += ZoneEqHeatorCool(iZone);
```
mjwitte: Add `TimeStepSys * SecInHour` here and in the `else`.

mjwitte (commented):
@jcyuan2020 A few initial comments.

src/EnergyPlus/OutputReportTabular.cc
jcyuan2020: @mjwitte Just a cautionary check: should the `ZnAirRpt(iZone).SumMCpDTsystem` be multiplied by a zone/zone-list multiplier `mult` the same way as `ZnAirRpt(iZone).SumNonAirSystem` does in Line 5283 (and in a few other places across that file as well)? Or the multiplier for the `SumMCpDTsystem` has been taken care of someplace else?

mjwitte: That's a very good question, this can be quite confusing. Everything related to air-side HVAC should already be multiplied, `SumMCpDTsystem` as calculated in  [`ZoneTempPredictorCorrector::CalcZoneComponentLoadSums`](https://github.com/NREL/EnergyPlus/blob/7badc378a280879046ae72d47b0b12e90e5ea128/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L6497) is the full value (already multiplied).  But `CalcZoneComponentLoadSums` does not account for multipliers when [calculating the optional zone heat balance imbalance](https://github.com/NREL/EnergyPlus/blob/7badc378a280879046ae72d47b0b12e90e5ea128/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L6672-L6673). Hmmm. And the I/O Reference for Zone Air Heat Balance System Air Transfer Rate says "This field is not multiplied ... `. 

Looks like `SumMCpDTsystem` should be divided by the zone multipliers the same way it is in [`CalcZoneSums`](https://github.com/NREL/EnergyPlus/blob/7badc378a280879046ae72d47b0b12e90e5ea128/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L6252-L6253). I wouldn't touch that for this PR. I would make some test files with and without multipliers (fully autosized) to prove that it's wrong for the Zone Air Heat Balance System Air Transfer Rate output variable and wrong for the imbalance ( Output:Diagnostics, DisplayZoneAirHeatBalanceOffBalance;) then post a new issue to clean this up. If I'm following everything correctly, the zone sensible heat gains report should work fine with or without multipliers as it is now.

src/EnergyPlus/OutputReportTabular.cc
jcyuan2020: This one has now been addressed.

src/EnergyPlus/OutputReportTabular.cc
jcyuan2020: This one has now been addressed as well.

src/EnergyPlus/OutputReportTabular.cc
jcyuan2020: @mjwitte Thanks! Yes, this new issue will not affect the fix in the current PR. The code revisions for the original comments are done now.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5214,6 +5214,7 @@ namespace OutputReportTabular {
         static Array1D<Real64> radiantCool;
         static Array1D<Real64> ATUHeat;
         static Array1D<Real64> ATUCool;
+        static Real64 ZoneEqHeatorCool(0.0);
```
mjwitte: Nit-picking, but we don't want to declare this here, and we don't want it to be static. Just declare it in the loop as noted below.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5242,9 +5243,23 @@ namespace OutputReportTabular {
         // clear the ATU accumulation variables
         ATUHeat = 0.0;
         ATUCool = 0.0;
+        // clear the temporary Zone Eq accumulatoin variable
+        ZoneEqHeatorCool = 0.0;
```
mjwitte: These lines can be deleted.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5254,12 +5269,12 @@ namespace OutputReportTabular {
             ZonePreDefRep(iZone).SHGSAnLiteAdd += ZnRpt(iZone).LtsTotGain * mult * timeStepRatio;
             // HVAC Input Sensible Air Heating
             // HVAC Input Sensible Air Cooling
-            if ((ZnAirRpt(iZone).SumMCpDTsystem + ZnAirRpt(iZone).SumNonAirSystem) > 0.0) {
-                ZonePreDefRep(iZone).SHGSAnHvacHt +=
-                    ZnAirRpt(iZone).SumMCpDTsystem * TimeStepSys * SecInHour + ZnAirRpt(iZone).SumNonAirSystem * mult * TimeStepSys * SecInHour;
+            ZoneEqHeatorCool =
```
mjwitte: Add `Real64` here to keep the scope tight.

JasonGlazer (approved):
Just reviewed the code changes and they look good.

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
```diff
@@ -8278,3 +8279,43 @@ TEST_F(EnergyPlusFixture, OutputReportTabularTest_GetDelaySequencesSurfaceOrder_
...
+    EnergyPlus::DataHeatBalance::ZoneWinHeatGainRepEnergy.allocate(1);
+    EnergyPlus::DataHeatBalance::ZoneWinHeatLossRepEnergy.allocate(1);
+    
+    GatherHeatGainReport(OutputProcessor::TimeStepType::TimeStepSystem);
+
+    EXPECT_EQ(1.0*(EnergyPlus::DataHVACGlobals::TimeStepSys)*SecInHour, DataHeatBalance::ZonePreDefRep(1).SHGSAnZoneEqHt);
+    EXPECT_EQ(0.0*(EnergyPlus::DataHVACGlobals::TimeStepSys)*SecInHour, DataHeatBalance::ZonePreDefRep(1).SHGSAnZoneEqCl);
+    EXPECT_EQ(1000.0, DataHeatBalance::ZonePreDefRep(1).SHGSAnHvacATUHt);
+    EXPECT_EQ(-2000.0, DataHeatBalance::ZonePreDefRep(1).SHGSAnHvacATUCl);
+}
```
Myoldmopar, line 8336: Would this unit test have failed before?  It's not a bad test, but I can't tell that it actually would fail to aggregate properly in the before case.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -5245,6 +5245,18 @@ namespace OutputReportTabular {
         //--------------------
         //     ANNUAL
         //--------------------
+        // HVAC annual heating by ATU
```
Myoldmopar, line 5248: So what actually ended up causing the fix here?  Was it moving this block up above the zone loop?  Or aggregating ATUHeat using `+=` instead of just an assignment?  Or the altered calculation of `total` below?

Myoldmopar (commented):
I think this _could_ be ready, but I could use a little more clarification on how this actually fixed the issue.

Myoldmopar:
There's a very minor merge conflict (two branches added a new unit test to the same spot).  If more commits are needed, go ahead and resolve it.  But if no other commits are needed, I'll resolve and test locally and push just before merging.

src/EnergyPlus/OutputReportTabular.cc
jcyuan2020: The actual fix is that the aggregation for the ZoneEqHt and ZoneEqCl now occurs in the `GatherHeatGainReport()` function at the HVAC time step level, instead of originally calculated by a subtraction of two annual sum in the `FillRemainingPredefinedEntries()` after all the time step accumulation was done (around Line 6764-6766 in `OutputReportTabular.cc`). 
`
\-        PreDefTableEntry(pdchSHGSAnHvacHt, "Total Facility", (totalHvacHt - totalHvacATUHt) * convertJtoGJ, 3);
\-        PreDefTableEntry(pdchSHGSAnHvacCl, "Total Facility", (totalHvacCl - totalHvacATUCl) * convertJtoGJ, 3);
\+       PreDefTableEntry(pdchSHGSAnZoneEqHt, "Total Facility", totalZoneEqHt * convertJtoGJ, 3);
\+       PreDefTableEntry(pdchSHGSAnZoneEqCl, "Total Facility", totalZoneEqCl * convertJtoGJ, 3); 
`

The original method would cause some "canceling" heating and cooling not counted toward the total sum. The revised method does this accumulation of heating and cooling at system time step level, so that the two are clearly separated and logged in for each system time step, where both heating and cooling accumulate to get their totals. 

The block order change (ATU block being moved up) serves as facilitating this accumulating this change in the time step level, in that the ATU heat and cool for each zone needs to be known before the ZoneEq can be calculated through a split subtraction. This can be done without moving the blocks, but would then require creating a few new temporary storing variables and adding an extra block to do the splitting after both the zone block and the ATU block finish.

mjwitte: @Myoldmopar Regarding 

> Or aggregating ATUHeat using += instead of just an assignment?

That change actually fixes a different bug where this report was wrong for any zone with more than one terminal unit. But I never posted a new issue for that or created a test file to prove it.

tst/EnergyPlus/unit/OutputReportTabular.unit.cc
jcyuan2020: It would be difficult to demonstrate in the test case on how it fails. The test unit was trying to assure that the major change works as it expected. For showing the difference, the originally attached file named `HVACTemplate-5ZoneFanCoil-DOAS.idf` in the associated issue # 6794 would show the correction---where original it showed an unexpected positive cooling energy in one zone but after the change the cooling energy values in all the zones are negative as expected.

JasonGlazer: Just a suggestion, maybe a unit test could call GatherHeatGainReport several times with positive and negative values and then show that they no longer cancel each other out.

Myoldmopar: That's what I was thinking too :+1:

jcyuan2020: Thanks! I will revise that part to reflect the aggregation. Btw, I solved the conflict and pushed the message before seeing this message. Will upload again after the test change is done.

Myoldmopar: @jcyuan2020 do not worry about pushing that test fix up right now.  We need to be getting rid of PRs in a hurry, and this is good to go.  Update the unit test whenever you make your next commits after the release.  Thanks.

Myoldmopar:
I'm feeling convinced.  I appreciate you all taking the time to clarify the changes here and how they impact the results.  I'm going to merge this one in.  Thanks!

jcyuan2020:
> 
> 
> I'm feeling convinced. I appreciate you all taking the time to clarify the changes here and how they impact the results. I'm going to merge this one in. Thanks!

Thank you!

PR #8296, opened 2020-09-22
Documentation changes - #7760

### LINKED ISSUE #7760: Documentation Issues - 2020a
Issue overview
--------------
A collection of small documentation issues. See checkboxed comments below.

jmythms:
Pull request overview
---------------------
 - Fixes #7760
 - A few fixes to correct documentation issues

### Pull Request Author

### Reviewer


mjwitte:
@jmythms Is this ready for review?

jmythms:
@mjwitte Yes, It would be great if you could review it for me. Thank you.

doc/input-output-reference/src/overview/group-airflow-network.tex
```diff
@@ -1,4 +1,4 @@
-\section{Group -- Airflow Network}\label{group-airflow-network}
+\section{Group -- Natural Ventilation and Duct Leakage}\label{group-natural-ventilation-and-duct-leakage}
```
mjwitte: Per @jasondegraw 's https://github.com/NREL/EnergyPlus/issues/7760#issuecomment-585356366 we should revert this change and update the IDD post v9.4.

doc/input-output-reference/src/overview/group-water-heaters.tex
```diff
@@ -1,4 +1,4 @@
-\section{Group -- Water Heaters}\label{group-water-heaters}
+\section{Group -- Water Heaters and Thermal Storage}\label{group-water-heaters}
```
mjwitte, line 1: Moving this is OK.

doc/input-output-reference/src/overview/group-humidifiers-and-dehumidifiers.tex
```diff
@@ -1,4 +1,288 @@
-\section{Group Desiccant Dehumidifiers}\label{group-desiccant-dehumidifiers}
+\section{Group -- Humidifiers and Dehumidifiers}\label{group-humidifiers-and-dehumidifiers}
```
mjwitte, line 1: Combining these two groups is Ok.

doc/input-output-reference/src/overview/group-air-distribution-equipment.tex
```diff
@@ -1,4 +1,4 @@
-\section{Group -- Air Distribution Equipment}\label{group-air-distribution-equipment}
+\section{Group -- Zone HVAC Air Loop Terminal Units}\label{group-zone-HVAC-air-loop-terminal-units}
```
mjwitte, line 1: I'm not a big fan of this group name - but there's also a group call "Air Distribution" which includes the main air loop components, so this name helps distinguish the terminal units.

doc/engineering-reference/src/loop-equipment-sizing-and-other-design-data/component-sizing.tex
```diff
@@ -649,6 +649,52 @@ \subsubsection{Rated Sensible Cooling Capacity}\label{rated-sensible-cooling-cap
...
+CPC2 = user input for Cooling Power Consumption Coefficient 2
+
+CPC3 = user input for Cooling Power Consumption Coefficient 3
+
+CPC4 = user input for Cooling Power Consumption Coefficient 4
+
+CPC5 = user input for Cooling Power Consumption Coefficient 5
+
+
+The 4\(^{th}\) and 5\(^{th}\) coefficient (CPC4 and CPC5) used in the above equation are multipliers for the load-side and source-side flow ratios, respectively. For sizing, these ratios are assumed to be 1.
```
mjwitte: I see you've followed the pattern of the other two sets of coefficients above, but all of these should include the VLratio and VSratio terms in the equations (even if the term is assumed to be 1 for sizing). These same equations are used during the simulation if I understand correctly.

doc/engineering-reference/src/loop-equipment-sizing-and-other-design-data/component-sizing.tex
```diff
@@ -716,9 +762,102 @@ \subsubsection{Rated Water Flow Rate}\label{rated-water-flow-rate-2}
...
 The calculation is identical to that done for \emph{Coil:Cooling:Water} , which is the coil design load divided by the \emph{Loop Design Temperature Difference} user input from \emph{Sizing:Plant.} For sizing the plant loop serving this coil, only one half of this flow rate is used since both the cooling and heating coil will save a flow rate but only one of these coils will operate at a time.
 
+%\subsubsection{Rated Total Heating Capacity}\label{rated-total-heating-capacity}
+
+%The rated total heating capacity is set equal to the rated total cooling capacity.
+
 \subsubsection{Rated Total Heating Capacity}\label{rated-total-heating-capacity}
 
-The rated total heating capacity is set equal to the rated total cooling capacity.
+The following calculations are performed to determine the rated total heating capacity.
```
mjwitte: Hmm, are you sure it's not set equal? Looks like the code sets [`RatedCapHeatDes = DXCoolCap;` here](https://github.com/NREL/EnergyPlus/blob/b1c7a5b4634ed662ef93c23b195ba084a09df3c6/src/EnergyPlus/WaterToAirHeatPumpSimple.cc#L1894). 
Actually, all of the coefficients are already explained in the Engineering Reference section titled "Single Speed Equation-Fit Model". Maybe not real obvious, but A1-F5 are noted after the equations and then identified in the following section "Coefficient estimation procedure". So the best remedy might be to add some text in the IO Reference pointing to these sections in the Engineering Reference.

mjwitte (commented):
This is ok, but the entire IO Ref could use an outline restructuring independent of the IDD order. There are sections (like EMS) that are plopped in the middle of other related groups. Not suggesting we change anything more here. Most of these changes are OK for now.

Same for the Engineering Reference (6 identical chapter headings of "Simulation Models - Encyclopedic Reference"?).

Regarding @jasondegraw 's question about groups in epJSON. The \group names are not currently preserved in the schema, but the objects do appear to be in the same order as the IDD. How that order plays out in tools for editing epJSON remains to be seen (some grouping as a navigational aid would be helpful IMHO).

There are some items checked off in #7760 that are not fixed here (I unchecked them):
https://github.com/NREL/EnergyPlus/issues/7760#issuecomment-594215711 
https://github.com/NREL/EnergyPlus/issues/7760#issuecomment-595318925

doc/engineering-reference/src/loop-equipment-sizing-and-other-design-data/component-sizing.tex
jmythms: You are right, my mistake. I got confused by the heating mode simulation code [here](https://github.com/NREL/EnergyPlus/blob/b1c7a5b4634ed662ef93c23b195ba084a09df3c6/src/EnergyPlus/WaterToAirHeatPumpSimple.cc#L2595).
Thanks, I'll make the changes in the IO reference.

jasondegraw:
@mjwitte Ordering within schemas is a tricky thing, so if there's going to be some grouping going forward we probably should discuss that sooner rather than later. It might be worth a separate issue.

mjwitte (approved):
@jmythms Thanks for addressing comments. This looks good and will merge after CI completes its rounds.

jmythms:
> @jmythms Thanks for addressing comments. This looks good and will merge after CI completes its rounds.

Thank you.

PR #8298, opened 2020-09-23
Parametric Preprocessor and EP-Launch error handling cleanup

mjwitte:
Pull request overview
---------------------
 - Fixes #6573 - cleanup error messages in Parametric Preprocessor
   See https://github.com/NREL/EnergyPlus/issues/6573#issuecomment-697503289 for testfiles.
 - Fixes #8251 - Adjust EP-Launch to not rely on the presence of eplusout.end to determine run completion status.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [ ] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [ ] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #8307, opened 2020-09-25
OutputControl documentation and bugfix for tabular file control

mbadams5:
Pull request overview
---------------------
This adds documentation for new OutputControl:Files object.
This also changes the approach for tabular files so you can turn off tabular file(s) output but still get tabular results in sqlite.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/EconomicTariff.cc
+++ b/src/EnergyPlus/EconomicTariff.cc
@@ -2898,6 +2898,11 @@ namespace EconomicTariff {
         Real64 annualAggregate;
         int annualCnt;
 
+        if (!(ioFiles.outputControl.tabular || ioFiles.outputControl.sqlite)) {
+            WriteTabularFiles = false;
+            return;
+        }
+
         hugeValue = HUGE_(Real64());
         //  Clear the isEvaluated flags for all economics variables.
         for (nVar = 1; nVar <= numEconVar; ++nVar) {
--- a/src/EnergyPlus/EconomicTariff.cc
+++ b/src/EnergyPlus/EconomicTariff.cc
@@ -2898,11 +2898,6 @@ namespace EconomicTariff {
         Real64 annualAggregate;
         int annualCnt;
 
-        if (!ioFiles.outputControl.tabular) {
-            WriteTabularFiles = false;
-            return;
-        }
-
         hugeValue = HUGE_(Real64());
         //  Clear the isEvaluated flags for all economics variables.
         for (nVar = 1; nVar <= numEconVar; ++nVar) {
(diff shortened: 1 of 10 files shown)


src/EnergyPlus/EconomicTariff.cc
```diff
@@ -2898,11 +2898,6 @@ namespace EconomicTariff {
         Real64 annualAggregate;
         int annualCnt;
 
-        if (!ioFiles.outputControl.tabular) {
-            WriteTabularFiles = false;
```
mjwitte, line 2902: @mbadams5 If I'm following this correctly, one side effect of these changes is that `WriteTabularFiles` will be `true` if there are any triggers in the idf file regardless of the output file settings. This flag is used to skip a bunch of timestep calls to [gather functions here](https://github.com/NREL/EnergyPlus/blob/b1c7a5b4634ed662ef93c23b195ba084a09df3c6/src/EnergyPlus/OutputReportTabular.cc#L792-L806). 

Is there a place after all of the Output* input processing is complete that we could test if "Output Tabular = No" and "Output SQLite = No" then set `WriteTabularFiles = false` ?

mbadams5: Yeah, that is an interesting idea. I think that is possible and I'll try it out.

mjwitte:
@mbadams5 There are no doc changes in this branch?

mbadams5:
@mjwitte I'm working on them now, I just pushed up the code changes last night to have CI chew on it.

mbadams5:
@mjwitte I pushed up my documentation changes and addressed your other comment. I think this should be ready to go. I tested the tabular and sqlite output permutations locally but having another set of eyes on it would be good.

mjwitte:
@mbadams5 This is in great shape. 

During testing of various combinations of Tabular=Yes/No, SQlite = Yes/No, Output:SQlite object present in idf (or not), I noticed messages to std:output about "Writing tabular output file results" or "Writing final SQL reports" at times when it shouldn't be. Also, saw some tabular report data gathering that wasn't needed.

I've pushed some changes to prevent almost all of that. e.g. outputSQlite = yes, but there is no sqlite object in the input file (which is kind of the default case) the sqlite instance is not created and the message about Writing SQL reports does not happen. Likewise if there's an sqlite object and sqlite=true, but output tabular = false, then the message about writing tabular output does not happen.

There's one combination I ran out of time to figure out: With Output:SQLite, Simple; and Output Tabular = no, then `WriteTabularFiles` is still true, so the Gathers in `UpdateTabularReports` are still executing for that case. But if 9.4 goes out like that, it's just fine.

Myoldmopar (approved):
This looks great to me.  Good docs and some code cleanup along with the fixes.  I did test it briefly locally as part of reviewing #8298 and it worked fine, but I haven't tried a gauntlet of variations yet.

Myoldmopar:
@mjwitte I think between your testing various combinations and me testing a couple combinations during other testing, this should be good to go.  Merging.

PR #8308, opened 2020-09-25
Remove python wrapper intermediary

Myoldmopar:


.github/workflows/mac_release.yml
```diff
@@ -13,13 +13,9 @@ jobs:
   mac_release:
     runs-on: macos-10.15
 
-    strategy:
-      matrix:
-        osx_target: ["10.13", "10.15"]
-
```
Myoldmopar, line 19: Indeed, with this PR we are going to just be building 10.15 packages, at least on Github Actions.  This doesn't limit us to never make a 10.13 package, but the 10.13 packages coming out of Github Actions 10.15 machines was flawed.  A developer with a 10.13 machine (or 10.14) could certainly build a full release package and we can test it and post it with our list of installers.

cmake/PythonFixUpOnMac.cmake
```diff
@@ -1,53 +1,75 @@
...
+        execute_process(COMMAND "install_name_tool" -change "${PREREQ}" "@executable_path/${PYTHON_LIB_FILENAME}" "${EXECUTABLE_PATH}")
+    endif()
+endforeach()
+
+# for the energyplus dylib, search for the python dylib prereq and change it to use @loader_path
+get_prerequisites("${ENERGYPLUS_API_PATH}" PREREQUISITES 1 1 "" "")
+foreach(PREREQ IN LISTS PREREQUISITES)
+    string(FIND "${PREREQ}" "${PYTHON_LIB_FILENAME}" PYTHON_IN_PREREQ)
+    if (NOT PYTHON_IN_PREREQ EQUAL -1)
+        execute_process(COMMAND "install_name_tool" -change "${PREREQ}" "@loader_path/${PYTHON_LIB_FILENAME}" "${ENERGYPLUS_API_PATH}")
```
Myoldmopar, line 62: Using @loader_path was 1 core piece to this.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -826,21 +829,6 @@ install( TARGETS energyplusapi
   ARCHIVE DESTINATION ./
 )
 
-# we will build the intermediate wrapper DLL that hard links to Python DLL here
```
Myoldmopar, line 829: Cleaning out the little Python wrapper blob from the build process and runtime process is the second important piece of this.

src/EnergyPlus/PluginManager.hh
```diff
@@ -135,27 +147,27 @@ namespace PluginManagement {
         bool bHasUserDefinedComponentModel = false;
         bool bHasUnitarySystemSizing = false;
 #if LINK_WITH_PYTHON
-        PyObjectWrap pModule = nullptr;  // reference to module
-        PyObjectWrap pClassInstance = nullptr; // reference to instantiated class -- *don't decref until the end of the simulation*
+        PyObject *pModule = nullptr;  // reference to module
```
Myoldmopar, line 150: No more *Wrap objects necessary, we can now call directly into the Python DLL again.

src/EnergyPlus/PluginManager.cc
```diff
@@ -78,396 +72,6 @@ namespace PluginManagement {
     bool fullyReady = false;
     bool apiErrorFlag = false;
 
-    // function pointers into the PythonWrapper DLL
```
Myoldmopar, line 81: That's right, 656 lines removed from `PluginManager.cc`

Myoldmopar (commented):
OK, this PR should now be ready to merge.  The core storyline:
- Mac 10.15 does not allow dlopen due to the hardened runtime, unless your binaries and package are signed and notarized.  We'll be looking into integrating that into our build process anyway to make a nicer install experience, but in either case, we never needed to be doing it anyway.
- The underlying assumption with how we've been "fixing up" executable blobs on Mac has been that the executable starting the process would _always_ be energyplus(.exe).  This has been fine for years, but now that we are calling into the EnergyPlus dylib as a library, the starting executable could be anywhere on the system.  This required an investigation into the Mac fixup process and ultimately led to a solution.
- The Github Actions Mac instances are only available in 10.15.  This is one clue that supporting older versions is not super necessary (plus the near 2/3 market share of 10.15 which will continue to climb).  In any case, using 10.15 to build packages targeting 10.13 on Github Actions is flawed, and results in a package that is trying to find a system declaration that is not (always) present on 10.13.  So we are only building 10.15 packages on Github Actions.  A developer with a 10.13 machine could make a full release package and we can test it and post it no problem.

OK, as long as CI is happy here this should be ready to go.

Myoldmopar:
Well, this went through _how many_ RC tests before everything was passing happily.  I am not going to expect anyone else to dig in to this change.  I'm merging this and will tag RC 10 tonight.  I expect 2 more PRs to drop in tomorrow, which will be followed up by RC 11 which will likely be the final SHA and tagged v9.4.0.  At least that's what I'm going to tell myself as I drift off to sleep tonight.

PR #8320, opened 2020-09-29
Fix Inlet and Outlet Node idf errors with AirLoopHVAC:UnitarySystem objects cause are not caught, resulting in faulty data members.

### LINKED ISSUE #8305: Inlet and Outlet Node idf errors with AirLoopHVAC:UnitarySystem objects cause are not caught, resulting in faulty data members.
Issue overview
--------------
When using an *AirLoopHVAC:UnitarySystem*, the error detection for inlet and outlet nodes do not seem to catch mismatched nodes, and instead result in the following error far downstream in the simulation:  

`** Severe  ** Duplicate internal variable was sent to SetupEMSInternalVariable.`

For example, in the attached idf, there is a typo in one of the *NodeList* items here:

    NodeList,
        Bath_ZN_1_FLR_1 Exhaust Nodes,  !- Name
        Bath_ZN_1_FLR_1 ZN-PTAC Inlet Node;  !- Node 1 Name

idf: [in.idf.txt](https://github.com/NREL/EnergyPlus/files/6145467/in.idf.txt)

The problem with this object is the *Node 1* name is misspelled. Instead of printing out that the there is some mismatch in the inlet and exhaust nodes, we instead see these other strange errors much further downstream in the simulation.

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): 9.3

### Checklist

ajscimone:
Pull request overview
---------------------
 - Fixes #8305
 - This pull request fixes a parsing issue in the Unitary System which fails to detect idf issues where the inlet and exhaust node names are misspelled or do not match. Previously, Unitary System would not find the proper equipment to set in the `getUnitarySystemInputData` function, but would not throw an error. This would result in errors being thrown for duplicate EMSInternalVariables downstream in the simulation. Now, when these mismatched node names do not allow the Unitary System to set the equipment properly, it will throw a severe error.

### Pull Request Author

### Reviewer


src/EnergyPlus/UnitarySystem.cc
```diff
@@ -3327,6 +3327,13 @@ namespace UnitarySystems {
                                 ZoneInletNodeFound = true;
                                 break;
                             }
+                        } else{
```
rraustad: This will be difficult to fix given the number of things that could be wrong. Either the inlet or outlet node could be misspelled, or both. The control zone name could be incorrect or misspelled but the node names are correct. The unit could be in the air loop and load controlled (i.e., has a control zone name) and it will hit this else looking for the unit location (I just checked using 5Zone_Unitary_HXAssistedCoil). So although this would work for the defect file, the fact that the UnitarySystem can be zone, air loop or OA system equipment makes this much harder.

rraustad:
These changes look good. Great functions.

Defect error file in this branch:

    ** Severe  ** Input errors for AirLoopHVAC:UnitarySystem:DOAS UNITARY
    **   ~~~   ** Incorrect or misspelled Air Outlet Node Name = DOAS UNITARY OUTLET NODE
    **   ~~~   ** Node name does not match any controlled zone inlet node name. Check ZoneHVAC:EquipmentConnections object inputs.
    **  Fatal  ** getUnitarySystemInputData: previous errors cause termination. Check inputs

I tried unsuccessfully to fix the error and I see:

    AirLoopHVAC,
      DOAS,                    !- Name
      ,                        !- Controller List Name
      DOAS Availability Manager List,  !- Availability Manager List Name
      Autosize,                !- Design Supply Air Flow Rate {m3/s}
      DOAS Branch List,        !- Branch List Name
      DOAS Supply Equipment Inlet Node,  !- Supply Side Inlet Node Name
      DOAS Zone Equipment Outlet Node,  !- Demand Side Outlet Node Name
      DOAS Zone Equipment Inlet Nodes,  !- Demand Side Inlet Node Names
      DOAS Equipment Outlet Nodes;  !- Supply Side Outlet Node Names

    NodeList,
      DOAS Equipment Outlet Nodes,  !- Name
      DOAS Unitary Outlet Node;!- Node 1 Name

That is the correct spelling of "DOAS Unitary Outlet Node". 

    AirLoopHVAC:UnitarySystem,
      DOAS Unitary,            !- Name
      ,                        !- Controlling Zone or Thermostat Location
      None,                    !- Dehumidification Control Type
      DOAS Unitary Operation Schedule,  !- Availability Schedule Name
      DOAS Unitary Inlet Node, !- Air Inlet Node Name
      DOAS Unitary Outlet Node,!- Air Outlet Node Name
 
So a NodeList isn't allowed for Supply Side Outlet Node Names? Or accounted for at the warning? What would change in the defect file to correct this warning?

matthew-larson:
@rraustad I'm taking over for Tony on this one and made a number of fixes and adjustments to get clean CI results so I'm marking this ready for review.

nrel-bot-2c:
@matthew-larson @ajscimone @lgentile it has been 28 days since this pull request was last updated.

rraustad:
@matthew-larson I ran the defect file in #8305 and the sim fails during plant sizing with runaway plant temps. I stepped into the UnitarySystem and see that ZoneExhaustNodeFound is false for Bath_ZN_1_FLR_1 ZN-PTAC Unitary, as it should be.

    AirLoopHVAC:UnitarySystem,
      Bath_ZN_1_FLR_1 ZN-PTAC Unitary,  !- Name
      Bath_ZN_1_FLR_1 ZN-PTAC Unitary Inlet Node,  !- Air Inlet Node Name
      Bath_ZN_1_FLR_1 ZN-PTAC Unitary Outlet Node,  !- Air Outlet Node Name

    ZoneHVAC:EquipmentConnections,
      Bath_ZN_1_FLR_1,         !- Zone Name
      Bath_ZN_1_FLR_1 Equipment,  !- Zone Conditioning Equipment List Name
      Bath_ZN_1_FLR_1 Inlet Nodes,  !- Zone Air Inlet Node or NodeList Name
      Bath_ZN_1_FLR_1 Exhaust Nodes,  !- Zone Air Exhaust Node or NodeList Name

    NodeList,
      Bath_ZN_1_FLR_1 Inlet Nodes,  !- Name
      Bath_ZN_1_FLR_1 ZN-PTAC Unitary Outlet Node;  !- Node 1 Name

    NodeList,
      Bath_ZN_1_FLR_1 Exhaust Nodes,  !- Name
      Bath_ZN_1_FLR_1 ZN-PTAC Inlet Node;  !- Node 1 Name  **<< incorrect node name**

However, there were no warnings in the error file wrt this UnitarySystem. So I moved up the call to `searchZoneInletNodesByEquipmentIndex` and see that ZoneInletNodeFound is true, so that equipment should have been found and warned against a node name issue. This also makes me think of a problem when both inlet and outlet names are wrong where warnings would not be posted. 

image

This change is important to the UnitarySystem in that it can be used as zone, airloop, OA system and OA unit equipment. I could be wrong but think only zone and airloop example files exist so we need to be sure the UnitarySystem finds the node connection issues in all those configurations. If the UnitarySystem is not found in any of these configurations a warning needs to be issued. 

I don't recall if or how the UnitarySystem was identified when used in the OA unit but I do know it is supposed to work. OutdoorAirUnit*.idf doesn't have a UnitarySystem, maybe one should be added in one of those example files. Maybe that configuration is beyond the scope of this issue and would require a new issue to correct that specific configuration if necessary.

    ZoneHVAC:OutdoorAirUnit:EquipmentList,
       \memo Equipment list for components in a ZoneHVAC:OutdoorAirUnit. Components are simulated
       \memo sequentially in the order given in the equipment list.
    A1 , \field Name
       \required-field
       \type alpha
       \reference OutdoorAirUnitEquipmentLists
    A2 , \field Component 1 Object Type
       \type choice
       \key Coil:Heating:Electric
       \key Coil:Heating:Fuel
       \key Coil:Heating:Steam
       \key Coil:Heating:Water
       \key Coil:Cooling:Water
       \key Coil:Cooling:Water:DetailedGeometry
       \key CoilSystem:Cooling:Water:HeatexchangerAssisted
       \key CoilSystem:Cooling:DX
       \key CoilSystem:Heating:DX
       \key HeatExchanger:AirToAir:FlatPlate
       \key HeatExchanger:AirToAir:SensibleAndLatent
       \key Dehumidifier:Desiccant:NoFans
       \key AirLoopHVAC:UnitarySystem

mjwitte:
@rraustad It seems that a better way to approach this would be to instantiate unitary systems one at a time when called from their parent. Then it could know exactly what configuration it should be confirming and not have to do all of this searching to find matches. Something to consider for a future cleanup.

rraustad:
@mjwitte point me to an example so I understand what you are suggesting.

mjwitte:
I don't see an example yet. Most places with factories are filling the vector for all instances. 

But if we look at [`UnitarySys::factory`](https://github.com/NREL/EnergyPlus/blob/637ea911c08986d9ffe0df06e7140b819888ed02/src/EnergyPlus/UnitarySystem.cc#L430-L431) it already has arguments for `ZoneEquipment` and `ZoneOAUnitNum`:
```
    HVACSystemData *UnitarySys::factory(
        EnergyPlusData &state, int const object_type_of_num, std::string const objectName, bool const ZoneEquipment, int const ZoneOAUnitNum)
    {
        if (state.dataUnitarySystems->getInputOnceFlag) {
            UnitarySys::getUnitarySystemInput(state, objectName, ZoneEquipment, ZoneOAUnitNum);
            state.dataUnitarySystems->getInputOnceFlag = false;
        }
```
`UnitarySys::getUnitarySystemInput` calls `UnitarySys::getUnitarySystemInputData` and that loops over *all* unitary system inputs and tries to process them. I guess the current logic is trying to instantiate all of the unitary system objects but only process inputs for the one that's been called? I'm having trouble following all of the logic for `getInputOnceFlag` etc. 

Looking at the call in [`SimAirServingZone`](https://github.com/NREL/EnergyPlus/blob/637ea911c08986d9ffe0df06e7140b819888ed02/src/EnergyPlus/SimAirServingZones.cc#L1191-L1199), more arguments could be added to the factory to include `AirSysNum` and `BranchNum` and `CompNum` so the unitary system can know exactly what node names to check without searching through all air loops and branches etc. I assume the call from OA system or zone equipment could also pass useful indexes.

I did find one example, I think, [Fan:SystemModel is done one at a time](https://github.com/NREL/EnergyPlus/blob/637ea911c08986d9ffe0df06e7140b819888ed02/src/EnergyPlus/SimAirServingZones.cc#L1147-L1158), but as I recall we didn't choose to follow this model for Coi:Cooling:DX.

nealkruis:
@rraustad @mjwitte It sounds like maybe we can continue this topic on a new/separate issue?

rraustad:
It seems the conversation would be the same on a different issue. Thinking about what @mjwitte said regarding new arguments, there are already these in the call to the sim function, which are 3 of the 4 possible configurations. So we already know all we need to know and can better target the configuration to look for node connections.

    void UnitarySys::simulate(EnergyPlusData &state, std::string const &Name,
                              int const &AirLoopNum,
                              int const ZoneOAUnitNum,
                              bool const ZoneEquipment,

Example of ZoneEquipmentManager

                        int AirLoopNum = 0;
                        int OAUnitNum = 0;
                        bool ZoneEquipFlag = true;
                        ZoneEquipList(CurZoneEqNum).compPointer[EquipPtr]
                            ->simulate(state, PrioritySimOrder(EquipTypeNum).EquipName, ...

Example in SimAirServingZones

        int OAUnitNum = 0;           // used only for UnitarySystem call
        bool ZoneEquipFlag = false;  // used only for UnitarySystem call

                CompPointer->simulate(state,
                                      CompName,
                                      FirstHVACIteration,
                                      AirLoopNum,

Example in MixedAir (which looks strange but I didn't spend time to see if anything was actually wrong)

                    bool ZoneEquipFlag = false;
                    OutsideAirSys(OASysNum).compPointer[CompIndex]->simulate(state, CompName,
                                                                             FirstHVACIteration,
                                                                             AirLoopNum,
                                                                             CompIndex,
                                                                             HeatingActive,
                                                                             CoolingActive,
                                                                             CompIndex,
                                                                             OAUCoilOutTemp,
                                                                             ZoneEquipFlag,
                                                                             sensOut,
                                                                             latOut);

Example in OutdoorAirUnit

                    OutAirUnit(OAUnitNum)
                        .OAEquip(SimCompNum)
                        .compPointer->simulate(state, EquipName,
                                               FirstHVACIteration,
                                               -1,
                                               DXSystemIndex,
                                               HeatActive,
                                               CoolActive,
                                               UnitNum,
                                               Dxsystemouttemp,
                                               false,
                                               sensOut,
                                               latOut);

nealkruis:
@rraustad I'm trying to figure out if something needs to be figured out still before @matthew-larson can resolve this defect.

matthew-larson:
@rraustad I merged in the latest develop and fixed all merge conflicts.  I also tested the fix with the defect file and it appears to work as expected.  I've reattached the defect file and weather file below.

[in.epw.txt](https://github.com/NREL/EnergyPlus/files/5913187/in.epw.txt)
[in.idf.txt](https://github.com/NREL/EnergyPlus/files/5913191/in.idf.txt)

rraustad:
I get the same result as I did 25 days ago [here](https://github.com/NREL/EnergyPlus/pull/8320#issuecomment-756836486) with no node connection warnings and a fatal with plant temps too  high.

You have to check for both nodes, the inlet and exhaust. Or change getInput as @mjwitte described and instantiate each individually, which would be more work.

Also, are you seeing these strange water temps? That looks like a separate issue. I've seen this before but don't know what causes it.

    **   ~~~   ** PlantLoop Outlet Node (DemandSide) "DHW DEMAND OUTLET NODE" has temperature=76892694952034048935729515505954623222351074218750000000000000000000000000000000000000000000000000000000000000000000000000. {C}

matthew-larson:
@rraustad I was able to recreate your error with the DHW when running it through Debug on Windows, this error never showed up on Mac.  Since this is unrelated to the issue and I was unsuccessful in eliminating it, I just removed the DHW system from the model.  Here is the working version of the in.idf.  Now that you can see the node errors that are now being generated, I will circle back on your comment regarding checking both inlet and exhaust nodes.  Let me know if you have any additional comments after testing this updated defect file.  Thanks again for all your help and feedback with this issue.

[in.idf.txt](https://github.com/NREL/EnergyPlus/files/5927271/in.idf.txt)

Myoldmopar:
I pulled develop into this branch and built it and ran tests.  Everything went just fine.  If I recall from the iteration meeting, this was going to be OK to go in with the possibility of additional follow-up work later.  If that's correct then I'll get this one merged directly.  I'll leave a small time window here in case someone thinks otherwise.

matthew-larson:
Thanks Edwin.  I verified the fix does provide a check for both the inlet node and the exhaust node that they do in fact match with the zone node names and provides a warning that references these nodes if not.

Myoldmopar:
This works for me, and creating the small worker functions makes it easy to read.  Going in.  Thanks @matthew-larson.  

@mjwitte  or @rraustad if I missed something on here, feel free to pop open a follow-up issue later.

PR #8322, opened 2020-09-29
Fix Unitary System with Heating Coil Temperature Setpoint Node Name Field Non-Blank

### LINKED ISSUE #8299: Zero heating rate in DOAS Unitary System with Temperature Setpoint Node Name
Issue overview
--------------
I have a model that has a Unitary System set up as a DOAS with fuel heating, DX cooling, and cycling fan operation. The fuel heating coil has it's setpoint node name specified as the heating coil's outlet node (see below).

       Coil:Heating:Fuel,
         DOAS Unitary Heating Coil, !- Name
         ...
         DOAS Unitary Heating Coil Outlet Node, !- Air Outlet Node Name
         DOAS Unitary Heating Coil Outlet Node; !- Temperature Setpoint Node Name

According to IO Ref for this input field:
Temperature Setpoint Node Name - If the coil is used in the air loop simulation and is temperature controlled using a Set Point Manager, then the node that is the control node needs to be specified here. If the coil is used in an air terminal unit, the coil is load controlled and a control node set point is not required.

The DOAS unitary system is setpoint controlled, and the fuel heating coils' outlet node is part of a node list used by the mixed air setpoint manager of the DOAS unitary system, so I would assume that the highlighted IO Ref sentence above applies. However, in simulating just a heating and cooling design day, the DOAS Unitary system's fuel heating coil doesn't turn on. See plot from Results Viewer below.
image

If the fuel heating coil's Temperature Setpoint Node Name input field is instead left blank, then the fuel heating coil works as intended. See plot from Results Viewer below.
image

This bug case is contained in the attached "setpoint-node-specified.idf". The IDF uses Chicago weather.

[setpoint-node-specified.idf.txt](https://github.com/NREL/EnergyPlus/files/5483499/setpoint-node-specified.idf.txt)

### Details
 - EnergyPlus v9.3

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #8299 

This pull requests addresses the situation where a heating coil object with a Temperature Setpoint Node Name field is referenced by a unitary system object with the Control Type set to Setpoint.  When the Temperature Setpoint Node Name field is not blank, the heating coil outlet node temperature is set to the heating coil outlet setpoint, which when compared to the unitary system outlet temperature setpoint, results in no heating rate being required.  The pull request creates a flag that is set to true when the coil is load controlled and the temperature setpoint node is not required and not used.  If this flag is set to true, the same logic is applied as if there was no temperature setpoint node name specified like it is done [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/HeatingCoils.cc#L1301) when the Temperature Setpoint Node Name is left blank, as recommended.

I also modified the documentation for the Temperature Setpoint Node Name field for further clarification when this field is required or can be left blank. 

### Pull Request Author

### Reviewer


src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12851,6 +12851,11 @@ namespace UnitarySystems {
                     }
                 }
 
+                // If the coil is temperature controlled, reset the desired outlet temperature so the temperature setpoint at the oulet node won't be used
+                if (DataLoopNode::SensedLoadFlagValue < 0) {
+                    DataLoopNode::Node(OutletNode).Temp = 0;
+                }
```
rraustad: I don't understand this logic. DataLoopNode::SensedLoadFlagValue = -999.0 so it is always < 0. Also, the component models should be responsible for setting a node condition not the parent so this code should be removed (unless there is an extremely good reason to do it). Can I assume that with this code the unit test you wrote still passes now?

matthew-larson: I originally was doing something similar in HeatingCoils.cc instead but ran into some issues with other unit tests so I switched to putting this in the UnitarySystem.cc since it occurred for unitary systems.  Based on your comment, I'm thinking now I should put what I had back in HeatingCoils.cc and try to address the other unit test issues.  This seems out of place here now.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15406,3 +15406,202 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTes
...
+                      ZoneEquipment,
+                      sensOut,
+                      latOut);
+
+    // check that heating coil air outlet node is at set point
+    EXPECT_NEAR(DataLoopNode::Node(2).Temp, DataLoopNode::Node(2).TempSetPoint, 0.001);
+    // heating coil air inlet node temp is less than heating coil air outlet node temp
+    EXPECT_GT(DataLoopNode::Node(2).Temp, DataLoopNode::Node(3).Temp);
+    // #6282 idle air flow rate for electric heating coils should equal 0
+    EXPECT_EQ(0.0, thisSys->m_IdleMassFlowRate);
```
rraustad: If you add another test here with the outlet node set point = 23.0 then the outlet temp should be 24 C + fan heat (heat coil is off so outlet T = inlet T + fan heat). It might be = 0 with current code. Or maybe not since calcUnitaryHeatingSystem is called again after your code change. Still, setting outlet node T does not look like a proper fix.

src/EnergyPlus/UnitarySystem.cc
rraustad: Certainly don't do this:

  if (DataLoopNode::SensedLoadFlagValue < 0)

rraustad: If you can briefly describe the issue you see in code it might help in the review process.

matthew-larson:
I moved the fix and unit test over to the HeatingCoils.cc module and updated the description to hopefully explain the fix.

src/EnergyPlus/HeatingCoils.cc
```diff
@@ -1360,10 +1361,16 @@ namespace HeatingCoils {
...
+                StptNodeErrorFlag = true;
             }
             MySPTestFlag(CoilNum) = false;
         }
 
+        // If the coil is temperature controlled, reset the desired outlet temperature so the temperature setpoint at the oulet node won't be used
+        if (StptNodeErrorFlag) {
+            HeatingCoil(CoilNum).DesiredOutletTemp = 0.0;
+        }
+
```
rraustad: I still don't understand why this is necessary. If the outlet node set point = -999 then the heating coil won't turn on, right? And if outlet node set point = 0 then it also won't turn on, except in winter if OA frac is large. So what's the point?

matthew-larson: Neal and I dug through this more and the problem as far as we can tell is that the process is running the SimulateHeatingCoilComponents function twice, once within the controlHeatingSystemToSP function and again in the calcUnitaryHeatingSystem function within the same iteration.  After the first call in the controlHeatingSystemToSP function, the PartLoadRatio gets set to 0 in [line 12855](https://github.com/NREL/EnergyPlus/blob/75fcd72b36440e1aaddf98e4b2fcd8341565340a/src/EnergyPlus/UnitarySystem.cc#L12855) since the outlet node matches the desired outlet node.  Then right before the second call for SimulateHeatingCoilComponents, QCoilReq, or HeatCoilLoad being passed in, is modified from -999 to 0 due to the PartLoadRatio being set to 0.  This is what is resulting in the coil not running in the CalcFuelHeatingCoil function due to not going into the `else if` statement on line [line 2178](https://github.com/NREL/EnergyPlus/blob/75fcd72b36440e1aaddf98e4b2fcd8341565340a/src/EnergyPlus/HeatingCoils.cc#L2178) to calculate a new TempAirOut and HeatingCoilLoad.  Our fix is to check if the `HeatCoilLoad` equals -999 or if the PartLoadRatio is 0, don't override the HeatCoilLoad to 0 and use -999 or the actual calculated PartLoadRatio.

I just pushed up the latest changes, all unit tests appear to pass.  Thank you for your feedback so far, it has helped weed through everything to figure out what is exactly happening.  Let me know your thoughts on the latest fix.

matthew-larson:
Neal and I dug through this more and the problem as far as we can tell is that the process is running the `SimulateHeatingCoilComponents` function twice, once within the `controlHeatingSystemToSP` function and again in the `calcUnitaryHeatingSystem` function within the same iteration.  After the first call in the `controlHeatingSystemToSP` function, the `PartLoadRatio` gets set to 0 in [line 12855](https://github.com/NREL/EnergyPlus/blob/75fcd72b36440e1aaddf98e4b2fcd8341565340a/src/EnergyPlus/UnitarySystem.cc#L12855) since the outlet node matches the desired outlet node.  Then right before the second call for `SimulateHeatingCoilComponents`, `QCoilReq`, or `HeatCoilLoad` being passed in, is modified from -999 to 0 due to the `PartLoadRatio` being set to 0.  This is what is resulting in the coil not running in the `CalcFuelHeatingCoil` function due to not going into the `else if` statement on line [line 2178](https://github.com/NREL/EnergyPlus/blob/75fcd72b36440e1aaddf98e4b2fcd8341565340a/src/EnergyPlus/HeatingCoils.cc#L2178) to calculate a new `TempAirOut` and `HeatingCoilLoad`.  Our fix is to check if the `HeatCoilLoad` equals -999 or if the `PartLoadRatio` is 0, don't override the `HeatCoilLoad` to 0 and use -999 or the actual calculated `PartLoadRatio`.

I just pushed up the latest changes, all unit tests appear to pass.  Thank you for your feedback so far, it has helped weed through everything to figure out what is exactly happening.  Let me know your thoughts on the latest fix.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -11066,7 +11066,9 @@ namespace UnitarySystems {
 
             } else if ((SELECT_CASE_var == DataHVACGlobals::Coil_HeatingGasOrOtherFuel) ||
                        (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingElectric)) {
-                HeatCoilLoad = PartLoadRatio * m_DesignHeatingCapacity;
+                if ((HeatCoilLoad != DataLoopNode::SensedLoadFlagValue) || (PartLoadRatio > 0)) {
+                        HeatCoilLoad = PartLoadRatio * m_DesignHeatingCapacity;
+                }
```
rraustad: OK, now I understand what is happening, the question is whether this is the correct fix or not.

The heating coil is controlled 2 ways, either by load or by set point.

        // Control output to meet load QCoilReq (QCoilReq is passed in if load controlled, otherwise QCoilReq=-999)
        if ((AirMassFlow > 0.0 && HeatingCoil(CoilNum).NominalCapacity > 0.0) && (GetCurrentScheduleValue(HeatingCoil(CoilNum).SchedPtr) > 0.0) &&
            (QCoilReq > 0.0)) {

or

            // Control coil output to meet a setpoint temperature.
        } else if ((AirMassFlow > 0.0 && HeatingCoil(CoilNum).NominalCapacity > 0.0) &&
                   (GetCurrentScheduleValue(HeatingCoil(CoilNum).SchedPtr) > 0.0) && (QCoilReq == SensedLoadFlagValue) &&
                   (std::abs(TempSetPoint - TempAirIn) > TempControlTol)) {

The code here in UnitarySystem is passing a load to the heating coil based on PLR * DesignHeatingCapacity. This seems correct to me.

I would question why the PLR is being set to 0 around UnitarySystem line 12850 in the else if since the previous call to the heating coil should have return the results with the heating coil off.

                //     IF outlet temp at no load is within ACC of set point, do not run the coil
                if (std::abs(DataLoopNode::Node(OutletNode).Temp - DesOutTemp) < Acc ||
                    this->m_HeatingCoilType_Num == DataHVACGlobals::Coil_UserDefined) {
                    // do nothing, coil is at the set point.
                } else if ((DataLoopNode::Node(OutletNode).Temp - DesOutTemp) > Acc) { // IF outlet temp is above set point turn off coil
                    PartLoadFrac = 0.0;
                } else { // ELSE get full load result
 
Up at line 12783 (ish) is this:

                    } else if ((SELECT_CASE_var == DataHVACGlobals::Coil_HeatingGasOrOtherFuel) ||
                               (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingElectric) ||
                               (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingDesuperheater)) {

                        HeatingCoils::SimulateHeatingCoilComponents(
                            state, CompName, FirstHVACIteration, DataLoopNode::SensedLoadFlagValue, CompIndex, _, _, FanOpMode);

It seems to me that this call is wrong and should be:

                        HeatingCoils::SimulateHeatingCoilComponents(
                            state, CompName, FirstHVACIteration, PartLoadFrac, CompIndex, _, _, FanOpMode);

or whatever it needs to be such that the coil is off. Also there are 3 coil types here, so be careful since one coil type may require a different form of the call. If that's the case then this else if should be split into 2. I would test all 3 coil types with whatever change is made here. I assume once that is done then this change (line 11069 above) can be reverted.

matthew-larson:
@rraustad I tested this latest heating coil fix with the 3 heating coil types (fuel, electric, and desuperheater) and it should be ready for another review.

rraustad:
@mjwitte there are no diff's which means no example file with this configuration. I am also guilty of not including example files for specific configurations. What should we do regarding example files? Modify one? Add one? Other?

mjwitte:
Either modify (preferred) or add one. And in this case, it would be great to duplicate the unit test for all 3 coil types.

matthew-larson:
I could modify an existing example file by just specifying a temperature setpoint node name for the outlet of the heating coil.  I'll go ahead and create duplicate unit tests as well for the other 2 coil types.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15406,3 +15406,200 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTes
      EXPECT_NEAR(thisSys->m_SpeedRatio, 0.228062, 0.02);
      EXPECT_NEAR(sensOut, -11998.0, 210.0);
 }
+
+TEST_F(ZoneUnitarySysTest, UnitarySystemModel_FuelHeatCoilStptNodeTest)
```
mjwitte, line 15568: The new unit test doesn't build. It needs to be brought up to speed with all the latest global data changes.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15406,3 +15406,200 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTes
...
+                      OAUCoilOutTemp,
+                      ZoneEquipment,
+                      sensOut,
+                      latOut);
+
+    // check that heating coil air outlet node is at set point
+    EXPECT_NEAR(DataLoopNode::Node(2).Temp, DataLoopNode::Node(2).TempSetPoint, 0.001);
+    // heating coil air inlet node temp is less than heating coil air outlet node temp
+    EXPECT_GT(DataLoopNode::Node(2).Temp, DataLoopNode::Node(3).Temp);
+}
```
mjwitte: Please add a return (end-line) at the end of the file.

mjwitte:
@matthew-larson Thanks for the defect file, but it fails early with this branch, so I can't verify the fix.
```
   **  Fatal  ** CheckForRunawayPlantTemps: Simulation terminated because of run away plant temperatures, too hot
```

matthew-larson:
@mjwitte That's interesting, not getting that on my end, I can investigate.  Attached is the EnergyPlus example file I modified for this pull request which should work then.
[5Zone_Unitary_VSDesuperheater.idf.txt](https://github.com/NREL/EnergyPlus/files/5484685/5Zone_Unitary_VSDesuperheater.idf.txt)

rraustad:
Both develop and this branch report this warning:

    ** Warning ** Coil:Heating:Fuel "DOAS UNITARY HEATING COIL"
    **   ~~~   **  The Temperature Setpoint Node Name input is not required for this heating coil because
    **   ~~~   **  this heating coil is controlled based on the load calculated by the thermostat.
    **   ~~~   ** ... this heating coil is not controlled by using a temperature setpoint manager.
    **   ~~~   ** ... if a temperature setpoint is placed at the outlet node of this heating coil, that temperature setpoint will not be used.
    **   ~~~   ** ... leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

I wonder if we should try to eliminate this warning by UnitarySystem setting the control node to 0 or try to modify the warning when UnitarySystem is using set point control since much of the warning text is inaccurate. This warning was developed before UnitarySystem so the warning message refers to load control. 

Another possibility would be going back to using SensedNodeFlag as the load variable and allowing the heating coil to control the coil PLR. This method would eliminate the need for iterating within UnitarySystem but require a re-write of the logic in controlHeatingSystemToSP and controlSuppHeatSystemToSP. Not sure if there would be any snags with this alternate method.

matthew-larson:
I'm hesitant to re-write the logic to eliminate the need for iterating in UnitarySystem as it goes beyond the scope of what I intended for this issue.  I can see how the warning is outdated however.  My thought would be to modify the warning for this pull request and look at opening up a new issue for re-writing the logic to eliminate the need for iterating within UnitarySystem.  I can also look at setting the control node to 0 as you mentioned, and/or the ability to expand the if statement with the warning as well to eliminate it when the heating coil is contained in the unitary system.

matthew-larson:
I can't see an easy way to change the logic to eliminate this warning when using the unitary system so we could generalize the warning and just remove the line about load-based control since it isn't applicable to every situation this warning appears, but the rest of the warning seems to still apply.  Let me know your thoughts, I'm happy to do whatever you guys think is best at this point.

```
** Warning ** Coil:Heating:Fuel "DOAS UNITARY HEATING COIL"
**   ~~~   **  The Temperature Setpoint Node Name input is not required for this heating coil because
**   ~~~   **  this heating coil is not controlled by using a temperature setpoint manager.
**   ~~~   ** ... if a temperature setpoint is placed at the outlet node of this heating coil, that temperature setpoint will not be used.
**   ~~~   ** ... leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.
```

rraustad:
Yeah, I think revising the warning to be more generic is the way to go.

rraustad:
The 2nd and 3rd line don't seem to apply to this configuration.

    ** Warning ** Coil:Heating:Fuel "DOAS UNITARY HEATING COIL"
    **   ~~~   **  The Temperature Setpoint Node Name input is not required for this heating coil because
    **   ~~~   **  this heating coil is not controlled by using a temperature setpoint manager.
    **   ~~~   ** ... if a temperature setpoint is placed at the outlet node of this heating coil, that temperature setpoint will not be used.
    **   ~~~   ** ... leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

How about simplifying it further as:

    ** Warning ** Coil:Heating:Fuel "DOAS UNITARY HEATING COIL"
    **   ~~~   **  The Temperature Setpoint Node Name input is not required for this heating coil.
    **   ~~~   **  Leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

If a load is passed in to SimulateHeatingCoilComponents, in this case based on set point manager control, then this warning is just short of unnecessary. This warning only tells the user that this specific input field isn't used. If a load is not passed in then a different warning is reported. 

@mjwitte can you think of anything else to say for this warning? 

I'm tempted to add to the test so this warning won't print when the heating coil's set point node name does point to the control node. But what if UnitarySystem does use load control and the user also uses a set point manager to control the heating coil? I think we would want this message in that case (which would not print unless the user made a mistake). 

            if (ControlNode > 0 && Node(ControlNode).TempSetPoint == SensedLoadFlagValue) {

mjwitte:
@rraustad I'm not sure I follow everything completely. It seems we could do away with this warning altogether.
When unitary system calls the heating coil, does it always set `QCoilReq` ? For both load-based and setpoint control?
If so, then the coil is going to ignore the setpoint node name, correct? And any setpoint that may have been placed there needlessly?

rraustad:
This branch changed 3 calls to SimulateHeatingCoilComponents. If I search that function name in develop I find more that still use SensedLoadFlagValue. So without more review I can't say why those calls still use set point control. When UnitarySystem does call the heating coil with a load value then the heating coil SP node name input will be ignored.

rraustad:
OA coils do not pass a load:

     //     stand-alone coils are temperature controlled (do not pass QCoilReq in argument list, QCoilReq overrides temp SP)
    SimulateHeatingCoilComponents(state, CompName, FirstHVACIteration, _, CompIndex);

Air loop coils also do not pass a load but request what load has been met:

    SimulateHeatingCoilComponents(state, CompName, FirstHVACIteration, _, CompIndex, QActual);

Most other components pass in a load (PTAC, PTHP, FanCoil, etc.). SingleDuct uses:

    SimulateHeatingCoilComponents(state, this->ReheatName, FirstHVACIteration, QZnReq, this->ReheatComp_Index);

Furnaces pass in a load:

    SimulateHeatingCoilComponents(state, BlankString, FirstHVACIteration, HeatCoilLoad, Furnace(FurnaceNum).SuppHeatCoilIndex, _, true, FanOpMode);

Multispeed heat pump uses a variation. Maybe this is where some of the set point control came to exist in heating coils.

    SimulateHeatingCoilComponents(state, MSHeatPump(MSHeatPumpNum).HeatCoilName, FirstHVACIteration, _,
                                                  0, _, _, MSHeatPump(MSHeatPumpNum).OpMode, PartLoadFrac, SpeedNum, SpeedRatio);

And then UnitarySystem combined all these control methods. And also allowed this parent to be controlled via a node set point.

So I think this warning does have it's place, otherwise a user might ask "Why isn't this coil controlling to my set point"?

rraustad:
This fix does correct the problem.

image

matthew-larson:
I think I follow you. I implemented the new if statement and the warning does still appear for the fuel heating coil and electric heating coil, but it does still fix the problem.  Were you thinking of using this new if statement and simplified warning?  Only the case with the Desuperheater heating coil does the set point node name point to the control node, the other 2 cases it is -999.  So maybe something does have to change elsewhere.

rraustad:
The only point here is to inform the user that the set point node name in the heating coil is not really what controls coil operation. In the UnitarySystem, if the control node is not at the heating coil outlet, the system will still control based on the control node set point temperature. There are many ways to set up the UnitarySystem, if the heating coil follows that set up, and the UnitarySystem uses set point control, then everything should be OK and the warning is not needed. If the UnitarySystem uses load control, and a user specifies a set point control node name, and places a set point at that node, then this warning does have meaning. In the desuperheater test file, which node controls the UnitarySystem and which node is specified as the set point node name in the heating coil object?

I think I would modify the warning to simplify it and leave it at that. That should be good enough for this issue. We can think more about what else needs to happen in the future. I'm still not sure about the change to the IF statement, I think the heating coil needs more information (e.g., is UnitarySystem load controlled or set point controlled) before reporting that warning. If UnitarySystem is load controlled then the original message was fine. If the UnitarySystem is set point controlled, and the user enters a set point control node name in the heating coil, then the user simply needs to know that the UnitarySystem controls heating coil operation, not the heating coil. The user could put a SPM at the UnitarySystem outlet node and then put another SPM at the heating coil outlet node, in this use case the UnitarySystem set point manager would control the heating coil.

rraustad:
Looking again at heating coil control, I notice my result was different from the issue description. there appeared to be heating coil operating in summer in the issue description figure. In this branch there is none. Then I looked at the heating coil outlet node set point and temperature. It looks like the heating coil should be off since it's within 0.001 C of the set point, so this does look correct and maybe investigating why it was operating before is not necessary.

image

rraustad:
@matthew-larson leave things as you now have them. I will probably merge this as is.

rraustad:
Latest warning text is less confusing:

    ** Warning ** Coil:Heating:Fuel "DOAS UNITARY HEATING COIL"
    **   ~~~   **  The "Temperature Setpoint Node Name" input is not required for this heating coil.
    **   ~~~   **  Leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

rraustad:
Defect file DOAS unit coils operate as expected:

image

rraustad:
The 2 defect files have changed text in the err file. These are both expected and explainable.

5Zone_Unitary_VSDesuperheater - new warning because the code was changed to send a load to the heating coil.

    +   ** Warning ** Coil:Heating:Fuel "MAIN HEATING COIL 1"
    +   **   ~~~   **  The "Temperature Setpoint Node Name" input is not required for this heating coil.
    +   **   ~~~   **  Leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

ChangeoverBypassVAV_AirToAir - a change in warning text in existing warning.

        ** Warning ** Coil:Heating:Fuel "FURNACE HEATING COIL 1"
    -   **   ~~~   **  The Temperature Setpoint Node Name input is not required for this heating coil because
    -   **   ~~~   **  this heating coil is controlled based on the load calculated by the thermostat.
    -   **   ~~~   ** ... this heating coil is not controlled by using a temperature setpoint manager.
    -   **   ~~~   ** ... if a temperature setpoint is placed at the outlet node of this heating coil, that temperature setpoint will not be used.
    -   **   ~~~   ** ... leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.
    +   **   ~~~   **  The "Temperature Setpoint Node Name" input is not required for this heating coil.
    +   **   ~~~   **  Leaving the input field "Temperature Setpoint Node Name" blank will eliminate this warning.

rraustad:
@matthew-larson there is only 1 remaining item, The new warning in 5Zone_Unitary_VSDesuperheater is because the temperature set point control node name was added to the heating coil. I believe you did that as a test, correct? And that change will be reverted?

matthew-larson:
@rraustad I did that originally as a test and then @mjwitte recommended modifying an example file to show this configuration so I kept it in.  Based on the CI results it looks like ChangeoverBypassVAV_AirToAir also demonstrates this situation.  If you don't think the 5Zone_Unitary_VSDesuperheater modification is necessary, I can revert it back.

rraustad:
@mjwitte @Myoldmopar  I thought we were moving towards a clean set of example files. Are you OK with the change to 5Zone_Unitary_VSDesuperheater where the control node name was added to the heating coil and now this warning is reported?

mjwitte:
@rraustad If I remember correctly, the reason for suggesting the change to the test file was because we didn't have any files with this configuration, so the problem was flying under the radar. The solution here has morphed a few times, so as long as we have a thorough unit test to catch any future problems, I'm ok with reverting the test file.

rraustad:
Well, that test file is less than clean even with this new warning. I'm Ok with it as is.

rraustad (approved):
Docs look good, unit tests run locally, revised warning gets right to the point and defect file runs correctly now.

rraustad:
So thinking about what @mjwitte just said about no test files with this configuration, I wonder why 5Zone_Unitary_VSDesuperheater didn't show diff's. That example file does have this configuration, that's why the new warning showed up. Things appear to work correctly, but why was this heating coil operating and the defect heating coil was not?

    AirLoopHVAC:UnitarySystem,
      GasHeat DXAC Furnace 1,  !- Name


image

matthew-larson:
The first time it calls `SimulateHeatingCoilComponents`, the desuperheater uses `ReqOutput` instead of `this->m_DesignHeatingCapacity`, so I'm not sure it was actually an issue for this coil type.  I can do some testing on this.

```
                        } else if ((SELECT_CASE_var == DataHVACGlobals::Coil_HeatingGasOrOtherFuel) ||
                                   (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingElectric)) {

                            HeatingCoils::SimulateHeatingCoilComponents(
                                state, CompName, FirstHVACIteration, this->m_DesignHeatingCapacity, CompIndex, _, _, FanOpMode);

                        } else if (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingDesuperheater) {

                            HeatingCoils::SimulateHeatingCoilComponents(state, CompName, FirstHVACIteration, ReqOutput, CompIndex, _, _, FanOpMode);

                        } else if (SELECT_CASE_var == DataHVACGlobals::Coil_HeatingWater) {
```

rraustad:
I ran this file in develop and the heating coil is off. Where the heck are the CI diff's? For this run, the humidistat heating coil is picking up the reheat load, which should have happened anyway, that is if it's reheat and not heating. So this branch does look correct. I think I'll merge this and move on.

image

rraustad:
Well, this example file is just all kinds of messed up. The control is CoolReheat, using set point control. This doesn't work very well since there is no humidistat in this file. No warnings of this in the err file. So the "humidistat" heating coil operation is due to a heating load. But in summer? This is ripe for a follow up issue.

    AirLoopHVAC:UnitarySystem,
      GasHeat DXAC Furnace 1,  !- Name
      ,                        !- Controlling Zone or Thermostat Location
      CoolReheat,              !- Dehumidification Control Type

image

rraustad:
So the 5Zone_Unitary_VSDesuperheater example file uses a scheduled humidity ratio manager. In this case it would not need a humidistat.

    SetpointManager:Scheduled,
      Supply Air Hum Rate Manager 1,  !- Name
      MaximumHumidityRatio,    !- Control Variable
      Supply Air Hum Rate Schedule,  !- Schedule Name
      VAV Sys 1 Outlet Node;   !- Setpoint Node or NodeList Name

rraustad:
A few more plots of this example file's HVAC operation. First in the develop branch.

    Schedule:Constant,Supply Air Hum Rate Schedule,,0.006;

image

image

rraustad:
And now with this branch.

image

image

PR #8323, opened 2020-09-29
Unitary system now reports warning when cycling supply fan schedule is used with setpoint control type

### LINKED ISSUE #8285: Near zero single speed cooling coil electric power for unitary system with cycling supply fan
Issue overview
--------------
I have a model that has a Unitary System set up as a DOAS with gas heating, DX cooling, and cycling fan operation. The cycling fan operation is set by the Supply Air Fan Operating Mode Schedule Name input field of the unitary system object with the schedule set to always zero. 

image
image

According to IO Ref for this input field:
Supply Air Fan Operating Mode Schedule Name - This alpha field specifies the name of the supply air fan operating mode schedule. The supply air fan operating mode may vary during the simulation based on time-of-day or with a change of season. Schedule values of 0 denote that the unitary system supply air fan and the heating or cooling coil cycle on and off together to meet the heating or cooling load (a.k.a. AUTO fan). Schedule values other than 0 denote that the supply fan runs continuously while the heating or cooling coil cycles to meet the load. The SingleZoneVAV control type is only active when the supply air fan runs continuously (i.e., during cycling fan operation the Control Type = Load model is used).

Simulating just a heating and cooling design day, the DOAS Unitary system works correctly on both days. However, in the cooling design day the Cooling Coil Electricity Rate is near zero. See table below.

image

If the fan operation schedule changes to always 1, indicating a constant flow supply fan per IO Ref, then the Cooling Coil Electricity Rate is much more realistic.

[CyclingFan_SetpointControl.zip](https://github.com/NREL/EnergyPlus/files/5453260/CyclingFan_SetpointControl.zip)

### Details
 - EnergyPlus v9.3

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #8285 
 - This pull request corrects a check for the unitary system based on the Supply Air Fan Operating Mode Schedule when the Control Type is set to Setpoint.  Previously, the fan operation mode schedule was not being checked, which was keeping the cooling coil from being enabled.

### Pull Request Author

### Reviewer


src/EnergyPlus/UnitarySystem.cc
```diff
@@ -7996,6 +7996,15 @@ namespace UnitarySystems {
...
 
+                if (this->m_FanOpModeSchedPtr > 0) {
+                    if (ScheduleManager::GetCurrentScheduleValue(this->m_FanOpModeSchedPtr) == 0.0) {
+                        this->m_FanOpMode = DataHVACGlobals::CycFanCycCoil;
+                    } else {
+                        this->m_FanOpMode = DataHVACGlobals::ContFanCycCoil;
+                        DataHVACGlobals::OnOffFanPartLoadFraction = 1.0;
+                    }
+                }
+
```
rraustad: Well, this works but now this same code is in 2 places, here and in initLoadBasedControl. Shouldn't these get moved to initUnitarySystems inside the FirstHVACIteration test (line 1084) so that both control types execute this, it's only done once per time step and there is only 1 copy of these lines of code.

matthew-larson: That makes sense, I'll make this adjustment once the CI finishes.  Thanks

matthew-larson:
@rraustad I moved the fix to initUnitarySystems as recommended to eliminate the duplicated check.  If you could make sure I got it in the right spot, it's ready for another review.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15406,3 +15406,234 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTes
...
+                      ZoneEquipment,
+                      sensOut,
+                      latOut);
+
+    EXPECT_NEAR(thisSys->m_CompPartLoadRatio, 0.35, 0.001);
+    // check that cooling coil air outlet node is at set point
+    EXPECT_NEAR(DataLoopNode::Node(2).Temp, DataLoopNode::Node(2).TempSetPoint, 0.001);
+    // cooling coil air inlet node temp is greater than cooling coil air outlet node temp
+    EXPECT_GT(DataLoopNode::Node(3).Temp, DataLoopNode::Node(2).Temp);
+}
```
rraustad: Shoot, this was ready to merge. Line feed needed at end of unit test.

rraustad (changes requested):
Revision to code location so load based and set point based control use same logic. This looks good. Unit test still needs mod to last line.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -991,6 +991,15 @@ namespace UnitarySystems {
             DataLoopNode::Node(this->m_HeatRecoveryInletNodeNum).MassFlowRate = mdotHR;
         }
 
+        if (this->m_FanOpModeSchedPtr > 0) {
```
rraustad: Well, I actually suggested this be moved to line 1006, or better yet 1010, below since it needs only be done once per time step. Might as well make that change while you correct the unit test.

rraustad:
Perfect, I think I'll run the defect file before merging this. Also, check off the pull request author items in the pull request description. Think about the PR title from the users perspective.

rraustad:
Thanks @matthew-larson. You are done with this one unless something else unexpectedly pops up. I'll get to testing the defect file when I can.

matthew-larson:
Sounds good, thanks for your help and walking me through the code, it was very valuable.  I'll keep tabs on this to make sure everything comes back clean.

matthew-larson:
@rraustad I updated the pull request for the new fix of making sure the correct error is caught when a cycling fan is used with Setpoint control.  I also updated the unit test to check for the error.  Let me know if you have any additional comments on this.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15406,3 +15406,159 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_MultiSpeedDXCoilsDirectSolutionTes
...
+    bool zoneEquipment = true;
+    UnitarySystems::UnitarySys::factory(state, DataHVACGlobals::UnitarySys_AnyCoilType, compName, zoneEquipment, 0);
+    UnitarySystems::UnitarySys *thisSys = &state.dataUnitarySystems->unitarySys[0];
+
+    DataGlobals::NumOfTimeStepInHour = 1;
+    DataGlobals::MinutesPerTimeStep = 60;
+    DataZoneEquipment::ZoneEquipInputsFilled = true;                                    // indicate zone data is available
+    thisSys->getUnitarySystemInputData(state, compName, zoneEquipment, 0, ErrorsFound); // get UnitarySystem input from object above
+    std::string const error_string = delimited_string({
+        "   ** Severe  ** AirLoopHVAC:UnitarySystem = UNITARY SYSTEM MODEL\n   **   ~~~   ** For FAN:ONOFF = SUPPLY FAN 1\n   **   ~~~   ** Fan operating mode must be continuous (fan operating mode schedule values > 0).\n   **   ~~~   ** Error found in Supply Air Fan Operating Mode Schedule Name ALWAYSZERO\n   **   ~~~   ** ...schedule values must be (>0., <=1.)"
```
rraustad, line 15561: @matthew-larson are you clang formatting as you go? It seems this line would have wrapped. Not sure if it's worth another push if everything else is OK.

matthew-larson: I have clang formatting ON in CLion.  I can double check that it's working right.

rraustad:
This seems correct now.

The DOAS unit uses setpoint control and FanOpMode schedule value = 0

    ** Severe  ** AirLoopHVAC:UnitarySystem = DOAS UNITARY
    **   ~~~   ** For FAN:ONOFF = DOAS UNITARY SUPPLY FAN
    **   ~~~   ** Fan operating mode must be continuous (fan operating mode schedule values > 0).
    **   ~~~   ** Error found in Supply Air Fan Operating Mode Schedule Name DOAS UNITARY FAN MODE SCHEDULE
    **   ~~~   ** ...schedule values must be (>0., <=1.)
    **  Fatal  ** getUnitarySystemInputData: previous errors cause termination. Check inputs

If value for FanOpMode schedule is changed to 1 or that input is left blank then there are no fan op mode schedule warnings and the coil turns on:

image

The check of the FanOpMode schedule for both set point control and load or SZVAV control is not necessary so I'm glad that code was not ultimately moved as discussed. For set point control, we know the fan op mode must be constant so only checking once that the schedule value is > 0 is the only thing needed.

rraustad:
Running unit tests locally multiple times to get a clean pass. Keeps stalling at LoggingSubStep4stepsPerHour. That unit test should not be related to this change.

mitchute:
> Keeps stalling at LoggingSubStep4stepsPerHour

@rraustad, FWIW, that test usually takes a few minutes to complete on my machine.

rraustad:
Yeah, that unit test finally finished after 15 minutes, using a release build.

image

rraustad (approved):
Simple changes made to catch issues with fan operating mode schedule when set point control is used.

mitchute:
15 minutes... 😮

PR #8347, opened 2020-10-19
Heat balance surface data structure refactoring

xuanluo113:
Pull request overview
---------------------
This pull request includes the following pieces
- ENUM surface class
- Reduce the scope of function level local variables in HeatBalanceSurface & SolarShading in a few functions including `ComputeIntThermalAbsorpFactors`, `ComputeIntSWAbsorpFactors`.
- Add the shading surface index list and ext solar surface list and (surface listed to be sorted in future PRs)
- Continue converting the variable names in solar shading functions according to the naming convention
- Remove unused solar heat balance variables and related unused calculations including
```
Array1D<Real64> DSZone; // Factor for sky diffuse solar radiation into a zone
Array1D<Real64> DGZone; // Factor for ground diffuse solar radiation into a zone
Array1D<Real64> CBZone; // Factor for beam solar absorbed by interior shades
```
- Refactoring of CalcInteriorSolarDistribution
** reduce the scope of function level variable.
** element-wise initialization.
** separate the main loop into opaque surface loop and window+dome surface loop (to be continued when air boundary surface is sorted)
** reduce branch condition checking by reordering the branches, moving the glass layer loop to the bottom.

### Pull Request Author

### Reviewer


src/EnergyPlus/DataHeatBalance.hh
```diff
@@ -704,6 +704,7 @@ namespace DataHeatBalance {
         std::vector<int> ZoneIZSurfaceList;          // List of interzone surfaces in this zone
         std::vector<int> ZoneHTNonWindowSurfaceList; // List of non-window HT surfaces related to this zone (includes adjacent interzone surfaces)
         std::vector<int> ZoneHTWindowSurfaceList;    // List of window surfaces related to this zone (includes adjacent interzone surfaces)
+        std::vector<int> ZoneExtSolarSurfaceList;    // List of window surfaces related to this zone (includes adjacent interzone surfaces)
```
mjwitte: `(includes adjacent interzone surfaces)` does not apply here - if there are other changes, please fix here, otherwise, fix in the next window-struct branch.

src/EnergyPlus/DataSurfaces.cc
```diff
@@ -445,6 +421,7 @@ namespace DataSurfaces {
     std::vector<int> AllHTNonWindowSurfaceList; // List of all non-window heat transfer surfaces
     std::vector<int> AllHTWindowSurfaceList;    // List of all window surfaces
     std::vector<int> AllSurfaceListReportOrder; // List of all surfaces - output reporting order
+    std::vector<int> AllShadingSurfList;        // List of all shading surfaces
```
mjwitte: Because the shading surfaces are are sorted together at the beginning of the Surface data, it should be possible to replace this with `ShadingSurfaceFirst` and `ShadingSurfaceLast`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -1262,42 +1261,44 @@ namespace HeatBalanceSurfaceManager {
                     computedNetArea(Surface(iSurf).BaseSurf) -= Surface(iSurf).GrossArea * mult;
                 }
             }
-            if ((Surface(iSurf).Class <= 20) && (Surface(iSurf).Class >= 1)) {
-                ++numSurfaces(Surface(iSurf).Class);
+            int currSurfaceClass = int(Surface(iSurf).Class);
+            if ((currSurfaceClass <= 20) && (currSurfaceClass >= 1)) {
```
mjwitte: This is a bit fragile. This statement should never be false if I'm following correctly. I assume this is here to protect against and array bounds problem, but it seems that an assert would be better here. And wherever `numExtSurfaces` and others are dimensioned, it should use the max value of `SurfaceClass` (is there a method for that?) so it automatically adjusts if ever a new surface class is added.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3579,77 +3552,66 @@ namespace HeatBalanceSurfaceManager {
             int const solEnclosureNum = Zone(zoneNum).SolarEnclosureNum;
             for (int SurfNum = firstSurfOpague; SurfNum <= lastSurfOpague; ++SurfNum) {
```
mjwitte: Not from this PR,  `Opague` --> `Opaque`

src/EnergyPlus/SolarShading.cc
```diff
@@ -6311,80 +6043,100 @@ namespace SolarShading {
                 SurfWinBmSolTransThruIntWinRep(SurfNum) = 0.0;
                 SurfWinBmSolTransThruIntWinRepEnergy(SurfNum) = 0.0;
             }
+            int const firstSurfOpague = Zone(zoneNum).NonWindowSurfaceFirst;
```
mjwitte: Here also, `Opague` --> `Opaque`

src/EnergyPlus/SolarShading.cc
```diff
@@ -5964,288 +5967,24 @@ namespace SolarShading {
...
-        int ConstrNumSh;                                    // Shaded construction number
-        int ConstrNumBack;                                  // Construction number of back surface
-        int ConstrNumBackSh;                                // Shaded construction number of back surface
-        int ShadeFlag;                                      // Shading flag for a window
-        int ShadeFlagBack;                                  // Shading flag for a window that is a back surface
-        int Lay;                                            // Glass layer number
-        Real64 SwitchFac;                                   // Switching factor for a window
-        Real64 SwitchFacBack;                               // Switching factor for a window that is a back surface
-        Real64 TransBeamWin;                                // Beam solar transmittance of a window
-        Real64 TransBeamWinSh;                              // Beam solar transmittance of a shaded window
```
mjwitte, line 5982: Sooooo many scope reductions - nice!

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -1346,17 +1344,19 @@ namespace SurfaceGeometry {
 
         // Move all shading Surfaces to Front
 
+        state.dataSurfaceGeometry->SurfaceTmpClassInvalid.dimension(TotSurfaces, false);
+        state.dataSurfaceGeometry->SurfaceTmpClassMoved.dimension(TotSurfaces, false);
```
mjwitte: Are these used anywhere outside this function? Could these be local here?

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -1456,11 +1455,9 @@ namespace SurfaceGeometry {
             ShowSevereError(state, format("{}Reordered # of Surfaces ({}) not = Total # of Surfaces ({})", RoutineName, MovedSurfs, TotSurfaces));
             SurfError = true;
             for (int Loop = 1; Loop <= TotSurfaces; ++Loop) {
-                if (state.dataSurfaceGeometry->SurfaceTmp(Loop).Class != SurfaceClass_Moved) {
-                    if (state.dataSurfaceGeometry->SurfaceTmp(Loop).Class > 100) {
-                        ShowSevereError(state, RoutineName + "Error in Surface= \"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).Name + "\" Class=" +
-                                        cSurfaceClass(state.dataSurfaceGeometry->SurfaceTmp(Loop).Class - 100) + " indicated Zone=\"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).ZoneName + "\"");
-                    }
+                if (state.dataSurfaceGeometry->SurfaceTmpClassMoved(Loop) && state.dataSurfaceGeometry->SurfaceTmpClassInvalid(Loop)) {
```
mjwitte: This should be `!state.dataSurfaceGeometry->SurfaceTmpClassMoved(Loop) && ...`

mjwitte (commented):
@xuanluo113 One real comment below (a missing !) and some nit-picky ones that can be addressed here or in the next round. That's up to you. But the recent global data merge just introduced some conflicts.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: I agree it's just to make sure the dimension of `numExtSurfaces` and `numSurfaces` arrays are no more than the max value of `SurfaceClass`. And assertion would definitely be better to be used here. I'll address this in this branch.

src/EnergyPlus/SurfaceGeometry.cc
xuanluo113: They are used in `GetSurfaceData`, `GetIntMassSurfaceData`, `GetRectSurfaces`, `GetHTSurfaceData`. The last three are called by `GetSurfaceData`. I don't know what is better, using them as state variables, or pass them as function parameters?

mjwitte: I think Moved is only used in GetSurfaceData, so it can be local. And the other one is fine as a state variable.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: While you are at it, you should change 1 and 20 to whatever the enumerations are.

xuanluo113:
@mjwitte @Myoldmopar When I merge upstream develop, I got the compiling error as such,

`~/EnergyPlus/src/EnergyPlus/PluginManager.cc:135:38: error: unknown type name 'state'
    std::string pythonStringForUsage(state)
`
`~/EnergyPlus/src/EnergyPlus/PluginManager.cc:141:62: warning: unused parameter 'state' [-Wunused-parameter]
    void PluginManager::setupOutputVariables(EnergyPlusData &state)
`

I don't know if it's something wrong with my merge.

I'm using a mac with Release build, and the Python related cmake settings are (default configuration):
Screen Shot 2020-11-16 at 12 11 01 PM

Myoldmopar:
I think it may be fixed in a different branch, as we are now compiling with link_with_python OFF along with some other lesser-used compiler configurations.  But for now, just make it:

`std::string pythonStringForUsage(EnergyPlusData &state)`

src/EnergyPlus/SurfaceGeometry.cc
mjwitte: Or would it make sense to add a new surface class enum for `Invalid`?

mjwitte: That one could be last in the list and then you could use it in the range test for that assert `if (>None && <Invalid)`, maybe?

xuanluo113: I don't know. I was gonna explicitly add a `const int NO_SURFACE_CLASS = 15;`.
First `Invalid` is not literally a `SurfaceClass`. And no implicit way to prevent more classes to append after it. 

Here's [a post](https://stackoverflow.com/questions/2102582/how-can-i-count-the-items-in-an-enum) about using an extra (last) item of an ENUM class to represent its length, as the example below shows. `Invalid` may be used this way, but again it's not a 100% safe solution. 
`enum class Example { A, B, C, D, E, ExampleCount };`

amirroth: No, you shouldn't add a const NO_SURFACE_CLASS just for MOVED. MOVED is temporary. In fact, you can probably just make the MOVED array local to the function that uses it. 

As far as using enum class Example { A, B, C, D, Count }; and then using Example::Count as the length of the array ... that works if A is set to 0.

mjwitte: My suggestion was not to replace MOVED, but rather to replace INVALID. Storing INVALID briefly during initial setup is one issue. Counting the length of the enum to size the various reporting arrays is another.

amirroth: Issue 1. INVALID. Is this a piece of information that needs to persist throughout the simulation or does it immediately trigger a fatal error and the simulation ends? If the former, I think you can have INVALID as part of the enum. Just set its value to -1 if you don't want to do any reporting on it. 

Issue 2. Enumeration counts. This is really only useful if the enumeration starts at 0 (or 1) and is dense.  [Ed: if an enumeration doesn't start at 0 (or 1) and/or is not dense, it probably shouldn't be an enumeration.] Using Count (I've usually called it Num, but whatever) is a good shorthand, and is automatic if you are starting the enumeration at 0. If you are starting at 1, you can always define Count to have the same value as the last element in your set, e.g., enum Example { A, B, C, D, Count = D };

mjwitte: INVALID is set during input processing for surfaces, and it eventually trips a fatal error. It's used in `GetSurfaceData` and a few functions called from there.

amirroth: As long as the simulation exits when there is an INVALID surface before any array that is indexed by enum SurfaceClass is read or written, it is okay to have INVALID as enum SurfaceClass -1.

xuanluo113: The [latest commit](https://github.com/NREL/EnergyPlus/pull/8347/commits/83d8eba37e731a54e1cb2728afd00b078f740552) added `SurfaceClassCount` (It's first item is None = 0). Looks like it's reasonable to add INVALID = -1 at the top.

src/EnergyPlus/DataSurfaces.cc
xuanluo113: Addressed.

src/EnergyPlus/SurfaceGeometry.cc
amirroth: You can probably just call it Count, because SurfaceClass is already the name of the enum.

src/EnergyPlus/SurfaceGeometry.cc
```diff
@@ -1455,11 +1456,9 @@ namespace SurfaceGeometry {
...
             SurfError = true;
             for (int Loop = 1; Loop <= TotSurfaces; ++Loop) {
-                if (state.dataSurfaceGeometry->SurfaceTmp(Loop).Class != SurfaceClass_Moved) {
-                    if (state.dataSurfaceGeometry->SurfaceTmp(Loop).Class > 100) {
-                        ShowSevereError(state, RoutineName + "Error in Surface= \"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).Name + "\" Class=" +
-                                        cSurfaceClass(state.dataSurfaceGeometry->SurfaceTmp(Loop).Class - 100) + " indicated Zone=\"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).ZoneName + "\"");
-                    }
+                if (!SurfaceTmpClassMoved(Loop) && state.dataSurfaceGeometry->SurfaceTmp(Loop).Class == SurfaceClass::INVALID) {
+                    ShowSevereError(state, RoutineName + "Error in Surface= \"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).Name + "\" Class=" +
+                                        cSurfaceClass(state.dataSurfaceGeometry->SurfaceTmp(Loop).Class) + " indicated Zone=\"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).ZoneName + "\"");
```
mjwitte: Hmmm, INVALID is getting used as an array index here in the error message. Sorry. I guess you should just go back to the bool array you had before for invalid.

amirroth: That's just a lookup to the string name of INVALID, which you can just hardcode as INVALID. That's not a good enough reason to go back to an arrays of bools.

xuanluo113: Uhhh.. The `SurfaceTmp(Loop).Class` it wants to print here was the original surface class. (SurfaceTmp(Loop).Class - 100)
```
if (state.dataSurfaceGeometry->SurfaceTmp(Loop).Class > 100) {
ShowSevereError(state, RoutineName + "Error in Surface= \"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).Name + "\" Class=" +
cSurfaceClass(state.dataSurfaceGeometry->SurfaceTmp(Loop).Class - 100) + " indicated Zone=\"" + state.dataSurfaceGeometry->SurfaceTmp(Loop).ZoneName + "\"");
 }
```
Can I just skip printing the Class in the error message? Surface Name and Zone Name should be enough to identify it?

amirroth: I should also say that adding 100 to SurfaceClass to indicate an error is exemplary programming. Exemplary of what not to do that is. If you want to retrieve the original class of the surface, just make a boolean array to indicate that there was an error. And don't do arithmetic on SurfaceClass (not saying that you, Xuan, did this to begin with).

mjwitte: @xuanluo113 Sure, just drop the class from the error message - good idea.

mjwitte (approved):
@xuanluo113 Code changes look good. I pulled in develop for one last round of CI, then this will get merged.

PR #8382, opened 2020-11-13
Remove EP_UNUSED and use C++17's [[maybe_unused]]

lefticus:
Pull request overview
---------------------

Move from `EP_UNUSED` macro to the C++17 `[[maybe_unused]]` attribute.

No expected changes.


Myoldmopar:
I'm super excited for this.  Every time we eliminate one of our custom solutions for something standard it makes me very happy.  Thanks for c++17, I know there are many things we can evolve in this way.

This looks totally ready, and since the other couple wide-reaching PRs look like they will need additional commits, I think I'll drop this in.  Thanks for this quick transition, @lefticus!

PR #8388, opened 2020-11-21
Global sim manager sim air

Myoldmopar:
Continued refactoring state.  Captured the following namespaces:

- EIRPlantLoopHeatPump
- PlantValves
- SetpointManager
- SimulationManager
- SingleDuct
- SizingManager
- SolarReflectionManager

In addition, started making better use of c++17 by using `make_unique`, and also added the start of a diagnostic script that will verify we are properly handling state clearing and instantiation.  The new script should run on Github Actions as part of the custom check gauntlet.

This is currently up to date with develop and could go in if clean.


Myoldmopar:
That is a _whole_ lotta green.  OK, merging this in, then pushing up an updated version of my unit test build speed up branch.

PR #8389, opened 2020-11-22
Unit test build speed improvement

Myoldmopar:
The state refactor is proceeding rapidly, but for anyone that routinely builds unit tests, you may have noticed a steady increase in build time.  It has just steadily increased to the point where building the unit tests is out of hand.  This has nothing to do with building EnergyPlus itself, just the unit tests.

The root reason for this is because the main unit test fixture, `EnergyPlusFixture`, which is needed in nearly every unit test file, itself needed to include the large `Data/CommonIncludes.hh` file.  This file is the master include file for only the few places that need to own and instantiate a state instance.  These include EnergyPlus' `main()` function, and the API entry points.  No other code should really _need_ that bulk, but because of the dependency between unit tests and the fixture, they essentially did.  

With this PR, we break that dependency by changing the fixture to hold a pointer to the state instance, rather than the actual instance itself.  This break means that each unit test CC file will no longer inherently include tons of unnecessary header bulk.  The build time changes are drastic, as seen on the dashboard, and summarized here for a few branches:

| Branch Name                | Windows Release | Ubuntu Release | Ubuntu Debug | Mac Release |
|----------------------------|-----------------|----------------|--------------|-------------|
| airwall-simplify           | 4644.87         | 2039.46        | 1623.61      | 1450.36     |
| fix_nocache_psychrometrics | 3495.13         | 1284.02        | 1435.78      | 1451.46     |
| ** UnitTestBuildSpeed      | 2615.19         | 448.12         | 614.18       | 736.16      |

**Massive** build time speed ups.  There is no significance found in runtime changes on CI, sometimes this branch is faster than other branches and sometimes it is slower.

OK, so what does this mean for the developers?  There are two changes you'll need to make in your unit test writing:
1. Accessing state members
 - State is a pointer now, so access the member using pointer member operator
 - Before: `state.dataSomething`
 - After: `state->dataSomething`
2. Passing state into E+ functions
 - State is a pointer now, so pass the contents of state using the indirection operator.  Note that all the functions inside EnergyPlus accept a reference to the state, so this will not _actually_ be sending the whole state object.
 - Before: `DoSomething(state, ...)`
 - After: `DoSomething(*state, ...)`

That's not a lot of differences, but this branch makes a world of difference for build time both locally as well as when getting CI results back.

I want to emphasize again that this changes **nothing** inside EnergyPlus code itself, this is purely for the unit test code.  I realize this will cause some conflicts, but they should be minor, easy to address, and as always I'm happy to help resolve them.


tst/EnergyPlus/unit/AdvancedAFN.unit.cc
```diff
@@ -85,28 +85,28 @@ TEST_F(EnergyPlusFixture, AirflowNetwork_AdvancedTest_Test1)
     int CurveNum;
 
     AirflowNetworkNumOfOccuVentCtrls = 1;
-    state.dataAirflowNetworkBalanceManager->OccupantVentilationControl.allocate(AirflowNetworkNumOfOccuVentCtrls);
-    state.dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinOpeningTime = 4;
-    state.dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinClosingTime = 4;
-    state.dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinTimeControlOnly = true;
+    state->dataAirflowNetworkBalanceManager->OccupantVentilationControl.allocate(AirflowNetworkNumOfOccuVentCtrls);
```
Myoldmopar, line 88: You can expect lots and lots and lots of these changes if you are perusing this PR.  No functional change, just operating on a pointer rather than a direct instance of state.

tst/EnergyPlus/unit/AdvancedAFN.unit.cc
```diff
@@ -85,28 +85,28 @@ TEST_F(EnergyPlusFixture, AirflowNetwork_AdvancedTest_Test1)
...
+    state->dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinOpeningTime = 4;
+    state->dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinClosingTime = 4;
+    state->dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).MinTimeControlOnly = true;
 
     TimeOpenElapsed = 3.0;
     TimeCloseElapsed = 0.0;
-    state.dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).calc(state, 1, TimeOpenElapsed, TimeCloseElapsed, OpenStatus, OpenProbStatus, CloseProbStatus);
+    state->dataAirflowNetworkBalanceManager->OccupantVentilationControl(1).calc(*state, 1, TimeOpenElapsed, TimeCloseElapsed, OpenStatus, OpenProbStatus, CloseProbStatus);
```
Myoldmopar, line 95: And here's the other change you'll see many, many times.  The guts of EnergyPlus expect a state instance to be passed in (technically a reference to it), so you just need to dereference the pointer with an asterisk and that's it.

Myoldmopar:
CI is super duper happy, so this should be ready to merge anytime.

mitchute:
Really great stuff @Myoldmopar! Merging.

PR #8390, opened 2020-11-24
Outside range warning and error when using linear interpolation with Schedule:Day:Interval

### LINKED ISSUE #6739: Incorrect outside range warning and error when using linear interpolation
Issue overview
--------------
When Schedule:Day:Interval using linear interpolation has values that are at the limits shown in referenced ScheduleTypeLimits object, get error messages:

```
    ** Warning ** ProcessScheduleInput: Schedule:Day:Interval="2LLO DAY", Values are outside of range for Schedule Type Limits Name=FRACTION
    ** Severe  ** ProcessScheduleInput: Schedule="2LLO-YR" has values outside its Schedule Type (FRACTION) range
    **   ~~~   **   Minimum should be >=0.000 and Maximum should be <=1.000
    **  Fatal  ** ProcessScheduleInput: Preceding Errors cause termination.
 

```
That are not correct, since the values should be ok at exactly the limits. This does not occur when the interpolation method gets changed to average

### Details
 - Windows
 - 8.9.0
 - https://unmethours.com/question/31336/values-are-outside-of-range-for-schedule-type-limits/

### Checklist
 - [ ] Defect file added (see devsupport)

jmythms:
Pull request overview
---------------------
 - Fixes #6739: Incorrect outside range warning and error when using linear interpolation with Schedule:Day:Interval.
 - There is a very tiny difference between the limit and maximum and minimum (~10^(-12)) that causes a warning (severe/fatal error). 
 - I changed the code to check if the difference between the values and the limits is (<= or >=) FLT_EPSILON, where FLT_EPSILON is defined as "Difference between 1 and the least value greater than 1 that is representable." ([Link](http://www.cplusplus.com/reference/cfloat/)), which is 1E-5 or smaller depending on system/compiler. I understand we are dealing with `Real64` variable type so not sure if this is ok. 
 - [Fixed] Unit test is failing: ZoneUnitarySysTest.UnitarySystemModel_SetpointControlCyclingFan. 

To Do:
 - [x] Warning still showing up if schedule touches lower limit. 
 - [x] Fix unit test. 
 - [x] Clean up code.
 - [x] Should I make a test case?

### Reviewer


DOAS_wNeutralSupplyAir_wFanCoilUnits.idf
```diff
@@ -0,0 +1,4577 @@
+!-Generator IDFEditor 1.50
```
mitchute: Is this new file needed to show the error? It's fine if it is, just want to make sure we need a new input file to demonstrate this or if we could get by with just adding a new unit test.

Also, I'm just now noticing that the PR description states that the ```ZoneUnitarySysTest.UnitarySystemModel_SetpointControlCyclingFan``` unit test was fixed, but I'm not seeing any changes to that file here. Should there be?

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -3827,11 +3827,11 @@ namespace ScheduleManager {
         }
 
         //  Min/max for schedule has been set.  Test.
-        MinValueOk = (Schedule(ScheduleIndex).MinValue >= Minimum);
+        MinValueOk = (FLT_EPSILON >= Minimum - Schedule(ScheduleIndex).MinValue);
```
mitchute, line 3791: These changes look OK.

third_party/ObjexxFCL/src/ObjexxFCL/Array.hh
```diff
@@ -1776,7 +1776,11 @@ public: // Comparison: Predicate: Any
 		assert( a.size_bounded() );
 		if ( a.empty() ) return false;
 		for ( size_type i = 0, e = a.size_; i < e; ++i ) {
-			if ( t < a[ i ] ) return true;
+			if (t < a[i]) {
```
mitchute: Hmm, we generally don't modify the third-party libraries directly. What's the reason for these changes? It would help if you could specifically outline your reasons for making these changes, such as pointing us to some code in E+ that's misbehaving so we can have some context.

mitchute (changes requested):
Thanks, @jmythms! This looks like you're on the right path. Please provide some additional details where noted.

third_party/ObjexxFCL/src/ObjexxFCL/Array.hh
Myoldmopar: I concur, I hope that this isn't needed.  I think what we probably need to do instead is just get rid of the call to `any_lt`, and instead do this loop/check ourselves right in EnergyPlus code.  That way we don't have to edit this.  But please speak up if I'm missing something here.  Otherwise this looks good.

jmythms: Yes. Getting rid of the `any_lt` call should fix the errors happening now. Just that there is a (very) remote possibility of this happening again in the future if someone else tries to use `any_lt` for an array that was linearly interpolated like the case here.

jmythms: As context, for my test, [`any_gt`](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/src/EnergyPlus/ScheduleManager.cc#L947) returns false which causes the warning in [this line](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/src/EnergyPlus/ScheduleManager.cc#L948). 

[`any_gt`](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/third_party/ObjexxFCL/src/ObjexxFCL/Array.hh#L1813) calls [`any_lt`](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/third_party/ObjexxFCL/src/ObjexxFCL/Array.hh#L1761) where the comparisons are done. 

I guess there are two `any_lt` functions as well ([here ](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/third_party/ObjexxFCL/src/ObjexxFCL/Array.hh#L1761) and [here](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/third_party/ObjexxFCL/src/ObjexxFCL/Array.hh#L1774)), and it's necessary to change both if we decide to change one. 

There's already a check in [this line](https://github.com/NREL/EnergyPlus/blob/7a650ed251d6fa927d61df916e81504ab3a55f9d/src/EnergyPlus/ScheduleManager.cc#L2278). I'm looking if this is enough to remove the other check.

DOAS_wNeutralSupplyAir_wFanCoilUnits.idf
jmythms: Thank you. I removed the IDF file and added a new unit test. It's my first unit test and I am not very sure if this is how it should be written. In the unit test, there are conditions (both upper and lower limit) that should make the test fail if the interpolation issue is not fixed. It fails on the current develop branch, passes on this branch. 

There shouldn't be any changes in the `ZoneUnitarySysTest.UnitarySystemModel_SetpointControlCyclingFan`. Sorry for the confusion, that was a message for myself 😅

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+        "  ,                               !- Condenser Air Inlet Node Name",
+        "  EvaporativelyCooled;            !- Condenser Type",
+
+        "ScheduleTypeLimits,",
+        "  Any Number;                     !- Name",
+
+        // This schedule should cause issues if interpolation issue is not fixed. Got the initial schedule from the unmethours issue.
+        "ScheduleTypeLimits, ",
+        "  Fractional,                     !-Name ",
+        "  0.1,                            !-Lower Limit Value",
```
jmythms: Lower limit is set here.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+        "  0.9,                            !-Upper Limit Value",
```
jmythms: Upper limit is set here.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+        "  Continuous;                     !-Numeric Type",
+
+        "Schedule:Day:Interval,",
+        "  2LLO Tuesday,                   !- Name",
+        "  Fractional,                     !- Schedule Type Limits Name",
+        "  Linear,                         !- Interpolate to Timestep",
+        "  06:00,                          !- Time 1 {hh:mm}",
+        "  0.1,                            !- Value Until Time 1",
```
jmythms: Lower limit is reached here.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+        "  07:15,                          !- Time 2 {hh:mm}",
+        "  0.9,                            !- Value Until Time 2",
```
jmythms: Upper limit.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+        "  1.5,                            !- Ratio of Initial Moisture Evaporation Rate and Steady State Latent Capacity {dimensionless}",
+        "  3,                              !- Maximum Cycling Rate {cycles/hr}",
+        "  45,                             !- Latent Capacity Time Constant {s}",
+        "  ,                               !- Condenser Air Inlet Node Name",
+        "  EvaporativelyCooled;            !- Condenser Type",
+
+        "ScheduleTypeLimits,",
+        "  Any Number;                     !- Name",
+
+        // This schedule should cause issues if interpolation issue is not fixed. Got the initial schedule from the unmethours issue.
```
jmythms: The schedule is the one that caused issues unmet hours: https://unmethours.com/question/31336/values-are-outside-of-range-for-schedule-type-limits/

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
     EXPECT_TRUE(compare_err_stream(error_string, true));
 }
 
+TEST_F(ZoneUnitarySysTest, UnitarySystemModel_ScheduleInterpolationTest)
```
jmythms: This unit test is based on the test before it: ([link](https://github.com/NREL/EnergyPlus/blob/f84f89f6d6cb935885ca7af3fea3664fe6f6516b/tst/EnergyPlus/unit/UnitarySystem.unit.cc#L15412))

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -15565,6 +15565,210 @@ TEST_F(ZoneUnitarySysTest, UnitarySystemModel_SetpointControlCyclingFan)
...
+    std::string compName = "UNITARY SYSTEM MODEL";
+    bool zoneEquipment = true;
+    UnitarySystems::UnitarySys::factory(*state, DataHVACGlobals::UnitarySys_AnyCoilType, compName, zoneEquipment, 0);
+    UnitarySystems::UnitarySys *thisSys = &state->dataUnitarySystems->unitarySys[0];
+
+    state->dataGlobal->NumOfTimeStepInHour = 1;
+    state->dataGlobal->MinutesPerTimeStep = 60;
+    DataZoneEquipment::ZoneEquipInputsFilled = true; // indicate zone data is available
+
+    try {
```
jmythms: Not sure if this is how it should be implemented. I also wanted to include an error message here, but I did not know how to do that.

src/EnergyPlus/ScheduleManager.cc
jmythms: Thanks.

third_party/ObjexxFCL/src/ObjexxFCL/Array.hh
jmythms: I removed these changes because the limits are being validated at a later part of the code, [here](https://github.com/NREL/EnergyPlus/blob/f84f89f6d6cb935885ca7af3fea3664fe6f6516b/src/EnergyPlus/ScheduleManager.cc#L2271).

nrel-bot-3:
@jmythms @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-2c:
@jmythms @lgentile it has been 28 days since this pull request was last updated.

mjwitte:
@jmythms Status on this one? It's still marked draft.

jmythms:
> @jmythms Status on this one? It's still marked draft.

Will pull in updates from latest develop and change this to ready for review. Sorry for the draft tag.

jmythms:
> @jmythms Status on this one? It's still marked draft.

This one is ready for review. Hoping I am still in the time limit for the 9.5 release.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
mjwitte: @jmythms It would be better for this test to focus only on the schedule without the extra baggage of a UnitarySystem. In fact, there are several existing unit tests for Schedule:Day:Interval in ScheduleManager.unit.cc `ScheduleDayInterval*`. I would either extend one of these (or copy one to make a new unit test) to specifically trip on this issue.

jmythms:
@mjwitte Thanks for the suggestion about the unit test, can I have another review on this please?

Myoldmopar:
This seems ready to go in.  @mitchute did you have anything further?  I'm currently building it locally and running tests.

mitchute:
I haven't tested this recently, but the changes look OK to me. As long as it's passing tests, I'd call it done.

mitchute (approved):

Myoldmopar:
OK, will report shortly.

Myoldmopar:
Everything looks good locally, merging.  Thanks @jmythms and @mitchute

jmythms:
Thank you @mitchute and @Myoldmopar

PR #8403, opened 2020-12-01
Refactoring of CalcInteriorSolarDistributions and reduce condition checkings

xuanluo113:
Pull request overview
---------------------
This pull request includes the following refactoring efforts
- Simplify the duplicated surface condition checking. (e.g. `ExtSolar` and ExtBoundary conditions)
- Continued branch/loop refactoring in `CalcInteriorSolarDistributions` by reordering code blocks and branches.
- Reduce solar array reinitialization by adding a flag to check if sunset at the current time step. The arrays once reset at sunset do not need to be reset afterward.

### Pull Request Author

### Reviewer


src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2474,52 +2405,53 @@ namespace HeatBalanceSurfaceManager {
...
         if (InitSurfaceHeatBalancefirstTime) {
             for (int SurfNum = 1; SurfNum <= TotSurfaces; ++SurfNum) {
                 SurfBmToDiffReflFacGnd(SurfNum) = Surface(SurfNum).ViewFactorGround;
                 SurfSkyDiffReflFacGnd(SurfNum) = Surface(SurfNum).ViewFactorGround;
             }
         }
+        bool currSolRadPositive = state.dataEnvrn->SunIsUp && (state.dataEnvrn->BeamSolarRad + state.dataEnvrn->GndSolarRad + state.dataEnvrn->DifSolarRad > 0.0);
+        bool sunset = (!currSolRadPositive) && state.dataEnvrn->PreviousSolRadPositive;
+        bool sunIsUpNoRad = state.dataEnvrn->SunIsUp && (!currSolRadPositive);
+        bool resetSolar = state.dataGlobal->BeginDayFlag || sunIsUpNoRad || sunset; // Reset at (1) Beginning of simulation (2) sunset time, and SunIsUp but not solar time.
```
mjwitte: The comment says reset at beginning of simulation, but `BeginDayFlag` is true at the beginning of every day. Perhaps this should be `BeginEnvrnFlag`?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2322,37 +2320,9 @@ namespace HeatBalanceSurfaceManager {
         Array1D<Real64> currGndSolarInc(TotSurfaces); // Ground diffuse solar incident on a surface
         // Always initialize the shortwave quantities
         for (int SurfNum = 1; SurfNum <= TotSurfaces; ++SurfNum) {
-
             SurfInitialDifSolInAbsReport(SurfNum) = 0.0;
```
mjwitte: Does this need to be reset every time?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2322,37 +2320,9 @@ namespace HeatBalanceSurfaceManager {
...
-            SurfQRadSWOutIncidentBeam(SurfNum) = 0.0;
-            SurfQRadSWOutIncidentSkyDiffuse(SurfNum) = 0.0;
-            SurfQRadSWOutIncidentGndDiffuse(SurfNum) = 0.0;
-
-            SurfQRadSWOutIncBmToDiffReflGnd(SurfNum) = 0.0;
-            SurfQRadSWOutIncSkyDiffReflGnd(SurfNum) = 0.0;
-            SurfQRadSWOutIncBmToBmReflObs(SurfNum) = 0.0;
-            SurfQRadSWOutIncBmToDiffReflObs(SurfNum) = 0.0;
-            SurfQRadSWOutIncSkyDiffReflObs(SurfNum) = 0.0;
             SurfCosIncidenceAngle(SurfNum) = 0.0;
```
mjwitte: Same for this, move to `ifResetSolar` block?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2377,83 +2347,44 @@ namespace HeatBalanceSurfaceManager {
...
-                SurfWinInsRevealDiffOntoFrame(SurfNum) = 0.0;
-                SurfWinInsRevealDiffOntoGlazingReport(SurfNum) = 0.0;
-                SurfWinInsRevealDiffIntoZoneReport(SurfNum) = 0.0;
-                SurfWinInsRevealDiffOntoFrameReport(SurfNum) = 0.0;
-                SurfWinBmSolAbsdInsRevealReport(SurfNum) = 0.0;
-                SurfWinBmSolTransThruIntWinRepEnergy(SurfNum) = 0.0;
-                SurfWinBmSolRefldOutsRevealRepEnergy(SurfNum) = 0.0;
-                SurfWinBmSolRefldInsRevealRepEnergy(SurfNum) = 0.0;
                 SurfWinProfileAngHor(SurfNum) = 0.0;
                 SurfWinProfileAngVert(SurfNum) = 0.0;
```
mjwitte: Move?
Also `SurfWinSysSolReflectance` and `SurfWinSysSolAbsorptance` below?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2377,83 +2347,44 @@ namespace HeatBalanceSurfaceManager {
         for (int zoneNum = 1; zoneNum <= state.dataGlobal->NumOfZones; ++zoneNum) {
             int const firstSurfOpaq = Zone(zoneNum).NonWindowSurfaceFirst;
             int const lastSurfOpaq = Zone(zoneNum).NonWindowSurfaceLast;
-            if (firstSurfOpaq == -1) continue;
             for (int SurfNum = firstSurfOpaq; SurfNum <= lastSurfOpaq; ++SurfNum) {
```
mjwitte, line 2344: There are ~not~ *now* four `for` loops with the same limits here. Combine them?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2529,22 +2461,87 @@ namespace HeatBalanceSurfaceManager {
             for (int zoneNum = 1; zoneNum <= state.dataGlobal->NumOfZones; ++zoneNum) {
                 int const firstSurfWin = Zone(zoneNum).WindowSurfaceFirst;
                 int const lastSurfWin = Zone(zoneNum).WindowSurfaceLast;
-                if (firstSurfWin == -1) continue;
+                for (int SurfNum = firstSurfWin; SurfNum <= lastSurfWin; ++SurfNum) {
```
mjwitte, line 2452: More loops that can be combined? I assume these were moved in separate chunks over the course of the development. If there's a reason to keep them separate, that's fine.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2474,52 +2405,53 @@ namespace HeatBalanceSurfaceManager {
...
+                SurfQRadSWOutIncident(SurfNum) = 0.0;
+                SurfQRadSWOutIncidentBeam(SurfNum) = 0.0;
+                SurfQRadSWOutIncidentSkyDiffuse(SurfNum) = 0.0;
+                SurfQRadSWOutIncidentGndDiffuse(SurfNum) = 0.0;
+
+                SurfQRadSWOutIncBmToDiffReflGnd(SurfNum) = 0.0;
+                SurfQRadSWOutIncSkyDiffReflGnd(SurfNum) = 0.0;
+                SurfQRadSWOutIncBmToBmReflObs(SurfNum) = 0.0;
+                SurfQRadSWOutIncBmToDiffReflObs(SurfNum) = 0.0;
+                SurfQRadSWOutIncSkyDiffReflObs(SurfNum) = 0.0;
```
mjwitte, line 2432: Many (all?) of these are also getting set to zero in `InitSolarCalculations`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2569,9 +2566,9 @@ namespace HeatBalanceSurfaceManager {
                     SurfBmToDiffReflFacGnd(SurfNum) = 0.0;
                 }
             }
+        }
```
mjwitte, line 2568: I was going to comment here that Lines 2552:2558 above are redundant since these variables have been reset above for all SurfWin, but then I wasn't sure if domes are included in `firstSurfWin::lastSurfWin`? If not, a comment here explaining that would be useful. (Github won't let me comment on untouched lines.)

src/EnergyPlus/SolarShading.cc
```diff
@@ -5974,6 +5970,8 @@ namespace SolarShading {
         static Array2D<Real64> AbsSolDiffEQL(2, CFSMAXNL + 1);     // absorbed exterior diffuse radiation by layers fraction
         static Array2D<Real64> AbsSolBeamBackEQL(2, CFSMAXNL + 1); // absorbed interior beam radiation by layers fraction from back
         // Array2D< Real64 > AbsSolDiffBackEQL( CFSMAXNL+1, 2 ); // absorbed exterior diffuse radiation by layers fraction from back //Unused
+        static Array1D<Real64> WinTransBmBmSolar; // Factor for exterior beam to beam solar transmitted through window, or window plus shade, into zone at current time (m2)
+        static Array1D<Real64> WinTransBmDifSolar; // Factor for exterior beam to diffuse solar transmitted through window, or window plus shade, into zone at current time (m2)
```
mjwitte, line 5974: Remind me why we want (some of) these arrays to be `static`?

src/EnergyPlus/SolarShading.cc
```diff
@@ -6069,9 +6039,6 @@ namespace SolarShading {
             // EXTERIOR BEAM SOLAR RADIATION ABSORBED ON THE OUTSIDE OF OPAQUE SURFACES
             //-------------------------------------------------------------------------
             // TODO: use opaq and window loop after airboundary is sorted
-//            int const firstSurfOpaq = Zone(enclosureNum).NonWindowSurfaceFirst;
-//            int const lastSurfOpaq = Zone(enclosureNum).NonWindowSurfaceLast;
-
             for (int const SurfNum : thisEnclosure.SurfacePtr) {
                 if (Surface(SurfNum).Class != SurfaceClass::Window && Surface(SurfNum).Class != SurfaceClass::TDD_Dome) {
                     if (!Surface(SurfNum).HeatTransSurf) continue;
```
mjwitte, line 6045: 1. It may be useful to sort `SurfacePtr` to group windows and domes together and then opaque surfaces together so this loop can be over just the appropriate surfaces and eliminate the first `if` here - something for the ToDo list.
2. Is it possible for a shading surface to end up in an enclosure?

src/EnergyPlus/SolarShading.cc
```diff
@@ -6087,10 +6054,11 @@ namespace SolarShading {
             }
 
             //--------------------------------------------------------------------------------------------------------
-            // EXTERIOR BEAM AND DIFFUSE SOLAR RADIATION ABSORBED IN THE GLASS LAYERS OF EXTERIOR WINDOWS OR TDD DOMES
+            // EXTERIOR WINDOWS OR TDD DOMES
             //--------------------------------------------------------------------------------------------------------
             for (int const SurfNum : thisEnclosure.SurfacePtr) {
                 if (Surface(SurfNum).Class != SurfaceClass::Window && Surface(SurfNum).Class != SurfaceClass::TDD_Dome) continue;
+                if (!Surface(SurfNum).ExtSolar && SurfWinOriginalClass(SurfNum) != SurfaceClass::TDD_Diffuser) continue;
```
mjwitte, line 6062: Another ToDo list item: sort enclosures by extsolar windows and domes, then non-extsolar windows and domes(?).

src/EnergyPlus/SolarShading.cc
```diff
@@ -6600,169 +6470,167 @@ namespace SolarShading {
...
-                            if (ShadeFlag != IntBlindOn && ShadeFlag != ExtBlindOn && ShadeFlag != BGBlindOn && ShadeFlag != ExtScreenOn) {
 
-                                // Shade on or switchable glazing
+                //-----------------------------------------------------------------
+                // BLOCK 3 - DIFFUSE SOLAR TRANSMITTANCE
+                //-----------------------------------------------------------------
 
-                                if (SunLitFract > 0.0) TBmAllShBlSc = POLYF(CosInc, state.dataConstruction->Construct(ConstrNumSh).TransSolBeamCoef);
 
+                if (SurfWinWindowModelType(SurfNum) != WindowBSDFModel && SurfWinWindowModelType(SurfNum) != WindowEQLModel) {
```
mjwitte: Another ToDo list item: `SurfWinWindowModelType` is initialized to `Window5DetailedModel` so this could be a simple `if ... ==Window5DetailedModel` and/or this should be a enum and make this a set of case blocks.
This shows up in many places in the code.

src/EnergyPlus/SolarShading.cc
```diff
@@ -6600,169 +6470,167 @@ namespace SolarShading {
...
+                        if (ShadeFlag == ExtScreenOn) {
+                            SurfWinScTsolDifDif(SurfNum) = SurfaceScreens(ScNum).DifDifTrans;
+                        } else {
+                            SurfWinBlTsolDifDif(SurfNum) = InterpSlatAng(SlatAng, VarSlats, Blind(BlNum).SolFrontDiffDiffTrans);
+                        }
+                    }
+                }
 
+                //-----------------------------------------------------------------
+                // BLOCK 4 - SCREEN, BLINDS AND GLAZING SYSTEM SOLAR TRANSMITTANCE
```
mjwitte: If I'm following correctly, this is "BLOCK 4 - SCREEN, BLINDS AND GLAZING SYSTEM **BEAM** SOLAR TRANSMITTANCE"

src/EnergyPlus/SolarShading.cc
```diff
@@ -6905,58 +6769,61 @@ namespace SolarShading {
...
-                        if (ShadeFlag == IntShadeOn || ShadeFlag == ExtShadeOn || ShadeFlag == BGShadeOn || SurfWinSolarDiffusing(SurfNum) ||
-                            SurfWinOriginalClass(SurfNum) == SurfaceClass::TDD_Diffuser || Surface(SurfNum).Class == SurfaceClass::TDD_Dome)
-                            continue;
+                // If shade is in place or there is a diffusing glass layer there is no interior beam
+                // from this exterior window since the beam-beam transmittance of shades and diffusing glass
+                // is assumed to be zero. The beam-beam transmittance of tubular daylighting devices is also
+                // assumed to be zero.
 
+                if (SurfWinWindowModelType(SurfNum) != WindowBSDFModel)
+                    if (ShadeFlag == IntShadeOn || ShadeFlag == ExtShadeOn || ShadeFlag == BGShadeOn || SurfWinSolarDiffusing(SurfNum) ||
```
mjwitte: There's an indentation problem here, because the line above isn't really doing anything:
`if (SurfWinWindowModelType(SurfNum) != WindowBSDFModel)` ??

src/EnergyPlus/SolarShading.cc
```diff
@@ -6905,58 +6769,61 @@ namespace SolarShading {
...
+                        SurfWinOriginalClass(SurfNum) == SurfaceClass::TDD_Diffuser || Surface(SurfNum).Class == SurfaceClass::TDD_Dome)
+                        continue;
+                if (SunLitFract > 0.0) {
```
mjwitte, line 6790: Wondering why this was moved here? This seems like a faster check than the compound if above which can be skipped if the surface doesn't have any beam solar.

src/EnergyPlus/SolarShading.cc
```diff
@@ -7006,21 +6869,25 @@ namespace SolarShading {
...
                                     // The layer order for interior windows is "outside" to "inside," where "outside" refers to
                                     // the adjacent zone and "inside" refers to the current zone.
                                     int ShadeFlagBack = SurfWinShadingFlag(BackSurfNum);
-                                    Real64 SlatAngBack = SurfWinSlatAngThisTS(BackSurfNum);
+//                                    int ConstrNum = Surface(SurfNum).Construction;
+//                                    int ConstrNumSh = Surface(SurfNum).activeShadedConstruction;
+//                                    if (SurfWinStormWinFlag(SurfNum) == 1) {
+//                                        ConstrNum = Surface(SurfNum).StormWinConstruction;
+//                                        ConstrNumSh = Surface(SurfNum).activeStormWinShadedConstruction;
+//                                    }
```
mjwitte: Ready to delete these lines?

src/EnergyPlus/SolarShading.cc
```diff
@@ -7038,7 +6905,6 @@ namespace SolarShading {
 
                                     // Interior beam absorptance of glass layers and beam transmittance of back exterior  &
                                     // or interior window WITHOUT SHADING this timestep
-
```
mjwitte, line 7041: A possible defect (from, long long ago). Lines 6899 and 6900 look suspect:
```
int ConstrNumBackSh = Surface(BackSurfNum).activeShadedConstruction;
                                    if (SurfWinStormWinFlag(BackSurfNum) == 1) {
                                        ConstrNum = Surface(BackSurfNum).StormWinConstruction;
                                        ConstrNumSh = Surface(BackSurfNum).activeStormWinShadedConstruction;
                                    }
```
Seems like these should be setting `ConstrNumBack` and `ConstrNumBackSh`.

src/EnergyPlus/SolarShading.cc
```diff
@@ -6852,50 +6730,36 @@ namespace SolarShading {
...
-                        TBmAll = TBmBm + TBmDif;
-                    }
-
-                    int FenSolAbsPtr = WindowScheduledSolarAbs(SurfNum, ConstrNum);
-
-                    // Window is schedule surface gained. Do not make addition to what enters into zone since that information is not
-                    // available
-                    if (FenSolAbsPtr == 0) {
+                    // Window is schedule surface gained. Do not make addition to what enters into zone since that information is not available
+                    if (WindowScheduledSolarAbs(SurfNum, ConstrNum) == 0) {
```
mjwitte: `WindowScheduledSolarAbs` only applies to BSDF windows (ComplexFenestrationProperty:SolarAbsorbedLayers) and it's a function so we don't want to call it if it's not necessary. So this section could be rearranged to check `WindowScheduledSolarAbs` only for BSDF windows.
Also since `WindowScheduledSolarAbs` is evaluated multiple times for BSDF windows to see if it's zero, it seems this could be set once at the top of the surface loop.

src/EnergyPlus/SolarShading.cc
```diff
@@ -7472,7 +7317,7 @@ namespace SolarShading {
                                         } else {
                                             if (state.dataConstruction->Construct(ConstrNumBack).TransDiff <= 0.0) {
                                                 // Do not take into account this window if it is scheduled for surface gains
-                                                SurfSolIncPtr = SurfaceScheduledSolarInc(BackSurfaceNumber, ConstrNumBack);
+                                                int SurfSolIncPtr = SurfaceScheduledSolarInc(BackSurfaceNumber, ConstrNumBack);
```
mjwitte: `SurfSolIncPtr` can be set at the top of this loop at line 7246 and then used in both if statements. And it's tempting to add a global bool for `AnyScheduledSurfaceSolar` to avoid the extra steps in `int SurfaceScheduledSolarInc` when there aren't any.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10775,21 +10608,21 @@ namespace SolarShading {
         //  REAL(r64)    :: TransGl,ReflGl,AbsGl ! Glass layer solar transmittance, reflectance, absorptance
 
         Real64 ViewFactor;                  // temp var for view factor
-        Real64 ViewFactorTotal;             // debug var for view factor total
+//        Real64 ViewFactorTotal;             // debug var for view factor total
```
mjwitte, line 10606: I'm assuming you want to keep these lines for the next phase of work.

mjwitte (commented):
@xuanluo113 This looks good. I've made some comments below. It's up to you whether to add anything here or wait until the next round. I've merged in develop and pushed for another round of CI results.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Amir suggested this way to reuse the registers for the array pointers, as there are 16 registers in each CPU, and ~12 can be used for each of the array pointers in the loop at most (the rest to be used for `int SurfNum `, `firstSurfOpaq`, etc). Maintaining ~12 pointers would avoid overwriting these registers to pointing to something else. Something like a cache level speed-up.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Modified.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Seems `InitSolarCalculations` is a one time initialization at `beginSimFlag`. Here is to reset at sunset time.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Same as above.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Domes are currently placed on the SurfOpaq list. My very next step is to sort the TDD surface. I think Dome should be placed at the end of a zone surface list, but not necessarily to be included in `firstSurfWin::lastSurfWin`, as they only partially share window solar behaviors.
I'd mark this as a TODO.

src/EnergyPlus/SolarShading.cc
xuanluo113: I think I just mindlessly copied the format of the arrays above. I don't really understand why they have to be static. I vaguely remember Edwin @Myoldmopar commented on my previous (unrelated) PR of avoiding using static arrays, as they are hard to reset during consecutive runs. But I think I should double-check.

src/EnergyPlus/SolarShading.cc
xuanluo113: 1. On my to-do list.
2. I won't be worried here as shading surfaces are non-HeatTransSurf.

src/EnergyPlus/SolarShading.cc
xuanluo113: Or `thisEnclosure.WinSurfPtr` (including dome) and `thisEnclosure.OpaqSurfPtr`? First, they are also index arrays as `.SurfacePtr`, and window and opaque surfaces are always treated separately.

src/EnergyPlus/SolarShading.cc
xuanluo113: I'll do this in THIS PR after this round of CI checking the array re-init are okay.

src/EnergyPlus/SolarShading.cc
xuanluo113: This is fine as the final statement is only one line. C++ seems okay to understand
```
    if (a > 0)
        if (b > 0)
            cout<<"Hello World";
``` 
without brackets or indentations.

src/EnergyPlus/SolarShading.cc
xuanluo113: Fixed.

src/EnergyPlus/SolarShading.cc
xuanluo113: Yes.

src/EnergyPlus/SolarShading.cc
xuanluo113: Great catch!

src/EnergyPlus/SolarShading.cc
xuanluo113: Fixed.

src/EnergyPlus/SolarShading.cc
xuanluo113: I forgot to mention this in the documentation. Here the issue was I found that the commented variable and calculations are only used for debugging by the developers. They are not actually used anywhere. I wanted to removed them but was hesitating. Looks like they are more suitable to be moved to a unit test.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Fixed.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Fixed.

xuanluo113:
@mjwitte Thanks for the thorough review. I addressed or replied to most of them. In this PR I also plan to address
- Window5DetailedModel
- check WindowScheduledSolarAbs only for BSDF windows
- add a global bool for AnyScheduledSurfaceSolar to avoid the extra steps in int SurfaceScheduledSolarInc when there aren’t any.

as you suggested after the CI is okay with the current changes.
The issue regarding sorting TDD surfaces is planed to be included in future PRs.

xuanluo113:
@mjwitte I finished addressing the rest of the comments. Specifically,

- for WinfowModelType, I converted the conditions from checking (not EQL and not Complex) windows to `if ModelType == Window5DetailedModel`. I plan to convert the model types to ENUM as well, but I think that could happen after the ShadingFlag->ENUM feature.
- Regarding "adding a global bool for AnyScheduledSurfaceSolar", I searched globally and there are only two places calling `SurfaceScheduledSolarInc` and both only happen when complex windows exist. I thought it would be fine to keep those two as they are.

src/EnergyPlus/SolarShading.cc
xuanluo113: @mjwitte I was wrong before. I think the `static` arrays used here are to make sure those arrays only got initialized once, but not at every time step (line 5979 here).

src/EnergyPlus/SolarShading.cc
mjwitte: Yes, `...WinSurfPtr` and `...OpagSurfPtr` sounds good.

mjwitte (approved):
@xuanluo113 This is a great piece of work. Thanks for addressing or explaining comments. CI is green. Anything more you want to do here before I merge?

xuanluo113:
@mjwitte No more in this PR. Thanks for the review.

PR #8404, opened 2020-12-01
Fix build with python on windows

mjwitte:
Pull request overview
---------------------
 - Fixes build with Python problem with VS2019 on windows

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/PluginManager.cc
+++ b/src/EnergyPlus/PluginManager.cc
@@ -705,7 +705,7 @@ namespace PluginManagement {
 
             for (unsigned long itemNum = 0; itemNum < numVals; itemNum++) {
                 PyObject *item = PyList_GetItem(pyth_val, itemNum);
-                if PyUnicode_Check(item) { // NOLINT(hicpp-signed-bitwise) -- something inside Python code causes warning
+                if (PyUnicode_Check(item)) { // NOLINT(hicpp-signed-bitwise) -- something inside Python code causes warning
                     std::string traceback_line = PyUnicode_AsUTF8(item);
                     if (!traceback_line.empty() && traceback_line[traceback_line.length()-1] == '\n') {
                         traceback_line.erase(traceback_line.length()-1);


src/EnergyPlus/PluginManager.cc
```diff
@@ -705,7 +705,7 @@ namespace PluginManagement {
 
             for (unsigned long itemNum = 0; itemNum < numVals; itemNum++) {
                 PyObject *item = PyList_GetItem(pyth_val, itemNum);
-                if PyUnicode_Check(item) { // NOLINT(hicpp-signed-bitwise) -- something inside Python code causes warning
+                if (PyUnicode_Check(item)) { // NOLINT(hicpp-signed-bitwise) -- something inside Python code causes warning
```
Myoldmopar, line 708: I do not understand how this was compiling before.  The macro expansion does not result in a parenthesized expression.  

Thanks for catching this!

Myoldmopar (approved):

Myoldmopar:
Just can't believe this ever compiled.  I don't get it.  Sheesh.  Great fix @mjwitte, thank you!

src/EnergyPlus/PluginManager.cc
jmarrec: @mjwitte Did you upgrade to python 3.9 by any chance?

I'm tracking the current cpython develop versus 3.8.6 and I think I know why.

Here is current develop of cpython, (I made adjustements to avoid python types): https://godbolt.org/z/ncYWzj
https://github.com/python/cpython/blob/5977a7989d49c3e095c7659a58267d87a17b12b1/Include/object.h#L672-L686

And here is 3.8.6, with an extra parenthesis in the macro expansion...: 

https://godbolt.org/z/Prrhao

https://github.com/python/cpython/blob/db455296be5f792b8c12b7cd7f3962b52e4f44ee/Include/object.h#L348-L351

mjwitte: @jmarrec I did, actually install 3.9. Not by choice, but because I had to reinstall everything  due to a system failure.

Myoldmopar: Thanks Python :laughing:

PR #8410, opened 2020-12-04
#8376 - EnergyPlus with option --expandobjects set fails when working directory is not on the same device as the output directory (on Unix)

jmarrec:
Pull request overview
---------------------

- Fix  #8376
 - EnergyPlus with option --expandobjects set fails when working directory is not on the same device as the output directory (on Unix)

This is a **minimal** fix for the specific `moveFile` issue that #8376 touches upon. 

(I have spent a lot of time implementing a solution that makes use of C++17 `std::filesystem`, but I'm going to put up a NFP for this to see if I can get buy in first)

I added a bunch of new tests for FileSystem.unit.cc, including some that I commented out or adjusted because they currently fail (I think they shouldn't). 

The specific issue for #8376 is not tested in a unit test... As it requires having an mount point, like a secondary hard drive or whatnot and we cannot ensure that the host running the test does have that. I have tested locally that this fixes it though.

I have a few other drives other than the one my OS (Ubuntu 18.04) is on, including one named 'DataExt4'. Before the fix, I can reproduce the issue as per below. After fix simulation works fine.

```bash
$ pwd
/home/julien/Software/Others/EnergyPlusDevSupport/DefectFiles/8000s

$ /bin/rm -Rf /media/DataExt4/Software/tmp/ && mkdir -p /media/DataExt4/Software/tmp/

$ energyplus -x -w /home/julien/Software/Others/EnergyPlus/weather/USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw -d /media/DataExt4/Software/tmp/out-ori HVACTemplate-5ZoneVAVFanPowered.idf

ExpandObjects Started.
 Begin reading Energy+.idd file.
 Done reading Energy+.idd file.
ExpandObjects Finished. Time:     0.044
EnergyPlus Starting
EnergyPlus, Version 9.4.0-998c4b761e, YMD=2020.12.02 15:42
**FATAL:Input file path /media/DataExt4/Software/tmp/out-ori/eplusout.expidf not found
EnergyPlus Run Time=00hr 00min  0.10sec
Program terminated: EnergyPlus Terminated--Error(s) Detected
```

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [ ] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
[...]

Issue #8376, opened 2020-11-12
EnergyPlus with option --expandobjects set fails when working directory is not on the same device as the output directory (on Linux)


MaximilianAzendorf:
Issue overview
--------------
ExpandObjects writes two files named expanded.idf and expandedidf.err into the working directory and then moves them (via rename() unix syscall) into the output directory. This does not work if the working directory is not on the same device as the output directory (e.g. a working directory in /tmp while the output directory is under an externally mounted /home).

### Details
Some additional details for this issue (if relevant):
 - Platform: Arch Linux x64 (linux 5.9.4)
 - Version of EnergyPlus: 9.4.0

### Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

jmarrec:
@MaximilianAzendorf Thanks for the issue report.

This the Fortran `OPEN` command that is indeed used to create the file.

https://github.com/NREL/EnergyPlus/blob/4e2af181c66c3a55748df2f7167ae5aae41a7e15/src/ExpandObjects/epfilter.f90#L1935-L1962

The CLI will then move it via `moveFile`: https://github.com/NREL/EnergyPlus/blob/734af19f9a7fc87715a5078c529a4083d2dff7ce/src/EnergyPlus/CommandLineInterface.cc#L560-L588

`moveFile` does indeed use a `rename` syscall on Unix: https://github.com/NREL/EnergyPlus/blob/734af19f9a7fc87715a5078c529a4083d2dff7ce/src/EnergyPlus/FileSystem.cc#L258-L266

PR #8411, opened 2020-12-07
Upgrade to C++17 std::filesystem library

jmarrec:
Pull request overview
---------------------

This is a NFP to upgrade `FileSystem` to use C++17 `std::filesystem` instead of relying on macros to call OS-specific functions, 

example: https://github.com/NREL/EnergyPlus/blob/7f35ab639d5a75b6e090a0db19ec83c3f0199025/src/EnergyPlus/FileSystem.cc#L258-L266

This potential upgrade was discovered via #8376. A minimal fix that doesn't use `std::filesystem` was made in #8410 and it is now proposed to make it a larger defect and do the upgrade to `std::filesystem`.

### Pull Request Author

### Reviewer

### DIFF
--- a/design/FY2020/NFP-filesystem.md
+++ b/design/FY2020/NFP-filesystem.md
@@ -1,0 +1,48 @@
+Use C++17 std::filesystem library
+==================================
+
+**Julien Marrec, EffiBEM**
+
+ - Original Date: 2020-12-07
+ - Revision Date
+
+## Justification for New Feature ##
+
+[#8376](https://github.com/NREL/EnergyPlus/issues/8376) identified a bug with the `EnergyPlus::FileSystem::moveFile(std::string const &filePath, std::string const &destination)` function on Unix that calls the stdio [`rename`](https://www.cplusplus.com/reference/cstdio/rename/). The rename method would fail when `filePath` and `destination` aren't on the same device (eg: two different hard drives).
+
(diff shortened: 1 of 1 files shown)


Myoldmopar (commented):
I personally love this idea, and frankly would not mind going all the way to # 2 and using fs::path everywhere.  The more we can push toward standard library and standard practice things, the lower the barrier for bringing on new devs, and in most cases at least, the lower our maintenance. 

I am curious about those handful of other places that use FileSystem.hh, most importantly the Schedule:File handling and the Python Plugin handling.
But I would love to know what others think.

Myoldmopar:
With a stray FileSystem unit test failing spuriously for me, I'm even more inclined to push for full `std::filesystem` implementation to reduce our maintenance with file system operations.  Anyone else have comment?  I'm also inclined to go ahead and push toward using the `std::filesystem::path` in the codebase to make it more familiar to modern C++ developers.

@jmarrec if you don't hear any reason why we **shouldn't** move to the standard C++ FileSystem library, then please proceed to fix this issue.  And consider the unit test failure another defect that this will resolve (https://github.com/NREL/EnergyPlus/issues/8430)

Myoldmopar:
@jmarrec no one has provided any reason to not push ahead and just bring in full file system library support.  Please go ahead and do that.

jmarrec:
@Myoldmopar I'm on it.

jmarrec:
Superseeded by #8480

PR #8414, opened 2020-12-09
Global state refactor continued

Myoldmopar:
Pull request overview
---------------------
 - This gets the number of global variables hanging out there down to around 5200, down from like 7000 before this branch!
 - A vast majority of the global variables were actually integers and characters hanging around that were converted into constexpr and enums wherever possible.
 - There will be an upcoming branch where the state argument is reworked, this branch does not address that, just stay tuned.

This is a big set of changes, but I think it will be totally clean on CI.


Myoldmopar:
Only CI issues are now the spurious diffs in the shading file and the integration coverage dipping below the limit.  This looks to be ready to go.  I will hold on merging momentarily while I do final review on the air wall PR and possibly drop that in first.  But as long as that one is good to go then these will both drop in sometime today.

Myoldmopar:
Alright got develop pulled in, everything is looking great here locally.  Enabled the unit test that was failing as well.  If CI is clean, this beast can drop in.

Myoldmopar:
OK, this looks good to go in.  This is a big set of changes, and there will be a follow-on branch that addresses the state-as-an-argument issue.  If you are having lots of conflicts, just let me know and I'll pull your branch and help get them addressed.  I'll be dropping this in now... Thanks @mitchute and @jmythms for the big effort here.  Among all the actual global "variables" that were addressed, another 400 or so `int const` variables were converted to enums. :100:

PR #8416, opened 2020-12-09
Add check for capacity limit of zero for ZoneHVAC:IdealLoadsAirSystem

### LINKED ISSUE #8365: Ideal Loads with zero cooling capacity is reporting values for Zone Air System Sensible Cooling Rate
Issue overview
--------------
The defect file ideal loads system has cooling capacity set to zero, yet there are significant amounts of "Zone Air System Sensible Cooling Rate" being reported. These should be zero or very close to it.

### Details
 - Version of EnergyPlus v9.4.0 release
 - @shorowit e-mail 23 Oct 2020

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #8365 
 - This pull request adds a check for the ZoneHVAC:IdealLoadsAirSystem object for when the 'Heating Limit' or 'Cooling Limit' is set to "LimitCapacity" and the 'Maximum Sensible Heating Capacity' or 'Maximum Total Cooling Capacity' is set to zero.  Previously, this setting would result in zero heat transfer but the "Zone Air System Sensible Cooling Rate" or "Zone Air System Total Heating Rate" would be nonzero.  This was due to `SupplyMassFlowRateForCool` being based on the cooling demand and set to a nonzero number.  This was being passed into the calculation in the ZoneTempPredictorCorrector.cc that was modifying the zone temperature and using the nonzero supply mass flow rate to compare to the inlet conditions and thus calculating a nonzero output for these variables.  This is fixed by adding a check after the initial calculations for `SupplyMassFlowRate*` to zero out `SupplyMassFlowRate*`if the heating/cooling capacity is limited to zero.

### Pull Request Author

### Reviewer


src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2245,6 +2245,15 @@ namespace PurchasedAirManager {
                     }
                 }
 
+                // If cooling capacity is limited to zero, SupplyMassFlowRate* should be set to zero
```
nealkruis, line 2169: @matthew-larson, @mjwitte I'm wondering if we scale the mass flow rate any time the load is greater than the capacity limit (e.g., `SupplyMassFlowRate = (CoolingLimit/QZnCoolSP) / CpAir / DeltaT`), and not just when the capacity limit is zero. This would modulate the flow rate to supply a consistent deltaT whereas, in the current solution, the DetlaT could drop considerably if the load is greater than the capacity limit.

nealkruis (changes requested):

src/EnergyPlus/PurchasedAirManager.cc
```diff
@@ -2245,6 +2245,15 @@ namespace PurchasedAirManager {
...
+                if (((PurchAir(PurchAirNum).CoolingLimit == LimitCapacity) ||
+                    (PurchAir(PurchAirNum).CoolingLimit == LimitFlowRateAndCapacity)) &&
+                    (PurchAir(PurchAirNum).MaxCoolTotCap == 0)) {
+                    SupplyMassFlowRateForCool = 0;
+                    SupplyMassFlowRateForDehum = 0;
+                    SupplyMassFlowRateForHumid = 0;
```
nealkruis, line 2175: Isn't humidification a heating process? I don't understand why its flow rate is zeroed out.

I imagine the same issue exists for heating limits.

src/EnergyPlus/PurchasedAirManager.cc
mjwitte: @nealkruis There are already adjustments for capacity limits in place further down in the function. I don't think this is a good reason to change that. That section should have been setting the deltaT to zero so even with some flow it should still have no impact on the zone. But the supply wasn't coming through exactly as the inlet so it reported some load. So this guarantees that the system is off if capacity is zero and there is no OA flow.  

@matthew-larson Did you track down why the inlet/outlet conditions weren't identical?

src/EnergyPlus/PurchasedAirManager.cc
mjwitte: This is the cooling section, so we want everything off here unless there's OA flow. Likewise, the dehumid mass flow is set to zero in the heating section.

src/EnergyPlus/PurchasedAirManager.cc
matthew-larson: @mjwitte I found where the zone temperature is modified in the [CorrectZoneAirTemp](https://github.com/NREL/EnergyPlus/blob/9d6116a760ca6317e55196ab157d33f570a42f8b/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L4739) function in ZoneTempPredictorCorrector.cc.  The "Zone Air System Sensible Cooling Rate" is based on `SNLoadCoolRate` which references back to the calculation for [SNLoad](https://github.com/NREL/EnergyPlus/blob/9d6116a760ca6317e55196ab157d33f570a42f8b/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L5059).  The `Node(ZoneNodeNum).Temp` of this calculation is from the corrected zone temperature, but `ZoneEnthalpyIn` traces back to `SumSysMCpT` calculated from the [CalcZoneSums](https://github.com/NREL/EnergyPlus/blob/9d6116a760ca6317e55196ab157d33f570a42f8b/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L4958) function.  This occurs before the corrected zone temperature is set [here](https://github.com/NREL/EnergyPlus/blob/9d6116a760ca6317e55196ab157d33f570a42f8b/src/EnergyPlus/ZoneTempPredictorCorrector.cc#L5049).  I'm not sure if there is a bug here and `SumSysMCpT` or `ZoneEnthalpyIn` should be updated based on the corrected zone temperature or if the zone temperature shouldn't be corrected?  I also could be missing something somewhere in the code.

mjwitte (approved):
This fixes the defect. Results now match the same file with cooling scheduled off.

Please merge in the latest develop branch and fix the new code. `LimitCapacity` and `LimitFlowRateAndCapacity` are now enums of type `LimitType::`. The new unit test needs some fixes also.

PR #8275, opened 2020-09-15
Update PVWatts using SAM Simulation Core library

nmerket:
Pull request overview
---------------------
 - Supercedes #7524 

This PR includes the [SAM Siimulation Core (SSC)](https://github.com/NREL/ssc) as a library in the `third_pary` directory using a git subtree of their latest release. It also connects the previously included PVWatts model to that library instead of using code copied from SSC into the EnergyPlus repo (PVWattsSSC.cc and PVWattsSSC.hh). This sets up a couple of useful things in the future:

1. Easier to maintain updates to PVWatts.
2. Lays the groundwork for [this NFP for including a lithium-ion battery model and controls](https://github.com/NREL/EnergyPlus/blob/Li_ion_battery_model/design/FY2020/NFP%20Li_ion%20battery%20model.md). The model is being implemented in SAM and we will make a equivalent model in EnergyPlus by calling the SSC library.
3. The ability to build in additional modeling capabilities into EnergyPlus that are already available in SAM. 

The goal is that this PR _should_ be zero diff.

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [x] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [x] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [x] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [x] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [x] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [x] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [x] If structural output changes, add to output rules file and add OutputChange label
 - [x] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
[...]

PR #8417, opened 2020-12-09
CondFD Surface Actuators

mitchute:
Pull request overview
---------------------
 - Add EMS actuators to the CondFD surfaces for thermal conductivity and specific heat
 - Add new API methods for querying CondFD data

### Pull Request Author

### Reviewer


mjwitte:
@mitchute This seems fine, but I'm concerned about energy conservation. One could create (or remove) energy in a given layer by simply changing the specific heat. 
I kind of get how that might be reasonable in a phase change material, and insulation is assumed to be lightweight so it might be negligible, but this could be limiting if someone wants to swap in a heavier layer. Perhaps a node temperature actuator would be good? Of course, that opens the door for even more shenanigans.

mitchute:
@mjwitte I believe that changing cp *won't* affect energy conservation. cp = dh/dT, so it will affect the path the solution takes but isn't actually a state variable. So, it won't have any effect until the solution is computed. If we were altering h or T directly, then yes, we would have an energy conservation problem.

mjwitte:
That would be fine if we track h at each node, but my understanding is that we track T. So, let's say the cp is low and a node is raised from T1 to T2. Then the cp is raised, it will release more energy as it falls from T2 to T1 than what went in. But maybe I'm wrong about only tracking T.

src/EnergyPlus/HeatBalFiniteDiffManager.cc
```diff
@@ -935,6 +952,25 @@ namespace HeatBalFiniteDiffManager {
...
+                                             SurfaceFD(SurfNum).condMaterialActuators(mat).isActuated,
+                                             SurfaceFD(SurfNum).condMaterialActuators(mat).actuatedValue);
+                EnergyPlus::SetupEMSActuator(state,
+                                             "CondFD Surface Material Layer",
+                                             SurfaceFD(SurfNum).specHeatMaterialActuators(mat).actuatorName,
+                                             "Specific Heat",
+                                             "[J/kg-C]",
+                                             SurfaceFD(SurfNum).specHeatMaterialActuators(mat).isActuated,
+                                             SurfaceFD(SurfNum).specHeatMaterialActuators(mat).actuatedValue);
+            }
```
Myoldmopar, line 972: :+1:

src/EnergyPlus/HeatBalFiniteDiffManager.cc
```diff
@@ -972,6 +1008,20 @@ namespace HeatBalFiniteDiffManager {
...
+                                    "Zone",
+                                    "State",
+                                    Surface(SurfNum).Name);
+                SetupOutputVariable(state,
+                                    format("CondFD Phase Change Node Specific Heat {}", Lay),
+                                    OutputProcessor::Unit::J_kgK,
+                                    SurfaceFD(SurfNum).specHeatNodeReport(Lay),
+                                    "Zone",
+                                    "State",
+                                    Surface(SurfNum).Name);
```
Myoldmopar, line 1024: :+1:

src/EnergyPlus/HeatBalFiniteDiffManager.cc
```diff
@@ -993,6 +1043,27 @@ namespace HeatBalFiniteDiffManager {
         } // End of the Surface Loop for Report Variable Setup
 
         ReportFiniteDiffInits(state); // Report the results from the Finite Diff Inits
+
+        if (!state.dataGlobal->eplusRunningViaAPI) {
+            state.dataPluginManager->condFDReadyFlag = true;
+        }
```
Myoldmopar: Is there a reason the other API ready flag didn't work for this?  Was it not ready in time?  Or was it ready too early?  I don't mind if we need another flag, but if we do then that's probably a clue we should plan for the possibility of different sections of the API having their own "ready" flag.

src/EnergyPlus/HeatBalFiniteDiffManager.cc
```diff
@@ -1530,6 +1601,8 @@ namespace HeatBalFiniteDiffManager {
                 int const MatLay(state.dataConstruction->Construct(ConstrNum).LayerPoint(Lay));
                 auto const &mat(state.dataMaterial->Material(MatLay));
                 auto const &matFD(MaterialFD(MatLay));
+                auto const &condActuator(SurfaceFD(Surf).condMaterialActuators(Lay));
+                auto const &specHeatActuator(SurfaceFD(Surf).specHeatMaterialActuators(Lay));
```
Myoldmopar, line 1602: Nice.

src/EnergyPlus/HeatBalFiniteDiffManager.hh
```diff
@@ -111,6 +111,16 @@ namespace HeatBalFiniteDiffManager {
...
 
+    struct MaterialActuatorData {
+        std::string actuatorName;
+        bool isActuated;
+        Real64 actuatedValue;
+
+        MaterialActuatorData () :
+            isActuated(false), actuatedValue(0.0)
+        {}
+    };
```
Myoldmopar, line 122: Reusing this for both Cp and K is great.

src/EnergyPlus/OutputProcessor.cc
```diff
@@ -5228,6 +5228,9 @@ namespace OutputProcessor {
         case OutputProcessor::Unit::clo:
             return "clo";
             break;
+        case OutputProcessor::Unit::W_mK:
+            return "W/m-K";
+            break;
```
Myoldmopar, line 5025: It would be great to get @mjwitte and @JasonGlazer to provide comment on this addition.

src/EnergyPlus/api/datatransfer.cc
```diff
@@ -141,6 +152,12 @@ void resetErrorFlag(EnergyPlusState state) {
     thisState->dataPluginManager->apiErrorFlag = false;
 }
 
+int getNumNodesInCondFDSurfaceLayer(EnergyPlusState state, const char* surfName, const char* matName) {
+    auto *thisState = reinterpret_cast<EnergyPlus::EnergyPlusData *>(state);
+    auto UCsurfName = EnergyPlus::UtilityRoutines::MakeUPPERCase(surfName);
+    auto UCmatName = EnergyPlus::UtilityRoutines::MakeUPPERCase(matName);
+    return EnergyPlus::HeatBalFiniteDiffManager::numNodesInMaterialLayer(*thisState, UCsurfName, UCmatName);
+}
```
Myoldmopar, line 152: Nice. Yes, a new API method to get the number of nodes in a specific material layer.

src/EnergyPlus/api/datatransfer.py
```diff
@@ -235,6 +251,31 @@ def reset_api_error_flag(self, state: c_void_p) -> None:
...
+            raise EnergyPlusException(
+                "`request_variable` expects `component_type` as a `str` or UTF-8 encoded `bytes`, not "
+                "'{}'".format(surface_name))
+        if isinstance(material_name, str):
+            material_name = material_name.encode('utf-8')
+        elif not isinstance(material_name, bytes):
+            raise EnergyPlusException(
+                "`request_variable` expects `component_type` as a `str` or UTF-8 encoded `bytes`, not "
+                "'{}'".format(material_name))
+        return self.api.getNumNodesInCondFDSurfaceLayer(state, surface_name, material_name)
```
Myoldmopar, line 262: All good here.

testfiles/CMakeLists.txt
```diff
@@ -417,6 +417,7 @@ ADD_SIMULATION_TEST(IDF_FILE PurchAirWithDaylightingAndShadeControl.idf EPW_FILE
 ADD_SIMULATION_TEST(IDF_FILE PurchAirWithDaylightingAngleFac.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
 ADD_SIMULATION_TEST(IDF_FILE PurchAirWithDoubleFacadeDaylighting.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)
 if (LINK_WITH_PYTHON)
+ADD_SIMULATION_TEST(IDF_FILE PythonPlugin1ZoneUncontrolledCondFD.idf EPW_FILE USA_CO_Golden-NREL.724666_TMY3.epw)
```
Myoldmopar, line 420: An example file with the usage, good.

testfiles/PythonPlugin1ZoneUncontrolledCondFD.py
```diff
@@ -0,0 +1,127 @@
...
+        for surface in self.surfaces:
+            for layer in surface.material_layers:
+                if layer.name is "WallStudAndCavityk1":
+                    if actuate_walls:
+                        k = 10
+                        cp = 1500
+                    else:
+                        k = 2.4
+                        cp = 1179
+                    self.api.exchange.set_actuator_value(state, layer.k_handle, k)
```
Myoldmopar, line 124: Nice.

Myoldmopar (commented):
I don't have any major issues with this.  With respect to the enthalpy issue previously discussed, this particular feature doesn't actually affect the ability of a user to alter Cp.  They already do this with _every_ phase change material simulation by entering values of T/h and letting EnergyPlus update the Cp value at each calculation.  This current change simply adds an API function to do it outside of the IDF.  If a deep discussion needs to happen regarding that calculation, it can be made outside of the context of this PR.  

One question I had was about the ready flag you added.  Can you clarify why that was needed?  If it is needed, then so be it, but we will need to think about how this looks as we add more to the API.

src/EnergyPlus/HeatBalFiniteDiffManager.cc
mitchute: I think I was seeing that the exiting flags weren't ready in time, but I'll go back and double-check just to be sure.

src/EnergyPlus/OutputProcessor.cc
mjwitte: I think this should have a corresponding `else if` in `unitStringToEnum`. @JasonGlazer ?

mitchute: Thanks, @mjwitte. Yeah, that looks right. I'll fix it up.

src/EnergyPlus/HeatBalFiniteDiffManager.cc
mitchute: @Myoldmopar following up on this. I've confirmed that the extra flag is not needed. The most recent commit removes it: https://github.com/NREL/EnergyPlus/pull/8417/commits/c1aa6928af1899cf040051f0fa9b08b518625df2

mitchute:
@Myoldmopar FYI, Windows failed to build packages here

```
EXEC : CPack error : Cannot find QtIFW compiler "binarycreator": likely it is not installed, or not in your PATH [D:\a\EnergyPlus\EnergyPlus\build\package.vcxproj]
```

See here: https://github.com/NREL/EnergyPlus/runs/1618821650?check_suite_focus=true#step:9:1895

Myoldmopar:
Hmmm, I would've thought this was previously tested and fleshed out, but something seems up for sure.  Let me look into it a bit.

mitchute:
@Myoldmopar ready for a final review here. I'll open a new PR for further changes.

Myoldmopar:
Building this now locally as it is otherwise ready to go.  Both this and #8275 should go in today.

Myoldmopar:
Pulled develop in, ran all the tests both serially and parallel, all good.  Merging.

PR #8423, opened 2020-12-14
ZoneCompTurnFansOff/On Availability Signal not working for VRFTU Fan

mitchute:
Pull request overview
---------------------
 - Fixes ```VRFTest_TU_NoLoad_OAMassFlowRateTest```
 - A bug was introduced [here](https://github.com/NREL/EnergyPlus/blame/9efd1c570b4bfc6993478ba51dfd9b6c8bdef0e6/src/EnergyPlus/HVACVariableRefrigerantFlow.cc#L8990) where the ```ZoneCompTurnFansOn``` availability signal was mistakenly replaced with ```ZoneCompTurnFansOff``` 🤦 . This may cause the OFF signal to actually turn ON the VRFTU fan and it prevents the ON signal from activating the fan (e.g. during night cycle operation).
 - Diffs in VariableRefrigerantFlow_5Zone because it has a night-cycle availability manager attached to a VRF terminal unit.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
+++ b/src/EnergyPlus/HVACVariableRefrigerantFlow.cc
@@ -9426,7 +9426,7 @@ namespace HVACVariableRefrigerantFlow {
 
         // if the terminal unit and fan are scheduled on then set flow rate
         if (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).SchedPtr) > 0.0 &&
-            (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).FanAvailSchedPtr) > 0.0 || DataHVACGlobals::ZoneCompTurnFansOff) &&
+            (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).FanAvailSchedPtr) > 0.0 || DataHVACGlobals::ZoneCompTurnFansOn) &&
             !DataHVACGlobals::ZoneCompTurnFansOff) {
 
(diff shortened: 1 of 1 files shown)


src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -9426,7 +9426,7 @@ namespace HVACVariableRefrigerantFlow {
 
         // if the terminal unit and fan are scheduled on then set flow rate
         if (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).SchedPtr) > 0.0 &&
-            (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).FanAvailSchedPtr) > 0.0 || DataHVACGlobals::ZoneCompTurnFansOff) &&
+            (GetCurrentScheduleValue(state, VRFTU(VRFTUNum).FanAvailSchedPtr) > 0.0 || DataHVACGlobals::ZoneCompTurnFansOn) &&
```
mjwitte, line 9429: Interesting. So, following the blame back, this was broken back in January. So, there's the potential that this introduced a bug where an availability signal of `ZoneCompTurnFansOff` would actually turn on the VRFTU fan. A rare case, for sure, and clearly not in our regression suite, but we should probably rename the title and describe the potential problem in the PR description just in case someone runs across it in v9.3 or 9.4.

mitchute: Title and description are updated.

mjwitte (approved):

PR #8431, opened 2020-12-18
Rethinking the Global State Instance

Myoldmopar:
OK, so per offline discussion, there were a couple issues that needed resolution surrounding the "state" business.

One thing was determined to be a non-issue.  That is the fact that we do need to continue to collect the global variables of EnergyPlus and bring them into a manageable state.  This will allow us to manage the state of the program so that EnergyPlus can be deployed in new workflows where state reset, warmup, persistence will all be important.  This will also make it so that EnergyPlus is thread-safe, so that clients of the API can spin off multiple simulations in parallel without having to worry about putting them in a completely separate process space.

As a status update, we have gone from ~15k global variables down to about 5k, with a significant amount of that occurring in late November and early December after team members expressed concern about the ongoing nature of that task.  Once this current PR is resolved, I anticipate ramping back up on that effort and getting the remaining global variables cleaned up very soon.  One side benefit of this global refactor is that we are simultaneously converting list of int const values into enumerations and cleaned up global strings into constexpr literals, both of which have some benefit on runtime, and I'd say program complexity.

The core issue was about how the state of a simulation is stored and passed.  With current EnergyPlus, when you execute EnergyPlus.exe, it will create a state instance, and pass a reference to that state everywhere through the code.  If you call to run EnergyPlus via API, you request a new instance be created, a pointer to that instance is returned, and then the client passes this state reference in and out of the program.  Multiple states could be created and executed, however, they can't be run parallel yet because of the remaining global variables in the codebase.

It was determined that we should change the "passing state around" and instead hold a global variable of states and access them inside each function.  The vector of states is to allow EnergyPlus to run in multiple threads and own the state instances internally.  A thread_local variable will then be created which will be the index into the vector for the currently running thread. 
 Basically create this in EnergyPlus:

```
std::vector<EnergyPlusData*> stateVector;
unsigned int thread_local currentStateIndex = -1;
EnergyPlusData & getCurrentState() {
    return *stateVector[currentStateIndex];
}
```

or similar, and then go from

```
void GetTESCoilInput(EnergyPlusData &state) {
```

to 

```
void GetTESCoilInput() {
    EnergyPlusData & state = getCurrentState();
```

in every function that needs access to the state of the program, which is nearly every function in the program.

This PR achieves this.  Every function call that had state passed in was converted to the new approach.  At the top of each function, we now have a line requesting the currently running state.  Every function call made into the functions was modified to no longer pass a state instance.  The unit test code was cleaned up to avoid passing state around, however, an instance of state is still set up by the unit test fixture and unit test main function so that each individual unit test does not need to be otherwise modified.  When calling EnergyPlus as an API client, the client still needs to know which currently running state instance to use, in the case of multiple runs in serial within one thread, or multiple runs in parallel in different threads.  Thus, instead of passing back a reference to a state, the program passes back an index into that global vector variables.  The client code doesn't really change much since it is still receiving some form of reference to an object that must be passed back in.  Underneath the typedef, the object is now an unsigned int instead of a pointer.

There are definitely some open questions about this, primarily about thread safety with this global vector of states, and runtime performance, but this code currently does compile and run, so it's a good time to have a discussion.  At this point, I'd like to open the floor to comments.


src/EnergyPlus/AirLoopHVACDOAS.cc
```diff
@@ -102,25 +103,25 @@ namespace AirLoopHVACDOAS {
...
 
-        this->CalcAirLoopDOAS(state, FirstHVACIteration);
+        this->CalcAirLoopDOAS(FirstHVACIteration);
     }
 
-    AirLoopMixer *AirLoopMixer::factory(EnergyPlusData &state, int object_num, std::string const &objectName)
+    AirLoopMixer *AirLoopMixer::factory(int object_num, std::string const &objectName)
     {
-
+        EnergyPlusData & state = getCurrentState();
```
lefticus, line 122: *I have strong concerns about these changes (picking this particular line-entry because it is the first one in the diff set, not because of this particular change).*

# Introduction

At the high level, in a TL;DR fashion:

 * It is nearly impossible to get the interaction between `static` and `thread_local` variables correct  (`stateVector`, the underlying data store is implicitly `static`). They have different lifetimes in the C++ standard. Aside: I teach an entire day class specifically on understanding these issues (which at least two people on the EnergyPlus team have taken).
 * There may be a very difficult to measure impact on performance in a multithreaded environment due to false sharing of state objects between threads. (I did a recent [C++ Weekly](https://youtu.be/5VrX_EXYIaM) episode around this topic.) Note that I do not believe the techniques shown in that video would help in this case.
 * We do not know when and how to destroy the heap-allocated `State` objects
 * Use of EnergyPlus as a library on Windows will result in issues due to differences in how globals are resolved at runtime. (issues we have already had to address multiple times in both EnergyPlus and OpenStudio)
 * It breaks the thread safety that we were on the verge of attaining, and adding back in thread-safety would have a significant impact on performance
 * It breaks fundamental compiler optimizations related to interprocedural optimization, value lookup, and function inlining.

📝 **Note** The passing-parameter version we were working towards already has none of the above issues, performs better, and brings thread-safety with it *for free*. It just has the minor inconvenience of an extra parameter.

## High Level Performance Overview 

When attempting to measure just the impact of accessing a local value, we can see with this [quick-bench test](https://quick-bench.com/q/R3K-rs4zihfc-FMGGi7GN1IJdu4) test

image

a considerable impact in the cost of accessing an object, based on the various options.

This test takes the cost of accessing an object out of context with the rest of the system, as all micro benchmarks do.

# Performance Investigation

However, it is easy to see with some compiler-explorer links where these differences in performance will come from:

## Investigation of Thread-Safe Version

Let's take this version that has `mutex` locks for thread safety:

```cpp
State *get_current_thread_safe() {
  const std::lock_guard<std::mutex> lock(global_mutex);
  return states[tl_current_state];
}

int do_1st_op_thread_safe() {
  return ++(*get_current_thread_safe()).some_value;
}
int do_2nd_op_thread_safe() {
  return ((*get_current_thread_safe()).some_value *= 2);
}

int do_ops() {
  do_1st_op_thread_safe();
  return do_2nd_op_thread_safe();
}
```

The compiler has no way to optimize out the two different `mutex` locks (it is not allowed to). So, while the code in `do_ops` does get inlined and merged, even at full optimizations, we end up with: 

[compiler-explorer link](https://godbolt.org/z/joPa36)

```asm
do_ops():                             # @do_ops()
        push    r14
        push    rbx
        push    rax
        mov     r14d, offset __pthread_key_create
        test    r14, r14
        je      .LBB4_2
        mov     edi, offset global_mutex
        call    pthread_mutex_lock
        test    eax, eax
        jne     .LBB4_9
.LBB4_2:
        movsxd  rcx, dword ptr fs:[tl_current_state@TPOFF]
        mov     rax, qword ptr [rip + states]
        mov     rbx, qword ptr [rax + 8*rcx]
        test    r14, r14
        je      .LBB4_5
        mov     edi, offset global_mutex
        call    pthread_mutex_unlock
        add     dword ptr [rbx], 1
        mov     edi, offset global_mutex
        call    pthread_mutex_lock
        test    eax, eax
        jne     .LBB4_9
        mov     rax, qword ptr [rip + states]
        jmp     .LBB4_6
.LBB4_5:
        add     dword ptr [rbx], 1
.LBB4_6:
        movsxd  rcx, dword ptr fs:[tl_current_state@TPOFF]
        mov     rbx, qword ptr [rax + 8*rcx]
        mov     eax, offset __pthread_key_create
        test    rax, rax
        je      .LBB4_8
        mov     edi, offset global_mutex
        call    pthread_mutex_unlock
.LBB4_8:
        mov     eax, dword ptr [rbx]
        add     eax, eax
        mov     dword ptr [rbx], eax
        add     rsp, 8
        pop     rbx
        pop     r14
        ret
.LBB4_9:
        mov     edi, eax
        call    std::__throw_system_error(int)
```



A lot is going on here, and we could walk through it step by step, but for now, let's pretend like we have made it thread-safe without using a mutex (we cannot, but let us pretend).

## Investigation of Thread-Unsafe Version

```cpp
State *get_current_thread_unsafe() { return states[tl_current_state]; }

int do_1st_op_thread_unsafe() {
  return ++(*get_current_thread_unsafe()).some_value;
}
int do_2nd_op_thread_unsafe() {
  return ((*get_current_thread_unsafe()).some_value *= 2);
}

int do_ops() {
  do_1st_op_thread_unsafe();
  return do_2nd_op_thread_unsafe();
}
```

We get assembly generated (`-O3`)

[compiler-explorer link](https://godbolt.org/z/f1h5ox)

```asm
do_ops():                             # @do_ops()
        movsxd  rax, dword ptr fs:[tl_current_state@TPOFF]
        mov     rcx, qword ptr [rip + states]
        mov     rax, qword ptr [rcx + 8*rax]
        add     dword ptr [rax], 1
        movsxd  rax, dword ptr fs:[tl_current_state@TPOFF]
        mov     rcx, qword ptr [rcx + 8*rax]
        mov     eax, dword ptr [rcx]
        add     eax, eax
        mov     dword ptr [rcx], eax
        ret
```

Note that it still has to do a lookup each time, even though it has merged the two functions! Why? Because the compiler cannot optimize the lookup of the thread-local variable. Another function in another translation unit might have access to that value and modify it in some way, for example.

## Investigation of Parameter-Passing Version

Let's now take this example back to something that is closer to the direction we had been moving with the state refactor up to this point:

```cpp
int do_1st_op_with_param(State &s) { 
    return ++s.some_value;
}
int do_2nd_op_with_param(State &s) { return (s.some_value *= 2); }

int do_ops(State &s) {
    do_1st_op_with_param(s);
    return do_2nd_op_with_param(s);
}
```

[compiler-explorer link](https://godbolt.org/z/bn5nxo)

```asm
do_ops(State&):                       # @do_ops(State&)
        mov     eax, dword ptr [rdi]
        add     eax, eax
        add     eax, 2
        mov     dword ptr [rdi], eax
        ret
```

In this version, the compiler can take advantage of calling conventions to know that the value being manipulated is the one that's pointed to by the first parameter, which is passed as a memory location in `edi` (aka `rdi`). It can dereference that value once (the first line), double the value, then add 2, and store the result back in the first parameter reference.

# Conclusion

The performance issues outlined cannot be mitigated due to what is required by the standard. Meanwhile, the parameter-passing currently in-progress works with the ABI and calling-conventions to produce the most optimizable code possible.

amirroth: @lefticus, I am the "brains" (if you will) behind this transition. We haven't met yet and that feels like an oversight given the prominent role you play in EnergyPlus and OpenStudio development. If you would like to schedule a zoom for after the holidays, I would like that. Just to give you a bit of context about where I am coming from. Other than C++ itself (certainly C++11 and beyond), I have many of the same computer-science/multi-threading/teaching chops that you appear to have. Edwin and others will tell you that I do not pretend to know things I that I don't and that I quickly incorporate new information and update my thinking if shown a better way. For now, I feel that I am firmly within my forest. Anyways, here is a short rebuttal.

**Re: performance.** Performance microbenchmarks are not entirely meaningless, but they are pretty close. I understand that you are orienting us towards the performance singularity here, but we are so far away from that in so many ways that this particular bit makes less than 1% difference right now and will continue to make less than 1% difference for the foreseeable future.

**Re: thread-safety.** This is far more important and I have to admit I do not understand where your concerns are coming from, i.e., what makes a design with externally managed state memory inherently more thread-safe than one with internally managed state memory that is externally referenced by handles. I actually do not see the handles discussed in this PR and so maybe that is what is worrying you (e.g., you are under the impression that EnergyPlus is responsible for creating, destroying, and toggling between states automatically). But if that is not the case, and there is actually a fundamental difference, I would like to understand what it is.

amirroth: @lefticus, can you explain why the lookup `state_vector[tl_current_state]` requires locking for thread safety? Maybe I don't understand what `thread_local` really implies. My understanding it is a thread-specific variable that will retain the last value that was set within the that thread. For this specific example, if another thread writes `tl_current_state`, that will not effect the value in the current thread. Is that not correct? If not, then what is the point of `thread_local`? it is correct, then this is exactly the way in which I intend to use `tl_current_state` and I do not see what locking achieves. I am sorry, I am mentally blocked on this. 

[Update: I wrote a short program with 64 threads in which each thread writes a unique `tl_current_state` and then saves that value in a vector indexed by the same. This part of the code is mutex protected. Each thread then runs 100000 iterations in which it temporarily modified `tl_current_state` and checked to see that the value in the vector still matched the position. This part of the code is not mutex protected, yet I was unable to reproduce a "thread unsafe" behavior in which a `tl_current_state` modification by one thread is observed by another thread.] 

[Update: you may not have gotten this yesterday because I initially misspelled your handle.]

amirroth: @lefticus, I also want to understand what the Windows-specific issue is. Most of my programming experience is on Mac/linux with g++/clang/llvm. In what way does Windows resolve global and/or thread-local variables differently than Mac/linux do?

lefticus: @amirroth

I admit I don't fully understand what the motivation is for these changes, but with regards to your specific points:

## Performance Questions

While it may seem like just an academic micro benchmark in the examples I showed, *we are already seeing at least a 3% performance loss just in this PR on the performance dashboards!*

We should see more performance improvements possible as we continue to remove globals and the compiler is able to perform more optimizations, and we should see even more optimizations possible with LTO enabled.

For a project that has performance as a primary design goal, why would we choose to take a 3% hit?

In my experience with optimizing C++ projects, these kind of apparently small, meaningless changes compound over time. For example, in my scripting engine I ended up getting over a [100x performance improvement](https://youtu.be/lNnBExDoNSQ) by slowly attacking each of these small things and aggressively making sure I didn't accept any regressions without a clear usability improvement.

One particular way that performing this lookup of the global via a vector inside of each function is going to hurt us is in the specific case of I-cache pressure. While the lookup may seem small, it is taking away some of the cache that could be used for other instructions and will ultimately result in more cache misses, which are very expensive.

It's reminiscent of how we would academically say that inserting into the middle of an array is more expensive than inserting into the middle of a linked-list, and therefor if we need to insert into the middle of a container, we "know" we should use a linked-list. However, on modern hardware it's actually [far more likely to be cheaper](https://www.youtube.com/watch?v=YQs6IC-vgmo) to insert into an array (aka C++ vector). 

## Thread Safety and Standard Containers

You are correct that the `tl_current_state` is safe to read from / write to via multiple threads, as that is the point of a `thread_local`.

However, if the `state_vector` is a global, `static`, non-`thread_local`, as seems to be the intention, then `state_vector` reads and writes *must* be protected via a mutex.

[It is safe only to access `const` members](http://eel.is/c++draft/res.on.data.races) of an `std::vector` via multiple threads without a lock. As soon as you add in the possibility that you might grow, shrink or mutate the `state_vector` in any way, then you must protect all accesses via mutexes or you will have a race condition on reading and resizing of the vector. (this is the case for all standard library containers in C++).

Here is a simple illustrative example: https://godbolt.org/z/We6sra

Without locks and having simultaneous readers and writers you have [undefined behavior](http://eel.is/c++draft/intro.multithread#intro.races-21) in the project and the [entire code base is invalid](http://eel.is/c++draft/intro.abstract#5).

## Thread Safety With Function Parameters vs Globals

With the proposed redesign we have 2 primary issues:

 * As mentioned above the holder of the state objects itself needs to be thread safe
 * It's easy for two different threads to set the same `tl_current_state` index value, and perform operations on the same `State` object in two different threads, necessitating that the `State` object itself become thread safe.

Because it's impossible for two different threads to see each other's `thread_local` values, there is no way to enforce that two threads not use the same index.

With parameter passing we know:

```cpp
// inside a thread
State some_state;
do_simulation(some_state);
```

The programmer has to go out of their way to choose to share `State` objects between threads, accidental sharing of state is not possible.

Because the state exists on the thread's local call stack, we don't have to worry about thread safety.

## Spooky Action At A Distance With Global State and Loggers

Loggers, while generally considered to be a safe thing to keep as global objects by most people, are currently the cause of a long standing bug in OpenStudio. 

Currently this test fails: https://github.com/NREL/OpenStudio/blob/develop/src/energyplus/Test/ForwardTranslator_GTest.cpp#L611

This is due to multiple threads all using a global logger object simultaneously. One thread performs an operation then tries to read back the number of warnings/errors that were generated, while another thread is also performing a similar operation. The first thread then reads back the value from the second thread. Because these two threads are executing on objects with the same type there is no way in the current architecture of OpenStudio's logger to actually solve this problem.

If, however, each thread had its own logger object there would be no mixing of data and no problem to fix.

## Static Objects in DLL's on Windows

On MacOS and Linux any static that is loaded in a common shared library is automatically shared between all other shared libraries that access the common library.

On Windows, each library creates its own copy of the static objects that are visible in a header. This can be addressed either by re-initializing the variable for each accessing library, or moving the global into a function local static that exists in exactly one .cpp file. 

This has come up very recently with OpenStudio having multiple views of SQLite's gobal state and our (working with @kbenne ) having to make sure we moved all accesses of SQLite into .cpp files.

## Static Initialization Order Fiasco (and related lifetime problems)

Because there is no way in C++ to guarantee the order of construction of static objects (and therefor the order of destruction also) then there is no way to ensure that the state vector is constructed before any other static object might have use for it.

Simultaneously, we risk that the `std::vector` that holds the state objects might be destroyed before another library is done using it. (A similar issue took me over 2 weeks to debug in a previous project).

So the only way to make this vector safe from the static initialization order issues is to make it a so-called "Meyer's Singleton" (named after Scott Meyers), which utilizes the Construct on First Use Idiom.

```cpp
std::vector<State *> &get_states() {
  static std::vector<State *> states; // constructed on first call to this function
  return states;
}
```

Unfortunately, a function-local static incurs is own intrinsic cost, because the compiler [must insert a thread safe initialization routine into the function itself](https://www.youtube.com/watch?v=B3WWsKFePiM).

And then providing thread-safe access to the data contained in the function local static becomes considerably more difficult, as any mixed read/writing access will still require a mutex (see above) but we have to figure out how to make sure the mutex itself is constructed before and destroyed after the global data we need it to protect! (I have also spent ~1 week chasing this issue on previous projects). 

Keep in mind, that due to the differences in how Windows deals with statics across DLL's, these operations have to be defined in the `.cpp` files, often preventing their inlining and further affecting performance.

## TL;DR

In short - I have spent many months of my life finding (and usually fixing) bugs related to the shared state between global variables and object lifetime issues related to global variables in C++. Many of these issues are intractable due to the nature of the language and our linker/compiler tool chains. This experience led to the creation of the class I teach that I mentioned earlier.

I'm trying to make sure we don't continue to make these same mistakes while simultaneously giving a path forward for better performance.

amirroth: @lefticus I appreciate the detailed response, including examples. Thanks also for the info on Windows specific and lifetime issues. I admit I have never run into these before. Anyways, I think we will get to the end of this "faster" if we decompose the discussion into two pieces, thread safety and performance.

### Performance

Performance is quicker so let's knock it out first. I admit that a 3% slowdown is more than I was expecting. I was expecting 1% or less. I don't know how this breaks down in terms of Icache vs. straight up instruction count or Dcache/load count, but it doesn't really matter. I will also admit that I didn't intuit the fact that keeping `& state` as the first argument allows you to keep it in `%rdi` across function calls, making passing it around essentially free. Personally, I think 3% says more about the granularity of functions in EnergyPlus as opposed to the intrinsic overhead of this approach, but at the same time EnergyPlus is big and its function structure is not likely to change anytime soon. I relent on this point. 

### Thread Safety

It will take me longer to relent on the thread-safety point because I still don't understand some of the issues and because you may not see the overall design. So let me explain the design very briefly. You're correct, it consists of a `static std::vector<EnergyPlusData state> state_vector` and an `int thread_local tl_state_handle`. The EnergyPlus API supports a function called (I think) `CreateRunnableState` which allocates a new entry in `state_vector`, sets `tl_state_handle` to the index of the new entry, and also returns this handle to the client. The client will use this handle in future calls. Two more facts about this function. First, the state entry saves the id of the thread that created it and `CreateRunnableState` allows each thread to create only one runnable state. Second, the function is mutex protected, because (as you suggested) the allocation of new state entries and handles needs to be serialized. However, because state and handle creation/modification is encapsulated in this function, references to the `state_vector` via existing handles do not need to be serialized. Third, there is a maximum limit on the number of simultaneously active EnergyPlus threads and `state_vector` is pre-allocated to this number and not resized (this is not in the current PR design but it was always a fallback and now that you've mentioned the issue it can be). 

[Side note: the plan is also to have a second vector of non-runnable state objects that could be used as checkpoints or for other purposes.]

[Another side note: threads created within EnergyPlus (e.g., if/when we end up multi-threading EnergyPlus itself do not allocate new state objects or handles, rather they inherit the state and handle of the parent thread via the thread function.]

This design keeps state and its structure internal to EnergyPlus and allows clients to interact with them via handles and an API (calls through the API can check that the given handle belongs to the calling thread). Threading aside, I believe that there is a certain measure of safety and aesthetic in not giving the client a direct pointer to state. Given the thread-safe nature of object/handle allocation and the protection at API boundaries, I also do not believe it is inherently less thread safe than punting state memory management to the client (including putting it on the client's stack). 

### Windows Shenanigans

Unless I am missing something, I do not believe that the way Windows deals with statics would break this design. Except for ...

### Meyer's Fiasco

I have fortunately never encountered this particular noise. I've picked up everything you've put down here and if this is a "show-stopper" then so be it. In an homage to you, I will end with this ...

### TL;DR (not really apt, but this is an homage, so ...)

Are the problems with static variables so fundamental that they are simply to be avoided at all costs? Or is there a workable hybrid design in which EnergyPlus manages states internally and clients reference handles, but API functions recover the state pointer, which is then passed around as a parameter within EnergyPlus itself. This design combines EnergyPlus control of state with a greatly reduced number of references to the `state_vector` itself (only at API calls, not in every function). This greatly reduced number of calls may enable a more robust solution to the static variable problem, e.g., Meyer's.

Thanks again for your patience with this line of questioning and for your past and continuing contributions to EnergyPlus, OpenStudio, and other DOE projects.

lefticus: ## Multiple Readers / Writers

If you pre-allocate the maximum number of threads that are allowed to run at a given time, then you *might* be able to get away with avoiding locks for readers, because it would avoid a change in the size of the allocated storage.

Why *might*? Because some platforms can only write data in specific units at a time. And therefor require a read and a write back, or a write over two operations. So it's possible that a write to location 1 would also require a write to location 0 at the same time, which could cause a race. It's also possible that a write to location 1 might take 2 (or more) operations, resulting in only a half read of the new data.

This might be solvable if it's a fixed-sized array of `std::atomic<>` id's?

## API Surprises

Do you have any way of alerting the user that if they call `CreateRunnableState` two times in the same thread they are actually getting back the same object? I imagine this will create a large number of support requests.

## Intractability of Static Lifetime Issues

An entire new keyword was added to the language in C++20 just to try and fix some of the static lifetime issues (`constinit`). And we wouldn't even be able to use it in this case. This has been a problem since C++ was created. I strongly recommend avoid 100% of all mutable global state because of these issues.

This is not specific to Windows, BTW, but Windows is slightly different than Linux (and MacOS has its own issues) so it's the one that comes up more often.

I've been involved with OpenStudio for over 11 years now. Early on we had many issues with static lifetimes of shared objects (notably the LoggerSingletons and the regular expressions used for parsing). But further leading into globals used by Qt, Ruby, and SQLite. It's an average of at least once a year that I am tasked with solving some newly discovered issue that comes down to global state of something.

At this point I would say I'm almost uniquely qualified to know how to implement what you want in this code specifically in a way that would *appear* to work

 * from inside a python call
 * on all the major platforms
 * via a shared library

While simultaneously knowing how to break it, with no known fixes, and hurting performance.

## Concerns Not Yet Mentioned Related To Statics

With this API, how do we when to destroy a state object? If only the ID is a thread_local, we have no way of knowing when a thread has shut down and we need to clean up its data object.

## Re your TL;DR:

> Or is there a workable hybrid design in which EnergyPlus manages states internally and clients reference handles, but API functions recover the state pointer, which is then passed around as a parameter within EnergyPlus itself. This design combines EnergyPlus control of state with a greatly reduced number of references to the state_vector itself (only at API calls, not in every function).

Sure, that might work, but why? All this accomplishes is limiting how the user of EnergyPlus can create and manage the simulation.

Unfortunately I'm still missing what the end goal of this PR is.

With the current `develop` branch you

 * create a state object
 * pass it into a function you want to call

With the proposed code:

 * You must call an `init` function of some sort
 * call the EnergyPlus function you want (and pass a handle? how is a handle different than an object reference?)
 * call a shutdown function when done (presumably)

With the current `develop` the API is "impossible" to use wrong. The API requires you to pass in a valid state object.

With the proposed API, if you forget to init the system, you get a crash because your function actually still takes an extra parameter, we just never communicated that to the user.

## Safety?

You mention safety as a concern when providing the user with pointers and passing those around, but the current `develop` does not operate in this way.

https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/main.cc#L57-L59

```cpp
int main() {
  EnergyPlus::EnergyPlusData state; // created and initialized automatically with a constructor
  
  ProcessArgs(state, argc, argv); // guaranteed to still be valid
  return EnergyPlusPgm(state); // guaranteed to still be valid  
} // state destroyed when main exits
```

The amazing bonus to the current code is that you get all of the single stepping / rolling back / whatever you want *for free* and the user can do anything they want.

```cpp
int main() {
  EnergyPlus::EnergyPlusData state; // created and initialized automatically with a constructor
  
  ProcessArgs(state, argc, argv); // guaranteed to still be valid
  
  // initialize with beginning state
  std::vector<EnergyPlus::EnergyPlusData> complete_history{state};
  
  while (timeStepsRemaining(state)) {
    DoTimeStep(state);
    complete_history.push_back(state); // copy current state!
  
    if (somethingWeirdOrUserInput()) {
      state = complete_history[somePreviousIndex]; // rewind state
    }
  }
  
  return EnergyPlusPgm(state);
}
```

The current API is perfectly set up to be able to do all of the interesting simulation things with complete safety.

## An Aside Regarding Trying to "automatically manage state lifetimes for the user"

I took a similar approach in my scripting engine for objects that live inside of the scripting context. In this case, and with that particular design of the scripting engine, it was a necessity, not just desirable.

It took approximately 6 years to find and resolve all of the bugs related to internal object lifetime management.

lefticus: @amirroth 

Upon further reflection and consideration of the concerns and motivations for the move back to global mutable state, I believe I have a solution that retains the performance, correctness, safety, and flexibility of the current work on `develop`.

It seems your concern lies mainly around a few public entry points into EnergyPlus, since you are willing to keep passing references around internally.

**EDIT** just before posting this I now wonder if your primary concern is actually with the C API functions, since those match the description you are talking about (re creating objects, management of lifetime, passing of pointers, all of which have no application in the C++ API).

If you want to give the users of EnergyPlus (who want to choose this option) the "automagic" feel of having the state maintained for them, as it currently is, then we can provide functions like this.

```cpp
namespace energyplus::threadlocal {
  EnergyPlusData &get_thread_local_state() {
    thread_local EnergyPlusData state;
    return state;
  }

  int EnergyPlusPgm() {
    return energyplus::EnergyPlusPgm(get_thread_local_state());
  }

  void ProcessArgs(const int argc, const char *argv[]) {
    energyplus::ProcessArgs(get_thread_local_state(), argc, argv);
  }

  // add as many others as we choose to add
}
```

Then for those users who choose this path:

```cpp
int main() {
  energyplus::threadlocal::ProcessArgs(argc, argv); // thread local is created on first call to get_thread_local_state
  return energyplus::threadlocal::EnergyPlusPgm(); // thread local is still valid and the same object
}
// thread local is destroyed when "main" thread is destroyed (just before the program exits and
// regular statics are destroyed
```


I have no concerns with an implementation like this, and I believe it addresses your concerns. I also cannot think of any way to break this after giving it a fair bit of thought.

* This can be easily extended to include other functions like psychometrics, etc.
* Maintenance will be easy: if the API changes, the code breaks.
* This is particularly helpful in the case of the C API (*which I just now realize seems to be the API you were concerned about? Since that is the one that actually passes around points instead of dealing with clean C++ objects?*)

lefticus: In case there is any confusion, I'm suggesting that the C++ API stay as it is on develop. We only wrap a few functions that provide an automatic thread_local version of the State object.

But my main concern is that we understand where the concern actually lies, as the C++ API does not have any of the issues brought up in the initial posts, only the C wrapper API does.

amirroth: @lefticus, thanks for the detailed replies. I don't know whether my problem is with the C++ API or the C API (I think it is with both). At this point, it may be simpler if I just state my motivation for the proposed alternative design rather than try to iterate over somewhat straw-man arguments.

Just as background, my active programming career lasted from 1992 to about 2010. I am not one of the EnergyPlus developers, and I would not even really call myself the project manager (@myoldmopar is that). I am a manager in the bureaucratic sense. However, I take a close interest in the internal details of EnergyPlus because it is the most important project in my portfolio (by a good bit), because I can (I know how to program), and because I can (I figuratively sign the checks so I can do whatever I want). I track the repo and various branches of interest and I make local modifications and try things out on my MacBook. Fun times!

Over the past X months, I have noticed the addition of the state argument to each and every function in EnergyPlus, including logically pure functions where the state argument is needed only for the purpose of printing error messages. This immediately gave me PTSD back to my undergrad functional programming course where state had to be passed as an argument and new state returned as a value! I asked why this was happening and was told that it was required to "make EnergyPlus thread-safe". Now, I have written a number of multi-threaded programs (although all of the significant ones have been in C, not C++, all have been for Mac/Linux, and now that I think of it, none of them have been used as a library to the best of my knowledge), all of them had global state and a few had detailed highly optimized internal memory management. That a program can only be "thread safe" if it has no global state seemed (and honestly still seems) ridiculous. 

Then there was the matter of externally exposed and managed state. This is something we've generally avoided with EnergyPlus in the past. I'm sure you're familiar with the PythonPlugin nee EMS feature. That API doesn't hand over pointers to internal EnergyPlus variables, in that API we use handles for safety. And now we're going to hand over the entire state object? Really? You gave checkpoint/recovery as an example of an "amazing" feature that the user could implement. Great. For the record, an internal array of non-runnable states also supports that. Other than this particular example (and serializing/deserializing to/from files) what else would a user want to do with the state object that would be considered "safe" and "good" and wouldn't generate umpteen mystery EnergyPlus crashes and subsequent UnmetHours posts? Direct access to EnergyPlus state is what bothers me here, memory management of that state is only part of that, this is why I think my problem is with both the C++ and C APIs.

Anyways, the combination of these two pieces of perceived nuttiness got me thinking about an alternative solution. One in which EnergyPlus state is kept internal with clients receiving handles, in which state operations are controlled via an API, and in which a state pointer does not need to be threaded through each and every one of EnergyPlus' roughly nine million functions, but can be quickly recovered in functions that actually need it. That is the genesis of the alternative design.

### TL;DR

Through this discussion I have come to understand/appreciate two things. First, passing a state reference around to every function may be distasteful (although not as distasteful as `auto`) but it ultimately is faster than doing even an unlocked `state_vector` lookup at the beginning of every function. And by 3X more than I anticipated. I am not going to argue with numbers or assembly code (as I said, I relent on this), although I still feel that it would be good to find an alternative error-reporting solution for logically stateless functions.

Second, maybe this is a C++ thing or a Windows thing or a library thing or a combination, but apparently there are issues with global variables than I have not encountered personally or even heard of. Again, my active programming days are behind me and although I have experience with multi-threaded programming, it seems that having done most of it in C and little of it in the form of libraries has insulated me from this mishegas. Some of the issues you've described make sense to me (although most of these seem specific to the program rather than fundamental to the language or system) and I have internalized while others I will take as articles of faith and will probably ask about again at some point in the future because I don't have a good mental model of them. Anyways, it seems that making EnergyPlus state into a local variable in the client obviates these issues. Whether the other things it enables are "amazing bonuses" or just "more potential issues" and which issues supersede is a matter for discussion. 

What is your assessment?

Myoldmopar:
After discussion, I think I can capture where things stand right now.

- The API is OK as it is in develop.  We are giving the user a reference to a state instance, and while it may not be the most ideal solution, we limit issues by restricting to our public API.  If a client does other things with it, then we can't stop them, but we don't have to support any of those efforts. 
- Passing around a state instance vs accessing from a global variable is an internal design choice and does not affect the API.  In general, passing around a state instance isn't so bad except in those functions that only need it for error handling reasons.  We need to come up with a plan for handling those functions by passing an optional error handler or something.

The changes in this branch are deep and wide reaching, changing the state argument to a function call and modifying the API. 
 Now that I have a better feel for where we'll end up, I don't think it will actually be worth using this branch in the end, but I will keep this PR open so we can continue discussion as needed.

I may be missing a note or two so I'll update this if I think of additional commentary.

amirroth:
@Myoldmopar has the most significant points nailed down. I did have a long-ish Slack chat with him about this and I want to enter some of the points we went over into the official GitHub record.

The basic argument is this:
- @amirroth: Giving API clients a direct reference to state is potentially really bad. We should be willing to use a small number of global variables (e.g., one fixed sized array) and one mutex to avoid this and give users handles. 
- @lefticus @Myoldmopar: Global variables are really bad (at least in libraries). We should be willing to give clients a reference to the global state object in order to avoid having any global variables in EnergyPlus.

After discussions with @lefticus in this thread, I understand that global variables are more problematic than I thought (especially in libraries). I still believe that with some stipulations they are not showstoppers, but that becomes less relevant given ... 

After discussions with @Myoldmopar on Slack, I see that giving API clients a state reference is not as bad as I thought it may be from a support/compatibility sense (I still have some questions about security implications, but I don't know enough about cybersecurity to even pose them properly). Specifically, clients will have to do some additional work to use the state reference as anything other than a black-box handle, and all of that code will be explicitly "unsupported".

nrel-bot-3:

nrel-bot-3:

PR #8461, opened 2021-01-11
Schedule manager refactor (try 3)

Myoldmopar:
This is a preliminary proof of concept.  I'll be polishing it up before requesting review.


src/EnergyPlus/ScheduleManager.hh
```diff
@@ -68,39 +68,18 @@ struct EnergyPlusData;
...
-    extern bool ScheduleInputProcessed; // This is false until the Schedule Input has been processed.
-    extern bool ScheduleDSTSFileWarningIssued;
+    int constexpr MaxDayTypes = 12;
+    auto constexpr ValidDayTypes(int i) {
+        constexpr const char* const dayTypesVector[] = {
+            "DUMMYZERO", "Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday",
+            "Holiday", "SummerDesignDay", "WinterDesignDay", "CustomDay1", "CustomDay2"
+        };
+        return dayTypesVector[i];
+    }
```
amirroth, line 79: Why is this a function rather that just a simple array lookup? So that the value can be a constexpr? I don't believe there is any meaningful difference between a `const char *` and a `constexpr const char *`. Is this because we are "not allowed" to have any global arrays? Even constant ones? Because of potential lifetime problems again? My eyes are literally doing 360s.

Myoldmopar: @amirroth I would say don't stare at this too deeply at the moment.  This is definitely in a proof-of-concept stage, so this is one of the several things to be polished up before reviewing.

As for this specific issue though, we have a number of places in the code where we keep an array of strings of things.  These are often `extern`d and constructed at program start-up, which, yes, does have potential lifetime issues.  But more importantly -- they probably don't need to be there in the current way they exist.  There are definitely better designs, but the improvement to be applied depends on the actual usage of such lists.  
- Sometimes they are just used as a plain array that is looked up by operator.  In these cases, it would be great to keep a constexpr list of strings, maybe std::array?, but then you end up with array index issues because they are currently 1-based Objexx array lookups.  Not a problem, just means extra effort is needed when converting them.  Masking them as a function call with a static constexpr list inside it actually gets you the proper functionality without having to change the code everywhere it is used.  This is not intended to be a final answer, but it is a nice stopgap.
- In other cases, the array itself is iterated, so you cannot provide a function that just returns a single value from it.  In these cases, there are other, better solutions.
- In still other cases, the array can be completely eliminated, or if not eliminated, then at least moved to a static const list inside a function, with a much tighter scope than just hanging out in global space for no reason.

Let me give this all another pass and then we can discuss the best path forward on this and other issues.

Thanks!

nrel-bot-3:

nrel-bot-3:

Myoldmopar:
Closing again, at least temporarily.

PR #8479, opened 2021-01-23
Plant Cleanups

Myoldmopar:
Pull request overview
---------------------
 - This is simply a series of non-function-changing cleanups and minor refactors.  Mostly extracting out the "one-time" initialization from inside the component logic into its own function in each component.  This will lead toward the bigger goal of gutting the "logic" from inside the calc/sim function inside each component.

For now this should be a no-diff PR, if any show up, I'll need to figure out where it occurred and clean it out.  I'll plan on adding some more components to this no-diff PR prior to merging as well.


Myoldmopar:
Unrelated FileSystem test failure that happens occasionally, otherwise pretty much what I was expecting.  No impact from these changes so far.

Myoldmopar:
Python plugin diffs are unrelated and fixed in #8493.  Once that merges in I'll get fresh CI results here and then this can merge in.

Myoldmopar:
I believe this is ready to go.  The Windows results are unrelated and occurring on other branches.  Linux is showing the build is good and there are no regressions.  This is just a refactor to continue pushing plant components into a better state for cool solver interaction.  If no one has an issue with this, it should drop in right away to get out of the way for other PRs.

mitchute:
Looks straightforward. Merging.

PR #8480, opened 2021-01-25
Upgrade to C++17 std::filesystem library

### LINKED ISSUE #8376: EnergyPlus with option --expandobjects set fails when working directory is not on the same device as the output directory (on Linux)

jmarrec:
Pull request overview
---------------------

 - Fixes #8376
 - Original NFP was in https://github.com/NREL/EnergyPlus/pull/8411, but NFP doc was merged into this PR

### Pull Request Author

### Reviewer


cmake/PythonCopyStandardLib.py
```diff
@@ -25,12 +26,37 @@
...
+    std_lib_ctypes_sos = find_libs(standard_lib_dir)
+    this_lib_ctypes_sos = find_libs(target_dir)
+
+    if ((set(std_lib_ctypes_sos) - set(this_lib_ctypes_sos)) or
+        (set(this_lib_ctypes_sos) - set(std_lib_ctypes_sos))):
+        print("Detected changes in the python libs, wiping and recopying")
+        shutil.rmtree(target_dir)
+    else:
+        # File names match
+        sys.exit(0)
```
jmarrec, line 112: I got burned pretty badly by this script. I built with python 3.8 originally. Then I changed in the cmake to python3.9, but the shared library files didn't get copied again, so when running PythonPlugin example files, my python3.9 couldn't load the standard library in `python_standard_library/lib-dynload` since they were the 3.8 ones... 

5 hours later I realized this was the problem and not my fs::path changes.

src/ConvertInputFormat/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.5.1)
 project(ConvertInputFormat)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
jmarrec: I started touching this subfolder without realizing (...), so instead of reverting the changes I just bumped to C++17.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -767,37 +767,61 @@ endif()
...
+  target_link_libraries( energyplusparser PUBLIC dl)
 endif()
 if (WIN32)
-  target_link_libraries( energyplusparser Shlwapi )
+  target_link_libraries( energyplusparser PUBLIC Shlwapi )
+endif()
+
+# See https://en.cppreference.com/w/cpp/filesystem#Notes
+if ((CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.1) OR
+    ( ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang") AND (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)))
```
jmarrec, line 758: Some shenanigans that are needed for GCC < 9.1 or Clang < 9 to link to std::filesystem.

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -238,40 +255,41 @@ namespace CommandLineInterface {
...
-            state.dataGlobal->isBSON = true;
-            DisplayString(state, "BSON input format is experimental and unsupported.");
-        } else {
-            DisplayString(state, "ERROR: Input file must have IDF, IMF, or epJSON extension.");
-            exit(EXIT_FAILURE);
+        DataStringGlobals::inputFilePathNameOnly = FileSystem::removeFileExtension(FileSystem::getFileName(DataStringGlobals::inputFilePath));
+        DataStringGlobals::inputDirPath = FileSystem::getParentDirectoryPath(DataStringGlobals::inputFilePath);
+
+        // TODO: would this be better?
+        // FileSystem::InputFileType fileType = FileSystem::getInputFileType(DataStringGlobals::inputFilePath);
```
jmarrec: TODO: not sure if worth it to provide a storngly type enumeration or not.

jmarrec: Note: similar block of code can be found in ConvertInputFormat/main.cpp

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -322,143 +331,168 @@ namespace CommandLineInterface {
...
+
+        // TODO: why is this relative?
+        state.files.iniFile.filePath = "Energy+.ini";
+
+        // Stat next to epw
+        // Careful! fs::path::replace_extension will **mutate the original object**
+        state.files.inStatFilePath.filePath = state.files.inputWeatherFilePath.filePath;
+        state.files.inStatFilePath.filePath.replace_extension(".stat");
+        // Or is it better to provide a helper like this?
+        // state.files.inStatFilePath.filePath = FileSystem::replaceFileExtension(state.inputWeatherFilePath.filePath, ".stat");
```
jmarrec: TODO: couldn't really decide whether the helper is a good idea, or leave the burden of not messing it up to the developper?

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -322,143 +331,168 @@ namespace CommandLineInterface {
...
-        state.files.inStatFileName.fileName = weatherFilePathWithoutExtension + ".stat";
-        eplusADSFileName = inputDirPathName + "eplusADS.inp";
+
+        state.files.screenCsv.filePath = composePath(screenSuffix + ".csv");
+        // TODO, why are these relative paths?
+        state.files.delightIn.filePath = "eplusout.delightin";
+        DataStringGlobals::outputDelightOutFilePath = "eplusout.delightout";
+
+        // TODO: why is this relative?
+        state.files.iniFile.filePath = "Energy+.ini";
```
jmarrec: I'm not sure I understand why these were different? I guess for iniFile I do, it's because it'll look for the Energy+.ini in the current directory or next to the IDF later, but the delightout?

src/EnergyPlus/ConfiguredFunctions.hh
```diff
@@ -49,12 +49,13 @@
 #define CONFIGURED_FUNCTIONS_HH
 
 #include <string>
+#include <EnergyPlus/FileSystem.hh>
 
 namespace EnergyPlus {
 
-std::string configured_source_directory();
+fs::path configured_source_directory();
```
jmarrec, line 56: TODO: shouldn't these two be made `inline` (and move to header-only) (or at least `static`)?

cmake/PythonCopyStandardLib.py
Myoldmopar: Sorry, yeah, that definitely needed some love.  I haven't looked at your changes yet, but if you need me to make some updates there still I'm happy to do it.

src/EnergyPlus/CommandLineInterface.cc
Myoldmopar: Are you saying you could accidentally stomp on a previous value because it is mutating the underlying reference?  I don't think a helper should be really necessary, primarily because if a developer messes with that the wrong way, the simulation and the CI system should complain with huge red badges alerting that things went terribly wrong.  But, all that said, if a worker function can help avoid that ever happening, then OK, maybe it's worth it.

src/EnergyPlus/CommandLineInterface.cc
Myoldmopar: :shrug:

src/EnergyPlus/ConfiguredFunctions.hh
Myoldmopar: Yes, they should.  That was added in a bit of a rush, and ended up the way they are because they needed to be fixed in a hurry.  If you want to clean it up that is totally fine.

Myoldmopar:
Obviously CI isn't quite happy with you yet on this one, but I'm excited to see the progress.  Let me know if I can help with any of the issues you've identified.

cmake/PythonCopyStandardLib.py
Myoldmopar: I see in the code now that you added this block to do a very intelligent check on whether to update the library.  This looks great!

scripts/dev/find_byref_bool_override.py
```diff
@@ -111,13 +111,6 @@
         ]
     },
 
-    # Could be changed to just return a string
```
Myoldmopar, line 168: One less to check!

src/EnergyPlus/CMakeLists.txt
```diff
@@ -765,37 +765,63 @@ endif()
...
+         epexpat
+         epfmiimport
+         re2
+         DElight
+         libkiva
+         btwxt
+         Windows-CalcEngine
+         airflownetworklib
+         ssc
+)
```
Myoldmopar: I do like cleanups in the cmake.  I'd like to know how badly this will conflict with the changes in the work from @lefticus  to modernize our cmake (#8425).  I'm not saying you have to change your approach here, I'm just wanting to prepare myself mentally :laughing:

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -169,15 +167,32 @@ namespace CommandLineInterface {
...
+        {
+            // TODO: should this vbe in IOFiles as an InputFile?
+            std::string inputIddFileName;
+            opt.get("-i")->getString(inputIddFileName);
+            DataStringGlobals::inputIddFilePath = fs::path{inputIddFileName};
+        }
 
-        if (!opt.isSet("-i") && !legacyMode) inputIddFileName = exeDirectory + inputIddFileName;
+        if (!opt.isSet("-i") && !legacyMode) {
+            DataStringGlobals::inputIddFilePath = DataStringGlobals::exeDirectoryPath / DataStringGlobals::inputIddFilePath;
```
Myoldmopar: :heart_eyes: I know it's silly, but this is just nice.  It's the same as when I started using `pathlib` in Python.  This just makes so much sense.

src/EnergyPlus/CommandLineInterface.cc
Myoldmopar: I mean, that would certainly be great I think.  I can't think of a reason why we would need these individual flags inside E+.  @mbadams5 ?

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -322,143 +331,168 @@ namespace CommandLineInterface {
...
+                DisplayString(state, "ERROR: Unrecognized argument for output suffix style: " + suffixType);
+                DisplayString(state, errorFollowUp);
+                exit(EXIT_FAILURE);
+            }
         }
 
+        // Helper to construct output file path
+        auto composePath = [&outputFilePrefixFullPath](const std::string& suffix) -> fs::path {
+            return fs::path(outputFilePrefixFullPath.string() + suffix);
+        };
```
Myoldmopar, line 417: :+1:

src/EnergyPlus/CommandLineInterface.cc
```diff
@@ -501,36 +535,39 @@ namespace CommandLineInterface {
...
-            } else {
-                CurrentWorkingFolder.erase(TempIndx + 1);
-            }
+            // TODO: not sure I understand this block
+            //const auto TempIndx = index(DataStringGlobals::CurrentWorkingFolder, DataStringGlobals::pathChar, true);
+            //if (TempIndx == std::string::npos) {
+                //DataStringGlobals::CurrentWorkingFolder = "";
+            //} else {
+                //DataStringGlobals::CurrentWorkingFolder.erase(TempIndx + 1);
+            //}
```
Myoldmopar, line 601: :shrug:

src/EnergyPlus/ConfiguredFunctions.hh
Myoldmopar: They can just be `constexpr char *` variables now actually, which is even simpler.

src/EnergyPlus/ExternalInterface.cc
```diff
@@ -1120,12 +1123,16 @@ namespace ExternalInterface {
             // in any case, the relative paths work fine here
 
             // post process as needed in case these are used later
-            FMURootWorkingFolder = "tmp-fmus";
-            FMURootWorkingFolder += pathChar; // getStringFromCharArray( FMUWorkingFolderCharArr );
+            FMURootWorkingFolder = fs::path("tmp-fmus"); // getStringFromCharArray( FMUWorkingFolderCharArr );
```
Myoldmopar: All these corners of the code...

src/EnergyPlus/FileSystem.cc
```diff
@@ -176,93 +188,102 @@ namespace FileSystem {
...
-        else
-            return false;
-#else
-        struct stat info;
-        if (stat(filePath.c_str(), &info) == 0) {
-            return !(info.st_mode & S_IFDIR);
-        } else
-            return false;
-#endif
+        return fs::exists(filePath) && !fs::is_directory(filePath);
```
Myoldmopar, line 276: This.  :+1: :+1: :+1:

Myoldmopar (commented):
This is **a lot** of changes.  Really good stuff.  I'm hesitant to approve it because CI is all red, but I don't see any problems from a review of it.  I love where this is going.

src/EnergyPlus/CommandLineInterface.cc
jmarrec: It mutates the path it's called on yes: https://en.cppreference.com/w/cpp/filesystem/path/replace_extension . Signature is `path& replace_extension( const path& replacement = path() );`

See an example on https://godbolt.org/z/Mvze3W

```c++
#include <filesystem>
#include <iostream>

namespace fs = std::filesystem;

int main() {

    fs::path inputWeatherFilePath("/path/to/myfile.epw");
    fs::path inStatFilePath = inputWeatherFilePath.replace_extension(".stat");
    std::cout << "inputWeatherFilePath=" << inputWeatherFilePath.string() << '\n';
    // !! => inputWeatherFilePath=/path/to/myfile.stat
    std::cout << "inStatFilePath=" << inStatFilePath.string() << '\n';

}
```

The thing with having a helper function is that nothing prevents anyone from **not** using it and directly calling the `fs::path::replace_extension` instead of the helper...

src/EnergyPlus/CMakeLists.txt
jmarrec: I have seen #8425 and aligned my use of PUBLIC/PRIVATE attributes to match, so it shouldn't be that bad. (The format of this block (switching to one per line) is also copied over from lefticus's PR)

Plus, I'm touching like 4 or 5 blocks in the CMakeLists.txts, so the merge conflicts should be fine. I'd be happy to fix conflicts on whichever PR is merged second.

src/EnergyPlus/ConfiguredFunctions.hh
jmarrec: If I make them `constexpr char *` then all calls would have to look like `fs:path(configured_source_directory()) / xxx`

lefticus: `constexpr string_view` would be better than `constexpr char *` but in either case, yes you'd have to call your conversion function to use them.

src/EnergyPlus/FileSystem.hh
```diff
@@ -49,43 +49,108 @@
 #define FileSystem_hh_INCLUDED
 
 #include <algorithm>
+#include <string>
+#include <fmt/format.h>
+#if __has_include(<filesystem>)
```
lefticus, line 54: `__has_include` might not necessarily work 100% how we want it to - many compiler vendors provide the headers for features they don't yet support, then give you a compile time warning if you try to use features from that header.

A better option is compiler feature test macros https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros#C.2B.2B17 but those are not officially supported until C++20, and MSVC doesn't enable them unless you use C++20 mode (other compilers support them before C++20).

TL;DR, don't change anything, this is probably your best option for today, but I thought it was worth mentioning in case we do have an issue in the future with this.

jmarrec:
I merged develop and resolved conflicts. This new warning pops up:

```c++
[ 38%][479/1240] Building CXX object src/EnergyPlus/AirflowNetwork/CMakeFiles/airflownetworklib.dir/src/Properties.cpp.o
In file included from /home/julien/Software/Others/EnergyPlus/src/EnergyPlus/IOFiles.hh:58:0,
                 from /home/julien/Software/Others/EnergyPlus/src/EnergyPlus/DataGlobals.hh:57,
                 from /home/julien/Software/Others/EnergyPlus/src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/../../../Psychrometrics.hh:62,
                 from /home/julien/Software/Others/EnergyPlus/src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Properties.hpp:52,
                 from /home/julien/Software/Others/EnergyPlus/src/EnergyPlus/AirflowNetwork/src/Properties.cpp:48:
/home/julien/Software/Others/EnergyPlus/src/EnergyPlus/FileSystem.hh:104:64: warning: ‘no_discard’ attribute directive ignored [-Wattributes]
     [[no_discard]] fs::path makeNativePath(fs::path const &path);
                                                                ^
/home/julien/Software/Others/EnergyPlus/src/EnergyPlus/FileSystem.hh:106:65: warning: ‘no_discard’ attribute directive ignored [-Wattributes]
     [[no_discard]] fs::path getFileName(fs::path const &filePath);
```


@lefticus is this equivalent to `set(CXX_STANDARD 17)`? 

https://github.com/NREL/EnergyPlus/blob/e8bf5eab3283afa744b5fa8a5cc6a6a9d5cff10c/CMakeLists.txt#L25

jmarrec:
I get the no_discard warnings with gcc 7 and gcc 10.


FYI, There are other unrelated warnings too in Eigen

https://github.com/NREL/EnergyPlus/runs/1888058655#step:6:229

lefticus:
> 
> 
> I get the no_discard warnings with gcc 7 and gcc 10.
> 
> FYI, There are other unrelated warnings too in Eigen
> 
> https://github.com/NREL/EnergyPlus/runs/1888058655#step:6:229

It should be `[[nodiscard]]` no `_`

jmarrec:
> It should be `[[nodiscard]]` no `_`

:facepalm: . Thank you..

src/EnergyPlus/FileSystem.hh
kbenne: I think cmake's check_include_file might be the best option https://cmake.org/cmake/help/latest/module/CheckIncludeFile.html

Myoldmopar:
Great work @jmarrec !  I will test this out if you feel like you are done with it.

jmarrec:
@Myoldmopar I think it's ready for review. the Clang build is failing because it's not targeting 10.15 and getting the `'path' is unavailable: introduced in macOS 10.15` stuff

Myoldmopar:
This will be merged right after 9.5 is released to avoid breaking support for 10.14 so close to the release.

nrel-bot-3:

jmarrec:
Test GHA build on my fork at https://github.com/jmarrec/EnergyPlus/releases/tag/FileSystem_Move_v5

jmarrec:
I can reproduce the windows failing test locally if I run several tests in // that call ManageSizing.

Single test: this works

```
$ ctest -C Debug -R DesiccantDehum_VSCoolingCoilOnPrimaryAirSystemTest
```

This fails:
```
$ ctest -C Debug  -j 6 -R '(DesiccantDehum_|WindowAC_VStest1)'
```

With the added file.ssz.open_as_stringstream() in https://github.com/NREL/EnergyPlus/pull/8480/commits/0dabeb010903c36a37215ad0d1a229941c3feedb it seems to work.

Myoldmopar:
Looks so clean!!!  I'll take a final look and see if I can get this merged in.

.decent_ci-MacOS.yaml
```diff
@@ -1,3 +1,3 @@
 compilers:
   - name: clang
-    cmake_extra_flags: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13 -DBUILD_FORTRAN=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=ON -DREGRESSION_BASELINE_PATH:PATH=$REGRESSION_BASELINE -DREGRESSION_SCRIPT_PATH:PATH=$REGRESSION_DIR -DREGRESSION_BASELINE_SHA:STRING=$REGRESSION_BASELINE_SHA -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF -DLINK_WITH_PYTHON=ON
+    cmake_extra_flags: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15 -DBUILD_FORTRAN=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=ON -DREGRESSION_BASELINE_PATH:PATH=$REGRESSION_BASELINE -DREGRESSION_SCRIPT_PATH:PATH=$REGRESSION_DIR -DREGRESSION_BASELINE_SHA:STRING=$REGRESSION_BASELINE_SHA -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF -DLINK_WITH_PYTHON=ON
```
Myoldmopar, line 3: Note to self for later release testing.

cmake/PythonCopyStandardLib.py
Myoldmopar: Still looks great :smile:

src/EnergyPlus/DataStringGlobals.hh
```diff
@@ -82,68 +83,71 @@ namespace DataStringGlobals {
...
+    fs::path idfDirPath;
+    fs::path outDirPath;
+    fs::path idfFilePathNameOnly;
+    fs::path inputDirPath;
+    fs::path outputDirPath;
+    fs::path inputFilePathNameOnly;
+    fs::path exeDirectoryPath;
+    fs::path ProgramPath;          // Path for Program from INI file
+    fs::path CurrentWorkingFolder; // Current working directory for run
+    fs::path FullPath;             // Full name of file to open, including path
```
Myoldmopar, line 129: :heart_eyes:

Myoldmopar (approved):
Built and tested locally, all good.  This is awesome and ready to go in.  Thanks for powering through @jmarrec !

Issue #7632, opened 2019-11-22
Add helper functions to validate and get an epJSON object and field value


mjwitte:
Issue overview
--------------
A couple of common patterns for input processing functions that are accessing epJSON data directly are:
````
From UnitarySystem::getUnitarySystemInputData
                auto const &fields = instance.value();

                if (fields.find("heating_coil_name") != fields.end()) { // not required field
                    loc_m_HeatingCoilName = UtilityRoutines::MakeUPPERCase(fields.at("heating_coil_name"));
                }

From WaterToWaterHeatPumpEIR::processInputForEIRWWHP:
                        if (fields.find("reference_coefficient_of_performance") != fields.end()) {
                            thisWWHP.referenceCOP = fields.at("reference_coefficient_of_performance");
                        } else {
                            Real64 defaultVal = 0.0;
                            if (!inputProcessor->getDefaultValue(cCurrentModuleObject,
                                                                 "reference_coefficient_of_performance", defaultVal)) {
                                // this error condition would mean that someone broke the input dictionary, not their
                                // input file.  I can't really unit test it so I'll leave it here as a severe error
                                // but excluding it from coverage
                                ShowSevereError(  // LCOV_EXCL_LINE
                                        "EIR WWHP: Reference COP not entered and could not get default value"); // LCOV_EXCL_LINE
                                errorsFound = true;  // LCOV_EXCL_LINE
                            } else {
                                thisWWHP.referenceCOP = defaultVal;
                            }
                        }
````
That's a lot of work to repeat just to get an input field value or it's default.

Also, as evidenced by #7624 a simple typo in the field name reference can result in an input field being missed, because there's no validation that the requested field is a valid field name for the object.

So, it would be useful to have an input processor function that 

1. Takes a field name, validates the field name, and returns the input value or it's default.
2. To validate field name, will likely have to pass in `cCurrentObjectItem` (object class name) as an argument.
2. If the field doesn't have a declared default, then numeric should return zero, and string should return blank.
3. The EIRWWHP test for no default and no input is unusual, but for objects that want to test this for a given field, perhaps this function could return a status flag? (0 = user input, 1 = default, -1 = no user input or default) or ???

For a field where you simply want to get the value or the default and move on, the goal would be something like:
```
loc_m_HeatingCoilName = inputProcessor->getFieldValue("heating_coil_name", fields, cCurrentObjectItem, status);
```

### Details
Some additional details for this issue (if relevant):
 - Platform (Operating system, version)
[...]

PR #8491, opened 2021-01-29
GHE Updates

mitchute:
Pull request overview
---------------------
 - Moves GHE data to state
 - Moves GHE SetupOutputVars to shared member method. Note that this causes rdd, mdd, and aud diffs. See here for the offending commit: 68ab73fd9c66b1ca3c5b466e950df1ea7b4e89ae
 - Moves GHE input processing to ctor's that accept JSON input. This is a potential prototype for future refactoring when moving away from the traditional input processing methods.
 - Cleans up local variables
 - Add some missing unit test support

### Pull Request Author

### Reviewer


src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -122,71 +119,433 @@ namespace GroundHeatExchangers {
...
+        this->pipe.thickness = j["pipe_thickness"];
+
+        std::string const hxConfig = UtilityRoutines::MakeUPPERCase(j["heat_exchanger_configuration"]);
+        if (UtilityRoutines::SameString(hxConfig, "VERTICAL")) {
+            this->verticalConfig = true;
+        } else if (UtilityRoutines::SameString(hxConfig, "HORIZONTAL")) {
+            this->verticalConfig = false;
+        }
+
+        this->coilDiameter = j["coil_diameter"];
```
mjwitte, line 178: @mitchute This is an interesting direction to move in. 
1. What happens if the field name string is misspelled? Will it throw an error or simply return a zero or blank?
2. What happens if the field isn't present in the incoming object? Will it be present in the object data with it's default if it has one?
I still feel that we need an input fetching utility to manage this as suggested in https://github.com/NREL/EnergyPlus/issues/7632 .
Or maybe fields with defaults are already populated in the JSON data structure when the input is read in? I've never been sure about that.

mitchute: @mjwitte

Re 1. I just tested this by misspelling one field, ```depth_of_top_of_borehole```. The following is the error message thrown.

```
** Severe  ** [json.exception.type_error.302] type must be number, but is number
```

Not super helpful, but one would hope that these sort of errors never make it past the developer/reviewer since the model likely would never run. We should give this a little more thought.

Re 2. We handle fields that could be missing and which do NOT have default values like the following:
```C++
if (j.find("ghe_vertical_responsefactors_object_name") != j.end())
```
In this case, the field will not be present if it's not present in the input. I'd assume that fields with default values will have the default filled in automatically, but I have not tested that.

FFR, some issues I had with this transition are listed here:
 - The string fields come in as-is from the input. I found myself needing to upper case everything before comparing strings, whereas before I think some of this was done automatically. We'll need a way to keep some fields case sensitive, such as paths. However, the majority of strings don't need that, so when we go to scale this up we'll need to decide how to best handle it.
 - Currently, we have ```cAlphaFieldNames``` and ```cNumericFieldNames``` which allows us to programmatically get the field names. I didn't figure out a way to get these using JSON, but it probably exists. Either way, the names are likely to come in as the JSON key names. We may want a way to provide the user with the IDF-like field names? I.e. "inlet_node_name" vs. "Inlet Node Name".

This was my first real attempt at using the JSON inputs, so perhaps these issues were just due to my inexperience.

Myoldmopar: I am really excited about this pattern.  And one of my favorite reasons is that unit testing the input processing and catching casing/other issues will be easy.  Just call the constructor will different JSON objects and verify the created object matches the expectation.  No need to do any other setup (unless the constructor is accessing something that needs to be set up).

Myoldmopar (approved):
I really like this.  Having the input processor routine rewritten as an object constructor that accepts JSON input is something we've wanted for quite some time.  We can critique this moving forward and optimize it as we improve more and more components.

src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -1386,31 +1747,22 @@ namespace GroundHeatExchangers {
...
-        sqrtAlphaT = std::sqrt(soil.diffusivity * t);
+        Real64 distance1 = distance(m, n, m1, n1, eta, theta);
+        Real64 sqrtAlphaT = std::sqrt(this->soil.diffusivity * t);
 
         if (!verticalConfig) {
 
-            sqrtDistDepth = std::sqrt(pow_2(distance1) + 4 * pow_2(coilDepth));
-            errFunc1 = std::erfc(0.5 * distance1 / sqrtAlphaT);
-            errFunc2 = std::erfc(0.5 * sqrtDistDepth / sqrtAlphaT);
+            Real64 sqrtDistDepth = std::sqrt(pow_2(distance1) + 4 * pow_2(this->coilDepth));
```
Myoldmopar, line 1755: Localized variables, great.

src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -1297,8 +1655,10 @@ namespace GroundHeatExchangers {
                                 J0 = 561;
                             }
 
+                            Real64 gFuncin;
+
                             // if the ring(n1, m1) is the near-field ring of the ring(n,m)
-                            if (disRing <= 2.5 + coilDiameter) {
+                            if (disRing <= 2.5 + this->coilDiameter) {
```
Myoldmopar, line 1661: Member variables, awesome.

src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -122,71 +119,432 @@ namespace GroundHeatExchangers {
...
+        this->soil.k = j["soil_thermal_conductivity"];
+        this->soil.rho = j["soil_density"];
+        this->soil.cp = j["soil_specific_heat"];
+        this->soil.rhoCp = this->soil.rho * this->soil.cp;
+        this->pipe.k = j["pipe_thermal_conductivity"];
+        this->pipe.rho = j["pipe_density"];
+        this->pipe.cp = j["pipe_specific_heat"];
+        this->pipe.outDia = j["pipe_outer_diameter"];
+        this->pipe.outRadius = this->pipe.outDia / 2.0;
+        this->pipe.thickness = j["pipe_thickness"];
```
Myoldmopar, line 169: Once again, really excited about this.

src/EnergyPlus/GroundHeatExchangers.cc
```diff
@@ -2127,703 +2472,207 @@ namespace GroundHeatExchangers {
...
+                auto const &objNameUC = UtilityRoutines::MakeUPPERCase(objName);
+                inputProcessor->markObjectAsUsed(currObj, objName);
+                state.dataGroundHeatExchanger->slinkyGLHE.emplace_back(state, objNameUC, instance);
             }
         }
     }
 
     //******************************************************************************
 
+    void GLHEBase::setupOutput(EnergyPlusData &state)
```
Myoldmopar, line 2622: Reorganization, causing rdd diffs.  OK.

src/EnergyPlus/GroundHeatExchangers.hh
```diff
@@ -326,40 +340,40 @@ namespace GroundHeatExchangers {
...
-        doubleIntegral(std::shared_ptr<GLHEVertSingleStruct> const &bh_i,
-                       std::shared_ptr<GLHEVertSingleStruct> const &bh_j, Real64 const &currTime);
+        doubleIntegral(std::shared_ptr<GLHEVertSingle> const &bh_i, std::shared_ptr<GLHEVertSingle> const &bh_j, Real64 const &currTime);
 
         void calcShortTimestepGFunctions(EnergyPlusData &state);
 
         void calcLongTimestepGFunctions(EnergyPlusData &state);
 
-        void calcGFunctions(EnergyPlusData &state);
+        void calcGFunctions(EnergyPlusData &state) override;
```
Myoldmopar, line 362: Proper override specification.  Good.

src/EnergyPlus/StateManagement.cc
```diff
@@ -183,7 +182,6 @@ void EnergyPlus::clearAllStates(EnergyPlusData &state)
     Furnaces::clear_state();
     General::clear_state();
     GeneralRoutines_clear_state(); // GeneralRoutines does not have a namespace
-    GroundHeatExchangers::clear_state();
```
Myoldmopar, line 186: One less state, woohoo!

Myoldmopar (approved):
All good here!  Let me make sure the branch is up to date with develop and if not I'll just build it locally anyway prior to merging.  Thanks @mitchute

Myoldmopar:
Branch is 36 commits ahead of develop so the latest CI results are still good.  Windows failures are unrelated, and otherwise everything seems happy.  Good to go, thanks @mitchute !

PR #8498, opened 2021-02-01
Fixes Zone Air Mass Balance report variables reporting issues

### LINKED ISSUE #8497: Report variable Zone Air Mass Balance Supply Mass Flow Rate is always zero
Issue overview
--------------
Testing a user file shows that "Zone Air Mass Balance Supply Mass Flow Rate" report variable is always zero when the zone supply air inlet node has flow. The value of this variable is always expected to match the zone supply air inlet node air mass flow rate.

### Details
 - Platform: WIN10
 - Version of EnergyPlus: 9.4
 - test file from email commuication

### Checklist

See a plot showing this issue.

ZoneAIrMassBalanceTest

Nigusse:
Pull request overview
---------------------
*Hoping to merge this before #8460 to reduce diffs there. MJW*

 - Fixes #8497
 - Zone Mass Balance reporting variables are set to zero when the infiltration treatment method selected is "None".
 - Fixes Zone Mass Balance issue when infiltration treatment method is adjust infiltration.
 - Revised the zone return node air flow calculation when the zone mass flow conservation is active.
 - Added a missing term to infiltration air mass flow rate calculation in infiltration treatment to enforce zone air flow balance. This code change item impacts results only when a zone is a source zone and a receiving zone at the same time.
 - Diffs are expected in example files that has ZoneAirMassFlowConservation object.

### Pull Request Author

### Reviewer


Nigusse:
Code review and testing shows that this is reporting issue. An IF logic [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L4000-L4048) used for infiltration treatment check is preventing from setting values to the supply and exhaust air flow rates report variables.

After moving the variable reset statements outside the if logic resolves the issue.  See the results below.

ZoneAIrMassBalanceTest_fixed

Nigusse:
Added defect file.

Nigusse:
Now the reporting issue resolved, I investigated the zone air flow balance issue. Run the defect file for the three infiltration treatment methods: None, AddInfiltrationFlow, and AdjustInfiltrationFlow.  Below are the results:
- **None**: the kitchen is balanced but not the dining zone. This is not unexpected because you need infiltration air flow to balance source zones in most cases.
- **AddInfiltrationFlow**: the kitchen and the dining zone are balanced. 
- **AdjustInfiltrationFlow**: the kitchen is balanced but not the dining zone. This needs investigation.

Nigusse:
See results of zone air flow mass balance before the fix. The dining zone air flow mass zone balance equation residual ranges between 0.0 to 1.488 kg/s as shown in the plot below. The residual is calculated as follows:

Residual = [m_{sup} - m_{exh} - m_{ret} + m_{zmreceiving} - m_{zmsource} + m_{inf}]


ZoneAIrMassBalanceTest_wAdjustInf_BeforeFix

Nigusse:
Looking at the code for **AdjustInfiltrationFlow** method [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/ZoneEquipmentManager.cc#L4006-L4017) the else part of the IF logic may need to change as follows:

```
                            if (ZoneAirMassFlow.InfiltrationTreatment == AdjustInfiltrationFlow) {
                                if (std::abs(ZoneInfiltrationMassFlowRate) > ConvergenceTolerance) {
                                    ZoneInfiltrationFlag(ZoneNum) = true;
                                    MassConservation(ZoneNum).InfiltrationMassFlowRate = ZoneInfiltrationMassFlowRate;
                                    MassConservation(ZoneNum).IncludeInfilToZoneMassBal = 1;
                                    Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate = ZoneInfiltrationMassFlowRate;
                                    Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate =
                                        max(0.0, Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate);
                                } else {
                                    //MassConservation(ZoneNum).InfiltrationMassFlowRate =
                                    //    Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate;
                                    Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate = 0.0;
                                    MassConservation(ZoneNum).InfiltrationMassFlowRate = 0.0;
                                }
```

With this change the residual of the zone air mass balance that includes the infiltration term yields a balanced flow in the dinning zone. Note that in this zone air flow balance equation, the infiltration term is no assumed self-balanced. See below the residual plot after the fix.

ZoneAIrMassBalanceTest_wAdjustInf_AfterFix

Nigusse:
Below are plots showing before (Develop) and after (Branch) results. Run the defect file for the three infiltration treatment methods: None, AddInfiltrationFlow, and AdjustInfiltrationFlow. Below are the results:

**Infiltration Treatment Method is set to None**: the kitchen is always balanced but the dining zone is not balance for both before and after the fix. The dining zone is unbalanced when the HVAC system off; otherwise, it is balanced. The unbalance is caused by limitation of the method.

Develop: before the fix.
RestaurantBLDG_NoneInfiltration_Develop

Branch #8497: after the fix.
RestaurantBLDG_NoneInfiltration_Branch

Nigusse:
**Infiltration Treatment Method is set to AdjustInfiltrationFlow**: the kitchen and the dinning zone are always balanced after the fix. Before the fix the dinning zone was not balanced (shows large residual) but after the fix that is not the case as shown by the second plot. Note that the dinning zone return flow is way small than the supply. The third infiltration treatment method also shows balanced flow results.

Develop: before the fix:
RestaurantBLDG_AdjustInfiltration_Develop

Branch #8497: after the fix:
RestaurantBLDG_AdjustInfiltration_Branch

mjwitte:
@Nigusse Are these runs with the "phantom" Dining zone exhaust fan present or not?

Nigusse:
Yes, the dinning room exhaust fan is present. I will post new results for one case (add infiltration method) without the dining room exhaust fan.

Nigusse:
Results without Dinning Room Exhaust Fan.

**Infiltration Treatment Method is set to AddInfiltrationFlow**: the kitchen and the dinning rooms are always balanced for both before and after the fix. However, the return air flows have changed significantly. The dinning room return air flow has reduced now because of increased zone air transfer from the dinning room to the kitchen. Before the fix, there was hardly any transfer of air. It is exactly for the same reason that the return air flow of the kitchen room has increased after the fix.

Develop: before the fix:
RestaurantBLDG_AddInf_woDiningExhFan_Develop



Branch: after the fix:
RestaurantBLDG_AddInf_woDiningExhFan_Branch

Nigusse:
I see big diffs in three example files:
 - 5ZoneAirCooled_ZoneAirMassFlowBalance.idf 
 - 5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized.idf
 - SolarShadingTestGPU.idf

The diffs in the first two example files is due to this fix and I am not sure if the SolarShadingTestGPU .idf diffs is related to this code change.  I will examine the example ZoneAirCooled_ZoneAirMassFlowBalance.idf diffs next.

mjwitte:
SolarShadingTestGPU.idf throws diffs often, but not always. My guess is that something's not getting initialized properly, but I wasn't able to determine that for sure. You can ignore those diffs here.

Nigusse:
Looking at example file 5ZoneAirCooled_ZoneAirMassFlowBalance.idf, it has the following diffs: EIO diffs, ESO big diffs, SSZ big diffs, ZSZ big diffs and Table big diffs.  The second example also shows similar diffs. 

First let us look at the zone air mass flow balance residuals for thermal zone SPACE1-1 and SPACE5-1. SPACE1-1 is a source zone, and has infiltration air specified. SPACE5-1 is a receiving zone and does not have infiltration air specified.  The plots below show the zone air flow balance equation residual for Develop and this Branch for SPACE1-1 and SPACE5-1. The residuals for SPACE1-1 are nearly identical except Develop has a small blip at 07/09 at 11 am. However, for SPACE5-1, the residual is almost zero all the time for the Branch whereas Develop has significant residual. This indicates the fix has improved the zone air mass flow balance equation convergence.

5ZoneAirCooled_ZoneAirMassFlowBalance_AdjustInf_Comparison

Nigusse:
Now let us look at the zone air temperature for SPACE1-1 and SPACE5-1.  The zone air temperature shows significant diffs during heating season and the diffs are larger for the SPACE1-1 than SPACE5-1.  For SPACE1-1, Branch shows better zone air temperature control than Develop. For SPACE5-1, both Develop and Branch show similar zone air temperature.

5ZoneAirCooled_ZoneAirMassFlowBalance_AdjustInf_ZoneAirTempComparison

Nigusse:
Now let us look at the zone return node air flow for SPACE1-1 and SPACE5-1.  The zone return node air flow for the Branch is reduced compared to Develop for all four perimeter zones, which are specified as source zones. For SPACE5-1, which is a receiving zone (from all the four zones), has increased return node air flow. The new return node air flow trend looks consistent with expectation, I think.

5ZoneAirCooled_ZoneAirMassFlowBalance_AdjustInf_ReturnAirFlowComparison

Nigusse:
Now let us look at the zone mixing source and receiving air flow for SPACE1-1 and SPACE5-1. The zone mixing source and receiving air flows show significant difference between Develop and this Branch. Noe that the ZoneMixing and Exhaust fan have always ON schedule. The Branch shows that the ZoneMixing flows show less variability compared to Develop. The user specified zone mixing flows are modified to balance the zone air flow but are always greater than zero and I believe this trend is consistent with the expectation. And this is the reason that we saw reduced return node air flow for the source zones (SPACE1-1 through SPACE4-1)and an increase for the receiving zone (SPACE5-1).

5ZoneAirCooled_ZoneAirMassFlowBalance_AdjustInf_ZoneMixingFlowComparison

Nigusse:
Now let us look at the zone infiltration air flow for SPACE1-1 and zone exhaust fan air flow for SPACE5-1. The infiltration air flows show significant difference between Develop and this Branch. Note that infiltration object schedule for all the four perimeter zones is always ON.   SPACE5-1 does not have infiltration specified. Again here the infiltration treatment method is AdjustInfiltrationFlow and the infiltration air flows were adjusted to balance the zone air flow.  The exhaust fan in SPACE5-1 is always ON and shows constant flow for both Develop and Branch, as expected.

5ZoneAirCooled_ZoneAirMassFlowBalance_AdjustInf_ZoneInfiltrationFlowComparison

Nigusse:
Regression run results shows significant diffs but the results after the fix makes more sense to me and also shows better zone air temperature control. @mjwitte Please let me know if you have any concern.

Nigusse:
Added a defect file with dining room exhaust fan removed.

Nigusse:
Diffs are expected in the two example files in the test suite that contain ZoneAirMassFlowConservation object. Ran unit test suite locally and all passed. A new unit test has been added. This branch is ready for review.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -3967,6 +3966,8 @@ namespace EnergyPlus::ZoneEquipmentManager {
                                     state.dataZoneEquip->ZoneEquipConfig(ZoneNum).TotInletAirMassFlowRate + MassConservation(ZoneNum).MixingSourceMassFlowRate);
                     }
                     CalcZoneMixingFlowRateOfReceivingZone(ZoneNum, ZoneMixingAirMassFlowRate);
+                    ZoneMixingNetAirMassFlowRate = MassConservation(ZoneNum).MixingMassFlowRate - MassConservation(ZoneNum).MixingSourceMassFlowRate;
+
```
Nigusse, line 3970: Calculates update the mixing objects net zone mass flow rate for immediate use, instead of waiting until the next iteration.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -3852,7 +3852,6 @@ namespace EnergyPlus::ZoneEquipmentManager {
         Real64 TotInletAirMassFlowRateMaxAvail;
         Real64 TotInletAirMassFlowRateMin;
         Real64 TotInletAirMassFlowRateMinAvail;
-        Real64 TotSupplyAirMassFlowRate;
```
Nigusse, line 3855: removed unused local variable.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4012,8 +4012,8 @@ namespace EnergyPlus::ZoneEquipmentManager {
                                     Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate =
                                         max(0.0, Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate);
                                 } else {
-                                    MassConservation(ZoneNum).InfiltrationMassFlowRate =
-                                        Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate;
+                                    Infiltration(MassConservation(ZoneNum).InfiltrationPtr).MassFlowRate = 0.0;
+                                    MassConservation(ZoneNum).InfiltrationMassFlowRate = 0.0;
```
Nigusse, line 4016: When infiltration treatment is AdjustInfiltrationFlow, the adjusted infiltration must be set to zero if the infiltration term calculated less than the convergence tolerance limit.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -3993,14 +3994,13 @@ namespace EnergyPlus::ZoneEquipmentManager {
...
                 CalcZoneReturnFlows(state, ZoneNum, StdTotalReturnMassFlow, FinalTotalReturnMassFlow);
-
                 MassConservation(ZoneNum).RetMassFlowRate = FinalTotalReturnMassFlow;
 
                 // Set zone infiltration flow rate
                 if (ZoneAirMassFlow.InfiltrationTreatment != NoInfiltrationFlow) {
                     if (MassConservation(ZoneNum).InfiltrationPtr > 0) {
                         if (MassConservation(ZoneNum).IsOnlySourceZone || (ZoneAirMassFlow.InfiltrationZoneType == AllZones)) {
-                            ZoneInfiltrationMassFlowRate = MassConservation(ZoneNum).MixingSourceMassFlowRate +
+                            ZoneInfiltrationMassFlowRate = MassConservation(ZoneNum).MixingSourceMassFlowRate - MassConservation(ZoneNum).MixingMassFlowRate +
```
Nigusse, line 4003: This is a missing term added. This would have been caught early if we had a test cases with a thermal zone that is used as a source and receiving zone.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4041,16 +4041,12 @@ namespace EnergyPlus::ZoneEquipmentManager {
                         // Zone has no infiltration objects
                         MassConservation(ZoneNum).InfiltrationMassFlowRate = 0.0;
                     }
-
+                }
+                //
+                if (ZoneMassBalanceFlag(ZoneNum)) {
                     MassConservation(ZoneNum).InMassFlowRate = state.dataZoneEquip->ZoneEquipConfig(ZoneNum).TotInletAirMassFlowRate;
                     MassConservation(ZoneNum).ExhMassFlowRate = state.dataZoneEquip->ZoneEquipConfig(ZoneNum).TotExhaustAirMassFlowRate;
-                    ZoneMixingNetAirMassFlowRate = MassConservation(ZoneNum).MixingMassFlowRate - MassConservation(ZoneNum).MixingSourceMassFlowRate;
```
Nigusse, line 4047: Moved up.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4041,16 +4041,12 @@ namespace EnergyPlus::ZoneEquipmentManager {
...
                 }
-                //
-
-                TotSupplyAirMassFlowRate = state.dataZoneEquip->ZoneEquipConfig(ZoneNum).TotInletAirMassFlowRate -
-                                           (state.dataZoneEquip->ZoneEquipConfig(ZoneNum).TotExhaustAirMassFlowRate - state.dataZoneEquip->ZoneEquipConfig(ZoneNum).ZoneExh) -
-                                           state.dataZoneEquip->ZoneEquipConfig(ZoneNum).PlenumMassFlow;
```
Nigusse, line 4053: Removed unused code section.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4234,6 +4230,10 @@ namespace EnergyPlus::ZoneEquipmentManager {
                         returnNodeMassFlow = inletMassFlow;
                         thisZoneEquip.FixedReturnFlow(returnNum) = true;
                     }
+                    // if zone mass balance true, set to expected return flow
+                    if (DataHeatBalance::ZoneAirMassFlow.EnforceZoneMassBalance ) {
+                        returnNodeMassFlow = ExpTotalReturnMassFlow;
+                    }
```
Nigusse, line 4236: Resets the local variable to the expected zone total return air mass flow rate determined by the zone air mass flow balance calculation. When the zone mass conservation is active, this new code overrides the equation that sets the return mass flow rate to the supply inlet node flow rate.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -4264,6 +4264,10 @@ namespace EnergyPlus::ZoneEquipmentManager {
                                 returnNodeMassFlow = max(0.0, (ExpTotalReturnMassFlow * returnSchedFrac * airLoopReturnFrac));
                             }
                         }
+                        // if zone mass balance true, set to expected return flow
+                        if (DataHeatBalance::ZoneAirMassFlow.EnforceZoneMassBalance ) {
+                            returnNodeMassFlow = ExpTotalReturnMassFlow; 
+                        }
```
Nigusse, line 4270: Same as the above explanation.

Nigusse (commented):
Code change walk through for PR8498 and is ready for review.

mjwitte:
@Nigusse Defect file RAT_VAVSys_wZnMxYes_AddInf_woDiningExhFan has some new warnings that look suspicious. Please check to see where these are coming from.
```
   ** Warning ** Enthalpy out of range (PsyTsatFnHPb)
   **   ~~~   **  Routine=Unknown, Environment=JANUARY RUN, at Simulation time=01/01 05:40 - 06:00
   **   ~~~   **  Enthalpy=-45222.26861 Pressure= 99500.00
   **   ~~~   **  Initial Resultant Temperature= -59.98
```

Nigusse:
@mjwitte OK.

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: Good catch. Yours truly left this hanging here some years back during the multilple airloops serving the same zone rewrite.

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: This may be fine when there's only one return node, but it won't work if there are multiple return nodes. `ExpTotalReturnMassFlow` is the sum of all return nodes serving the zone.

mjwitte: This kind of adjustment needs to happen below in the `// Adjust return flows if greater than expected....` block starting at line 4280. Maybe it needs a different method to adjust the return flow when `EnforceZoneMassBalance` is true?

mjwitte:
Looking at diffs for 5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized
Sizing diffs are small (e.g. cooling coil design load is 18618 vs 18620).
I see how some (all?) zones were not fully balanced before this fix, but it looks like the new method is ignoring the Zone Return Air Node 1 Flow Rate Fraction Schedule of 0.8.
But the exhaust fan clouds everything in that example file. So, here's an example without the exhaust fan. With develop, the returns flow are all 0.8 of the supply flow, and the mass balance infiltration rates are negative (exfiltration) which is what's expected. With this branch, the return flow are nearly equal to the supply flows.
[5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized-NoExhaustFan.idf.txt](https://github.com/NREL/EnergyPlus/files/5991237/5ZoneAirCooled_ZoneAirMassFlowBalance_Pressurized-NoExhaustFan.idf.txt)

Nigusse:
I did not plan to include the return air flow schedule modifier in the calculation. My understanding and @EnergyArchmage request was to not include any schedule modifier in the return air calculation. Instead, let the zone air flow balance equation determine the return air flow rate that balances the flow. Furthermore, if the zone air flow is balanced by adjusting the zone total return air flow, then the infiltration air flow rate is set to zero.

I ran the test file provided and the supply and return air flow rates are different. See the plots for zones SPACE1-1 and SPACE2-1. Am I looking at different example file?

SupplyAndReturnAirFlows

mjwitte:
But the total return flow is nearly equal to supply. And Space 5-1 return is much higher than supply (0.32 return vs 0.085 supply). This eliminates the possibility of modeling a pressurized system which was the whole point of adding the return node schedule and later the design return flow ratio on AirLoopHVAC.

Nigusse:
This is about the Psych error message thrown when running the defect file RAT_VAVSys_wZnMxYes_AddInf_woDiningExhFan. This defect file is a constant volume system with a variable speed DX cooling coil. 

```
   ** Warning ** Enthalpy out of range (PsyTsatFnHPb)
   **   ~~~   **  Routine=Unknown, Environment=JANUARY RUN, at Simulation time=01/01 05:40 - 06:00
   **   ~~~   **  Enthalpy=-45222.26861 Pressure= 99500.00
   **   ~~~   **  Initial Resultant Temperature= -59.98
```

I ran the defect file and confirmed the Psych error message shown above. This error message does exist in develop. So, it is introduced in this branch. But investigating led me to be suspicious about a code section in VS DX Cooling Coil. See the code snippet below and a link [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/VariableSpeedCoils.cc#L5470-L5483). 

```
        // calculate coil outlet state variables
        state.dataVariableSpeedCoils->LoadSideOutletEnth = state.dataVariableSpeedCoils->LoadSideInletEnth - state.dataVariableSpeedCoils->QLoadTotal / state.dataVariableSpeedCoils->LoadSideMassFlowRate;
        state.dataVariableSpeedCoils->LoadSideOutletDBTemp = state.dataVariableSpeedCoils->LoadSideInletDBTemp - state.dataVariableSpeedCoils->QSensible / (state.dataVariableSpeedCoils->LoadSideMassFlowRate * CpAir);

        MaxHumRat = PsyWFnTdbRhPb(state, state.dataVariableSpeedCoils->LoadSideOutletDBTemp, 0.9999, state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).InletAirPressure, RoutineName);
        MaxOutletEnth = PsyHFnTdbW(state.dataVariableSpeedCoils->LoadSideOutletDBTemp, MaxHumRat);
        if (state.dataVariableSpeedCoils->LoadSideOutletEnth > MaxOutletEnth) {
            state.dataVariableSpeedCoils->LoadSideOutletEnth = MaxOutletEnth;
            // QLoadTotal = state.dataVariableSpeedCoils->LoadSideMassFlowRate * (LoadSideInletEnth - LoadSideOutletEnth)
        }
        state.dataVariableSpeedCoils->LoadSideOutletHumRat = PsyWFnTdbH(state, state.dataVariableSpeedCoils->LoadSideOutletDBTemp, state.dataVariableSpeedCoils->LoadSideOutletEnth, RoutineName);
        if (state.dataVariableSpeedCoils->LoadSideOutletHumRat > MaxHumRat) {
            state.dataVariableSpeedCoils->LoadSideOutletHumRat = MaxHumRat;
        }
```
The above code section is evaluated regardless of the compressor PartLoadRatio value. Stepping through the  code confirms that the VS DX cooling coil outlet condition is calculated even when the PartLoadRatio is zero leading to very low temperature (e.g. -34C, -46C) when the flow rate is smaller. The very low cooling coil outlet node air temperature leads to Psych error. But what is interesting is that in the same code section a few line below [here](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/VariableSpeedCoils.cc#L5487-L5499) and also the code snippet shown below, the VS DX cooling coil outlet is set to coil inlet condition when the PartLoadRatio is zero and the supply fan operating mode is ContFanCycCoil. 

```
        if (CyclingScheme == ContFanCycCoil) {
            // continuous fan, cycling compressor
            state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).OutletAirEnthalpy = PartLoadRatio * state.dataVariableSpeedCoils->LoadSideOutletEnth + (1.0 - PartLoadRatio) * state.dataVariableSpeedCoils->LoadSideInletEnth;
            state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).OutletAirHumRat = PartLoadRatio * state.dataVariableSpeedCoils->LoadSideOutletHumRat + (1.0 - PartLoadRatio) * state.dataVariableSpeedCoils->LoadSideInletHumRat;
            state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).OutletAirDBTemp = PsyTdbFnHW(state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).OutletAirEnthalpy, state.dataVariableSpeedCoils->VarSpeedCoil(DXCoilNum).OutletAirHumRat);
            state.dataVariableSpeedCoils->PLRCorrLoadSideMdot = state.dataVariableSpeedCoils->LoadSideMassFlowRate;
        } else {
```

Therefore, I do not see a value executing this section of the code (the first snippet) to get the coil outlet condition when the PartLoadRatio is zero and the supply fan operating mode is ContFanCycCoil while allowing for CycFanCycCoil. I am proposing the following code modification to avoid running at a condition that leads Psych error:
```
        if ( (PartLoadRatio > 0.0 && CyclingScheme == ContFanCycCoil) || (CyclingScheme == CycFanCycCoil) ) {
            // calculate coil outlet state variables
            state.dataVariableSpeedCoils->LoadSideOutletEnth = state.dataVariableSpeedCoils->LoadSideInletEnth - state.dataVariableSpeedCoils->QLoadTotal / state.dataVariableSpeedCoils->LoadSideMassFlowRate;
            state.dataVariableSpeedCoils->LoadSideOutletDBTemp = state.dataVariableSpeedCoils->LoadSideInletDBTemp - state.dataVariableSpeedCoils->QSensible / (state.dataVariableSpeedCoils->LoadSideMassFlowRate * CpAir);

            MaxHumRat = PsyWFnTdbRhPb( state, state.dataVariableSpeedCoils->LoadSideOutletDBTemp, 0.9999, state.dataVariableSpeedCoils->VarSpeedCoil( DXCoilNum ).InletAirPressure, RoutineName );
            MaxOutletEnth = PsyHFnTdbW( state.dataVariableSpeedCoils->LoadSideOutletDBTemp, MaxHumRat );
            if ( state.dataVariableSpeedCoils->LoadSideOutletEnth > MaxOutletEnth ) {
                state.dataVariableSpeedCoils->LoadSideOutletEnth = MaxOutletEnth;
                // QLoadTotal = state.dataVariableSpeedCoils->LoadSideMassFlowRate * (LoadSideInletEnth - LoadSideOutletEnth)
            }
            state.dataVariableSpeedCoils->LoadSideOutletHumRat = PsyWFnTdbH( state, state.dataVariableSpeedCoils->LoadSideOutletDBTemp, state.dataVariableSpeedCoils->LoadSideOutletEnth, RoutineName );
            if ( state.dataVariableSpeedCoils->LoadSideOutletHumRat > MaxHumRat ) {
                state.dataVariableSpeedCoils->LoadSideOutletHumRat = MaxHumRat;
            }
        }
```
 
@mjwitte @rraustad @EnergyArchmage @shenb your thought?

Nigusse:
> But the total return flow is nearly equal to supply. And Space 5-1 return is much higher than supply (0.32 return vs 0.085 supply). This eliminates the possibility of modeling a pressurized system which was the whole point of adding the return node schedule and later the design return flow ratio on AirLoopHVAC.

I will modify the code to include a zone total return flow schedule value multiplier and see how it works.

shenb:
Niguesse,

I agree with you correction below.

The VS coil has a logic to return if
state.dataVariableSpeedCoils->LoadSideMassFlowRate <=0.
But this seems not adequate in the case when the PartLoadRatio == 0.0 along with a very small air flow rate.

Thanks for finding this out.

Bo

Nigusse:
@mjwitte Below is what I have found when added the zone total return air multiplied schedule and reverted the AdjustInfiltration code change.
 - Now the return flow rate is 80% times the supply air flow rates
 - The zone mixing flow rate becomes 0.0 since there is exhaust fan. The solution converges to zero mixing flow.
 - SPACE5-1 is now not balanced while SPACE1-1 is balanced
 - SPACE4-1 has exfiltration while SPACE1-1 through SPACE3-1 have 0.0 infiltration rate

Nigusse:
@shenb Good that you agree with the proposed change. I will post new issue to address it separately.

src/EnergyPlus/ZoneEquipmentManager.cc
mjwitte: I disagree with this. The mass balance should allow `MassConservation(ZoneNum).InfiltrationMassFlowRate` to be negative (or create a new variable for `MassConservation(ZoneNum).ExfiltrationMassFlowRate`) to allow a pressurized zone to be balanced.

Nigusse: The code change still allows negative infiltration flow rates (exfiltration flow rates). Rather this change simply sets the infiltration flow rate to zero if flow rates is smaller than the flow tolerance limits. Note that the code checks for "std::abs(ZoneInfiltrationMassFlowRate) > ConvergenceTolerance".

src/EnergyPlus/ZoneEquipmentManager.cc
Nigusse: I agree its the zone total return air flow and may need a different method of treatment. I tried a different approach but did not work and that is why I moved the code up in the loop such the flow adjustment catches it if needed. This is a work in progress.

src/EnergyPlus/ZoneEquipmentManager.cc
Myoldmopar: :+1:

src/EnergyPlus/ZoneEquipmentManager.cc
Myoldmopar: OK, well in an effort to try to get #8460 merged in for IO freeze, I'm going to table this discussion for right now, but I would like to see this discussed further on the next review call to make sure we all feel comfortable with the changes before 9.5 goes out.

src/EnergyPlus/ZoneEquipmentManager.cc
Myoldmopar: Another conversation that I will accept for the moment to keep things moving, but I want to make sure everyone is happy with this before 9.5 :smile: :smile: :smile:

Myoldmopar (commented):
OK, I am going to leave my review as a 'comment' here because I don't want to mark it approved.  But I think we'll get it merged in so we can get fresh results in #8460 and continue the IO Freeze push.  @mjwitte @Nigusse please plan on discussing this on next week's technicalities call so we can make sure we have it how we want it for release.

Myoldmopar:
Pulled in latest develop and ran everything fine.  I'm going to merge this, then pull develop into #8460 and let CI have a proper run with an updated branch there.

Nigusse:
I just saw this after pushing the latest local branch changes. I have been working to address Mike's concern and comments.

PR #8501, opened 2021-02-02
Pass correct mass flow rate to calculate flow fraction for DX coil performance curves

### LINKED ISSUE #8335: Airflow curve input defect in AirloopHVAC:UnitarySystem when MS coils operate at Speed = 1
Issue overview
--------------
Multispeed coil with a cycling fan should have a constant air flow curve input value per speed, while in the current simulation, the  input values varies in a linear relationship to node air flow rate for speed #1, but kept constant for speed #2. This happens when multispeed coil is modeled with an AirloopHVAC:UnitarySystem as a parent.

mscoil_defect
mscoil_defect2

### Details
 - Win10
 - E+ 9.4

### Checklist
 - [x] [Defect file added](https://drive.google.com/file/d/1eUHtkgH78go2QOBoS-qBADA6hGMFx62v/view?usp=sharing)

lgu1234:
Pull request overview
---------------------
Fixes #8335.

 The flow fraction performance curves at speed 1 in multiple speed DX cooling coil require actual mass flow rate as input to calculate performance modifiers. The mass flow rate should be independent of part load ratio. The fix provide correct actual mass flow rate.

### Pull Request Author

### Reviewer


lgu1234:
@yzhou601 @shorowit I upload the fix. Please review it.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -10260,11 +10260,7 @@ namespace UnitarySystems {
                 DataHVACGlobals::MSHPMassFlowRateLow = state.dataUnitarySystems->CompOnMassFlow * PartLoadRatio; // proportional to PLR when speed = 1,  #5518
             }
         } else if (this->m_DiscreteSpeedCoolingCoil && (state.dataUnitarySystems->CoolingLoad && CoolSpeedNum == 1)) {
```
yzhou601: @lgu1234 I tested with a multispeed air source heat pump test file, it seems that cooling curve inputs are fixed as expected, now airflow curve inputs for both speeds are constant, good job. However, I noticed that the heating curve inputs are showing the same issue at low speed. Can you look into heating as well?

Myoldmopar:
@lgu1234 is there an open issue for this, and also will you be addressing the comment from @yzhou601 about heating curves as well?

lgu1234:
@Myoldmopar Here is an open issue: https://github.com/NREL/EnergyPlus/issues/8335.

I noticed that heating curve inputs have the same problem during the fix. I prefer to verify the cooling curve fix and merge it. Then I will make a new issue to fix the heating curve inputs.

Myoldmopar:
In that case I guess that since @yzhou601 already confirmed the fix, this is ready for final review/merge?

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -10260,11 +10260,7 @@ namespace UnitarySystems {
...
-            if (this->m_FanOpMode == DataHVACGlobals::ContFanCycCoil) {
-                DataHVACGlobals::MSHPMassFlowRateLow = state.dataUnitarySystems->CompOnMassFlow; // #5737
-            } else {
-                DataHVACGlobals::MSHPMassFlowRateLow = state.dataUnitarySystems->CompOnMassFlow * PartLoadRatio; // proportional to PLR when speed = 1,  #5518
-            }
+            DataHVACGlobals::MSHPMassFlowRateLow = state.dataUnitarySystems->CompOnMassFlow;
```
Myoldmopar: I'm all for simplified code if this is indeed the proper flow rate to use.  I defer to @rraustad or @mjwitte who are more in tune with the intricacies of the different flow rates for unitary system calcs.  Thanks @lgu1234

lgu1234:
@Myoldmopar Yes. It is ready for final review.

Myoldmopar:
It looks like this is done.  @yzhou601 has confirmed the output changes.  I will build locally to test with latest develop.  @rraustad  or @mjwitte if you have any further comment on the changes here, go ahead and make them, otherwise I'll proceed toward merge.  Thanks!

tst/EnergyPlus/unit/UnitarySystem.unit.cc
```diff
@@ -10136,6 +10157,47 @@ TEST_F(EnergyPlusFixture, UnitarySystemModel_MultiSpeedCoils_SingleMode)
...
+                      sensOut,
+                      latOut);
+    EXPECT_NEAR(0.528942, thisSys->m_CycRatio, 0.0001); // cycling ratio
+    EXPECT_EQ(1, thisSys->m_CoolingSpeedNum);           // Speed number
+    EXPECT_NEAR(1.02,
+                state->dataCurveManager->PerfCurve(21).CurveInput1,
+                0.0001); // Speed 1 Total Cooling Capacity Function of Flow Fraction Curve input value
+    EXPECT_NEAR(1.02, state->dataCurveManager->PerfCurve(22).CurveInput1,
+                0.0001);                                               // Speed 1 Total EIR Function of Flow Fraction Curve input value
+    EXPECT_NEAR(0.4896, DataHVACGlobals::MSHPMassFlowRateLow, 0.0001); // cycling ratio
```
rraustad, line 10479: This unit test uses cycling fan operation so I would think the air flow rate would be proportional to PLR. The m_CycRatio = 0.528942 and speed 1 volume air flow rate = 0.4. I would expect the air mass flow rate to be around 0.24. Maybe it is but this unit test doesn't show the value. The MSHPMassFlowRate* variables are confusing. At speed 1, why isn't MSHPMassFlowRateLow = 0?

rraustad:
@Myoldmopar there are no plots to show how the air flow changed so I can't answer the question of why MSHPMassFlowRateLow needed to change. If I look at DXCoils at line 12279 I don't even see where MSHPMassFlowRateLow is used for speed 1 so I don't see how anything could change at speed 1. I do see where there were issue #'s at the place of the code change so I would wonder what the code fixed for those issues and how this change affects those issues. @lgu1234 can you add a plot of air flow and cycling ratio before and after so we can see how this change affected the coil model?

lgu1234:
@rraustad The object for this issue is to provide correct input values for flow function multiplier calculation at speed 1. Based on calculation requirements, it requires to pass mass flow rate at speed 1, which is independent of cycling ratio:

flow ratio = real mass flow rate / rated mass flow rate

Let me repeat it again, the flow ratio is independent of cycling ratio.

What I have done is to pass correct values for function calculation. I don't need to plot airflow and cycling ratios to prove my fix. The test file presents that the correct input values are passed after the fix.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
rraustad: The coil outlet mass flow rate = 0.25897 so that seems in line with what I would expect.

Myoldmopar:
@rraustad thank you for continuing to look over this.  I will defer to you for when you feel satisfied with the changes.  If you specifically need me to run anything, I will be happy to do it.

rraustad:
@lgu1234 there is a check list above in the PR description for the PR author. Also realize that it is hard for a reviewer to understand what a change means to the simulation so we may at times ask the author for some help.

I took 5518-in.idf and transitioned up to 9.4. It appears to have the correct air flow rate vs cycling ratio.

image

I assume #8335 was corrected with this change? If so, then a plot of the results so that we know it is correct now would be very helpful.

tst/EnergyPlus/unit/UnitarySystem.unit.cc
lgu1234: @rraustad The value of MSHPMassFlowRateLow is the mass flow rate when fan is on by definition, while the mass flow rate in the node is MSHPMassFlowRateLow * PLR.

The current setting requires MSHPMassFlowRateLow = MSHPMassFlowRateHigh at speed 1. If you would like to make MSHPMassFlowRateLow = 0, we may open a new issue for further discussion. The topic is beyond the fix.

lgu1234:
@rraustad Thanks for making a plot.

rraustad:
@lgu1234 you still need to verify that #8335 was fixed by providing a plot to compare with the description in that issue. That way we can see that this fix corrects that issue. See the Pull Request Author checklist above in the description. Those checkboxes needs to be filled in by the author, not the reviewer.

lgu1234:
@rraustad Two plots are provided with the same output variables after the fix.
Picutre1
Pinture2

rraustad:
@lgu1234 that looks really good. You still have to check off the boxes in the description. After that, @Myoldmopar I am OK with this change.

rraustad:
@Myoldmopar of course the reviewer will have to follow up on the close out of this PR (i.e., the reviewers check boxes).

lgu1234:
The check boxes were checked a while ago.

rraustad:
@Myoldmopar unit tests run here locally and the CI failures don't look related. Can you confirm?

rraustad:
@lgu1234 I am confused why this report variable is not in the csv?

Output:Variable,Cool-Cap-fFF1,Performance Curve Output Value,detailed; !- HVAC Average[]

rraustad:
The 8335 defect file isn't creating the eso file. I can't figure out why.

shorowit:
@rraustad Does it have ESO turned off via the `OutputControl:Files` object?

lgu1234:
@rraustad The csv file on my local computer has such outputs.

Pinture4

rraustad:
@lgu1234 thanks. I had another eplusout.csv file open when I ran the defect file and the csv was not created. Its working now.

rraustad:
Now I am confused. I ran the 8335 defect file in develop and this branch and get the same answer. Not really sure what to do in a situation like this. I guess I could try the unit test without the changes.

image

lgu1234:
It is better to check fan operation mode to ensure that an on/off fan is used.

shorowit:
> @rraustad Two plots are provided with the same output variables after the fix.
> Picutre1
> Pinture2

I’m confused, these before/after graphs plot different output variables.

rraustad:
I just figured out I was running the 1-speed defect file from 8335. That file should have been the same before and after. I am now running the 2-speed defect file which should show the problem in develop and correct operation in this branch. We'll see shortly.

rraustad:
Much better. The 2-spd defect file is the one that shows this was corrected.

image

rraustad:
@Myoldmopar except for my question on the CI results, my review is complete and this looks ready to merge.

Myoldmopar:
This is great!  Even though it isn't conflicted, I think with all the moving parts it's probably worth a local build with develop and running the test suite.  I'll do that and if all good then this should be mergeable directly.

Thanks @lgu1234 and @rraustad !

Myoldmopar:
Alright, built just fine with updated develop, no issues to resolve, and unit test suite passes.  Merging.  Thanks all!

PR #8508, opened 2021-02-04
Update dry air thermal conductivity correlation

### LINKED ISSUE #8504: Constant thermal conductivity of dry air below -20 deg. C
Issue overview
--------------
The thermal conductivity of dry air (used in airflow networks) is currently obtained using a correlation that uses a minimum outdoor air dry bulb temperature of -20 deg. C, below this temperature the conductivity is assumed to be constant. In cold climates, the outdoor air dry bulb is often below -20 deg. C. Sources such as [Kadoya et al.](https://aip.scitation.org/doi/10.1063/1.555744) show that the thermal conductivity of dry air is not constant below -20 deg. C.

### Details
 - Platform (Operating system, version): all
 - Version of EnergyPlus (if using an intermediate build, include SHA): 9.4

### Checklist

lymereJ:
Pull request overview
---------------------
 - Fixes #8504 
 - This pull request update the correlation used to calculate the thermal conductivity of dry air used by EnergyPlus in airflow networks. 

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/AirflowNetwork/src/Properties.cpp
+++ b/src/EnergyPlus/AirflowNetwork/src/Properties.cpp
@@ -59,15 +59,15 @@ namespace AirflowNetwork {
     )
     {
         // Dry air thermal conductivity {W/m-K}
-        // Correlated over the range -20C to 70C
-        // Reference Cengel & Ghajar, Heat and Mass Transfer. 5th ed.
+        // Correlated over the range -73C to 76.9C
+        // Reference doi/10.1063/1.555744
 
-        Real64 const LowerLimit = -20;
-        Real64 const UpperLimit = 70;
+        Real64 const LowerLimit = -73;
+        Real64 const UpperLimit = 76.9;
 
-        Real64 const a = 0.02364;
-        Real64 const b = 0.0000754772569209165;
-        Real64 const c = -2.40977632412045e-8;
+        Real64 const a = 0.02420195316;
+        Real64 const b = 0.0000766872;
+        Real64 const c = -4.4e-08;
 
         if (T < LowerLimit) {
             if (lowerLimitErrIdx == 0) {
--- a/src/EnergyPlus/AirflowNetwork/src/Properties.cpp
+++ b/src/EnergyPlus/AirflowNetwork/src/Properties.cpp
@@ -59,10 +59,10 @@ namespace AirflowNetwork {
     )
     {
         // Dry air thermal conductivity {W/m-K}
-        // Correlated over the range -73C to 76.9C
+        // Correlated over the range -73.1C to 76.9C
         // Reference doi/10.1063/1.555744
 
-        Real64 const LowerLimit = -73;
+        Real64 const LowerLimit = -73.1;
         Real64 const UpperLimit = 76.9;
 
         Real64 const a = 0.02420195316;
--- a/tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
+++ b/tst/EnergyPlus/unit/AirflowNetworkBalanceManager.unit.cc
@@ -4544,14 +4544,14 @@ TEST_F(EnergyPlusFixture, AirflowNetworkBalanceManager_AirThermalDiffusivity)
 
     Real64 const tol = 0.000001;
 
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, -30, 0.001, 101000), 1.578e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, -20, 0.001, 101000), 1.578e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 0, 0.001, 101000), 1.818e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 20, 0.001, 101000), 2.074e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 40, 0.001, 101000), 2.346e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 60, 0.001, 101000), 2.632e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 70, 0.001, 101000), 2.780e-5, tol);
-    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 80, 0.001, 101000), 2.780e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, -30, 0.001, 101000), 1.621e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, -20, 0.001, 101000), 1.621e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 0, 0.001, 101000), 1.869e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 20, 0.001, 101000), 2.132e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 40, 0.001, 101000), 2.408e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 60, 0.001, 101000), 2.698e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 70, 0.001, 101000), 2.848e-5, tol);
+    EXPECT_NEAR(AirflowNetwork::airThermalDiffusivity(*state, 80, 0.001, 101000), 2.848e-5, tol);
 }
 
 TEST_F(EnergyPlusFixture, AirflowNetworkBalanceManager_AirPrandtl)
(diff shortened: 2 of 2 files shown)


src/EnergyPlus/AirflowNetwork/src/Properties.cpp
```diff
@@ -59,15 +59,15 @@ namespace AirflowNetwork {
...
-        Real64 const UpperLimit = 70;
+        Real64 const LowerLimit = -73;
+        Real64 const UpperLimit = 76.9;
 
-        Real64 const a = 0.02364;
-        Real64 const b = 0.0000754772569209165;
-        Real64 const c = -2.40977632412045e-8;
+        Real64 const a = 0.02420195316;
+        Real64 const b = 0.0000766872;
+        Real64 const c = -4.4e-08;
```
lymereJ, line 70: The proposed correlation was obtained by fitting a linear regression to the values reported in Table 12 of [Kadoya et al.](https://aip.scitation.org/doi/10.1063/1.555744) between -~73 and +~77 deg. C. The figure below shows how the thermal conductivity of dry air differ from what EnergyPlus currently uses. Most of the differences occurs below -20 deg. C.

image

src/EnergyPlus/AirflowNetwork/src/Properties.cpp
```diff
@@ -59,15 +59,15 @@ namespace AirflowNetwork {
...
-        Real64 const UpperLimit = 70;
+        Real64 const LowerLimit = -73.1;
+        Real64 const UpperLimit = 76.9;
 
-        Real64 const a = 0.02364;
-        Real64 const b = 0.0000754772569209165;
-        Real64 const c = -2.40977632412045e-8;
+        Real64 const a = 0.02420195316;
+        Real64 const b = 0.0000766872;
+        Real64 const c = -4.4e-08;
```
lymereJ, line 70: The proposed correlation was obtained by fitting a linear regression to the values reported in Table 12 in [Kadoya et al.](https://aip.scitation.org/doi/10.1063/1.555744) between -~73 and +~77 deg. C. The figure below shows the difference between the proposed correlation and the one currently used by EnergyPlus. Most of the differences occurs below -20 deg. C.

image

lgu1234: @lymereJ The reference you cited was published in 2009. The previous reference, Yunus A. Çengel, Afshin J. Ghajar - Heat and mass transfer fundamentals and applications (2015 , Mc Graw-Hill Education), was published in 2015. I think latest publication should represent the latest progress.

jasondegraw: @lgu1234 I tend to agree with you. I find the graph to be particularly unpersuasive. Why only 4 points? That's not telling me a whole lot.

lgu1234:
When I search air thermal conductivity in the Internet, I find the website provides data: https://www.engineeringtoolbox.com/air-properties-viscosity-conductivity-heat-capacity-d_1509.html:
Air thermal conductivity

src/EnergyPlus/AirflowNetwork/src/Properties.cpp
lymereJ: @lgu1234, @jasondegraw - Thanks for the review. Here's a little bit more information.

Table 12 in Kadoya et al. provides the thermal conductivity of dry air for a wide range of temperatures (85K to 1200K) and pressures. Fitting a linear regression to the whole range of temperature could potentially increase the residuals within the temperature range that we typically see in the building simulations. The graph shown above shows the results of a regression fitted to subset of data points from Table 12 (-~73 and +~77 deg. C, at 0.1 MPa), this correlation is almost a perfect fit (R2 = 0.99). I can revise the regression to be based on more, or all points if need be.

Kadoya et al. was actually published in 1985, not in 2009.  Despite that, I found several very recent publications citing it as a reference (see [here](https://www.sciencedirect.com/science/article/pii/S0017931018308883), [here](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.701.1324&rep=rep1&type=pdf), or [here](https://www.sciencedirect.com/science/article/pii/S0020722520301841?via%3Dihub)). I do not have access to  Yunus A. Çengel, Afshin J. Ghajar - Heat and mass transfer fundamentals and applications (2015 , Mc Graw-Hill Education), you mentioned that it was published in 2015, but is the correlation provided in the book actually based on work done in 2015?

I found several other sources before landing on Kadoya et la. which were either older, had a too narrow range of temperature limits, or did not specify and limits. Here's a summary:

* ASHRAE Fundamentals, 2017: Figure 13 in Chapter 1 provides a chart of the thermal conductivity for temperatures between -50 deg. C to 300 deg. C but is based on older work by Mason and Monchick (1965)
* [Handbook of Physical Constants and Functions for Use in Atmospheric Boundary Layer Studies, Edgar L Andreas, Army Corps of Engineers, 2005](https://apps.dtic.mil/dtic/tr/fulltext/u2/a440352.pdf): Equation 11.1, is also based on older work by Hilsenrath et al. (1960)
* [Atiyeh Hoseini et al.](http://www.sfu.ca/~mbahrami/pdf/2017/Effects%20of%20humidity%20on%20thermal%20performance%20of%20aerogel%20insulation%20blankets.pdf): Equation 24, cites [this paper](https://www.sciencedirect.com/science/article/pii/S0017931007003614#bib36) which seems to refer to a 1992 publication in German (I don't speak German) and doesn't seem to provide any limits
* [Thermophysical and transport properties of humid air at temperature range between 0 and 100 °C](https://www.sciencedirect.com/science/article/pii/S0196890407003329): Equation 39, is only applicable between -23 deg. C and 777 deg. C
* Aerosol Technology, Properties, Behavior, and Measurement of Airborne Particles, Second Edition: Table 8, no temperature limits

I also came across [this presentation](https://stage.thermofluidprop.com/fileadmin/Dateien/KCE-Dateien/Slides_ASHRAE_Chicago_180122_2auf1.pdf) introducing results from ASHRAE RP-1767, see slide 17. I didn't think that using the picture provided on the slide and a tool like [WebPlotDigitizer](https://github.com/ankitrohatgi/WebPlotDigitizer) would provide good enough results. I did find out that I do have access the full ASHRAE RP-1767 report, it provides tabulated thermal conductivity of dry air for temperatures between -60 deg. C and 90 deg. C at atmospheric pressure at 1 degree interval. The table will be included in ASHRAE Fundamental 2021. This is probably the most recent data. The figure below compares this data, to Kadoya et al, and what EnergyPlus currently uses. It shows that both Kadoya et al. and ASHRAE Fundamentals 2021 are very similar.

image

Please, let me know if you want me to revise the proposed correlation, source, or if you wish to keep the current one and cancel this pull request. All sources that I reviewed, and that provided thermal conductivity for temperatures below -20 deg. C, seem to show that it is not constant below that value.

jasondegraw: @lymereJ Is this limitation of the correlation causing a failure? If it's not, then I'd prefer to leave it alone, especially since there's a possibility that CoolProp could be available for this and other properties soon. We can put in an issue and keep the branch until the CoolProp thing works itself out. If it is causing a failure, then I'd really prefer we use a correlation that we can just refer to rather than one we've developed ourselves. If there isn't one, then OK. As for the graph, you still need to provide some clue as to how many points you're showing.

Looks like my connection to our lab library is having an issue, so I can't see much in terms of the literature on this today, will have to try again Monday. I believe the current form was used with the expectation that it would not hit these temperatures too often. Is that not the case?

lymereJ: @jasondegraw - No, it's not causing failures, it is causing a lot of warnings. We've seen upwards of 350,000 warnings due to the limitation of the correlation in models run in climate zone 7. I'm not sure what is the impact on simulation results of assuming that the conductivity is constant below -20 deg. C versus what other correlation seem to show, @yanchenpnnl might know.

I'm afraid I don't know what CoolProp is, is it a new upcoming feature?

On the chart, the ASHRAE curve is based on 151 data points. They are all provided in Table B.1 of the ASHRAE RP-1767 RP report.

yanchenpnnl: @jasondegraw Hi Jason, so far, I did not see much energy difference (<1% changes). However, it generated a significant number of warnings (10000+ more) and took more computational effort.  I think what Jeremy proposed here makes sense, especially if you look at the plots, EnergyPlus is doing something weird (being saturated?) at the lower bound and upper bound, which seems incorrect.  

image

Regarding the book page you guys mentioned.  I also do not have a copy of the 2015 version, but I did find a 2010 version of a similar book by Gengel.  They are in IP units. and please see below.  If the 2015 book was using the same data source for air property, it then would refer to the original data published in 1970 (please see image below).  I did not see any saturated points (around -20C or 70C) from the data below. 

@jasondegraw @lgu1234  Let me know if this does not make sense.

jasondegraw: @yanchenpnnl I think you'll find that the "E" in that table title is significant.

There are two things here. First, we're having to guess what the underlying issue is. It would have been helpful to know that you were getting hundreds of thousands of warnings up front. Instead, all we got was "here's a different correlation". Second, the proposed correlation is different than the current correlation throughout the entire range. Adopting the proposed correlation changes results for every model using this part of AFN. Then if/when CoolProp gets adopted, all the results will change again. The worst case scenario is three successive versions with different results from the same model.

Had all of the information that we have now been included in the original issue, I would have told you that the best path forward would be to extend the correlation while leaving the current range as is. The code might look a little ugly (depending on the details), but it gives us at least a little stability. There'd only be diffs for those models that were going outside the -20C to 70C range. It may not be what you want, but given that this calculation could soon be replaced by something else entirely this is the best path forward.

yanchenpnnl: @jasondegraw  Haha, you are right!  Page 930 has the SI table~
Sorry that the (10,000+) warning issue was not written down in the first place of this PR.  It was mentioned in the email to Lixing and Mike earlier, and Jeremy did quite some investigations on getting the thermal property data, updated on the correlation, testing, and thus created this PR.  Yes, this issue is mostly only increasing the number of warnings.  So far, there is no observation of a big energy difference.  If the CoolProp gets adopted and solve this warning message in V9.5, I will agree with you that we do not need this PR.  I understand the potential of stability issue, so maybe if CoolProp did not get adopted in V9.5, we shall do an extension here, and leaving the current range as it is to avoid creating stability issue @lymereJ   

@jasondegraw BTW, comparing to the other PR on zone ERV and HPWH in AFN, this is less of a priority.  I will send more information about the other one, and definitely need your help to get that PR ready for V9.5.   Thanks!

jasondegraw: @yanchenpnnl It's up to you if you want to drop this, but fixing the warnings is still possible. Unless I'm mistaken, the current correlation does pretty well on the low temperature side down to much lower temperatures (when compared to the data from the textbook). I checked it with Excel, but you should check it too. If we set the limit to -50C or -70C it should only cause diffs for models that were getting warnings before. I'm not sure if the high temperature side is the same, but maybe that doesn't matter here.

lymereJ: > @yanchenpnnl I think you'll find that the "E" in that table title is significant.
> 
> There are two things here. First, we're having to guess what the underlying issue is. It would have been helpful to know that you were getting hundreds of thousands of warnings up front. Instead, all we got was "here's a different correlation". Second, the proposed correlation is different than the current correlation throughout the entire range. Adopting the proposed correlation changes results for every model using this part of AFN. Then if/when CoolProp gets adopted, all the results will change again. The worst case scenario is three successive versions with different results from the same model.
> 
> Had all of the information that we have now been included in the original issue, I would have told you that the best path forward would be to extend the correlation while leaving the current range as is. The code might look a little ugly (depending on the details), but it gives us at least a little stability. There'd only be diffs for those models that were going outside the -20C to 70C range. It may not be what you want, but given that this calculation could soon be replaced by something else entirely this is the best path forward.

@jasondegraw - The underlying issue has been provided in #8504 which is linked to this PR. The warnings are not the issue, they are results of the shortcoming of the current implementation. Actually, without them we would have never realized that the thermal conductivity is currently assumed to be constant below -20 deg. C. We proposed a different correlation which was developed based on data going down to -73.1 deg. C (or -60 deg. C for the one based on the ASHRAE RP-1767) which is a possible solution to the shortcoming of the current implementation. I understand that you are worried the proposed changes would lead to different results for all temperatures above -20 deg. C for models using AFNs. I am fine with your suggestion of extrapolating the current correlation, this will ensure that the physics is correctly represented which is really what #8508 and #8504 are about. @yanchenpnnl, is that fine with you too?

I just want to clarify that our intent was never to replace the current correlation because we thought that it was "inaccurate". I still don't know what CoolProp is (any information you could provide would be useful) and what it'll do/include, but @lgu1234 seem to imply that using the most recent work on the subject is important, and the ASHRAE RP-1767 (published in 2019) seems to be the most up-to-date source I could find.

Please let me know if you'd me to revise the current proposed changes to instead extrapolate the current correlation down to ... -70deg. C? Or, if you prefer to do it.

jasondegraw: @lymereJ We are now 10+ comments in on what could have been a one line fix. @yanchenpnnl has indicated that #8505 is more important, so my suggestion is to close this PR and focus on that one. Barring different feedback by COB today (2/8), I'll be clicking that button myself.

PR #8517, opened 2021-02-06
Windows package build

Myoldmopar:
Pull request overview
---------------------
 - Fixes package building on Windows, and Mac, which I didn't realize was broken...

Pretty much two things were changed:
 - On Windows, the MikTeX version we used was removed from the ctan mirror, which is silly, so I incremented it to the current version.  For reference, I'm in agreement with @jmarrec, we should just store our own version of the miktexsetup tool so we don't have to get burned every time they increment, but anyway.  Once I got the new version I found out the internal binary name had changed, so I needed to update that as well.
 - On Mac, the brew command needed a new form if you are installing from a cask.

Already verified that this makes packages again (https://github.com/NREL/EnergyPlus/releases/tag/9.5.0-TestBuildOnly4) so I'll merge this directly.


Myoldmopar:
This is only touching the github action workflow files, nothing we use in our CI testing, so I won't wait and waste Linux CI cycles.  Merging.

PR #8521, opened 2021-02-08
Reduce Alternate Build Configurations Built by Github Actions

mitchute:
Pull request overview
---------------------
 - Reduces the number of alternate build configurations built on Github Actions from 6 to 1
 - Eliminates duplicate base build which is already being built by DecentCI
 - Moves builds from Mac to Linux in hopes that GH Actions will be more responsive

### Pull Request Author

### Reviewer


Myoldmopar:
Yay!! It looks like it is still creating quite a few builds.  Having it as push and pull is fine, but I don't think we need both Ubuntu 18 and 20.  And I'm torn on whether we need the base build.  What do you think?

mitchute:
I was just going to make sure that the base build completes then we can turn it off. I also need to eliminate the Python install on the alternate build, but I'm waiting to make sure this all completes OK before making another commit. I'll also turn off the 18.04 build when I do that.

Myoldmopar:
I think you'll still need the Python install on the alternate build because CI looks for it just to do the build itself.  I think.  Hmm.  Yeah I think so.  OK, cool on the Ubuntu version as well.  This will be super helpful!  Thanks!

mitchute:
Hmm, maybe we just keep it building on either "push" OR "pull_request"? Turning on "pull_request" will turn on builds on PRs from forked repos but likely won't build anything in this repo until a PR is opened. Thoughts @Myoldmopar

mitchute:
Also, it appears that CMake was happy *without* a Python install. The current commit seems to be building OK so far.

mitchute:
** Without an explicit Python 3.7 install. It looks like 20.04 comes pre-installed with a version of 3.8, unless this is a stripped-down version. 

https://wiki.ubuntu.com/FocalFossa/ReleaseNotes#Python3_by_default

Myoldmopar:
I think keeping it like you have it right now is good.  Doing a PR build and a push build ensures we get results for all the commits we expect, but now we are down to just two builds per PR.  Quite reasonable to me!

mitchute:
Sounds good. Merge at will.

Myoldmopar:
I'll let everything finish up and if it's all green, it'll go right in.  Thanks for the quick effort getting all these runs cleaned up.

Myoldmopar:
Hey yo, look at all that green.  This is a great thing.  Thanks @mitchute!  Merging.

mitchute:
🤘

PR #8532, opened 2021-02-12
Update file version and IDD configuration and fix Construction:AirBoundary transition

Myoldmopar:
Pull request overview
---------------------
 - Updates version number to 9.5
 - Fixes IDD configuring for the build tree
 - Updates test files to 9.5
 - Fixes broken transition for Construction:AirBoundary

As usual, I expect the script to have missed a few files which will cause version warnings in the ERR files that I'll fix with a follow-up commit.


Myoldmopar:
EIO diffs are just the version change:

image

Two files were missed by the automatic version updater, they are now updated locally and I'll push.  Once the ERR diffs are gone I think this is ready to go.

Thanks for the reminder @mjwitte

Myoldmopar:
ERR diffs are gone, EIO and table diffs are just the version change.  I don't quite get why only two files are showing IDF diffs, probably because the test scripts ignore some special version and timestamp lines, but those files might have a different indentation or spacing or something.  But anyway, they definitely did change and everything looks all good here.  Building locally produces 9.4 and 9.5 IDD files in the build tree.  Anyone have concerns before this goes in later once CI is all happy?  @mjwitte do you want me to show anything from the changes?

mjwitte:
@Myoldmopar Yeah, I can't see what's different about the two files that are showing idf diffs. So, on a whim (or a premonition) I decided to test the new transition. Big fail. Fixes pushed up here. I can't imagine *who* would check in such broken code. Feel free to test and merge after CI finishes up.

Myoldmopar:
I'm not quite sure about the IDF diffs showing up, but it's not important to understand them right now.  This is good.  Merging.

PR #8541, opened 2021-02-16
Fix compiler warning on MSVC

Myoldmopar:
I updated CMake and MSVC on Windows to get the build running again.  Everything looks good but there is one warning being emitted that appears to be a complete false positive.  I wrapped this in a pragma which will hopefully hush up the compiler and get Windows all green again.


Myoldmopar:
We aren't the only ones who hit this on a recent update either: https://developercommunity.visualstudio.com/content/problem/961455/spurious-c4723-warning.html

Myoldmopar:
OK, it was still showing up, which was odd.  It's on me for assuming the pragma would work right there.  But anywho, I verified the warning on Windows with a release build, then moved the pragma to the function level and it muted it.  So I expect it to be all green here.

Myoldmopar:
There we go, all green on Windows!  Merging this.

PR #8556, opened 2021-02-22
Absorption chillerheater plant loop fixes

### LINKED ISSUE #8554: Exhaust (and gas fired) absorption chillerheater simulation fatal out on zero condenser flow rate
Issue overview
--------------
For the exhaust fired absorption chillerheater (ChillerHeater:Absorption:DoubleEffect), when a condenser flow is found to be zero during the simulation (or iteration), the program will report severe error and also throw fatal errors to terminate the simulation. The same issue also occurs on the gas fired absorption chillerheater (ChilerHeat:Absorption:DirectFired). 

While for other types of chiller models—such as electric chiller—this issue will still be reported as a severe warning, but does not cause a fatal-out and simulation termination. It will allow the simulation to continue—theoretically this gives the iteration loop a chance to correct the flow rate in later iterations if the zero flow is caused by numerical convergence issues. 

In general, if it is suggested to allow the simulation to continue, and possibly apply some mechanism to shut down the chiller (or take the chiller offline) when the condenser is zero. It is also expected to have all the chiller model types (electric and/or absorption) to have such a mechanism to deal with zero condenser flows in a consistent manner. 

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): V9.4, possibly earlier versions as well
 - Helpdesk ticket number 15605, and user direct email message

### Checklist
 - [x] Defect file added  to EnergyPlusDevSupport/DefectFiles

### LINKED ISSUE #8555: Exhaust or gas fired chillerheater plant loop branch check preventing pumps or other elements to be placed in front of the chillerheater
Issue overview
--------------
For the exhaust absorption chillerheater (ChillerHeater:Absorption:DoubleEffect), there is a "get design capacities" procedure in the chiller branch simulation, where a “hacky” way is used to determine what operation mode (or branch) which loop (chilled water, heating water, or condensing water) is being simulated for the absorption chillerheater. The same checking concept/procedure also occurs in the “simulation” module of the chillerheater.

However, this checking procedure is flawed somehow. As a direct (unexpected) outcome, it will prevent a pump to be placed in front of the exhaust absorption chiller. If a pump is placed in front of the chillerheater, the simulation program will fatal out while checking for this procedure, even though it is a very common pump configuration in many real simulation cases. But more generally, even a simple pipe placed in front of the chillerheater will be enough trigger such an unexpected fatal error.

The gas fired absorption chillerheater module (ChillerHeater:Absorption:DirectFired) also have similar issues. 

The fix would be to modify the checking procedure to accommodate more general branch setup scenarios.

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): v9.4, and possibly all versions.
 - Helpdesk ticket number 15605, and the user's direct email message

### Checklist
 - [x] Defect file added to EnergyPlusDevSupport/DefectFiles

jcyuan2020:
Pull request overview
---------------------
 - Fixes #8555 
 - Fixes #8554
 - The PR addressed an unexpected outcome (#8555) in the exhaust fired absorption chiller heater loop branch. The original code prevents another element--mostly typically a pump, but could be anything such as a simple pipe) to be placed in front of the exhaust absorption chillerheater. Otherwise, the simulation program will report fatal errors and terminate the simulation. The PR also addressed issue #8554 that when condenser flow rate becomes zero for whatever reasons, the program would report fatal error and terminate the program. With the fix in this PR, the program will use a similar treatment for other water cooled chiller models, in which it will still report severe errors but allow the simulation to continue. 
 - These changes are applied to both the exhaust-fired absorption chillerheater and the gas-fired absorption chillerheater.

### Pull Request Author

### Reviewer


src/EnergyPlus/ChillerGasAbsorption.cc
```diff
@@ -173,24 +197,47 @@ namespace EnergyPlus::ChillerGasAbsorption {
...
-        // kind of a hacky way to find the location of this, but it's what plantloopequip was doing
-        int BranchInletNodeNum =
-            state.dataPlnt->PlantLoop(calledFromLocation.loopNum).LoopSide(calledFromLocation.loopSideNum).Branch(calledFromLocation.branchNum).NodeNumIn;
+        bool matchfound(false);
+
+        int branchTotalComp = state.dataPlnt->PlantLoop(calledFromLocation.loopNum)
+                                  .LoopSide(calledFromLocation.loopSideNum)
+                                  .Branch(calledFromLocation.branchNum)
+                                  .TotalComponents;
+
```
jcyuan2020, line 206: The revised code will check all components in a branch, rather than just checking the branch head element as before.

src/EnergyPlus/ChillerGasAbsorption.cc
```diff
@@ -173,24 +197,47 @@ namespace EnergyPlus::ChillerGasAbsorption {
...
+        for (int iComp = 1; iComp <= branchTotalComp; iComp++) {
+            // kind of a hacky way to find the location of this, but it's what plantloopequip was doing
+            int compInletNodeNum = state.dataPlnt->PlantLoop(calledFromLocation.loopNum)
+                                       .LoopSide(calledFromLocation.loopSideNum)
+                                       .Branch(calledFromLocation.branchNum)
+                                       .Comp(iComp)
+                                       .NodeNumIn;
+
```
jcyuan2020, line 214: Still using the `hacky' way, but applied to all the components in the loop branch.

src/EnergyPlus/ChillerGasAbsorption.cc
```diff
@@ -173,24 +197,47 @@ namespace EnergyPlus::ChillerGasAbsorption {
...
+            } else if (compInletNodeNum == this->CondReturnNodeNum) { // called from condenser loop
+                MinLoad = 0.0;
+                MaxLoad = 0.0;
+                OptLoad = 0.0;
+                matchfound = true;
+                break;
+            } else {
+                matchfound = false;
+            }
+        }
```
jcyuan2020, line 237: If match found, update the loads values accordingly, break and return. Otherwise, continue to the next component check.

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -131,23 +131,47 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
...
     void ExhaustAbsorberSpecs::simulate(EnergyPlusData &state, const PlantLocation &calledFromLocation, bool FirstHVACIteration, Real64 &CurLoad, bool RunFlag)
     {
+        int brIdentity(0);
+
+        int branchTotalComp = state.dataPlnt->PlantLoop(calledFromLocation.loopNum)
+                                  .LoopSide(calledFromLocation.loopSideNum)
+                                  .Branch(calledFromLocation.branchNum)
+                                  .TotalComponents;
+
+        for (int iComp = 1; iComp <= branchTotalComp; iComp++) {
```
jcyuan2020, line 141: The revised code checks all the elements in an (exhaust or gas-fired) absorption chiller heat plant loop branch, to identity which (chilled water, hot water, or condensing water) branch it is simulating. This is in contrast to the original code where only head branch component is checked.

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -131,23 +131,47 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
...
+            } else if (compInletNodeNum == this->HeatReturnNodeNum) { // Operate as heater
+                brIdentity = 2; // heater
+                break;
+            } else if (compInletNodeNum == this->CondReturnNodeNum) { // called from condenser loop
+                brIdentity = 3; // condenser
+                break;
+            } else {
+                brIdentity = 0;
+            }
+        }
```
jcyuan2020, line 162: Use an integer number (`brIndentity`) to identify the loop branch property: 1 for chilled water; 2 for hot water; 3 for condensing water; 0 indicates problem (no match).

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -131,23 +131,47 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
...
 
-        // kind of a hacky way to find the location of this, but it's what plantloopequip was doing
-        int BranchInletNodeNum =
-            state.dataPlnt->PlantLoop(calledFromLocation.loopNum).LoopSide(calledFromLocation.loopSideNum).Branch(calledFromLocation.branchNum).NodeNumIn;
-
-        // Match inlet node name of calling branch to determine if this call is for heating or cooling
-        if (BranchInletNodeNum == this->ChillReturnNodeNum) { // Operate as chiller
+        if (brIdentity == 1) {
             this->InCoolingMode = RunFlag != 0;
```
jcyuan2020, line 165: After the current branch is identified, run the corresponding functions/code blocks to execute the simulation. Slightly reorganized from the original code.

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -131,23 +131,47 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
 
     void ExhaustAbsorberSpecs::simulate(EnergyPlusData &state, const PlantLocation &calledFromLocation, bool FirstHVACIteration, Real64 &CurLoad, bool RunFlag)
```
jcyuan2020, line 132: The changes are applied to both the Exhaust fired absorption chiller heater, and the gas fired absorption chiller heater. For each types of absorption chiller heater, the getDesignCapacities() and the simulate() functions would be check to identity the loop/branch type (chilled water, hot water, or condenser water).

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -1626,7 +1672,7 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
                 } else {
                     ShowSevereError(state, "CalcExhaustAbsorberChillerModel: Condenser flow = 0, for Exhaust Absorber Chiller=" + this->Name);
                     ShowContinueErrorTimeStamp(state, "");
-                    ShowFatalError(state, "Program Terminates due to previous error condition.");
+                    // ShowFatalError(state, "Program Terminates due to previous error condition.");
```
jcyuan2020, line 1678: Now do not show the fatal message and terminate any more--this is aligned with the kind of common treatment used in other type of chiller (electric or absorption) models.

src/EnergyPlus/ChillerExhaustAbsorption.cc
Myoldmopar: This is kinda odd.  If it is just an identifier and the value doesn't matter, this should be an `enum class`.  I would like to see that changed before this is merged.  I can make the change if needed, just let me know.

Myoldmopar (changes requested):
I think we should fix the magic numbers and make them an enum.  If you aren't able to make that change, I can do it, just let me know.

jcyuan2020:
> 
> 
> I think we should fix the magic numbers and make them an enum. If you aren't able to make that change, I can do it, just let me know.

Thanks! Updating it. Will test and push again.

src/EnergyPlus/ChillerExhaustAbsorption.cc
```diff
@@ -131,7 +131,14 @@ namespace EnergyPlus::ChillerExhaustAbsorption {
...
     {
-        int brIdentity(0);
+        enum brLoopType
+        {
+            CHILLER,
+            HEATER,
+            CONDENSER,
+            NOMATCH
+        };
+        brLoopType brIdentity(NOMATCH);
```
Myoldmopar: This is much closer, but not quite there.  I'm confused why this enum is repeated in the two cases.  It should be declared once.  Please:
- move this one enum declaration to either the top of the ChillerExhaustAbsorption.cc file, or in the header,
- make it an `enum class` instead of a raw `enum`,
- change the enum name from `camelCase` to `PascalCase`,
- change the member names from `SCREAMINGCASE` to `PascalCase`,
- and make the accompanying change for the `enum class` by using the namespace below (`BrLoopType::Chiller`)

jcyuan2020: > 
> 
> This is much closer, but not quite there. I'm confused why this enum is repeated in the two cases. It should be declared once. Please:
> 
>     * move this one enum declaration to either the top of the ChillerExhaustAbsorption.cc file, or in the header,
> 
>     * make it an `enum class` instead of a raw `enum`,
> 
>     * change the enum name from `camelCase` to `PascalCase`,
> 
>     * change the member names from `SCREAMINGCASE` to `PascalCase`,
> 
>     * and make the accompanying change for the `enum class` by using the namespace below (`BrLoopType::Chiller`)

I will make the changes. For why it was done twice in the two functions: actually these variables are very "local" to the simulate() functions so I did not want to bump up their scope. 
It might be a little bit debatable since it is just a type definition. Or we can make the scope an even higher level up, so that both the ChillerExhaustAbsorption.cc (or hh) and the ChillerGasAbsorption.cc (or .hh) can see the definition. 

Any comments or thoughts?

Myoldmopar: Wherever it makes sense is fine with me.  But duplicating the same list of enumarations in two places smells like it needs to be DRYd up.  For now, I say just move it up to the file level here so it can be shared between these two functions without duplication.  

And you are correct about the type declaration comment, the scope here doesn't matter since an enum is just optimized away quickly, so it won't affect runtime.  The scope should simply be determined by how many components this can be applied to.

jcyuan2020: So putting the definition in EnergyPlus/Plant/Enum.hh seems to be a good choice; then add `#inlcude<EnergyPlus/Plant/Enum.hh>` to both ChillerExhaustAbsorption.cc and ChillerGasAbsorption.cc.

jcyuan2020: Or definition in EnergyPlus/Plant/DataPlant.hh, which is already included in both both ChillerExhaustAbsorption.cc and ChillerGasAbsorption.cc.

jcyuan2020: See the new commits.

src/EnergyPlus/ChillerExhaustAbsorption.cc
Myoldmopar: Much nicer than the magic numbers, thank you!

src/EnergyPlus/ChillerExhaustAbsorption.cc
Myoldmopar: Hmmm.... OK, so if we show a severe then we need to fatal.  If we aren't going to fatal then this needs to be a warning.  But if it's a warning that could occur multiple times should we do a recurring warning message to avoid huge error files?

src/EnergyPlus/Plant/Enums.hh
```diff
@@ -189,6 +189,15 @@ namespace EnergyPlus::DataPlant {
...
     };
 
+    // branch loop type for absorption chillerheater models
+    enum class BrLoopType
+    {
+        Chiller,
+        Heater,
+        Condenser,
+        NoMatch
+    };
```
Myoldmopar, line 199: Yup, great!  In a future set of changes the actual type name might be tailored a little more to the actual application, but this is fine for now.

Myoldmopar (commented):
OK, I need one more reply back about the severe/warning/recurring comment.  I actually do think it's worth changing to at least a warning message, but I'm curious about the recurring change as well.  @jcyuan2020 what do you think?  Would this warning end up coming out in the error file many many times?  Or just once?  If it's just a one time sort of message then a regular warning is fine, and I can quickly make that change and get this merged.  If it needs to be a recurring message I may just push this back for that one more change from you.  

Or, I could have totally missed something and you can correct me.

Thanks for this effort, either way this is very nearly done.

src/EnergyPlus/Plant/Enums.hh
jcyuan2020: @Myoldmopar Thanks! Yes, the warnings are usually recurring. So it makes sense to condense them.

Myoldmopar:
My concern was still not fully addressed.  It looks to be like we have a ShowSevereError that now no longer fatals.  That is an invalid pattern -- even though yes there may still be a few of them lurking in the code.  If it is a severe error, then it requires a fatal.  If it is not going to fatal then it needs to be a warning.

I suspect it needs to be a warning.

Now, if it is a warning, it could occur many times in the course of a single simulation.  We don't want our error files to be thousands of lines long, so we need to instead use the ShowRecurringWarning style error messages so that we only print the first one plus a summary at the end.

Right now I think a user could end up with dozens?  thousands?? of severe errors in their error file, when it should've just been a recurring warning.

I don't think this is quite ready.

jcyuan2020:
> 
> 
> My concern was still not fully addressed. It looks to be like we have a ShowSevereError that now no longer fatals. That is an invalid pattern -- even though yes there may still be a few of them lurking in the code. If it is a severe error, then it requires a fatal. If it is not going to fatal then it needs to be a warning.
> 
> I suspect it needs to be a warning.
> 
> Now, if it is a warning, it could occur many times in the course of a single simulation. We don't want our error files to be thousands of lines long, so we need to instead use the ShowRecurringWarning style error messages so that we only print the first one plus a summary at the end.
> 
> Right now I think a user could end up with dozens? thousands?? of severe errors in their error file, when it should've just been a recurring warning.
> 
> I don't think this is quite ready.

Thanks for the comments! I thought the original question was whether this would be a recurring warning, and my answer was that these warnings could indeed be recurring: 
See here https://github.com/NREL/EnergyPlus/pull/8556#discussion_r588820527

As to whether the severe warnings should report fatal, currently this change is trying to make it similar to the way how other chiller and regular absorption chiller models handle such a stiuation---I think for those chiller models, it would be just report severe error but does not fatal out on it. 

So for the recurrent ones, do you want me change it to using the recurrent errors mode? I can make that change for the two absorption chillerheater models involved in this PR. 

BTW, if related, I think for the other chiller models (e.g. electric or regular absorption), the use of recurrent ones had not propagated in yet.

Myoldmopar:
@jcyuan2020 thanks for the further details.  I'm not excited about dropping this in with the severe-no-fatal and non-recurring errors, but indeed this is happening with the other objects, so this is not a widespread issue.  Given our timing I think I will merge this in after I finish testing it locally.

Myoldmopar (approved):

Myoldmopar:
This passes fine locally with develop pulled in.  I'll merge it now.  @jcyuan2020 please open an issue to address after the release where the severe-no-fatal code paths are cleaned up in relevant objects and also the messages should become recurring.  Thanks!

jcyuan2020:
@Myoldmopar Thanks! The issue has been posted---Issue #8638

PR #8570, opened 2021-02-26
Static lib ssc

Myoldmopar:
Pull request overview
---------------------
 - Changes SSC to build as a static library so that EnergyPlus can link up with it and we don't have to carry around another dll/so/dylib.  

It looks like it is building and working fine now, but I will still need to update the EnergyPlus packaging scripts to avoid trying to grab the now-nonexistent dylib on Mac.  

@nmerket these are very minor changes, my suggestion is to not directly backport these into SSC, but rather, investigate adding a proper CMake variable in SSC so that a tool wrapping it can switch it to build as a static library if the dynamics are not needed. 
 Then the EnergyPlus build process would simply switch on that flag and we'd be good to go.  Unless someone feels *very* strongly, I will plan on merging this soon as it is, since it is meeting our needs and cleaning up the installer, but am open to further refinements at any time.

FYI @jmarrec


Myoldmopar:
OK, so I think a push I'm about to make will wrap this up.  The SSC library is now built as a static library and linked at build time to EnergyPlus.  It is no longer packaged up for install since it is not needed.  We can have further discussions about how to port these changes back to SSC proper, and if there are better solutions, I'm all ears, but this will get it going for now.  As long as my local build finishes clean, and CI is still happy, this will go in.

Myoldmopar:
This change isn't going to affect anything that we test on CI, so there is literally no reason to wait for it and use up CI cycles.  I'm merging and closing this.

nmerket:
I'm happy with this. I pulled this change into my branch `li_ion_nmc2` and did a `git subtree pull` to get some changes from ssc and there were no merge conflicts.

PR #8572, opened 2021-02-27
Fix Thermal Comfort zone temperature and humidity lag

### LINKED ISSUE #8516: Fanger thermal comfort calculation misaligned with zone air (and other) temperature values
Issue overview
--------------
In calculating the Fanger thermal comfort values, such as PMV and PPD in a simulation run, the calculated PMV values look like if it is one time-step behind the corresponding temperature values (such as zone air temperature or zone radiant temperature). This is especially the case when the air temperature in a zone abruptly changes (e.g. when the thermostat setpoint changes). 

image

 
The cause of the problem is that the calculation module uses as a saved zone air temperature value for the previous time step, which causes the calculated PMV lagging behind. The calculated PMV value is thus off from what is “supposed” to be, depending on how quickly the zone air temperature changes. 

There is limited impact to the average users who do not need precise real-time thermal comfort information. Currently no workaround is available. The bug is expected to be fixed in future releases. 

### Details
 - Platform (Operating system, version) All
 - Version of EnergyPlus (if using an intermediate build, include SHA): 9.4, probably all
 - Helpdesk ticket number 15616

### Checklist

jcyuan2020:
Pull request overview
---------------------
 - Fixes #8516
 - Changes are made to allow the Fanger PMV and PPD calculation to use up-to-date temperature information, to make accurate (and correct) PMV and PPD calculations. 
 - The following results indicated how the issue was fixed: 
image
 
vs. the original results before the fix: 
image
(See #8516 as well)

  - Diffs: The common types of diffs that have been observed in many cases are the ``uncomfortable hours`` diffs (e.g. based on ASHRAE Standard 55), such as: 
 
image

These diffs could be expected as the PMV values would slightly change after the zone temperature and humidity corrections in the current PR. 

Since the uncomfortable hours are based on computed thermal comfort values (such as PMVs), a graphical comparison of the newly computed PMVs vs. the original PMVs is demonstrated in the following example: 
image
image
image

Additional diff comparisons: 
- Originally reported the psychometric relative humidity out of bound warning increase. After a code change these diffs are now eliminated. 
- No energy use related difference was observed. 

### Pull Request Author

### Reviewer


jcyuan2020:
New code changes are applied. Hopefully this should solve the increased counts of psychometric RH out of bound warnings.

jcyuan2020:
The psychometric relative humidity ratio out of bound warnings are finally gone--this is good as these warnings are not much expected. 
Now with the changes there are more cases having diffs---comparatively this is not that much unexpected. 
May also due to the fast changing upstream "develop" in the merging flurry--will try some local comparisons for this reason.

jmythms:
For the next reviewer:
I tested out some of the PMV csv diffs on this branch. This branch does better than how the current develop does in generating the PMV value.

For example, looking at the absolute csv diffs in the regression test for RadHiTempElecTermReheat.idf ([CI test results link](http://energyplus.s3-website-us-east-1.amazonaws.com/regressions/2021-03/5f676dbee178472c5a05097e88c8fb5b11644b0f-db8285f0f1546b9f3ebf0996e8edaa2f438bdd84/RadHiTempElecTermReheat/x86_64-MacOS-10.15-clang-11.0.0/)),
the largest PMV diff was 0.818 followed by a diff of 0.747 in the next timestep. 
image

This is how develop performs here:
image

This branch: 
image

As can be observed, the effect of the large jump in temperature on PMV happened one timestep later in the develop branch while it occurred in the same timestep in this branch. Similar big PMV diffs were observed in other regression tests, but could be justified with the temperature changes. 

Looks good!

Myoldmopar:
The Linux failed tests results are unrelated, but the diffs are present.  Would be good to get another set of CI results again.  @jmythms thanks for the results comparison, that helps a lot.  @mjwitte do you have anything further on this before it gets in line for final review?  Anything else you know you'd like to see from @jcyuan2020 ?

src/EnergyPlus/HVACManager.cc
```diff
@@ -243,8 +243,8 @@ namespace EnergyPlus::HVACManager {
 
         state.dataHeatBalFanSys->ZT = state.dataHeatBalFanSys->MAT;
         // save for use with thermal comfort control models (Fang, Pierce, and KSU)
-        state.dataHeatBalFanSys->ZTAVComf = state.dataHeatBalFanSys->ZTAV;
-        state.dataHeatBalFanSys->ZoneAirHumRatAvgComf = state.dataHeatBalFanSys->ZoneAirHumRatAvg;
+        // state.dataHeatBalFanSys->ZTAVComf = state.dataHeatBalFanSys->ZTAV;
+        // state.dataHeatBalFanSys->ZoneAirHumRatAvgComf = state.dataHeatBalFanSys->ZoneAirHumRatAvg;
```
mjwitte: @jcyuan2020 These lines should be deleted instead of commented, but otherwise, I'm good with this.

mjwitte (approved):

src/EnergyPlus/HVACManager.cc
jcyuan2020: @mjwitte The redundant commented code lines are now removed.

src/EnergyPlus/HVACManager.cc
```diff
@@ -491,6 +489,9 @@ void ManageHVAC(EnergyPlusData &state)
         FirstTimeStepSysFlag = false;
     } // system time step  loop (loops once if no downstepping)
 
+    state.dataHeatBalFanSys->ZTAVComf = state.dataHeatBalFanSys->ZTAV;
+    state.dataHeatBalFanSys->ZoneAirHumRatAvgComf = state.dataHeatBalFanSys->ZoneAirHumRatAvg;
+
```
Myoldmopar, line 494: A small change to make such an impact.  But it seems reasonable, and analysis of the diffs shows better results.  This is fine.

tst/EnergyPlus/unit/ThermalComfort.unit.cc
```diff
@@ -1053,3 +1053,524 @@ TEST_F(EnergyPlusFixture, ThermalComfort_CalcThermalComfortASH55)
...
+    CalcThermalComfortFanger(*state);
+
+    EXPECT_NEAR(state->dataHeatBalFanSys->ZTAVComf(1), 14.863733439268286, 0.001);
+
+    EXPECT_NEAR(state->dataHeatBalFanSys->ZoneAirHumRatAvgComf(1), 0.010564839505489259, 0.0001);
+
+    EXPECT_NEAR(state->dataThermalComforts->ThermalComfortData(1).FangerPMV, -5.5896341565108720, 0.001);
+
+    EXPECT_NEAR(state->dataThermalComforts->ThermalComfortData(1).FangerPPD, 100.0, 0.001);
+}
```
Myoldmopar, line 1576: OK

Myoldmopar (approved):

Myoldmopar:
Pulled in latest develop here and built and ran tests fine.  Code changes look good.  Merging.  Thanks @jcyuan2020 !

PR #8595, opened 2021-03-08
Fix VS2017 compile error

mjwitte:
Pull request overview
---------------------
Fixes compiler error with windows Visual Studio 2017.
```
\Develop\src\EnergyPlus\EvaporativeCoolers.cc(2804): error C2440: '<function-style-cast>': cannot convert from 'const T' to 'EnergyPlus::EvaporativeCoolers::OperatingMode'
        with
        [
            T=double
        ]
\Develop\src\EnergyPlus\EvaporativeCoolers.cc(2804): note: Conversions between enumeration and floating point values are no longer allowed  
```
Introduced in https://github.com/NREL/EnergyPlus/commit/9a2f824555804ae187d197dc69d3f2acc45e3ed1. VS2019 is fine with the current code.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/EvaporativeCoolers.cc
+++ b/src/EnergyPlus/EvaporativeCoolers.cc
@@ -2801,7 +2801,7 @@ namespace EvaporativeCoolers {
         auto &EvapCond(state.dataEvapCoolers->EvapCond);
 
         EvapCoolIndex = int(Par(1));
-        DryOrWetOperatingMode = OperatingMode(Par(2));
+        DryOrWetOperatingMode = OperatingMode(int(Par(2)));
         SysTempSetPoint = Par(3);
         EDBTSecAirSide = Par(4);
         EWBTSecAirSide = Par(5);


src/EnergyPlus/EvaporativeCoolers.cc
```diff
@@ -2801,7 +2801,7 @@ namespace EvaporativeCoolers {
         auto &EvapCond(state.dataEvapCoolers->EvapCond);
 
         EvapCoolIndex = int(Par(1));
-        DryOrWetOperatingMode = OperatingMode(Par(2));
+        DryOrWetOperatingMode = OperatingMode(int(Par(2)));
```
Myoldmopar, line 2804: This is fine...for now.  We need a targeted overhaul of the root-finding methods in EnergyPlus at some point.  The ways we packaged up a payload of data was fine in Fortran, but there are much better ways in C++.  In any case, I didn't actually realize we could even compile in VS2017 anymore, so if this is all it takes to get it compiling again, then so be it.  Thanks for the fix.

Myoldmopar (approved):

PR #8615, opened 2021-03-12
Global HVACController, RootFinder, and others

mitchute:
Pull request overview
---------------------
 - Moves the following to state:
   - Fans
   - DataReportingFlags
   - DataMoistureBalanceEMPD
   - DemandManager
   - HeatBalanceAirManager
   - HVACInterfaceManager
   - DataSurfaceLists
   - RootFinder
   - DataRootFinder
   - HVACControllers

### Pull Request Author

### Reviewer


tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -8074,8 +8074,8 @@ TEST_F(EnergyPlusFixture, VRFTU_CalcVRFSupplementalHeatingCoilWater)
     state->dataWaterCoils->WaterCoil.allocate(state->dataWaterCoils->NumWaterCoils);
     state->dataWaterCoils->WaterCoil(CoilNum).Name = thisVRFTU.SuppHeatCoilName;
     state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilType_Num = state->dataWaterCoils->WaterCoil_SimpleHeating;
-    state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilModel = state->dataWaterCoils->CoilType_Heating;
-    state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilType = state->dataWaterCoils->CoilType_Heating;
+    state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilModel = state->dataWaterCoils->CoilModel_Cooling;
+    state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilType = HVACControllers::iCoilType::Heating;
```
mitchute: @mjwitte I think it's worth drawing your attention to this particular change. I changed ```WaterCoilType``` here to use an enum value instead of const ints. In doing so, I found that in this unit test, ```WaterCoilModel``` was using the "CoilType" int const value instead of the "CoilModel" int const values at this location. In order to replicate the same behavior, ```CoilType_Heating``` which had a value of 2 had to be moved to ```CoilModel_Cooling```, which also has a value of 2. So now this unit test looks a little funny, with CoilType values being set to "Heating" and the CoilModel value being set to "Cooling".

I suppose the main question I have is whether you think this looks like just a quirk of this unit test, or if this points to a larger issue that needs to be investigated?

src/EnergyPlus/DataRootFinder.hh
```diff
@@ -54,124 +54,102 @@
...
-    // Type declaration for the root finder solution technique.
-
-    // Types
-
-    struct ControlsType
+namespace EnergyPlus::DataRootFinder {
+
+constexpr int iSlopeNone(0);        // Undefined slope specification
+constexpr int iSlopeIncreasing(1);  // For overall increasing function F(X) between min and max points
+constexpr int iSlopeDecreasing(-1); // For overall decreasing function F(X) between min and max points
```
mitchute, line 61: @amirroth @Myoldmopar These were not changed to an enum class because the values are used in error reporting messages, which frankly are not super useful. But that's what we've got. For example: https://github.com/NREL/EnergyPlus/blob/83402156ae773d8b87568ffac3f867fe68b1910b/src/EnergyPlus/RootFinder.cc#L244-L245 
I don't expect either of you to propose a solution at this time, but just noting FFR. We'll run into this as we push more on getting these converted over, so we could start thinking about how situations like this might be handled.

src/EnergyPlus/HVACInterfaceManager.hh
```diff
@@ -64,26 +64,14 @@ struct EnergyPlusData;
...
 
-    // Data
-    // MODULE PARAMETER DEFINITIONS:
     // Common Pipe Recirc Flow Directions
-    extern int const NoRecircFlow;
-    extern int const PrimaryRecirc;   // flow from Supply-outlet/Demand-inlet to Supply-inlet/demand-outlet
-    extern int const SecondaryRecirc; // flow from Supply-inlet/Demand-oulet to Supply-outlet/demand-inlet
+    constexpr int NoRecircFlow(0);
+    constexpr int PrimaryRecirc(1);   // flow from Supply-outlet/Demand-inlet to Supply-inlet/demand-outlet
+    constexpr int SecondaryRecirc(2); // flow from Supply-inlet/Demand-outlet to Supply-outlet/demand-inlet
```
mitchute, line 70: Same issue here, but this time due to the variables these are assigned to being used as reporting variables.

src/EnergyPlus/DataRootFinder.hh
amirroth: Thx.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
mjwitte: @mitchute No, that's bad.  Why is there a coil type enum class in `HVACControllers` now? 

Earlier in this same unit test a different coil type is set which is probably the source of the inconsistency.
https://github.com/NREL/EnergyPlus/blob/b7b900ecde5abcfdaa232726a85f5b2dd12590f0/tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc#L8058

And why does `dataWaterCoils->WaterCoil` have two different type variables, `WaterCoilType_Num` and `WaterCoilType`?

But you had to change `WaterCoilModel` to cooling? Now I'm really confused.

amirroth: Here's a high level comment. EnergyPlus and its API should deal with strings as little as possible. Enumerated values should not be strings, they should be enums. That's what the keyword `enum` means. The only things that should be strings are (dynamic) names. This is probably not the right time to do this sweep, but a sweep is needed.

mitchute: Let's see if I can recreate the chain of events that lead to this:
1. Change HVACController::CoilType* const ints to enum class.
2. Those are checked against ```iNodeType``` in ```HVACControllers::GetControllerInput``` which is set by ```WaterCoils::CheckActuatorNode```.
3. Internally, ```CheckActuatorNode``` sets ```iNodeType``` from ```WaterCoils::WaterCoil(coilNum).WaterCoilType```.
4. Changing ```WaterCoilType``` from ```int``` to ```HVACControllers::iCoilType``` then resulted in noticing this problem.

No idea why we have two different type variables.

mjwitte: <sigh> I see, `WaterCoilType` is only heating or cooling, `WaterCoilType_Num` is
```
        int const WaterCoil_SimpleHeating;
        int const WaterCoil_DetFlatFinCooling;
        int const WaterCoil_Cooling;
```
So, `WaterCoilType` could be combined with the other with some OR logic added where needed. FFR I guess.
But `WaterCoilModel` is essentially the same list of options?
```
        int const CoilModel_Simple; <-- which is used for WaterCoil_SimpleHeating
        int const CoilModel_Cooling;
        int const CoilModel_Detailed;
```
So, it looks like `WaterCoil(CoilNum).WaterCoilModel` and `WaterCoil(CoilNum).WaterCoilType_Num` are redundant and one or the other should be removed. At a minimum, they should bot work off the same enum class.

Myoldmopar:
I'll watch for this to get some CI results and if it's clean then this can go in next.

Myoldmopar:
It feels like the changes above are settled for now, in the context of this PR.  We certainly still have room to improve the code to further reduce the lists of constant ints, but the goal of this is to get rid of the data from the global struct, and leaving a few integers around, as long as they are constexpr, is fully reasonable for this stage.  This is a ton of work and is looking good.  Does anyone have anything further on this?  I'm building and doing a few tests locally to add onto what CI is already doing.

amirroth:
None from me, thanks.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
mjwitte: @Myoldmopar @mitchute We still have these lines in this unit test.
```
 state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilModel = state->dataWaterCoils->CoilModel_Cooling;
    state->dataWaterCoils->WaterCoil(CoilNum).WaterCoilType = HVACControllers::iCoilType::Heating;
```
What's the plan for fixing this?

amirroth: What is the semantic distinction between these two variables? Why are they both necessary?

mjwitte: They aren't really necessary. One has choices of "heating" or "cooling" and the other has choices of "heating", "cooling-simple" and "cooling-detailed". With a few logic changes the binary one could be removed. See https://github.com/NREL/EnergyPlus/pull/8615#discussion_r593356185

src/EnergyPlus/WaterCoils.hh
```diff
@@ -65,20 +66,27 @@ struct EnergyPlusData;
 
 namespace WaterCoils {
 
+    enum class iCoilModel
+    {
+        Unassigned,
+        Simple,
+        Cooling,
+        Detailed,
```
mjwitte: Never mind my earlier comment - my page wasn't refreshed.
But these enum names here should be "HeatingSimple" "CoolingSimple" and "CoolingDetailed", I'm fairly certain - haven't gone looking for their usage.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
Myoldmopar: Verified in GetWaterCoilInput where these are assigned: @mjwitte's suggested names make great sense.  I'll make that change alone here and this should be ready to go in.  Thank you all for the review!

amirroth: Alternatively replace all names with `auto` and let the compiler figure it out.

Myoldmopar:
OK, my commit went from:

```
    enum class iCoilModel
    {
        Unassigned,
        Simple,
        Cooling,
        Detailed,
    };
```

to 

```
    enum class iCoilModel
    {
        Unassigned,
        HeatingSimple,
        CoolingSimple,
        CoolingDetailed,
    };
```

I built and ran unit tests successfully.  I then ran a few hundred regressions with zero diffs of any kind.  This is ready to merge in without waiting on any CI cycles.  Thanks for the review everyone.  I do understand there is lots of additional valuable work to be made in the enumeration realm, and look forward to that work, but this PR successfully cleans out a bunch of global variables, which is awesome.

PR #8618, opened 2021-03-14
Global daylight manager and runtime language processor

Myoldmopar:
Encountered diffs when doing the external interface, so reverting that, at least for now.  Wanting CI to confirm that this is back to no diffs for now.


Myoldmopar:
The regression here is a bit elusive.  I bisected to find the commit, but I can't figure out what went awry.  So for now I pushed a new cleaner branch with that stuff out of there.  Should be no diffs.  Closing this PR.

PR #8619, opened 2021-03-14
Global Daylighting Manager, External Interface, Erl Processor

Myoldmopar:
Deglobalizes daylighting manager, external interface, and the runtime language processor.  This leaves some statics in the daylighting manager for right now as they were causing regressions when cleaned up.


src/EnergyPlus/DaylightingManager.hh
```diff
@@ -589,6 +657,60 @@ struct DaylightingManagerData : BaseGlobalStruct {
         this->TDDFluxTrans.deallocate();
         this->MapErrIndex.deallocate();
         this->RefErrIndex.deallocate();
+        this->MySunIsUpFlag = false;
+        this->CalcDayltgCoeffsMapPointsMySunIsUpFlag = false;
+        // seems like a reasonable initialization for the Vector variables
+        this->AR = 0.0;
```
Myoldmopar, line 663: These assignments reset the three vector coefficients to zero using `a=b=c=0` assignment.  But it really doesn't make a performance difference anyway because this is in the `clear_state` function which is only called in between unit tests or in between simulations in the API.

src/EnergyPlus/DaylightingManager.hh
```diff
@@ -543,6 +543,74 @@ struct DaylightingManagerData : BaseGlobalStruct {
     Array2D_int MapErrIndex;
     Array2D_int RefErrIndex;
 
+    bool MySunIsUpFlag = false;
+    bool CalcDayltgCoeffsMapPointsMySunIsUpFlag = false;
+
+    // static variables extracted from functions
+    Vector3<Real64> AR;  // Inside surface area sum for floor/wall/ceiling (m2)
```
Myoldmopar, line 550: I think a much better design would be to have an actual DaylightingManager object that holds data and persists it, much like the input processor, SQLite manager, or plugin manager.  But this is fine for now to at least get these statics off of the global state.

Myoldmopar (commented):
This one should be ready to go.

Myoldmopar:
OK, the only conflict was a tiny change in the state manager.  I resolved that and built locally.  I ran the full unit test suite and it is still passing.  I ran the full regression suite and got no diffs. This is in agreement with what CI was producing before pulling in develop.  I'm going to rely on that set of results and not waste CI cycles.  This is going to merge.

PR #8633, opened 2021-03-18
Fix clear_state missing vars in pollution module

Myoldmopar:
Pull request overview
---------------------
 - Fixes an API test failure.  The API test in question would purposefully run E+ via API, clear the state, then run it again.  This failure revealed that a couple variables were not being reset in a clear_state method.

Quick fix, to be merged directly.

### DIFF
--- a/src/EnergyPlus/PollutionModule.hh
+++ b/src/EnergyPlus/PollutionModule.hh
@@ -480,6 +480,10 @@ struct PollutionModuleData : BaseGlobalStruct {
 
     void clear_state() override
     {
+        this->PollutionReportSetup = false;
+        this->GetInputFlagPollution = true;
+        this->NumEnvImpactFactors = 0;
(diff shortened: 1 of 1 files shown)


src/EnergyPlus/PollutionModule.hh
```diff
@@ -480,6 +480,10 @@ struct PollutionModuleData : BaseGlobalStruct {
 
     void clear_state() override
     {
+        this->PollutionReportSetup = false;
+        this->GetInputFlagPollution = true;
+        this->NumEnvImpactFactors = 0;
+        this->NumFuelFactors = 0;
```
Myoldmopar, line 486: Just reinitializing these during the clear_state.

Myoldmopar:
Not worth waiting on CI here, I'll check develop once CI gets around to it, and it should be cleaned out there.

PR #8636, opened 2021-03-20
Globals hvac globals

Myoldmopar:
Pull request overview
---------------------
 - Cleans out all the global variables from DataHVACGlobals except a few const string arrays that I'll address soon.


Myoldmopar:
I built this locally, ran all tests, and regressions, with no issue whatsoever.  I'm merging this directly.  Thanks @Myoldmopar

amirroth:
I'm not sure what you are planning to do with the const string arrays, but my suggestion is to convert them to `constexpr const char *array[3] = {"string1", "string2", "string3"};`. This way no constructors are needed, you can stick them into the .hh files, and they show up just once in the static text segment (I ran a little test). Now, you will have to access them with `[]` rather than `()`, but in my opinion that is a good thing!

Myoldmopar:
I fully agree, that is the best possible path.  And as long as they are just used as simple lookups, then that's very easy to accomplish, and we can even put a dummy value in the zeroth item so we don't have to adjust array index values right now.  The only problem is if the list is passed into a utility function like "FindItemInList", in which case the list obviously needs to be an actual object, either vector or Array.  For those cases, we just have to decide, do we:

1. move the Array1D object into state, which is usually a 1 minute change, but adds a small object into state that really doesn't need to be there, or
2. change all the usages to be able to search over a `constexpr const char *`, which is better for state, but might require 10-15 minutes of refactoring and testing to make sure it's working well, and possibly more prone to a small regression to investigate.

And yes, moving from `()` to `[]` is a fantastic thing!

amirroth:
Why do you say that the list has to be an object in order to be passed into a function like `FindItemInList`? Because of the current signature of `FindItemInList`? Because of the need to pass the length or the positions of the first and last valid elements? Those could be passed separately. That's what we did in C, and we liked it!

Anyways, you can do whatever is expedient for now in the name of deglobalizing all of EnergyPlus for 9.5. I was just wondering what the ultimate plan here is.

Myoldmopar:
You are correct, it's because we'd need to have a separate FindItemInList.  Definitely not a problem, just something we'd have to do...and we'd like it!

PR #8642, opened 2021-03-22
Deglobalizing EnergyPlus

Myoldmopar:
Pull request overview
---------------------
 This was one of the big remaining ones.  Just because it's in nearly every file.  The DataIPShortCuts.  I had hoped to refactor it at the same time to eliminate the use of the global variables, but after a couple hours I realized this was a terrible idea right now.  The DataIPShortcuts are used in very fun ways throughout the code.  This way it does the same old deglobalization, but should show no diffs.  It is clean locally on ~100 regression tests.


Myoldmopar:
It looks like there could be a very tiny VRF err output diff, which I have now fixed in my next commit that I'll push soon.

Myoldmopar:
Ended up deglobalizing a lot more modules here:
- DataIPShortcuts
- heat bal finite diff
- data hvac controllers
- ground temp model manager
- data global constants
- hvac cooled beam
- hvac dx assisted coil
- HWBaseboardRadiator
- electric baseboard radiator
- humidifiers
- dataoutputs
- hvac dx system
 
I'll fix up the test issues and get this cleaned up shortly.

Myoldmopar:
The debug failures are all fixed up.  Windows builds are interesting, I'm getting a memory access violation in the RE2 destructor as it tries to clean up.  EnergyPlus has already completed successfully, but when the destructor executes in RE2 it hits a bad memory spot.  I'm bisecting on Windows now.  Hopefully clear it up shortly.

Myoldmopar:
OK, I verified the Windows issue locally and was able to isolate which commit caused the problem and I reverted it.  I'll push it and see if Windows agrees.

Myoldmopar:
OK, not sure when CI is going to get around to my commit.  I feel pretty confident after being able to reproduce the issues locally and getting them clean.  Given the timing, I'm going to merge this in so we can move on.

Myoldmopar:
Yes, Windows was happy, now as long as integration is happy too then I made the right decision to merge this!

PR #8649, opened 2021-03-24
Grounding EnergyPlus (removing the static), Cont'd

Myoldmopar:
Pull request overview
---------------------
Removes statics from a bunch of modules.  Builds and runs fine, no regressions.  Merging directly.


Myoldmopar:
Merging.  If something terribly wrong happens I can revert, but I've tested it pretty heavily and it should be A OK.

PR #8652, opened 2021-03-24
Continued Grounding of EnergyPlus

Myoldmopar:
Pull request overview
---------------------
Removes static from a bunch more files, check the commit log to see which ones.


Myoldmopar:
Builds fine, unit tests run fine, no regressions, merging right away to keep things moving.

PR #8663, opened 2021-03-26
Reduce dynamic allocations

lefticus:
Pull request overview
---------------------

This is a small code refactoring based on testing the tool heaptrack against EnergyPlus. With these small changes we are able to reduce dynamic allocations by up to about 87%.

5b41a64a535c3489378d6e50386fa6007d6ccb54 has the full description of what was found and what future recommendations are.

Copied here for convenience 

---------

File tested: BenchmarkHospitalNew_USA_CA_SAN_FRANCISCO.idf

 * Dynamic allocations reduced from 38,231,848 to 5,569,555

Moving from `static const std::string RoutineName` to
`constexpr auto RoutineName` is a good idea, however it ended up
resulting in extra conversions from `const char *` to `std::string`
which is silent.

`static const std::string` has in inherent cost in that each
time the function is entered a thread-safe check must be performed.

Similarly, constructing a new `std::string` from a `const char *`
is going to have potential dynamic allocation cost, which is why
we did not notice a performance degredation when the change was made.

This patch set

 * moves from `constexpr auto RoutineName` to
   `constexpr static std::string_view RoutineName`
 * avoids automatic conversions to `std::string` because that
   does not exist for `std::string_view`
 * similarly to how `const` makes you think about "`const`-correctness",
   `std::string_view` pushes a further set of changes

Because of the move to `std::string_view` we needed to update the
function signatures of a few functions that do logging with
RoutineName.

Because of the move to `std::string_view`, which does not have a `+`
operator, we then needed to modernize some of the logging function
calls to use `fmt::format`.

For any future update of `const static std::string`, I recommend
a move to `constexpr static std::string_view` and updating
all downstream code to avoid as many dynamic allocations
as possible.

The benefit is clear, with a measurable performance improvement
and potentially drastic decrease in dynamic allocations.


lefticus:
5b41a64a535c3489378d6e50386fa6007d6ccb54 has the full description of what was found and what future recommendations are.

Copied here for convenience 

---------

File tested: BenchmarkHospitalNew_USA_CA_SAN_FRANCISCO.idf

 * Dynamic allocations reduced from 38,231,848 to 5,569,555

Moving from `static const std::string RoutineName` to
`constexpr auto RoutineName` is a good idea, however it ended up
resulting in extra conversions from `const char *` to `std::string`
which is silent.

`static const std::string` has in inherent cost in that each
time the function is entered a thread-safe check must be performed.

Similarly, constructing a new `std::string` from a `const char *`
is going to have potential dynamic allocation cost, which is why
we did not notice a performance degredation when the change was made.

This patch set

 * moves from `constexpr auto RoutineName` to
   `constexpr static std::string_view RoutineName`
 * avoids automatic conversions to `std::string` because that
   does not exist for `std::string_view`
 * similarly to how `const` makes you think about "`const`-correctness",
   `std::string_view` pushes a further set of changes

Because of the move to `std::string_view` we needed to update the
function signatures of a few functions that do logging with
RoutineName.

Because of the move to `std::string_view`, which does not have a `+`
operator, we then needed to modernize some of the logging function
calls to use `fmt::format`.

For any future update of `const static std::string`, I recommend
a move to `constexpr static std::string_view` and updating
all downstream code to avoid as many dynamic allocations
as possible.

The benefit is clear, with a measurable performance improvement
and potentially drastic decrease in dynamic allocations.

Myoldmopar:
This is so cool.  I second the question from @amirroth, and I believe you have mentioned that indeed we should be using arrays of string views, but it would be good to lock that down here.  The number of dynamic allocations removed is clearly a huge win.  

The branch has got lots of conflicts, do you feel like it is worth going back through and resolving them as conflicts or is it easy enough to just reapply the changes?  I'm happy to resolve them as conflicts if that's the best approach rather than spending your time on that.

amirroth:
@Myoldmopar I did a mini test and `constexpr std::array<std::string_view>` looked good, i.e., no constructor/destructor. In fact, it did an optimization that I did not think it could do.

lefticus:
I missed Amir's question originally, but yeah, it's a compile-time constant. `constexpr std::array<std::string_view>` can be almost magical, but it is technically possible to do a little better.

To do better, you have to encode the length of the string into the type, then you end up with something like:

```
constexpr std::tuple<static_string<5>, static_string<7>, static_string<29>> data;
```

But this is much more difficult to work with, because you cannot do runtime indexing into a tuple (since each element has its own type, it's not stored like a normal array, it's more like elements of a struct).

Myoldmopar:
I will soon resolve the conflicts here unless @lefticus has any local changes already in motion.  Or if anyone has any reason to hold it up.  I'm really excited about this change though.

scripts/dev/apply_formatting.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
 
-find src tst -name "*.hpp" -o -name "*.h" -o -name "*.hh" -o -name "*.cc" -o -name "*.cpp" -o -name "*.c" | xargs -n 1 -I '{}' clang-format -i "{}"
+find src \( -name "*.hpp" -o -name "*.h" -o -name "*.hh" -o -name "*.cc" -o -name "*.cpp" -o -name "*.c" \) | xargs -n 1 -P 10 -I '{}' clang-format -style=file -i "{}"
+find tst \( -name "*.hpp" -o -name "*.h" -o -name "*.hh" -o -name "*.cc" -o -name "*.cpp" -o -name "*.c" \) | xargs -n 1 -P 10 -I '{}' clang-format -style=file -i "{}"
```
Myoldmopar, line 4: Nevermind this for now, I need to tweak the command anyway.  Nothing related to this work.

src/EnergyPlus/Boilers.cc
```diff
@@ -161,7 +161,7 @@ void GetBoilerInput(EnergyPlusData &state)
     // standard EnergyPlus input retrieval using input Processor
 
     // Locals
-    constexpr auto RoutineName("GetBoilerInput: ");
+    static constexpr std::string_view RoutineName("GetBoilerInput: ");
```
Myoldmopar, line 164: These minimal changes to string views and such a great impact.

src/EnergyPlus/CurveManager.cc
```diff
@@ -2802,10 +2802,10 @@ namespace CurveManager {
...
                         int const CurveIndex,
                         std::vector<int> validDims,
-                        std::string routineName,
-                        std::string objectType,
-                        std::string objectName,
-                        std::string curveFieldText)
+                        std::string_view routineName,
+                        const std::string &objectType,
+                        const std::string &objectName,
+                        const std::string &curveFieldText)
```
Myoldmopar, line 2808: Clang tidy warnings about const arguments in headers were cleaned up here while I was resolving conflicts.

src/EnergyPlus/Psychrometrics.cc
```diff
@@ -1005,7 +1005,7 @@ namespace Psychrometrics {
                 if (state.dataPsychrometrics->iPsyErrIndex(iPsyTsatFnHPb) == 0) {
                     ShowWarningMessage(state, "Enthalpy out of range (PsyTsatFnHPb)");
                     if (!CalledFrom.empty()) {
-                        ShowContinueErrorTimeStamp(state, " Routine=" + CalledFrom + ',');
+                        ShowContinueErrorTimeStamp(state, format(" Routine={},", CalledFrom));
```
Myoldmopar, line 1008: Added in commas here in a couple spots that were showing ERR diffs.  They should be all clean now.

Myoldmopar (approved):
OK, got the conflicts resolved here, this is in great shape.  Everything passes locally, including regressions.  I'll wait until CI, including GIthub Actions, gets done with the commit, and get this merged in.  Thanks @lefticus!

Myoldmopar:
All super green, ready to go in.  Thanks @lefticus !

PR #8672, opened 2021-03-30
Apply clang format

Myoldmopar:
### Summary 

Applies clang-format to the src/EnergyPlus and tst/EnergyPlus/unit folders.

For the record, it was applied using:
```
$ find . | grep -E ".*(\.cc|\.cpp|\.c|\.h|\.hpp|\.hh)$" | xargs clang-format -i --style=file
```

So it gets all the c, cc, cpp, h, hh, and hpp files recursively in both folders.

### Reason

This was requested by an important partner who uses a fork of EnergyPlus code.  Their tools use clang-format much like our tools should, but our code has gotten out of sync and a bunch of it is not using the prescribed format in our `.clang-format` files.  I will work on our style validation script to make sure it is properly checking the commits moving forward and have it give warnings if it finds anything invalid to help developers keep the code up to the proper style (I'm looking at you @Myoldmopar).  

### Effect

This has a lot of file diffs.  The primary reason is because modern C++ prefers us to change from:

```
namespace EnergyPlus {

namespace MyCoolModule {
```

to

```
namespace EnergyPlus::MyCoolModule {
```

anytime the contents of the namespace can be collapsed directly.  However, Clang-format gets a little confused here.  Our style settings say that we should _not_ indent code that is only one namespace deep.  This is a great style because for most libraries, you will have one big namespace, and you don't want to have your entire codebase indented just for the sake of your own single namespace.  But then if you start nesting namespaces, the code should become indented to reflect that.  In our case, the two namespace keywords told Clang-Format to apply one level of indentation to the contents.  But when we move that to a single nested statement with `::`, Clang-Format thinks it is a single namespace, so it pulls the indentation back out.  

I actually prefer this, and I think most people will.  Because if you are specifically looking at file contents where there is only one namespace, even if it is nested like `EnergyPlus::Whatever`, there's really nothing to gain from having the whole contents indented.  With this style properly applied, you get the lines closer to the line numbers and more real estate for great things like `state.dataGlobalsData->dataGlobalStateVariable1;`.  OK, OK, that's a little jab at myself and we already have plans in motion to clean that up as well.  Good things are happening, don't worry.

### Conclusion

If you have an open PR, and you don't want to deal with the conflicts, just let me know!  I will be happy to apply Clang-Format to your branch and then do the merge, or at least help significantly with the merge, as long as it is after the release.  Once I get our custom checks doing proper validation of the style on our Github Action tests, this should never happen again as the code won't be able to get out of sync without warnings popping up.

Thanks all!


Myoldmopar:
Once Linux finishes I'll merge this in.  I'm currently preparing my final branch for the 9.5 release now.

PR #8674, opened 2021-03-30
Storm window refactoring

xuanluo113:
Pull request overview
---------------------
This branch refactors the logic of checking active structure and shaded structure of window surfaces when there is a storm window applies to the surface.

### Pull Request Author

### Reviewer


src/EnergyPlus/DataSurfaces.hh
```diff
@@ -1584,6 +1581,9 @@ struct SurfacesData : BaseGlobalStruct
     Array1D<Real64> SurfWinSpecTemp;        // The specification temperature of the TC layer glass Added for W6 integration June 2010
     Array1D<Real64> SurfWinWindowModelType; // if set to WindowBSDFModel, then uses BSDF methods
     Array1D<Real64> SurfWinTDDPipeNum;      // Tubular daylighting device pipe number for TDD domes and diffusers
+    Array1D<int> SurfWinStormWinConstr;                  // Construction with storm window (windows only)
+    Array1D<int> SurfWinActiveConstruction;            // The currently active construction with storm window (windows only)
+    Array1D<int> SurfWinActiveShadedConstruction;      // The currently active shaded construction with storm window (windows only)
```
mjwitte: These comment descriptions seem wrong. If I'm following correctly
```
 SurfWinActiveConstruction;            // The currently active unshaded construction with or without storm window (windows only)
SurfWinActiveShadedConstruction;      // The currently active shaded construction with or without storm window (windows only)

File: src/EnergyPlus/HeatBalanceSurfaceManager.cc [hunk 2]
```diff
@@ -2856,8 +2855,7 @@ void InitSolarHeatGains(EnergyPlusData &state)
             int const firstSurfOpaq = state.dataHeatBal->Zone(zoneNum).OpaqOrIntMassSurfaceFirst;
             int const lastSurfOpaq = state.dataHeatBal->Zone(zoneNum).OpaqOrIntMassSurfaceLast;
             for (int SurfNum = firstSurfOpaq; SurfNum <= lastSurfOpaq; ++SurfNum) {
-                int ConstrNum = Surface(SurfNum).Construction;
-                if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1) ConstrNum = Surface(SurfNum).StormWinConstruction;
+                int ConstrNum = state.dataSurface->Surface(SurfNum).Construction;
```
mjwitte: Should be? `int ConstrNum = state.dataSurface->SurfWinActiveConstruction(SurfNum);`

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3891,13 +3881,10 @@ void InitIntSolarDistribution(EnergyPlusData &state)
             state.dataHeatBal->SurfInitialDifSolInAbsReport(SurfNum) = 0.0;
             state.dataHeatBal->SurfWinInitialDifSolInTransReport(SurfNum) = 0.0;
             state.dataHeatBal->SurfWinInitialDifSolInTransReport(SurfNum) +=
-                state.dataHeatBalSurf->SurfWinInitialDifSolInTrans(SurfNum) * Surface(SurfNum).Area;
+            state.dataHeatBalSurf->SurfWinInitialDifSolInTrans(SurfNum) * Surface(SurfNum).Area;
```
mjwitte: Formatting?

src/EnergyPlus/SolarShading.cc
```diff
@@ -8554,7 +8542,7 @@ void CalcAbsorbedOnExteriorOpaqueSurfaces(EnergyPlusData &state)
                 continue;
             int ConstrNum = state.dataSurface->Surface(SurfNum).Construction;
             if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1) {
-                ConstrNum = state.dataSurface->Surface(SurfNum).StormWinConstruction;
+                ConstrNum = state.dataSurface->SurfWinStormWinConstr(SurfNum);
```
mjwitte: Can these lines be replaced with `int ConstrNum = state.dataSurface->SurfWinActiveConstruction(SurfNum);`?

src/EnergyPlus/SolarShading.cc
```diff
@@ -10980,12 +10968,9 @@ void CalcBeamSolarOnWinRevealSurface(EnergyPlusData &state)
...
             W = state.dataSurface->Surface(SurfNum).Width;
             d1 = OutsReveal + 0.5 * GlazingThickness;
-            ConstrNum = state.dataSurface->Surface(SurfNum).Construction;
-            ConstrNumSh = state.dataSurface->Surface(SurfNum).activeShadedConstruction;
-            if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1) {
-                ConstrNum = state.dataSurface->Surface(SurfNum).StormWinConstruction;
-                ConstrNumSh = state.dataSurface->Surface(SurfNum).activeStormWinShadedConstruction;
-            }
+            ConstrNum = state.dataSurface->SurfWinActiveConstruction(SurfNum);
+            ConstrNumSh = state.dataSurface->SurfWinActiveShadedConstruction(SurfNum);
```
mjwitte: Ideally, it should be possible to localize the scope of every instance of `ConstrNum=` and `ConstrNumSh =` (or similar variables). And they should all be `int const ConstrNum = ` Yes?

tst/EnergyPlus/unit/DaylightingManager.unit.cc
```diff
@@ -855,6 +855,9 @@ TEST_F(EnergyPlusFixture, DaylightingManager_GetDaylParamInGeoTrans_Test)
     ScheduleManager::UpdateScheduleValues(*state);
     InternalHeatGains::GetInternalHeatGainsInput(*state);
     state->dataInternalHeatGains->GetInternalHeatGainsInputFlag = false;
+    for (int SurfNum = 1; SurfNum <= state->dataSurface->TotSurfaces; SurfNum++) {
+        state->dataSurface->SurfWinActiveConstruction(SurfNum) = state->dataSurface->Surface(SurfNum).Construction;
+    }
```
mjwitte: Perhaps `state->dataSurface->SurfWinActiveConstruction(SurfNum) = state->dataSurface->Surface(SurfNum).Construction` should be added to whatever function is setting `Surface(SurfNum).Construction` initially. Then it will always be initialized.

mjwitte (commented):
@xuanluo113 This is a great improvement - another example of setting something up front instead of checking and re-checking in multiple places.

src/EnergyPlus/DataSurfaces.hh
xuanluo113: I don't see why they are wrong. Can you suggest?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Originally I intended to only use `SurfWinActiveConstruction` to overwrite the exterior window constr index. The opaq surface loop would not need to be overwritten. It may be possible to use `SurfActiveConstruction` to overwrite any use case of `Surface(i).Construction`, because this logic 
```
if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1) {
    state.dataSurface->SurfWinActiveConstruction(SurfNum) = state.dataSurface->SurfWinStormWinConstr(SurfNum);
} else {
    state.dataSurface->SurfWinActiveConstruction(SurfNum) = state.dataSurface->Surface(SurfNum).Construction;
}
```
is set at the beginning of the simulation loop.

However, I don't know if the storm window index, when exists, is supposed to overwrite any construction index in any calculations. For example, I noticed that only when `WindowModelType(SurfNum) == Window5DetailedModel`, the constr index is overwritten, but for BSDF and EQL windows, the construction are still set as the original one (`Surface(SurfNum).Construction`). I checked the IO Ref and IDD notes and I don't see any restriction to attach a storm window when `ComplexFenestration` or `EQL layers` are applied.

src/EnergyPlus/DataSurfaces.hh
mjwitte: The comment lines currently state "... with storm window" but these variables can be with or without storm window, correct? My comment wasn't clear, I was proposing new comments above.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
mjwitte: Ok, one step at a time here.

mjwitte: But this is inside a loop for opaque surfaces, so state.dataSurface->SurfWinStormWinFlag(SurfNum) does not apply here. I think line 2860 ` if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1) ConstrNum = Surface(SurfNum).StormWinConstruction;` can be deleted.

xuanluo113: @mjwitte Regarding this, I just mocked up a test case to apply StormWindow to an exterior window with equivalent layers (idf file attached) and no error checkings prevent me from doing so. The storm index is assigned to the window, but triggers a segfault [here](https://github.com/NREL/EnergyPlus/blob/6334396b7bcebeb5eb801747eff3d316911f0956/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L7152) to get the `Construct(ConstrNum).EQLConsPtr` for the storm window construction. So I guess I should fix it by checking the window model [here](https://github.com/NREL/EnergyPlus/blob/6334396b7bcebeb5eb801747eff3d316911f0956/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L7133) and wherever the storm window is overwriting the original construction number. This should be an existing defect in the develop branch.

[EquivalentLayerWindow_test.idf.txt](https://github.com/NREL/EnergyPlus/files/6281192/EquivalentLayerWindow_test.idf.txt)

src/EnergyPlus/SolarShading.cc
xuanluo113: Same here. The logic is to assign `ConstrNum` to the original construction number at the beginning of the loop, but whether to overwrite depends on if the surface is a `Window5DetailedModel` window. Though `ConstrNumSh` is, in general, a const.
Maybe a better way of doing so is to check `if 5DETAILSMODEL && EXTERIOR WINDOW` when assigning. 
```
if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1 && 5DETAILSMODEL && ???) {
    state.dataSurface->SurfWinActiveConstruction(SurfNum) = state.dataSurface->SurfWinStormWinConstr(SurfNum);
} else {
    state.dataSurface->SurfWinActiveConstruction(SurfNum) = state.dataSurface->Surface(SurfNum).Construction;
}
```

tst/EnergyPlus/unit/DaylightingManager.unit.cc
xuanluo113: The assignment is added to the end of `GetSurfaceData` and these initialization are removed.

src/EnergyPlus/SolarShading.cc
xuanluo113: I added the checking of `5DETAILSMODEL` and make most of the assignment local and static.

src/EnergyPlus/SolarShading.cc
xuanluo113: This is removed.

src/EnergyPlus/DataSurfaces.hh
xuanluo113: This is modified. Besides, I renamed `SurfWinActiveConstruction` to `SurfActiveConstruction` to change the current logic of 
```
int constNum = Surface(i).Construction;
if (surface is window) constNum = SurfWinActiveConstruction(i);
```
to
`int const constNum = SurfActiveConstruction(i);`

xuanluo113:
@mjwitte Thanks for the review. The comments are addressed.

src/EnergyPlus/DataSurfaces.hh
mjwitte: OK, that makes sense. So, do these active construction variables get updated correctly if there is an EMS actuator for "Construction State"? I guess if it wasn't, the regressions should catch it, but just wondering if you traced that logic to be sure.

xuanluo113: Yes. This [time step reset](https://github.com/NREL/EnergyPlus/blob/d5da8dd3730a127bed0ecdb784dc6ce0396c8121/src/EnergyPlus/HeatBalanceManager.cc#L5567) of `SurfWinActiveConstruction` happens after `InitEMSControlledConstructions` ([code](https://github.com/NREL/EnergyPlus/blob/d5da8dd3730a127bed0ecdb784dc6ce0396c8121/src/EnergyPlus/HeatBalanceManager.cc#L5548)), where `Surface(i).Construction` is overwritten by the EMS value. And yes a unit test `HeatBalanceManager_EMSConstructionTest` assured that.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -226,9 +223,9 @@ void ManageSurfaceHeatBalance(EnergyPlusData &state)
     }
 
     if (state.dataHeatBal->AnyCondFD) {
-        for (SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
-            if (Surface(SurfNum).Construction <= 0) continue; // Shading surface, not really a heat transfer surface
-            ConstrNum = Surface(SurfNum).Construction;
+        for (int SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
```
mjwitte, line 226: Is this on your do-list for later to move this loop to a separate function and simply loop over opaque surfaces?

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5563,6 +5564,13 @@ namespace HeatBalanceManager {
                 state.dataHeatBalMgr->ChangeSet = true;
             }
         }
+        for (SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; SurfNum++) {
+            if (state.dataSurface->SurfWinStormWinFlag(SurfNum) == 1 && state.dataSurface->SurfWinWindowModelType(SurfNum) == DataSurfaces::Window5DetailedModel) {
+                state.dataSurface->SurfActiveConstruction(SurfNum) = state.dataSurface->SurfWinStormWinConstr(SurfNum);
+            } else {
+                state.dataSurface->SurfActiveConstruction(SurfNum) = state.dataSurface->Surface(SurfNum).Construction;
+            }
+        }
```
mjwitte: There is a slight performance slowdown for this branch in the performance suite test files (except performance.BenchmarkLargeOfficeNew_USA_CA_SAN_FRANCISCO_10_windows_per_zone which is slightly faster here). I wonder if this is the source of the few extra steps.
Can this loop be moved up inside the ` if (state.dataSurface->TotStormWin > 0` ? Also, this loop could be over zones then only window surfaces.
Or better yet, delete this entire loop and set `state.dataSurface->SurfActiveConstruction(SurfNum)` in `SetStormWindowControl`?

amirroth: Any of those three approaches would work. Or rather option 1+2, or option 3.

Generally speaking, there is something very intuitively clean and compelling about moving to a "push" model (option 3). However, in practice, theory and practice are different and a push model often results either in the distribution of logic and the proliferation of dependences (which can sometimes be circular) or the use of a "callback style" of programming that is strange in different ways.

All that is to say, that we should probably go with option 1+2 for now because it is local, but then think of whether we want to go to option 3 in the near/medium future and what pattern we want to use to do that.

mjwitte:
@xuanluo113 So, the last set of changes caused diffs in some EMS test files (that modify constructions). Which is correct, before or after?

xuanluo113:
@mjwitte Yes. I'm debugging that issue.

xuanluo113:
Fixed a bug to synchronize the change to `surface.Contruction` in run time to `SurfActiveConstrucion(surfnum)`. The dependency is a little annoying but does not happen very often (in EMS, ThermochromicWindow, and adding a HT shelf).

mjwitte (approved):
@xuanluo113 Thanks, looks good.

mjwitte:
Arrgh, I didn't check the clangformat test before merging. Will push a fix shortly.

PR #8677, opened 2021-03-30
Fix uninit water mains

lefticus:
This fixes use of uninitialized WaterMainsTempMethod

 * I do know this is definitely a real problem, it was found with runtime analysis with valgrind
 * I do ***not*** know if this is a correct solution.


Myoldmopar:
That's an interesting error.  I'm going to get a redo on Windows before investigating.

mitchute:
It builds fine on my Windows VM FWIW

Myoldmopar:
Windows built fine on CI, this is good to go as far as CI goes.  This also doesn't cause any diffs, which is a bit surprising if it's actually being used uninitialized.  Perhaps it's only being used uninitialized at the very beginning?  Before reporting happens?  In any case, it looks like this hasn't affected any actual results, so I'm feeling OK with this.  I'll take one more look at the change but I think this will go in shortly.

src/EnergyPlus/WeatherManager.hh
```diff
@@ -1024,8 +1025,9 @@ struct WeatherManagerData : BaseGlobalStruct
         this->GroundTempsFCFromEPWHeader = Array1D<Real64>(12, 0.0);
         this->GroundReflectances = Array1D<Real64>(12, 0.2);
 
-        this->SnowGndRefModifier = 1.0;              // Modifier to ground reflectance during snow
-        this->SnowGndRefModifierForDayltg = 1.0;     // Modifier to ground reflectance during snow for daylighting
+        this->SnowGndRefModifier = 1.0;          // Modifier to ground reflectance during snow
+        this->SnowGndRefModifierForDayltg = 1.0; // Modifier to ground reflectance during snow for daylighting
+        this->WaterMainsTempsMethod = WeatherManager::WaterMainsTempCalcMethod::FixedDefault;
```
Myoldmopar, line 1030: Got it, this is during the clear_state call, and then...

src/EnergyPlus/WeatherManager.hh
```diff
@@ -1207,14 +1209,15 @@ struct WeatherManagerData : BaseGlobalStruct
...
-          WaterMainsTempsScheduleName(""), wthFCGroundTemps(false), TotRunPers(0), TotRunDesPers(0), NumSpecialDays(0), SpecialDayTypes(366, 0),
-          WeekDayTypes(366, 0), DSTIndex(366, 0), NumDataPeriods(0), NumIntervalsPerHour(1), UseDaylightSaving(true), UseSpecialDays(true),
-          UseRainValues(true), UseSnowValues(true), EPWDaylightSaving(false), IDFDaylightSaving(false), DaylightSavingIsActive(false),
-          WFAllowsLeapYears(false), curSimDayForEndOfRunPeriod(0), Envrn(0), NumOfEnvrn(0), NumEPWTypExtSets(0), NumWPSkyTemperatures(0),
-          RptIsRain(0), RptIsSnow(0), RptDayType(0), HrAngle(0.0), SolarAltitudeAngle(0.0), SolarAzimuthAngle(0.0), HorizIRSky(0.0),
-          TimeStepFraction(0.0), NumSPSiteScheduleNamePtrs(0), EndDayOfMonth(12, {31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31}), LeapYearAdd(0),
-          DatesShouldBeReset(false), StartDatesCycleShouldBeReset(false), Jan1DatesShouldBeReset(false), RPReadAllWeatherData(false)
+          SnowGndRefModifierForDayltg(1.0), WaterMainsTempsMethod{WeatherManager::WaterMainsTempCalcMethod::FixedDefault}, WaterMainsTempsSchedule(0),
```
Myoldmopar, line 1212: This was during the default constructor of the data class (which I don't think we actually need...but that's a different story).

Myoldmopar (approved):
Yeah, again, I think this must've only been using it uninitialized early in the program before outputs were generated because this hasn't caused any strange behavior or spurious diffs that I know of.  I'm good with this quick fix.

Myoldmopar:
Thanks @lefticus , merging.

PR #4843, opened 2015-04-07
Updates to ObjexxFCL and 2+D arrays for row-major

Myoldmopar:
This passes my test suite with only a few _small_ diffs.  I will merge shortly.

PR #8692, opened 2021-04-03
Multi-dimensional array reordering and refactoring

xuanluo113:
Pull request overview
---------------------
This branch refactors some inefficient use of multi-dimensional array, including:
- Reordering some 2D window arrays to improve cache locality
- Change some 2D construction arrays to array of array to avoid slicing and reconstruction.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/Construction.cc
+++ b/src/EnergyPlus/Construction.cc
@@ -2002,9 +2002,25 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
     this->BlAbsDiffBack.dimension(DataSurfaces::MaxSlatAngs, state.dataHeatBal->MaxSolidWinLayers, 0.0);
     this->AbsBeamCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
     this->AbsBeamBackCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->tBareSolCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->tBareVisCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->rfBareSolCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->rfBareVisCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->rbBareSolCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->rbBareVisCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->afBareSolCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
+    this->abBareSolCoef.allocate(state.dataHeatBal->MaxSolidWinLayers);
     for (int Layer = 1; Layer <= state.dataHeatBal->MaxSolidWinLayers; ++Layer) {
         this->AbsBeamCoef(Layer).allocate(DataSurfaces::MaxPolyCoeff);
         this->AbsBeamBackCoef(Layer).allocate(DataSurfaces::MaxPolyCoeff);
+        this->tBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->tBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rfBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rfBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rbBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rbBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->afBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->abBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
     }
 }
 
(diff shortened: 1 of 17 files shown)


src/EnergyPlus/General.cc
```diff
@@ -1255,26 +1255,12 @@ Real64 BlindBeamBeamTrans(Real64 const ProfAng,        // Solar profile angle (r
 }
 
 Real64 POLYF(Real64 const X,         // Cosine of angle of incidence
```
amirroth: Given that in a number of callsites `X` is 1, I wonder if it makes sense to make this function `constexpr`.

src/EnergyPlus/Construction.cc
```diff
@@ -2000,8 +2000,28 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+        this->tBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->tBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rfBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rfBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rbBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->rbBareVisCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->afBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+        this->abBareSolCoef(Layer).dimension(DataSurfaces::MaxPolyCoeff, 0.0);
+    }
 }
```
amirroth, line 2052: Probably doesn't make much of a difference given that this is done for constructions and only one time, but in general we should move away from combined allocation and initialization using `dimension` and towards separate allocation and explicit initialization in which all arrays can be initialized together in a single (vectorized) loop.

amirroth:
This is a nice new code pattern! Thanks @xuanluo113!

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -7206,6 +7204,13 @@ namespace HeatBalanceManager {
                 state.dataConstruction->Construct(ConstrNum).TotGlassLayers = NGlass(IGlSys);
                 state.dataConstruction->Construct(ConstrNum).TotSolidLayers = NGlass(IGlSys);
 
+                for (int Layer = 1; Layer <= state.dataHeatBal->MaxSolidWinLayers; ++Layer) {
+                    for (int index = 1; index <= DataSurfaces::MaxPolyCoeff; ++index) {
+                        state.dataConstruction->Construct(ConstrNum).AbsBeamCoef(Layer)(index) = 0.0;
```
mjwitte, line 7209: This syntax will take a while to sink in (for me anyway). `x(i)(j)`

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -6234,8 +6234,8 @@ namespace HeatBalanceManager {
             if (thisSurface.Class == DataSurfaces::SurfaceClass::Window) {
                 auto &thisConstruct(thisSurface.Construction);
                 if (!state.dataConstruction->Construct(thisConstruct).WindowTypeBSDF) {
-                    state.dataHeatBal->SurfWinFenLaySurfTempFront(1, SurfNum) = state.dataHeatBalSurf->TH(1, 1, SurfNum);
-                    state.dataHeatBal->SurfWinFenLaySurfTempBack(state.dataConstruction->Construct(thisConstruct).TotLayers, SurfNum) =
+                    state.dataHeatBal->SurfWinFenLaySurfTempFront(SurfNum, 1) = state.dataHeatBalSurf->TH(1, 1, SurfNum);
```
mjwitte, line 6237: Interesting that these arrays (and many others) were originally (pre-c++) `x(SurfNum, LayerNum) (or whatever)`, but back in #4383, they were [reversed](https://github.com/NREL/EnergyPlus/pull/4843/files#diff-18fbb8a5b1334b62513de2441bdb4bb4ef1fbe5fbec24d5dd0effcd019dfc33dL721-R715) (find SolarShading.cc and expand it then the lines should show there) to be row-major, and now we're going back to the original order.

src/EnergyPlus/WindowManager.cc
```diff
@@ -714,49 +714,49 @@ namespace WindowManager {
                              6,
                              1,
                              TotalIPhi,
-                             state.dataConstruction->Construct(ConstrNum).tBareSolCoef(_, IGlass));
+                             state.dataConstruction->Construct(ConstrNum).tBareSolCoef(IGlass));
```
mjwitte, line 717: Always nice to get rid of `_ ,`

mjwitte (approved):
@xuanluo113 I'm ok with these changes. The swapped indexes in some of the 2d arrays is a bit scary, but I guess if any were missed there would be an array bounds error for sure in the CI tests.

Do you want to change the initialization method (per @amirroth [comment](https://github.com/NREL/EnergyPlus/pull/8692/files#r606707491) before this merges?

src/EnergyPlus/HeatBalanceManager.cc
xuanluo113: `x[i][j]` would make more sense after we change those arrays to vectors.

xuanluo113:
@mjwitte I changed the initialization method in the construction array initialization. More on this topic, I'm switching to a branch to convert surface object fields to surface HB arrays. I found the location of initialization of `dataSurface`, `dataHeatBal`, `dataHeatBalSurf`, `dataHeatBalSurfMgr` arrays are sometimes random (some in SurfaceGeometry, some in SolarShading, some in HBManger or HBSurfaceManager), and it's a little misleading to justify where to add new surface HB arrays. I'll probably revisit the initialization issue later.

Issue #3468, opened 2014-01-14
Documentation - Shortwave from Lights uses surface material solar absorptance rather than visible absorptance in heat balance calcs. (CR #8695)


axelstudios:
## 
#### Description

from code inspection while working on CR 8676, found that the shortwave energy from lights is combined with short wave energy from solar early on in the calculations and then the combined radiation is applied to surface using material property for solar absorptance.  But the Material input for visible absorptance implies that Lights should use that in heat balance calcs, not solar absorptance.  This may require substantial code changes to separate out. 

Issues:

VMULT is constructed based on solar absorptance.  would need to mirror same for visible absorptance

It may be fine to use solar absorptance for the surface heat transfer calcs.   It will depend on the light source, a high temperature source would have a lot of power in near infrared, but a flourescent or LED fixture would not.  

It will be a really big job to rewrite, marking as unconfirmed. 

MJW 21 Mar 2012
Since this is a big job to rewrite - seems that separating these should be an enhancement list item, and this CR should become a doc CR that clarifies how the visible output from lights is treated in the heat balance.  (BG agrees)

**\* Tasks for this CR include doc updates (I/O Ref and Engineering Ref) plus writing up a new Enhancement List Item **\*  

MJW 29 Aug 2012
So, if visible absorptance is not used for the visible energy from lights, what is it used for?  Daylighting calcs only?
Search of the source code shows that the only place it is currently used is in daylighting calcs for reflected solar from opaque surfaces.  In DaylightingManager there are a couple lines like this:
                    ObsVisRefl = 1.0d0 - Material(Construct(ObsConstrNum)%LayerPoint(1))%AbsorpVisible

I can't find anywhere else that uses this.  
Bob - please see if the engineering ref discusses the use of visible absorptance anywhere.
## 

External Ref: CR8676
Last build tested: `11.12.19 V7.1.0.001`

PR #8701, opened 2021-04-07
Movable insulation logic refactoring

xuanluo113:
Pull request overview
---------------------
The branch improves the logic to calculate movable insulation properties. The ext and int heat transfer coefficients and absorptances only need to calculated once at the beginning of the heat balance surface manager when any movable insulation exists. The subroutines of calculating these properties are transformed from external function calls to loop-wise initializations. The functions are moved to HeatBalanceSurfaceManager and the original movable insulation related files are removed.

### Pull Request Author

### Reviewer


xuanluo113:
@mjwitte This branch is ready for review. CI and performance tests look green. Thanks very much!

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -1467,11 +1475,17 @@ void AllocateSurfaceHeatBalArrays(EnergyPlusData &state)
...
     state.dataSurface->SurfSkySolarInc.dimension(state.dataSurface->TotSurfaces, 0);
     state.dataSurface->SurfGndSolarInc.dimension(state.dataSurface->TotSurfaces, 0);
 
+    // allocate movable insulation arrays
+    state.dataSurface->SurfMovInsulIntPresent.dimension(state.dataSurface->TotSurfaces, false);
+    state.dataSurface->SurfMovInsulIntPresentPrevTS.dimension(state.dataSurface->TotSurfaces, false);
+    state.dataSurface->SurfMovInsulHExt.dimension(state.dataSurface->TotSurfaces, 0.0);
+    state.dataSurface->SurfMovInsulAbsExt.dimension(state.dataSurface->TotSurfaces, 0.0);
+    state.dataSurface->SurfMovInsulHInt.dimension(state.dataSurface->TotSurfaces, 0.0);
+    state.dataSurface->SurfMovInsulAbsInt.dimension(state.dataSurface->TotSurfaces, 0.0);
```
mjwitte: Seems excessive to allocate all of these arrays when typically there will be no surfaces with movable insulation, and at most only a few of the surfaces in the entire model will have it.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2292,6 +2306,55 @@ void InitThermalAndFluxHistories(EnergyPlusData &state)
     }
 }
 
+void EvalOutsideMovableInsulation(EnergyPlusData &state)
+{
+    // This subroutine determines whether or not outside movable insulation on opaque surfaces is present at the current time.
+    // The SurfNum is passed in and then the rest of the parameters are set if movable insulation is present.
```
mjwitte: This comment is not relevant here.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5940,7 +5942,7 @@ namespace HeatBalanceManager {
         // Update interior movable insulation flag--needed at the end of a zone time step so that the interior radiant
         // exchange algorithm knows whether there has been a change in interior movable insulation or not.
         for (SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
-            state.dataSurface->Surface(SurfNum).MovInsulIntPresentPrevTS = state.dataSurface->Surface(SurfNum).MovInsulIntPresent;
+            state.dataSurface->SurfMovInsulIntPresentPrevTS(SurfNum) = state.dataSurface->SurfMovInsulIntPresent(SurfNum);
```
mjwitte: This loop can be skipped if `!AnyMovableInsulation`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2292,6 +2306,55 @@ void InitThermalAndFluxHistories(EnergyPlusData &state)
...
+    for (int zoneNum = 1; zoneNum <= state.dataGlobal->NumOfZones; ++zoneNum) {
```
mjwitte: Even if we keep the full arrays for movable insulation, it would be useful to build a list of surfaces with movable insulation and simply loop over that list here instead of looping over every opaque surface.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -2854,19 +2916,15 @@ void InitSolarHeatGains(EnergyPlusData &state)
...
                     Real64 AbsExt;              // Absorptivity of outer most layer (or movable insulation if present)
-                    if (Surface(SurfNum).MaterialMovInsulExt > 0) EvalOutsideMovableInsulation(state, SurfNum, HMovInsul, RoughIndexMovInsul, AbsExt);
-
-                    if (RoughIndexMovInsul <= 0) { // No movable insulation present
-
+                    if (state.dataSurface->SurfMovInsulHExt(SurfNum) > 0) {
+                        AbsExt = state.dataSurface->SurfMovInsulAbsExt(SurfNum);
+                    } else { // No movable insulation present
                         AbsExt = state.dataMaterial->Material(state.dataConstruction->Construct(ConstrNum).LayerPoint(1)).AbsorpSolar;
                     }
```
mjwitte: This if construct could be eliminated if we had a `SurfAbsExt` array that is intialized to the base construction `AbsorpSolar` and modified or reset as needed in `EvalOutsideMoveableInsulation`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3567,22 +3615,18 @@ void InitIntSolarDistribution(EnergyPlusData &state)
...
-            Real64 AbsInt;          // Inside opaque surface solar absorptance
-            Real64 AbsExt;          // Solar absorptance of outermost layer (or movable insulation if present)
-            int RoughIndexMovInsul; // Roughness index of movable insulation
-            if (Surface(SurfNum).MaterialMovInsulInt > 0) EvalInsideMovableInsulation(state, SurfNum, HMovInsul, AbsInt);
-            if (HMovInsul > 0.0) AbsIntSurf = AbsInt;
+            Real64 AbsIntSurfVis = state.dataConstruction->Construct(ConstrNum).InsideAbsorpSolar; // Inside opaque surface visible absorptance to fix CR 8695 change to
+
+            if (state.dataSurface->SurfMovInsulHInt(SurfNum) > 0) {
+                AbsIntSurf = state.dataSurface->SurfMovInsulAbsInt(SurfNum);
+            }
```
mjwitte: Similar comment here. `SurfIntAbs` and `SurfIntAbsVis` would eliminate this if and the repeated construction lookups above.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3567,22 +3615,18 @@ void InitIntSolarDistribution(EnergyPlusData &state)
...
+
             state.dataHeatBalSurf->SurfOpaqQRadSWInAbs(SurfNum) += state.dataHeatBal->QS(solEnclosureNum) * AbsIntSurf;
             state.dataHeatBalSurf->SurfOpaqQRadSWLightsInAbs(SurfNum) += state.dataHeatBal->QSLights(solEnclosureNum) * AbsIntSurfVis;
 
             // Calculate absorbed solar on outside if movable exterior insulation in place
-            HMovInsul = 0.0;
-            if (Surface(SurfNum).MaterialMovInsulExt > 0) EvalOutsideMovableInsulation(state, SurfNum, HMovInsul, RoughIndexMovInsul, AbsExt);
-            if (HMovInsul > 0) { // Movable outside insulation in place
+            if (state.dataSurface->SurfMovInsulHExt(SurfNum) > 0) { // Movable outside insulation in place
```
mjwitte: Because exterior movable insulation doesn't have a `Present` variable like inside movable insulation, `SurfMovInsulHExt >0` is being used to test for presence. In the old `EvalOutsideMovableInsulation` function, this was reset to zero when `MovInsulSchedVal <= 0.0` but I don't see that in the new version of this function.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3979,13 +4020,8 @@ void ComputeIntThermalAbsorpFactors(EnergyPlusData &state)
...
-            Real64 HMovInsul = 0.0;                                              // Conductance of movable insulation
-            Real64 AbsInt;                                                       // Solar absorptance of movable insulation
-            if (state.dataConstruction->Construct(ConstrNum).TransDiff <= 0.0) { // Opaque surface
-                if (Surface(SurfNum).MaterialMovInsulInt > 0) EvalInsideMovableInsulation(state, SurfNum, HMovInsul, AbsInt);
-                if (HMovInsul > 0.0)
-                    state.dataHeatBal->ITABSF(SurfNum) =
-                        state.dataMaterial->Material(Surface(SurfNum).MaterialMovInsulInt).AbsorpThermal; // Movable inside insulation present
+            if (state.dataSurface->SurfMovInsulHInt(SurfNum) > 0.0) {
+                state.dataHeatBal->ITABSF(SurfNum) = state.dataMaterial->Material(Surface(SurfNum).MaterialMovInsulInt).AbsorpThermal; // Movable inside insulation present
             }
```
mjwitte: Seems that `EvalInsideMovableInsulation` could manage setting `ITABSF`.  Seems this entire loop could be eliminated then. But what about EMS construction changes.

For your todo list: `ComputeIntThermalAbsorpFactors` and `ComputeIntSWAbsorpFactors` appear to be completely recomputed every timestep? Seems these function are ripe for skipping a zone if nothing has changed with window shade status or movable insulation? Not suggesting you try that here.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -5667,14 +5701,10 @@ void ReportSurfaceHeatBalance(EnergyPlusData &state)
...
 {
-    int SurfNum;
-    auto &Surface(state.dataSurface->Surface);
     state.dataHeatBalSurf->TempSurfInMovInsRep = state.dataHeatBalSurf->TempSurfIn;
-    for (SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
-        if (Surface(SurfNum).MaterialMovInsulInt > 0) {
-            if (GetCurrentScheduleValue(state, Surface(SurfNum).SchedMovInsulInt) > 0.0) {
-                state.dataHeatBalSurf->TempSurfInMovInsRep(SurfNum) = state.dataHeatBalSurf->TempSurfInTmp(SurfNum);
-            }
+    for (int SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
```
mjwitte: Another spot for a list-driven update.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -6136,13 +6165,11 @@ void CalcHeatBalanceOutsideSurf(EnergyPlusData &state,
...
                     Real64 AbsThermSurf = state.dataMaterial->Material(state.dataConstruction->Construct(ConstrNum).LayerPoint(1)).AbsorpThermal;
-
+                    HMovInsul = state.dataSurface->SurfMovInsulHExt(SurfNum);
                     // Check for outside movable insulation
-                    if (Surface(SurfNum).MaterialMovInsulExt > 0) {
-                        EvalOutsideMovableInsulation(state, SurfNum, HMovInsul, RoughSurf, AbsThermSurf);
-                        if (HMovInsul > 0)
-                            AbsThermSurf = state.dataMaterial->Material(Surface(SurfNum).MaterialMovInsulExt)
-                                               .AbsorpThermal; // Movable outside insulation present
+                    if (state.dataSurface->SurfMovInsulHExt(SurfNum) > 0) {
```
mjwitte: Another place where arrays of surface roughness and absthermal could eliminate this if and repeated lookups from `dataConstruction`.

mjwitte (commented):
@xuanluo113 This is a good start. I think there's room for more improvement. See comments below.

mjwitte:
And there are clang formatting check errors - need to apply formatting to all of the changed files.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: The alternative would be to ... have a single array that indexes to a set of smaller arrays (or even one arrays of structures) sized to fit only the number of surfaces that have `MovableInsulation`? :thinking: How do we feel about that design pattern in general? How broadly should we apply it? I have been thinking about whether we should do something like this for all of the Window specific arrays. :thinking: :thinking:

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: Just factoring this code into a separate loop and guarding it with `if (MovableInsulationAnywhereInTheModel)` will speed up models that have no `MovableInsulation` in them. A "push" implementation like the one @mjwitte suggests will also speed up models that do have `MovableInsulation` in them, although do we really care about such models right now?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: We should probably implement inside and outside `MovableInsulation` in similar ways, no? Consistency is the hobgoblin of little minds. And of Amir.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Five HB arrays are added including `SurfAbsSolarExt`, `SurfAbsThermalExt`, `SurfAbsSolarInt`, `SurfAbsThermalInt`, SurfRoughnessExt`. They are initialized with material properties one time at the beginning of `InitSurfaceHeatBalance`, and are set to movable insulation properties (and set back) at run time when `AnyMovableInsulation`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Five HB arrays are added including SurfAbsSolarExt, SurfAbsThermalExt, SurfAbsSolarInt, SurfAbsThermalInt, SurfRoughnessExt. They are initialized with material properties one time at the beginning of InitSurfaceHeatBalance, and are set to movable insulation properties (and set back) at run time when AnyMovableInsulation`.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: `SurfMovInsulExtPresent`, `SurfMovInsulIntPresent`, `SurfMovInsulIntPresentPrevTS`, `SurfMovInsulHExt`, `SurfMovInsulHInt` are allocated only when `AnyMovableInsulation`, and whenever they are used (e.g. before checking if `SurfMovInsulExtPresent`), `AnyMovableInsulation` is checked first.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Deleted.

src/EnergyPlus/HeatBalanceManager.cc
xuanluo113: Flag added.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: A vector `SurfMovInsulIndexList` is initialized at the input stage and the loops are reduced.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Like explained in the previous comment, `AbsorpSolar` and `AbsorpThermal` are saved to runtime HB arrays. I removed `SurfMovInsulAbsInt` array here and added `SurfAbsSolarInt` instead.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: I added `SurfMovInsulExtPresent` and used the boolean arrays in checking presence for both inside and outside surface.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Used the index list instead.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Runtime `SurfAbsThermalExt` is used instead here and initialized with `state.dataSurface->SurfActiveConstruction` at the beginning of simulation or when any EMS override is on.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: In [develop branch](https://github.com/NREL/EnergyPlus/blob/b622e27ca3c96566d6bbc8c6429ae971f880d445/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L3570), `AbsIntSurfVis` is set as `Construct(ConstrNum).InsideAbsorpSolar` instead of `InsideAbsorpVis`, and is not reset when movable insulation is on. Although the comment here says "Inside opaque surface visible absorptance to fix CR 8695 change to this = Construct(ConstrNum)%InsideAbsorpVis". I wonder if this is a bug or it's intended to be fixed like this. @mjwitte Could you help take a look?

xuanluo113:
@mjwitte Thanks for the review. I've finished addressing most of the comments as replied, except for two items to work on here or later.

1. The `AbsIntSurfVis` bug (?) as above.
2. `dataHeatBal->ITABSF` seems to have the same physical meaning as the newly added `SurfAbsThermalExt` array, while `ITABSF` is used in window surfaces and `SurfAbsThermalExt` in opaque surfaces. I intend to remove `ITABSF`, but this can be done along with the todo item listed above - reducing `ComputeIntThermalAbsorpFactors` and `ComputeIntSWAbsorpFactors`?

src/EnergyPlus/DataGlobals.hh
```diff
@@ -141,6 +141,7 @@ struct DataGlobal : BaseGlobalStruct
     bool RedoSizesHVACSimulation = false;           // doing kick off simulation for redoing sizes as part of sizing
     bool FinalSizingHVACSizingSimIteration = false; // when doing HVAC sizing Simulation
     bool AnyEnergyManagementSystemInModel = false;  // true if there is any EMS or Erl in model.  otherwise false
+    bool SurfPropOverridesPresent = false;          // true if there is any EMS or Erl in model.  otherwise false
```
mjwitte: This endline comment should be more specific. And probably should be named `AnySurfPropOverridesInModel` to be consistent with the others? This can be be a TODO for later if you want.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -395,13 +394,35 @@ void InitSurfaceHeatBalance(EnergyPlusData &state)
         state.dataRoomAirMod->IsZoneCV.dimension(state.dataGlobal->NumOfZones, false);
         state.dataRoomAirMod->IsZoneUI.dimension(state.dataGlobal->NumOfZones, false);
     }
+    if (state.dataGlobal->BeginSimFlag || state.dataGlobal->SurfPropOverridesPresent) {
+        for (int zoneNum = 1; zoneNum <= state.dataGlobal->NumOfZones; ++zoneNum) {
+            int const firstSurf = state.dataHeatBal->Zone(zoneNum).HTSurfaceFirst;
+            int const lastSurf = state.dataHeatBal->Zone(zoneNum).HTSurfaceLast;
+            for (int SurfNum = firstSurf; SurfNum <= lastSurf; ++SurfNum) {
+                int ConstrNum = state.dataSurface->SurfActiveConstruction(SurfNum); // SurfActiveConstruction set above
+                if (ConstrNum > 0) {
```
mjwitte: Is it necessary to check for `ConstrNum>0`? All heat transfer surfaces should have a valid construction.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -3566,23 +3622,18 @@ void InitIntSolarDistribution(EnergyPlusData &state)
         for (int SurfNum = firstSurfOpaque; SurfNum <= lastSurfOpaque; ++SurfNum) {
             int const ConstrNum = Surface(SurfNum).Construction;
 
-            Real64 AbsIntSurf = state.dataConstruction->Construct(ConstrNum).InsideAbsorpSolar; // Inside opaque surface solar absorptance
+            Real64 AbsIntSurf = state.dataHeatBalSurf->SurfAbsSolarInt(SurfNum);
+            // TODO - AbsIntSurfVis = InsideAbsorpSolar instead of InsideAbsorpVis?
```
mjwitte, line 3624: If I'm following correctly, `AbsIntSurfVis` is only used for reporting, not in the actual surface heat gain calculations. Per the comments in [CR8695](https://github.com/NREL/EnergyPlus/issues/3468) it sounds like these needs to remain AbsSolar so that the reporting is consistent with the actual heat balance. But it's certainly worth looking at. It might be possible to fix that (later).

mjwitte (approved):
@xuanluo113 This is so much better with much cleaner logic. A few minor comments below.
I haven't looked at the ITABSF question yet.
I'll wait for CI to finish before merging. Or if you want to address the SurfPropOverridesPresent comment in this PR, let me know.

src/EnergyPlus/DataGlobals.hh
xuanluo113: I've fixed this comment and naming. Waiting for the local tests to finish for removing `ConstrNum > 0`. Will make a push later.

mjwitte:
@xuanluo113 Looks good. Merging.

jcyuan2020:
This PR gets an extra credit for fixing a newly reported issue #8758.

PR #8706, opened 2021-04-09
Refactor existing extern int const to enum classes

jmythms:
Changes all remaining extern int const-s to enum classes.

### Pull Request Author

### Reviewer


src/EnergyPlus/Pumps.hh
```diff
@@ -91,16 +91,20 @@ namespace Pumps {
...
+
+    enum class iPumpType : int
+    {
+        Unassigned = 0,
+        VarSpeed = 101,
+        ConSpeed = 102,
+        Cond = 103,
+        Bank_VarSpeed = 104,
+        Bank_ConSpeed = 105
+    };
```
amirroth, line 107: Try to make these just 1-5, instead of 101-105. See what happens. Live a little!

src/EnergyPlus/TARCOGCommon.cc
```diff
@@ -70,14 +70,15 @@ namespace TARCOGCommon {
...
 
         // Using/Aliasing
         using namespace TARCOGParams;
 
-        return layertype == VENETBLIND_HORIZ || layertype == VENETBLIND_VERT || layertype == WOVSHADE || layertype == PERFORATED ||
-               layertype == BSDF || layertype == DIFFSHADE;
+        return layertype == TARCOGLayerType::VENETBLIND_HORIZ || layertype == TARCOGLayerType::VENETBLIND_VERT ||
+               layertype == TARCOGLayerType::WOVSHADE || layertype == TARCOGLayerType::PERFORATED || layertype == TARCOGLayerType::BSDF ||
+               layertype == TARCOGLayerType::DIFFSHADE;
     }
```
amirroth: Perfect spot for the `BITF` trick.

src/EnergyPlus/TARCOGParams.hh
```diff
@@ -65,39 +65,56 @@ int constexpr maxlay2(maxlay * 2);  // maximum number of glass surfaces (maxlay*
...
+{
+    UNASSSIGNED = -1,
+    SPECULAR = 0,
+    VENETBLIND_HORIZ = 1,
+    WOVSHADE = 2,
+    PERFORATED = 3,
+    DIFFSHADE = 4,
+    BSDF = 5,
+    VENETBLIND_VERT = 6
+};
```
amirroth, line 79: So ... in some places we are setting `UNASSIGNED` as `0`, in others we are using `-1`. We should settle on one, preferably `-1` but we could make `0` work also.

src/EnergyPlus/TARCOGCommon.cc
jmythms: I did not understand. Do you mean bit fields?

src/EnergyPlus/TARCOGParams.hh
jmythms: Got it. Will change to `-1`, and let the group know as well 👍🏽

src/EnergyPlus/Pumps.hh
jmythms: Will remember consistency is key!

Myoldmopar: I'd say even go farther and do -1 for the invalid case and 0-4 for the valid entries.

src/EnergyPlus/TARCOGCommon.cc
amirroth: Yes, `BITF_TEST_ANY(BITF(layertype), BITF(TARCOGLayerType::VENETBLIND_HORIZ) | BITF(TARCOGLayerType::VENETBLIND_VERT) | ... )`. One instruction instead of 12.

jmythms: Awesome. Will do!

src/EnergyPlus/Pumps.hh
mjwitte: Just beware any if statements that may be checking for `=0` or `<0`  or `<100` etc. instead of testing against an enum name.

jmythms: Thanks. Most of these comparison checks usually pop up as IDE/compiler errors when we convert the `extern int const` definitions to `enum`s, saying that enums cannot be compared with ints. I have tried to refactor any checks like `<0` and `<100` to check against the actual enumerators, but I will keep an eye out. If I did things right (fingers crossed), converting the (101-105) to (-1 to 4) should be simple and I would not have to make changes anywhere else, since the associated `int`s are not being used for comparisons anymore. Thank you for checking and the advice,

amirroth: What is the i in `iPumpType`? `int`? I'm not against [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation) (they made us use it at Microsoft back in the day) but I didn't think we use it in EnergyPlus. Neither here nor there.

mitchute: That's probably a precedent I started or at least perpetuated. I believe I kept running into naming conflicts, so I just added the "i" to help distinguish the enum.

jasondegraw: Let's please not use Hungarian Notation. There are other ways to address conflicts.

jmythms: I think another small advantage being the distinction between enum class names and variables that use them (like [here](https://github.com/NREL/EnergyPlus/blob/f4a11a4781cf31579144724e559705ed1b9ad4f2/src/EnergyPlus/Pumps.hh#L148)). If there is a standard format for naming enum class names that we would like to follow, I'm glad to change to that.

amirroth: So, now that I think of it this isn't really HN, because `iPumpType` is a type not a variable. If this were HN, it would be `PumpType_t`. @jasondegraw, your thoughts?

jasondegraw: I'm pretty sure that a more standard approach would be to use lower case initial characters for variables (`thisIsMyVariable`) and upper case initial characters for types (`ThisIsMyType`).

Myoldmopar: I support that simple style.  In this case we are already in the pump's namespace, so you could probably go pretty simple and still have clarity:

```
namespace EnergyPlus {
 namespace Pumps {
  enum class Type { Constant, Variable, Etc }; // or PumpType if you really want
  struct PumpData {
   ...
   Type type; // or myType if you want
   ...
  }
 }
}

Comment (mitchute) [reply src/EnergyPlus/Pumps.hh:107]: I'll name them to whatever convention is generally agreed upon as best practice. I'm not aware of anything in our Wiki that describes this, although, those pages are somewhat out of date. I believe I was just looking at the Google C++ style guide on enumerators (https://google.github.io/styleguide/cppguide.html#Enumerator_Names) that says they can be named like constants (see here: https://google.github.io/styleguide/cppguide.html#Constant_Names)

Comment (jasondegraw) [reply src/EnergyPlus/Pumps.hh:107]: @amirroth I'd take `Type_t` over `iType` any day of the week.

Comment (jasondegraw) [reply src/EnergyPlus/Pumps.hh:107]: @mitchute I don't really find that to be too persuasive. I believe I've seen cases where the enumeration is getting named something like `iThingTypeNum`. It's no longer a number, so the obvious name to me would be `Thing` or `ThingType`. I've been slowly changing variable names to start with lower case as I work on objects and starting out that way with new objects, so from my point of view this method of naming is heading the wrong way.

Comment (mitchute) [reply src/EnergyPlus/Pumps.hh:107]: You'll get no disagreement from me that there are plenty of bad examples of type names inside of EnergyPlus. In fact, that's probably what lead me down this road to begin with. Probably something like ```PumpType``` not being an enum type, but being a poorly named struct or local variable that was causing naming problems. This was a quick way of getting around the issue. Arguably it wasn't the best, but it was *A* way.

That being said, I don't think the leading "i" or "k" or whatever is really that bad. At least for me, since I've been working with these so much over the last year, my brain is already programmed to understand that the leading "i" means that I'm dealing with an enumerator, and that I don't have to drill in and see whether this really is an enumerator "type", or some other poorly named struct.

However, as I said, I'll do whatever.

File: src/EnergyPlus/TARCOGCommon.cc (hunk 2 above)
Comment (jmythms) [reply src/EnergyPlus/TARCOGCommon.cc]: Is it recommended to use `BITF` in other places as well?

File: src/EnergyPlus/Pumps.cc [hunk 4]
```diff
@@ -2051,7 +2051,7 @@ void CalcPumps(EnergyPlusData &state, int const PumpNum, Real64 const FlowReques
                             state.dataPumps->PumpEquip(PumpNum).PartLoadCoef(4) * pow_3(PartLoadRatio);
         state.dataPumps->Power = FracFullLoadPower * state.dataPumps->PumpEquip(PumpNum).NomPowerUse;
 
-    } else if (PumpType == iPumpType::Bank_ConSpeed || PumpType == iPumpType::Bank_VarSpeed) {
+    } else if BITF_TEST_ANY(BITF(PumpType), BITF(iPumpType::Bank_ConSpeed) | BITF(iPumpType::Bank_VarSpeed)) {
```
amirroth: If you have an if-else ladder of these things, you can also use switch/case:

```
switch(PumpType) {
   case iPumpType::ConSpeed:
   case iPumpType::VarSpeed:
   case iPumpType::Cond: 
        blahblahblah;
        break;
   case iPumpType::Bank_ConSpeed:
   case iPumpType::Bank_VarSpeed:
       yadayadayadah;
       break;
   default:
      thisisprobablyanerror;
      break;
}
```

src/EnergyPlus/Pumps.cc
```diff
@@ -2041,8 +2042,7 @@ void CalcPumps(EnergyPlusData &state, int const PumpNum, Real64 const FlowReques
     //****************************!
     //***** CALCULATE POWER (1) **!
     //****************************!
-    if (PumpType == iPumpType::ConSpeed || PumpType == iPumpType::VarSpeed || PumpType == iPumpType::Cond) {
-
+    if (BITF_TEST_ANY(BITF(PumpType), BITF(iPumpType::ConSpeed) | BITF(iPumpType::VarSpeed) | BITF(iPumpType::VarSpeed))) {
         VolFlowRate = state.dataPumps->PumpMassFlowRate / LoopDensity;
```
amirroth, line 2054: The last one should be iPumpType::Cond, no?

src/EnergyPlus/TARCOGArgs.cc
```diff
@@ -425,10 +425,15 @@ int ArgCheck(EnergyPlusData &state,
...
+        if BITF_NOT_TEST_ANY(BITF(LayerType(i)), BITF(TARCOGLayerType::SPECULAR) |
+                                                   BITF(TARCOGLayerType::WOVSHADE) |
+                                                   BITF(TARCOGLayerType::VENETBLIND_HORIZ) |
+                                                   BITF(TARCOGLayerType::PERFORATED) |
+                                                   BITF(TARCOGLayerType::DIFFSHADE) |
+                                                   BITF(TARCOGLayerType::BSDF) |
+                                                   BITF(TARCOGLayerType::VENETBLIND_VERT))
+
+        {
             ArgCheck = 22;
```
amirroth, line 434: So, `BITF_NOT_TEST_ANY` is a bad name, a better one would be `BITF_TEST_NONE` (i.e., it matches none of the values). Alternatively you can also just use `!BITF_TEST_ANY`.

jmythms (commented):
Accidentally started reviewing my own PR.

src/EnergyPlus/Pumps.cc
jmythms: Yes, thanks for noticing this.

src/EnergyPlus/TARCOGArgs.cc
jmythms: Makes sense! Thank you.

src/EnergyPlus/Pumps.cc
jmythms: Thank you. Is there a preferred/optimum way to do it (Or decide between if-else/switch/BITF methods)? In this case, I am guessing BITF doesn't give many advantages since there are just three+two items being compared. So it is best to change to switch/case with the most common cases put first?

amirroth: There are actually two separate issues here. 

Issue 1. Should you use `Switch/case` or `if-else`? You should use `switch/case` if the `if-else` ladder has at least three or four cases and one of them is not dynamically dominant. There is also the limitation that `switch/case` is also for single variable equality tests, you cannot have a compound conditional in a `case`. That is the specific case here. For all other situations you either have to use `if-else` or you don't have to, but you should. This brings us to issue 2.

Issue 2. If you are using `if-else`, should you use `BITF`? The test here is how many things you are comparing against. `BITF` is three instructions including one branch for any number of comparisons. `X == A || X == B ...` is two instructions including one branch _per_ comparison. By the numbers (especially the branch numbers) you want to use `BITF` anytime you are comparing against two or more things. The performance advantage for two comparisons is admitted small, but it is there. One disclaimer is that this only true if the things you are comparing against are enumerations or other small compile-time constants.

src/EnergyPlus/Pumps.hh
amirroth: @Myoldmopar, so with your example "PumpType" enumerations would be `Type::Constant` within the Pumps namespace and `Pumps::Type::Constant` outside of it? I don't know if I like that... 

PS I would similarly complain about the type name itself but I realize that all of that is just going to be `auto` anyway so what's the point ...

src/EnergyPlus/Pumps.cc
jmythms: Thank you very much again. Last question: does dynamically dominant mean that the probability of a case occurring is higher than others?

amirroth: That's right. If one of the cases occurs 90% of the time, you want to use an `if-then` ladder and put that case first. The remaining cases will be a bit slower than they would have been had you used a `switch\case`, but this case will be faster and that is the one that matters for performance.

jmythms: Got it. Thanks so much!

Myoldmopar:
Looks like most of the review comments have been addressed and CI is all clean.  I haven't looked over the final changes but will be looking it over soon.  If anyone else has anything further go ahead and take a look today if you can.

src/EnergyPlus/DXCoils.hh
```diff
@@ -223,14 +224,14 @@ namespace DXCoils {
...
-        int DefrostStrategy;               // defrost strategy; 1=reverse-cycle, 2=resistive
-        int DefrostControl;                // defrost control; 1=timed, 2=on-demand
-        int EIRFPLR;                       // index of energy input ratio vs part-load ratio curve
-        int DefrostEIRFT;                  // index of defrost mode total cooling capacity for reverse cycle heat pump
-        int RegionNum;                     // Region number for calculating HSPF of single speed DX heating coil
-        Real64 MinOATCompressor;           // Minimum OAT for heat pump compressor operation
-        Real64 OATempCompressorOn;         // The outdoor tempearture when the compressor is automatically turned back on,
+        Real64 HeatingCoilRuntimeFraction;                          // Run time fraction of the DX heating unit
+        int DefrostStrategy;                                        // defrost strategy; 1=reverse-cycle, 2=resistive
+        Optional<StandardRatings::HPdefrostControl> DefrostControl; // defrost control; 1=timed, 2=on-demand
```
Myoldmopar: Hmmmm, I don't really like this optional being added.  Can you explain why that was added here?

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -8881,8 +8882,8 @@ namespace HeatBalanceManager {
                                          state.dataIPShortCut->rNumericArgs(10)));
             }
 
-            if (state.dataHeatBal->ComplexShade(Loop).LayerType == csVenetianHorizontal ||
-                state.dataHeatBal->ComplexShade(Loop).LayerType == csVenetianVertical) {
+            if (BITF_TEST_ANY(BITF(state.dataHeatBal->ComplexShade(Loop).LayerType),
```
Myoldmopar, line 8895: I still have questions about using these macros vs using constexpr functions, but they can wait for another time.  And I need to go back and redigest the discussion from when they were discussed previously before I make any further commentary.

src/EnergyPlus/TARCOGCommon.hh
```diff
@@ -54,6 +54,11 @@
 // EnergyPlus Headers
 #include <EnergyPlus/Data/BaseData.hh>
 #include <EnergyPlus/EnergyPlus.hh>
+#include <EnergyPlus/TARCOGParams.hh>
+
+#define BITF(B) (1 << (int(B)))
```
Myoldmopar: Is this a second definition of BITF and BITF_TEST_ANY?  If so they need to be pulled into a single location and reused.

Myoldmopar (changes requested):
@jmythms this is a large effort, thanks!   Well over 99% of it looks great, I just had a request to get rid of, or at least explain, the addition of a new optional argument, and then determining if we duplicated the definition of BITF_TEST_ANY and BITF.

jmythms:
> @jmythms this is a large effort, thanks! Well over 99% of it looks great, I just had a request to get rid of, or at least explain, the addition of a new optional argument, and then determining if we duplicated the definition of BITF_TEST_ANY and BITF.

Thank you very much for reviewing. Working on this now.

src/EnergyPlus/DXCoils.hh
jmythms: This was a mistake. It crept in while trying to refactor optional_int to optional<enum class>. I removed it and scanned through the rest of the PR to find any other instances of this as well.

src/EnergyPlus/TARCOGCommon.hh
jmythms: Done 👍🏽

jmythms:
> I would not include `DataSurfaces.hh` anywhere you want to use `BITF`. I would move the `BITF` definitions to `General.hh` or even to a new `BITF.hh` file.

I added the file and moved the definitions [here](https://github.com/NREL/EnergyPlus/blob/44347107eb0b8a2fd24289bbfcf6c711b1e92e06/src/EnergyPlus/BITF.hh).

Myoldmopar (approved):

Myoldmopar:
This looks great to me.  CI is super happy.  This PR will help push the codebase much further in the right direction.  Thanks for pushing on this @jmythms.  I will go ahead and merge this now, and if there are any follow-up issues, let's tackle them in the next PR!

jmythms:
@Myoldmopar Thank you very much! 🎉

PR #8707, opened 2021-04-09
Prototype simplified array1d

### LINKED ISSUE #8598: EnergyPlus Build Time Increasing
Issue overview
--------------
The build time for EnergyPlus has steadily increased as of late.  It's not a crazy amount, but it is becoming annoying, and starting to impact our CI's ability to produce tests in reasonable time.

lefticus:
The goal of this PR is to reduce compile times by using a simplified version of the Array1D and enabling PCH, and close #8598 (see previous discussion on #8598)

The results (posted in 571f04137954511ae2bee4def009ae6896111fff)

Older timing data: e62f0f24312d95afcc527af7a507b3a8a00b8500

To do

* consider using `extern template` with the expensive functions in nlohmann::json to further reduce compile times
* consider getting rid of the `unique_ptr<>`'s for the state objects. They might be more expensive to compile and parse then just merging everything into one object.
* Replace more usage of `Array1D` with `EPVector` then further replace `EPVector` with `std::vector` (and use ranged for loops instead of indexed loops)

### EPVector

```
3059.39user 149.65system 8:15.61elapsed 647%CPU (0avgtext+0avgdata 2220812maxresident)k
16inputs+13190552outputs (0major+35758085minor)pagefaults 0swaps

 34251576 Apr  9 13:17 Products/energyplus-9.5.0
312094912 Apr  9 13:15 Products/libenergyplusapi.so.9.5.0
```

### Array1D

```
4988.46user 166.81system 13:41.42elapsed 627%CPU (0avgtext+0avgdata 2431812maxresident)k
512inputs+13347056outputs (0major+38409650minor)pagefaults 0swaps

 37482448 Apr  9 13:25 Products/energyplus-9.5.0
313572560 Apr  9 13:25 Products/libenergyplusapi.so.9.5.0
```

-----------------------

### EPVector

```
4234.06user 197.40system 10:15.15elapsed 720%CPU (0avgtext+0avgdata 2213988maxresident)k
0inputs+13004280outputs (0major+45937739minor)pagefaults 0swaps
```

### Array1D

```
6166.44user 216.58system 15:37.60elapsed 680%CPU (0avgtext+0avgdata 2430876maxresident)k
400inputs+13077808outputs (5major+48555472minor)pagefaults 0swaps
```


src/EnergyPlus/EPVector.hh
```diff
@@ -0,0 +1,292 @@
...
+#define EPVector_hh_INCLUDED
+
+#include <numeric>
+#include <vector>
+
+
+namespace EnergyPlus {
+
+template <typename T> struct EPVector : private std::vector<T>
+{
```
amirroth, line 59: Can we use an aligned allocator? Asking for a friend. Named Amir.

Myoldmopar: @amirroth I did bring this up before you were on that call the other day.  The consensus was that the standard library will already use aligned allocators and that we don't need to add anything to our custom class.  If you are actually seeing behavior in the code that is counter to this then we will definitely need to address it, but it seems it should be acting properly as coded right now.  @lefticus do you have anything else to add?

Myoldmopar:
@lefticus we are still building on Ubuntu 18.04, which only has CMake 3.10, and thus does not have the PCH stuff baked in.  With CI, the PCH stuff shouldn't matter anyway, so for now I just turned that flag off.

src/EnergyPlus/EPVector.hh
amirroth: Standard library aligns to size of _type_. Does it also align to size of _vector_?

amirroth: Ok, so I did a test with a loop on 100 `std::vector<double>`s of various sizes and everything came out 32-byte aligned, which makes sense given that my 6 year old Mac Core i7 only has AVX1.0. I will drop this for now and loop back if I see any alignment issues with vectorization. Thanks.

Myoldmopar:
Looks like this would be fully ready to go if not for the diffs in one file that look significant enough to hold this up.  We should be able to get absolutely no diffs so it's worth investigating.  I ran a quick bisection but found the diffs started appearing in the first commit to this branch a while back.  (That commit didn't build so I had to take the first commit plus cherry pick the header inclusion fix that is like the fifth commit in this branch).  That still leaves a lot of possible array changes that could be the culprit.  I went through and found just a few that looked suspicious and tried to revert them to get the diff to go away but couldn't get there quickly.  I wonder if you have any thoughts before we just brute force start reverting chunks of files in that one commit to eliminate the diff.

Myoldmopar:
@lefticus the file system test failure is something that occurs very randomly, not related to your changes.  The one file still showing regressions is the same one that has been problematic so far.  And the new Github-Actions based Clang-Format checks are showing some files that need formatting applied before it is mergeable.

As I said earlier, other than the regression file, I think this branch is ready to make a huge dent in the build time and will be a major benefit to our local builds as well as our CI testing.  It should also be made clear that this branch is not the end-all to getting build time improved, but one major step forward.  If anyone wants to look at this before it goes in, I'd say go ahead and do it while we are investigating the diffs so that we can get it in promptly afterwards.

lefticus:
@Myoldmopar I was seeing some anomalies over here and wanted a fresh set of eyes (CI) on a new (freshly merged from develop) build. I now have a situation where I can reproduce the specific issue, and think I'll be able to figure it out shortly.

Myoldmopar:
Yay!  Diffs are gone.  Some Clang-formatting to apply, but I can do that if it turned out that was the only thing remaining.  @lefticus can you confirm whether or not you are set on this branch for now?

lefticus:
I was waiting for CI to finish to double check my side, otherwise I think
I'm done.

Hopefully the commit messages make sense.

.decent_ci-Linux.yaml
```diff
@@ -22,7 +22,7 @@ compilers:
   - name: "gcc"
     version: "7.5"
     build_type: Debug
-    cmake_extra_flags: -DLINK_WITH_PYTHON=ON -DBUILD_FORTRAN=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=OFF -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_COVERAGE:BOOL=ON -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF
+    cmake_extra_flags: -DLINK_WITH_PYTHON=ON -DBUILD_FORTRAN=ON -DBUILD_TESTING:BOOL=ON -DENABLE_REGRESSION_TESTING:BOOL=OFF -DCOMMIT_SHA=$COMMIT_SHA -DENABLE_COVERAGE:BOOL=ON -DENABLE_GTEST_DEBUG_MODE:BOOL=OFF -DENABLE_PCH:BOOL=OFF
```
Myoldmopar, line 25: I added these lines because PCH isn't necessary on our CI system and also the 18.04 image doesn't have an updated enough CMake for it anyway.

CMakeLists.txt
```diff
@@ -24,6 +24,21 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
...
+  #
+  # consider breaking this out per project as necessary
+  target_precompile_headers(
+    cpp_pch_files
+    INTERFACE
+    "${PROJECT_SOURCE_DIR}/src/EnergyPlus/Data/EnergyPlusData.hh"
+    "${PROJECT_SOURCE_DIR}/third_party/ObjexxFCL/src/ObjexxFCL/Array.functions.hh"
+    "${PROJECT_SOURCE_DIR}/third_party/nlohmann/json.hpp"
+    )
+endif()
```
Myoldmopar, line 40: Using PCH where appropriate will be a really nice benefit.

cmake/CompilerFlags.cmake
```diff
@@ -8,6 +8,13 @@ endif()
 
 # Make sure expat is compiled as a static library
 target_compile_definitions(project_options INTERFACE -DXML_STATIC)
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  option(BUILD_TIME_TRACE "Enable -ftime-trace for investigating build times on clang" OFF)
+  mark_as_advanced(BUILD_TIME_TRACE)
+  if (BUILD_TIME_TRACE)
+    target_compile_options(project_options INTERFACE -ftime-trace)
+  endif()
+endif()
```
Myoldmopar, line 17: Really excited to try this out myself.

src/EnergyPlus/AirflowNetworkBalanceManager.hh
```diff
@@ -297,17 +299,17 @@ struct AirflowNetworkBalanceManagerData : BaseGlobalStruct
...
-    Array1D<AirflowNetwork::AirflowNetworkExchangeProp> exchangeData;
-    Array1D<AirflowNetwork::AirflowNetworkExchangeProp> multiExchangeData;
-    Array1D<AirflowNetwork::AirflowNetworkLinkReportData> linkReport;
-    Array1D<AirflowNetwork::AirflowNetworkNodeReportData> nodeReport;
-    Array1D<AirflowNetwork::AirflowNetworkLinkReportData> linkReport1;
+    EPVector<AirflowNetwork::AirflowNetworkExchangeProp> exchangeData;
+    EPVector<AirflowNetwork::AirflowNetworkExchangeProp> multiExchangeData;
+    EPVector<AirflowNetwork::AirflowNetworkLinkReportData> linkReport;
+    EPVector<AirflowNetwork::AirflowNetworkNodeReportData> nodeReport;
+    EPVector<AirflowNetwork::AirflowNetworkLinkReportData> linkReport1;
```
Myoldmopar, line 311: This is what I was expecting :+1:

src/EnergyPlus/Autosizing/Base.cc
```diff
@@ -236,7 +236,7 @@ void BaseSizer::preSize(EnergyPlusData &state, Real64 const _originalValue)
                 }
             }
         }
-        if (this->unitarySysEqSizing.allocated())
+        if (allocated(this->unitarySysEqSizing))
```
Myoldmopar, line 239: I take it this change allows EPVector to stay slimmed down by not needing a member method for `allocated`?  Then instead just include a header to access a utility function for it when needed?  Or something?

src/EnergyPlus/EPVector.hh
```diff
@@ -0,0 +1,293 @@
...
+
+    void deallocate() noexcept
+    {
+        this->clear();
+    }
+
+    [[nodiscard]] bool allocated() const noexcept
+    {
+        return m_allocated || !this->empty();
+    }
```
Myoldmopar, line 100: But then I do see the allocated member method here.  Can you clarify @lefticus?  Mostly just so I know what to look for if we spread this further into the codebase.

Myoldmopar (approved):
This all looks great to me.  I'm curious about the allocated change, but I don't think it needs any additional changes related to that, just a comment.

Myoldmopar:
Applied Clang Format to the code.  Should be all green now.  I also ran the unit test suite and everything passes, and the regression suite and no diffs are present.  This is good to go in.  @lefticus if you could just comment on my allocated question, (and if Github Actions is green now) I think we can mark this closed and continue down the path of improving build time in following branches.  Thanks!!

Myoldmopar:
And fun.  Clang Format is now green but the alternate build is failing.

Myoldmopar:
OK, so the only thing I could see that the Clang Format affected that could've broken the build was reordering the header includes.  Primarily reordering the EnergyPlus.hh and EPVector.hh includes.  I looked in EPVector.hh and found that it is indeed referencing an Array1D, yet no Objexx includes are listed there or forward declarations.  The EnergyPlus.hh has a bunch of type defs and such so I think there was an implicit inclusion of EnergyPlus.hh ahead of each use of EPVector.hh.  So I just directly added `include EnergyPlus.hh` to the EPVector file and we'll see if it is happy again.  It builds fine on my machine still either way.

Myoldmopar:
OK, I think this is ready to go.  I'll probably let Clang finish since it already started but then merge it in.

Myoldmopar:
Good to go.

PR #8712, opened 2021-04-13
Greenify CI Results

Myoldmopar:
HospitalLowEnergy is timing out every time, and not doing us any good at all, except showing as a false error that might muddy up some other actual failing test, and also eating a chunk of CI time.  This disables that extremely long running test for now, but obviously we want to fix it to just make it run faster, whether by cleaning the input file or improving the runtime performance.


Myoldmopar:
And of course that freaking file system test had to come in and ruin the greeniness of the PR.  :smile_cat: Assuming the integration build is happy when it is done in a few minutes this can merge in to turn the dashboard green again.

Myoldmopar:
Real missed opportunity there for 100% green, but it's green enough to clean up the dashboard.  Merging.

PR #8714, opened 2021-04-13
Correct heating coil sizing in AirTerminal:SingleDuct:ConstantVolume:Reheat

### LINKED ISSUE #8713: Heating coil in ConstantVolume:Reheat terminal unit is undersized when air flow rate is set by user
Issue overview
--------------
Third party testing has shown that heating coils are undersized when the user sets the terminal air flow rate and that user specified air flow rate is higher than the design value for that zone.

### Details

### Checklist
 - [x] Defect file added ([8713-5ZoneAirCooledConvCoef-Modified.txt](https://github.com/NREL/EnergyPlus/files/6313556/8713-5ZoneAirCooledConvCoef-Modified.txt))

rraustad:
Pull request overview
---------------------
 - Fixes #8713
 - Coil sizing for constant volume terminal units was using zone design air flow rate to size heating coil. This value could be significantly different from the actual air flow rate through the terminal unit.

### Pull Request Author
 - [ ] ~If any defect files are updated to a more recent version, upload new versions here or on DevSupport~

### Reviewer


rraustad:
The first diff in 5ZoneAirCooledConvCoef shows the change. The first line of this diff shows the TU air flow rate was sized to 0.34048 m3/s. However, the heating coil sizing was not using that value and is now. Air flow used to size the heating coil was 7.45604E-002 and is now 0.34048. This has the effect of increasing the heating coil capacity and UA.

      Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Air Flow Rate [m3/s], 0.34048
    - Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 5.35851E-005
    - Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 7.45604E-002
    + Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 7.88169E-005
    + Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 0.34048
      Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Temperature [C], 16.70000
      Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Humidity Ratio [kgWater/kgDryAir], 8.00000E-003
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 2425.34724
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 5.35851E-005
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 59.39590
    -Water Heating Coil Capacity Information,Coil:Heating:Water,SPACE5-1 ZONE COIL,2439.50
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 3567.38127
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 7.88169E-005
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 68.87058
    +Water Heating Coil Capacity Information,Coil:Heating:Water,SPACE5-1 ZONE COIL,3587.01

rraustad:
I am admittedly having a problem showing the need to increase the size of the TU water heating coil using 5ZoneAirCooledConvCoef as the test file. I've increased the amount of outdoor air and lowered the supply air temperature in heating mode in the Sizing:System object with no success. The third party defect file used a steam coil in the ConstantVolume:Reheat TU so maybe that had something to do with the issue. I'm still looking for validation of the code change other than what looks to be correct based on the code.

rraustad:
The HW coil is performing the same on the air-side, but the water-side delta T is about 25% lower now which is in the right direction. With an 11 C design plant delta T it seems that this coil could be a little larger to meet that design intent (or at least a larger water flow rate). SPACE5-1 IN NODE is the outlet of the HW heating coil and zone inlet node.

  Sizing:Plant,
    Hot Water Loop,          !- Plant or Condenser Loop Name
    heating,                 !- Loop Type
    82.,                     !- Design Loop Exit Temperature {C}
    11;                      !- Loop Design Temperature Difference {deltaC}


image

src/EnergyPlus/SingleDuct.cc
```diff
@@ -1219,7 +1219,7 @@ void GetSysInput(EnergyPlusData &state)
         state.dataSingleDuct->sd_airterminal(state.dataSingleDuct->SysNumGSI).ReheatAirOutletNode =
             state.dataSingleDuct->sd_airterminal(state.dataSingleDuct->SysNumGSI).OutletNodeNum;
         state.dataSingleDuct->sd_airterminal(state.dataSingleDuct->SysNumGSI).MaxAirVolFlowRate = Numbers(1);
-        state.dataSingleDuct->sd_airterminal(state.dataSingleDuct->SysNumGSI).ZoneMinAirFracDes = 0.0;
+        state.dataSingleDuct->sd_airterminal(state.dataSingleDuct->SysNumGSI).ZoneMinAirFracDes = 1.0;
```
rraustad: For constant volume terminal units this value, ZoneMinAirFracDes, is used to set the air flow used for sizing the heating coil. If this value is 0 then the air flow rate used to size the heating coil is always equal to NonAirSysDesHeatVolFlow which is the zone design air flow rate, which could be significantly different from the flow rate of the TU(e.g. user specified flow, etc.). This value was changed to 1 for all TUs that did not have an input for Minimum Air Flow Fraction. The NoReheat version does not have a heating coil and this wouldn't matter, but for consistency, all were changed.

    TermUnitSizing(CurTermUnitSizingNum).AirVolFlow =
        max(TermUnitFinalZoneSizing(CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
            this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2814,3 +2814,166 @@ TEST_F(EnergyPlusFixture, setATMixerSizingProperties_Test)
...
+    state->dataSize->ZoneSizingRunDone = true;
+    state->dataSize->CurZoneEqNum = 1;
+    state->dataSize->CurTermUnitSizingNum = 1;
+    state->dataSize->TermUnitFinalZoneSizing(1).NonAirSysDesHeatVolFlow = 0.5;
+    state->dataSingleDuct->sd_airterminal(SysNum).SizeSys(*state);
+    // constant volume TUs should have ZoneMinAirFracDes = 1 (now set correctly in GetSysInput)
+    EXPECT_EQ(1.0, state->dataSingleDuct->sd_airterminal(SysNum).ZoneMinAirFracDes);
+    // if ZoneMinAirFracDes = 0, TermUnitSizing(1).AirVolFlow would = 0.5 (NonAirSysDesHeatVolFlow)
+    // since ZoneMinAirFracDes = 1, TermUnitSizing(1).AirVolFlow = hard sized value (i.e., 1.0)
+    EXPECT_EQ(1.0, state->dataSize->TermUnitSizing(1).AirVolFlow);
```
rraustad, line 2894: Tests NoReheat version.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2814,3 +2814,166 @@ TEST_F(EnergyPlusFixture, setATMixerSizingProperties_Test)
...
+                                                              state->dataSingleDuct->sd_airterminal(SysNum).CtrlZoneNum);
+    // TUSizing.AirVolFlow gets set like this. If ZoneMinAirFracDes = 0, then autosized air flow is passed to heating coil
+    // this->MaxAirVolFlowRate = 1.0 here which is set by user. This is a factor of 2 higher than design/auosized value
+    // TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
+    //     max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
+    //         this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
+
+    EXPECT_GT(actualCoilSize, desCoilSize);
+    EXPECT_NEAR(actualCoilSize, 12192.0, 0.1);
+    EXPECT_NEAR(actualCoilSize, 2.0 * desCoilSize, 0.1); // heating coil is twice the size of autosized value
```
rraustad, line 3014: Tests coil sizing for Reheat version.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2814,3 +2814,166 @@ TEST_F(EnergyPlusFixture, setATMixerSizingProperties_Test)
...
+    state->dataSize->ZoneSizingRunDone = true;
+    state->dataSize->CurZoneEqNum = 1;
+    state->dataSize->CurTermUnitSizingNum = 1;
+    state->dataSize->TermUnitFinalZoneSizing(1).NonAirSysDesHeatVolFlow = 0.5;
+    state->dataSingleDuct->sd_airterminal(SysNum).SizeSys(*state);
+    // constant volume TUs should have ZoneMinAirFracDes = 1 (now set correctly in GetSysInput)
+    EXPECT_EQ(1.0, state->dataSingleDuct->sd_airterminal(SysNum).ZoneMinAirFracDes);
+    // if ZoneMinAirFracDes = 0, TermUnitSizing(1).AirVolFlow would = 0.5 (NonAirSysDesHeatVolFlow)
+    // since ZoneMinAirFracDes = 1, TermUnitSizing(1).AirVolFlow = hard sized value (i.e., 1.0)
+    EXPECT_EQ(1.0, state->dataSize->TermUnitSizing(1).AirVolFlow);
```
rraustad, line 2983: Same tests as above for NoReheat version.

src/EnergyPlus/SingleDuct.cc
dareumnam: I agree.

dareumnam (commented):
This looks reasonable. I built and tested this locally and the tests look great, If there is nothing to update with Engineering Reference Documentation, I think this is good to go in. @Myoldmopar

rraustad:
Thank you @dareumnam for the review. @mjwitte I would like your thoughts on this change.

mjwitte:
@rraustad So, I started searching through SingleDuct.cc for `ZoneMinAirFracDes` to understand how it's used. And I don't understand much yet. I'll try again next week. 

The change in coil capacity [above](https://github.com/NREL/EnergyPlus/pull/8714#issuecomment-819061303) is quite large from 2425 to 3567W. On the design day, what's the peak load on that reheat coil vs. the capacity (before and after this fix)?

rraustad:
What are the sizing parameters for a heating coil? Sufficient capacity to meet the load and maintain zone temperature. But shouldn't there also be an outlet water temperature specification/limit ? We could get the same amount energy out of a HW coil as a smaller coil with 2x the water-side delta T. So which size is correct? All we have ever looked at to determine if heating capacity is sufficient is zone temperature. In this case the coil air-side performance is the same, but water-side performance differs. I'm trying to talk myself out of this change but I keep coming back to what appears to be a correct adjustment to heating capacity for this TU type. Would a Mech Eng size a HW coil to give a 34 C water temperature drop? I doubt it.

5ZoneAirCooledConvCoef

    Zone Sizing Information, SPACE1-1, Heating, 2565.55580, 3335.22254, 8.84572E-002, 0.11499, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 9.96609E-004, 99.16000, 11.00000, 0.10384, 0.00000
    Zone Sizing Information, SPACE2-1, Heating, 1125.12599, 1462.66378, 3.87930E-002, 5.04309E-002, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 9.96609E-004, 42.73500, 5.00000, 4.72000E-002, 0.00000
    Zone Sizing Information, SPACE3-1, Heating, 2533.92356, 3294.10063, 8.73665E-002, 0.11358, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 9.96609E-004, 96.48000, 11.00000, 0.10384, 0.00000
    Zone Sizing Information, SPACE4-1, Heating, 1123.37225, 1460.38393, 3.87325E-002, 5.03523E-002, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 9.96609E-004, 42.73500, 5.00000, 4.72000E-002, 0.00000
    Zone Sizing Information, SPACE5-1, Heating, 1627.72138, 2116.03780, 5.61217E-002, 0.18880, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 9.96609E-004, 182.49000, 20.00000, 0.18880, 0.00000

It's interesting that the heating coil sizes are far greater than the zone heating loads. The 1st 4 TUs are AirTerminal:SingleDuct:ParallelPIU:Reheat, the 5th is AirTerminal:SingleDuct:ConstantVolume:Reheat.

    Component Sizing Information, Coil:Heating:Water, SPACE1-1 ZONE COIL, Design Size Rated Capacity [W], 9650.92968
    Component Sizing Information, Coil:Heating:Water, SPACE2-1 ZONE COIL, Design Size Rated Capacity [W], 9156.67916
    Component Sizing Information, Coil:Heating:Water, SPACE3-1 ZONE COIL, Design Size Rated Capacity [W], 9141.66308
    Component Sizing Information, Coil:Heating:Water, SPACE4-1 ZONE COIL, Design Size Rated Capacity [W], 10319.21407
    Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 3578.35767

Heating rate is 1897 W on the winter design day. Also note the water outlet node temp is at 48 C which seems low to me since it should be closer to 70 C. In this branch leaving water temp is 59 C. That seems more reasonable.

  Sizing:Plant,
    Hot Water Loop,          !- Plant or Condenser Loop Name
    heating,                 !- Loop Type
    82.,                     !- Design Loop Exit Temperature {C}
    11;                      !- Loop Design Temperature Difference {deltaC}


image

I was trying to find info on what a typical leaving water temp is for HW coils. I found https://www.trane.com/content/dam/Trane/Commercial/global/products-systems/education-training/engineers-newsletters/waterside-design/en_26-01.pdf which shows this table. The 116 F leaving water temp shown here highlights a problem for boilers when return water temps fall below 90F (32.2C).  I've never really thought about the implications of HW coil leaving water temps. If all the HW coils had a return temp of 48 C then the plant would not be sized to meet that HW plant load.

image

mjwitte:
> In this case the coil air-side performance is the same, but water-side performance differs.
That says to me that the coil is adequately sized in both cases and the terminal unit "controller" is doing it's job.

>But shouldn't there also be an outlet water temperature specification/limit ?
Yes. If plant sizing says the deltaT at the design load is 11C with 82C HW temp, the I'd expect something very close to 11C at the design condition.

I'm still not following all of these changes correctly, but my understanding is that the air flow rate used for heating in this case will determine the design load but will also impact what the hot water heating coil uses when it's sizing the water side and internal UA value of the coil. In this case, the lower air flow rate is (maybe) correct for determining the design load on the coil, but that's not the airflow rate the coil will see when operating (since this is constant volume). Does the hot water coil sizing know this?

And in general, the coil water-side delta-T has been corrected in the past when the "Rated Ratio for Air and Water Convection" field was added. Plus this coil is over specified having 4 temperature inputs that may or may not align with the plant sizing inputs? If there's any need for improvement here, perhaps the issue is in the HW coil model/sizing.

Also, didn't you say this fix was prompted by a user file with a steam coil?

rraustad:
Yes a steam coil. But as far as I can tell the HW and steam coils are sized the same, except for the conversion of steam in the steam flow calculation. I think I'll update this defect file to run in V9.5 and recreate this problem here. Before and after of zone temp and load for the TU in question.

image

rraustad:
I am getting the same result in V9.5. The steam coil capacity is too small. I don't see a difference between water coil and steam coil fluid sizing. They both use the same air volume flow and account for the zone load, NonAirSysDesHeatLoad. The large water-side delta T bothers me and I think points to an undersized HW water coil.  So is it only the steam coil that is undersized or are we missing something for sizing any heating coil type?

    DesMassFlow = state.dataEnvrn->StdRhoAir * TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow;
    state.dataSingleDuct->DesZoneHeatLoadSS =
                            state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatLoad;


    state.dataSingleDuct->DesCoilLoadSS =
        state.dataSingleDuct->DesZoneHeatLoadSS + PsyCpAirFnW(state.dataSingleDuct->ZoneDesHumRatSS) * DesMassFlow *
            (state.dataSingleDuct->ZoneDesTempSS - state.dataSingleDuct->CoilInTempSS);

    MaxReheatWaterVolFlowDes =
        state.dataSingleDuct->DesCoilLoadSS / (state.dataSize->PlantSizData(PltSizHeatNum).DeltaT * Cp * rho);

    TempSteamIn = 100.00;
    EnthSteamInDry = GetSatEnthalpyRefrig(state, fluidNameSteam, TempSteamIn, 1.0, this->FluidIndex, RoutineNameFull);
    EnthSteamOutWet = GetSatEnthalpyRefrig(state, fluidNameSteam, TempSteamIn, 0.0, this->FluidIndex, RoutineNameFull);
    LatentHeatSteam = EnthSteamInDry - EnthSteamOutWet;
    SteamDensity = GetSatDensityRefrig(state, fluidNameSteam, TempSteamIn, 1.0, this->FluidIndex, RoutineNameFull);

    MaxReheatSteamVolFlowDes = state.dataSingleDuct->DesCoilLoadSS /
        (SteamDensity * (LatentHeatSteam + state.dataSize->PlantSizData(PltSizHeatNum).DeltaT * Cp));

image

Defect file uses La Crosse weather. 
[8714-in.txt](https://github.com/NREL/EnergyPlus/files/6417242/8714-in.txt)

rraustad:
The above is Jan 6 - Jan 19

rraustad:
Steam coil size changed from:

    Component Sizing Information, Coil:Heating:Steam, GUID10_1_E_CVR-1_HEATING COIL, Maximum Steam Flow Rate [m3/s], 5.88003E-003

To:

    Component Sizing Information, Coil:Heating:Steam, GUID10_1_E_CVR-1_HEATING COIL, Maximum Steam Flow Rate [m3/s], 7.45719E-002

nrel-bot-2:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

mjwitte:
@rraustad Looking at the HW test file,  8713-5ZoneAirCooledConvCoef-Modified, this change seems wrong. 
With develop, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 2822.83314
With this branch, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 4946.33405
And with both branches, SPACE5-1 ZONE COIL:Heating Coil Heating Rate [W] is 1928 W on the winter design day.
The zone sizing factor is 1.3 x 1928 = 2506.
So, the coil is already slightly oversized in develop, and way oversized with this branch.

I think focusing on the water deltaTs is a distraction here (that may need some attention but some other way). 

The oversized coil water flow rate also causes the boiler to be even larger.
At 53,216W (with develop) with a winter design day load of 10,630W it's already way oversized. Not sure why, but making it ever larger isn't good.
I'll look at the steam file next.

mjwitte:
Part of the reason the HW coil is oversized is because the design heating supply temp (entering the reheat coil) is 12.8C and that's what is used in sizing the coil, but the simulation results show that VAV SYS 1 OUTLET NODE:System Node Temperature is 18.294C on the winter design day. Just noting that here for now.

And, the sizing thinks the DesZoneHeatLoadSS is 2193W, but the output shows SPACE5-1:Zone Air System Sensible Heating Rate = 1002W. Hmmm.

rraustad:
So maybe it's steam coil specific? I keep coming back to a change in TU flow rate (i.e., hard-size higher than autosize) has to impact htg coil capacity. As airflow increases, coil outlet T decreases (at constant capacity) and eventually zone T cannot be maintained. Thanks for looking.

nrel-bot-2b:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-2b:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

Myoldmopar:
@rraustad @mjwitte from a quick scan I can't quite tell from the last comments whether the issues are resolved or if there still needs to be some investigation.  I just want to know if I should pull this up to develop to push toward review/merge.  Thanks!

rraustad:
@Myoldmopar @mjwitte put this one last on the list given that the change here affects coil sizing. A 3rd party vendor, based on a coil sizing concern, is using these changes now and has not highlighted any issues. These changes may be correct, but I don't want these changes accepted without some thought/investigation, etc.

Myoldmopar:
How far down on the list does this go?  Does it get punted until later after release?  If so we should go ahead and change the milestone to trim down the review queue.

rraustad:
Punted.

nrel-bot-3:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@rraustad @lgentile it has been 28 days since this pull request was last updated.

src/EnergyPlus/SingleDuct.cc
mjwitte: @rraustad Ok, I see how this makes sense. For singleduct, an alternative fix would be to add an if block for the 2 constant volume terminal units [here](https://github.com/NREL/EnergyPlus/blob/b03bdf040164a5ca370b6ea69bf41e08423ad0f6/src/EnergyPlus/SingleDuct.cc#L3493-L3504) similar to the one a few lines below. Or maybe not, maybe that if can be removed from lines 3502-03 because `this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac` now = 1 for these terminal unit types?

In general, this sizing function has lots of overlapping and/or incomplete paths, sometimes testing on `SysType_Num`, sometimes `DamperHeatingAction` (but this doesn't always (ever?) have a case for `HeatingActionNotUsed` which these terminal unit types are set to, sometimes on `ZoneMinAirFracMethod`, sometimes these are nested. I'm not suggesting this be refactored here, just saying.  

Now to look at the big diffs more closely to be certain they are as desired.

mjwitte:
Reviewing diffs.
Defect file 8714-in-Annual.idf, steam reheat coils are 40-60% larger, heating not met hours drop from 5456 hrs to 436 hrs, gas consumption goes up 25%. All seems reasonable since the base case was obviously undersized and not meeting setpoint.

Defect file 8713-5ZoneAirCooledConvCoef-Modified-Annual.idf, SPACE5-1 hot water reheat coil is 75% larger, tiny increase in gas use, before and after both have zero hrs of heating setpoint not met. So, the larger reheat coil seems unnecessary. Need to look at this again.

mjwitte:
Part of the problem with 8713-5ZoneAirCooledConvCoef-Modified-Annual.idf is that the supply air setpoint is 16C in winter, but the Sizing:System Central Heating Design Supply Air Temperature is 12.8C so the reheat coils are oversized to begin with. Will test again with consistent sizing to see if the old method fails to meet setpoint.

mjwitte:
So, changed the Heating Design Supply Temp to 16, also changed the Sizing factors to 1.0, and still no unmet heating setpoint hours with this branch or with develop. Odd.
Space 5-1 Zone Coil - Before: Rated Capacity 1988W, peak annual heating output 4025W?!
This branch: Rated Cap 2992W, peak annual heating output 3082W - huh? 

The peak reheat coil loads are typically at 8am recovery from setback. So, why would a larger coil end up with a smaller peak output? And why with the original sizing can the reheat coil put out 2x it's rated capacity? Need to review the earlier discussions about hot water coil sizing. If I could get this to have unmet heating hours with develop and fewer with this branch, I'd be confident in the fix.

rraustad:
Confidence is key. If I had good evidence that this was the wrong fix I'd happily revert this change. Then I would try to prove that the coils were undersized or find the reason they could meet the loads with existing sizing. If I recall, the defect file was from Trane and this issue only showed up with a certain configuration (or they never noticed it before, I do recall it was an unmet hours issue). The defect file here looks to be a modified version of an existing example file.

rraustad:
I do appreciate the review and you have found some interesting results. Maybe for the sake of time I should resurrect the Trane defect file, convert it to E+, and recreate the original issue.

mjwitte:
I have the (or some) defect file with a steam coil that does [show expected results](https://github.com/NREL/EnergyPlus/pull/8714#issuecomment-1016972147). So the only puzzle here is the hot water reheat coil from 5ZoneAirCooledConvCoef....

rraustad:
I did find the old defect file. That file was converted to E+ in [this comment](https://github.com/NREL/EnergyPlus/pull/8714#issuecomment-831485797) above. The only difference I see between the original defect file and this one is the Sizing:Zone Heating Maximum Air Flow Fraction. I changed it from 0 to 0.3, I can't recall why. That same comment also has some plots so maybe that is sufficient as far as a before and after plot. I did see in the original defect description that the air loop air flow was hard sized while the zone equipment was autosized. Maybe that has something to do with what fix is implemented.

mjwitte:
Back to the HW reheat coil in 
[8713-5ZoneAirCooledConvCoef-Modified-Annual-FixSizingTemp.idf.txt](https://github.com/NREL/EnergyPlus/files/7908261/8713-5ZoneAirCooledConvCoef-Modified-Annual-FixSizingTemp.idf.txt)
Interesting that the UA does not increase in the same proportion as the water flow rate and rated capacity, hmmm.
```
Develop:  
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Air Flow Rate [m3/s], 0.21475
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 4.39221E-005
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 4.91278E-002
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Temperature [C], 16.00000
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Humidity Ratio [kgWater/kgDryAir], 8.00000E-003
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 1987.98338
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 4.39221E-005
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 48.32542
Water Heating Coil Capacity Information,Coil:Heating:Water,SPACE5-1 ZONE COIL,1976.91

This branch:
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Air Flow Rate [m3/s], 0.21475
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 6.61126E-005
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 0.21475
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Temperature [C], 16.00000
 Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Inlet Air Humidity Ratio [kgWater/kgDryAir], 8.00000E-003
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 2992.36254
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 6.61126E-005
 Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 57.53551
Water Heating Coil Capacity Information,Coil:Heating:Water,SPACE5-1 ZONE COIL,2979.93

File: src/EnergyPlus/DualDuct.cc [hunk 5]
```diff
@@ -316,7 +316,7 @@ namespace DualDuct {
                                                                                                   cAlphaFields(5));
 
                 state.dataDualDuct->dd_airterminal(DDNum).MaxAirVolFlowRate = NumArray(1);
-                state.dataDualDuct->dd_airterminal(DDNum).ZoneMinAirFracDes = 0.0;
+                state.dataDualDuct->dd_airterminal(DDNum).ZoneMinAirFracDes = 1.0;
```
mjwitte: Ran a local regression suite with annual simulations and monthly output. Somehow this change is impacting dual duct even through there is no reheat coil. DDAutoSize shows *no* sizing diffs but the annual energy use is way up, district heating went from 76 to 306 GJ.

rraustad: That's not a good sign. I did not focus on DD when making this change, just thought if the SD TUs needed this change then so goes the dual duct TUs. I will have to review how the ZoneMinAirFracDes is used in DD sizing. This is why unexplained changes in sizing are NEVER merged with develop until there is full consensus from the team.

rraustad:
It appears that the DD system, at least with existing logic, cannot set the ZoneMinAirFracDes =1 since there are 2 ducts and one duct will have priority over the other (wrt massflowrate) and the unneeded duct flow rate should go towards 0. Here the min avails are set for each duct type.

                // update to the current dual duct minimum air flow fraction
                this->ZoneMinAirFrac = this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac;
                // The last item is to take care of the Min Avail Flow Rates
                if ((state.dataLoopNodes->Node(HotInNode).MassFlowRate > 0.0) && (GetCurrentScheduleValue(state, this->SchedPtr) > 0.0)) {
                    state.dataLoopNodes->Node(HotInNode).MassFlowRateMinAvail =
                        this->dd_airterminalHotAirInlet.AirMassFlowRateMax * this->ZoneMinAirFrac;
                } else {
                    state.dataLoopNodes->Node(HotInNode).MassFlowRateMinAvail = 0.0;
                }
                if ((state.dataLoopNodes->Node(ColdInNode).MassFlowRate > 0.0) && (GetCurrentScheduleValue(state, this->SchedPtr) > 0.0)) {
                    state.dataLoopNodes->Node(ColdInNode).MassFlowRateMinAvail =
                        this->dd_airterminalColdAirInlet.AirMassFlowRateMax * this->ZoneMinAirFrac;
                } else {
                    state.dataLoopNodes->Node(ColdInNode).MassFlowRateMinAvail = 0.0;
                }

Then in the calc routine, the cold air duct mass flow rate is set according to the load. Then the hot air duct should be set to the remaining flow such that the total flow is not greater than AirMassFlowRateMaxAvail. Since the dual duct system modulates flow in each duct to maintain the total flow it appears that, without other logic added, the dual duct system does have a min flow = 0 (for one or the other ducts). In this branch, both ducts are operating at the max flow rate since min avail and max avail = max. See last 2 if blocks checking against max and min avail and the final calculation for MassFlow = sum of duct flow. That mass flow rate should not be more than the max flow through one of the ducts (since the sum of both ducts should equal the max flow through one of the ducts). The DD system modulates the flow thorugh each duct to keep the total flow constant.

I think the best thing to do for now is to revert the change to ZoneMinAirFracDes for DD systems.

        // Get the calculated load from the Heat Balance from ZoneSysEnergyDemand
        QTotLoad = state.dataZoneEnergyDemand->ZoneSysEnergyDemand(ZoneNum).RemainingOutputRequired;
        // Need the design MassFlowRate for calculations
        if (GetCurrentScheduleValue(state, this->SchedPtr) > 0.0) {
            MassFlow = this->dd_airterminalHotAirInlet.AirMassFlowRateMaxAvail / 2.0 + this->dd_airterminalColdAirInlet.AirMassFlowRateMaxAvail / 2.0;
        } else {
            MassFlow = 0.0;
        }
        // If there is massflow then need to provide the correct amount of total
        //  required zone energy
        if (MassFlow > SmallMassFlow) {
            CpAirZn = PsyCpAirFnW(state.dataLoopNodes->Node(ZoneNodeNum).HumRat);
            QZnReq = QTotLoad + MassFlow * CpAirZn * state.dataLoopNodes->Node(ZoneNodeNum).Temp;
            // If the enthalpy is the same for the hot and cold duct then there would be a
            //  divide by zero so for heating or cooling set the damper to one max flow
            //  or the other.
            if (std::abs(this->dd_airterminalColdAirInlet.AirTemp - this->dd_airterminalHotAirInlet.AirTemp) > SmallTempDiff) {
                // CpAirSysHot = PsyCpAirFnWTdb(dd_airterminalHotAirInlet(DDNum)%AirHumRat,dd_airterminalHotAirInlet(DDNum)%AirTemp)
                // CpAirSysCold= PsyCpAirFnWTdb(dd_airterminalColdAirInlet(DDNum)%AirHumRat,dd_airterminalColdAirInlet(DDNum)%AirTemp)
                CpAirSysHot = CpAirZn;
                CpAirSysCold = CpAirZn;
                // Determine the Cold Air Mass Flow Rate
                this->dd_airterminalColdAirInlet.AirMassFlowRate =
                    (QZnReq - MassFlow * CpAirSysHot * this->dd_airterminalHotAirInlet.AirTemp) /
                    (CpAirSysCold * this->dd_airterminalColdAirInlet.AirTemp - CpAirSysHot * this->dd_airterminalHotAirInlet.AirTemp);
            } else if ((QTotLoad > 0.0) && (this->dd_airterminalHotAirInlet.AirMassFlowRate > 0.0)) {
                this->dd_airterminalColdAirInlet.AirMassFlowRate = 0.0;
            } else {
                this->dd_airterminalColdAirInlet.AirMassFlowRate = MassFlow;
            }
            // Check to make sure that the calculated flow is not greater than the available flows
            if (this->dd_airterminalColdAirInlet.AirMassFlowRate > this->dd_airterminalColdAirInlet.AirMassFlowRateMaxAvail) {
                this->dd_airterminalColdAirInlet.AirMassFlowRate = this->dd_airterminalColdAirInlet.AirMassFlowRateMaxAvail;
            } else if (this->dd_airterminalColdAirInlet.AirMassFlowRate < this->dd_airterminalColdAirInlet.AirMassFlowRateMinAvail) {
                this->dd_airterminalColdAirInlet.AirMassFlowRate = this->dd_airterminalColdAirInlet.AirMassFlowRateMinAvail;
            }
            // Using Mass Continuity to determine the other duct flow quantity
            this->dd_airterminalHotAirInlet.AirMassFlowRate = MassFlow - this->dd_airterminalColdAirInlet.AirMassFlowRate;
            if (this->dd_airterminalHotAirInlet.AirMassFlowRate > this->dd_airterminalHotAirInlet.AirMassFlowRateMaxAvail) {
                this->dd_airterminalHotAirInlet.AirMassFlowRate = this->dd_airterminalHotAirInlet.AirMassFlowRateMaxAvail;
            } else if (this->dd_airterminalHotAirInlet.AirMassFlowRate < this->dd_airterminalHotAirInlet.AirMassFlowRateMinAvail) {
                this->dd_airterminalHotAirInlet.AirMassFlowRate = this->dd_airterminalHotAirInlet.AirMassFlowRateMinAvail;
            }
            MassFlow = this->dd_airterminalColdAirInlet.AirMassFlowRate + this->dd_airterminalHotAirInlet.AirMassFlowRate;
        } else {
            // System is Off set massflow to 0.0
            MassFlow = 0.0;
        }

rraustad:
This branch is now specific to constant volume single duct TUs. Let's be careful to be sure this change is correct for sizing and performance for that TU type.

mjwitte:
@rraustad Re-ran annual regressions and summarized the changes in energy use and number of heating hours not met. 2 files are as expected - more heating energy and lower heating hrs not met. But about 10 files had significantly *more* heating hours not met. Will look further.
[2022-01-24 Annual Simulation Diffs.xlsx](https://github.com/NREL/EnergyPlus/files/7935359/2022-01-24.Annual.Simulation.Diffs.xlsx)

mjwitte:
Hold that thought. At least one of these was a typo in the summary spreadsheet.

mjwitte:
@rraustad Another clue. About half the files showing big diffs in energy use are *not* autosized, or they have autosized components that don't change. So perhaps the same as the dual duct, the min fraction variable has other side effects? Revised spreadsheet showing which files are not autosized or have no sizing diffs.
[2022-01-24 Annual Simulation Diffs.xlsx](https://github.com/NREL/EnergyPlus/files/7935739/2022-01-24.Annual.Simulation.Diffs.xlsx)

rraustad:
I'll have a look.

rraustad:
ZoneWSHP_wDOAS - Things look roughly the same except this branch (left figure in each figure set) shows a slightly higher zone temp, WSHP mass flow rate and WSHP coil heating energy than develop (right figure), This is probably due to the DOAS system now providing the rated flow rate at 5x of what develop provides because it uses AirTerminal:SingleDuct:ConstantVolume:NoReheat. So it looks like it's working correctly now, but those are certainly famous last words. The OA air flow increased because I changed the TU min flow fraction, but the OA controller should have done that based on it's inputs?

Now to figure out what the DOAS system should do regarding OA flow modulation. Maybe this is the wrong TU type for this configuration? Should that TU type allow a varied flow rate set by the parent or not? This input shows a constant OA flow rate but what if it were not constant? And then what about cycling fan cases for other system types where air flow is proportional to PLR? I guess this TU type could accept varying flow if we allowed it to (i.e., reverted my change). But then how to fix the defect file issue, an undersized heating coil because of the flow chosen for this TU type? I wonder if this OA system allowed the flow to vary (to test flow rates below the TU min flow), how would this TU respond?

image

    Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:NoReheat, ZONE1DIRECTAIR, Design Size Maximum Air Flow Rate [m3/s], 1.21596
    Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:NoReheat, ZONE1DIRECTAIR, User-Specified Maximum Air Flow Rate [m3/s], 6.25000E-002

    Controller:OutdoorAir,
      OA Controller 1,         !- Name
      Relief Air Outlet Node,  !- Relief Air Outlet Node Name
      Outdoor Air Mixer Inlet Node,  !- Return Air Node Name
      Mixed Air Node,          !- Mixed Air Node Name
      Outside Air Inlet Node,  !- Actuator Node Name
      0.0625,                  !- Minimum Outdoor Air Flow Rate {m3/s}
      0.0625,                  !- Maximum Outdoor Air Flow Rate {m3/s}

    SetpointManager:MixedAir,
      Mixed Air Temp Manager 1,!- Name
      Temperature,             !- Control Variable
      Air Loop Outlet Node,    !- Reference Setpoint Node Name
      Mixed Air Node,          !- Fan Inlet Node Name
      Air Loop Outlet Node,    !- Fan Outlet Node Name
      Mixed Air Node;          !- Setpoint Node or NodeList Name

    SetpointManager:SingleZone:Reheat,
      Supply Air Temp Manager 1,  !- Name
      Temperature,             !- Control Variable
      14.0,                    !- Minimum Supply Air Temperature {C}
      23.0,                    !- Maximum Supply Air Temperature {C}
      SPACE1-1,                !- Control Zone Name
      Zone 1 Node,             !- Zone Node Name
      Zone 1 OA Inlet Node,    !- Zone Inlet Node Name
      Air Loop Outlet Node;    !- Setpoint Node or NodeList Name


image

image

rraustad:
I did try changing the min OA flow and then max OA flow. The mixed air node did not change but the OA node flow did change and tracked the min flow rate input. So here, the TU is setting the return/mixed node flow rates since the TU min flow frac was changed. This is a passive air loop with no parent on the main branch, except the fan, so the TU should set the flow. Now I'm sure there are example files with unitary systems (all types) on the main branch that use cycling fan with this TU type. No diffs in those files tells me that flow can vary in those cases otherwise we would have seen diffs. It does make me wonder if this change will choke flow in some cases (choked at the max flow that is). This also shows some issue with this example file regarding the OA system flow rate, it should not have been that low regardless of TU type.

mjwitte:
Here's the terminal unit and the DSOA spec:
```
  AirTerminal:SingleDuct:ConstantVolume:NoReheat,
    Zone1DirectAir,          !- Name
    FanAndCoilAvailSched,    !- Availability Schedule Name
    Zone 1 OA Inlet Node ATInlet,  !- Air Inlet Node Name
    Zone 1 OA Inlet Node,    !- Air Outlet Node Name
    0.0625,                  !- Maximum Air Flow Rate {m3/s}
   <this is blank in the file - the default is CurrentOccupancy which may vary> ;                        !- Per Person Ventilation Rate Mode

    SUM,                     !- Outdoor Air Method
    0.00015,                 !- Outdoor Air Flow per Person {m3/s-person}
    0.00005,                 !- Outdoor Air Flow per Zone Floor Area {m3/s-m2}
```

If there is a DSOA spec on this terminal unit, then it should act like a VAV terminal and request a specific flow rate but then the airloop can alter that within max/min limits. Min flow rate should be allowed to go to zero in that case. If there is not a DSOA spec on this terminal unit, then it should be passive and take whatever flow it gets (like the old ...Uncontrolled terminal unit).

But this file has a Fan:ConstantVolume and there's only one zone on the DOAS, and the fan has a hard flow rate of 0.0625, so the DOAS should be at 0.0625 whenever it's running.  But it's not (in develop).  The flow is actually varying, so maybe Fan:ConstantVolume honors the flow request?  

The expected DOAS flow rate is 0.00005 * 232.25760 m2 = 0.1161288 m3/s when unoccupied (which is what we see with develop). And it should be 0.1161288 + 0.00015 * Number of occupants. That's what we get with develop. So, I guess Fan:ConstantVolume adjusts to match the terminal unit limits, which is maybe unexpected, but that is the intent here. Probably should change to a Fan:VariableVolume or Fan:SystemModel, but not here.

With this branch, the  Zone 1 OA Inlet Node flow rate is always 0.0625, the design flow rate. So that explains it.

For this particular terminal unit, AirTerminal:SingleDuct:ConstantVolume:NoReheat, it should be allowed to go below the design flow rate. It could be a passive terminal unit on, say, a multizone furnace, or it could be requesting variable flow rates with a DSOA specification.

For other constant volume terminal units, they really should be passive as well and take whatever the airloop gives them.

mjwitte:
So, perhaps it's better to leave `ZoneMinAirFracDes` as-is everywhere and modify only the [reheat coil sizing block](https://github.com/NREL/EnergyPlus/blob/fadc5973b85c70e8cc923efb69c144e808a26078/src/EnergyPlus/SingleDuct.cc#L3427-L3458) in SingleDuct.cc

Looking at this, there are two similar main paths, one `if (state.dataSize->ZoneSizingRunDone)` and an else. But the top block is missing several special cases that are in the second block:
```
            } else if (this->SysType_Num == SysType::SingleDuctConstVolReheat || this->SysType_Num == SysType::SingleDuctConstVolNoReheat) {
                if (this->DamperHeatingAction == Action::Reverse) {
                } else if (this->DamperHeatingAction == Action::ReverseWithLimits) {
```
Seems these cases should be handled differently in the top block as well.

rraustad:
So I should revert this change and select the flow rate passed to the heating coil via TermUnitSizing by expanding this if block and removing `* this->ZoneMinAirFracDes` ?

    if (this->SysType_Num == SysType::SingleDuctVAVReheatVSFan) {
        TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
            max(state.dataSingleDuct->UserInputMaxHeatAirVolFlowRateSS,
                state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
    } else {
        TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
            max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
    }

    enum class SysType
    {
        Invalid = -1,
        SingleDuctVAVReheat,
        SingleDuctConstVolReheat,
        SingleDuctConstVolNoReheat,
        SingleDuctVAVNoReheat,
        SingleDuctVAVReheatVSFan,
        SingleDuctCBVAVReheat,
        SingleDuctCBVAVNoReheat,
        Num,
    };

    switch (this->SysTypeNum) {
    case SysType::SingleDuctVAVReheatVSFan: {
        TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
            max(state.dataSingleDuct->UserInputMaxHeatAirVolFlowRateSS,
                state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
    } break;
    case SingleDuctConstVolReheat:
    case SingleDuctCBVAVReheat: {
        TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
            max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                this->MaxAirVolFlowRate * this->ZoneTurndownMinAirFrac);
    } break;
    default: { // SingleDuctVAVReheat and all NoReheat terminal unit types
        TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
            max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
    } break;
    }

mjwitte:
Yes, that's the proposal, but I think you also need to add ifs for `DamperHeatingAction == Action::Reverse and Action::ReverseWithLimits` for `SingleDuctVAVReheat`.

rraustad:
I see that. And yes I think the math regarding air flow should match between a sizing and hard sized run. Let's see what happens to the diffs with this change.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -3428,15 +3428,30 @@ void SingleDuctAirTerminal::SizeSys(EnergyPlusData &state)
         TermUnitSizing(state.dataSize->CurTermUnitSizingNum).ReheatAirFlowMult = 1.0;
         TermUnitSizing(state.dataSize->CurTermUnitSizingNum).ReheatLoadMult = 1.0;
         if (state.dataSize->ZoneSizingRunDone) {
+            // set air flow rate used to size heating coils
+            // ZoneTurndownMinAirFrac defaults to 1 for those TU types that do not use it
             if (this->SysType_Num == SysType::SingleDuctVAVReheatVSFan) {
                 TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
                     max(state.dataSingleDuct->UserInputMaxHeatAirVolFlowRateSS,
                         state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                         this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
```
rraustad, line 3436: Kept the if/else format below just so it would be easier to see changes

src/EnergyPlus/SingleDuct.cc
```diff
@@ -3428,15 +3428,30 @@ void SingleDuctAirTerminal::SizeSys(EnergyPlusData &state)
...
+                        max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
+                            this->MaxAirVolFlowRate * this->ZoneTurndownMinAirFrac);
+                } else if (this->DamperHeatingAction == Action::ReverseWithLimits) {
+                    TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
+                        max(this->MaxAirVolFlowRateDuringReheat, this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
+                } else {
+                    TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
+                        max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
+                            this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
+                }
```
rraustad: Note the ReverseWithLimits is the only one that does not compare to NonAirSysDesHeatVolFlow. I think that should be added but wasn't sure.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2517,7 +2517,7 @@ TEST_F(EnergyPlusFixture, SingleDuct_VAVWaterCoilSizing)
     EXPECT_EQ(DesZoneHeatLoad, 0.0);
     EXPECT_EQ(ZoneDesTemp, 17.0);
     EXPECT_EQ(ZoneDesHumRat, 0.008);
-    EXPECT_NEAR(DesCoilLoad, 120.5, 0.1);
+    EXPECT_NEAR(DesCoilLoad, 247.1, 0.1);
```
rraustad: And unit tests now change because DamperHeatingAction logic was added.

tst/EnergyPlus/unit/SizeWaterHeatingCoil.unit.cc
```diff
@@ -1080,7 +1080,7 @@ TEST_F(EnergyPlusFixture, TestSizingRoutineForHotWaterCoils4)
     state->dataSingleDuct->sd_airterminal(1).ZoneFloorArea = state->dataHeatBal->Zone(1).FloorArea;
     state->dataSingleDuct->sd_airterminal(1).SizeSys(*state);
     SizeWaterCoil(*state, 1);
-    EXPECT_NEAR(state->dataWaterCoils->WaterCoil(1).MaxWaterVolFlowRate, .0000850575, 0.000000001);
+    EXPECT_NEAR(state->dataWaterCoils->WaterCoil(1).MaxWaterVolFlowRate, .000094839, 0.000000001);
     EXPECT_NEAR(state->dataWaterCoils->WaterCoil(1).UACoil, 300.00, 0.01);
```
rraustad: Not sure why UA didn't change here as it did elsewhere.

rraustad:
5ZoneAirCooledConvCoef - the heating coil size increased in all categories. I think now I should figure out a way to prove that this increase is necessary.

    - Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 5.36759E-005
    + Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Reheat Water Flow Rate [m3/s], 7.89269E-005
    - Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 7.46867E-002
    + Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Reheat Coil Sizing Air Volume Flow Rate [m3/s], 0.34080
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 2429.45835
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 3572.35734
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 5.36759E-005
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Maximum Water Flow Rate [m3/s], 7.89269E-005
    - Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 59.49659
    + Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size U-Factor Times Area Value [W/K], 68.97126

Everything is looking the same, except the amount of reheat to the zone (first occurrence), and a little on zone temp at the same time. Note above the eio size of the water coil in develop is 2429 W and this branch is 3572 W. In the figure the reheat coil in develop (dotted) shows 4455 W, this branch 3694 W. So how can a 2492 W water coil yield 4455 W of output? 

image

The answer lies in something we have discussed in the past, water-side delta T. Which coil is acting more like the plant dT design data provided to the progam? This branch shows a 3572 W coil providing 3694 W. I would say this branch has a water-side delta T closer to design meaning the coil is sized properly. Unfortunately this example file does not have the water node data I need to prove that point but I think we all know what we would find. And now for the comments ...

    Sizing:Plant,
      Hot Water Loop,          !- Plant or Condenser Loop Name
      heating,                 !- Loop Type
      82.,                     !- Design Loop Exit Temperature {C}
      11;                      !- Loop Design Temperature Difference {deltaC}

mjwitte:
I was expecting this to have fewer big diffs, not more.

rraustad:
Yeah, more review needed. I just wanted to get some proof of change for now.

src/EnergyPlus/SingleDuct.cc
rraustad: The change above may be the reason for more diffs than before. I tried to change it to be correct, not to minimize diffs. Maybe that would be a better starting point as this code migrates to a "fixed" state. First the CV TUs, then include the DamperHeatingAction change, then any other changes that are needed. At least then we could follow the diffs better. That would not be hard to do given the limited lines of code here.

rraustad:
OK, lets look at only the CV terminal units for now, agree on the results, merge this in, then open a new issue and address the other issues we saw with other TU types. Then we don't have to deal with so many diffs.

rraustad:
All eio diffs show the same thing. An increase in coil air flow used for sizing, water flow and capacity for the CV w/reheat TUs. The TU max air flow did not change. That's in the right direction regarding the original defect, but does show a large change in heating coil capacity. Next I want to look at cfm/sqft to see if this moved into a reasonable range, or now exceeds that range. The only validation I have thought to look for so far is pulling the water-side deltaT in line with the design criteria and as @mjwitte points out the change in unmet hours. This is not an incidental change to sizing so other thoughts on how to validate are welcome.

rraustad:
This is a system design exercise looking for typical sizing results based on example files that are not representative of real buildings. Not sure this exercise shows anything other than now the CV TU air flow is used for sizing the heating coil. 

The TU air flow is typically sized based on the cooling air flow so to ensure the rated heating coil outlet temp the higher air flow should be used since the CV TU could operate at that air flow. The historic sizing requirement was to simply meet the heating load indicated by the zone heating load and zone heating air flow with some other adjustments in some cases. So which method is correct/adequate/user desired, etc.

The typical heating air flow to capacity ratio is around 0.02 - 0.03 m3/W as shown [here](https://www.engineeringtoolbox.com/air-heating-systems-d_1136.html) in the log plot of Heat kW versus Air Flow m3/s. Develop shows those numbers while this branch is higher. Likely because the actual heating air flow is not used in this branch and instead, the cooling air flow is used. Develop was using the actual zone heating air flow rate for this TU type.

Typical heating loads are [50 heating btuh per square ft](https://rc.cfmdistributors.com/helpful-tips/hvac-rules-of-thumb/) or about 150 W/m2 and the example files don't always reflect that. The [1.7x rule](https://rc.cfmdistributors.com/helpful-tips/hvac-rules-of-thumb/) for heating versus cooling capacity seems to lean towards this branch.

ElectricEIRChillerHeatRecoveryAuto does show near 150 W/m2 and is heating dominated yet the cooling air flow is used because the design heating deltaT is 3x higher than cooling (i.e., 10 dT cooling and 30 dT heating). However, 50 C is not that uncommon given mechanical engineers and contractors desire to reduce call backs on low heating SAT that is uncomfortable to occupants but that is not representative of heating loads. The question is what is the air flow rate at the time heating coils are active. For a constant volume TU I would think the heating air flow would be at the CV TU design air flow but this issue has proven that is not always the case. What if a Unitary System used a heating air flow lower than the cooling air flow? The heating coil sizing method may be more system dependent than based solely on the TU type (out of scope but important).

    Sizing:Zone,
      14.,                     !- Zone Cooling Design Supply Air Temperature {C}
      50.,                     !- Zone Heating Design Supply Air Temperature {C}

    Zone Sizing Information, WEST ZONE,  Cooling, 5525.88638, 6354.76934, 0.46009, 0.52910
    Zone Sizing Information, WEST ZONE,  Heating, 6392.79409, 7990.99261, 0.17566, 0.21957
    Zone Sizing Information, EAST ZONE,  Cooling, 4912.75449, 5649.66766, 0.40904, 0.47040
    Zone Sizing Information, EAST ZONE,  Heating, 6531.79927, 8164.74909, 0.17951, 0.22438
    Zone Sizing Information, NORTH ZONE, Cooling, 6233.75677, 7168.82029, 0.51900, 0.59685
    Zone Sizing Information, NORTH ZONE, Heating, 9266.86654, 11583.58318, 0.25456, 0.31820

[8714-CorrectHeatingCoilSizing-CVTUs.xlsx](https://github.com/NREL/EnergyPlus/files/8045227/8714-CorrectHeatingCoilSizing-CVTUs.xlsx)

image

nrel-bot-2b:
@rraustad @myoldmopar it has been 28 days since this pull request was last updated.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -3428,15 +3428,30 @@ void SingleDuctAirTerminal::SizeSys(EnergyPlusData &state)
...
                         state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                         this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
-            } else {
+            } else if (this->SysType_Num == SysType::SingleDuctConstVolReheat || this->SysType_Num == SysType::SingleDuctConstVolNoReheat) {
+                TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
+                    max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
+                        this->MaxAirVolFlowRate * this->ZoneTurndownMinAirFrac);
+            } else { // SingleDuctVAVReheat, and other NoReheat terminal unit types (shouldn't matter for NoReheat TUs)
+                     // if (this->DamperHeatingAction == Action::Reverse) {
+                //    TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
```
mjwitte: This commented code can be deleted.

mjwitte (approved):
@rraustad The diffs here all make sense. 
The files with diffs all have AirTerminal:SingleDuct:ConstantVolume:Reheat with an autosized reheat coil.

Looking at 8713-5ZoneAirCooledConvCoef-Modified:
Space 5-1 design heating load is 2197W
The terminal air flow rate is 0.27917 m3/s
The reheat coil design inlet temp is 12.8C and the zone design heating temp is 21.1C
So, the reheat coil needs to raise the 12.8C flow to 21.1C just to provide zero heating, and then add the zone design heating load:
0.27917*(21.1-12.8)*1200 (density) + 2197W = 4977W
With this branch, the reheat coil design size is 4952W (correct) and with develop the size was only 2827W (too small).

As an additional test, I changed the Sizing:System design heating supply temp to be 21.1C (the zone heat setpoint). The reheat coil size is 2197W which is the zone design heating load, as expected, with both this branch and develop.

With some cleanup, this will be ready to drop in.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2835,3 +2835,183 @@ TEST_F(EnergyPlusFixture, setATMixerSizingProperties_Test)
...
+    // test air flow sizing of TU with hard sized input for Maximum Air Flow Rate {m3/s}
+    state->dataSize->ZoneSizingRunDone = true;
+    state->dataSize->CurZoneEqNum = 1;
+    state->dataSize->CurTermUnitSizingNum = 1;
+    state->dataSize->TermUnitFinalZoneSizing(1).NonAirSysDesHeatVolFlow = 0.5;
+    state->dataSingleDuct->sd_airterminal(SysNum).SizeSys(*state);
+    // constant volume TUs should have ZoneMinAirFracDes = 1 (now set correctly in GetSysInput)
+    EXPECT_EQ(0.0, state->dataSingleDuct->sd_airterminal(SysNum).ZoneMinAirFracDes);
+    // if ZoneMinAirFracDes = 0, TermUnitSizing(1).AirVolFlow would = 0.5 (NonAirSysDesHeatVolFlow)
+    // since ZoneMinAirFracDes = 1, TermUnitSizing(1).AirVolFlow = hard sized value (i.e., 1.0)
```
rraustad: I missed these comments on the last pass and these are no longer true. I still don't like that ZoneMinAirFracDes = 0 for CV TUs but at this point I guess "ZoneMinAirFracDes is not used for CV TUs" is where we are at. I am now thinking about these unit tests and whether they provide fool proof checks going forward.

src/EnergyPlus/SingleDuct.cc
```diff
@@ -3428,11 +3428,16 @@ void SingleDuctAirTerminal::SizeSys(EnergyPlusData &state)
...
+            // set air flow rate used to size heating coils, ZoneTurndownMinAirFrac defaults to 1 for those TU types that do not use it
             if (this->SysType_Num == SysType::SingleDuctVAVReheatVSFan) {
                 TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
                     max(state.dataSingleDuct->UserInputMaxHeatAirVolFlowRateSS,
                         state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                         this->MaxAirVolFlowRate * this->ZoneMinAirFracDes * this->ZoneTurndownMinAirFrac);
+            } else if (this->SysType_Num == SysType::SingleDuctConstVolReheat || this->SysType_Num == SysType::SingleDuctConstVolNoReheat) {
+                TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
+                    max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
+                        this->MaxAirVolFlowRate * this->ZoneTurndownMinAirFrac);
```
rraustad, line 3440: Lots of work went into 1 line of code but well worth it.

tst/EnergyPlus/unit/SingleDuct.unit.cc
```diff
@@ -2835,3 +2835,181 @@ TEST_F(EnergyPlusFixture, setATMixerSizingProperties_Test)
...
+    EXPECT_EQ(1.0, state->dataSize->TermUnitSizing(1).AirVolFlow); // doesn't matter for NoReheat TUs
+    EXPECT_EQ(1.0, state->dataSingleDuct->sd_airterminal(SysNum).MaxAirVolFlowRate);
+
+    state->dataSingleDuct->sd_airterminal(SysNum).MaxAirVolFlowRate = DataSizing::AutoSize;
+    state->dataSize->TermUnitFinalZoneSizing(1).DesHeatVolFlow = 0.25;
+    state->dataSingleDuct->sd_airterminal(SysNum).SizeSys(*state);
+    // TermUnitSizing(1).AirVolFlow should be greater of DesHeatVolFlow (0.25) and NonAirSysDesHeatVolFlow (0.5)
+    EXPECT_EQ(0.5, state->dataSize->TermUnitSizing(1).AirVolFlow);
+    // the TU will size to DesHeatVolFlow (0.25) even though NonAirSysDesHeatVolFlow = 0.5
+    EXPECT_EQ(0.25, state->dataSingleDuct->sd_airterminal(SysNum).MaxAirVolFlowRate);
```
rraustad, line 2914: These last 2 tests here are bothering me. TU air flow (sd_airterminal(SysNum).MaxAirVolFlowRate) sizes to greater of TUFZS cooling or heating (0.25). 

    MaxAirVolFlowRateDes = max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).DesCoolVolFlow,
                               state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).DesHeatVolFlow);

and TermUnitSizing.AirVolFlow sizes to greater of that and NonAirSysDesHeatVolFlow (0.5). Feels like an ill formed unit test or a logic problem in sizing. Darn these sizing changes. I hope that this is caught in setting TermUnitFinalZoneSizing air flow.

                TermUnitSizing(state.dataSize->CurTermUnitSizingNum).AirVolFlow =
                    max(state.dataSize->TermUnitFinalZoneSizing(state.dataSize->CurTermUnitSizingNum).NonAirSysDesHeatVolFlow,
                        this->MaxAirVolFlowRate * this->ZoneTurndownMinAirFrac);

rraustad: NonAirSysDesHeatVolFlow is the raw zone heating air flow set in EndZoneSizingCalc

    z.NonAirSysDesHeatVolFlow = c.DesHeatVolFlow;

And then later in that function is multiplied by the heating sizing factor.

    // update non air system design load and air flow to include the sizing factor
    state.dataSize->FinalZoneSizing(CtrlZoneNum).NonAirSysDesHeatLoad *= state.dataSize->FinalZoneSizing(CtrlZoneNum).HeatSizingFactor;
    state.dataSize->FinalZoneSizing(CtrlZoneNum).NonAirSysDesHeatVolFlow *= state.dataSize->FinalZoneSizing(CtrlZoneNum).HeatSizingFactor;

which could be < 1

    Sizing:Zone,
    N7, \field Zone Heating Sizing Factor
         \note if blank or zero, global heating sizing factor from Sizing:Parameters is used.
         \minimum 0

    Sizing:Parameters,
      N1, \field Heating Sizing Factor,
          \type real
          \minimum> 0.0
          \default 1.0

So it appears it is possible that FinalZoneSizing.NonAirSysDesHeatVolFlow could be less than FinalZoneSizing.DesHeatVolFlow, at this point in the code. DesHeatVolFlow is multiplied by HeatSizingFactor somewhere in this mix I just don't see it right now.

Then in UpdateTermUnitFinalZoneSizing the TermUnitSizing array is set. 

        auto &thisTUFZSizing(state.dataSize->TermUnitFinalZoneSizing(termUnitSizingIndex));
        auto &thisTUSizing(state.dataSize->TermUnitSizing(termUnitSizingIndex));
        // Copy everything from FinalZoneSizing to TermUnitFinalZoneSizing
        thisTUFZSizing = thisFZSizing;

To test this I used 8713-5ZoneAirCooledConvCoef-Modified

    Zone Sizing Information, SPACE5-1, Cooling, 2552.94399, 3318.82718, 0.21475, 0.27917, CHICAGO ANN CLG 1% CONDNS DB=>MWB, 7/21 15:00:00, 31.50000, 1.46163E-002, 182.49000, 20.00000, 0.18880, 0.00000
    Zone Sizing Information, SPACE5-1, Heating, 1690.05388, 2197.07005, 4.91278E-002, 0.18880, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 8.38927E-004, 182.49000, 20.00000, 0.18880, 0.00000
    Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Air Flow Rate [m3/s], 0.27917
    Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 4952.51303

and then changed sizing factors of 1.3 by exactly 1/2 to 0.65 expecting TU flow to reduce 0.27917 by 1/2

    Sizing:Parameters,
      0.65,                     !- Heating Sizing Factor
      0.65,                     !- Cooling Sizing Factor
      ;                        !- Timesteps in Averaging Window

I also have to cut the DesignSpecification:OutdoorAir for zone 5 in half (0.00944 per person to 0.00472) so that zone OA flow is not dominant and I do get what I expect. Air flow and heating coil capacity are 1/2 of what they were before.

    Zone Sizing Information, SPACE5-1, Cooling, 2552.94399, 1659.41359, 0.21475, 0.13959, CHICAGO ANN CLG 1% CONDNS DB=>MWB, 7/21 15:00:00, 31.50000, 1.46163E-002, 182.49000, 20.00000, 9.44000E-002, 0.00000
    Zone Sizing Information, SPACE5-1, Heating, 1690.05388, 1098.53502, 4.91278E-002, 9.44000E-002, CHICAGO ANN HTG 99% CONDNS DB, 1/21 24:00:00, -17.30000, 8.38927E-004, 182.49000, 20.00000, 9.44000E-002, 0.00000
    Component Sizing Information, AirTerminal:SingleDuct:ConstantVolume:Reheat, SPACE5-1 CV REHEAT, Design Size Maximum Air Flow Rate [m3/s], 0.13959
    Component Sizing Information, Coil:Heating:Water, SPACE5-1 ZONE COIL, Design Size Rated Capacity [W], 2476.25651

This PRs code change seems OK for now. The unit test is doing something that apparently can't happen in code.

PR #8715, opened 2021-04-14
Update the function of GetInternalGainDeviceIndex

### LINKED ISSUE #8705: A room air model crashed due to different names between internal gain objects and room air objects for internal gains
Issue overview
--------------
When the names defined by internal gain objects (such as people, light, other equipment) and those referred by the object "RoomAir:Node:AirflowNetwork:InternalGains" are different in an EnergyPlus model, the model crashes without an error message.  The model breaks when it sums all the convective internal heat gains in RoomAirModelAirflowNetwork.cc 
SumInternalConvectionGainsByIndices. The function of GetInternalGainDeviceIndex for checking the consistency between the object names of internal gains needs to return ErrorsFound when this happens.

### Details
 - v9.5
 - ticket #15660

### Checklist

LipingWang:
Pull request overview
---------------------
 - Fixes #8705 
 - Update the function of GetInternalGainDeviceIndex which checks the consistency between the names of internal gain objects (such as people, light, other equipment) and those referred by the object "RoomAir:Node:AirflowNetwork:InternalGains."

### Pull Request Author

### Reviewer


src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -8077,6 +8077,8 @@ namespace InternalHeatGains {
                 Found = true;
                 DeviceIndex = DeviceNum;
                 break;
+            } else {
+                ErrorFound = true;
```
Myoldmopar: I guess this change works technically, but it seems out of place.  With this change, you basically have this:

```
Loop over all devices
  If device.name matches searchName
    If device.type matches searchType
      ErrorsFound = false, break, and return
    But if device.type doesn't match searchType
      ErrorsFound = true, break, and return
    EndIf
  But if device.name doesn't match searchName    << new line
    ErrorsFound = true, and continue the loop    << new line
  EndIf
EndLoop
```

So it's weird because you are going to loop over all the devices and you are going to end up setting ErrorsFound to true a good number of times before you find a match.  Why go through the trouble of setting ErrorsFound so many times?  It should've just been initialized to true before the loop starts and that's all you needed to do.  Then it would remain false if the loop terminates without finding a match.  And in fact, you could just simplify the logic and remove the line that sets ErrorsFound to false because it would already be initialized to false.  It should be reduced to something like:

```
ErrorsFound = true  << assume we don't find a match
Loop over all devices
  If device.name matches searchName
    If device.type matches searchType
      ErrorsFound = false
    EndIf
    break and return
  EndIf
EndLoop
```

Seems like a simpler solution, and probably worth making the change before this merges.  Or at least let me know why you think we shouldn't do it.

mjwitte: @Myoldmopar When I first saw the comment about `ErrorFound` I was concerned that this may be a case where the argument is passed along to many functions and might already be true when this function is called (as so many other places in the code). But in this case, the argument is [localized when the function is called](https://github.com/NREL/EnergyPlus/blob/6b0e48b11868effc547dade7162791519d232536/src/EnergyPlus/RoomAirModelManager.cc#L1713-L1728). Seems it would be better for this function to be a `bool` and remove `ErrorFound` as an argument, just to keep it clear for future devs, but maybe that's nit-picking. Or maybe it should be an `int` function and keep the error arg.

Myoldmopar: I think it would be reasonable for the function to just return an integer and eliminate the error found bool altogether.  Return -1 as an invalid device index.  That would eliminate the need to return **both** an error flag and a valid index flag.  And that is something we do quite often in E+ already.  (Not that that alone would be precedent...)

One additional note is that if you make a change to initialize the variable to true you may need to add it to the approve list in `scripts/dev/find_byref_bool_override.py`.

LipingWang: @mjwitte @Myoldmopar Thank you for the comments and notes! I will update this as an int function. Should I consider moving the related error message from RoomAirModelManager.cc inside this function?

mjwitte: > Should I consider moving the related error message from RoomAirModelManager.cc inside this function?

I would leave it where it's at, since it uses some local information to build the error message.

LipingWang: @mjwitte @Myoldmopar  The function `GetInternalGainDeviceIndex` was updated. Please let me know for any further suggestions or comments. Thanks a lot!

src/EnergyPlus/InternalHeatGains.cc
```diff
@@ -8054,31 +8053,24 @@ namespace InternalHeatGains {
...
+            if (UtilityRoutines::SameString(state.dataHeatBal->ZoneIntGain(ZoneNum).Device(DeviceNum).CompObjectName, IntGainName) &&
+                state.dataHeatBal->ZoneIntGain(ZoneNum).Device(DeviceNum).CompTypeOfNum == IntGainTypeOfNum) {
                 DeviceIndex = DeviceNum;
                 break;
+            } else {
+                DeviceIndex = -1;
             }
         }
+        return DeviceIndex;
     }
```
amirroth, line 8074: If you are returning `DeviceIndex`, why does it also need to be a reference parameter?

LipingWang: @amirroth  Thank you for your comments! The returned value for `DeviceIndex` was to replace `ErrorFound` in the original function `GetInternalGainDeviceIndex`. In the same time, `DeviceIndex` in this function (`GetInternalGainDeviceIndex`) needs to be passed back to `state.dataRoomAirMod->RoomAirflowNetworkZoneInfo(ZoneNum).Node(RAFNNodeNum).IntGainsDeviceIndices(gainsLoop)` within the function `void GetRoomAirflowNetworkData` in `RoomAirModelManager`. 
What would be a better way of handling this? Any recommendations? Thanks a lot!

amirroth: Well, if a returned index of -1 implies error then, you can use that to infer the `IntGainError` flag at the call site. If there is a single error mode that can be specified using a return value (the two most common ones are -1 for IndexNotFound and NULL for ObjectNotFound or MemoryReferenceNotFound) then this is a a good pattern, especially if the error must be handled in the direct caller. 

I've noticed EnergyPlus often passes Error parameters through multiple levels of functions. This pattern is less good and has been subsumed by exceptions. I assume that at one point we will do a pass that cleans that up that pattern as well, but probably not in the very near future.

LipingWang:
@mjwitte @amirroth  @Myoldmopar  Many thanks for helping me improve the code! The function `GetInternalGainDeviceIndex` was updated by removing `&DeviceIndex` as one of the function argument, returning an integer `DeviceIndex` from the function, and replacing `std::string` with `std::string_view` passed by reference.  Thank you so much!

mjwitte:
@LipingWang Please update this branch for @Myoldmopar 's final review.

Myoldmopar:
Tested this locally with develop pulled in, all good.  Thanks!

PR #8723, opened 2021-04-20
VariableRefrigerantFlow Terminal Unit on air loop does not test node connections

### LINKED ISSUE #8718: VRF TU on AirLoopHVAC supply branch will force loop off if terminal OA mixer not set and OA inputs not autosized
Issue overview
--------------
@rraustad I'm not sure this is intended or just a misinterpretation of the input requirement. 

A ZoneHVAC:TerminalUnitVariableRefrigerantFlow on the supply branch of an AirLoopHVAC will force the loop off if the terminal unit OA mixer is left blank **and** the terminal unit 'Cooling/Heating/No Load Outdoor Air Flow Rate' inputs are not 'autosized'. 

The documentation for VRF Terminal Unit Field: Outdoor Air Mixer Object Type states:
"If this field is left blank, an outdoor air mixer is not simulated. If the VRF terminal unit is used as air loop equipment and this terminal unit’s outdoor air flow rates are autosized then the autosized flow rates will be set to 0 when an outdoor air system is used in the air loop"

From this, I expected that not specifying an OutdoorAir:Mixer at the terminal unit will cause any input in the 'outdoor air flow rate' inputs to be ignored, since "an outdoor air mixer is not simulated". Instead, if those inputs are, say, hard-sized to 0, the air loop operation will be overridden and (in the example files below) the air loop will never turn on:

image

[VariableRefrigerantFlow_5Zone_wAirloop.zip](https://github.com/NREL/EnergyPlus/files/6320039/VariableRefrigerantFlow_5Zone_wAirloop.zip)

### Details
 - Windows 10
 - 9.5

### Checklist

rraustad:
Pull request overview
---------------------
 - Fixes #8718
 - When the ZoneHVAC:TerminalUnit:VariableRefrigerantFlow object is place in AirloopHVAC the node connections are not tested for incorrect node names. This can lead to incorrect operation or no operation at all.

### Pull Request Author

### Reviewer


rraustad:
New warning for defect file.

    ** Severe  ** ZoneHVAC:TerminalUnit:VariableRefrigerantFlow "TU5INAIRLOOP"
    **   ~~~   ** ... For draw thru fan when no OA mixer is specified and a cooling coil is present the terminal unit inlet node name must match the cooling coil inlet node name.
    **   ~~~   ** ... Terminal unit inlet node name = AIRLOOP MIXED AIR OUTLET.
    **   ~~~   ** ... Cooling coil inlet node name = TU5 VRF DX CCOIL INLET NODE.
    **  Fatal  ** GetVRFInput: Errors found in getting AirConditioner:VariableRefrigerantFlow system input. Preceding condition(s) causes termination.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -422,7 +424,7 @@ class AirLoopFixture : public EnergyPlusFixture
         int VRFTUOutletNodeNum = 31;
         int VRFTUOAMixerOANodeNum = 32;
         int VRFTUOAMixerRelNodeNum = 33;
-        int VRFTUOAMixerRetNodeNum = 34;
+        int VRFTUOAMixerRetNodeNum = VRFTUInletNodeNum;
```
rraustad, line 427: Existing unit test was not configured correctly. When testing new function the failure was due to the OA mixer return node not connected to the TU inlet node. Correcting this changed the setup needed for the unit test.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -545,7 +548,7 @@ TEST_F(AirLoopFixture, VRF_SysModel_inAirloop)
     state->dataScheduleMgr->Schedule(state->dataHVACVarRefFlow->VRF(curSysNum).SchedPtr).CurrentValue = 1.0; // enable the VRF condenser
     state->dataScheduleMgr->Schedule(thisTU.SchedPtr).CurrentValue = 1.0;                                    // enable the terminal unit
     state->dataScheduleMgr->Schedule(thisTU.FanAvailSchedPtr).CurrentValue = 1.0;                            // turn on fan
-    state->dataScheduleMgr->Schedule(thisTU.FanOpModeSchedPtr).CurrentValue = 0.0;                           // set cycling fan operating mode
+    state->dataScheduleMgr->Schedule(thisTU.FanOpModeSchedPtr).CurrentValue = 1.0;                           // set constant fan operating mode
```
rraustad, line 551: Changed to constant fan for set point control.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -583,7 +587,8 @@ TEST_F(AirLoopFixture, VRF_SysModel_inAirloop)
     state->dataLoopNodes->Node(thisTU.VRFTUOutletNodeNum).TempSetPoint = 20.0; // select 20 C as TU outlet set point temperature
 
     InitVRF(*state, curTUNum, curZoneNum, FirstHVACIteration, OnOffAirFlowRatio, QZnReq); // Initialize all VRFTU related parameters
-
+    thisTU.NoCoolHeatOutAirMassFlow = 0.0;
+    // the above needs correcting for air loop equipment. OA flow should be set to 0 if autosized as it is here.
```
rraustad: This will require a follow up issue.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -646,15 +656,56 @@ TEST_F(AirLoopFixture, VRF_SysModel_inAirloop)
...
+    ErrorsFound = false;
+    thisTU.coolCoilAirOutNode += 1; // revert previous change
+    thisTU.heatCoilAirInNode -= 1;  // change index of comp node (watch for array bounds)
+    CheckVRFTUNodeConnections(*state, curTUNum, ErrorsFound);
+    EXPECT_TRUE(ErrorsFound); // nodes are not connected correctly
+    ErrorsFound = false;
+    thisTU.heatCoilAirInNode += 1;  // revert previous change
+    thisTU.heatCoilAirOutNode -= 1; // change index of comp node (watch for array bounds)
+    CheckVRFTUNodeConnections(*state, curTUNum, ErrorsFound);
+    EXPECT_TRUE(ErrorsFound); // nodes are not connected correctly
```
rraustad, line 708: New tests to exercise node connection function. This actually found a few missed cases in function.

tst/EnergyPlus/unit/HVACVariableRefrigerantFlow.unit.cc
```diff
@@ -4186,6 +4237,35 @@ TEST_F(EnergyPlusFixture, VRFTest_SysCurve)
...
+    ErrorsFound = false;
+    thisTU.coolCoilAirOutNode += 1; // revert previous change
+    thisTU.heatCoilAirInNode -= 1;  // change index of comp node (watch for array bounds)
+    CheckVRFTUNodeConnections(*state, VRFTUNum, ErrorsFound);
+    EXPECT_TRUE(ErrorsFound); // nodes are not connected correctly
+    ErrorsFound = false;
+    thisTU.heatCoilAirInNode += 1;  // revert previous change
+    thisTU.heatCoilAirOutNode -= 1; // change index of comp node (watch for array bounds)
+    CheckVRFTUNodeConnections(*state, VRFTUNum, ErrorsFound);
+    EXPECT_TRUE(ErrorsFound); // nodes are not connected correctly
```
rraustad, line 4268: Same functions tests for a TU as zone equipment.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -5566,6 +5568,247 @@ void GetVRFInputData(EnergyPlusData &state, bool &ErrorsFound)
     }
 }
 
+void CheckVRFTUNodeConnections(EnergyPlusData &state, int const VRFTUNum, bool &ErrorsFound)
```
mjwitte, line 5577: Would it make sense to make this a member function and remove `VRFTUNum` as an argument?
And make this a `bool` function and remove `ErrorsFound` as an argument?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -5566,6 +5568,247 @@ void GetVRFInputData(EnergyPlusData &state, bool &ErrorsFound)
...
+{
+
+    std::string const cTerminalUnitType("ZoneHVAC:TerminalUnit:VariableRefrigerantFlow");
```
mjwitte: Checking with @Myoldmopar about our current best practice for declaring string constants. This is nit-picking for a one-time function, but I'd like to make sure new code doesn't perpetuate old patterns.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
```diff
@@ -5566,6 +5568,247 @@ void GetVRFInputData(EnergyPlusData &state, bool &ErrorsFound)
...
+                    if (VRFTUInletNodeNum > 0 && heatCoilAirInNode > 0) {
+                        ShowContinueError(state, "... Terminal unit inlet node name = " + nodeID(VRFTUInletNodeNum) + ".");
+                        ShowContinueError(state, "... Heating coil inlet node name = " + nodeID(heatCoilAirInNode) + ".");
+                    }
+                    ErrorsFound = true;
+                }
+            }
+        }
+    }
+    if (FanPlace == DataHVACGlobals::BlowThru && !OAMixerUsed) {
```
mjwitte, line 5686: I'm not seeing a check when `FanPlace == DataHVACGlobals::BlowThru && OAMixerUsed`

mjwitte: The logic here works, but it seems somewhat complicated, so many nested ifs, some with dual comparisons. Wouldn't it be easier to just start with the inlet node as the `currentNode` and walk through the components one by one, skipping those that don't exist and check that the next component's inlet matches `currentNode`, then update `currentNode` to the current component's outlet and mode to the next component, and so on.

mjwitte (commented):
@rraustad I've tested this with the defect file (and with a corrected defect file) and everything works as it should here now. Unit tests all run fine. A few comments below.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
Myoldmopar: @lefticus could you document in a GitHub comment here the best practice for local variable strings?  We have discussed them on calls and maybe chats, but it would be good to lock it down here.  IIRC it's going to be a constexpr string_view, but I can't recall all the implications.

Myoldmopar: @mjwitte I found where I had seen @lefticus document the best practice.  The message is:

> For any future update of const static std::string, I recommend a move to constexpr static std::string_view and updating all downstream code to avoid as many dynamic allocations as possible.

For a fuller discussion, check the main comment body of #8663.  I now have a better feel for how to utilize string views; this is a really exciting change with lots of potential in our code.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: This function is inside GetVRFInputData which is not a member function. I think when models gather all instances of the object in a for loop, instead of just getting the current object data, then non-member functions will be used.

Re a bool function:

ErrorsFound is set throughout GetVRFInputData and then at the end this new function is called to test the node connections (within the for loop for all TUs). If this were a bool function then the call would look like this:

ErrorsFound = ErrorsFound || CheckVRFTUNodeConnections(EnergyPlusData &state, int const VRFTUNum);

I guess either way would work. Not sure I like the implications of this syntax where developers may forget to include "ErrorsFound || boolFunctionCall" to retain the current value of ErrorsFound.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: So constexpr static std::string_view cTerminalUnitType("string"); should be used here?

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
mjwitte: That's fine as-is.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: I'll check on a missing test for blow through. I did try this several ways and found the nested ifs were significant regardless of the organization. I think this was my 3rd attempt. Is there a better way? Possibly, I just didn't come across one.

BlowThru or DrawThru, FanExists, CoolingCoilExists, HeatingCoilExists, SuppHeatingCoilExists. These cause nested IFs regardless of what I do.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
rraustad: @lefticus, use constexpr static std::string_view cTerminalUnitType("string"); here?

lefticus: @Myoldmopar @rraustad I wrote up a full explanation of what is best and what the implications may or may not be on this other PR: 

https://github.com/NREL/EnergyPlus/pull/8663

rraustad:
Defect warning with revised warning format.

    ** Severe  ** ZoneHVAC:TerminalUnit:VariableRefrigerantFlow="TU5INAIRLOOP",
    **   ~~~   ** ... For draw thru or no fan when no OA mixer is specified and a cooling coil is present the terminal unit inlet node name must match the cooling coil inlet node name.
    **   ~~~   ** ... Terminal unit inlet node name = AIRLOOP MIXED AIR OUTLET.
    **   ~~~   ** ... Cooling coil inlet node name = TU5 VRF DX CCOIL INLET NODE.
    **  Fatal  ** GetVRFInput: Errors found in getting AirConditioner:VariableRefrigerantFlow system input. Preceding condition(s) causes termination.

src/EnergyPlus/HVACVariableRefrigerantFlow.cc
mjwitte: It's fine like it is.

rraustad:
@Myoldmopar @mjwitte this should be ready now.

mjwitte (approved):
@rraustad Looks good. I'll let CI make more green and the merge this.
And thanks for cleaning up the stray unit test formatting.

PR #8724, opened 2021-04-20
Add support for Coil:Cooling:DX in ComponentCost:LineItem

### LINKED ISSUE #8736: ComponentCost:LineItem Uses Wrong SpeedNum for Coil:DX and Coil:Cooling:DX:SingleSpeed
Issue overview
--------------
ComponentCost:LineItem is using the wrong speed number for some of the current DX coils. This should be using the highest speed, but it's always using the lowest.

https://github.com/NREL/EnergyPlus/blob/2381b2417047abcaf5cbd43415202fb15cefd4d1/src/EnergyPlus/CostEstimateManager.cc#L645

### Details
 - Version of EnergyPlus (2381b2417047abcaf5cbd43415202fb15cefd4d1)

mitchute:
Pull request overview
---------------------
 - Adds support for Coil:Cooling:DX in ComponentCost:LineItem
 - PR Fixes #8736 

### Pull Request Author

### Reviewer


mitchute:
@Myoldmopar @mjwitte @rraustad 

Ready for review.

src/EnergyPlus/CostEstimateManager.cc
```diff
@@ -619,7 +631,6 @@ namespace CostEstimateManager {
                     SurfMultipleARR.deallocate();
 
                 } else if ((SELECT_CASE_var == "COIL:DX") || (SELECT_CASE_var == "COIL:COOLING:DX:SINGLESPEED")) {
-                    WildcardObjNames = false;
```
rraustad: The reset was deleted

src/EnergyPlus/CostEstimateManager.cc
```diff
@@ -683,6 +694,85 @@ namespace CostEstimateManager {
...
                     }
 
+                } else if (SELECT_CASE_var == "COIL:COOLING:DX") {
+
+                    auto &parentObjName = state.dataCostEstimateManager->CostLineItem(Item).ParentObjName;
+                    bool coilFound = false;
+                    thisCoil = 0;
+                    //  check for wildcard * in object name..
+                    if (parentObjName == "*") { // wildcard, apply to all such components
+                        WildcardObjNames = true;
```
rraustad, line 713: ... following WildcardObjNames reset from above, this bool will be false getting here, right? Meaning if a previous coil type had a wild card, this variable will get reset before looking for the new coil model.

src/EnergyPlus/CostEstimateManager.cc
```diff
@@ -683,6 +694,85 @@ namespace CostEstimateManager {
...
+                    } else if (!state.dataCostEstimateManager->CostLineItem(Item).ParentObjName.empty()) {
+                        auto &v = state.dataCoilCooingDX->coilCoolingDXs;
+                        auto isInCoils = [&parentObjName](const CoilCoolingDX &coil) { return coil.name == parentObjName; };
```
rraustad, line 720: isInCoils sounds like a bool but is a coil name string. Just a little confusing.

src/EnergyPlus/CostEstimateManager.cc
```diff
@@ -683,6 +694,85 @@ namespace CostEstimateManager {
...
+                            state.dataCostEstimateManager->CostLineItem(Item).Qty * state.dataCostEstimateManager->CostLineItem(Item).ValuePer;
+                        state.dataCostEstimateManager->CostLineItem(Item).Units = "Ea.";
+                    }
+
+                    if (state.dataCostEstimateManager->CostLineItem(Item).PerKWCapPerCOP > 0.0) {
+                        if (WildcardObjNames) {
+                            Real64 Qty(0.0);
+                            for (auto const &e : state.dataCoilCooingDX->coilCoolingDXs) {
+                                // TODO: Using 0 index to match above Coil:DX object
+                                Real64 COP = e.performance.normalMode.speeds[0].original_input_specs.gross_rated_cooling_COP;
```
rraustad: Is 0 the highest speed? I think I would use numSpeeds here. Not sure what to do for VS coils. For a cost estimate it seems reasonable to use the rated COP and TotCap at highest speed.

src/EnergyPlus/CostEstimateManager.cc
mitchute: Yes, the reset above should ensure it's always false before it gets here.

src/EnergyPlus/CostEstimateManager.cc
mitchute: ```isInCoils``` is a lambda that returns the ```coil.name == parentObjName``` bool.

src/EnergyPlus/CostEstimateManager.cc
mitchute: Coil:DX in the previous block always uses the first speed. i.e ```DXCoil(thisCoil).RatedTotCap(1)```.

That doesn't mean that we should continue with that precedent. I'll defer to your recommendations.

src/EnergyPlus/CostEstimateManager.cc
mitchute: Good catch. I'll add that back.

src/EnergyPlus/CostEstimateManager.cc
mjwitte: That seems wrong. It should be full capacity. But for this coil, that comes from the performance object, I think.

amirroth:
Here's a meta-question. How many people use this cost feature? This doesn't seem to me like something that belongs in EnergyPlus.

bonnema:
> Here's a meta-question. How many people use this cost feature? This doesn't seem to me like something that belongs in EnergyPlus.

How many I am not sure. I know that in the past it has been used it as a part of optimization routines to explore the cost benefits of different efficiency levels for a given scenario.

amirroth:
Yeah. That was before OpenStudio Measures were invented. My sense is that this is vestigial. As are a number of other feature sets that we continue to carry forward for sentimental reasons. I think it's well past time to start tossing some of this dead weight overboard. Why waste time compiling it, updating it, having this conversation, etc.?

mitchute:
@amirroth you're probably right, but my guess is that there are still plenty of non-OS workflows out there and it's impossible to know how often objects like this are used. That would be useful data if we could ever figure out how to get it.

amirroth:
Here's an idea, take the feature away and see if anyone complains. We're not going to have this conversation now. I'm going to put an agenda item on the next EnergyPlus technicalities meeting.

Myoldmopar:
@amirroth I think we are all in support of removing extra pieces that don't need to be inside the physics engine.  Let's definitely have that discussion on the technicalities call in 2 weeks.  I'll add it to the agenda right now.

At this moment, we are trying to put in effort to get further down the road of eliminating all the dx coil objects.  We are looking at all the objects that reference the DX cooling coils (there are like 18!), and either:
- Adding in support for the new general coil to the parent object (as few as possible)
- Transitioning those parent objects to use the unitary system code behind the scenes, where it will essentially adopt support for all the coils in one step (probably most things will fall here)
- Actually eliminating parent objects and transitioning the IDF objects into the unitary system.  (We love this approach, except that in many cases, the huge set of unitary inputs is pretty daunting for anyone using one of the more specific parent objects, so we're balancing that.)

In this particular work that @mitchute completed, we have got the component cost line item object which tracks the equipment cost for coils scaled with efficiency and capacity.  All he's done is add support for the new generic coil object there which is a key step in eliminating all the other coils soon.  I think it would be best to continue the discussion of the cost object ASAP and maybe even set a formal plan for eliminating some objects we think we shouldn't need.  But for now I'd like to go ahead and get this reviewed and merged.  Worst case is if we decide to abandon the cost stuff, his work is deleted back off, but I believe the effort was on the order of a couple hours, so not a big problem there.

Let me know if you want me to hold up this PR until we discuss it further, and thanks!

amirroth:
No, please do not hold up the PR for this.

src/EnergyPlus/CostEstimateManager.cc
Myoldmopar: Do we need to address this change then?  Mine out the full capacity based on nominal speed or whatever it is called?  Otherwise this coding looks good.  It would be great to go ahead and add a unit test or at least an example file with this in place.

mitchute: Following up @mjwitte @rraustad @Myoldmopar 

Do we fix the old coil to use the highest speed as part of this?

mjwitte: Yes. If you haven't yet, please post an issue to note that it wasn't using the full capacity before.

rraustad: I agree. This needs to be fixed for any existing VS coil models.

mitchute:
@Myoldmopar @rraustad @mjwitte ready for review (again).

The DXCoil speed bug from #8736 should be handled in https://github.com/NREL/EnergyPlus/pull/8724/commits/f9effafb88c03563f9c9109cb070e5ad28c043f2

Myoldmopar:
IDD change makes sense, code changes look good, added to an existing IDF rather than adding a new test file :+1: , unit test in place and passing, CI is happy, and I pulled develop in one more time locally and everything passes there.  This appears good to go.  Thanks @mitchute !  Merging!

PR #8728, opened 2021-04-26
CTF Caching

mitchute:
Pull request overview
---------------------
 - Adds previous NFP regarding CTF caching
 - Prototypes a caching method for CTFs
 
At a minimum, the PR still needs the following:
 - Investigation into the sensitivity of matching materials properties precision on results
 - Full test suite re-runs to verify the caching is working as expected for all files
 - Results from timing analysis
 - Unit tests
 - Move GHE cached results over to this method (maybe)

### Pull Request Author

### Reviewer


Myoldmopar:
Interesting, I wonder if that unit test would fail if it ran a second time.  I'm going to rerun that test to see...

Myoldmopar:
It kinda went the wrong way on that one :shrug:.  Windows seemed to be having some trouble deleting old builds or something so I gave it a reboot just now and I deleted these results one more time.  I bet it'll be OK now.

Myoldmopar:
@amirroth 

> Why have the key as a `std::string` and not as an `unsigned long long`?

In JSON, the key **has** to be [a string](https://www.w3schools.com/js/js_json_objects.asp), [afaik](https://www.json.org/json-en.html):

```
Keys must be strings, and values must be a valid JSON data type (string, number, object, array, boolean or null).
```


The JSON library treats it accordingly and uses string keys, even if we'd like to use integers.  I don't _think_ there is a way around this, but I'm definitely open to suggestion.  In all seriousness, I thought JSON allowed numeric keys, but I guess every time I've seen numeric keys, they were quoted and I just didn't comprehend it.

Let me know if I misunderstood your question though.

amirroth:
@Myoldmopar @mitchute Right, I keep forgetting you're storing this as JSON/CBOR rather than rolling your own binary cache format. Carry on, my wayward sons.

mitchute:
Here's an update on the timing results of this work.

From the current code at 36b915f85fed89402e84c07290f788e710f7ad10, I ran roughly 600 of our example files as design-day simulations letting EnergyPlus both compute the CTFs and then look up and use the cached CTF values. Both the calculated CTF runs and the cached runs were each run three times for each file. The time to compute or look up each CTF value was measured using C++ internal timing features and then output to stdout; the total simulation time was determined from EnergyPlus's self-reported simulation time in the stdout.

From this data for each construction, I was able to average the three times for either calculating the CTFs or looking them up from the cache. Then, the sum of these averaged values for each example file is what I'm calling "Cumulative CTF Lookup (or Calc) Time". The simulation time I'm using is just the average stdout time from each of the three runs.

The first plot shows the cumulative time to compute the CTFs for each of the files. The second plot shows the time to find and load the data from the cache.

image

image

All of the files show performance gains in the time to get the CTFs, with most of the files going from 1-5 ms to under 0.1 ms. The highest performers were the radiant system files (RadLoTempHydrCtrlOpt3.idf, RadLoTempHydrHeatCool2D.idf) going from about 1800 ms to about 0.05 ms. 

Most of the files similarly showed performance gains in overall simulation time, but since I was only running design days it's hard to say how much significance we should give to those numbers. For the radiant systems files, the overall simulation time was meaningful going from about 2.5 s to about 0.8 s. 

There were also a few files that regressed in overall simulation time, although I'm not sure of the cause of that at the moment. It does appear that most of these files all ran in under 1 s when computing the CTFs. I'm not sure if loading the cache and looking up the value is overwhelming the calculation because these files have very simple and/or only a few constructions, or what other issue could be causing this.

src/EnergyPlus/Cache.hh
```diff
@@ -90,6 +90,27 @@ namespace Cache {
 
     void writeCache(EnergyPlusData &state);
 
+#define USE_REINTERPRET_CAST
```
mitchute: @Myoldmopar feel free to review this commit if you have a minute since there have been a few options floated around. It's been noted [here](https://github.com/NREL/EnergyPlus/blob/develop/third_party/ObjexxFCL/src/ObjexxFCL/bit.hh#L258) that the reinterpret_cast usage is UB, but it's obviously been working OK since we've been using it for a long while now. @amirroth noted offline that the UB is due to mismatched sizes of the cast-to and cast-from types. In this case ```sizeof(double) == sizeof(unsigned long long)``` so things should be OK from that perspective.

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,128 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+
+    // data from material layers
+    // adding the layer num to get a unique key for objects with reversed layers
+    for (int Layer = 1; Layer <= this->TotLayers; ++Layer) {
+        int CurrentLayer = this->LayerPoint(Layer);
+        auto &mat = state.dataMaterial->Material(CurrentLayer);
+
+#ifdef USE_REINTERPRET_CAST
+        key ^= Cache::prepFloatForCacheKey(mat.Thickness, 15) + Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Conductivity, 15) + Layer;
```
amirroth: Doesn't matter so much, but it makes more sense to mulitply or shift left by `Layer` than to add `Layer`.

mitchute: Will this address the layer swapping issues you had noted previously?

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,128 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+        key ^= Cache::prepFloatForCacheKey(mat.Resistance, 15) + Layer;
+#endif
+#ifdef USE_STATIC_CAST_AND_FP_MULT
+        key ^= Cache::prepFloatForCacheKey(mat.Thickness, 1E5) + Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Conductivity, 1E5) + Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Density, 1E5) + Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.SpecHeat, 1E5) + Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Resistance, 1E5) + Layer;
+#endif
+#ifdef USE_C_CAST_AND_FP_MULT
```
amirroth: This block looks identical to previous one. Can combine using `#if defined(USE_STATIC_CAST_AND_FP_MULT) || defined(USE_C_CAST_AND_FP_MULT)`. Also, I am not entirely sure that there is any difference between C cast and `static_cast` for scalar data types. In this specific case, I believe those two are the same option, so this may not be necessary anyway.

src/EnergyPlus/Construction.cc
mitchute: Sorry, that was supposed to be an actual question, not a poorly punctuated statement. I missed the "this" which made it confusing.

src/EnergyPlus/Construction.cc
mitchute: Thanks for pointing that out. I think these ```#ifdef```s are going away before this PR merges. I just wanted to give anyone interested an opportunity to comment on the different approaches.

src/EnergyPlus/Construction.cc
amirroth: Yes, it will disambiguate constructions with swapped layers. It will actually do more than that. It will disambiguate constructions with multiple instances of the same layer from each other, even if those layers are different. If you have a construction ABA and another CBC, without multiplication or shifting they will look the same because A^A and C^C are both zero.

mitchute: OK, thanks for pointing that out. I'll get that changed.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -114,6 +114,8 @@ set(SRC
     BranchNodeConnections.hh
     CTElectricGenerator.cc
     CTElectricGenerator.hh
+    Cache.cc
+    Cache.hh
```
mitchute, line 118: New files

src/EnergyPlus/Cache.cc
```diff
@@ -0,0 +1,114 @@
...
+#include <EnergyPlus/Cache.hh>
+#include <EnergyPlus/Data/EnergyPlusData.hh>
+#include <EnergyPlus/DataGlobals.hh>
+#include <EnergyPlus/DataStringGlobals.hh>
+#include <EnergyPlus/FileSystem.hh>
+#include <EnergyPlus/UtilityRoutines.hh>
+
+namespace EnergyPlus::Cache {
+
+void readJSONfile(EnergyPlusData &state, std::string &filePath, nlohmann::json &j)
```
mitchute: New methods for reading/writing JSON files, and loading/writing the cache.

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+namespace Cache {
+
+    std::string const cacheName = "eplusout.cache";
+    std::string const CTFKey = "CTFs";
+
+    void readJSONfile(EnergyPlusData &state, std::string &filePath, nlohmann::json &j);
+
+    void writeJSONfile(nlohmann::json &j, std::string &fPath);
+
+    template <typename T> void jsonToArray(EnergyPlusData &state, Array1D<T> &arr, nlohmann::json &j, std::string const &key)
```
mitchute: New function for reading a JSON list into a 0-based ```Array1D<T>```. Type ```T``` should be limited to a string, number, or bool here, so feel free to comment if we need to add some sort of check for this.

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+            for (auto &v : j.at(key)) {
+                arr(idx) = v;
+                ++idx;
+            }
+        } catch (const nlohmann::json::out_of_range &e) {
+            ShowFatalError(state, format("From eplusout.cache, key: \"{}\" not found", key));
+        }
+    }
+
+    template <typename T> void jsonToArray1(EnergyPlusData &state, Array1D<T> &arr, nlohmann::json &j, std::string const &key)
```
mitchute: Same, but for a 1-based ```Array1D<T>```

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+            for (auto &v : j.at(key)) {
+                arr(idx) = v;
+                ++idx;
+            }
+        } catch (const nlohmann::json::out_of_range &e) {
+            ShowFatalError(state, format("From eplusout.cache, key: \"{}\" not found", key));
+        }
+    }
+
+    template <typename T> void jsonToData(EnergyPlusData &state, T &data, nlohmann::json &j, std::string const &key)
```
mitchute: New function for getting a JSON value into a variable.

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+    {
+        try {
+            data = j.at(key);
+        } catch (const nlohmann::json::out_of_range &e) {
+            ShowFatalError(state, format("From eplusout.cache, key: \"{}\" not found", key));
+        }
+    }
+
+    template <typename T> void arrayToJSON(Array1D<T> const &arr, nlohmann::json &j, std::string const &key)
```
mitchute: A new method for taking an ```Array1D``` to a JSON list.

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+        for (auto v : arr)
+            vect.push_back(v);
+        j[key] = vect;
+    }
+
+    void loadCache(EnergyPlusData &state);
+
+    void writeCache(EnergyPlusData &state);
+
+    inline unsigned long long prepFloatForCacheKey(double const x, unsigned int const precision_bits)
```
mitchute: A new method for taking a floating point number to an unsigned long long with a given number of precision bits.

src/EnergyPlus/Cache.hh
```diff
@@ -0,0 +1,154 @@
...
+    {
+        assert(sizeof(double) == sizeof(unsigned long long));
+        return *reinterpret_cast<unsigned long long const *>(&x) >> (64 - 12 - precision_bits);
+    }
+
+} // namespace Cache
+
+struct CacheData : BaseGlobalStruct
```
mitchute, line 93: A new struct for the state cache data.

src/EnergyPlus/Construction.cc
```diff
@@ -1064,6 +1060,9 @@ void ConstructionProps::calculateTransferFunction(EnergyPlusData &state, bool &E
     if (allocated(this->Gamma1)) this->Gamma1.deallocate();
     if (allocated(this->Gamma2)) this->Gamma2.deallocate();
     if (allocated(this->s)) this->s.deallocate();
+
+    // write cache data
+    this->writeCacheData(state);
```
mitchute, line 1069: This generates the CTF cached data at the end of CTF generation.

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,113 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
     }
 }
 
+void ConstructionProps::writeCacheData(EnergyPlusData &state)
```
mitchute, line 2096: New construction member method for writing the cache data.

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,113 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+
+    // long verification key
+    cacheData["full_data_key"] = this->getCacheKeyString(state);
+
+    // write to cache
+    std::string key = this->getCacheKey(state);
+    state.dataCache->cache[Cache::CTFKey][key] = cacheData;
+}
+
+void ConstructionProps::loadFromCache(EnergyPlusData &state)
```
mitchute, line 2121: New construction member method for loading the CTF and other cached construction-specific data from the cache.

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,113 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+        Cache::jsonToData(state, this->UValue, thisConstrData, "u_value");
+
+        this->CTFLoadedFromCache = true;
+    } catch (const nlohmann::json::out_of_range &e) {
+        // should already be defaulted to false, but this makes sure
+        this->CTFLoadedFromCache = false;
+    }
+}
+
+std::string ConstructionProps::getCacheKey(EnergyPlusData &state)
```
mitchute, line 2152: New construction member method to create a short key for this construction based on its material layer properties and their ordering.

src/EnergyPlus/Construction.cc
```diff
@@ -2063,4 +2062,113 @@ void ConstructionProps::setArraysBasedOnMaxSolidWinLayers(EnergyPlusData &state)
...
+        key ^= Cache::prepFloatForCacheKey(mat.Conductivity, precision_bits) << Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Density, precision_bits) << Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.SpecHeat, precision_bits) << Layer;
+        key ^= Cache::prepFloatForCacheKey(mat.Resistance, precision_bits) << Layer;
+    }
+
+    return format("{}", key);
+}
+
+std::string ConstructionProps::getCacheKeyString(EnergyPlusData &state)
```
mitchute: New construction member method with all material layer properties for a final check to verify we've found the right cached data.

src/EnergyPlus/Construction.hh
```diff
@@ -268,6 +271,7 @@ namespace Construction {
         // with a heat source/sink.
         int NodeSource;   // Node at which a source or sink is present
         int NodeUserTemp; // Node where user wishes to calculate a temperature (for constructions with sources/sinks only)
+        bool CTFLoadedFromCache;
```
mitchute, line 273: New construction member variable for checking whether the CTFs have been loaded from the cache.

src/EnergyPlus/Data/EnergyPlusData.cc
```diff
@@ -54,13 +54,13 @@ namespace EnergyPlus {
 
 EnergyPlusData::EnergyPlusData()
 {
+    this->dataAFNProps = std::make_unique<DataAFNProps>();
```
mitchute, line 57: Added new methods for the new ```dataCache``` struct. Sorted all other lines.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -76,6 +76,7 @@ struct BoilersData;
 struct BranchInputManagerData;
 struct BranchNodeConnectionsData;
 struct CTElectricGeneratorData;
+struct CacheData;
```
mitchute, line 79: New ```CacheData``` struct. Sorted the file.

src/EnergyPlus/DataGlobals.hh
```diff
@@ -166,6 +166,8 @@ struct DataGlobal : BaseGlobalStruct
     int FDnumIterYears = 0;
     bool printConsoleOutput = true;
 
+    bool useCache = true;
```
mitchute, line 175: New ```DataGlobals``` flag to tell when we should be using the cache. Currently, only used to turn off the cache during unit tests.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -2041,7 +2043,7 @@ namespace HeatBalanceManager {
                 ShowContinueError(state, state.dataIPShortCut->cNumericFieldNames(14) + " not > 0.0 and < 1.0");
             }
 
-            if (MaterialNames(4) == "") {
+            if (MaterialNames(4).empty()) {
```
mitchute, line 2050: Minor, IDE-suggested cleanup.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -5974,7 +5974,7 @@ namespace HeatBalanceManager {
         // na
 
         // SUBROUTINE PARAMETER DEFINITIONS:
-        Real64 const MinLoad(100.0); // Minimum laods for convergence check
+        constexpr Real64 MinLoad(100.0); // Minimum loads for convergence check
```
mitchute, line 6192: A few random ```constexpr``` changes.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -9912,8 +9912,29 @@ namespace HeatBalanceManager {
     {
         bool ErrorsFound(false); // Flag for input error condition
         bool DoCTFErrorReport(false);
-        for (auto &construction : state.dataConstruction->Construct) {
-            construction.calculateTransferFunction(state, ErrorsFound, DoCTFErrorReport);
+
+        if (state.dataCache->ctfObjectsInCache) {
```
mitchute: Updated to check whether the CTFs can be loaded. If they cannot, compute them as normal.

tst/EnergyPlus/unit/CMakeLists.txt
```diff
@@ -17,16 +17,16 @@ set(test_src
...
-    AirflowNetworkSimulationControl.unit.cc
-    AirflowNetworkBalanceManager.unit.cc
-    AirflowNetworkSolver.unit.cc
     AirLoopHVACDOAS.unit.cc
     AirTerminalSingleDuct.unit.cc
     AirTerminalSingleDuctConstantVolumeNoReheat.unit.cc
     AirTerminalSingleDuctMixer.unit.cc
     AirTerminalSingleDuctPIUReheat.unit.cc
-    Autosizing/AutosizingFixture.hh
+    AirflowNetworkBalanceManager.unit.cc
```
mitchute, line 25: Two new unit test files. Sorted the rest of these lines.

tst/EnergyPlus/unit/Cache.unit.cc
```diff
@@ -0,0 +1,160 @@
...
+// JSON Header
+#include <nlohmann/json.hpp>
+
+// EnergyPlus Headers
+#include "Fixtures/EnergyPlusFixture.hh"
+#include <EnergyPlus/Cache.hh>
+
+using namespace EnergyPlus;
+
+TEST_F(EnergyPlusFixture, Cache_TestJSONToData)
```
mitchute: New unit tests.

tst/EnergyPlus/unit/Construction.unit.cc
```diff
@@ -0,0 +1,135 @@
...
+// EnergyPlus Headers
+#include "Fixtures/EnergyPlusFixture.hh"
+#include <EnergyPlus/Cache.hh>
+#include <EnergyPlus/Construction.hh>
+#include <EnergyPlus/Data/EnergyPlusData.hh>
+#include <EnergyPlus/HeatBalanceManager.hh>
+
+using namespace EnergyPlus;
+
+TEST_F(EnergyPlusFixture, Construction_getCacheKeys)
```
mitchute, line 65: New unit tests.

mitchute (commented):
This is ready for review. Feel free to discuss and test it out.

nrel-bot-2b:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

mitchute:
@Myoldmopar @amirroth is there any appetite to get this merged? It doesn't really make any difference to me either way, I'm just looking to clean up the PR list. If you want to try to get it in, then I can update it. If not, I can close it, or just mark it as draft and move the milestone.

amirroth:
> @Myoldmopar @amirroth is there any appetite to get this merged? It doesn't really make any difference to me either way, I'm just looking to clean up the PR list. If you want to try to get it in, then I can update it. If not, I can close it, or just mark it as draft and move the milestone.

So, if we are going to merge this at all, there is no reason to wait. We're not going to find a universal caching file format. I would merge this, making sure to pick up the file reading/writing improvements that Mark Adams just added. My $200.

mitchute:
This has been updated and is ready for review. I ran 5ZoneTDV locally multiple times and compared regression diffs. In the first run, the CTF's were generated and the cache file created. In the second and subsequent runs, they were not and no csv diffs were shown.

```diff
EnergyPlus, Version 9.6.0-a32acb81c5, YMD=2021.09.03 14:28
- Initializing Response Factors
- Calculating CTFs for "ROOF-1"
- Calculating CTFs for "WALL-1"
- Calculating CTFs for "FLOOR-SLAB-1"
- Calculating CTFs for "INT-WALL-1"
Initializing Window Optical Properties
Initializing Solar Calculations
Allocate Solar Module Arrays
```

Then, since most of our example files copy/paste the same construction and material objects, I took 5ZoneAirCooled which has identical constructions to 5ZoneTDV and ran it with the previously generated cache file. The CTF generation step was skipped and the regressions showed zero diffs. Finally, I deleted the cached file and ran again. The CTFs were generated and cached, again with zero diffs.

```diff
EnergyPlus, Version 9.6.0-a32acb81c5, YMD=2021.09.03 15:05
Initializing Response Factors
+ Calculating CTFs for "ROOF-1"
+ Calculating CTFs for "WALL-1"
+ Calculating CTFs for "FLOOR-SLAB-1"
+ Calculating CTFs for "INT-WALL-1"
Initializing Window Optical Properties
Initializing Solar Calculations
```

Myoldmopar:
I was hoping to pop onto this PR and get it merged, but the latest round of commits has got a test failure.  I can try to investigate, but my time may be better suited looking over other PRs.  I'll hold for now and see if @mitchute or @mbadams5 quickly see what happened, and if not I'll try to tackle it later.

mitchute:
I can update it right now. The failure is just mismatched cache keys due to the recent updates Mark made. The actual values keys are arbitrary, so I'll just make them match what E+ is now computing.

mitchute:
Oh, well that's interesting. Using ```std::hash``` is implementation-defined, and not likely to produce the same result on multiple platforms. (https://en.cppreference.com/w/cpp/utility/hash).

It passes as-is on my local machine, but even the CI Mac is failing so it must be creating the hash values in some other way. We probably need to rethink that unit test, at a minimum.

Myoldmopar:
@mitchute @mbadams5 I was able to reproduce the API test failure on Windows.  This comes out of EnergyPlus:

```
** Severe  ** [json.exception.parse_error.113] parse error at byte 216: syntax error while parsing CBOR string: expected length specification (0x60-0x7B) or indefinite string type (0x7F); last byte: 0xFA
```

If you have any quick ideas @mbadams5 that obviously that would be great.  But if this seems like a strange thing that we don't have time to investigate then I'm inclined to suggest punting until after release.

To reproduce the issue: On Windows, build the EnergyPlus library and then from the `build/Products/Debug` folder, run:

```
PYTHONPATH=. /c/Python38-64/python.exe ../../../../tst/EnergyPlus/api/TestRuntime.py -D ../../../../testfiles/1ZoneUncontrolled.idf
```

nrel-bot-3:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-3:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

nrel-bot-2c:
@mitchute @lgentile it has been 28 days since this pull request was last updated.

Myoldmopar:
OK, after some thought, I think I'll table this PR.  The last conversations seemed to indicate that while it seems a harmless change, it also doesn't seem to be as beneficial as we would have hoped.  With plenty of other things to do, and this needing some repairs, I'm just going to the close the PR but leave the branch.  At some point we'll want to bring this conversation back up and will be able to bring back in these changes.  If anyone disagrees, obviously you are welcome to reopen the PR, resolve conflicts, and fix the unit tests.

PR #8729, opened 2021-04-27
Simulate CoilSystem:Cooling:DX using AirloopHVAC:UnitarySystem code base

### LINKED ISSUE #8751: UnitarySystem used in OutdoorAirUnit does not size cooling coil air flow properly
Issue overview
--------------
Testing shows when UnitarySystem is used as child component in OutdoorAirSystem the cooling coil in the UnitarySystem does not autosize air flow properly.

Difference when CoilSystem (-) is changed to UnitarySystem (+):

    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Rated Air Flow Rate [m3/s], 0.42000
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Gross Rated Total Cooling Capacity [W], 9186.05557
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Rated Air Flow Rate [m3/s], 1.26703
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Gross Rated Total Cooling Capacity [W], 27711.88366

### Details

### Checklist
 - [ ] Defect file added (OutdoorAirUnit.idf change CoilSystem to UnitarySystem)

### LINKED ISSUE #8761: UnitarySystem does not correctly size Coil:Cooling:DX:TwoStageWithHumidityControlMode
Issue overview
--------------
The UnitarySystem pre-calculates cooling and heating air flow and capacity. These values are set for use by the coil object. However, for multimode coils the low speed modes are sized first and the saved values set by the UnitarySystem, representing the high speed capacity, is then used by the coil object to set the low speed capacity.

A secondary issue for sizing coils is that the air flow per capacity limits for normal and 100% OA coils is not set in UnitarySystem. This will cause a difference in capacity when UnitarySystem is the parent of a CoilSystem.

        if (this->m_ISHundredPercentDOASDXCoil) {
            state.dataHVACGlobal->DXCT = 2; // uses 100% DX coil flow limits
        } else {
            state.dataHVACGlobal->DXCT = 1; // uses normal DX coil flow limits
        }

### Details

### Checklist
 - [ ] Defect file added (DOASDXCOIL_wADPBFMethod.idf)

### LINKED ISSUE #8849: CoilSystem:Cooling:DX with Coil:Cooling:TwoSpeed does not control humidity
Issue overview
--------------
Coil model does not correctly account for available speeds when calculating operating speed for humidity control.

### Details

### Checklist
 - [x] Defect file added (example file _DOASDXCOIL_wUserSHRMethod.idf)

### LINKED ISSUE #8883: AirloopHVAC:UnitarySystem does not control humidity with Coil:Cooling:DX:VariableSpeed when more than 1 speed is used
Issue overview
--------------
Unit testing showed error when controlling to a humidity ratio set point. The VS coil would not increase compressor speed to meet the humidity ratio set point.

### Details

### Checklist

rraustad:
Pull request overview
---------------------
 - Fixes #8751
 - Fixes #8761
 - Fixes #8849
 - Fixes #8883
 
 - DESCRIBE PURPOSE OF THIS PULL REQUEST: To execute CoilSystem:Cooling:DX within UnitarySystem to reduce maintenance requirements. Several issues, as shown above, were discovered during this transition.

### Pull Request Author

### Reviewer


rraustad:
ASHRAE9012016_SchoolSecondary_Denver error file diff on CI is confused. They are nearly the same. 
[Compare-eplusout.zip](https://github.com/NREL/EnergyPlus/files/6418248/Develop-eplusout.zip)

rraustad:
Impact of attempt to fix row order using Mac results.

image

rraustad:
Progress since yesterday. It's getting harder to find real diff's. I found one this morning with sizing a coil inside a UnitarySystem inside an OutdoorAirUnit.

image

rraustad:
Summary of Diffs. All others not listed are Warmup Convergence diffs. I did not try to determine if Table diffs were a result of row changes.
[Summary-CoilSystemToUnitarySystem.txt](https://github.com/NREL/EnergyPlus/files/6428094/Summary-CoilSystemToUnitarySystem.txt)


    ASHRAE9012016_RestaurantFastFood_Denver
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 25758.87372
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 24602.74334

    ASHRAE9012016_RestaurantSitDown_Denver
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:2_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 29150.58010
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:2_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 27842.22044

    ASHRAE9012016_RetailStandalone_Denver - this new report seems correct but confuses me as to why now
      Component Sizing Information, Coil:Heating:Fuel, PSZ-AC:1_UNITARY_PACKAGE_HEATCOIL, Design Size Nominal Capacity [W], 41416.77458
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Air Flow Rate [m3/s], 1.74507
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC:1_UNITARY_PACKAGE_COOLCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 25957.12624
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Gross Rated Total Cooling Capacity [W], 28887.06000
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Sensible Heat Ratio, 0.79866

    ASHRAE9012016_RetailStripmall_Denver - same as above
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Air Flow Rate [m3/s], 1.75855
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_UNITARY_PACKAGE_COOLCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 24297.89259
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Gross Rated Total Cooling Capacity [W], 29110.29000
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_1:1_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Sensible Heat Ratio, 0.79866

    ASHRAE9012016_SchoolPrimary_Denver - same as above with 1st one removed. Still confusing as to why. The 1st one is greater than 10% different so it should have reported.
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:5_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Air Flow Rate [m3/s], 0.88609
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:5_UNITARY_PACKAGE_COOLCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 17255.86627
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:5_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Gross Rated Total Cooling Capacity [W], 15055.82000
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:5_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Sensible Heat Ratio, 0.78918
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:5_UNITARY_PACKAGE_COOLCOIL, Design Size Low Speed Rated Air Flow Rate [m3/s], 0.29533
    @@ -2050,7 +2049,7 @@
      Component Sizing Information, Fan:ConstantVolume, PSZ-AC_2:7_UNITARY_PACKAGE_FAN, Design Size Maximum Flow Rate [m3/s], 1.48809
      Component Sizing Information, Coil:Heating:Fuel, PSZ-AC_2:7_UNITARY_PACKAGE_HEATCOIL, Design Size Nominal Capacity [W], 83555.97596
      Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:7_UNITARY_PACKAGE_COOLCOIL, User-Specified High Speed Rated Air Flow Rate [m3/s], 1.48809
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:7_UNITARY_PACKAGE_COOLCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 33935.71212
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, PSZ-AC_2:7_UNITARY_PACKAGE_COOLCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 30701.57114

    ASHRAE9012016_SchoolSecondary_Denver - same as above with this new warning which occurred twice
    +   ** Warning ** Enthalpy out of range (PsyTsatFnHPb)
    +   **   ~~~   **  Routine=CalcMultiSpeedDXCoil:newdewpointconditions,
    +   **   ~~~   **  Environment=DENVER-AURORA-BUCKLEY.AFB_CO_USA ANN CLG .4% CONDNS DB=>MWB, at Simulation time=07/21 22:00 - 22:10
    +   **   ~~~   **  Enthalpy=-90928.11610 Pressure= 82237.00
    +   **   ~~~   **  Initial Resultant Temperature= -59.98

    AirflowNetwork_MultiAirLoops - new warning but similar to what was there before
    +   ** Warning ** The mass flow rate difference is found between System Node = 'ZONE EQUIPMENT INLET NODE' and AFN Link = 'MAIN LINK 1'.
    +   **   ~~~   ** The system node max mass flow rate = 0.753 kg/s. The AFN node mass flow rate = 1.364 kg.s.
        ** Warning ** The mass flow rate difference is found between System Node = 'ZONE 1 INLET NODE ATINLET' and AFN Link = 'ZONE1SUPPLY2LINK'.
        **   ~~~   ** The system node max mass flow rate = 0.753 kg/s. The AFN node mass flow rate = 0.764 kg.s.
        ** Warning ** Please adjust the rate of Maximum Air Flow Rate field in the terminal objects or duct pressure resistance.

    AirflowNetwork_MultiZone_SmallOffice_VAV - sizing difference
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, ACDXCOIL 1, Design Size High Speed Rated Air Flow Rate [m3/s], 1.06000
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, ACDXCOIL 1, Design Size High Speed Gross Rated Total Cooling Capacity [W], 17546.76378
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, ACDXCOIL 1, Design Size High Speed Rated Air Flow Rate [m3/s], 1.00461
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, ACDXCOIL 1, Design Size High Speed Gross Rated Total Cooling Capacity [W], 16629.89824

    CmplxGlz_SmOff_IntExtShading - big diffs in Coil System Part Load Ratio

    DOASDXCOIL_wADPBFMethod - sizing differences in all 4 performance modes, this is the 1st mode
    - Component Sizing Information, Coil:Cooling:DX:TwoStageWithHumidityControlMode, DOAS COOLING COIL:DOAS STANDARD PERF 1, Design Size Gross Rated Total Cooling Capacity [W], 8428.75217
    - Component Sizing Information, Coil:Cooling:DX:TwoStageWithHumidityControlMode, DOAS COOLING COIL:DOAS STANDARD PERF 1, Design Size Gross Rated Sensible Heat Ratio, 0.62575
    + Component Sizing Information, Coil:Cooling:DX:TwoStageWithHumidityControlMode, DOAS COOLING COIL:DOAS STANDARD PERF 1, Design Size Gross Rated Total Cooling Capacity [W], 12189.71939
    + Component Sizing Information, Coil:Cooling:DX:TwoStageWithHumidityControlMode, DOAS COOLING COIL:DOAS STANDARD PERF 1, Design Size Gross Rated Sensible Heat Ratio, 0.58872

    DOASDXCOIL_wADPBFMethod_NoReturnPath - same as above

    DisplacementVent_VAV - table diffs with DX Coil Capacity Increase Ratio from Too Low Flow/Capacity Ratio - new issue #8750

    EMSThermochromicWindow - same as above

    HVACTemplate-5ZonePTAC-DOAS - same sizing issue as DOASDXCOIL_wADPBFMethod

    HVACTemplate-5ZonePackagedVAV - same sizing issue as DOASDXCOIL_wADPBFMethod

    HybridVentilationControlGlobalSimple - not sure I caught this in previous files, warmup convergence diffs, point to control differences?
    -Environment:WarmupDays,  6
    +Environment:WarmupDays, 13

    IndEvapCoolerRTUoffice - lots of new warnings
    +   ** Warning ** CalcDoe2DXCoil: Coil:Cooling:DX:SingleSpeed "ACDXCOIL ZN1" - Full load outlet air dry-bulb temperature &lt; 2C. This indicates the possibility of coil frost/freeze. Outlet temperature = -0.12 C.
    +   **   ~~~   **  ...Occurrence info = PHOENIX ANN HTG 99.6% CONDNS DB, 12/21 06:00 - 06:05
    +   **   ~~~   ** ... Possible reasons for low outlet air dry-bulb temperatures are: This DX coil
    +   **   ~~~   **    1) may have a low inlet air dry-bulb temperature. Inlet air temperature = 19.985 C.
    +   **   ~~~   **    2) may have a low air flow rate per watt of cooling capacity. Check inputs.
    +   **   ~~~   **    3) is used as part of a HX assisted cooling coil which uses a high sensible effectiveness. Check inputs.

    OutdoorAirUnit - fixed this morning. And didn't break anything else?
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Rated Air Flow Rate [m3/s], 0.42000
    - Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Gross Rated Total Cooling Capacity [W], 9186.05557
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Rated Air Flow Rate [m3/s], 1.26703
    + Component Sizing Information, Coil:Cooling:DX:SingleSpeed, ACDXCOIL 1, Design Size Gross Rated Total Cooling Capacity [W], 27711.88366

    OutdoorAirUnitwithAirloopHVAC - same as above, must not have noticed this wasn't working correctly when example file was created

    PythonPluginThermochromicWindow - same sizing issue as DOASDXCOIL_wADPBFMethod

    ReliefIndEvapCoolerRTUoffice - new warnings
    +   ** Warning ** CalcDoe2DXCoil: Coil:Cooling:DX:SingleSpeed "ACDXCOIL ZN5" - Full load outlet air dry-bulb temperature < 2C. This indicates the possibility of coil frost/freeze. Outlet temperature = 1.63 C.
    +   **   ~~~   **  ...Occurrence info = PHOENIX ANN HTG 99.6% CONDNS DB, 12/21 06:00 - 06:05
    +   **   ~~~   ** ... Possible reasons for low outlet air dry-bulb temperatures are: This DX coil
    +   **   ~~~   **    1) may have a low inlet air dry-bulb temperature. Inlet air temperature = 20.347 C.
    +   **   ~~~   **    2) may have a low air flow rate per watt of cooling capacity. Check inputs.
    +   **   ~~~   **    3) is used as part of a HX assisted cooling coil which uses a high sensible effectiveness. Check inputs.

    SmOffPSZ - odd results
    DX COOLING COIL SYSTEM 1:Coil System Part Load Ratio  DX COOLING COIL SYSTEM 2:Coil System Part Load Ratio
    1.32E+40	999
    1.32E+40	999
    1.32E+40	999

    SmOffPSZ_OnOffStagedControl - similar to CmplxGlz_SmOff_IntExtShading but the diffs are 1, yuck this is staged control Tstat
    Coil System Part Load Ratio

    SpectralAngularOpticalProperties_TableData - warmup convergence diffs
    -Environment:WarmupDays,  6
    +Environment:WarmupDays, 13

    TermRHDXSystem - same as above

    ThermochromicWindow - same as CmplxGlz_SmOff_IntExtShading, DX Coil Capacity Increase Ratio from Too Low Flow/Capacity Ratio

    TranspiredCollectors - lots of new warnings
    +   ** Warning ** CalcDoe2DXCoil: Coil:Cooling:DX:SingleSpeed "ACDXCOIL ZN1" - Air-cooled condenser inlet dry-bulb temperature below 0 C. Outdoor dry-bulb temperature = -20.00
    +   **   ~~~   **  ... Occurrence info = DENVER_STAPLETON ANN HTG 99.6% CONDNS DB, 12/21 00:00 - 00:15
    +   **   ~~~   ** ... Operation at low ambient temperatures may require special performance curves.

    ZoneCoupledKivaRefBldgMediumOfficeNoClg - sizing issue
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size High Speed Rated Air Flow Rate [m3/s], 2.13856
    - Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 53105.51758
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size High Speed Rated Air Flow Rate [m3/s], 1.11728
    + Component Sizing Information, Coil:Cooling:DX:TwoSpeed, VAV_1_COOLC DXCOIL, Design Size High Speed Gross Rated Total Cooling Capacity [W], 27744.78393

    _DOASDXCOIL_wUserSHRMethod - same as above

    _FanCoilHybridVentAFN - warmup convergence diffs
    -Environment:WarmupDays,  6
    +Environment:WarmupDays, 10

src/EnergyPlus/Autosizing/CoolingAirFlowSizing.cc
```diff
@@ -621,16 +621,16 @@ Real64 CoolingAirFlowSizer::size(EnergyPlusData &state, Real64 _originalValue, b
                         }
                     }
                 } else {
-                    if (state.dataSize->ZoneCoolingOnlyFan) {
+                    if (this->zoneEqSizing(this->curZoneEqNum).CoolingAirFlow) {
+                        this->autoSizedValue = this->zoneEqSizing(this->curZoneEqNum).CoolingAirVolFlow;
+                    } else if (state.dataSize->ZoneCoolingOnlyFan) {
```
rraustad, line 626: #8751 The order here makes a difference. Makes me wonder what the hierarchy should be.  This affected OA Unit child sizing when a child was UnitarySystem.

src/EnergyPlus/Autosizing/HeatingAirFlowSizing.cc
```diff
@@ -167,16 +167,16 @@ Real64 HeatingAirFlowSizer::size(EnergyPlusData &state, Real64 _originalValue, b
                                                    this->dataFlowPerHeatingCapacity * this->dataAutosizedHeatingCapacity);
                     }
                 } else {
-                    if (state.dataSize->ZoneCoolingOnlyFan) {
+                    if (this->zoneEqSizing(this->curZoneEqNum).HeatingAirFlow) {
+                        this->autoSizedValue = this->zoneEqSizing(this->curZoneEqNum).HeatingAirVolFlow;
+                    } else if (state.dataSize->ZoneCoolingOnlyFan) { // probably shouldn't be here
```
rraustad, line 172: #8751 Same here for OA Unit heating coil sizing.

src/EnergyPlus/DXCoils.cc
```diff
@@ -14936,8 +14936,6 @@ void GetFanIndexForTwoSpeedCoil(
     auto &NumPrimaryAirSys = state.dataHVACGlobal->NumPrimaryAirSys;
 
     // SUBROUTINE PARAMETER DEFINITIONS:
-    int const DXSystem(14);      // must match SimAirServingZones.cc (not public)
-    int const UnitarySystem(19); // must match SimAirServingZones.cc (not public)
```
rraustad: Deleted these. Using SimAirServingZones parameters for consistency everywhere.

src/EnergyPlus/DXCoils.cc
```diff
@@ -14966,10 +14965,15 @@ void GetFanIndexForTwoSpeedCoil(
...
-                    break;
+                } else if (state.dataAirSystemsData->PrimaryAirSystems(AirSysNum).Branch(BranchNum).Comp(CompNum).CompType_Num ==
+                           SimAirServingZones::UnitarySystemModel) {
+
+                    if (UtilityRoutines::SameString(state.dataAirSystemsData->PrimaryAirSystems(AirSysNum).Branch(BranchNum).Comp(CompNum).Name,
+                                                    state.dataDXCoils->DXCoil(CoolingCoilIndex).CoilSystemName)) {
+                        FoundBranch = BranchNum;
+                        FoundAirSysNum = AirSysNum;
+                        break;
+                    }
```
rraustad: Added check on name here to match what DXSystem did.

src/EnergyPlus/MixedAir.cc
```diff
@@ -569,7 +569,31 @@ void SimOAComponent(EnergyPlusData &state,
...
+                                                                                            FirstHVACIteration,
+                                                                                            AirLoopNum,
+                                                                                            CompIndex,
+                                                                                            HeatingActive,
+                                                                                            CoolingActive,
+                                                                                            CompIndex,
+                                                                                            OAUCoilOutTemp,
+                                                                                            ZoneEquipFlag,
+                                                                                            sensOut,
+                                                                                            latOut);
```
rraustad: Typical factory and simulate call for OA System, PrimaryAirSystem, Zone equipment and OA Unit.

src/EnergyPlus/MixedAir.cc
```diff
@@ -1097,7 +1121,7 @@ void GetOutsideAirSysInputs(EnergyPlusData &state)
                 } else if (SELECT_CASE_var == "COILSYSTEM:COOLING:DX") {
                     state.dataAirLoop->OutsideAirSys(OASysNum).ComponentType_Num(CompNum) = DXSystem;
                     // set the data for 100% DOAS DX cooling coil
-                    CheckDXCoolingCoilInOASysExists(state, state.dataAirLoop->OutsideAirSys(OASysNum).ComponentName(CompNum));
+                    // CheckDXCoolingCoilInOASysExists(state, state.dataAirLoop->OutsideAirSys(OASysNum).ComponentName(CompNum));
```
rraustad, line 1122: Wasn't sure if this needs to be done here (timing) but appears to be done later.

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -965,6 +965,11 @@ namespace OutdoorAirUnit {
                                               "UNDEFINED");
                             }
                         }
+                        // Must call after SetUpCompSets since this will add another CoilSystem:Cooling:DX object in CompSets
+                        if (OutAirUnit(OAUnitNum).OAEquip(InListNum).ComponentType == "COILSYSTEM:COOLING:DX") {
+                            UnitarySystems::UnitarySys::checkUnitarySysCoilInOASysExists(
+                                state, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName, OAUnitNum);
+                        }
```
rraustad, line 969: Tricky node connections warnings in err file issue.

tst/EnergyPlus/unit/EvaporativeCoolers.unit.cc
```diff
@@ -930,7 +930,12 @@ TEST_F(EnergyPlusFixture, EvapCoolerAirLoopPumpCycling)
 
     // Simulate air loop component calls SimEvapCooler
     // SimEvapCooler calls InitEvapCooler(EvapCoolNum) and CalcDirectEvapCooler
-    SimAirServingZones::SimAirLoopComponent(*state, EvapCond(EvapCoolNum).EvapCoolerName, Evap_Cooler_CompType, false, AirLoopNum, EvapCoolNum, 0);
+    // temporary arguments
+    int airLoopNum = 0;
+    int branchNum = 0;
+    int compNum = 0;
+    SimAirServingZones::SimAirLoopComponent(
+        *state, EvapCond(EvapCoolNum).EvapCoolerName, Evap_Cooler_CompType, false, AirLoopNum, EvapCoolNum, 0, airLoopNum, branchNum, compNum);
```
rraustad: Added last 3 arguments.

src/EnergyPlus/SimAirServingZones.hh
```diff
@@ -153,7 +153,11 @@ namespace SimAirServingZones {
                              bool FirstHVACIteration,     // TRUE if first full HVAC iteration in an HVAC timestep
                              int AirLoopNum,              // Primary air loop number
                              int &CompIndex,              // numeric pointer for CompType/CompName -- passed back from other routines
-                             HVACSystemData *CompPointer);
+                             HVACSystemData *CompPointer, // equipment actual pointer
+                             int const &airLoopNum,       // index to AirloopHVAC
+                             int const &branchNum,        // index to AirloopHVAC branch
+                             int const &compNum           // index to AirloopHVAC branch component
+    );
```
rraustad, line 166: Since this passes a copy of CompPointer, added last 3 arguments so CompPointer could be directly set inside this function in PrimaryAirSystem struct.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -242,7 +242,10 @@ namespace UnitarySystems {
         // MassFlowRateMaxAvail issues are impeding non-VAV air loop equipment by limiting air flow
         // temporarily open up flow limits while simulating, and then set this same value at the INLET after this parent has simulated
         Real64 tempMassFlowRateMaxAvail = state.dataLoopNodes->Node(this->AirInNode).MassFlowRateMaxAvail;
-        state.dataLoopNodes->Node(this->AirInNode).MassFlowRateMaxAvail = this->m_DesignMassFlowRate;
+        // this is not working for CoilSystem simulated with UnitarySystem. Try to protect when this happens.
+        if (AirLoopNum > 0 && this->m_ControlType != ControlType::Setpoint) {
+            state.dataLoopNodes->Node(this->AirInNode).MassFlowRateMaxAvail = this->m_DesignMassFlowRate;
+        }
```
rraustad, line 248: Something is stomping on MaxAvail. Even with this change some files show evidence of issues with flow rate.

E.g., this new warning in AirflowNetwork_MultiAirLoops seems to show an issue with flow rate.

    +   ** Warning ** The mass flow rate difference is found between System Node = 'ZONE EQUIPMENT INLET NODE' and AFN Link = 'MAIN LINK 1'.
    +   **   ~~~   ** The system node max mass flow rate = 0.753 kg/s. The AFN node mass flow rate = 1.364 kg.s.
        ** Warning ** The mass flow rate difference is found between System Node = 'ZONE 1 INLET NODE ATINLET' and AFN Link = 'ZONE1SUPPLY2LINK'.
        **   ~~~   ** The system node max mass flow rate = 0.753 kg/s. The AFN node mass flow rate = 0.764 kg.s.
        ** Warning ** Please adjust the rate of Maximum Air Flow Rate field in the terminal objects or duct pressure resistance.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -515,6 +518,9 @@ namespace UnitarySystems {
                 if (this->m_FanExists && (this->m_CoolCoilExists && (this->m_HeatCoilExists || this->m_SuppCoilExists)))
                     state.dataAirLoop->AirLoopControlInfo(AirLoopNum).UnitarySys = true;
                 state.dataAirLoop->AirLoopControlInfo(AirLoopNum).UnitarySysSimulating = true;
+                if (this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_CoolingTwoSpeed) {
+                    DXCoils::SetCoilSystemCoolingData(state, this->m_CoolingCoilName, this->Name);
+                }
```
rraustad, line 523: I can't recall why I put this here and not in getInput. Probably a late night. This just sets the CoilSystem name in state.dataDXCoils->DXCoil(WhichCoil).CoilSystemName which is then used in SimAirServingZones. Not typical.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1428,6 +1434,9 @@ namespace UnitarySystems {
 
         bool errorsFound(false);
 
+        UnitarySys::getDXCoilSystemData(state, objectName, ZoneEquipment, ZoneOAUnitNum, errorsFound);
+        // call UnitarySystem getInput last, where all report variables will be set up, even if no UnitarySystem objects exist in input.
+        // see function setupAllOutputVars to add new system specific report variables.
         UnitarySys::getUnitarySystemInputData(state, objectName, ZoneEquipment, ZoneOAUnitNum, errorsFound);
```
rraustad: New objects will each have their own getInput and read that object, translate to UnitarySystem inputs and then call processInputSpec. Object specific report variables will be processed last after memory is finalized. Where/when report variable processing occurs is up for debate.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1497,6 +1506,21 @@ namespace UnitarySystems {
...
+            select_EqSizing->HeatingAirFlow = false;
+            select_EqSizing->AirVolFlow = 0.0;
+            select_EqSizing->CoolingAirVolFlow = 0.0;
+            select_EqSizing->HeatingAirVolFlow = 0.0;
+            select_EqSizing->Capacity = false;
+            select_EqSizing->CoolingCapacity = false;
+            select_EqSizing->HeatingCapacity = false;
+            select_EqSizing->DesCoolingLoad = 0.0;
+            select_EqSizing->DesHeatingLoad = 0.0;
+            select_EqSizing->OAVolFlow = 0.0; // UnitarySys doesn't have OA
```
rraustad: #8751 This general reset was zeroing globals set up by OutdoorAirUnit. I think the parent should do the reset, set these up, then size all children. The next parent will reset these.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1506,31 +1530,24 @@ namespace UnitarySystems {
         // Object Data, points to specific array
         DataSizing::ZoneEqSizingData &EqSizing(*select_EqSizing);
 
-        // can't hurt to initialize these going in, probably redundant
```
rraustad, line 1509: Note the comment.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1506,31 +1530,24 @@ namespace UnitarySystems {
...
-        EqSizing.HeatingCapacity = false;
-        EqSizing.DesCoolingLoad = 0.0;
-        EqSizing.DesHeatingLoad = 0.0;
-        EqSizing.OAVolFlow = 0.0; // UnitarySys doesn't have OA
+        // coil sizing requires this information to check proper flow/capacity limits (#8761)
+        if (this->m_ISHundredPercentDOASDXCoil) {
+            state.dataHVACGlobal->DXCT = 2; // uses 100% DX coil flow limits
+        } else {
+            state.dataHVACGlobal->DXCT = 1; // uses normal DX coil flow limits
+        }
```
rraustad: #8761 Missing from UnitarySystem. So when UnitarySystem sized cooling capacity the flow/capacity check was not using the correct min/max limit. Then the coil would size and would use the correct limits.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1506,31 +1530,24 @@ namespace UnitarySystems {
...
 
         bool anyEMSRan;
         EMSManager::ManageEMS(state, EMSManager::EMSCallFrom::UnitarySystemSizing, anyEMSRan, ObjexxFCL::Optional_int_const()); // calling point
         bool HardSizeNoDesRun; // Indicator to a hard-sized field with no design sizing data
 
         // Initiate all reporting variables
-        if (state.dataSize->SysSizingRunDone || state.dataSize->ZoneSizingRunDone) {
+        if (((state.dataSize->CurOASysNum > 0 || state.dataSize->CurSysNum > 0) && state.dataSize->SysSizingRunDone) ||
+            (state.dataSize->CurZoneEqNum > 0 && state.dataSize->ZoneSizingRunDone)) {
```
rraustad: Corrected logic to be more targeted.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1595,7 +1612,10 @@ namespace UnitarySystems {
             if (!this->m_HeatCoilExists) state.dataSize->ZoneCoolingOnlyFan = true;
             TempSize = this->m_MaxCoolAirVolFlow;
             SaveCurDuctType = state.dataSize->CurDuctType;
-            state.dataSize->CurDuctType = DataHVACGlobals::Cooling;
+            // might want to rethink this method. Tries to find the larger of cooling or heating capcity
+            // however, if there is no heating coil the cooling air flow rate is used, not the main flow rate
+            // this is fine if there are no other systems on the branch. CoilSystem does not do this (#8761).
+            if (this->UnitType == "AirLoopHVAC:UnitarySystem") state.dataSize->CurDuctType = DataHVACGlobals::Cooling;
```
rraustad: UnitarySystem is the only parent that tries to find a seasonal capacity by using the actual cooling or heating flow rate. This hack got diffs back in line. I have a feeling that doing this the way UnitarySystem does could size the system flow rate for cooling only systems different than an external fan on the same branch.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1685,12 +1705,19 @@ namespace UnitarySystems {
                 sizerCoolingCapacity.initializeWithinEP(state, CompType, CompName, PrintFlag, RoutineName);
                 CoolCapAtPeak = sizerCoolingCapacity.size(state, TempSize, errorsFound);
                 state.dataSize->DXCoolCap = CoolCapAtPeak;
-                EqSizing.CoolingCapacity = true;
-                EqSizing.DesCoolingLoad = CoolCapAtPeak;
+                // CoilSystem does not size the cooling coil (#8761)
+                if (this->UnitType == "AirLoopHVAC:UnitarySystem") {
+                    EqSizing.CoolingCapacity = true;
+                    EqSizing.DesCoolingLoad = CoolCapAtPeak;
+                }
```
rraustad: Another difference based on parent type. These must get resolved so that these types of checks don't get spread throughout UnitarySystem. Also need an enum here, I got lazy when trying to fix diffs.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1685,12 +1705,19 @@ namespace UnitarySystems {
...
             } else {
                 if (!HardSizeNoDesRun &&
                     (CoolingSAFlowMethod != state.dataUnitarySystems->FlowPerCoolingCapacity && this->m_DesignCoolingCapacity > 0.0)) {
-                    if (this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_MultiSpeedCooling) {
+                    // corrected code for #8756
+                    if (this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_CoolingSingleSpeed ||
+                        this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_MultiSpeedCooling ||
+                        this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_CoolingTwoSpeed ||
+                        this->m_CoolingCoilType_Num == DataHVACGlobals::CoilDX_CoolingTwoStageWHumControl) {
```
rraustad: #8756 Place in code where UnitarySystem was not picking up the index for CapTempModFac for some DX coil types.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1866,7 +1893,7 @@ namespace UnitarySystems {
             state.dataSize->DXCoolCap = EqSizing.DesHeatingLoad;
         }
 
-        if (this->m_OKToPrintSizing) PrintFlag = true;
+        if (this->m_OKToPrintSizing && this->UnitType != "CoilSystem:Cooling:DX") PrintFlag = true;
```
rraustad: CoilSystem does not report any parent sizing information. Could relax this in the future.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -2935,30 +2964,193 @@ namespace UnitarySystems {
                                                                    DataLoopNode::NodeConnectionType::Outlet,
                                                                    1,
                                                                    DataLoopNode::ObjectIsParent);
+        }
+
+        // need to read in all information needed to SetupOutputVariable in setupAllOutputVars
```
rraustad: This became really important over the last few days for setting up report variables.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -2935,30 +2964,193 @@ namespace UnitarySystems {
...
+                this->m_HeatingCoilType_Num = DataHVACGlobals::CoilDX_MultiSpeedHeating;
+            } else if (UtilityRoutines::SameString(loc_heatingCoilType, "Coil:Heating:Water")) {
+                this->m_HeatingCoilType_Num = DataHVACGlobals::Coil_HeatingWater;
+                if (this->m_DesignSpecMSHPIndex > -1) {
+                    this->m_NumOfSpeedHeating = this->m_CompPointerMSHP->numOfSpeedHeating;
+                    if (this->m_NumOfSpeedHeating > 1) {
+                        this->m_MultiSpeedHeatingCoil = true;
+                        this->m_MultiOrVarSpeedHeatCoil = true;
+                    }
+                }
```
rraustad: When setting up report variables the m_HeatingCoilType_Num and m_CoolingCoilType_Num must be known. So they must be read in the first pass through. Also note that m_MultiSpeedHeatingCoil and m_MultiOrVarSpeedHeatCoil are set here which are also used to set up report variables. Bottom line is anything that is needed for report variables must be read in early.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -2935,30 +2964,193 @@ namespace UnitarySystems {
...
+        }
+
+        Real64 loc_m_DesignHRWaterVolumeFlow = input_data.design_heat_recovery_water_flow_rate;
+        this->m_DesignHRWaterVolumeFlow = loc_m_DesignHRWaterVolumeFlow;
+        if (this->m_DesignHRWaterVolumeFlow > 0.0) {
+            this->m_HeatRecActive = true;
+        }
+
         // Early calls to ATMixer don't have enough info to pass GetInput. Need to get the data next time through.
         if (sysNum == -1 || !state.dataZoneEquip->ZoneEquipInputsFilled) {
```
rraustad, line 3155: On first pass UnitarySystem will stop reading here and return back to the main getInput function. All systems will get processed, but all data will not get read in the first time through. Must process all data needed to set up report variables above here. For things like ATMixer, the data is not yet available and need to wait until it is available. Could not think of a way around this.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -7051,400 +7247,19 @@ namespace UnitarySystems {
                 } else {
                     state.dataUnitarySystems->unitarySys[sysNum] = thisSys;
                 }
+            }
+        }
+        if (numAllSystemTypes == state.dataUnitarySystems->numUnitarySystems && state.dataZoneEquip->ZoneEquipInputsFilled)
+            setupAllOutputVars(state, numAllSystemTypes);
+    }
```
rraustad: Call setupAllOutputVars to set up report variables. This might not be the best place for this call but it seems to work for now. I think by the time we get here that all system types have been read in so all that is needed here is if (state.dataZoneEquip->ZoneEquipInputsFilled) then call set up reports.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -14697,27 +14512,29 @@ namespace UnitarySystems {
...
-                    OASysEqSizing(CurOASysNum).HeatingCapacity = false;
-                    this->m_FirstPass = false;
-                } else if (state.dataSize->CurSysNum > 0) {
-                    state.dataAirLoop->AirLoopControlInfo(state.dataSize->CurSysNum).UnitarySysSimulating = false;
-                    DataSizing::resetHVACSizingGlobals(state, state.dataSize->CurZoneEqNum, state.dataSize->CurSysNum, this->m_FirstPass);
-                } else if (state.dataSize->CurZoneEqNum > 0) {
-                    DataSizing::resetHVACSizingGlobals(state, state.dataSize->CurZoneEqNum, state.dataSize->CurSysNum, this->m_FirstPass);
-                } else {
-                    this->m_FirstPass = false;
+            if (AirLoopNum > -1) {
```
rraustad: OutdoorAirUnit passes AirLoopNum = -1. Do not reset globals set by OutdoorAirUnit (this objects parent).

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -5174,17 +5260,22 @@ namespace UnitarySystems {
...
-            } else if (this->m_CoolingCoilType_Num == DataHVACGlobals::Coil_CoolingWaterToAirHPVSEquationFit ||
-                       this->m_CoolingCoilType_Num == DataHVACGlobals::Coil_CoolingAirToAirVariableSpeed) {
-                this->m_ContSpeedCoolingCoil = true;
-            } // CoilDX_Cooling is set above
-
-            if (SetPointManager::NodeHasSPMCtrlVarType(state, this->AirOutNode, SetPointManager::iCtrlVarType::Temp))
-                this->m_SystemCoolControlNodeNum = this->AirOutNode;
-            if (SetPointManager::NodeHasSPMCtrlVarType(state, CoolingCoilOutletNode, SetPointManager::iCtrlVarType::Temp))
-                this->m_SystemCoolControlNodeNum = CoolingCoilOutletNode;
+            if (loc_m_SystemCoolControlNodeName.size() > 0) {
```
rraustad: CoilSystem allows control node to be other than outlet node which is same as coil outlet node. One example file was not running since the control node wasn't known.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -6812,12 +7005,15 @@ namespace UnitarySystems {
         static std::string const getUnitarySystemInput("getUnitarySystemInputData");
 
         std::string cCurrentModuleObject = "AirLoopHVAC:UnitarySystem";
+        int numUnitarySystems = state.dataInputProcessing->inputProcessor->getNumObjectsFound(state, cCurrentModuleObject);
+        int numCoilSystems = state.dataInputProcessing->inputProcessor->getNumObjectsFound(state, "CoilSystem:Cooling:DX");
+        int numAllSystemTypes = numUnitarySystems + numCoilSystems;
```
rraustad, line 7010: Added to keep track of number of systems. Some may not be used but numAllSystemTypes will be equal to state.dataUnitarySystems->numUnitarySystems after all systems are read in. I am thinking this is not needed but have not tried to remove it yet.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -1073,7 +1079,7 @@ namespace UnitarySystems {
                             }
                         }
                     } else {
-                        state.dataLoopNodes->Node(this->AirInNode).MassFlowRate = 0.0;
+                        if (this->UnitType != "CoilSystem:Cooling:DX") state.dataLoopNodes->Node(this->AirInNode).MassFlowRate = 0.0;
```
rraustad: This caused those 4 files to not operate in winter. The CoilSystem:Cooling:DX has a system availability manager that turns off the cooling coil in winter. This shut down the UnitarySystem. The CoilSystem:Cooling:DX object does not set the inlet node flow rate, it just uses what is there. I guess it (or the fan) must also set a flow rate during FirstHVACIteration and then use the node flow rate afterwards. This means that if there are 2 UnitarySystems, one with a cooling coil and the other with a heating coil AND the individual system availability managers are restrictive to their respective operating period then the air loop will shut down.

mjwitte: Seems that a UnitarySystem this is coil-only (no fan) should pass through the inlet node mass flow and conditions even when it's off.

mjwitte: And CoilSystem:Cooling:DX should do the same (native or in UnitarySystem).

rraustad: The air loop must be setting the loop flow on FirstHVACIteration. I don't see in HVACDXSystem where node flow rate is set. I think this is a carry over from when unitary equipment needed to set the inlet node mass flow rate. Will have to figure out when that should not happen for both CoilSystem and UnitarySystem.

rraustad: I think I should be using the fan availability schedule to determine if flow is on or off. The coil system is set point based constant fan mode. Would like to find a way to use the air loop flow if possible. But that only works if the air loop availability schedule matches the fan availability schedule. In HybridVentilationControlGlobalSimple they match.

  Fan:ConstantVolume,
    Supply Fan 1,            !- Name
    FanAndCoilAvailSched,    !- Availability Schedule Name

  AvailabilityManagerAssignmentList,
    Reheat System 1 Avail List,  !- Name
    AvailabilityManager:Scheduled,  !- Availability Manager 1 Object Type
    Reheat System 1 Avail;   !- Availability Manager 1 Name

  AvailabilityManager:Scheduled,
    Reheat System 1 Avail,   !- Name
    FanAndCoilAvailSched;    !- Schedule Name

mjwitte: I disagree. Only the fan should care about availability manager status. Any other component on the branch (including fan-less unitary systems and coil systems) should just do their thing. and HVACDXSystem should not be setting a flow rate, neither should a fan-free unitary system.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -11863,6 +11722,13 @@ namespace UnitarySystems {
                             OutletTemp = state.dataLoopNodes->Node(OutletNode).Temp;
                             if (OutletTemp < DesOutTemp && SensibleLoad) break;
                         }
+                        if (SpeedNum == 1) {
+                            CycRatio = 1.0;
+                            SpeedRatio = 0.0;
+                        } else {
+                            CycRatio = 0.0;
+                            SpeedRatio = 1.0;
+                        }
```
rraustad: Compressor speed ratio was not lining up with DXSystem.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12217,6 +12096,7 @@ namespace UnitarySystems {
                                 this->m_CoolingSpeedRatio = SpeedRatio;
                                 Par[4] = SpeedRatio;
                                 General::SolveRoot(state, Acc, MaxIte, SolFla, CycRatio, this->DXCoilCyclingResidual, 0.0, 1.0, Par);
+                                SpeedRatio = 0.0;
```
rraustad: Same compressor speed ratio fix here where code iterates.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12402,7 +12408,7 @@ namespace UnitarySystems {
                                 //                   RegulaFalsi may not find latent PLR when the latent degradation model is used.
                                 //                   IF iteration limit is exceeded, find tighter boundary of solution and repeat RegulaFalsi
                                 TempMaxPLR = -0.1;
-                                TempOutletHumRatDXCoil = OutletHumRatDXCoil;
+                                TempOutletHumRatDXCoil = DesOutHumRat;
```
rraustad, line 12411: This was causing SolveRoot failure = -2 because OutletHumRatDXCoil is operating outlet w, not the target w. also a tighter min/max limit of 0.01 in the SolveRoot call was too tight (e.g., min=0.78, max=0.79).

rraustad:
I see some big table diffs but when I run a file, 5Zone_Unitary_HXAssistedCoil, in this branch and develop and compare I don't see any difference. This makes sense (that I don't see any differences) since the eio hasn't changed. So I don't know what is causing this.

image

mjwitte:
Hmm, well, that's, um, impossible, unless there's some initialization problem that's showing only on linux/mac. @Myoldmopar may need to confirm the diffs locally.

rraustad:
I would agree if I hadn't seen this myself. I hope it's my mistake. FYI, I have 2 repos, one this branch and one develop. I have 2 MSVS projects open at the same time. I drop the idf and epw in both folders. Then run each, sometimes at the same time, but these ran so fast that they ran independent. Then open each html and compare. The path of each shows I am looking a 2 unique html's in 2 different folders. I was not expecting to see no diffs in my results.

image

rraustad:
As a test I pulled develop on my develop repo. I still get these same results.

rraustad:
Another file with big ESO diffs that looks like the cooling coil outlet is the same compared to develop.

image

Myoldmopar:
Got this tab opened too for review tomorrow morning.

Myoldmopar:
Pulled upstream develop in and everything passes and looks good.  Running regressions locally against latest develop so that I can spot check any remaining concerns before this merges.

rraustad:
I am just looking for comments and suggestions at this point. I want to make another pass at the diffs before this gets merged.

Myoldmopar:
Excellent.  Well I just ran the full regression suite on design days, reporting outputs averaged to the daily aggregate so that tiny little time step diffs might get washed out, and got 8 files with big mathdiffs.  We'll let CI confirm if something went awry with my runs because CI was only reporting 2 big ESO diffs on the last run.

Myoldmopar:
Looking through 5Zone_Unitary_HXAssistedCoil the daily max diff was 66 W of total heat transfer which was a tiny percent change, so not a problem.  I could go back and re-run to get time series diffs on these specific files if it is helpful.

rraustad:
Thanks. Save the names of those 8 files to reduce my effort.

Myoldmopar:
```
  "diffs": {
    "big_math": [
      "5Zone_Unitary_HXAssistedCoil",
      "AirflowNetwork_MultiZone_SmallOffice_VAV",
      "RefBldgFullServiceRestaurantNew2004_Chicago",
      "RefBldgQuickServiceRestaurantNew2004_Chicago",
      "RefBldgPrimarySchoolNew2004_Chicago",
      "RefBldgOutPatientNew2004_Chicago",
      "ShopWithSimplePVT",
      "ShopWithPVandStorage",
    ],
```

rraustad:
I am interested in report variables that I have not looked at that have big diff's. For example, I checked coil outlet air temp and that looks nearly identical. But what about things like cycling ratio or speed ratio that might report different than before but only because setting these reports is different now, not that the coil is operating differently. I saw that in 1 file that @mitchute plotted and I fixed it. I am worried I missed something else, or that 2 different coil types report these differently, etc.

rraustad:
Yeah, those are the files I have been looking at. I'll make a final pass at those before this is merged.

Myoldmopar:
OK, well I am available to create some plots, and @mitchute is of course very good at making the plots as well :+1:

mitchute:
I'm happy to make some plots if it would be helpful.

rraustad:
I'd feel very comfortable with these changes if those 8 files did not show big diff's in any of the reports in the eso/csv. I am also worried that some of these files don't have all the UnitarySystem reports that they should. I also found 1 thing that I can't easily fix that affects the err and html (#8750) and wonder if I have found all those. As you can see I am paranoid of missing something.

Myoldmopar:
OK, so back to just two files with big ESO diffs, that's great news.  Apparently when I made the outputs average daily, it was intended to wash out some of the short time scale diffs, but what it probably did was disproportionately skew the mathdiff tolerances and cause them to show as big when they are actually not as big.  Interesting interaction there.

rraustad:
This looks good so far. The eio diff's in the failed list of 28 files is all warmup convergence. That's a good sign but I still need to look through other diff's (e.g., Table big diffs) for anything I have missed. Also, I am still looking for comments on implementation that can be cleaned up now or through a follow up issue. There are certainly things I did in this branch only to reduce the number of diff's, those will have to be cleaned up at some point.

rraustad:
The 2 files with big eso diffs are _DOASDXCOIL_wUserSHRMethod and DOASDXCOIL_wADPBFMethod. Could one of you plot those to see if something stands out? I would also like to know how to set that up and do that myself since it is a good tool.

mitchute:
I'll make some plots, and also clean up my script so it's a little more user-friendly.

rraustad:
DOASDXCOIL_wADPBFMethod has only 1 CoilSystem in the main air loop. So I looked at the coil outlet conditions. These look nearly identical except for the SPM HumRat set point.

image

And the reason these are different is because HVACDXSystem is overwriting the node humratmax set point while UnitarySystem does not. They both use the adjusted set point based on the result of FrostControlSetPointLimit so this is looking very good now.

    void FrostControlSetPointLimit(EnergyPlusData &state,
                                   int const DXSystemNum,      // dx cooling coil system index
                                   Real64 &TempSetPoint,       // temperature setpoint of the sensor node
                                   Real64 &HumRatSetPoint,     // humidity ratio setpoint of the sensor node

HVACDXSystem (overwrites the node):

    FrostControlSetPointLimit(state,
                              DXSystemNum,
                              state.dataLoopNodes->Node(ControlNode).TempSetPoint,
                              state.dataLoopNodes->Node(ControlNode).HumRatMax,

UnitarySystem (overwrites a local):

    this->frostControlSetPointLimit(state,
                                    state.dataLoopNodes->Node(ControlNode).TempSetPoint,
                                    humRatMaxSP,

rraustad:
Definite diffs in _DOASDXCOIL_wUserSHRMethod that I can't put my finger on yet. Still looking.

rraustad:
I am getting very close on this last file diff. There are mistakes in the logic when humidity control is used for coils with more than 1 speed. I just hope I am not breaking any other files. These traces previously showed big diff's in coil outlet humidity ratio. Now there are only a few blips where something is still slipping through unaccounted for.

These results are at the timestep frequency.

image

@mitchute plots were at hourly frequency but shows how big the diffs were.

image

rraustad:
1 line of code missing where needed causes hours of frustration.

                } else if (FullLoadHumRatOut > (DesOutHumRat - tempHumRatAcc)) {
                    PartLoadFrac = 1.0;
                    SpeedRatio = 1.0;      <---- needed for 2-speed coils

image

rraustad:
I see that 5Zone_Unitary_HXAssisted is now showing diffs again on the CI. I check CMakeLists.txt and the weather file is Chicago_Ohare not Chicago IL as shown in the example file. The weather makes a difference when comparing to results from CI. On the sizing call I said I see no diffs in the html. With the same weather file as CI uses I now see those diff's in mechanical ventilation. With the results I have now I do see that the supply air flow rate has a small difference. However, the supply air T and w are nearly identical so I think that would explain these diffs. Not sure if I can track down why with such a small difference in supply T and w.

image

rraustad:
An even bigger reason for differences in 5Zone_Unitary_HXAssisted is RootSolver warnings no longer occur. 

    ** Warning ** AirLoopHVAC:UnitarySystem - DX unit latent part-load ratio calculation failed unexpectedly: part-load ratio limits exceeded, for unit = GASHEAT DXAC FURNACE 1
    **   ~~~   ** Estimated part-load ratio = 0.525
    **   ~~~   ** The estimated part-load ratio will be used and the simulation continues. Occurrence info:
    **   ~~~   **  Environment=CHICAGO ANN CLG .4% CONDNS WB=>MDB, at Simulation time=07/21 05:56 - 06:00
    ** Warning ** AirLoopHVAC:UnitarySystem - sensible part-load ratio calculation failed: part-load ratio limits exceeded, for unit = GASHEAT DXAC FURNACE 1
    **   ~~~   ** Estimated part-load ratio = 0.525
    **   ~~~   ** The estimated part-load ratio will be used and the simulation continues. Occurrence info:

rraustad:
I am seeing the same similar results in this file and don't see anything worth investigating.

image

rraustad:
RefBldgPrimarySchoolNew2004_Chicago.idf does have diffs in the HW plant water temps, but that's it. All other reports look nearly identical. I don't know what is causing these diffs. For sure the CoilSystem outlet node temps (not humidity controlled) look the same. Just the plant inlet/outlet temps. Scrolling through the plots of all reports and some I added show nothing has changed except HW plant temps. Not sure where to look for the reason for that. Blue lines are this branch in the boiler temp plots. Only diffs in eio are warmup convergence, boiler size did not change.

image

image

rraustad:
For RefBldgQuickServiceRestaurantNew2004_Chicago there are diffs but reviewing all reports shows that they are the same just different by such a small amount that the difference is not discernable by eye. Here's an example of @mitchute plots. Brilliant new method of review, I suggest everyone try it out. Much faster than what I have done in the past with plotting in excel to find diffs.

image

mitchute:
For anyone interested, you can give it shot here: https://github.com/mitchute/plot-eplusout-csv

src/EnergyPlus/DXCoils.cc
mjwitte: @rraustad Note that this used to be hard coded to 19 inside this function. UnitarySystem hasn't been 19 for years. So, when this was enumed, it was set to `Furnace_UnitarySys_HeatOnly` which makes no sense inside a cooling coil function. So, basically this if has been wrong for-ever for UnitarySystem and this fixes it here.

mjwitte:
For RefBldgPrimarySchoolNew2004_Chicago, the boiler diffs happen on the first summer day. In develop, the hot water heating coils are always off and so is the boiler. In this branch, three coils come on randomly a few times with a small load, VAV_POD_1_HEATC, and 2, and 3, especially VAV_POD_2_HEATC which has loads on the order of 500W from 14:30-16:00. The cooling coil is overshooting the setpoint.

image


[RefBldgPrimarySchoolNew2004_Chicago-MoreOutput.idf.txt](https://github.com/NREL/EnergyPlus/files/6787211/RefBldgPrimarySchoolNew2004_Chicago-MoreOutput.idf.txt)

rraustad:
This was a subtle error in logic. Needed to recalculate ReqOutput at the current node humrat. These load variables, Watts: NoOutput, FullOutput and ReqOutput, should be eliminated in a function that is solving for Temp or humrat. Down to 1 test had: ESO big diffs (5Zone_Unitary_HXAssistedCoil). @Myoldmopar would you please look at the other change I made in [e6a9b81](https://github.com/NREL/EnergyPlus/pull/8729/commits/e6a9b81ead54274f5e8e2545b77ddd68a6c598cc) where I switch from int to real in the loop counter. Is that OK or should that be reverted?

image

rraustad:
Hmm, Matt's plot routine says there are no diffs in any of the reports in 5Zone_Unitary_HXAssistedCoil (no plots were generated) yet CI says there are diffs?

First 3 from Make_Plots.ipynb (all are the same as this):

    Skipping: SPACE3-1 ATU IN NODE:System Node Mass Flow Rate [kg/s](Hourly) - no diffs
    Skipping: SPACE5-1:Zone Air Relative Humidity [%](Hourly) - no diffs
    Skipping: SPACE2-1 VAV REHEAT:Zone Air Terminal VAV Damper Position [](Hourly) - no diffs

and I don't see any diff's in the coil outlet temp?

image

rraustad:
Oh yeah, this is the one that uses Chicago IL for weather but runs Chicago Ohare on the CI.

  Site:Location,
    CHICAGO_IL_USA TMY2-94846,  !- Name

add_simulation_test(IDF_FILE 5Zone_Unitary_HXAssistedCoil.idf EPW_FILE USA_IL_Chicago-OHare.Intl.AP.725300_TMY3.epw)

image

rraustad:
Ah Ha! I made a change in this branch to correct when SolFlag == -1. This branch is working correctly. See 2nd to last line in each code. In develop the while loop is comparing TempOutletHumRatDXCoil to OutletHumRatDXCoil, 

    while ((OutletHumRatDXCoil - TempOutletHumRatDXCoil)

but in develop these are the same after the incorrect line of TempOutletHumRatDXCoil = OutletHumRatDXCoil; In this branch it is working correctly.

Develop:

    if (SolFla == -1) {
        //                   RegulaFalsi may not find latent PLR when the latent degradation model is used.
        //                   IF iteration limit is exceeded, find tighter boundary of solution and repeat RegulaFalsi
        TempMaxPLR = -0.1;
        TempOutletHumRatDXCoil = OutletHumRatDXCoil;
        while ((OutletHumRatDXCoil - TempOutletHumRatDXCoil) >= 0.0 && TempMaxPLR <= 1.0) {

This branch:

    if (SolFla == -1) {
        //                   RegulaFalsi may not find latent PLR when the latent degradation model is used.
        //                   IF iteration limit is exceeded, find tighter boundary of solution and repeat RegulaFalsi
        TempMaxPLR = -0.1;
        TempOutletHumRatDXCoil = DesOutHumRat;
        while ((OutletHumRatDXCoil - TempOutletHumRatDXCoil) >= 0.0 && TempMaxPLR <= 1.0) {

Note the smooth PLR in this branch.

image

rraustad:
If I change that 1 line back to the way it was before I get very very small diffs.

image

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12650,27 +12648,34 @@ namespace UnitarySystems {
...
-                                                                      1,
-                                                                      1.0,
-                                                                      SpeedNum,
-                                                                      1.0,
-                                                                      ReqOutput,
-                                                                      dummy,
-                                                                      OnOffAirFlowRatio);
-
-                            OutletHumRatHS = state.dataLoopNodes->Node(this->CoolCoilOutletNodeNum).HumRat;
+                            for (int speedNum = SpeedNum; speedNum <= this->m_NumOfSpeedCooling; ++speedNum) {
```
rraustad, line 12651: Issue found while converting HVACDXSystem.unit.cc tests. The VS coil was not looping through higher speeds to get to the humrat set point. Thanks goodness for unit tests. Documented in #8883.

src/EnergyPlus/MixedAir.cc
```diff
@@ -565,7 +563,6 @@ void SimOAComponent(EnergyPlusData &state,
         OACoolingCoil = true;
     } else if (CompTypeNum == SimAirServingZones::CompType::DXSystem) { // CoilSystem:Cooling:DX  old 'AirLoopHVAC:UnitaryCoolOnly'
```
mjwitte, line 564: @rraustad Do we have a regression test file that uses CoilSystem:Cooling:DX in an AirloopHVAC:OutdoorAirSystem?

rraustad: I have only found water coils in the OA system. I could create one using an existing example file. I also need a test file for the faults model as well.

rraustad: There is something strange happening. I added CoilSystem:Cooling:DX to OA System in FurnaceWithDXSystemRHControl. In this file there is a UnitarySystem in the air loop, and I added CoilSystem to the OA system. That's a good test of multiple objects in the same file.

Existing:

    Branch,
      Air Loop Main Branch,    !- Name
      ,                        !- Pressure Drop Curve Name
      AirLoopHVAC:OutdoorAirSystem,  !- Component 1 Object Type
      OA Sys 1,                !- Component 1 Name
      Outdoor Air Mixer Inlet Node,  !- Component 1 Inlet Node Name
      Mixed Air Node,          !- Component 1 Outlet Node Name
      AirLoopHVAC:UnitarySystem,  !- Component 2 Object Type
      GasHeat DXAC Furnace 1,  !- Component 2 Name
      Mixed Air Node,          !- Component 2 Inlet Node Name
      Air Loop Outlet Node;    !- Component 2 Outlet Node Name

New CoilSystem added to OA system:

    AirLoopHVAC:OutdoorAirSystem:EquipmentList,
      OA Sys 1 Equipment,      !- Name
      CoilSystem:Cooling:DX,   !- Component 1 Object Type
      OA DX Cooling Coil,      !- Component 1 Name
      OutdoorAir:Mixer,        !- Component 2 Object Type
      OA Mixing Box 1;         !- Component 2 Name

Sim is crashing on this->AirInNode = 0 here.

    Real64 tempMassFlowRateMaxAvail = state.dataLoopNodes->Node(this->AirInNode).MassFlowRateMaxAvail;

The thing is I see it setting that node num during getInput. I wonder if this is like report vars that this needs to be set once all systems are gotten. Still looking at this one.

image

Nigusse:
@rraustad I have run to similar problem for coilSystem water coil. I suspect that you may need to add the following code in getDXCoilSystemData() as well:

```
                int sysNum = getUnitarySystemIndex(state, thisObjectName);
                UnitarySys thisSys;
                if (sysNum == -1) {
                    ++state.dataUnitarySystems->numUnitarySystems;
                    auto const &thisObjName = instance.key();
                    state.dataInputProcessing->inputProcessor->markObjectAsUsed(cCurrentModuleObject, thisObjName);
                } else {
                    thisSys = state.dataUnitarySystems->unitarySys[sysNum];
                }
```

rraustad:
I looked at AtticRoof_RadiantBarriers for why it has big table diffs. This input file has a UnitarySystem and the rows are switched in the Annual and Peak Values - Electricity table. The values were the same as before.

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -11807,15 +11693,20 @@ namespace UnitarySystems {
...
 
                         CycRatio = 1.0;
-                        for (SpeedNum = 1; SpeedNum <= this->m_NumOfSpeedCooling; ++SpeedNum) {
-                            SpeedRatio = double(SpeedNum) - 1.0;
+                        for (int speedRatio = 0; speedRatio < this->m_NumOfSpeedCooling; ++speedRatio) {
+                            SpeedRatio = Real64(speedRatio);
                             DXCoils::SimDXCoilMultiSpeed(state, CompName, SpeedRatio, CycRatio, this->m_CoolingCoilIndex);
                             OutletTemp = state.dataDXCoils->DXCoilOutletTemp(this->m_CoolingCoilIndex);
-                            if (OutletTemp < DesOutTemp && SensibleLoad) break; // this isn't going to work IF dehumidIFying
+                            if (SpeedRatio == 1) {
```
Myoldmopar, line 11705: This should be speedRatio not SpeedRatio

src/EnergyPlus/OutdoorAirUnit.cc
```diff
@@ -867,7 +865,8 @@ namespace OutdoorAirUnit {
                             } else if (SELECT_CASE_var == "COILSYSTEM:COOLING:DX") {
                                 OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentType_Num = CompType::DXSystem;
                                 // set the data for 100% DOAS DX cooling coil
-                                CheckDXCoolingCoilInOASysExists(state, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName);
+                                // is a different function call needed here? similar to one in HVACDXSystem
+                                // CheckDXCoolingCoilInOASysExists(state, OutAirUnit(OAUnitNum).OAEquip(CompNum).ComponentName);
```
Myoldmopar, line 869: Need to follow-up on this.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -14409,7 +14409,7 @@ void ComputeTableBodyUsingMovingAvg(EnergyPlusData &state,
             // if exterior is other side coefficients using ground preprocessor terms then
             // set it to ground instead of other side coefficients
             if (curExtBoundCond == OtherSideCoefNoCalcExt || curExtBoundCond == OtherSideCoefCalcExt) {
-                if (has_prefixi(state.dataSurface->OSC(state.dataSurface->Surface(kSurf).OSCPtr).Name, "surfPropOthSdCoef")) {
+                if (has_prefixi(AsString(state.dataSurface->OSC(state.dataSurface->Surface(kSurf).OSCPtr).Name), "surfPropOthSdCoef")) {
```
Myoldmopar, line 14412: Try taking this back out...maybe a fluke?

src/EnergyPlus/SimAirServingZones.cc
```diff
@@ -3464,7 +3463,10 @@ void SimAirLoopComponents(EnergyPlusData &state,
                                 FirstHVACIteration,
                                 AirLoopNum,
                                 PrimaryAirSystems(AirLoopNum).Branch(BranchNum).Comp(CompNum).CompIndex,
-                                PrimaryAirSystems(AirLoopNum).Branch(BranchNum).Comp(CompNum).compPointer);
+                                PrimaryAirSystems(AirLoopNum).Branch(BranchNum).Comp(CompNum).compPointer,
+                                AirLoopNum,
```
Myoldmopar, line 3467: Take out the extra AirLoopNum here.

Myoldmopar (approved):
I walked through this with @rraustad line-by-line.  What a massive effort.  I see nothing red-flag.  There are some yellow-flags that I have noted that we should consider as we clean this up, but this should be ready to go.  I'll do a final check after all the recent merges to develop, but this will drop soon. :balloon:

Myoldmopar:
@Nigusse when this does merge, it will cause conflicts in your work.  Let me know if you want some assistance proceeding with resolving them.  Because of the major work in the unitary system, it may be better to pull your changes in manually.  (Make sure to save your changes before trying the merge).

CI results are looking great, but we want to spot check a couple things:
 - @rraustad will look at DesiccantDehumidifierWithAirToAirCoil tabular output -- it seems the supply fan is not being reported in a certain table row
 - @Myoldmopar will look at UnitarySystem_SubcoolReheatDX to see if the numerics are changing or if the rows are reordered.

Myoldmopar:
We are all good with UnitarySystem_SubcoolReheatDX

Develop:
image

This branch:
image

Just a reorder!  @rraustad if you get that supply fan thing working then we should be ready to merge!

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -523,6 +523,55 @@ namespace UnitarySystems {
...
+                    if (this->m_CoolCoilExists) { // copy this block for heating coil also on future commit
+                        bool coilDataSet = false;
+                        bool errorsFound = false;
+                        std::string coilName;
+                        std::string coilType;
+                        if (this->m_CoolingCoilType_Num == DataHVACGlobals::Coil_CoolingAirToAirVariableSpeed) {
+                            VariableSpeedCoils::getCoilTypeAndName(state, this->m_CoolingCoilIndex, coilType, coilName, errorsFound);
+                            if (!errorsFound) coilDataSet = true;
+                        } else if (this->m_CoolingCoilIndex == DataHVACGlobals::CoilDX_CoolingSingleSpeed) {
+                            // need to check for all other coil types
```
rraustad, line 537: HVACDXSystem only did this for VS coils. Will open this up in a future pass.

        if (!state.dataHVACDXSys->DXCoolingSystem(DXSystemNum).VSCoilFanInfoSet && AirLoopNum > 0) {
            if (state.dataHVACDXSys->DXCoolingSystem(DXSystemNum).CoolingCoilType_Num == Coil_CoolingAirToAirVariableSpeed) {

rraustad: DesiccantDehumidifierWithAirToAirCoil tabular output now reports the fan associated with the DXSystem.

image

There are some files not reporting this information and now I know why. E.g., UnitarySystem_DXCoilSystemAuto which uses a UnitarySystem to mimic the CoilSystem file DXCoilSystemAuto.

image

Myoldmopar: Great!!

Myoldmopar:
Alllright, I think that was the last thing.  This should be ready.  I built once again with latest develop and ran the unit test suite then ran the full regression suite.  Everything looks good.  Still similar to what CI reported last, but now some improvements.  Now DesiccantDehumidiferWithAirToAirCoil is small table diffs instead of large, thanks to @rraustad last change.  Nice.  I'm going to merge this down now.  This is a crucial step forward as we continue to refactor the coils and parent objects in a more manageable and robust direction.

rraustad:
Thank goodness! Who wants to do the next one?

PR #8691, opened 2021-04-02
Update VRF min/max outdoor unit inlet temp field description

rraustad:
Pull request overview
---------------------
 - Fixes #8687
 - User suggestion to clarify min/max operating limit field description

**NOTE: ENHANCEMENTS MUST FOLLOW A SUBMISSION PROCESS INCLUDING A FEATURE PROPOSAL AND DESIGN DOCUMENT PRIOR TO SUBMITTING CODE**

### Pull Request Author
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [x] Title of PR should be user-synopsis style (clearly understandable in a standalone changelog context)
 - [x] Label the PR with at least one of: Defect, Refactoring, NewFeature, Performance, and/or DoNoPublish
 - [x] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [x] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process
 - [ ] If any diffs are expected, author must demonstrate they are justified using plots and descriptions
 - [ ] If changes fix a defect, the fix should be demonstrated in plots and descriptions
 - [ ] If any defect files are updated to a more recent version, upload new versions here or on DevSupport
 - [ ] If IDD requires transition, transition source, rules, ExpandObjects, and IDFs must be updated, and add IDDChange label
 - [ ] If structural output changes, add to output rules file and add OutputChange label
 - [ ] If adding/removing any LaTeX docs or figures, update that document's CMakeLists file dependencies

### Reviewer
This will not be exhaustively relevant to every PR.
 - [ ] Perform a Code Review on GitHub
 - [ ] If branch is behind develop, merge develop and build locally to check for side effects of the merge
 - [ ] If defect, verify by running develop branch and reproducing defect, then running PR and reproducing fix
 - [ ] If feature, test running new feature, try creative ways to break it
 - [ ] CI status: all green or justified
 - [ ] Check that performance is not impacted (CI Linux results include performance check)
 - [ ] Run Unit Test(s) locally
 - [ ] Check any new function arguments for performance impacts
 - [ ] Verify IDF naming conventions and styles, memos and notes and defaults
 - [ ] If new idf included, locally check the err file and other outputs

PR #8738, opened 2021-04-30
epJSON helper functions to get field values

mjwitte:
Pull request overview
---------------------
 - Addresses #7632 
 - Added three new input processor functions to simplify input processing using epJSON field_names instead of `getObjectItem`.
    `getObjectSchemaProps(state, CurrentModuleObject)`
    `getAlphaFieldValue(state, CurrentModuleObject, objectFields, objectSchemaProps, "alpha_field_name")`
    `getRealFieldValue(state, CurrentModuleObject, objectFields, objectSchemaProps, "real_field_name")`
    `getIntFieldValue(state, CurrentModuleObject, objectFields, objectSchemaProps, "integer_field_name")`

ToDos/Questions:
- [x] Add support for extensible fields - do we even have a pattern for that somewhere - grabbing extensible fields without using `getObjectItem`?
- [ ] ~Look for more code re-use opportunities in InputProcessor (e.g. getDefaultValue also searches for the object and field schema)~
- [x] Undo the Airflownetwork getinput changes to avoid conflicts with other AFN PRs.
- [x] Demonstrate the new functions in a getInput function somewhere
- [x] Add extensible fields to the unit tests

### Pull Request Author

### Reviewer


rraustad:
Your on a roll. Is getFieldName on your list? I'd love that for eio reporting.

rraustad:
OK, so I guess that would just be a reformatting of the fieldname passed to these functions.

mjwitte:
> Is getFieldName on your list? I'd love that for eio reporting.

Do you mean pass in the epJSON field name "thermal_conductivity" and return the IDD field name "Thermal Conductivity"?

rraustad:
Yes, I had an issue with getting field names during the sizer function refactor.

rraustad:
And what happens when we change thermal_conductivity to material_thermal_conductivity? Are the field names frozen from here forward? I better get #8691 wrapped up quick.

rraustad:
Seems it needs to be an index/vector, but then that index can't change in the future.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -396,22 +400,22 @@ namespace AirflowNetworkBalanceManager {
                 auto const &thisObjectName = UtilityRoutines::MakeUPPERCase(instance.key());
                 state.dataInputProcessing->inputProcessor->markObjectAsUsed(CurrentModuleObject, instance.key()); // Temporary workaround
 
-                Real64 coeff{fields.at("air_mass_flow_coefficient_at_reference_conditions")}; // Required field
-                Real64 expnt{0.65};
-                if (fields.find("air_mass_flow_exponent") != fields.end()) { // not required field, has default value
-                    expnt = fields.at("air_mass_flow_exponent");
-                }
+                Real64 coeff = state.dataInputProcessing->inputProcessor->getRealFieldValue(
+                    state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions");
```
mjwitte: @amirroth When trying to apply the new `getFieldValue` functions in an actual getInput function, it seemed silly to have patterns like
```
Real64 coeff = 0.0;
state.dataInputProcessing->inputProcessor->getFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions", coeff);
```
By moving to explicitly typed functions, this reduces to one line. Seems more readable.
```
Real64 coeff = state.dataInputProcessing->inputProcessor->getRealFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions");
```

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,86 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+                value = field_value.get<std::int64_t>();
+            } else {
+                value = field_value.get<double>();
+            }
+        } else {
+            bool is_empty = field_value.get<std::string>().empty();
+            if (is_empty) {
+                isDefaulted = findDefault(value, schema_field_obj);
+            } else {
+                value = -99999; // autosize and autocalculate
```
rraustad, line 595: I guess you assume here if !is_empty then the input is either blank or is set to Autosize.

I explicitly tested for that but maybe with this method you don't have to test. Anything other than the word autosize would be caught by the inputprocessor.

                    auto tempFieldVal = fields.at("heating_supply_air_flow_rate");
                    if (tempFieldVal == "Autosize") {
                        loc_m_HeatingSAFMethod_SAFlow = DataSizing::AutoSize;
                    } else {
                        loc_m_HeatingSAFMethod_SAFlow = fields.at("heating_supply_air_flow_rate");
                    }

Do you need to check for?

       \autosizable

mjwitte: I'm not 100% certain about that. In `getRealFieldvalue` I mimicked the code in [`findDefault`](https://github.com/NREL/EnergyPlus/blob/b622e27ca3c96566d6bbc8c6429ae971f880d445/src/EnergyPlus/InputProcessing/InputProcessor.cc#L490-L506) for handling \autosize. It seems a bit fragile to assume if there's a string in a numeric field then it's autosize (or autocalculate). I'm not sure if illegal use of autosize or nonesense words get trapped during initial input processing (and that could be different if the incoming file is idf vs epJSON). Need to do some testing on that.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -4406,6 +4405,202 @@ TEST_F(InputProcessorFixture, epJSONgetObjectItem_minfields)
...
+    // Water heater object
+    objectSchemaProps = state->dataInputProcessing->inputProcessor->getObjectSchemaProps(*state, obj_type1);
+    // User inputs from above
+    // If the field is autosizable and alpha input will return -99999
+    numericFieldValue = state->dataInputProcessing->inputProcessor->getRealFieldValue(*state, obj_type1, wh1, objectSchemaProps, "tank_volume");
+    EXPECT_EQ(numericFieldValue, -99999);
+    numericFieldValue =
+        state->dataInputProcessing->inputProcessor->getRealFieldValue(*state, obj_type1, wh1, objectSchemaProps, "heater_maximum_capacity");
+    EXPECT_EQ(numericFieldValue, -99999);
+    // Even a field that is not autoszable will return -99999 here (assuming that gets checked upon epJSON input processing)
```
rraustad, line 4500: I guess it is safe to assume this will always be caught. I wonder if this is actually unit tested somewhere.

    ** Severe  ** <root>[Building][Building][loads_convergence_tolerance_value] - Value type "string" for input "autosize" not permitted by 'type' constraint.

rraustad: Not sure why it says this input field type is "constraint"?

    "loads_convergence_tolerance_value": {
        "type": "number",

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -4406,6 +4405,202 @@ TEST_F(InputProcessorFixture, epJSONgetObjectItem_minfields)
...
+    state->dataGlobal->isEpJSON = true;
+    state->dataInputProcessing->inputProcessor->initializeMaps();
+    std::string alphaFieldValue;
+    Real64 numericFieldValue = 0.0;
+    json objectSchemaProps;
+
+    // Building object
+    objectSchemaProps = state->dataInputProcessing->inputProcessor->getObjectSchemaProps(*state, obj_type1);
+    // User inputs from above
+    // Note even though choice keys are case-sensitive during epJSON processing, getFieldValue pushes Alphas to UPPERcase
```
rraustad, line 4436: How is retain case handled?

            "group": "Simulation Parameters",
            "name": {
                "type": "string",
                "retaincase": true,
                "default": "NONE"
            },

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,81 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+    if (schema_field_obj.empty()) {
+        ShowFatalError(state, "InputProcessor::fieldValue: Invalid field name = " + fieldName + "for object type = " + objectWord);
+    }
+    bool isDefaulted = false;
+    std::string value;
+    auto it = ep_object.find(fieldName);
+    if (it != ep_object.end()) {
+        auto const &field_value = it.value();
+        if (field_value.is_string()) {
+            auto valuePair = getObjectItemValue(field_value.get<std::string>(), schema_field_obj);
```
rraustad, line 557: I don't see in here where uppercase is set. Is getObjectItemValue where a string gets pushed to uppercase? Can retain case be used here so developers don't make a mistake?

rraustad: Never mind.

    std::pair<std::string, bool> InputProcessor::getObjectItemValue(std::string const &field_value, json const &schema_field_obj)
    {
        if (schema_field_obj.find("retaincase") == schema_field_obj.end()) {
            output.first = UtilityRoutines::MakeUPPERCase(output.first);
        }

src/EnergyPlus/InputProcessing/InputProcessor.hh
```diff
@@ -127,6 +127,14 @@ public:
 
     bool getDefaultValue(EnergyPlusData &state, std::string const &objectWord, std::string const &fieldName, std::string &value);
 
+    std::string getAlphaFieldValue(
+        EnergyPlusData &state, std::string const &objectWord, json const &ep_object, json schema_obj_props, std::string const &fieldName);
+
+    Real64 getRealFieldValue(
+        EnergyPlusData &state, std::string const &objectWord, json const &ep_object, json schema_obj_props, std::string const &fieldName);
```
rraustad: So schema_obj_props is a copy of the original. Should also be referenced?

mjwitte: Yes, good catch. And `const`.

tst/EnergyPlus/unit/InputProcessor.unit.cc
```diff
@@ -4406,6 +4405,202 @@ TEST_F(InputProcessorFixture, epJSONgetObjectItem_minfields)
...
+                *state, obj_type1, fields, objectSchemaProps, "source_side_effectiveness");
+            EXPECT_EQ(numericFieldValue, 0.9);
+            // Also check a field that defaults to autosize (not input above)
+            numericFieldValue = state->dataInputProcessing->inputProcessor->getRealFieldValue(
+                *state, obj_type1, fields, objectSchemaProps, "use_side_design_flow_rate");
+            EXPECT_EQ(numericFieldValue, -99999);
+            // Check an alpha field
+            alphaFieldValue = state->dataInputProcessing->inputProcessor->getAlphaFieldValue(
+                *state, obj_type1, fields, objectSchemaProps, "on_cycle_parasitic_fuel_type");
+            EXPECT_TRUE(UtilityRoutines::SameString(alphaFieldValue, "Electricity"));
```
rraustad, line 4581: Interesting, this gets pushed to lowercase since the field is an enum choice.

    "on_cycle_parasitic_fuel_type": {
        "type": "string",
        "enum": [
            "Electricity",

rraustad:
This all looks very good. I see one more push (const &) and clang format and this is done.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,81 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+        isDefaulted = findDefault(value, schema_field_obj);
+        if (!isDefaulted) {
+            value = 0.0;
+        }
+    }
+    return value;
+}
+json InputProcessor::getObjectSchemaProps(EnergyPlusData &state, std::string const &objectWord)
+{
+    auto const &schema_properties = schema.at("properties");
```
rraustad, line 648: Not sure how you would do this only once but this one line, auto const &schema_properties = schema.at("properties");, is applicable to all objects. Here it is set up again for each call.

mjwitte:
This is ready for review. There is a design doc that shows prototype getInput blocks using these new functions. It still seems like a lot of lines of code are required to set things up before processing the individual fields. Suggestions are welecome.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,82 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+        isDefaulted = findDefault(value, schema_field_obj);
+        if (!isDefaulted) {
+            value = "";
+        }
+    }
+    return value;
+}
+
+Real64 InputProcessor::getRealFieldValue(
+    EnergyPlusData &state, std::string const &objectWord, json const &ep_object, json const &schema_obj_props, std::string const &fieldName)
```
amirroth: Does it make sense to have separate functions for `Real` and `Int`?

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -1560,73 +1560,68 @@ namespace HeatBalanceManager {
...
-                state.dataHeatBal->NominalR(MaterNum) =
-                    state.dataMaterial->Material(MaterNum).Thickness / state.dataMaterial->Material(MaterNum).Conductivity;
-                state.dataMaterial->Material(MaterNum).Resistance = state.dataHeatBal->NominalR(MaterNum);
-            } else {
-                ShowSevereError(state, "Positive thermal conductivity required for material " + state.dataMaterial->Material(MaterNum).Name);
-                ErrorsFound = true;
+                // Load the material derived type from the input data.
+                state.dataMaterial->Material(MaterNum).Group = RegularMaterial;
+                state.dataMaterial->Material(MaterNum).Name = materialName;
+
```
amirroth, line 1596: Just a stylistic thing, but defining `auto & Mat = state.dataMaterial->Material(MaterNum);` and `auto & ip = state.dataInputProcessing->inputProcessor; ` will make the rest of statements in this function into one liners.

src/EnergyPlus/HeatBalanceManager.cc
```diff
@@ -1560,73 +1560,68 @@ namespace HeatBalanceManager {
...
+                std::string roughness = state.dataInputProcessing->inputProcessor->getAlphaFieldValue(
+                    state, state.dataHeatBalMgr->CurrentModuleObject, objectFields, objectSchemaProps, "roughness");
+                ValidateMaterialRoughness(state, MaterNum, roughness, ErrorsFound);
```
amirroth, line 1598: Not for this PR, but just noting that `ValidateMaterialRoughness` and functions of similar ilk will be going soon.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,82 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+        if (field_value.is_number()) {
+            if (field_value.is_number_integer()) {
+                value = field_value.get<std::int64_t>();
+            } else {
+                value = field_value.get<double>();
+            }
+        } else {
+            bool is_empty = field_value.get<std::string>().empty();
+            if (is_empty) {
+                isDefaulted = findDefault(value, schema_field_obj);
```
amirroth, line 593: Need to do something if `isDefaulted` is `false`?

src/EnergyPlus/InputProcessing/InputProcessor.cc
mjwitte: Good question. I started to do that and then realized that the epJSON schema does not differentiate between real and integer: "type": "number"

src/EnergyPlus/InputProcessing/InputProcessor.cc
mjwitte: Another good question. This block of code is borrowed from elsewhere (somewhere in the `getObjectItem` chain. `isDefaulted` gets used elsewhere for some statistics on the number of defaulted fields in the input file. This could probably be simplified a bit here.

src/EnergyPlus/InputProcessing/InputProcessor.cc
rraustad: getRealFieldVal could be changed to getNumericFieldVal if there is no differentiation between int and real. I wonder if returning a real will have any effect when reading into an int variable but since we do that now anyway with getObjectItem's Numbers(x) it shouldn't be a problem.

mbadams5: If you wanted to return `int` or `Real`, you would need to have separate functions since they'd likely have the same input parameter signature and only change the return type.

mbadams5: I will also say with the following, which I saw in an outdated comment in this PR:
```
Real64 coeff = 0.0;
state.dataInputProcessing->inputProcessor->getFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions", coeff);
```
It does seem odd to not directly return the fieldValue, however, this function could be overloaded to include `int` or `string` as input types with their own unique functions. Such as below:
```
int coeff = 0.0;
state.dataInputProcessing->inputProcessor->getFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions", coeff);

std::string coeff = "look a coefficient!";
state.dataInputProcessing->inputProcessor->getFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions", coeff);
```
This means there is only ever one function name to call.
Otherwise, you'll need to have `getRealFieldValue`, `getIntFieldValue`, and `getAlphaFieldValue` functions like below:
```
Real64 coeff = state.dataInputProcessing->inputProcessor->getRealFieldValue(state, CurrentModuleObject, fields, objectSchemaProps, "air_mass_flow_coefficient_at_reference_conditions");
```

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,82 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+    } else {
+        isDefaulted = findDefault(value, schema_field_obj);
+        if (!isDefaulted) {
+            value = 0.0;
+        }
+    }
+    return value;
+}
+
+json InputProcessor::getObjectSchemaProps(EnergyPlusData &state, std::string const &objectWord)
```
mbadams5: This will create a copy of the schema json everytime it is called. It should really return `json const &`, especially since `getPatternProperties` returns `json const &`:
```
json const &InputProcessor::getPatternProperties(EnergyPlusData &state, json const &schema_obj)
```
and you return from `getPatternProperties`
```
json schema_obj_props = getPatternProperties(state, object_schema);
return schema_obj_props;
```
If this current function is called many times it will cause performance issues.

mbadams5 (approved):

mjwitte:
> Otherwise, you'll need to have getRealFieldValue, getIntFieldValue, and getAlphaFieldValue functions like below:

@mbadams5 Is there anywhere in the current input processing that distinguishes between int and real? `getObjectItem` return an array of reals, regardless of the field type in the idd and the schema seems to just have alpha or numeric. So, I'm planning to change `getRealFieldValue` to `getNumericFieldValue` unless I'm missing something related to integer fields.

mbadams5:
@mjwitte No, but mostly that is an artifact of the old InputProcessor and IDD. The IDD does not distinguish between real and int, however, json schema can use `integer` type for fields that should be integer. So as long as we are auto-generating the epjson schema from the IDD, it will likely only ever be alpha or numeric. The same goes for internally within InputProcessor, but there could be a future where there are expanded types usable by epjson schema.

mjwitte:
@mbadams5 Actually, there are some fields in the IDD that are tagged with `\type integer` so we could be using that information in the epJSON schema. I'll add an integer function here for now and then in the future the real and integer functions can validate against the numeric type.

I think I've addressed your `const &` comment in `getObjectSchemaProps`. Please confirm that I got it right.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,73 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
     return defaultFound;
 }
 
+std::string InputProcessor::getAlphaFieldValue(json const &ep_object, json const &schema_obj_props, std::string const &fieldName)
+{
+    auto const &schema_field_obj = schema_obj_props[fieldName];
+    assert(!schema_field_obj.empty()); // Check that field name exists in the schema for this object type
```
mbadams5, line 550: Why did you convert this to an `assert` statement instead of the `FatalErrorMessage`? Asserts only work in debug builds, so is this something only developers will ever hit? If not, then it should probably be the FatalError.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,73 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+    std::string value;
+    auto it = ep_object.find(fieldName);
+    if (it != ep_object.end()) {
+        auto const &field_value = it.value();
+        if (field_value.is_string()) {
+            auto valuePair = getObjectItemValue(field_value.get<std::string>(), schema_field_obj);
+            value = valuePair.first;
+            isDefaulted = valuePair.second;
+        } else {
+            assert(false); // String value requested but field type in numeric
```
mbadams5: Same comment about change to `assert` here. This `assert` statement provides little information except for the inline comment.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,73 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+            value = "";
+        }
+    }
+    return value;
+}
+
+Real64 InputProcessor::getRealFieldValue(json const &ep_object, json const &schema_obj_props, std::string const &fieldName)
+{
+    auto const &schema_field_obj = schema_obj_props[fieldName];
+    assert(!schema_field_obj.empty()); // Check that field name exists in the schema for this object type
```
mbadams5, line 578: Same question about change to `assert`.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -542,6 +542,73 @@ bool InputProcessor::getDefaultValue(EnergyPlusData &state, std::string const &o
...
+        }
+    }
+    return value;
+}
+
+const json &InputProcessor::getObjectSchemaProps(EnergyPlusData &state, std::string const &objectWord)
+{
+    auto const &schema_properties = schema.at("properties");
+    const json &object_schema = schema_properties.at(objectWord);
+    assert(!object_schema.empty()); // If this fails, the object type does not exist in the schema
```
mbadams5, line 650: Same question about change to `assert`.

mbadams5:
@mjwitte I see in the IDD where it has `\type integer` and I also see in `idd_parser.py` script where it finds it but only uses `number` for both real and int. In the future it seems easy to break these apart to include a type integer for epjson schema.

The `getObjectSchemaProps` change looks correct.

src/EnergyPlus/InputProcessing/InputProcessor.cc
mjwitte: I changed to `assert` because this is a developer error (same for the others). The function arguments for `state` and `objectType` were only being used for these error messages that should never happen unless the developer makes a mistake. So it seemed reasonable to use an `assert` and drop those.

src/EnergyPlus/InputProcessing/InputProcessor.cc
mjwitte: I can revert this, but I like the leaner function signature. If there's a better approach, I'm open to suggestions.

rraustad:
@mjwitte and @mbadams5 (thanks for the review) are there any remaining issues? I would like this merged as soon as possible. I am available to click merge at a moments notice.

src/EnergyPlus/InputProcessing/InputProcessor.cc
rraustad: @mjwitte can you give me an example of why schema_field_obj.empty() would ever be false? I get the misspelled field name issue but are there other instances? Inputs past min-fields?

rraustad:
Unit tests run to completion on Win64. This seems ready and any remaining items can be addressed as needed with follow up issues. I am a little surprised there is an indication of a speed improvement albeit small. I will wait on CI for a bit.

rraustad (approved):
I see no significant issues. I am not sure int vs real functions is even an issue.

rraustad:
Failed unit test PackagedTerminalHP_VSCoils_Sizing on Mac is likely unrelated.

src/EnergyPlus/InputProcessing/InputProcessor.cc
mjwitte: I see this is merged, but just for the record, this line is looking at the schema (not the user input) in `schema_obj_props`. So `empty` will be true if the field name does not exist in `schema_obj_props`.

PR #8741, opened 2021-04-30
Remove unused objexx

lefticus:
Pull request overview
---------------------

 - removes parts of objexxfcl that are unused
 - 3 lines changed in EnergyPlus

I'm seeing approximately another 30% reduction in build times on Linux (PCH) and 5-10% smaller binaries.


Myoldmopar:
This is just getting better and better!!  On my Ubuntu 20.04 machine I pulled this branch and built the whole project, including unit tests, in 7.3 minutes!!!  This is a stunning improvement!  Thank you @lefticus, let me know when you are set with your changes here and I'll review them and help get this in.

Myoldmopar (commented):
Stop.  This is madness.  These gains without changing like anything in EnergyPlus.  I mean I read it in your original comment, but I thought maybe you were exaggerating.  Crazy.

lefticus:
@Myoldmopar I think it's ready to go. I could probably keep going, but it feels like a good stopping point.

Myoldmopar:
This is great.  The Windows CI build time has been hovering in the 2000 to 3000 second range, mostly above 2800s.  With this PR, the build time is 1660s.  On Mac, the build time has been around 1000s, and with this it is 815.  On Linux the values are scattered a lot more but the build time here is awesome, and I have the numbers in my local build to support the improvement here.  That is a massive, important, change.  Thanks @lefticus!  Can't wait to see what's next!

PR #8742, opened 2021-04-30
Fix HVAC Sizing Summary Thermostat Setpoint Temperature at Peak Load when no peak heating load

### LINKED ISSUE #8688: HVAC Sizing Summary Thermostat Setpoint Temperature at Peak Load outputs 0C when no peak load
Issue overview
--------------
The HVAC Sizing Summary Zone Sensible Heating Report shows a value of 32F, or 0C, for the Thermostat Setpoint Temperature at Peak Load column when there is no peak heating load in the zone.  Option to either set this to the normal thermostat temperature during the first (or other) timestep, or have this field blank since there is no actual value since there is no peak.

image

[in.zip](https://github.com/NREL/EnergyPlus/files/6407939/in.zip)

### Details
 - EnergyPlus v8.5

### Checklist

matthew-larson:
Pull request overview
---------------------
 - Fixes #8688 
 - This pull request sets the Thermostat Setpoint Temperature at Peak Load to the `ZoneThermostatSetPointLo` variable if there is no peak heating load.  Originally, if `HeatLoad` equals 0, `HeatTstatTemp` never gets set, so the default value of 0 used and reported to the Zone Sensible Heating report.  I added a line to set the `HeatTstatTemp` to the `ZoneThermostatSetPointLo` variable if `HeatLoad` equals 0 so as to override the default value and provide a more reasonable value to the Zone Sensible Heating report.

The alternative is to set the "... Peak Load" values to blank, similar to the Date/Time Of Peak column in the report.

**Note: Need to commit unit test**

Before:
image

After:
image

### Pull Request Author

### Reviewer


src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -668,6 +668,8 @@ void SizeZoneEquipment(EnergyPlusData &state)
             if (state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum) > 0.0) {
                 state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolTstatTemp =
                     state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum);
+                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatTstatTemp =
+                    state.dataHeatBalFanSys->ZoneThermostatSetPointLo(ActualZoneNum);
```
rraustad: I have a question. If TempZoneThermostatSetPoint > 0, and CoolTstatTemp = (is set equal to) TempZoneThermostatSetPoint (line 669 and 670), then why would HeatTstatTemp be set = to ZoneThermostatSetPointLo ? What if ZoneThermostatSetPointLo = 0 ? It hasn't been tested for being > 0. Why not set it to TempZoneThermostatSetPoint ? which is already tested for > 0. Also, you only made 1 change here, that gives me pause given the multiple possible load results. If (heatingload > 0), else take care of everything else.

I realize you are following this existing code. Where ZoneThermostatSetPointLo is used if TempZoneThermostatSetPoint = 0. But here, TempZoneThermostatSetPoint > 0.

        if (state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatLoad > 0.0) {
            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatZoneRetTemp = RetTemp;
            if (state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum) > 0.0) {
                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatTstatTemp =
                    state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum);
            } else {
                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatTstatTemp =
                    state.dataHeatBalFanSys->ZoneThermostatSetPointLo(ActualZoneNum);
            }
        } else {
            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolZoneRetTemp = RetTemp;
            if (state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum) > 0.0) {
                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolTstatTemp =
                    state.dataHeatBalFanSys->TempZoneThermostatSetPoint(ActualZoneNum);
                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatTstatTemp =
                    state.dataHeatBalFanSys->ZoneThermostatSetPointLo(ActualZoneNum);   <<<<<<<
            } else {
                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolTstatTemp =
                    state.dataHeatBalFanSys->ZoneThermostatSetPointHi(ActualZoneNum);
            }
        }

This is my first pass at this and was rather quick. Just pointing out what I saw on this pass.

matthew-larson: Thanks for doing a quick initial pass on this.  I see your reasoning with TempZoneThermostatSetPoint and there are some additional situations I need to test for.  I did try using this variable and the Thermostat Setpoint Temperature at Peak Load results in a somewhat random value that doesn't match any setpoint values.

image

My thinking was adding these lines in the else statement, it basically provides a default for this value that matches the heating setpoint value since it is covering when HeatLoad or CoolLoad is equal to 0.  Maybe there is a better way to achieve this however as I was trying to following existing code.  Is it better to have a separate if statement for if HeatTstatTemp is equal to 0?  This does seem to achieve the same thing.  I can discuss with Neal in more detail this week as well.

matthew-larson:
@rraustad Thinking about this more, is it more appropriate to have the columns related to the Peak Load be blank if there is no Date/Time Of Peak?  There really is no value that is correct in this situation when there is no peak load.  I feel this could be accomplished with a simple check if there is no heating or cooling peak load but wanted to get your thoughts on this first.

rraustad:
Not sure about that. Should probably do the same thing Indoor Temperature at Peak Load is doing Just because there is no load does not mean there is no thermostat.

nealkruis:
@rraustad If there are no loads (and therefore no peak loads), maybe it would make sense to report the values corresponding to the min difference between the indoor temperature and the set point temperature. That way, the report would illustrate how far away from the setpoint the zone actually gets when there is no load.

rraustad:
That sounds like a good solution. For a Tstat temp of 70F, the indoor temp might be 71.2F. That would definitely show why there is no zone load.

matthew-larson:
That makes sense and provides the most value to the user.  I'll work on going this route.

rraustad:
So then it would look like this.

image

nealkruis:
Right, except we might as well add the date/time of the min temperature difference as well (and all the other coincident values in the other columns).

matthew-larson:
I've made updates as suggested to output the Thermostat Setpoint Temperature at Peak Load and the Indoor Temperature at Peak Load to show that there is no heat/cool load and the date/time where the min difference between the indoor temperature and the set point temperature is.  I've resolved the unit test failure but need to look into the integration test and all of the regression test failures to see if there are any issues there.  Below show updated tables from the model based on changes made.  I set 2 zones with high cooling setpoints to ensure there is no cooling load to observe the resulting table.  Let me know if you have any additional thoughts or feedback on anything.

image

image

rraustad:
These diff's make me wonder why Thermostat Setpoint Temperature at Peak Load was 0 in the first place since there is a Tstat in the zone (ZoneTempPredictorCorrector). Either the diff's point to a mistake that was corrected or these changes incorrectly changed the results. I would focus on big diffs in zone and system sizing results to see if those results make sense.

nrel-bot-3:
@matthew-larson @lgentile it has been 28 days since this pull request was last updated.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -1979,8 +1984,128 @@ void UpdateZoneSizing(EnergyPlusData &state, DataGlobalConstants::CallIndicator
...
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).ZoneRetTempAtHeatPeak =
+                                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).HeatZoneRetTempSeq(TimeStepIndex);
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).ZoneHumRatAtHeatPeak =
+                                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).HeatZoneHumRatSeq(TimeStepIndex);
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).OutHumRatAtHeatPeak =
+                                state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).HeatOutHumRatSeq(TimeStepIndex);
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).TimeStepNumAtHeatMax = TimeStepIndex;
+                            NoLoadSizingData = true;
+                        }
+                    }
```
Nigusse, line 2018: First register the first time step value as peak, and then update the peak values only if the current time step values is less the registered peak. Makes sense.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -1979,8 +1984,128 @@ void UpdateZoneSizing(EnergyPlusData &state, DataGlobalConstants::CallIndicator
...
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).TimeStepNumAtHeatMax =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).TimeStepNumAtHeatMax;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).DesHeatCoilInTemp =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).DesHeatCoilInTemp;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).DesHeatCoilInHumRat =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).DesHeatCoilInHumRat;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).HeatTstatTemp =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).HeatTstatTemp;
+                    }
+                }
```
Nigusse, line 2057: OK, then save the updated peak values in CalcFinalZoneSizing() variable.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -1979,8 +1984,128 @@ void UpdateZoneSizing(EnergyPlusData &state, DataGlobalConstants::CallIndicator
...
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).TimeStepNumAtCoolMax =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).TimeStepNumAtCoolMax;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).DesCoolCoilInTemp =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).DesCoolCoilInTemp;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).DesCoolCoilInHumRat =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).DesCoolCoilInHumRat;
+                        state.dataSize->CalcFinalZoneSizing(CtrlZoneNum).CoolTstatTemp =
+                            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, CtrlZoneNum).CoolTstatTemp;
+                    }
+                }
```
Nigusse, line 2117: @matthew-larson For cooling, you are not using the same logic as as the heating. I mean you don't have the *NoLoadSizingData* variable used for cooling. I am wondering why not?

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -673,6 +673,11 @@ void SizeZoneEquipment(EnergyPlusData &state)
                 state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolTstatTemp =
                     state.dataHeatBalFanSys->ZoneThermostatSetPointHi(ActualZoneNum);
             }
+        } else {
+            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatTstatTemp =
+                state.dataHeatBalFanSys->ZoneThermostatSetPointLo(ActualZoneNum);
+            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolTstatTemp =
+                state.dataHeatBalFanSys->ZoneThermostatSetPointHi(ActualZoneNum);
```
Nigusse, line 686: @matthew-larson I am wondering why you did not set the following variable under the *else* condition? This king of changes may lead to unexplained differences (I see big cooling system sizing diffs). The previous code (before the change) is not the same as the current one. Unless there is good reason not to do so, I prefer to maintain the original logic, i.e. set  .CoolZoneRetTemp variable under the *else* logic as well.


`state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolZoneRetTemp = RetTemp;
`

Also in the *else" logic you do not check whether the TempZoneThermostatSetPoint() value is greater than 0. Any reason for not doing it?

tst/EnergyPlus/unit/ZoneEquipmentManager.unit.cc
```diff
@@ -4406,3 +4406,135 @@ TEST_F(EnergyPlusFixture, ZoneAirMassFlowBalance_ZoneMixingInfiltrationFlowsFlag
...
+    state->dataHeatBalFanSys->TempControlType.deallocate();
+    state->dataHeatBalFanSys->TempZoneThermostatSetPoint.deallocate();
+    state->dataHeatBalFanSys->ZoneThermostatSetPointLo.deallocate();
+    state->dataHeatBalFanSys->ZoneThermostatSetPointHi.deallocate();
+    state->dataZoneEnergyDemand->ZoneSysEnergyDemand.deallocate();
+    state->dataZoneEnergyDemand->ZoneSysMoistureDemand.deallocate();
+    state->dataZoneEnergyDemand->DeadBandOrSetback.deallocate();
+    state->dataZoneEnergyDemand->CurDeadBandOrSetback.deallocate();
+    state->dataHeatBalFanSys->ZoneMassBalanceFlag.deallocate();
+    state->dataHeatBal->MassConservation.deallocate();
```
Nigusse: I don't know if you need to deellocate() these variables here. I thought the clear_state() function in the various modules already handle it for you.

Nigusse (changes requested):
@matthew-larson I see around 220 example files show SSZ big diffs. This is concerning.  Have you investigated as to what is causing the big diffs. After all you were trying to extract data values. "Theoretically" it should not change the sizing results, should it? This needs further investigation or justification for the level big diffs observed.

matthew-larson:
@Nigusse Thank you for taking a pass through this, it has turned into a bit of a struggle getting this to work as intended.  The SSZ big diffs are all due to the system sizing resulting in higher cooling capacities, though not by much.  I've been going through these to figure out why.  I'll make some changes based on your comments and hopefully that helps resolve these as well.

tst/EnergyPlus/unit/ZoneEquipmentManager.unit.cc
matthew-larson: I think you're right, I removed the deallocate() methods.

src/EnergyPlus/ZoneEquipmentManager.cc
matthew-larson: It didn't seem to make a difference in the cooling logic, but probably best to include it here as well for consistency.  Added the same logic here.

src/EnergyPlus/ZoneEquipmentManager.cc
matthew-larson: I believe this is where many of the diffs came from, not setting the .`CoolZoneRetTemp` and `.HeatZoneRetTemp` in the `else` logic.  I added this to the `else` logic and this seemed to reduce the number of diffs significantly.  Need to investigate the remaining diffs however.

The reason I didn't check the `TempZoneThermostatSetPoint()` value is that the value of `.HeatTstatTemp` and `.CoolTstatTemp` would be set to `ZoneThermostatSetPointLo` and `ZoneThermostatSetPointHi` respectively no matter the value of `TempZoneThermostatSetPoint()`.

nrel-bot-3:
@matthew-larson @lgentile it has been 28 days since this pull request was last updated.

rraustad:
Now there are only 60 files with SSZ diffs. I agree there should be no diffs there unless there is a good explanation.

matthew-larson:
I agree, I've been working my way back to this one to try to get it finished up.  Many of the diffs are slight differences in heating design capacities, which shouldn't be there.

Myoldmopar:
Windows CI got a little sick the last couple days (Covid negative, just allergies I promise), so ignore that red CI result for now.  If I'm understanding it correctly though, we are expecting this to get to **no diffs**?  Or at least no big diffs?  We will be starting moving toward code freeze early-mid week, so if this feels like it's going to be too much, then go ahead and change the milestone.  Otherwise I'll just be watching for commits to come in.  Thanks @matthew-larson and @rraustad

matthew-larson:
There will be some diffs for the zone sizing information to include the date/time and change in Thermostat Setpoint Temperature at Peak Load but still seeing some diffs in the system sizing that shouldn't be there, hoping to resolve this today.

Myoldmopar:
@matthew-larson I can run the latest commit to check for regressions locally if you would like.  Let me know if you think it's ready for that or not.

Myoldmopar:
I did go ahead and kick off a build/test of this, but didn't make it to regressions.  Even when just running unit tests I encountered a few that failed with:

```
685: ...for Sizing Period: #2 MIAMI INTL AP ANN CLG .4% CONDNS DB=>MWB
685: **FATAL:Please send your input file to the EnergyPlus support/development team for further investigation.
685: EnergyPlus Run Time=00hr 00min  0.14sec
```

That error message only occurs in a few random spots in the code (and obviously the message could be cleaned up, but that's a different day).  Anyway, it feels like this is still not ready.  If it is unlikely to get addressed over the weekend, then this is likely to get punted to after release.

matthew-larson:
I found that too, working through this right now.  I'll do what I can to try to get it resolved but if it looks like too much too quickly, I'll push it back.

matthew-larson:
I was able to resolve the issue with the unit tests.  I'm expecting a handful of diffs with the EIO and HTML files for the Zone Sizing Information adding the date/time and the Component Sizing Information for any system coils,  but will check those in detail once the CI does it's thing and provide justification.

src/EnergyPlus/ZoneEquipmentManager.cc
```diff
@@ -605,22 +605,23 @@ void SizeZoneEquipment(EnergyPlusData &state)
...
             state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatLoad = SysOutputProvided;
             state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatMassFlow = MassFlowRate;
-            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatZoneTemp = Node(ZoneNode).Temp;
-            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).HeatZoneHumRat = Node(ZoneNode).HumRat;
             state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolLoad = 0.0;
             state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolMassFlow = 0.0;
-            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolZoneTemp = 0.0;
-            state.dataSize->CalcZoneSizing(state.dataSize->CurOverallSimDay, ControlledZoneNum).CoolZoneHumRat = 0.0;
-        } else {
+        } else if (SysOutputProvided < 0.0) {
```
matthew-larson, line 610: Broke out parameters for when there is a cooling load and when there is no heating or cooling load.  Always defining the ZoneTemp and ZoeHumRat temp to determine min/max temperature to output minimum temperature at no heating load condition and maximum temperature at no cooling load condition.

Myoldmopar:
Yeah, unit tests do run a bit better now :+1:   Regressions are currently running and I'll report back shortly.

Myoldmopar:
My results, trying to aggregate on the fly so the numbers might be off by a couple:
- 23 files with big math diffs
- 6 files with small math diffs
- ~40 table diffs
- 50 text diffs, mostly EIO, but some ERR and one EDD

src/EnergyPlus/ZoneEquipmentManager.cc
rraustad: Yes, this does give better control of which vars get set based on load, or no load. HeatZoneTemp and CoolZoneTemp are set all the time now, that's what we are doing over in a 3rd party spin off. Nice!

matthew-larson:
Found a small issue that was causing some of the model diffs that shouldn't be there (5ZoneSwimmingPool.idf for example).  Just pulled in develop and pushed up the latest, should get rid of some additional diffs without hopefully causing more with any models.

matthew-larson:
These latest diffs look better as they are only for models that contain a zone(s) that don't contain a heating or cooling load.  I can go into more detail later tonight/tomorrow as well.

Myoldmopar:
I am happy the diffs count has gone down.  This is definitely looking better.  I ran tests locally and everything seems fine.  I can look back over the changes, but would love if anyone else wants to chime in.  @rraustad it looks like above you are pleased with some of the changes as that is what you are already doing in the fork.  If I don't hear otherwise then I'll look this over in a bit and see if I can drop it in.

src/EnergyPlus/ZoneEquipmentManager.cc
Myoldmopar: :+1:

tst/EnergyPlus/unit/ZoneEquipmentManager.unit.cc
```diff
@@ -4556,3 +4556,110 @@ TEST_F(EnergyPlusFixture, CZoeEquipmentManager_CalcZoneLeavingConditions_Test)
...
+
+    // New calculated design values that get reported in the Zone Sensible Heating/Cooling table
+    // When no load, equal to the minimal temperature difference between zone temp and thermostat temp
+    EXPECT_DOUBLE_EQ(23.0, state->dataSize->CalcFinalZoneSizing(1).ZoneTempAtHeatPeak);
+    EXPECT_DOUBLE_EQ(23.0, state->dataSize->CalcFinalZoneSizing(1).ZoneTempAtCoolPeak);
+
+    // Final design values that get passed to the equipment, same as before
+    EXPECT_DOUBLE_EQ(22.0, state->dataSize->FinalZoneSizing(1).ZoneTempAtHeatPeak);
+    EXPECT_DOUBLE_EQ(24.0, state->dataSize->FinalZoneSizing(1).ZoneTempAtCoolPeak);
+}
```
Myoldmopar, line 4665: :+1:

Myoldmopar (approved):

matthew-larson:
Thanks @Myoldmopar!

PR #8745, opened 2021-05-03
Add API Endpoint to Mute Console Output

Myoldmopar:
Pull request overview
---------------------
 - Adds ability to quiet all standard output and standard error messages when calling EnergyPlus as a library

When calling E+ from a script, the EnergyPlus output can get in the way of other diagnostic messages, and currently there was no way to mute them because they wrote directly to cout and cerr.  (Yes there was of course a way, but having to fiddle with overwriting streams is silly when we can just offer an API endpoint to mute the output.)

This mutes standard output messages by adding a single flag in the E+ state that is checked prior to writing a message to either cout or cerr.  

New C Function: `ENERGYPLUSLIB_API void muteConsoleOutput(EnergyPlusState state);`
Python Wrapper: `api.runtime.mute_console_output(state)`

I'm open to naming convention, and I also thought about making it take a bool argument to determine whether to turn off or on, but felt like maybe that was overkill?  I'm open to suggestion based on comments.

### Pull Request Author

### Reviewer


mitchute:
I think I'd prefer that we allow an *optional* bool argument, and make it default ```True```. This gives the user a simple way to switch it on/off if they need it, rather than making them wrap it in an IF/ELSE.

src/EnergyPlus/DataGlobals.hh
```diff
@@ -164,6 +164,7 @@ struct DataGlobal : BaseGlobalStruct
     bool CountNonZoneEquip = true;
     int FDsimDay = 0;
     int FDnumIterYears = 0;
+    bool muteConsoleOutput = false;
```
Myoldmopar: I really prefer using "positive" boolean variable names, as it is much easier to read.  But in this case, the default action is "positive", as in "yes, do print console output".  And the purpose of this flag is to mute the console output, or turn it off.  I could change it everywhere to just be `if (printConsoleOutput) {`, and set it to true.  That's probably the best approach, but at the same time, this variable is only used like 4 almost never touched locations in the code, so it isn't worth a dramatic discussion either.

src/EnergyPlus/api/runtime.h
```diff
@@ -136,6 +136,14 @@ ENERGYPLUSLIB_API int energyplus(EnergyPlusState state, int argc, const char *ar
 /// \param[in] state An active EnergyPlusState instance created with `stateNew`.
 ENERGYPLUSLIB_API void stopSimulation(EnergyPlusState state);
 
+/// \brief Mutes console output (stdout and stderr) to keep terminal clear when running EnergyPlus as a library
+/// \details Normally EnergyPlus will print lots of status update messages while running.  These can get in the way
+///          when calling EnergyPlus as a library.  This function mutes the standard output messages.  Note that these
+///          can still be retrieved using callback functionality.
+/// \param[in] state An active EnergyPlusState instance created with `stateNew`.
+/// \param[in] muteOutput An integer describing whether the console output should be muted or not: nonzero integer for yes, 0 for no.
+ENERGYPLUSLIB_API void setConsoleOutputMuteState(EnergyPlusState state, int muteOutput);
```
Myoldmopar: However, this one could use some wordsmithing.  When the function was purely there to mute the output, it was nicely named `muteConsoleOutput`.  But now that it can mute it, but also allow it to come back on, then that makes the naming a bit funny.  Should I abandon the word mute entirely and just have it be `setConsoleOutputState`?

tst/EnergyPlus/api/TestRuntime.c
```diff
@@ -190,8 +191,15 @@ int main(int argc, const char *argv[])
...
+    // reset and run a test with an external hvac manager
     stateReset(state);
     registerExternalHVACManager(state, externalHVAC);
     energyplus(state, argc, argv);
+    // create a new state and run it with console output muted
+    printf("Running EnergyPlus with Console Output Muted...\n");
+    EnergyPlusState state2 = stateNew();
+    setConsoleOutputMuteState(state2, 1);
+    energyplus(state2, argc, argv);
+    printf("...and it is done.");
```
Myoldmopar, line 203: The C version of this test works very nicely.

tst/EnergyPlus/api/TestRuntime.py
```diff
@@ -80,3 +80,12 @@ def error_handler(severity: int, message: bytes) -> None:
...
 api.runtime.run_energyplus(state, sys.argv[1:])
+
+print("MUTING CONSOLE OUTPUT")
+state2 = api.state_manager.new_state()
+api.runtime.clear_callbacks()
+api.functional.clear_callbacks()
+api.runtime.set_console_output_mute_state(state2, True)
+print("RUNNING MUTED ENERGYPLUS...")
+api.runtime.run_energyplus(state2, sys.argv[1:])
+print("MUTED E+ RUN DONE")
```
Myoldmopar, line 91: And this Python version works nicely too.  When I am running multiple threads of E+ in a single script this really helps.

Myoldmopar (commented):
Open to suggestion on variable naming and more importantly API function naming, if anyone has any thoughts.  I think I will change it to just `setConsoleOutputState` if no one has any specific comments.

src/EnergyPlus/DataGlobals.hh
mitchute: I like ```printConsoleOutput```. That will make code like this,

```C++
if (!state.dataGlobal->muteConsoleOutput) std::cerr << "EnergyPlus Completed Successfully." << std::endl;
``` 

into this,

```C++
if (state.dataGlobal->printConsolOutput) std::cerr << "EnergyPlus Completed Successfully." << std::endl;
``` 

which I think is more natural and easier to read.

src/EnergyPlus/api/runtime.h
mitchute: I agree, the "mute" isn't necessary. ```setConsoleOutputState``` is what I would prefer.

mitchute (commented):
I had a few minor comments on the naming conventions, but have no issues with this otherwise.

Myoldmopar:
I think I addressed it now, I'll do another walk-through of the changes shortly.

src/EnergyPlus/Data/EnergyPlusData.cc
```diff
@@ -312,6 +312,7 @@ EnergyPlusData::EnergyPlusData()
 
 void EnergyPlusData::clear_state()
 {
+    this->ready = true;
```
Myoldmopar, line 315: A new flag is added to make sure someone doesn't try to re-run an EnergyPlus simulation with an already used state.  The same state can be used to make API calls routinely, but you can't kick off a second simulation run without resetting the state.

src/EnergyPlus/Data/EnergyPlusData.hh
```diff
@@ -318,6 +318,7 @@ struct ZoneTempPredictorCorrectorData;
 
 struct EnergyPlusData : BaseGlobalStruct
 {
+    bool ready = true;
```
Myoldmopar, line 321: It is initially true, and reset to true during the call to resetState.

src/EnergyPlus/DataGlobals.hh
```diff
@@ -164,6 +164,7 @@ struct DataGlobal : BaseGlobalStruct
     bool CountNonZoneEquip = true;
     int FDsimDay = 0;
     int FDnumIterYears = 0;
+    bool printConsoleOutput = true;
```
Myoldmopar, line 167: A new state variable for specifying whether we should print messages to the console.  For normal EnergyPlus operation this will be true, but now this can be toggled via API.  This makes it nice to mute the console output while running as a library.

src/EnergyPlus/DisplayRoutines.cc
```diff
@@ -78,6 +78,7 @@ void DisplayString(EnergyPlusData &state, std::string const &String) // String t
     if (state.dataGlobal->messageCallback) state.dataGlobal->messageCallback(String.c_str());
 
     if (state.dataGlobal->KickOffSimulation && !state.dataSysVars->DeveloperFlag) return;
+    if (!state.dataGlobal->printConsoleOutput) return;
```
Myoldmopar, line 81: If we aren't printing to the console right now then return here.

src/EnergyPlus/EPVector.hh
```diff
@@ -91,6 +91,7 @@ template <typename T> struct EPVector : private std::vector<T>
 
     void deallocate() noexcept
     {
+        m_allocated = false;
```
Myoldmopar, line 94: @lefticus I encountered an issue trying to run EnergyPlus twice in a row.  In our clear state methods, we routinely ask any allocated state arrays to deallocate, or clear.  In my testing, I found that even though I was calling the method to deallocate the array, somehow the object was still showing that it was indeed allocated, and caused the second run of EnergyPlus to fail.  I found that the EPVector m_allocated member was not set back to false in a couple places.  Can you confirm this looks OK?

src/EnergyPlus/PlantCentralGSHP.cc
```diff
@@ -1463,7 +1463,6 @@ void GetChillerHeaterInput(EnergyPlusData &state)
                 ShowContinueError(state, "PLR          =    0.00   0.10   0.20   0.30   0.40   0.50   0.60   0.70   0.80   0.90   1.00");
 
                 const auto curve_output = format("Curve Output = {:7.2F}", fmt::join(CurveValArray, ","));
-                std::cout << curve_output << '\n';
```
Myoldmopar, line 1466: Stray print statement removed.

src/EnergyPlus/UtilityRoutines.cc
```diff
@@ -887,7 +887,7 @@ int EndEnergyPlus(EnergyPlusData &state)
 #ifdef EP_Detailed_Timings
     epSummaryTimes(Time_Finish - Time_Start);
 #endif
-    std::cerr << "EnergyPlus Completed Successfully." << std::endl;
+    if (state.dataGlobal->printConsoleOutput) std::cerr << "EnergyPlus Completed Successfully." << std::endl;
```
Myoldmopar, line 890: If we are indeed printing console output, print this message.

src/EnergyPlus/api/runtime.cc
```diff
@@ -66,6 +66,11 @@ int energyplus(EnergyPlusState state, int argc, const char *argv[])
     //    argv[6] = epcomp->iddPath.c_str();
     //    argv[7] = epcomp->idfInputPath.c_str();
     auto *thisState = reinterpret_cast<EnergyPlus::EnergyPlusData *>(state);
+    if (!thisState->ready) {
+        std::cerr << "Attempted to re-run EnergyPlus using a state that was not yet cleared, call stateReset() on this instance and try again\n";
+        return 1;
+    }
+    thisState->ready = false;
```
Myoldmopar, line 73: In this section, we simply return 1 from the `energyplus` function if you try to start a second simulation with a state that is still full of results (and flags, and timesteps, and other data) from a previous run.

src/EnergyPlus/api/runtime.cc
```diff
@@ -75,6 +80,16 @@ void stopSimulation(EnergyPlusState state)
...
 
+void setConsoleOutputState(EnergyPlusState state, int outputStatus)
+{
+    auto thisState = reinterpret_cast<EnergyPlus::EnergyPlusData *>(state);
+    if (outputStatus == 0) {
+        thisState->dataGlobal->printConsoleOutput = false;
+    } else {
+        thisState->dataGlobal->printConsoleOutput = true;
+    }
+}
```
Myoldmopar, line 91: In this new API function we take a 0 for no and a nonzero for yes to toggle on or off the console output.

src/EnergyPlus/api/runtime.h
```diff
@@ -136,6 +136,14 @@ ENERGYPLUSLIB_API int energyplus(EnergyPlusState state, int argc, const char *ar
 /// \param[in] state An active EnergyPlusState instance created with `stateNew`.
 ENERGYPLUSLIB_API void stopSimulation(EnergyPlusState state);
 
+/// \brief Allows disabling and enabling console output (stdout and stderr) during a run.
+/// \details Normally EnergyPlus will print lots of status update messages while running.  These can get in the way
+///          when calling EnergyPlus as a library.  This function allows the client to mute (and unmute) the standard output messages.
+///          Note that output messages can still be retrieved using callback functionality, even with the console output is muted.
+/// \param[in] state An active EnergyPlusState instance created with `stateNew`.
+/// \param[in] outputStatus An integer describing whether the console output should be printed: nonzero integer for yes, 0 for no.
+ENERGYPLUSLIB_API void setConsoleOutputState(EnergyPlusState state, int outputStatus);
```
Myoldmopar, line 145: C function exposed via API.

src/EnergyPlus/api/runtime.py
```diff
@@ -189,6 +191,18 @@ def run_energyplus(self, state: c_void_p, command_line_args: List[Union[str, byt
     # def stop_simulation(self, state: c_void_p) -> None:
     #     pass
 
+    def set_console_output_status(self, state, print_output: bool) -> None:
```
Myoldmopar, line 194: API function with minimal wrapping in the Python API.  For Python we use a simple boolean to toggle the console output on or off.

Myoldmopar (commented):
I believe this is ready to go, but I did do a small bug fix on EPVector and would love @lefticus to bless it.  I'm like 99% sure it's right though.  If this goes in and he looks at it later we can always follow-up with a tweak.

Myoldmopar:
All good on CI's end, and I am done with it.  Does anyone have any comments on it before this drops?

mitchute:
Nothing else from me.

Myoldmopar:
Thanks @mitchute, I say we hold it for a few hours and if no one has issues, go ahead and merge.

mitchute:
tick, tick, tick... ding! Time's up. Merging.

PR #8746, opened 2021-05-03
int constexpr to enum class

dareumnam:
Pull request overview
---------------------
- `int constexpr` to `enum class` in

  - DataHVACControllers
  - DataRoomAirModel
  - DataRuntimeLanguage
  - DataSurfaceColers
  - DataWindowEquivalentLayer
  - FaultsManager
  - FiniteDifferenceGroundTemperatureModel
  - GroundTemperatureModelManager
  - HeatRecovery
  - Humidifiers
  - HVACCooledBeam

### Pull Request Author
 - [X] Author should provide a "walkthrough" of relevant code changes using a GitHub code review comment process

### Reviewer


src/EnergyPlus/DataSurfaceColors.hh
```diff
@@ -92,10 +96,10 @@ namespace DataSurfaceColors {
...
 struct SurfaceColorData : BaseGlobalStruct
 {
-    Array1D_int const defaultcolorno = Array1D_int(DataSurfaceColors::NumColors, {3, 43, 143, 143, 45, 8, 15, 195, 9, 13, 174, 143, 143, 10, 5});
-    Array1D_int DXFcolorno = Array1D_int(DataSurfaceColors::NumColors, SurfaceColorData::defaultcolorno);
+    Array1D_int const defaultcolorno = Array1D_int({0, 14}, {3, 43, 143, 143, 45, 8, 15, 195, 9, 13, 174, 143, 143, 10, 5});
+    Array1D_int DXFcolorno = Array1D_int({0, 14}, SurfaceColorData::defaultcolorno);
 
-    Array1D_string const colorkeys = Array1D_string(DataSurfaceColors::NumColors,
+    Array1D_string const colorkeys = Array1D_string({0, 14},
                                                     {"Text",
```
amirroth, line 103: What's the benefit of `{0,14}` over `ColorNo::NUM` in this constructor?

src/EnergyPlus/GroundTemperatureModeling/GroundTemperatureModelManager.cc
```diff
@@ -81,27 +81,33 @@ namespace GroundTemperatureManager {
...
-        } else if (objectType_str_UPPERCase == UtilityRoutines::MakeUPPERCase(CurrentModuleObjects(objectType_SiteShallowGroundTemp))) {
-            objectType = objectType_SiteShallowGroundTemp;
-        } else if (objectType_str_UPPERCase == UtilityRoutines::MakeUPPERCase(CurrentModuleObjects(objectType_SiteDeepGroundTemp))) {
-            objectType = objectType_SiteDeepGroundTemp;
-        } else if (objectType_str_UPPERCase == UtilityRoutines::MakeUPPERCase(CurrentModuleObjects(objectType_SiteFCFactorMethodGroundTemp))) {
-            objectType = objectType_SiteFCFactorMethodGroundTemp;
-        } else if (objectType_str_UPPERCase == UtilityRoutines::MakeUPPERCase(CurrentModuleObjects(objectType_XingGroundTemp))) {
-            objectType = objectType_XingGroundTemp;
+        if (objectType_str_UPPERCase == UtilityRoutines::MakeUPPERCase(CurrentModuleObjects(static_cast<int>(GroundTempObjType::KusudaGroundTemp)))) {
+            objectType = GroundTempObjType::KusudaGroundTemp;
```
amirroth, line 92: Probably can eliminate `GroundTemp` from the enumerated value itself since it is part of the enumeration class name.

src/EnergyPlus/DataSurfaceColors.hh
amirroth: Is this because `Array1D_string` by default starts at 1 otherwise? I think this was going to be part of the next pass, but if you want you can go ahead and change these Array1D_string const instances to `constexpr std::array<std::string_view>`, which will both start at zero by default and also not construct strings at runtime.

src/EnergyPlus/GroundTemperatureModeling/GroundTemperatureModelManager.cc
mitchute: One thing to note is that most of these ground temp files are hanging out on the EnergyPlus namespace. That's probably something we should correct at some point, but that may affect how we handle your comment here.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -8264,7 +8267,8 @@ void TestSurfTempCalcHeatBalanceInsideSurf(EnergyPlusData &state, Real64 TH12, S
                     } else {
                         ShowContinueError(state, format("...Internal Heat Gain (no floor) [{:.3R}] W", zone.InternalHeatGains / zone.FloorArea));
                     }
-                    if (state.dataAirflowNetwork->SimulateAirflowNetwork <= AirflowNetwork::AirflowNetworkControlSimple) {
+                    if (static_cast<int>(state.dataAirflowNetwork->SimulateAirflowNetwork) <=
+                        static_cast<int>(AirflowNetwork::AirflowNetworkControl::Simple)) {
                         ShowContinueError(state, format("...Infiltration/Ventilation [{:.3R}] m3/s", zone.NominalInfilVent));
```
amirroth: So ... IMO rather than do range comparisons on enumerations, it's clearer to do combined equality tests using the `BITF` and `BITF_TEST_ANY` macros. That way you can see at the call site which options are being compared rather than having to remember which options are `<= AFNControl::Simple`.

src/EnergyPlus/DataSurfaceColors.hh
dareumnam: Yes, it starts at 1 by default. I'll use `constexpr std::array<std::string_view>` so that I could go ahead next pass. Thank you for the feedback.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
dareumnam: I see. I'll figure out this with using `BITF`. Thank you for your comment.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
+    //  MULTIZONE      DISTSYS          Perform distribution system during system on time and multizone calculations during off time
 
     // Types
 
     struct AirflowNetworkSimuProp // Basic parameters for AirflowNetwork simulation
     {
         enum class Solver
         {
-            SkylineLU,
+            Unassigned = -1,
```
jasondegraw: Please remove Unassigned

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
+        SPI, // Splitter inlet Node
+        SPO  // Splitter Outlet Node
     };
 
     enum class iWPCCntr : int
     {
-        Unassigned = 0,
-        Input = 1,
-        SurfAvg = 2
+        Unassigned = -1,
```
jasondegraw: Please revert changes to this enum, it is renamed elsewhere.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
-        CPD = 12, // Distribution system constant pressure drop component
-        COI = 13, // Distribution system coil component
-        TMU = 14, // Distribution system terminal unit component
-        EXF = 15, // Zone exhaust fan
-        HEX = 16, // Distribution system heat exchanger
-        HOP = 17, // Horizontal opening component
-        RVD = 18, // Reheat VAV terminal damper
-        OAF = 19, // Distribution system OA
-        REL = 20  // Distribution system relief air
+        Unassigned = -1,
```
jasondegraw: Revert these changes, this enum is removed entirely elsewhere.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
-        CPD,     // Distribution system constant pressure drop component
-        COI,     // Distribution system coil component
-        TMU,     // Distribution system terminal unit component
-        EXF,     // Zone exhaust fan
-        HEX,     // Distribution system heat exchanger
-        HOP,     // Horizontal opening component
-        RVD,     // Reheat VAV terminal damper
-        OAF,     // Distribution system OA
-        REL      // Distribution system relief air
+        Unassigned = -1,
```
jasondegraw: Revert these changes, there is no need for Unassigned and the new design removes it as a possibility.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
+        Input,
+        SurfAvg
     };
 
-    int constexpr PressureCtrlExhaust = 1;
-    int constexpr PressureCtrlRelief = 2;
+    enum class PressureCtrl
+    {
+        Unassigned = -1,
```
jasondegraw: I am generally against Unassigned, but this one I will leave alone until I can be certain that it's handled properly.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -103,125 +106,132 @@ namespace AirflowNetwork {
...
+        Exhaust,
+        Relief
+    };
 
-    // DERIVED TYPE DEFINITIONS:
+    enum class AirflowNetworkControl
+    {
+        Unassigned = -1,
```
jasondegraw: Same thing here, introducing Unassigned is not a good idea since there is no way it can be unassigned. I'll need to review to be certain but I'd prefer not to add this.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -59,36 +59,39 @@ struct AirflowNetworkData;
...
-        KeepStatus = 2   // Keep status at the previous time step from probability check
+        Unassigned = -1,
+        NoAction,    // No action from probability check
+        ForceChange, // Force open or close from probability check
+        KeepStatus   // Keep status at the previous time step from probability check
     };
 
     enum class EquivRec
     {
+        Unassigned = -1,
```
jasondegraw: I am not sure this is needed, I'll check on my branch.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -59,36 +59,39 @@ struct AirflowNetworkData;
...
+        MinCheckForceOpen, // Force open when opening elapsed time is less than minimum opening time
+        MinCheckForceClose // Force open when closing elapsed time is less than minimum closing time
     };
 
     enum ProbabilityCheck // TODO: make enum class
     {
-        NoAction = 0,    // No action from probability check
-        ForceChange = 1, // Force open or close from probability check
-        KeepStatus = 2   // Keep status at the previous time step from probability check
+        Unassigned = -1,
```
jasondegraw: If it wasn't needed before, I tend to think it is not needed now. I'll check to be certain.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
```diff
@@ -59,36 +59,39 @@ struct AirflowNetworkData;
...
+        AdjEnth    // Enthalpy venting control based on adjacent zone conditions
     };
 
-    enum OpenStatus // TODO: make enum class
+    enum OpenStatus // TODO: make enum class, since this enum is used in SetupOutputVariable()
     {
-        FreeOperation = 0,     // Free operation
-        MinCheckForceOpen = 1, // Force open when opening elapsed time is less than minimum opening time
-        MinCheckForceClose = 2 // Force open when closing elapsed time is less than minimum closing time
+        Uninitialized = -1,
```
jasondegraw: If it wasn't needed before, I tend to think it is not needed now. I'll check to be certain.

jasondegraw (changes requested):
There are some poor choices being extended here that will only cause more merge disasters later. Please revert/undo the changes indicated.

src/EnergyPlus/MundtSimMgr.cc
```diff
@@ -297,12 +293,12 @@ namespace MundtSimMgr {
                         }
 
                         // count air nodes connected to walls in each zone
-                        if (state.dataMundtSimMgr->LineNode(NodeNum, MundtZoneIndex).ClassType == MundtRoomAirNode) {
+                        if (state.dataMundtSimMgr->LineNode(NodeNum, MundtZoneIndex).ClassType == DataRoomAirModel::AirNodeType::MundtRoomAirNode) {
                             ++RoomNodesCount;
```
amirroth, line 297: Same as before, probably don't need `AirNode` in the enumerated values themselves if using scoped enum.

src/EnergyPlus/OutputReportTabular.cc
```diff
@@ -13416,7 +13416,8 @@ void GatherComponentLoadsHVAC(EnergyPlusData &state)
                 ((state.dataHeatBal->ZnAirRpt(state.dataOutRptTab->iZoneGCLH).InfilHeatGain -
                   state.dataHeatBal->ZnAirRpt(state.dataOutRptTab->iZoneGCLH).InfilHeatLoss) /
                  (TimeStepSys * DataGlobalConstants::SecInHour)); // zone infiltration
-            if (state.dataAirflowNetwork->SimulateAirflowNetwork > AirflowNetwork::AirflowNetworkControlSimple) {
+            if (static_cast<int>(state.dataAirflowNetwork->SimulateAirflowNetwork) >
+                static_cast<int>(AirflowNetwork::AirflowNetworkControl::Simple)) {
                 ort->infilInstantSeq(state.dataSize->CurOverallSimDay, state.dataOutRptTab->TimeStepInDayGCLH, state.dataOutRptTab->iZoneGCLH) +=
```
amirroth: Same comment as before about using inequalities on enumerated values. I just think it's less readable.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
amirroth: @jasondegraw, we are going to create a new pattern for all enumerated values that searches `constexpr std::array<std::string_view>` to map string names to enumerated values. That singular lookup function will return -1 if the enumerated string is not found, and the idea was to use -1 as the constructor initialization value as well. I think you can ignore the Unassigned value for all intents and purposes. What problems do you think it will create?

jasondegraw: That pattern (which is not a good pattern at any rate) might work for some things where values can be "unassigned", but in all of the cases I'm pointing to here cannot be unassigned. It's got to be one of the valid choices, and there's a default already assigned. There are IDEs/compilers out there that will warn about not all enumeration values being accounted for in switch statements, so ignoring a superfluous value reinforces bad behavior.

It's a bad pattern because using magic values is always a bad pattern.  Always. It would be more proper C++ to return a `std::optional` and require the caller to take affirmative steps to obtain the value.

amirroth: Ummm, returning an index of -1 when an item (or a NULL if you are dealing with objects and not indices) is not found in list is not a "magic value", it's a pattern as old as C itself. `std::optional` may or may not be "more proper" C++ (is it really more proper for list searches to return an optional? really? why?) but it's also responsible for all of the get nonsense in OpenStudio measures. You can do whatever you want in AFN and you can propagate the `std::optional` pattern to the rest of EnergyPlus after I leave DOE.

jasondegraw: I know the OpenStudio experience with optionals has been... uh... difficult, but I'd argue that there was some overuse there. Maybe a lot of overuse. Dan probably agrees. Anyway, I'd say yes, returning an optional from a search would be a good use of an optional, at least until the "future of errors" presentations that I've seen become reality (static exceptions and all that jazz). It requires the caller to do something about what got returned and is not quite the blunt instrument that `[[nodiscard]]` is.

amirroth: Well, I would argue that until the future of errors is decided, -1 is better than `std::optional` for both sanity and the program critical path. And however distasteful and "bad" this pattern may seem, I am sure we can find at least 29,000 other worse ones to argue about.

src/EnergyPlus/AirflowNetwork/include/AirflowNetwork/Elements.hpp
dareumnam: Thanks for your feedback. I'll revert it on the next commit.

dareumnam: I reverted all the changes that I made in AirflowNetwork.

jasondegraw (approved):

src/EnergyPlus/SQLiteProcedures.cc
```diff
@@ -2672,9 +2672,9 @@ int SQLiteProcedures::sqlitePrepareStatement(sqlite3_stmt *&stmt, const std::str
     return rc;
 }
 
-int SQLiteProcedures::sqliteBindText(sqlite3_stmt *stmt, const int stmtInsertLocationIndex, const std::string &textBuffer)
+int SQLiteProcedures::sqliteBindText(sqlite3_stmt *stmt, const int stmtInsertLocationIndex, const std::string_view &textBuffer)
 {
-    int rc = sqlite3_bind_text(stmt, stmtInsertLocationIndex, textBuffer.c_str(), -1, SQLITE_TRANSIENT);
+    int rc = sqlite3_bind_text(stmt, stmtInsertLocationIndex, format(textBuffer).c_str(), -1, SQLITE_TRANSIENT);
```
dareumnam: Since it passes a string into a C library (`sqlite3.c`), I ended up using a format statement at the deepest level.

amirroth: `std::string_view` supports the `.data()` method which I is functionally equivalent to `.c_str()` for this purpose (you may need to add `static_cast<const char *>`). Can you try that and see if it works? Thx!

Myoldmopar: Since the `data()` method has some caveats about passing it down to C libraries, I had advised her it would be OK to create a string, or a const char* or whatever, to pass down into the C library.  As long as the underlying data is a null terminated string (it should be, right??) then passing `data()` is fine.  However this resolves it will be good to set the precedent for if we have to do this in other places.  Thanks for watching this @amirroth!

amirroth: I see. It's not that hard to create a run-on string in C (a character buffer that is shorter than the string you want will often do it, actually it will often do worse than that). It's harder to do in C++ with `std::string` and `std::string_view` so I would assume this doesn't happen in EnergyPlus, but who knows. Something to put in the parking lot and look at later, maybe. Thx.

dareumnam:
Once I'd like to lock this down so that this can get merged in, and I'll start another PR with reportSizerOutput argument type with `string_view`.

mitchute:
I've got no issue with this - I think it's already been discussed plenty. @amirroth @Myoldmopar @jasondegraw anything further on this one before we continue on with the next set of changes?

I'll go ahead and merge this if no objections arise before CI turns green.

amirroth:
None from me, thank you @dareumnam and all!

mitchute:
Looks great! Merging.

Issue #7673, opened 2020-01-10
Add end-use subcategory fields to all remaining objects that consume electricity or a fuel


JasonGlazer:
Issue overview
--------------
The use of the end-use subcategories is used extensively to show the disaggregation of energy in building models. Currently the user can define end-use subcategories for internal and external load objects (Lights, ElectricEquipment, Exterior:Lights, etc.) and some zone equipment (ZoneBaseboard:OutdoorTemperatureControlled, Fans:OnOff) and central plant equipment (Pump:VariableSpeed, Boiler:HotWater, Chiller:Electric:EIR, CoolingTower:TwoSpeed, WaterHeater:Mixed, Refrigeration:Condenser:WaterCooled).  Unfortunately, this excludes several other classes of objects that do consume (or generate) electricity or fuels such as coils (Coil:Cooling:DX:SingleSpeed, etc.) humidifiers, PV objects, generators. The first step should be to review which objects should have the end-use subcategory field.

## Checklist
Add to this list or remove from it as applicable.  This is a simple templated set of guidelines.
 - [ ] Defect file added (list location of defect file here)
 - [ ] Ticket added to Pivotal for defect (development team task)
 - [ ] Pull request created (the pull request will have additional tasks related to reviewing changes that fix this defect)

mjwitte:
Closely related to https://github.com/NREL/EnergyPlus/issues/4951

benbrannon-arup:
Has there been any discussion of this feature request recently?  It's a pretty fundamental feature in other software and would make E+ substantially more usable in certain projects where differentiating heating and cooling energy is critical (and avoid having to sub up coils, or other workarounds)

PR #8757, opened 2021-05-07
MELs Categorization for Reporting

yujiex:
Pull request overview
---------------------
Energy use by miscellaneous electric loads (MELs) has become a growing portion of energy consumption in residential and commercial buildings in the U.S. and worldwide. Currently,  EnergyPlus reports aggregated plug loads (under General end-use subcategory) without a way to break it down into major subcategories such as set-top boxes, TVs, PCs, etc. reported in the National Energy Modeling System (NEMS). This new feature proposal will implement end-use subcategorization to enable detailed MELs energy calculations and reporting in EnergyPlus to support GEBs research and application of tools such as Scout. The MELs categorization will be standardized based on previous research, DOE appliances standards, building energy codes and standards, as well as their significance in total energy use.

Currently, MELs are aggregated into the Interior Equipment end-use category in EnergyPlus. The default subcategory Interior Equipment does not provide a breakdown of MELs end uses. We propose to add a categorization for MELs which cover major plug loads in residential and commercial buildings

### Pull Request Author

### Reviewer

### DIFF
--- a/design/FY2021/MELsCategorizationForReporting.md
+++ b/design/FY2021/MELsCategorizationForReporting.md
@@ -107,7 +107,10 @@ ElectricEquipment,
        \retaincase
        \default General
 A6 ;   \field User-defined End-Use Subcategory
+       \note when the End-Use Subcategory is UserDefined , A6 - User-defined End-Use Subcategory is required.
+       \note Any text may be used here to categorize the end-uses in the ABUPS End Uses by Subcategory table.
        \type alpha
+       \retaincase
 ```
 
 ## Testing/Validation/Data Source(s): ##
--- a/design/FY2021/MELsCategorizationForReporting.md
+++ b/design/FY2021/MELsCategorizationForReporting.md
@@ -1,0 +1,132 @@
+MELs Categorization for Reporting
+================
+
+**Yujie Xu, Xuan Luo, Tianzhen Hong**
+
+**Lawrence Berkeley National Laboratory**
+
+ - Original Date: April 15, 2021
+ - Modified Date: May 6, 2021
+
+## Justification for New Feature ##
+Energy use by miscellaneous electric loads (MELs) has become a growing portion of energy consumption in residential and commercial buildings in the U.S. and worldwide. Currently,  EnergyPlus reports aggregated plug loads (under General end-use subcategory) without a way to break it down into major subcategories such as set-top boxes, TVs, PCs, etc. reported in the National Energy Modeling System (NEMS) [1]. This new feature proposal will implement end-use subcategorization to enable detailed MELs energy calculations and reporting in EnergyPlus to support GEBs research and application of tools such as Scout. The MELs categorization will be standardized based on previous research, DOE appliances standards, building energy codes and standards, as well as their significance in total energy use.
+
+## Overview ##
+EnergyPlus, up to version 9.5, reports major energy uses as shown in Table 1 where MELs are aggregated into the Interior Equipment end-use category.
+
+|                    | Electricity [GJ] | Natural Gas [GJ] |
+|--------------------|------------------|------------------|
+| Heating            | 0.00             | 5293.02          |
+| Cooling            | 1842.85          | 0.00             |
+| Interior Lighting  | 5636.33          | 0.00             |
+| Exterior Lighting  | 805.32           | 0.00             |
+| Interior Equipment | 8384.16          | 0.00             |
+| Exterior Equipment | 1865.08          | 0.00             |
+| Fans               | 615.79           | 0.00             |
+| Pumps              | 896.10           | 0.00             |
+| Heat Rejection     | 486.64           | 0.00             |
+| Humidification     | 0.00             | 0.00             |
+| Heat Recovery      | 0.00             | 0.00             |
+| Water System       | 0.00             | 246.60           |
+| Refrigeration      | 0.00             | 0.00             |
+| Generators         | 0.00             | 0.00             |
+| Total End Uses     | 20532.27         | 5539.62          |
+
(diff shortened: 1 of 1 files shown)


shorowit:
I don't know how far it makes sense to go with this, but on the residential side see Table 22 from the [Building America House Simulation Protocols](https://www1.eere.energy.gov/buildings/publications/pdfs/building_america/house_simulation.pdf) for two and a half pages of MELs.

JasonGlazer:
The proposal is not clear, will the End Uses By Subcategory table in the Annual Building Utility Performance Summary report includes all of these predefined subcategories with zeros when the result is zero? I hope not it will make the report almost unreadable with so many extra lines. If that is not the goal, then I don't understand the reason for providing specific a specific subcategory list. What will that do that the simple string approach that is used cannot do now? 

My other comment is that this list of subcategories is far from complete and seems to be primarily for residential and offices. There are probably 1000s of additional subcategories if you expand into hospitals/medical buildings, laboratories, commercial kitchens, theaters, museums, police stations, fire stations, airports, auto repair, etc.. It is almost endless. Once the list gets long enough to capture them all, it becomes nearly impossible to find something specific. Lets say you think of a "portable fan" as simply a "fan" and look through the list of 100s of items and do not find "fan" then the user needs to think "what else could a fan be called?"

In general, I don't see a benefit to this addition.

mjwitte:
I agree with @JasonGlazer. It seems this belongs in an interface or other input file generator rather than the EnergyPlus input schema.

hongtz68:
> 
> 
> The proposal is not clear, will the End Uses By Subcategory table in the Annual Building Utility Performance Summary report includes all of these predefined subcategories with zeros when the result is zero? I hope not it will make the report almost unreadable with so many extra lines. If that is not the goal, then I don't understand the reason for providing specific a specific subcategory list. What will that do that the simple string approach that is used cannot do now?
> 
> My other comment is that this list of subcategories is far from complete and seems to be primarily for residential and offices. There are probably 1000s of additional subcategories if you expand into hospitals/medical buildings, laboratories, commercial kitchens, theaters, museums, police stations, fire stations, airports, auto repair, etc.. It is almost endless. Once the list gets long enough to capture them all, it becomes nearly impossible to find something specific. Lets say you think of a "portable fan" as simply a "fan" and look through the list of 100s of items and do not find "fan" then the user needs to think "what else could a fan be called?"
> 
> In general, I don't see a benefit to this addition.

hongtz68:
Jason, thanks for your comments.

Brief response to your questions:
1. The end use by subcategory reporting shows up at this table with those added sub categories when selected by users:
Annual Building Utility Performance Summary - End Uses By Subcategory

The table structure is not changed, only more rows may show up if MELs categories are provided in the IDF files.

2. The MELs list is prioritized based on MELs energy use reported in EIA's Annual Energy Outlook 2017. It is not an inclusive list. Users can define new categories as needed (last choice is UserDefined which allows customization)

3. The MELs standard categorization will support tools such as Scout which uses native EnergyPlus w/o interfaces, and support modeling/analysis of MELs for BTO's MELs subprogram.

Edwin, can we have 10 to 15 minutes to discuss this at May 19's EnergyPlus Technicalities meeting?
 
> 
> The proposal is not clear, will the End Uses By Subcategory table in the Annual Building Utility Performance Summary report includes all of these predefined subcategories with zeros when the result is zero? I hope not it will make the report almost unreadable with so many extra lines. If that is not the goal, then I don't understand the reason for providing specific a specific subcategory list. What will that do that the simple string approach that is used cannot do now?
> 
> My other comment is that this list of subcategories is far from complete and seems to be primarily for residential and offices. There are probably 1000s of additional subcategories if you expand into hospitals/medical buildings, laboratories, commercial kitchens, theaters, museums, police stations, fire stations, airports, auto repair, etc.. It is almost endless. Once the list gets long enough to capture them all, it becomes nearly impossible to find something specific. Lets say you think of a "portable fan" as simply a "fan" and look through the list of 100s of items and do not find "fan" then the user needs to think "what else could a fan be called?"
> 
> In general, I don't see a benefit to this addition.

xuanluo113:
@Myoldmopar

amirroth:
In this example Scout _is_ the interface. I believe the point @JasonGlazer was trying to make is that there is already a facility for any client application to define end-use sub-categories which Scout can use to define the EIA typology. What is the marginal benefit of hardwiring this subset into EnergyPlus?

hongtz68:
One significant benefit is to standardize the naming of MELs for reporting purpose (which serve other purposes, e.g., comparison of MELs energy use/efficiency/control). As Jason said before, fan as user named it, can mean different things, e.g., HVAC fan, ceiling fan, portable fan. By allowing the choice of CeilingFan and PortableFan, we try to standardize the naming; and if user will, it can be customized into any term say MyFan.

> 
> 
> In this example Scout _is_ the interface. I believe the point @JasonGlazer was trying to make is that there is already a facility for any client application to define end-use sub-categories which Scout can use to define the EIA typology. What is the marginal benefit of hardwiring this subset into EnergyPlus?

shorowit:
If there was work being done to improve categorization of EnergyPlus outputs, I would personally prefer work done on:
- Allowing more end use subcategory fields (https://github.com/NREL/EnergyPlus/issues/7673)
- Allowing pump/fan energy to be disaggregated by heating vs cooling -- A) this is needed for a variety of calculations, like HERS/ERI, B) using EMS as a workaround becomes surprisingly complicated when dealing with a large variety of different HVAC systems, and C) this question comes up a lot (like [this Unmet Hours post](https://unmethours.com/question/54329/how-to-calculate-total-space-heatingair-conditioning-in-ep/) from less than a week ago)

These would both have larger benefits in my opinion.

PR #8764, opened 2021-05-10
Improved solve root formulation

Myoldmopar:
Pull request overview
---------------------
 - Changes the formulation of SolveRoot - kinda.  By the end of this set of changes the calls to SolveRoot really didn't need to change for almost every call.  
 - There will be a good amount of code changes, but they are only around the SolveRoot calls, not any widespread changes.

Most of the readers of this will know, but just a little background on SolveRoot.  SolveRoot is simply a root finder that takes in a residual function and performs appropriate iteration and solves the function for when the residual goes to zero.  The function to be passed into the function must be a specific signature to match a SolveRoot capable call.  Over the years, and through several refactors, we ended up with not just one SolveRoot function, but 6 implementations, nearly copy/pasted in General.cc.  However, that's not all, very early on in the state refactor, some preliminary changes took hold and four more copies of SolveRoot were added into the TempSolveRoot namespace.  That means there were **10 full copies** of SolveRoot!!!!!!!  All with very very subtle overload signatures.  

In 226 spots in EnergyPlus, we utilize this solver, and pass in either a static function, or a member function via std::bind, along with a few parameters and an extra parameter Array1D, to perform a wide variety of root solver operations.  The function is passed in as a std::function, which is not an efficient operation.  The std::function is also passed by value, but maybe that in itself is not relevant because it was having to be constructed each call anyway from the raw function pointer.  

Lots of opportunity for improvement here.  
- This work started as me looking at whether we should be passing the function pointers more efficiently.  I tried passing them by reference, but it seemed that the bottleneck was the std::function argument itself, which was silently being constructed from the function pointer.  I tried a few various forms, but it ultimately led me to a template approach.  I had this working, but looked for more impact.
- Having 10 different TempSolveRoots was gross, so those went away.  Unfortunately the different variations were used in different, sometimes dusty, corners of the code, so they couldn't just be deleted.  Spoiler alert though, we are down to **1 single copy of SolveRoot** now.
- My next step was to address the fact that in every single call to SolveRoot, an Array1D must be allocated to hold the parameter array being passed into the SolveRoot function.  Of course the Array1D itself is passed by reference, so there is no cost there, but the Array1D is allocated prior to each call to the solver, leaving the door open for a runtime impact.  Even if it isn't a dramatic runtime impact, the fact of the matter is that we know the size of each of these parameter arrays at compile time, and so they shouldn't be an Array1D, or even a vector, they should be using std::array.  So I started converting these over.  I ended up changing a ton of the Array1Ds into properly fixed size array<Real64, N> objects, which can be more intelligently optimized by the compiler since the size is known at compile time.
- However this led to the templates getting a bit complex, because an `array<Real64, 2>` is not the same type as an `array<Real64, 3>`.  The compiler could not perform deduction on the size of the array, so the size of the parameter array was specified as a template parameter.  Definitely doable, but not great.
- After a conversation and some prototyping with @lefticus, we came up with a better template solution.  The solution allows for the compiler to do automatic deduction on the template parameters, so there is actually **no need** for a modification to the SolveRoot call.  Much of the code changes in this PR are from back when I was making the `Array1D` into a `std::array`.  This should definitely continue for code optimization, but it's not necessary to get it compiled and running.

Code changes you will find below include:
- Lots of changes from `TempSolveRoot::SolveRoot` to `General::SolveRoot`
- Changing Par array objects from `Array1D` to `std::array<Real64, N>`, which must be made in the code calling SolveRoot as well as in the residual function that accepts the array while solving the equation.  So obviously this comes with accompanying changes from `Par(2)` to `Par[1]`.
- 9 implementations of SolveRoot deleted!
- To make sure the template approach works with automatic type deduction, the implementation is in the header General.hh.  Because of this I extracted out a small class for the root finding algorithm input data to its own file.  No functional change here, and pulling the implementation into the header was acceptable during conversation with @lefticus.  
- In a few cases where the function was a member method, I was on the path of converting the std::bind into plain lambdas, but did not end up needing to handle every case once the template design was improved.

To Do:
- Evaluate build time and runtime impact
- Add a few static asserts in the template implementation that will help narrow down compiler errors should someone use SolveRoot wrong
- I think this should be merged soon with a follow-on effort that will:
  - Convert all the Par arrays from Array1D to std::array.  This can be enforced using additional static_asserts in the template implementation.
  - Convert all the std::bind into lambdas

### Pull Request Author

### Reviewer


scripts/dev/check_formatting.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "***Processing files in directory: ${dir}"
+  files=$(find "${dir}" | grep -E '\.((c|C)c?(pp|xx|\+\+)*$|(h|H)h?(pp|xx|\+\+)*$)')
+  for file in $files; do
+    if ! /usr/bin/clang-format-10 --dry-run --Werror --style=file "${file}";
+    then
+		  exit_code=1
+		fi
+  done
+done
+exit "$exit_code"
```
Myoldmopar, line 24: A new script for locally checking the clang-format of the source code.  This can be implemented as a pre-commit hook in your Git repo if you would like, but I am just trying to get into the habit of running this before pushing.

src/EnergyPlus/AirflowNetworkBalanceManager.cc
```diff
@@ -6854,15 +6853,15 @@ namespace AirflowNetworkBalanceManager {
...
-                    TempSolveRoot::SolveRoot(state,
-                                             ErrorToler,
-                                             MaxIte,
-                                             SolFla,
-                                             state.dataAirflowNetwork->ExhaustFanMassFlowRate,
-                                             AFNPressureResidual,
-                                             MinExhaustMassFlowrate,
-                                             MaxExhaustMassFlowrate,
-                                             Par);
+                    General::SolveRoot(state,
```
Myoldmopar, line 6856: No more TempSolveRoot!  Not a functional change here, just points to one single SolveRoot function in General namespace now.

src/EnergyPlus/BaseboardRadiator.cc
```diff
@@ -939,8 +935,7 @@ namespace BaseboardRadiator {
                         // pick an air  mass flow rate that is twice the water mass flow rate (CR8842)
                         baseboard->Baseboard(BaseboardNum).DesAirMassFlowRate = 2.0 * rho * baseboard->Baseboard(BaseboardNum).WaterVolFlowRateMax;
                         // pass along the coil number and the design load to the residual calculation
-                        Par(1) = DesCoilLoad;
-                        Par(2) = BaseboardNum;
+                        std::array<Real64, 2> Par2 = {DesCoilLoad, Real64(BaseboardNum)};
```
Myoldmopar, line 938: Creating a presized `std::array` now instead of an Array1D that is then allocated to a dynamic size.  This array gets passed into the SolveRoot function.  The SolveRoot function will ultimately then forward these along to the specified residual functions, so the residual functions must be modified to take the correctly matched size of `std::array`.  

However note that the final version of SolveRoot does not actually **require** the argument be a `std::array`.  I believe they should all be made `std::array` to allow the compile to do the best optimization, but it can be deferred until later to keep this PR tractable and review-able.

src/EnergyPlus/BaseboardRadiator.cc
```diff
@@ -1235,8 +1230,8 @@ namespace BaseboardRadiator {
     }
 
     Real64 HWBaseboardUAResidual(EnergyPlusData &state,
-                                 Real64 const UA,           // UA of coil
-                                 Array1D<Real64> const &Par // par(1) = design coil load [W]
+                                 Real64 const UA,                 // UA of coil
+                                 std::array<Real64, 2> const &Par // par(1) = design coil load [W]
```
Myoldmopar, line 1234: Because the call to SolveRoot for this function included a `std::array<Real64, 2>` argument, the parameter argument in the residual function here must also accept a `std::array<Real64, 2>`.  As such, the array access is changed from one-based parentheses to zero-based square brackets.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -331,6 +331,7 @@ set(SRC
     HVACSizingSimulationManager.hh
     HVACStandAloneERV.cc
     HVACStandAloneERV.hh
+    HVACSystemRootFindingAlgorithm.hh
```
Myoldmopar, line 334: A new small header only where I extracted a single class out of a larger file and put it into its own file.  No new functionality.

src/EnergyPlus/CMakeLists.txt
```diff
@@ -598,8 +599,6 @@ set(SRC
     TARCOGParams.hh
     TarcogShading.cc
     TarcogShading.hh
-    TempSolveRoot.cc
-    TempSolveRoot.hh
```
Myoldmopar, line 602: No new TempSolveRoot.  W00t!

src/EnergyPlus/DataHVACGlobals.hh
```diff
@@ -65,14 +65,6 @@ namespace DataHVACGlobals {
     // Data
     // -only module should be available to other modules and routines.
     // Thus, all variables in this module must be PUBLIC.
-    enum class HVACSystemRootSolverAlgorithm : int
```
Myoldmopar, line 68: Moved this to a standalone file to keep inclusion clean.

src/EnergyPlus/DesiccantDehumidifiers.hh
```diff
@@ -242,49 +242,49 @@ namespace DesiccantDehumidifiers {
...
+                                  std::string const &CompName, // name of the dehumidifier unit
+                                  bool FirstHVACIteration,     // TRUE if 1st HVAC simulation of system timestep
                                   int &CompIndex);
 
     void GetDesiccantDehumidifierInput(EnergyPlusData &state);
 
     void InitDesiccantDehumidifier(EnergyPlusData &state,
-                                   int const DesicDehumNum,      // number of the current dehumidifier being simulated
-                                   bool const FirstHVACIteration // TRUE if 1st HVAC simulation of system timestep
+                                   int DesicDehumNum,      // number of the current dehumidifier being simulated
```
Myoldmopar, line 252: There are a few places in here where I was just cleaning up Clang-tidy warnings such as `const` in header function declarations.

src/EnergyPlus/FuelCellElectricGenerator.cc
```diff
@@ -2667,70 +2666,6 @@ namespace FuelCellElectricGenerator {
         HProdGases = tempHprodGases;
     }
 
-    void FCDataStruct::FigureProductGasHeatCap(EnergyPlusData &state, Real64 const FluidTemp, Real64 &Cp)
```
Myoldmopar, line 2670: Function was not used at all.

src/EnergyPlus/FuelCellElectricGenerator.hh
```diff
@@ -574,27 +574,25 @@ namespace FuelCellElectricGenerator {
 
         void FigureAirEnthalpy(EnergyPlusData &state, Real64 FluidTemp, Real64 &Hair);
 
-        void FigureFuelHeatCap(EnergyPlusData &state, Real64 FluidTemp, Real64 &Cp);
+        void FigureFuelHeatCap(EnergyPlusData &state, Real64 FluidTemp, Real64 &Cp) const;
```
Myoldmopar, line 577: Minor - a few member methods marked as `const`.

src/EnergyPlus/General.hh
```diff
@@ -79,73 +80,174 @@ namespace WeatherManager {
...
                    Real64 Eps,   // required absolute accuracy
                    int MaxIte,   // maximum number of allowed iterations
                    int &Flag,    // integer storing exit status
                    Real64 &XRes, // value of x that solves f(x,Par) = 0
-                   std::function<Real64(Real64 const, std::vector<Real64> const &)> f,
-                   Real64 X_0,                    // 1st bound of interval that contains the solution
-                   Real64 X_1,                    // 2nd bound of interval that contains the solution
-                   std::vector<Real64> const &Par // array with additional parameters used for function evaluation
-    );
+                   const Function &f,
```
Myoldmopar, line 90: The function was previously a fixed `std::function` type, with a specific argument list.  This limited the possible parameters going into the function to an Array1D of floats.  Now the formulation of the function is actually a template parameter.  Right now we are still using just a list of floats, either the remaining `Array1D<Real64>` or the cleaned up versions that use `std::array<Real64, N>`.  In the long run, we need to make better use of `std::array`, and/or maybe even create some modified version that allows passing data in a better way into the SolveRoot residual function.  Right now we have to convert all variables into floats, including indexes and boolean values, then convert them back to be used inside the residual function.  There is still lots of room for better C++ practices that we can employ once this single version of SolveRoot is in place.

Myoldmopar:
Walk-through done, I believe CI will be all happy (it was locally), so I would just love some comments on it.  As I've said, the final version of this is pretty great because it looks just like the existing SolveRoot, but allows for better patterns to develop.

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -5690,17 +5690,17 @@ namespace CondenserLoopTowers {
 
         // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
         int SolFla;                                // Flag of solver
-        Array1D<Real64> Par(4);                    // Parameter array for regula falsi solver
+        std::array<Real64, 6> Par;                 // Parameter array for regula falsi solver
```
mjwitte, line 5693: Did you want 4 here?

src/EnergyPlus/CondenserLoopTowers.cc
```diff
@@ -6103,14 +6103,14 @@ namespace CondenserLoopTowers {
...
+        Real64 const OutWaterTemp = this->calculateSimpleTowerOutletTemp(state, Par[2], Par[3], UA);
+        Real64 const CoolingOutput = Par[4] * Par[2] * (this->WaterTemp - OutWaterTemp); // tower cooling output [W]
+        return (Par[0] - CoolingOutput) / Par[0];
     }
 
     Real64 CoolingTower::residualTr(EnergyPlusData &state,
-                                    Real64 Trange,             // cooling tower range temperature [C]
-                                    Array1D<Real64> const &Par // par(1) = tower number
+                                    Real64 Trange,                   // cooling tower range temperature [C]
+                                    std::array<Real64, 6> const &Par // par(1) = tower number
```
mjwitte, line 6113: 4 here? Does it matter if it's a little larger and some elements are unused?

src/EnergyPlus/DXCoils.cc
```diff
@@ -17585,11 +17584,7 @@ void ControlVRFIUCoil(EnergyPlusData &state,
         if (QCoilSenCoolingLoad > QinSenMin1) {
             // Increase fan speed to meet room sensible load; SH is not updated
 
-            Par(1) = QCoilSenCoolingLoad;
-            Par(2) = Ts_1;
-            Par(3) = Tin;
-            Par(4) = Garate;
-            Par(5) = BF;
+            Par = {QCoilSenCoolingLoad, Ts_1, Tin, Garate, BF};
```
mjwitte, line 17587: Getting tricky here with the initialization. It works either way.

src/EnergyPlus/FuelCellElectricGenerator.hh
mjwitte: So, exactly what does this do (adding `const` to a function)?

src/EnergyPlus/General.cc
```diff
@@ -91,945 +91,6 @@ using DataHVACGlobals::Bisection;
...
-
-void SolveRoot(Real64 const Eps, // required absolute accuracy
-               int const MaxIte, // maximum number of allowed iterations
-               int &Flag,        // integer storing exit status
-               Real64 &XRes,     // value of x that solves f(x,Par) = 0
-               std::function<Real64(Real64 const, Array1D<Real64> const &)> f,
-               Real64 const X_0,           // 1st bound of interval that contains the solution
-               Real64 const X_1,           // 2nd bound of interval that contains the solution
-               Array1D<Real64> const &Par, // array with additional parameters used for function evaluation
-               int const AlgorithmTypeNum, // ALgorithm selection
```
mjwitte, line 604: For places that were passing in `AlgorithmTypeNum` and that flavor of `SolveRoot` only supported bisection and regula falsi), do those components need some documentation changes?

src/EnergyPlus/UnitarySystem.cc
```diff
@@ -12452,17 +12450,21 @@ namespace UnitarySystems {
...
-                        // or fatal out in GetInput
-                        Par[3] = 1.0; // DehumidMode
-                        Par[4] = double(FanOpMode);
-                        Par[5] = this->m_CoolingSpeedNum;
-                        Par[6] = 1.0; //  this->m_CoolingSpeedRatio;
-                        Par[7] = 0.0;
-                        General::SolveRoot(state, Acc, MaxIte, SolFla, PartLoadFrac, &this->genericDXCoilResidual, 0.0, 1.0, Par);
+                        std::array<Real64, 8> Par2 = {
+                            double(this->m_CoolingCoilIndex),
+                            DesOutTemp,
```
mjwitte, line 12455: Is this assignment format done for performance reasons? It makes it a little harder to read if one wants to compare the `Par` elements in the residual function back to the values/variables passed in here.

mjwitte (commented):
So - much - nice - er! Some questions:

src/EnergyPlus/UnitarySystem.cc
Myoldmopar: As I said above, I'm fine using individual assignments.  When the assignment is a clear set of values, the initializer list is, practically speaking, just a nicer look.

src/EnergyPlus/General.cc
Myoldmopar: If I recall correctly, the only place I found it being used like that was actually _inside_ a unit test.  I could pull develop, delete that one overload and try to build to verify unless you happen to know that there are definitely components using it.

src/EnergyPlus/DXCoils.cc
Myoldmopar: Yep, works either way.  I _think_ that the initializer list is going to be like nanoscopically faster but it's nothing worth worrying about.  In places where they are clear assignments, this is a nice look.  In places where you are doing logic on each assignment then I would definitely leave them piecemeal.

src/EnergyPlus/CondenserLoopTowers.cc
Myoldmopar: I don't think it would matter to any noticeable amount, but definitely better to be precise with the amount of data we need in the payload.

src/EnergyPlus/CondenserLoopTowers.cc
Myoldmopar: Probably just a little too quick with the paste, I'll make it back to 4.

src/EnergyPlus/FuelCellElectricGenerator.hh
Myoldmopar: For a class member method, adding the const attribute tells the compiler that this function will access member variables on the class instance, but does not modify them.  It is basically making the `this->` instance "read-only".  And so therefore it can only call other methods on the class if they themselves are also `const` methods.  Great question!

And I'm not sure if it impacts performance at all.  Surely not in a significant way in the context of E+.  Just helps make sure the intent of the function is known to users of it.

src/EnergyPlus/General.cc
mjwitte: No, not necessary.

mjwitte (approved):

src/EnergyPlus/FuelCellElectricGenerator.hh
lefticus: It almost certainly will not affect performance, but besides documentation to let users know that the function will not modify the object, it also enables the function to be used on a const object

```cpp
const SomeThing st;
// can only call `const` member functions on st.
```

amirroth:
General question using `FigureFuelHeatCap` as a mule. Why does this function return a value via a reference parameter as opposed to, I don't know, the return value?

lefticus:
> 
> 
> General question using `FigureFuelHeatCap` as a mule. Why does this function return a value via a reference parameter as opposed to, I don't know, the return value?

I've also wondered that kind of thing throughout EnergyPlus as I've been working in the code base. Return by value (particularly for something like a `double`) is likely to result in more efficient function calls.

lefticus:
@Myoldmopar I went for a more general approach for a static_assert, instead of the ones that check for `begin()` and `end()` as you had.

I can see no intrinsic reason to require that it's a vector/array like thing that is the Payload passed into the Function. With it templated as it is now, it can be pretty much anything, which might open up other use cases. The only requirement is that it can be passed to the Function as the 3rd parameter.

jasondegraw:
@amirroth @lefticus Hazarding a guess here, but subroutines and functions in FORTRAN are separate things, and subroutines don't return values.

amirroth:
It’s also less readable at the call site.

Myoldmopar:
> @Myoldmopar I went for a more general approach for a static_assert, instead of the ones that check for `begin()` and `end()` as you had.

Yeah that makes good sense.  I like where this is at now -- the compiler should give a really nice readable message if this is used incorrectly.

As for the Windows build, it looks like it was unrelated.  I'm going to kick it to try a new build and we'll see.

Myoldmopar:
@lefticus so it looks like Windows is getting some compiler warnings on those static assertions.  Do you think they should be muted or should we make a change to address them?

lefticus:
> 
> 
> @lefticus so it looks like Windows is getting some compiler warnings on those static assertions. Do you think they should be muted or should we make a change to address them?

I'll take a look and see what reasonable options there are.

mitchute (approved):
Fantastic work here, @Myoldmopar! Merging.

PR #8765, opened 2021-05-11
Remove hack that limits simulate() when InitLoopEquip is true, Plant-Component

jmythms:
This PR removes a [hack](https://github.com/NREL/EnergyPlus/pull/8765/files#diff-919b9010a53636f05fbf93e731ae1d428eaa46311def39e8af6edadf1a855964L127-L140) that was placed to ensure no diffs when the Plant code was refactored. 

[A line that checks for](https://github.com/NREL/EnergyPlus/pull/8765/files#diff-919b9010a53636f05fbf93e731ae1d428eaa46311def39e8af6edadf1a855964L117) `compPtr` was also removed, and the errors caused by its removal were fixed by adding a [dummy plant component](https://github.com/NREL/EnergyPlus/pull/8765/files#diff-cad31c16b3d6de86a812724335842941e2a1aa7727348ae5b2e7d9f32df33588R110-R122) object for the plant components that were not being simulated earlier. 

Details:
When (InitLoopEquip), this hack was a check to call the simulate() function only if the plant component being simulated belonged to:
1. DataPlant::TypeOf_Pipe, 
2. DataPlant::TypeOf_PipeSteam, 
3. DataPlant::TypeOf_SolarCollectorICS, 
4. DataPlant::TypeOf_SolarCollectorFlatPlate
([link in code](https://github.com/NREL/EnergyPlus/pull/8765/files#diff-3c5684b1a060d0f8e5c5414ac06bf78b62251b7ca80a7f39a8c0eb47f501733dL230-L231))

Removing the hack will call the simulate() function for the other components as well. 

The Table big diffs are caused by re-ordering of outputs in the tables. 

### Pull Request Author

### Reviewer


src/EnergyPlus/Plant/PlantManager.cc
```diff
@@ -974,15 +979,19 @@ void GetPlantInput(EnergyPlusData &state)
                     } else if (UtilityRoutines::SameString(this_comp_type, "Coil:Cooling:Water")) {
                         this_comp.TypeOf_Num = TypeOf_CoilWaterCooling;
                         this_comp.CurOpSchemeType = DemandOpSchemeType;
+                        this_comp.compPtr = EmptyPlantComponent::factory();
```
mitchute: I assume that there's a place in the code where maybe,

```C++
if (this_comp.compPtr == nullptr) ShowFatalError(...);
```

and that's the reason for the EmptyPlantComponent rather than just setting it to ```nullptr```?

mitchute (commented):
One minor question regarding when we should be checking for nullptr's, but this is a concise set of changes, and overall looks great.

amirroth:
It will look even greater when CompType becomes an `enum class`, the names become `constexpr std::array<std::string_view>` indexed by that enumeration and the entire `if-else` ladder collapses to three lines of code (except for pumps maybe). But yes.

src/EnergyPlus/Plant/PlantManager.cc
jmythms: Thanks @mitchute . I missed this, thank you for noticing, adding it in the next push.

Myoldmopar: Yeah, in my mind, we actually had every single component type that could live on a plant loop already setting this compPtr.  But there are some coils and things that are fully air-side, and don't yet implement the PlantComponent base class interface, because they don't need to interact with the plant solver (at least not yet).  So I just had @jmythms create a "no-op" plant component for the remaining components.  I think it would be good to add a check somewhere around [here](https://github.com/NREL/EnergyPlus/pull/8765/files#diff-b614b9cbdea8251f873e0f96030fd410b5c3f16d5a060ee59d9bb3331fb1b3a6R1389) @jmythms before this is called completely done.  That will ensure that if new component types are added, they will definitely get a pointer set on there.

And then we can get rid of this "passive" check on nullptr for every single plant component ever simulated, and the no-op plant components won't do anything.  I realize there may be a very small penalty for calling the no-op function, but I think it will be a net gain as most of the plant components will now simulate straight through without the interruption of the conditional check.  It will certainly make the simulate method on the topology component simpler anyway, even if the gains aren't significant.

jmythms: @Myoldmopar Thank you for explaining the performance side of this. 

You had mentioned adding the check earlier, I tried it in this [commit](https://github.com/NREL/EnergyPlus/pull/8765/commits/747edc8f9e018ca8cfc11df3f9caf538bae741e2#), but for some reason, my build kept going past the Fatal Error and crashing. So I used a std::cout to tell me which were the components not simulating, then cleaned it up. Then forgot to re-add the check back in.

jmythms: Actually I see the problem in that commit, it was because I used a SevereError(I used this because I thought I could diagnose multiple components at once) instead of the FatalError. And also understand why we need both SevereError and FatalError :) thanks!

src/EnergyPlus/Plant/PlantManager.cc
```diff
@@ -1351,16 +1379,20 @@ void GetPlantInput(EnergyPlusData &state)
...
                     } else {
                         // discover unsupported equipment on branches.
                         ShowSevereError(state, "GetPlantInput: Branch=\"" + BranchNames(BranchNum) + "\", invalid component on branch.");
                         ShowContinueError(state, "...invalid component type=\"" + this_comp_type + "\", name=\"" + CompNames(CompNum) + "\".");
                         //            ErrorsFound=.TRUE.
                     }
 
+                    if (!this_comp.compPtr) ShowFatalError(state, format(" Plant component \"{}\" was not assigned a pointer.", this_comp_type));
```
jmythms, line 1394: FatalError added here. This will help remind the next person who adds a component to assign the compPtr as well.

mitchute: Thanks. When I commented before I assumed that this check already existed somewhere, but I see now that it didn't. I'm good with adding this here.

mitchute:
I don't see any issues with the code, but we do still have some diffs here that go beyond just reordering of output files. Is that expected, or is that why this is still a draft?

src/EnergyPlus/Plant/PlantManager.hh
```diff
@@ -107,6 +107,27 @@ namespace PlantManager {
...
+                      [[maybe_unused]] bool RunFlag) override
+        {
+            // this is empty on purpose
+        }
+
+        static EmptyPlantComponent *factory()
+        {
+            // returns a newly allocated dummy plant component, again this is only used for plant topology components
+            // that do not need any plant side calcs done
+            return new EmptyPlantComponent;
```
jmythms: I'm trying to figure out if there is a better way to get around the use of this `new` keyword.

jmythms:
Explaining the numerical diffs in regression tests:

There are only small diffs, which means differences in results were less than the threshold limit of 0.001(units).

 A few graphs are attached to show how these differences are insignificant when compared to actual values. The top plot shows how the difference changes with date/time values for each figure, the bottom shows the actual simulation results for this branch vs develop branch. 

The y-axis limits for the top and bottom plots in each figure show the insignificance of the differences. 

(In the plots, the faint vertical red line shows the point of maximum absolute difference, if not evident from the graph.)

5ZoneDetailedIceStorage

ASHRAE9012016_ApartmentHighRise_Denver
EMSDemandManager_LargeOffice
GasTurbChillerHeatRecoveryAuto
HeatPumpWaterHeater
HospitalBaseline



ShopWithPVandBattery

mitchute:
1e-12 diffs... unacceptable... 😉 

Looks like this is ready. Thanks for the plots, @jmythms! Go ahead and merge in develop one more time, then we can get this merged.

Myoldmopar:
Awesome!!

Myoldmopar (approved):

mitchute:
Diffs as expected. Thanks, @jmythms! Merging.

jmythms:
Thank you! :)

PR #8769, opened 2021-05-13
Move EnergyPlus to Next Version

Myoldmopar:
Pull request overview
---------------------
We've been talking about getting away from the incremental tenth release numbering scheme for a while.  The only hesitation I have is that some tools (including some of our own) are parsing the version number from the install path or something else with the assumption that it is N_X_Y or N.X.Y.  Where N is a single digit.  But that shouldn't hold us up.  We should just bite the bullet and move to a major version increment pattern now, and do it early in this release cycle so that we have time to catch any issues.  

For those wondering why the change, well, since our "API" has traditionally been the input file, and we change the input file structure every time we make a release, it would've been better just to do major version number changes each release.  With the move to epJSON happening sometime, we will actually have the possibility of doing some (maybe rare) releases where the input file from an older version still works in the next release.  In this case we would have the option to start doing minor version releases where we don't break the "API".  I'm not sure if that will actually happen or not, but it doesn't matter, either way we should be doing major version changes.

Beyond the number being two digits now, this is a standard version change PR, with tons of IDF changes and accompanying transition and cmake updates.  For CI, the final clean version will have table diffs and EIO diffs, maybe IDF diffs, but no ERR diffs or any functional change.


nealkruis (approved):
@Myoldmopar so, we're adopting [Semantic Versioning](https://semver.org/)?

Myoldmopar:
> @Myoldmopar so, we're adopting [Semantic Versioning](https://semver.org/)?

Not exactly yet, but I think moving to major version changes will definitely push us in the right direction.  And maybe we can practically make that jump once we are purely taking JSON inputs.

jasondegraw (approved):
I think I would probably not live up to expectations if I didn't say (one last time) that we should call the next version 21.10 and transition to Ubuntu-style version numbering. We release on a set schedule, make a lot of substantial changes every release, and I think that getting caught up in what is breaking vs. what is not breaking is just not all that constructive.

mitchute (approved):

nealkruis:
I don't feel strongly, but @jasondegraw makes some good points. His approach is consistent with [how CLion does it's versions](https://www.jetbrains.com/clion/whatsnew/) too. I think it's worth considering for future releases.

amirroth:
EnergyPlus MMXXI.X

JasonGlazer:
EnergyPlus NT, EnergyPlus 2000, EnergyPlus XP, EnergyPlus Vista, EnergyPlus 10 

:-)

What ever we do, lets not change the scheme for at least another decade. It seems like the easiest to explain and understand is just to continue what we have been doing: 9.6.0, 9.7.0, 9.8.0, 9.9.0, 10.0.0, 10.1.0, 10.2.0, etc. From the users perspective, nothing is changed with that and we just update the utilities as needed.

But I would also be fine with @jasondegraw Ubuntu style proposal also. I think that approach has the good side effect of encouraging more frequent updates by users because they are frequently reminded how old their software is.

amirroth:
What about MacOS-style animal or California locations naming scheme? Except we would use names associated with the EnergyPlus development team. EnergyPlus Golden. EnergyPlus Knoxville. EnergyPlus Arlington Heights. EnergyPlus Cashion.

Myoldmopar:
It's surprising that 10.0, 11.0, 12.0, 13.0 is the least popular option for naming a piece of software. :shrug: FWIW, my main intent here is to get us away from the arbitrary tenth increment without disrupting any workflows out there.  We were actually ready to go to 10.0 right after 9.0 came out for this very reason, but I was afraid that the double digit version could break some of our own utilities that parse out version.  

Unless we get this resolved before then we can discuss it on the technicalities call.  And if it ends up being 9.6, or 2021.09, or 2021.2?, or 10.0, it's not going to make a big difference to me.

jasondegraw:
How about EnergyPlus Home Premium SP1? The next one can be SP2. I agree with @JasonGlazer that going to YYYY.MM (or YY.MM) would help with awareness on the age of versions, that was not something I had thought of. If I had, I would have added that to my list. I'd also rather get to something less arbitrary sooner rather than later, and going this route seems like the quicker way (but I'm also fine if we continue with the current numbering). 

Are we following parliamentary procedure? I move we adopt a year+month version numbering scheme.

mjwitte:
I tested this by building an installer (on Windows) and tried to transition a v9.5.0 idf to v10.0.0 using :
EP-Launch (File->Transition Version). It failed "Transition utility file not found." 
Then I tried using IDFVersionUpdater and it showed 9.5.0 as the latest version.

@JasonGlazer What are EP-Launch and IDFVersionUpdater looking for here? Do they need code changes to accommodate a change in version number format?

@Myoldmopar I have no opinion on version numbering schemes. As long as the installers and transition utilities work, I'm happy.

JasonGlazer:
I believe that EP-Launch 2 and IDFVersionUpdater would probably need to have some code changes just to accept a version number in the form of NN.N.N. Not major changes but they both make assumptions on versions being in the form N.N or N.N.N To get a sense of the age of this code, it shows example version numbers like "3.0.0". I think EP-Launch 3 would not need any code updates for NN.N.N

If the version numbering were to change to YYYY.MM or YY.MM, I would guess that all three programs would need some code changes. 

That said, I would expect the amount of code changes is relatively minor no matter which way we go.

I did not check Transition but would not be surprised if it looks at version numbers at all, that code probably would need an update.

Myoldmopar:
OK, I think I've got everything pushed back to target 9.6.  I'll watch to see the CI results in case I missed a test file version or something, but locally regressions were good.  If it's all clean it could merge.

mitchute:
Were we adding a note to the README about the upcoming versioning scheme change, or is that just with 9.6's release notes?

mjwitte (approved):
Built a windows installer and tested transition both from EP-Launch (classic) and IDFVersionUpdater. All good.

Myoldmopar:
@mjwitte thanks for testing it out!

@mitchute yeah good call.  I added a preliminary paragraph that will suffice for this branch and we can refine it closer to release time when I overhaul the readme.

CI looks very happy here, the big thing is no ERR diffs popping up.  At this point the `scripts/dev/change_version.py` script is really working well to automatically update the version number in the IDFs and IMFs.  I need to modify it to edit the epJSON test files as I have to do that manually still.  

Anyway, this looks fully ready, merge at will.

mitchute:
Diffs are as expected here. Merging. Thanks, @Myoldmopar!

PR #8771, opened 2021-05-17
8472 Missing a Step in the CTF Calculation

### LINKED ISSUE #8472: Missing Step in Conduction Transfer Function Calculation
Issue overview
--------------
During a review of the EnergyPlus code that uses the state space method to calculate the conduction transfer functions, a user noticed that one of the steps did not match the relevant procedure as outlined in the doctoral dissertation of John Seem upon which the procedure is based.  In Seem's dissertation, on page 128, there is a four step procedure for calculating the exponential matrix.  In Step 4, the term "L" is calculated based not on the original value of the A(delta) matrix but rather the modified version of the A(delta) matrix after it has been divided by 2^k.  In the EnergyPlus code, L is based on the on the original value of the A(delta) matrix.  This is incorrect according to Seem's dissertation and should be fixed.

The user has suggested added another section of code at line 1202 of Construction.cc (click here to get there: https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/Construction.cc#L1202 ) that repeats the code in lines 1166-1179 which calculates AMatRowNormMax.  It may be possible (though this should be checked as part of the fix) to simply adjust AMatRowNormMax by the same factor as was used to update AMat1 with a single assignment statement.  Thorough testing of this will be needed to make sure that nothing strange happens, even though it does appear that the code is wrong.  It is not anticipated that this fix will have large impacts on the results and some preliminary testing by @mitchute backs that up.

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): V9.4 and all previous versions

### Checklist
 - [none needed] Defect file added (list location of defect file here)

RKStrand:
Pull request overview
---------------------
 - Fixes #8472 
 - DESCRIBE PURPOSE OF THIS PULL REQUEST:
When looking through the code and comparing the state space method for calculating conduction transfer functions (CTFs) to the document that the code was based on (John Seem's doctoral dissertation, UW-Madison), a user noticed that a step was missing.  While it is unclear whether this was intentional or not, what is clear is that adding this step in will actually not make EnergyPlus more accurate.  Instead, it will actually have the potential to make it less accurate and potentially cause instabilities.  The missing step affects the calculation of the number of terms that must be calculated for an exponential matrix that is part of the CTF determination.  The missing step would actually reduce the number of terms and it isn't 100% clear that this missing step is necessary or that it is correct given the equations in this part of the code.  So, it was decided to leave the code as is (making this a documentation only bug fix--no new code, no unit tests) except for the addition of comments to clarify the difference between EnergyPlus and Seem's dissertation.  In addition, more documentation was added to the Engineering Reference to provide more detail about this portion of the calculation.

### Pull Request Author

### Reviewer


doc/engineering-reference/src/surface-heat-balance-manager-processes/conduction-through-the-walls.tex
```diff
@@ -61,7 +61,27 @@ \subsection{Calculation of Conduction Transfer Functions}\label{calculation-of-c
...
 
-This formulation can be used to solve the transient heat conduction equation by enforcing a finite difference grid over the various layers in the building element being analyzed.~ In this case, the state variables are the nodal temperatures, the environmental temperatures (interior and exterior) are the inputs, and the resulting heat fluxes at both surfaces are the outputs.~ Thus, the state space representation with finite difference variables would take the following form:
+Seem (1987) summarizes the steps required to obtain the output vector (y) from the A, B, C, and D matrices.~ Briefly, the solution for the first order differential system of equations with constant coefficients shown above is:
+
+\begin{equation}
+x_{t + \delta} = e^{A \delta} x_t + \int\limits_t^{t + \delta } {e^{A (t + \delta - \tau)} B u(\tau) d \tau}
+\end{equation}
+
+where t is a particular point in time, \(\delta}\) is the time step, and \(e^{A \delta}\) is the exponential matrix which is calculated using a power series:
```
mitchute: Looks like there's a stray curly brace here ```\(\delta}\)```

mitchute (approved):
Thanks, @RKStrand! This should give the next interested reader (in another 20 years when E+ turns 40!) some additional details. There was one stray curly brace that I went ahead and fixed, but assuming docs build, this is good from my perspective.

mitchute:
@mjwitte do you want to take a look here before this goes in?

src/EnergyPlus/Construction.cc
```diff
@@ -1197,7 +1197,25 @@ void ConstructionProps::calculateExponentialMatrix()
...
+    // l, the integer value of CheckVal, is used to define the number of
+    // terms in the exponential matrix of AMat that are required for an
+    // accurate estimation of that matrix.  In practice, additional terms
+    // probably won't have a large effect but in general should make the
+    // calculation MORE accurate.  Also, if the new terms are too small,
+    // then as noted above the cut-off criteria will not use them.
+    // Given that there is no harm in making l (integer of CheckVal) larger
+    // but it could result in a more accurate solution, it was decided to
+    // keep the value of AMatRowNormMax as it has been traditionally done
+    // in EnergyPlus since it could lead to higher accuracy.
```
mjwitte: My preference would be to just state the facts:



// exponentiation algorithm.  One thing to note is that the EnergyPlus    // code is slightly different here than what is presented in Seem's    // dissertation.  In Seem's dissertation, AMatRowNormMax (delta, the    // timestep, is already factored into this term above) is divided by    // 2^k with k defined in the code above.  ~This should have been done    // here to follow Seem's algorithm exactly.  However, in EnergyPlus,    // this was not done and there is some uncertainty as to why it wasn't.~    // Dividing AMatRowNormMax by 2^k would have the net effect of decreasing    // the value of AMatRowNormMax and thus potentially CheckVal as well.    // l, the integer value of CheckVal, is used to define the number of    // terms in the exponential matrix of AMat that are required for an    // accurate estimation of that matrix.  In practice, additional terms    // ~probably~ won't have a large effect but in general should make the    // calculation more accurate.  Also, if the new terms are too small,    // then as noted above the cut-off criteria will not use them.    ~// Given that there is no harm in making l (integer of CheckVal) larger    // but it could result in a more accurate solution, it was decided to    // keep the value of AMatRowNormMax as it has been traditionally done    // in EnergyPlus since it could lead to higher accuracy.~

doc/engineering-reference/src/surface-heat-balance-manager-processes/conduction-through-the-walls.tex
RKStrand: Yup--that's a stray.  Not sure how I missed that, but it doesn't surprise me that I accidentally had something like that in there.  Thanks for catching it and fixing it!

doc/engineering-reference/src/surface-heat-balance-manager-processes/conduction-through-the-walls.tex
```diff
@@ -61,7 +61,27 @@ \subsection{Calculation of Conduction Transfer Functions}\label{calculation-of-c
...
...quaring" approach to calculate the exponential matrix.  This scaling would also affect L meaning that if the matrix row norm has been scaled so that it is never greater than 1 then L would never go higher than 9.  This seems somewhat unsual (especially given the equation for L has a limit of 100) and when Seem's approach was first implemented in EnergyPlus, the matrix row norm was NOT scaled when it was used to calculate L.  So, while the actual implementation in EnergyPlus deviates slightly from what was recommended by Seem, this approach can perhaps be seen to reflect the improvements in computer technology since that work was published or simply that there appears to be a potential incongruency in the method that Seem suggested for that one detail.  In addition, to avoid excessive terms in the exponential matrix, the terms that are added in the exponential matrix calculation in EnergyPlus are limited so that if a new term is smaller than will impact the variable based on the precision of the variable it is then ignored.  To see the implementation of these details in EnergyPlus, consult both Appendix A of Seem's dissertation and the routine calculateExponentialMatrix in the code.
```
mjwitte: A suggested edit:
~This seems somewhat unsual (especially given the equation for L has a limit of 100) and when Seem's approach was first implemented~ In EnergyPlus, the matrix row norm ~was~ *is* not scaled when ~it was~ used to calculate L.  ~So, while the actual implementation in EnergyPlus deviates slightly from what was recommended by Seem, this approach can perhaps be seen to reflect the improvements in computer technology since that work was published or simply that there appears to be a potential incongruency in the method that Seem suggested for that one detail.~ *This can result in additional terms which in general will make the calculation slightly more accurate.* ~In addition,~ To avoid excessive terms in the exponential matrix,  . . .

src/EnergyPlus/Construction.cc
RKStrand: Sounds good, @mjwitte.  It is better to just stick with the facts.  I can save the editorializing for the incredibly exciting paper that I will publish on this topic (when I have nothing better to do 🤣).  I'll make this change (and the other) shortly.

doc/engineering-reference/src/surface-heat-balance-manager-processes/conduction-through-the-walls.tex
RKStrand: I'll make these changes as requested also.  Thanks for reviewing all this, @mjwitte and @mitchute!

mitchute:
Looks like GH Actions is having problems: https://www.githubstatus.com/incidents/m16jzl31gnqt

I built the Eng Ref locally, and it looks good. Merge at will @mjwitte.

mitchute:
GH Actions is back online and the docs built A-OK. Thanks, @RKStrand! Merging.

PR #8773, opened 2021-05-19
Fix one source that gives the leap year simulation day missing problem.

### LINKED ISSUE #8772: Missing simulation days when starting with a leap year and end with the next year
Issue overview
--------------
It was reported that there are missing simulation days when dealing with leap starting year. For a full year simulation (expecting 365 days), one to two days of simulations were found missing at the end of the running period. 

After some experiment runs, one common problematic scenario is when the simulation start year is a leap year, but the end year is next year. In this case, the number of total simulation days would get a 364 instead of 365 if a full year is specified—for example, running from Oct 1, 2020 to Sep 30, 2021 will only get simulation runs (and reports) s from Oct 1, 2020 to Sep 29, 2021, with the last day (Sep 30, 2021) missing. 

Actually this would happen to any ordinary idf file---just ordinary year and no special specification for "leap year weather" or  "actual weather treatment". See the attached idf file modified from the example file 1ZoneUncontrolled_DD2009.idf. The problem will occur when the simulation period from Oct 1, 2016 to Sep 30, 2021 in any file with default setting with typical default (blank) and typical weather file (default Atlantic weather with `HOLIDAYS/DAYLIGHT SAVINGS,No,0,0,0`--so it is just a case for ordinary year and not for leap year at all, but the problem happens and the simulation only run from 2016-10-01 to 2017-09-29.

The root cause is the processing of the start Julian day inputs have some missing logic in handling the case of having a starting leap year and then a different ending year. 

[1ZoneUncontrolled_DD2009_rev1.idf.txt](https://github.com/NREL/EnergyPlus/files/6509220/1ZoneUncontrolled_DD2009_rev1.idf.txt)

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): v9.4 v9.5
 - Refer to Helpdesk ticket number 15720

### Checklist

jcyuan2020:
Pull request overview
---------------------
 - Tries to fix #8772
 - The very general case was fixed. Some further fix might needed and tested for some other corner cases and combinations with the weather file settings.

### Pull Request Author

### Reviewer

### DIFF
--- a/src/EnergyPlus/WeatherManager.cc
+++ b/src/EnergyPlus/WeatherManager.cc
@@ -9198,7 +9198,8 @@ namespace WeatherManager {
                 } else {
                     // Environment crosses year boundaries
                     env.RollDayTypeOnRepeat = runPer.RollDayTypeOnRepeat;
-                    env.StartJDay = General::OrdinalDay(runPer.startMonth, runPer.startDay, runPer.isLeapYear ? 1 : 0);
+                    env.StartJDay = General::OrdinalDay(
+                        runPer.startMonth, runPer.startDay, isLeapYear(env.StartYear) && state.dataWeatherManager->WFAllowsLeapYears ? 1 : 0);
                     env.EndJDay = General::OrdinalDay(
                         runPer.endMonth, runPer.endDay, isLeapYear(runPer.endYear) && state.dataWeatherManager->WFAllowsLeapYears ? 1 : 0);
                     env.TotalDays = 366 - env.StartJDay + env.EndJDay + 365 * std::max(env.NumSimYears - 2, 0);


src/EnergyPlus/WeatherManager.cc
```diff
@@ -9117,7 +9117,8 @@ namespace WeatherManager {
                 } else {
                     // Environment crosses year boundaries
                     env.RollDayTypeOnRepeat = runPer.RollDayTypeOnRepeat;
-                    env.StartJDay = General::OrdinalDay(runPer.startMonth, runPer.startDay, runPer.isLeapYear ? 1 : 0);
+                    env.StartJDay = General::OrdinalDay(
+                        runPer.startMonth, runPer.startDay, isLeapYear(env.StartYear) && state.dataWeatherManager->WFAllowsLeapYears ? 1 : 0);
```
jasondegraw, line 9202: This change means that `env.IsLeapYear`/`runPer.isLeapYear` aren't definitive on whether the first year of the environment is a leap year or not, so this is not going to work by itself. The reported issue needs more information before it's at all clear what the fix is.

Myoldmopar:
Because of the potential impact of this, should we try to get it in early for IO Freeze, or punt until after 9.6?  I don't want something potentially volatile to squeeze into bug fix release time.

jcyuan2020:
The complete fix could be more complex. The major function related to this, is functioning fine if the user take a little bit caution. So I think this might not be that urgent and could be left to future development after v9.6. 

BTW, the user also reported another corner case abnormality issue, which I will post so someone could also take a look in the future if needed. Again, the major functionality related to this leap year is functioning fine if the users take some cautions to avoid going to these rare corner case `traps`.

Myoldmopar:
Excellent, let's push this to after 9.6 then.  Thanks for the full update!

mjwitte:
Just a one-liner here, so closing this PR. Will open a new on when work resumes, post v9.6.

PR #8781, opened 2021-05-21
Use gross roof area for ave ceiling height

### LINKED ISSUE #8770: Use gross ceiling surface area to calculate average ceiling height
Issue overview
--------------
In the surface geometry, the ceiling height of a zone is calculated using the averaged ceiling height for all the ceiling (roof surfaces. For the case of multiple surfaces, a weight to each ceiling surface is given by the fraction of the individual ceiling surface area among the summed area of all the ceiling surfaces. 

However, the area used for this weighting is the “net” surface area—rather than the gross one. In the case of a roof surface contains a window (a skylight, for example), the corresponding weight of that surface is then would be reduced. This will make the calculated results to be skewed toward the roof that has no windows or less window areas. 

Since the “area” of the ceiling in this context is just a geometric concept that does need to consider how much of window surface area it contains, it is suggested that the ceiling gross areas rather than the (net) area should be to be used for this calculation. 

### Details
 - Version of EnergyPlus (if using an intermediate build, include SHA): v9.5 and before, probably all
 - Related Helpdesk ticket: 15707 (related but not exactly the same issue) 

### Checklist

jcyuan2020:
Pull request overview
---------------------
 - Fixes #8770 
 - 

### Pull Request Author

### Reviewer


mitchute:
@jcyuan2020 can you pull in develop one more time so I can do a final review and merge?

jcyuan2020:
> 
> 
> @jcyuan2020 can you pull in develop one more time so I can do a final review and merge?
@mitchute Thanks! The branch is now up-to-date.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -7021,3 +7021,532 @@ TEST_F(EnergyPlusFixture, GetSurfaceData_SurfaceOrder)
...
+        "    4,                       !- Number of Vertices",
+        "    2, 4, 4,                         !- X,Y,Z  1 {m}",
+        "    2, 0, 4,                         !- X,Y,Z  2 {m}",
+        "    6, 0, 4,                         !- X,Y,Z  3 {m}",
+        "    6, 4, 4;                         !- X,Y,Z  4 {m}",
+
+        "BuildingSurface:Detailed,",
+        "06Roof_E,                !- Name",
+        "Roof,                    !- Surface Type",
+        "R13WALL,                 !- Construction Name",
```
mitchute, line 7457: Is this a mistake? Surface name "06Roof_E" and Construction Name "R13Wall"?

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -7021,3 +7021,532 @@ TEST_F(EnergyPlusFixture, GetSurfaceData_SurfaceOrder)
...
+        "4,                       !- Number of Vertices",
+        "6, 0, 4,                         !- X,Y,Z  1 {m}",
+        "8, 0, 2.5,                          !- X,Y,Z  2 {m}",
+        "8, 4, 2.5,                          !- X,Y,Z  3 {m}",
+        "6, 4, 4;                         !- X,Y,Z  4 {m}",
+
+        "BuildingSurface:Detailed,",
+        "07Roof_W,                !- Name",
+        "Roof,                    !- Surface Type",
+        "R13WALL,                 !- Construction Name",
```
mitchute, line 7473: Same here.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
```diff
@@ -7021,3 +7021,532 @@ TEST_F(EnergyPlusFixture, GetSurfaceData_SurfaceOrder)
...
+    Real64 totalGrossCeilingArea = 0.0;
+    totalGrossCeilingArea =
+        state->dataSurface->Surface(6).GrossArea + state->dataSurface->Surface(7).GrossArea + state->dataSurface->Surface(8).GrossArea;
+
+    Real64 totalNetCeilingArea = 0.0;
+    totalNetCeilingArea = state->dataSurface->Surface(6).Area + state->dataSurface->Surface(7).Area + state->dataSurface->Surface(8).Area;
+
+    Real64 ceilingHeight_expected = 0.0;
+    ceilingHeight_expected = 3.25 * (state->dataSurface->Surface(7).GrossArea + state->dataSurface->Surface(8).GrossArea) / totalGrossCeilingArea +
+                             4.0 * state->dataSurface->Surface(6).GrossArea / totalGrossCeilingArea;
```
mitchute, line 7544: What do the "3.25" and "4.0" signify here?

mitchute (commented):
This fix seems appropriate based on the issue description. There were a few minor comments to address; however, one larger question is why there are no math diffs here? Are there any example files with zones containing multiple ceiling sub-surfaces where this would apply?

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
jcyuan2020: No, the construction actually does not matter here as far it is a normal construction that would be counted to gross or net area of a surface. Here only a simple (or simplest) version.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
jcyuan2020: Same as above--construction does not matter here.

tst/EnergyPlus/unit/SurfaceGeometry.unit.cc
jcyuan2020: 3.25 and 4.0 (in meters) are the average height of the respective roof surfaces. This is an example of multiple roof surfaces in the same zone, with different heights--two ceiling surfaces has height of 3.25 and one has a height of 4m in the unit test case.

jcyuan2020:
> 
> 
> This fix seems appropriate based on the issue description. There were a few minor comments to address; however, one larger question is why there are no math diffs here? Are there any example files with zones containing multiple ceiling sub-surfaces where this would apply?

@mitchute Thanks for the review and the comments! There are two special conditions simultaneously required for the diff to show up: the first one is that a zone that contains two or more roof surfaces at different heights; and second one is that the ceiling surfaces contain a window (or skylight)--at least one of them needs to have some window (which make the net surface area not equal to the gross surface area. 

The above two conditions rarely happens (or do not happen at all) in the test file examples. But the situation is not difficult to imagine or encounter in reality (may not be a rare case). Actually it could be a commonly seen scenario, such as in residential house wit asymmetric pitched roofs where the two roofs surfaces could start at different height and also one side may have a skylight (or window). In helpdesk ticket 15707 the user also modeled such a case (consider this is an actual simulation configuration that reflected a real case from a modeler). The defect file of the issue post also shows such an example (it is the same three-roof/ceiling surfaces example in the test file).

mitchute (approved):

mitchute:
OK, thanks for the explanation. Unit tests all pass locally. Merging.

PR #8786, opened 2021-05-25
Solar and thermal absorptance calculation

xuanluo113:
Pull request overview
---------------------
 - This branch reduces unnecessary loop iterations in `ComputeIntThermalAbsorpFactors` and `ComputeIntThermalAbsorpFactors` by prechecking `state.dataHeatBal->EnclRadReCalc` (if an enclosure needs to be recalculate at a certain time step) at the end of `WindowShadingManager`. The solar and thermal absorptances do not need to be recalculated when no shading or glazing control status changes.
 - Better namings of related HB arrays and initialization processes.
 
### Pull Request Author

### Reviewer
 - [x] If branch is behind develop, merge develop and build locally to check for side effects of the merge~


xuanluo113:
@mjwitte Mike, this branch is also ready for review. Pulled upstream develop branch and updated PR overview. CI looks happy. Tested locally with 10x incremental models, the branch reduces 90% of `ComputeIntThermalAbsorpFactors` and `ComputeIntThermalAbsorpFactors` run time, at the cost of prechecking a few shading status change conditions.

src/EnergyPlus/DataGlobals.hh
```diff
@@ -142,6 +142,7 @@ struct DataGlobal : BaseGlobalStruct
     bool FinalSizingHVACSizingSimIteration = false; // when doing HVAC sizing Simulation
     bool AnyEnergyManagementSystemInModel = false;  // true if there is any EMS or Erl in model.  otherwise false
     bool AnySurfPropOverridesInModel = false;       // true if there is any EMS or Erl overriding the surface properties for any surface.
+    bool AnyConstrOverridesInModel = false;         // true if there is any EMS or Erl overriding the surface properties for any surface.
```
mjwitte: The comments here need updating to distinguish between these two flags.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4017,29 +4021,30 @@ void ComputeIntThermalAbsorpFactors(EnergyPlusData &state)
...
-    using General::InterpSw;
     auto &Surface(state.dataSurface->Surface);
-
-    if (!allocated(state.dataHeatBal->ITABSF)) {
-        state.dataHeatBal->ITABSF.dimension(state.dataSurface->TotSurfaces, 0.0);
-        state.dataHeatBal->TMULT.dimension(state.dataGlobal->NumOfZones, 0.0);
-    }
-
     for (int zoneNum = 1; zoneNum <= state.dataGlobal->NumOfZones; ++zoneNum) {
-        int const firstSurf = state.dataHeatBal->Zone(zoneNum).HTSurfaceFirst;
```
mjwitte, line 4029: If `EnclRadReCalc` is false for the zones in a particular enclosure, these loops can be skipped, if I'm following correctly. So, can these updates be moved inside the enclosure loop after 
`if (!state.dataHeatBal->EnclRadReCalc(radEnclosureNum)) continue;` ?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4166,24 +4165,13 @@ void ComputeIntSWAbsorpFactors(EnergyPlusData &state)
...
-
-    if (!allocated(state.dataHeatBal->EnclSolVMULT)) {
-        state.dataHeatBal->EnclSolVMULT.dimension(state.dataViewFactor->NumOfSolarEnclosures, 0.0);
-    }
-    if (state.dataHeatBalSurfMgr->ComputeIntSWAbsorpFactorsfirstTime) {
-        state.dataHeatBalSurfMgr->FirstCalcZone.dimension(state.dataGlobal->NumOfZones, true);
-        state.dataHeatBalSurfMgr->ComputeIntSWAbsorpFactorsfirstTime = false;
-    }
+    // Avoid a division by zero of the user has entered a bunch of surfaces with zero absorptivity on the inside
+    Real64 const SmallestAreaAbsProductAllowed(0.01);
```
mjwitte: `constexpr` ?

src/EnergyPlus/HeatBalanceSurfaceManager.cc
```diff
@@ -4556,18 +4545,15 @@ void InitEMSControlledConstructions(EnergyPlusData &state)
...
 
-    // SUBROUTINE LOCAL VARIABLE DECLARATIONS:
-    bool SurfConstructOverridesPresent(false); // detect if EMS ever used for this and inits need to execute
-
     auto &Surface(state.dataSurface->Surface);
 
     for (int SurfNum = 1; SurfNum <= state.dataSurface->TotSurfaces; ++SurfNum) {
         if (state.dataSurface->SurfEMSConstructionOverrideON(SurfNum)) {
-            SurfConstructOverridesPresent = true;
+            state.dataGlobal->AnyConstrOverridesInModel = true;
```
mjwitte, line 4557: This loop can be skipped once `AnyConstrOverridesInModel = true`. I did not understand that this flag (and `AnySurfPropOverridesInModel`) are continuously checked and only flip true if an EMS control takes action. But after that, it's never reset to `false` so this loop is redundant once the flag goes true. Same in `InitEMSControlledSurfaceProperties`.

Also, in this function, `InitEMSControlledConstructions`, `SurfActiveConstruction` is set to `SurfEMSConstructionOverrideValue` when the EMS override is on, but it's not reset to `ConstructionStoredInputValue` when the actuator is inactive (see line 4737).

src/EnergyPlus/SolarShading.cc
```diff
@@ -10018,6 +10019,29 @@ void WindowShadingManager(EnergyPlusData &state)
             }
         } // End of surface loop
     }
+    for (int enclosureNum = 1; enclosureNum <= state.dataViewFactor->NumOfSolarEnclosures; ++enclosureNum) {
+        state.dataHeatBal->EnclRadReCalc(enclosureNum) = false;
+        if (state.dataGlobal->BeginEnvrnFlag || state.dataGlobal->AnyConstrOverridesInModel ||
```
mjwitte: Seems like this should also check `|| `AnySurfPropOverridesInModel` .

Also, because the triggers here relate to things other than window shading, this enclosure recalc check should probably be in its own function, not in `WindowShadingManager`.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10018,6 +10019,29 @@ void WindowShadingManager(EnergyPlusData &state)
...
+            (state.dataWindowManager->inExtWindowModel->isExternalLibraryModel() && state.dataWindowManager->winOpticalModel->isSimplifiedModel())) {
```
mjwitte: Curious, why does `winOpticalModel->isSimplifiedModel()))` requires `EnclRadReCalc` every time?

src/EnergyPlus/SolarShading.cc
```diff
@@ -10018,6 +10019,29 @@ void WindowShadingManager(EnergyPlusData &state)
...
+            state.dataHeatBal->EnclRadReCalc(enclosureNum) = true;
+            continue;
+        }
+        for (int const SurfNum : state.dataViewFactor->ZoneRadiantInfo(enclosureNum).SurfacePtr) {
+            if (state.dataConstruction->Construct(state.dataSurface->Surface(SurfNum).Construction).TCFlag == 1 ||
+                state.dataConstruction->Construct(state.dataSurface->Surface(SurfNum).Construction).WindowTypeEQL) {
```
mjwitte: Same question here, why do `WindowTypeEQL` trigger recalc?

Should these checks be looking at `Construction(SurfActiveConstruction(SurfNum)`?

If one of these conditions is true, it's likely to be true for the entire simulation. Seems like there should be a way to skip this surface loop if we already know that. I suppose this could change status if there's a surface construction override, but that would be caught above  by `AnyConstrOverridesInModel `.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10018,6 +10019,29 @@ void WindowShadingManager(EnergyPlusData &state)
...
+                state.dataHeatBal->EnclRadReCalc(enclosureNum) = true;
+                break;
+            }
+            bool surfShadingStatusChange =
```
mjwitte, line 10062: Is there a flag to skip these checks if there are no window shades or blinds or .... in the model?

mjwitte (commented):
Some questions on first review.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: Within function scope, there is probably not a significant difference between `const` and `constexpr`. However, it is still a good idea to get into the habit of using `constexpr` for these things.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
amirroth: Is that the right fix? Or is the right fix to reset these flags to `false` at the end of `InitEMSControlledConstructions` property re-evaluation function and have EMS changes set them to `true` if and when construction/surface variables are updated?

src/EnergyPlus/SolarShading.cc
xuanluo113: The `transdiff` of windows gets reevaluated based on sun angle at [CalcInteriorSolarDistributionWCESimple](https://github.com/NREL/EnergyPlus/blob/6dadedaca31b6ee3823de26ab1e57f6f99256bd3/src/EnergyPlus/SolarShading.cc#L8630) and is used in the solar absorptance function.

src/EnergyPlus/SolarShading.cc
xuanluo113: `AbsDiffTotWin` is calculated differently in [ComputeIntSWAbsorpFactors](https://github.com/NREL/EnergyPlus/blob/6dadedaca31b6ee3823de26ab1e57f6f99256bd3/src/EnergyPlus/HeatBalanceSurfaceManager.cc#L4293) and the regression tests catch the diffs if the condition not applied.

mjwitte: OK, I will trust you on this. I don't understand why the current sun angles should have any impact on the diffuse-diffuse transmittance, but what do I know.  I have much to learn about the many layers of window models.

src/EnergyPlus/DataGlobals.hh
xuanluo113: Fixed.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: I added a skip condition check at this loop, so `if (state.dataSurface->AnyMovableSlat)` can still be handled separately after this loop.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: Modified.

src/EnergyPlus/HeatBalanceSurfaceManager.cc
xuanluo113: `AnyConstrOverridesInModel` and `AnySurfPropOverridesInModel` are set to false at the begining of each time step. Also `ConstructionStoredInputValue` is restored when EMS override is off.

src/EnergyPlus/SolarShading.cc
xuanluo113: Moved to a separate function and added`|| AnySurfPropOverridesInModel`.

src/EnergyPlus/SolarShading.cc
xuanluo113: I added a global condition of `EnclRadAlwaysReCalc` to be calculated at the beginning of the whole simulation. All immutable conditions (over the entire simulation) are checked once here. And if `EnclRadAlwaysReCalc`, the checks can be bypasses to always calculate.

src/EnergyPlus/SolarShading.cc
xuanluo113: I added a check of if `state.dataGlobal->AndShadingControlInModel`. If not, `recalc` is always false and the following checks are bypassed.

xuanluo113:
@mjwitte I've addressed all the comments so far. Ready for another round of review after the CI is clean.

src/EnergyPlus/SolarShading.cc
```diff
@@ -10020,6 +10021,54 @@ void WindowShadingManager(EnergyPlusData &state)
     }
 }
 
+void CheckGlazingShadingStatusChange(EnergyPlusData &state)
+{
+    if (state.dataGlobal->BeginSimFlag) {
+        if (state.dataWindowManager->inExtWindowModel->isExternalLibraryModel() && state.dataWindowManager->winOpticalModel->isSimplifiedModel()) {
+            state.dataHeatBal->EnclRadAlwaysReCalc = true;
+        }
+        for (int ZoneNum = 1; ZoneNum <= state.dataGlobal->NumOfZones; ++ZoneNum) {
```
mjwitte: Minor detail, this `for` loop can be inside an `else` block. If `EnclRadAlwaysReCalc` has already been set true, there's no need for this loop. I know, this happens just once at the beginning of the simulation, so it's fine if there are no other changes needed.

mjwitte (approved):
@xuanluo113 This looks good. I'm merging in develop one more time hoping the odd CI test failures will go away. If it looks good in the morning, this will merge.

mjwitte:
@xuanluo113 There are two array abounds failures in the integration coverage tests, 5ZoneAirCooled_AirBoundaries, 5ZoneAirCooled_AirBoundaries_Daylighting. Since I may have had something to do with air boundaries, let me know if you need any help fixing these.

xuanluo113:
@mjwitte Fixed the air boundary issue by allocating `NumOfZones` instead of `NumOfSolarEnclosures`. However, `Encl` or `Zone` array allocations are somehow a little confusing for me. These [AllocateZoneHeatBalArrays](https://github.com/NREL/EnergyPlus/blob/b253fb567c4e4cd4fd657cc04b496780ebc42461/src/EnergyPlus/HeatBalanceManager.cc#L5708) functions here includes these array allocations, some by the number of zones and other by the number of enclosures. When I sorted this function with some naming convention fixes in the previous branch, I used the arrays original dimensions (in dev branch), and some `Encl-` arrays were allocated with `NumOfZones`.
So I was thinking if I wrongly renamed some arrays with intuition. For example, `EnclSolDBSSG` looks like to be evaluated/assigned at enclosure level, but the initialization and usage are sometimes by `zoneNum`.

mjwitte:
@xuanluo113 (Without looking closely at the code) `NumOfZones` is always safe (for now), because `NumOfSolarEnclosures` and `NumOfRadiantEnclosures` are always = `NumOfZones` if there are not air boundaries, but can be < `NumOfZones` when there are air boundaries. 

So, the question here is whether using `NumOfSolarEnclosures` failed because it wasn't set yet, or because the array(s) really should be for `NumOfZones` or some other reason.

(Ok, started looking at code). Part of the problem is that there are places that used to loop over Zones (long ago) that now loop over enclosures, but the air boundary developer (was too lazy and) did not change the loop index, e.g. InitSolarHeatGains `for (int ZoneNum = 1; ZoneNum <= state.dataViewFactor->NumOfSolarEnclosures; ++ZoneNum)`.

`EnclSolDBSSG` is clearly being misused when it is accessed by `ZoneNum` inside a loop on `NumOfZones` which I see at the top of `CalcInteriorSolarDistribution`. But there it's just setting to zero. So, I think (based on a few minutes of searching) that `EnclSolDBSSG` truly is an enclosure variable and should be sized to `NumOfSolarEnclosures`.

These lines (SolarShading 6290:6292)
```
        state.dataHeatBal->EnclSolDB(zoneNum) = 0.0;
        state.dataHeatBal->EnclSolDBSSG(zoneNum) = 0.0;
        state.dataHeatBal->EnclSolDBIntWin(zoneNum) = 0.0;
```
should move out of the `NumOfZones` loop down to the `NumOfSolarEnclosures` loop (line 6320).

And someday all of the local loop variables named like `BTOTZone` should be renamed to `BTOTEnclosure`.

All of this will come back to haunt me when Space changes things to allow `NumOfSolarEnclosures` to be `< || > NumOfZones`.

xuanluo113:
Thanks! I will leave this branch clean (hopefully) and fix the `EnclSolDB` (and other arrays) index domain issue in my heat balance data structure branch.

mjwitte:
CI is all green and happy. Merging.

PR #8791, opened 2021-05-28
Support HVACTemplate objects in converter and trap unsupported objects in epJSON input

### LINKED ISSUE #8743: No Error Message for epJSON files with HVACTemplate, Parametric, or GroundHeatTransfer objects
Issue overview
--------------
The command line tool appears to accept and simulate epJSON files that contain HVACTemplate objects.  However, from the outputs, it looks like the HVACTemplate objects are ignored during processing, but no error or warning messages are explicitly provided.  The example provided below is using a converted epJSON file from the ExampleFiles directory of the same name.

`wine ../energyplus.exe -w ../WeatherData/USA_CO_Golden-NREL.724666_TMY3.epw ../ExampleFiles/HVACTemplate-5ZoneVAVWaterCooled.epJSON`

Console Message:

`EnergyPlus Starting`
`EnergyPlus, Version 9.4.0-998c4b761e, YMD=2021.05.02 21:19`
`Adjusting Air System Sizing`
`Adjusting Standard 62.1 Ventilation Sizing`
`Initializing Simulation`
`Reporting Surfaces`
`Beginning Primary Simulation`
`Initializing New Environment Parameters`
...
`EnergyPlus Completed Successfully.`

### Details
 - Platform : Linux
 - EnergyPlus 9.4.0
 - Command was executed via wine CLI tool
 - Example file is located [here](https://github.com/john-grando/pyExpandObjects/blob/main/simulation/ExampleFiles/HVACTemplate-5ZoneVAVWaterCooled.epJSON)
 - Similar file attached here [HVACTemplate-5ZoneVAVWaterCooled.epJSON.txt](https://github.com/NREL/EnergyPlus/files/6563094/HVACTemplate-5ZoneVAVWaterCooled.epJSON.txt)

mjwitte:
Pull request overview
---------------------
Various updates related to HVACTemplate objects and epJSON in support of the new epJSON-based HVACTemplate expander.

 - Fixes #8743 - If HVACTemplate:\*, GroundHeatTransfer:\*, and Parametric:\* objects are present in an incoming epJSON or idf file, throw severe errors and quit prior to simulation.
 - Allow ConvertInputFormat.exe to convert HVACTemplate:* objects by default.
 - Add command line option to ConvertInputFormat.exe to *not* convert HVACTemplate objects.
   `-n` or `--noHVACTemplate`
 - Fix ConvertInputFormat output line endings on Windows. The converted file (epJSON or idf) had non-DOS line endings, that's fixed here.
 - Add documentation for ConvertInputFormat to the EnergyPlus Essentials doc.
 - Fix the cmake depedencies and rename the main source file for the Essentials doc.

Questions
-------------
1. EnergyPlus -c will still reject HVACTemplate objects and not convert - are we ok with that?

2. ConvertInputFormat will throw errors and not convert the entire file if it finds unsupported objects. Do we want a method available that converts everything except the unsupported object?

### Pull Request Author

### Reviewer


src/ConvertInputFormat/main.cpp
```diff
@@ -145,8 +145,12 @@ void cleanEPJSON(json &epjson)
...
 }
 
-bool processInput(
-    std::string const &inputFilePath, json const &schema, OutputTypes outputType, std::string outputDirectory, std::string &outputTypeStr)
+bool processInput(std::string const &inputFilePath,
+                  json const &schema,
+                  OutputTypes outputType,
+                  std::string outputDirectory,
+                  std::string &outputTypeStr,
+                  bool convertHVACTemplate)
```
mjwitte, line 273: Add parameter to control whether HVACTemplate objects are converted or rejected.

src/ConvertInputFormat/main.cpp
```diff
@@ -219,26 +223,20 @@ bool processInput(
...
-    std::string const NL("\n"); // Platform newline
-#endif
-
     try {
         if (!isEpJSON) {
             std::string input_file;
             std::string line;
             while (std::getline(input_stream, line)) {
-                input_file.append(line + NL);
+                input_file.append(line + "\n");
```
mjwitte, line 351: Use `\n` instead of `NL`.

src/ConvertInputFormat/main.cpp
```diff
@@ -275,14 +273,14 @@ bool processInput(
         auto const input_file = idf_parser->encode(epJSON, schema);
         std::string convertedEpJSON(outputDirectory + inputFileNameOnly + ".idf");
         EnergyPlus::FileSystem::makeNativePath(convertedEpJSON);
-        std::ofstream convertedFS(convertedEpJSON, std::ofstream::out | std::ofstream::binary);
+        std::ofstream convertedFS(convertedEpJSON, std::ofstream::out);
```
mjwitte, line 398: Remove `binary` option for epJSON and idf output so they have dos line endings on windows. Note this matches [similar lines](https://github.com/NREL/EnergyPlus/blob/develop/src/EnergyPlus/InputProcessing/InputProcessor.cc#L303) in the main EnergyPlus code.

src/ConvertInputFormat/main.cpp
```diff
@@ -375,6 +373,15 @@ int main(int argc, const char *argv[])
             "--format",                                                        // Flag token.
             outputTypeValidation);
 
+    opt.add("",                              // Default.
+            0,                               // Required?
+            0,                               // Number of args expected.
+            0,                               // Delimiter if expecting multiple args.
+            "Do not convert HVACTemplate objects.", // Help description.
+            "-n",                            // Flag token.
+            "--noHVACTemplate"               // Flag token.
```
mjwitte: New command line option for ConvertInputFormat to disable converting HVACTemplate objects.

src/ConvertInputFormat/main.cpp
```diff
@@ -428,6 +435,11 @@ int main(int argc, const char *argv[])
         files = parse_input_paths(input_paths_file);
     }
 
+    bool convertHVACTemplate = true;
```
mjwitte, line 560: ConvertInputFormat converts HVACTemplate objects by default.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -229,11 +229,16 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success,
                     token = next_token(idf, index);
                 continue;
             } else if (obj_name.find("Parametric:") != std::string::npos) {
-                errors_.emplace_back(fmt::format("Line: {} You must run Parametric Preprocessor for \"{}\"", cur_line_num, obj_name));
+                errors_.emplace_back(fmt::format("Line: {} You must run the Parametric Preprocessor for \"{}\"", cur_line_num, obj_name));
                 while (token != Token::SEMICOLON && token != Token::END)
                     token = next_token(idf, index);
                 continue;
-            } else if (obj_name.find("Template") != std::string::npos) {
+            } else if (obj_name.find("GroundHeatTransfer:") != std::string::npos) {
```
mjwitte: Trap GroundHeatTransfer:\* objects.

src/EnergyPlus/InputProcessing/IdfParser.hh
```diff
@@ -67,7 +67,7 @@ public:
 
     json decode(std::string const &idf, json const &schema);
 
-    json decode(std::string const &idf, json const &schema, bool &success);
+    json decode(std::string const &idf, json const &schema, bool &success, bool const convertHVACTemplate = false);
```
mjwitte: EnergyPlus built-in file conversion defaults to not convert HVACTemplate objects.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -1494,6 +1484,31 @@ void InputProcessor::reportOrphanRecordObjects(EnergyPlusData &state)
             }
         }
 
+        // there are some orphans that we are deeming as special, in that they should be warned in detail even if !DisplayUnusedObjects and
+        // !DisplayAllWarnings
+        if (has_prefix(object_type, "ZoneHVAC:")) {
+            ShowSevereError(state, "Orphaned ZoneHVAC object found.  This was object never referenced in the input, and was not used.");
+            ShowContinueError(state, " -- Object type: " + object_type);
+            ShowContinueError(state, " -- Object name: " + name);
+        } else if (has_prefix(object_type, "Parametric:")) {
```
mjwitte: Add warnings for Parametric:\*, GrounHeatTransfer:\*, and HVACTemplate:\* to the orphan (unused) objects report. If these types of objects are present in an incoming epJSON file, they will be ignored, the simulation will run, and these warnings will be generated. If they are present in an incoming idf file, the input processor rejects them and stops with a fatal error.

mjwitte (commented):
Code walkthrough.

Myoldmopar:
@mbadams5 it would be good if you could take a look at the proposed changes here, at least from a design perspective.

src/ConvertInputFormat/main.cpp
mbadams5: This is fine. I likely just copied and pasted for all the various `ofstream` since cbor, msgpack, ubjson, and bson all require the binary flag since they are binary formats.

src/EnergyPlus/InputProcessing/InputProcessor.cc
mbadams5: Why the difference in how epJSON and IDF are treated when they contain these objects? Why does epJSON ignore them and continue to run, however, IDF will stop with a fatal error? Shouldn't the functionality be the same between the input formats if they are supposed to be identical input formats currently (aka do the same thing)?

Or is this comment only specific to the ConvertInputFormat program?

mbadams5 (commented):
Overall I think this is fine except for my one question about difference between how epJSON and IDF treat unsupported objects.

Regarding your two questions in the PR:

1. EnergyPlus -c will still reject HVACTemplate objects and not convert - are we ok with that?

> If the input file is IDF, my opinion is that they should run the external programs first then run E+ with `-c`. I think if the input is epJSON with HVACTemplate objects, then it should convert these to IDF BUT I don't think it should ignore them and continue the simulation. The simulation should fatal error without the external programs running these objects first, but it could be confusing to users if it did complete. 

2. ConvertInputFormat will throw errors and not convert the entire file if it finds unsupported objects. Do we want a method available that converts everything except the unsupported object?

> This is a good question. Should ConvertInputFormat just convert everything back and forth without caring about validity of the simulation? Within E+, it obviously should care if the inputs are valid and the simulation will complete or not. I think the bigger issue is whether or not epJSON ignores and continues simulation or fatal errors like IDF when it encounters unsupported objects.

nrel-bot-3:

doc/essentials/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 # to get the list of input files, go to the root of the document folder, and:
 # grep -r '\\input{' * | cut -d ':' -f 2 | cut -c 8- | rev | cut -c 2- | rev | sed 's|^|  ${PROJECT_SOURCE_DIR}/acknowledgments/|g' | sed 's/$/.tex/g'
-set(INCLUDED_TEX ${PROJECT_SOURCE_DIR}/header.tex ${PROJECT_SOURCE_DIR}/essentials/essentials.tex ${PROJECT_SOURCE_DIR}/title.tex)
+set(INCLUDED_TEX 
+    ${PROJECT_SOURCE_DIR}/header.tex 
+    ${PROJECT_SOURCE_DIR}/essentials/essentials.tex 
+    ${PROJECT_SOURCE_DIR}/essentials/src/essentials.tex
```
Myoldmopar, line 6: :+1:

doc/essentials/src/essentials.tex
```diff
@@ -692,14 +692,53 @@ \subsection*{epJSON}
...
+                                   location by default.
+  -v, --version                    Display version information
+Example: ConvertInputFormat in.idf
+\end{verbatim}
+
+On Windows, you can also drag a file (IDF or epJSON) onto ConvertInputFormat.exe.
+Note that ConvertInputFormat is version-specific and will not convert an epJSON file that is not the same version.
+
+There are three groups of special input objects which are preprocessor commands: GroundHeatTransfer:*,
+HVACTemplate:* Parametric:*. ConverInputFormat converts HVACTemplate objects.  EnergyPlus does not convert any of these.
```
Myoldmopar, line 739: Nice.

src/ConvertInputFormat/main.cpp
Myoldmopar: :+1:

src/ConvertInputFormat/main.cpp
Myoldmopar: OK.  I'm curious how this parameter is assigned in the calling routine.

src/ConvertInputFormat/main.cpp
```diff
@@ -220,26 +225,20 @@ bool processInput(std::string const &inputFilePath, json const &schema, OutputTy
...
+                input_file.append(line + "\n");
             }
             if (input_file.empty()) {
                 displayMessage("Failed to read input file: " + inputFilePath);
                 return false;
             }
 
             bool success = true;
-            epJSON = idf_parser->decode(input_file, schema, success);
+            epJSON = idf_parser->decode(input_file, schema, success, convertHVACTemplate);
```
Myoldmopar: OK, so decode also has this now.

src/ConvertInputFormat/main.cpp
```diff
@@ -377,6 +376,15 @@ int main(int argc, const char *argv[])
             "--format",                                                        // Flag token.
             outputTypeValidation);
 
+    opt.add("",                                     // Default.
+            0,                                      // Required?
+            0,                                      // Number of args expected.
+            0,                                      // Delimiter if expecting multiple args.
+            "Do not convert HVACTemplate objects.", // Help description.
+            "-n",                                   // Flag token.
+            "--noHVACTemplate"                      // Flag token.
```
Myoldmopar, line 504: Alright, so a `--noHVACTemplate` flag on the converter command line.

src/ConvertInputFormat/main.cpp
```diff
@@ -430,6 +438,11 @@ int main(int argc, const char *argv[])
         files = parse_input_paths(input_paths_file);
     }
 
+    bool convertHVACTemplate = true;
+    if (opt.isSet("-n")) {
+        convertHVACTemplate = false;
+    }
```
Myoldmopar, line 563: But then if you pass `-n` it won't convert HVACTemplate objects.

src/EnergyPlus/InputProcessing/InputProcessor.cc
Myoldmopar: It probably shouldn't be treated all that much different, but maybe the messaging should be slightly different between epJSON and IDF inputs.  For IDF inputs, yes, you need to just run ExpandObjects and you'll get a new IDF with those objects expanded.  For epJSON inputs, you can't run the IDF-based ExpandObjects, and we don't quite have PyExpandObjects yet for EpJSON files.

Myoldmopar (approved):
This actually looks quite good.  There is some uncertainty around making sure we are treating both file types in a similar way, but that may need to be polished after IO freeze.  I'm testing this locally now.

Myoldmopar:
Pulled in develop and built, no problem.  Full test suite passes fine.

I then ran started with an HVAC Template IDF file, and ran some tests:
 - Converted the file from IDF to EpJSON
   - Ran `./ConvertInputFormat template.idf` successfully
   - Output epJSON file does have the HVAC template objects in place
 - Took that epJSON output and converted it back to IDF
   - Ran `./ConvertInputFormat converted.epJSON` successfully
   - Output IDF file again still has HVACTemplates sitting there
 - Converted IDF while excluding template objects
   - Ran `./ConvertInputFormat -n template.idf` successfully
   - Resulted in a failed conversion with several output messages like:
     - Line: 1853 You must run the ExpandObjects program for "HVACTemplate:Thermostat"
 - Converted EpJSON while excluding Template objects
   - Ran the IDF->EpJSON conversion again successfully
   - Ran `./ConvertInputFormat -n template.epJSON` successfully
   - Expected it to fail, I guess this is the missing piece.
 - Ran EnergyPlus with an IDF with template objects
   - Ran: `./energyplus -D hvactemplate.idf` with a failure
   - Similar output messages about ExpandObjects in the error file
 - Ran EnergyPlus with an EpJSON with template objects
   - Ran `./energyplus -D hvactemplate.epJSON` and it succeeded.  Which seems to be the incorrect behavior.
   - The error file **does** have a bunch of severe errors about the found HVACTemplate objects, but it doesn't fatal error.

OK, so the code really isn't ready yet.  I will hold off from reviewing or testing any further at this moment and move onto another PR.  Let me know if I can do anything else here, @mjwitte.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -228,16 +228,6 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success,
                 while (token != Token::SEMICOLON && token != Token::END)
                     token = next_token(idf, index);
                 continue;
-            } else if (obj_name.find("Parametric:") != std::string::npos) {
```
mjwitte, line 231: Move checks for unsupported objects out of `parse_idf` so the same checks can apply for idf and epJSON input.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -248,11 +238,11 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success,
                 auto found_index = idf.find_first_of('\n', beginning_of_line_index);
                 std::string line;
                 if (found_index != std::string::npos) {
-                    line = idf.substr(beginning_of_line_index, found_index - beginning_of_line_index);
+                    line = idf.substr(beginning_of_line_index, found_index - beginning_of_line_index - 1);
```
mjwitte, line 241: Fix extra newline char in error messages.

src/EnergyPlus/InputProcessing/IdfParser.cc
```diff
@@ -248,11 +238,11 @@ json IdfParser::parse_idf(std::string const &idf, size_t &index, bool &success,
...
                 }
                 errors_.emplace_back(fmt::format("Line: {} Index: {}", cur_line_num, index_into_cur_line) + " - Error parsing \"" + obj_name +
                                      "\". Error in following line.");
-                errors_.emplace_back("~~~ " + line);
+                errors_.emplace_back(fmt::format("~~~ {}", line));
```
mjwitte, line 245: This was the first attempt to fix the newline problem, decided to just leave it.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -322,8 +322,9 @@ void InputProcessor::processInput(EnergyPlusData &state)
     bool is_valid = validation->validate(epJSON);
     bool hasErrors = processErrors(state);
     bool versionMatch = checkVersionMatch(state);
+    bool unsupportedFound = checkForUnsupportedObjects(state);
```
mjwitte, line 325: Check for unsupported objects after idf/epJSON input is loaded.

src/EnergyPlus/InputProcessing/InputProcessor.cc
```diff
@@ -383,6 +384,121 @@ bool InputProcessor::checkVersionMatch(EnergyPlusData &state)
     return true;
 }
 
+bool InputProcessor::checkForUnsupportedObjects(EnergyPlusData &state)
```
mjwitte, line 387: New function to check for unsupported objects prior to simulation.

src/ConvertInputFormat/main.cpp
```diff
@@ -109,6 +109,121 @@ bool checkVersionMatch(json const &epJSON)
     return true;
 }
 
+bool checkForUnsupportedObjects(json const &epJSON, bool convertHVACTemplate)
```
mjwitte, line 112: New function to check for unsupported objects in format converter instead of relying on `parse_idf`.

src/ConvertInputFormat/main.cpp
```diff
@@ -109,6 +109,121 @@ bool checkVersionMatch(json const &epJSON)
...
+    std::string objectType;
+    for (size_t count = 0; count < hvacTemplateObjects.size(); ++count) {
+        objectType = hvacTemplateObjects[count];
+        auto it = epJSON.find(objectType);
+        if (it != epJSON.end()) {
+            objectFound = true;
+            break;
+        }
+    }
+    if (objectFound && !convertHVACTemplate) {
```
mjwitte: Here is where `convertHVACTemplate` is used to skip this. Looking at it now, this could actually be moved up to skip the entire HVACTemplate block. *Done*

src/ConvertInputFormat/main.cpp
```diff
@@ -109,6 +109,124 @@ bool checkVersionMatch(json const &epJSON)
...
+                                                                      "HVACTemplate:Plant:HotWaterLoop",
+                                                                      "HVACTemplate:Plant:Boiler",
+                                                                      "HVACTemplate:Plant:Boiler:ObjectReference",
+                                                                      "HVACTemplate:Plant:MixedWaterLoop"};
+
+    bool objectFound = false;
+    std::string objectType;
+
+    // For ConvertInputFormat, skip this unless -n option is present to not allow conversion of HVACTemplate objects
+    if (!convertHVACTemplate) {
```
mjwitte, line 152: This is where the -n command line argument triggers rejecting HVACTemplate objects for both idf and epJSON input.

mjwitte (commented):
@Myoldmopar Here's plan B for the unsupported object checks. Both EnergyPlus and ConvertInputFormat will reject GroundHeatTransfer and Parametric objects. EnergyPlus will always reject HVACTemplate objects, and ConvertInputFormat will accept them by default by reject them with -n.

Myoldmopar:
This all looks great.  I've pulled develop locally and am building and will run it through a series of tests and get this merged next hopefully.

Myoldmopar:
Built fine, tested the input converter tool as well as EnergyPlus itself.  On the converter, it runs fine by default but complains with the new command line argument, for both input forms, as expected.  EnergyPlus always complains regardless of input syntax now.

Windows CI had a hiccup there, but it's not because of this branch.  I'm going to merge this now.  Thanks @mjwitte !

PR #8795, opened 2021-06-02
Performance refactor for Schedule:File

jmythms:
Pull request overview
---------------------
This pull request refactors the Schedule:File module in ScheduleManager.cc mainly focusing on improving run times.

The current Schedule:File module mainly has three steps:
1. Get input from IDF.
2. Parse CSV according to input from IDF from the last step.
3. Save parsed information to schedule data structure. 

Out of these three, parsing the CSV is the most expensive step, because it involves reading information from potentially very big external files. This step also has the potential for generating maximum performance gain with a refactor, mainly because of problems with:
1. Repeating the file open-close operations for each Schedule:File object in the IDF
2. Repeating the unnecessary delimiter checks each time the schedule is run

We address both of these problems in one step by reading in information from the CSV file in one step. This is explained below:

#### Current Method as in Develop: 
It parses all Schedule:File in the following way:

1. For each Schedule:File Object, it
2. Opens file
3. Parses each line, scans the line looking for the specific column
4. Once it finds the correct column, it converts the token in that cell to a Real64 value and saves it to an array
5. Repeat till end of file
6. Close file
7. Repeat for next Schedule:File object in IDF

#### Fig 1.
Screenshot from 2021-06-01 13-05-47

This is how it looks in the above example:

0. Open file
1. Skip header row
2. Read the second line, process 1, skip 2,3,4,5,
3. Read the third line, process 6, skip 7,8,9,10
4. Repeat till end of file
5. Close file
6. Open the file again.
7. Skip header row
8. Read the second line, skip 1, process 2, skip 3,4,5
9. Read the third line, skip 6, process 7, skip 8,9,10
10. Repeat till end of file
11. Close file... 

Repeat till all Schedule:File objects are processed

#### Proposed Method:

1. Scan IDF for all files associated with Schedule:File
2. For each file:
3. Find the schedule:File objects associated with it.
4. Open file
5. Read the row, process all tokens of interest in that specific row in the file to Real64.
6. Repeat for every row
7. Go till the end of the file
8. Repeat for every file
  
This is how it looks for the example:
0. Open file
1. Skip header row
2. Read first line, process 1,2,3,4,5
3. Read second line, process 6,7,8,9,10
4. Repeat for every row
5. Close file
6. Repeat for any other file

Advantages of this method:

1.For IDFs that have a large number of Schedule:File objects, decrease file open and file close operations significantly. These are very expensive IO operations.
2. Decrease the number of rows read, delimiter checks from N*(N+1)/2 to N. Where N is the number of columns in a file. 

### std::string_view
I tried to use std::string_view for the CSV parse part, since that will be the operation that is repeated most and it would be great to cut down the dynamic allocations made by string arrays. But I had to cast it to a string at some point, since replace_if needs a not-read-only object to replace things. 

## Performance Improvements
Percent SpeedUp for Schedule:File module and total simulation time when comparing this branch to develop is shown in Fig.2. Where percentage speed-up is calculated as (t_develop - t_branch)*100/t_branch.

#### Fig 2. 
Fig2 

We observe that the percentage speed-up is greater as the size and number of schedules to be parsed become greater. 
We also observe that the improvements in run-time don't affect the IDFs with a smaller number of schedules very much. This is because of two reasons:
i. The percentage of time spent on schedule manager is low, compared to the rest of the simulation as seen in Fig 3. Therefore, even huge improvements in the Schedule:Manager can only improve the total simulation run time by so much. 
ii. The improvement in schedule manager simulation is also low since the file open-close operations are very less. 

#### Fig 3.
Fig3

Time measured using the Chrono library between points [a](https://github.com/NREL/EnergyPlus/blob/b8f984adb5fe45cfebf838ab1fcf21d104e5091a/src/EnergyPlus/ScheduleManager.cc#L1644) and [b](https://github.com/NREL/EnergyPlus/blob/b8f984adb5fe45cfebf838ab1fcf21d104e5091a/src/EnergyPlus/ScheduleManager.cc#L1997).
The test IDFs (added in [this commit](https://github.com/NREL/EnergyPlus/commit/ca7692e319a7bfc61cce8933777f2c0dc060995b) ) were created by adding Schedule:File objects that used information from the SolarShadingTest_Shading_Data.csv. The schedule manager will process any Schedule:File object that is present in the IDF irrespective of whether it is used or not. This 'problem' was exploited to create this performance chart. 

## Going Forward:
1. This refactor seems to show a great improvement for bigger schedules. The File:Schedule:Shading module will greatly benefit from this methodology, mainly because the HybridZoneModel_shadow113.idf references the whole SolarShadingTest_Shading_Data.csv, and the average run time decreased from 17s to 7s on my local machine. 
2. My local machine ran through the smaller schedule test files quickly. I think someone with a slower machine might find improvements (by shaving off a few seconds) in the smaller IDFs also. (Referencing the ERR files from CI here).
3. There were slight variations in the total simulation time for similar runs. The data shown in the graphs are averages of three runs, the complete data file is linked below. An interesting observation was that the simulation time slightly increased sometimes even when the time to run schedule manager was decreased. I thought about using callgrind here but it will not measure performance improvements from reducing file operations. 
4. I don't understand why I am getting the unused variable warning. 
5. There is an issue if someone pointed multiple schedule objects to the same CSV column. I set up a fatal error for this scenario. This would have worked with the code in the current develop. 

## TO DO

- [x]  Implement many to one mapping support. 

### Pull Request Author

### Reviewer


src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
-                                         "ProcessScheduleInput: For " + CurrentModuleObject + "=\"" + Alphas(1) + "\", " + cAlphaFields(2) + "=\"" +
-                                             Alphas(2) + "\" not found -- will not be validated");
-                    } else {
-                        ShowWarningError(state,
-                                         "For " + CurrentModuleObject + "=\"" + Alphas(1) + "\", Blank " + cAlphaFields(2) +
-                                             " input -- will not be validated.");
-                    }
-                } else {
-                    state.dataScheduleMgr->Schedule(SchNum).ScheduleTypePtr = CheckIndex;
+
```
jmythms, line 1638: The main concept is that the processing moves from a Schedule:File object perspective to a CSV file perspective since we need to decrease file open and close operations as much as possible.

src/EnergyPlus/ScheduleManager.hh
```diff
@@ -328,6 +331,109 @@ namespace ScheduleManager {
...
+    private:
+        std::string m_line;
+        std::vector<int> m_data;
+    };
+
+    std::istream &operator>>(std::istream &str, CSVRow &data);
+
+    void PreProcessIDF(EnergyPlus::EnergyPlusData &state, int &SchNum, int NumCommaFileSchedules);
+
+    struct schedInputIdfPreprocessObject
```
jmythms, line 350: Objects of this struct will hold the IDF file info about a particular Schedule:File that is read by the getObjectItem function. Implementing it like this hoping to bring a more object-oriented approach to this module.

src/EnergyPlus/ScheduleManager.hh
```diff
@@ -355,6 +461,10 @@ struct ScheduleManagerData : BaseGlobalStruct
     Array1D<ScheduleManager::WeekScheduleData> WeekSchedule; // Week Schedule Storage
     Array1D<ScheduleManager::ScheduleData> Schedule;         // Schedule Storage
 
+    // Schedule:File variables
+    std::vector<ScheduleManager::schedInputIdfPreprocessObject> allIdfSchedData;
```
jmythms, line 465: A vector of schedInputIdfPreprocessObject type. We will add the IDF data about each Schedule:File object to this vector as schedInputIdfPreprocessObject elements.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+        // Runs getObjectItem for Schedule:File and saves it to vector in state: allIdfSchedData
+        PreProcessIDF(state, SchNum, NumCommaFileSchedules);
```
jmythms, line 1640: I wrapped the getObjectItem function and all the variable assignments into this PreProcessIDF function. This will populate the allIdfSchedData vector. We can access every Schedule:File object data from here, since these have been saved as elements of the allIdfSchedData vector. The main module also looks cleaner.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+
+        // Because we are focusing on saving performance by decreasing number of file open-close operations, get set of filenames
+        std::set<std::string> setOfFilenames;
+        PopulateSetOfFilenames(state.dataScheduleMgr->allIdfSchedData, setOfFilenames);
```
jmythms, line 1644: This function runs through the IDF data and saves all the CSV file names to the std::set setOfFilenames. std::set makes sure that the list only has unique members in it.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -5233,6 +5103,199 @@ namespace ScheduleManager {
...
+                                                                ColumnSep,
+                                                                SchNum);
+        }
+    }
+
+    void PopulateSetOfFilenames(const std::vector<schedInputIdfPreprocessObject> &allIdfSchedData, std::set<std::string> &setOfFilenames)
+    {
+        for (const schedInputIdfPreprocessObject &idfObject : allIdfSchedData) {
+            if (setOfFilenames.find(idfObject.fileName) == setOfFilenames.end()) {
+                setOfFilenames.insert(idfObject.fileName);
```
jmythms, line 5293: If we do not find the file name in the set, add it to the set...

src/EnergyPlus/ScheduleManager.hh
```diff
@@ -328,6 +331,109 @@ namespace ScheduleManager {
...
+        }
+        void setColumnarDataIndex(int incomingData)
+        {
+            this->columnarDataIndex = incomingData;
+        }
+    };
+
+    void PopulateSetOfFilenames(const std::vector<schedInputIdfPreprocessObject> &allIdfSchedData, std::set<std::string> &setOfFilenames);
+
+    struct PreProcessedColumn
```
jmythms, line 395: Objects of this struct will have a <Real64> vector (vals) that holds the actual schedule values from the CSV file.

src/EnergyPlus/ScheduleManager.hh
```diff
@@ -355,6 +461,10 @@ struct ScheduleManagerData : BaseGlobalStruct
     Array1D<ScheduleManager::WeekScheduleData> WeekSchedule; // Week Schedule Storage
     Array1D<ScheduleManager::ScheduleData> Schedule;         // Schedule Storage
 
+    // Schedule:File variables
+    std::vector<ScheduleManager::schedInputIdfPreprocessObject> allIdfSchedData;
+    std::vector<ScheduleManager::PreProcessedColumn> columnarData;
```
jmythms, line 466: A vector of PreProcessedColumn objects. This will hold the schedule data that will be later inserted into the schedule data structure.

src/EnergyPlus/ScheduleManager.hh
```diff
@@ -328,6 +331,109 @@ namespace ScheduleManager {
...
+        bool FileIntervalInterpolated = false;
+        int ScheduleTypePtr = -1;
+        int SchNum = -1;
+        int rowCnt = 0;
+        int numerrors = 0;
+        std::string fileName;
+        std::string delimiter;
+        std::vector<Real64> vals; // rows to skip + number of hours*items/hour
+
+        PreProcessedColumn(std::string name, // <- can inherit from parent sched class
```
jmythms, line 411: We implement constructors here, so emplace_back can be called, which is better than push_back because creating a copy is avoided.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+                                                                     item.columnOfInterest,
+                                                                     item.numHourlyValues,
+                                                                     item.MinutesPerItem,
+                                                                     item.rowsToSkip,
+                                                                     item.FileIntervalInterpolated,
+                                                                     item.fileName,
+                                                                     item.columnSeperator,
+                                                                     item.ScheduleTypePtr,
+                                                                     item.SchNum,
+                                                                     vals);
```
jmythms, line 1667: The main purpose of this code is 

1. To pre-allocate the very large vectors in columnarData using the reserve method. Calling reserve earlier in the code to improve performance. The other option would have been to allocate the vals vector using push back, which decreases performance. 

2. To connect allIdfSchedData and columnarData: using the setColumnarDataIndex(). This index is used later to create a map between the column of interest in the CSV and the element index of columnarData. 

The reason for this: so I could iterate through the CSV files while knowing which column of the CSV file corresponds to which column of columnarData.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+                std::adjacent_find(vectorOfCSVColNumbers.begin(), vectorOfCSVColNumbers.end()); // what if there are multiple last elements?
+            if (duplicate !=
+                vectorOfCSVColNumbers
+                    .end()) { // TODO : what if there are multiple last elements? - remove last element and check if duplicate is still the same
+                ShowFatalError(state,
+                               format("Multiple schedules pointing to the same column number : {} in file {}. Please modify the IDF file and make "
+                                      "sure multiple schedules do not point to the same column in the file.",
+                                      *duplicate,
+                                      fileNameItem));
+            }
```
jmythms: Having multiple schedules pointing to the same column in a CSV creates a many-to-one mapping, which breaks the code here. So I set up a check and a fatal error if this was happening. The only reason I see someone wanting this would be if they wanted to stack multiple schedules in one CSV column, using the "rows to skip" field.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
-                ShowSevereError(state,
-                                RoutineName + CurrentModuleObject + "=\"" + Alphas(1) + "Invalid value for \"" + cAlphaFields(5) + "\" field=\"" +
-                                    Alphas(5) + "\"");
-                ErrorsFound = true;
-            } else if (Alphas(5) != "YES") { // No validation done on the value of the interpolation field
-                FileIntervalInterpolated = false;
-            } else {
-                FileIntervalInterpolated = true;
-            }
+                    state.dataScheduleMgr->columnarData[colNumToColDataIndex[colNum]].vals.emplace_back(columnValue); // row[colNum]
```
jmythms: Go through every row of the file, 
For every column number that we know are important,
  if it needs to be skipped, skip it.
Or else, process the token or cell to a Real64 value and 
put it in the appropriate vals vector.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+        for (const PreProcessedColumn &schedule : state.dataScheduleMgr->columnarData) {
+            rowLimitCount = (schedule.numHourlyValues * 60) / schedule.MinutesPerItem;
+            if (schedule.numerrors > 0) {
+                ShowWarningError(state,
+                                 format("{}{}=\"{}\" {} records had errors - these values are set to 0.",
+                                        RoutineName,
+                                        CurrentModuleObject,
+                                        schedule.name,
+                                        schedule.numerrors));
+                ShowContinueError(state, "Use Output:Diagnostics,DisplayExtraWarnings; to see individual records in error.");
```
jmythms, line 1761: Warnings if there were issues with processing string to Real64.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+                                        RoutineName,
+                                        CurrentModuleObject,
+                                        schedule.name,
+                                        schedule.numHourlyValues));
+                ShowContinueError(state,
+                                  format("..Specified Number of Hourly Values={} Actual number of hourly values included={}",
+                                         schedule.numHourlyValues,
+                                         (rowCnt * 60) / MinutesPerItem));
+            }
+        }
```
jmythms, line 1775: Warnings if the number of rows specified in IDF is not equal to the number of rows processed.

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -1641,354 +1636,229 @@ namespace ScheduleManager {
...
+                    if (schedule.FileIntervalInterpolated) {
+                        for (int Hr = 1; Hr <= 24; ++Hr) {
+                            int minuteEnd = schedule.MinutesPerItem;
+                            SCount = 1;
+                            CurMinute = state.dataGlobal->MinutesPerTimeStep;
+                            for (iRowNum = 1; iRowNum <= hrLimitCount; ++iRowNum) {
+                                MinuteValue({SCount, minuteEnd}, Hr) = schedule.vals[rowCountFld];
+                                SCount = minuteEnd + 1;
+                                minuteEnd += schedule.MinutesPerItem;
+                                ++rowCountFld;
```
jmythms: Same as the previous code, except that I combined the loops that were iterating between for (int Hr = 1; Hr <= 24; ++Hr).

src/EnergyPlus/ScheduleManager.cc
mjwitte: Would it be difficult to let two schedule names share the same data column? I could see doing that intentionally as a base case and then changing the column for one schedule but not another in another run. It's not a huge deal, the user could simply duplicate a column as needed.

jmythms: Thanks for the review!
I spent a bit of time thinking about how to implement this, and also debating whether the extra code complexity was worth the feature. I will add it to the TODO list :)

mitchute:
There's one warning on Mac that I'm seeing locally that CI isn't reporting.

```
/Users/mmitchel/Projects/EnergyPlus/dev/EnergyPlus/third_party/ObjexxFCL/src/ObjexxFCL/string.functions.cc:481:14: warning: returning address of local temporary object [-Wreturn-stack-address]
      return std::string(); // Return empty string
             ^~~~~~~~~~~~~
```

I'm not sure what to make of the other warnings on the character separator constexpr's (e.g. ```CharSemicolon```) being unused. Those look like they *are* used in at least one location.

jmythms:
Any suggestions on how to have/debug a Mac build without having a Mac? 😅

Myoldmopar:
You shouldn't need a Mac, just Clang.  You can likely investigate this with Compiler Explorer.

I've started a sample [here](https://godbolt.org/z/16dMoP3eY) with that function, but I haven't picked out which version of Clang is issuing the warning.  Or whatever.

For the record, I'm not sure why the compiler thinks that is returning the address of the string.  It should be making a copy of the newly created string.  To be fair though I haven't investigated farther than I have described in this comment.  If I took a few more minutes I could figure out what it is doing.

mitchute:
All it takes is 💵  💵  💵 .

Maybe it has to do with the fact that you're returning a ```std::string``` but the new function has a ```std::string_view``` return type?

src/EnergyPlus/ScheduleManager.cc
```diff
@@ -231,7 +231,7 @@ namespace ScheduleManager {
         Array1D_bool AllDays(MaxDayTypes);
         Array1D_bool TheseDays(MaxDayTypes);
         bool ErrorHere;
-        int SchNum;
+        int SchNum{};
```
mitchute, line 232: I'm curious why this is necessary if you're not actually initializing to a value?

jmythms:
> All it takes is    .
> 
> Maybe it has to do with the fact that you're returning a `std::string` but the new function has a `std::string_view` return type?

I will check it out.

src/EnergyPlus/ScheduleManager.cc
jmythms: It would automatically initialize it to 0 if I used the `{}`. I will change it to `{0}` for clarity 👍🏽

Myoldmopar:
Yup, that's it.

https://godbolt.org/z/TqfqWWoTz

@amirroth has requested that we make a pass through the code to add string_view to places which may involve adding specialized functions for string_view as an alternate to the existing Objexx methods.  I believe this is one such case.  Perhaps for the sake of this PR, we can avoid going to string_view right now but use this as a first example use to dig in to string_view.  Or maybe in this case, string_view is not actually the right thing to use and it really should be a string itself here.  I've still got to learn a little bit more to be able to answer that without looking at the documentation at the same time.

src/EnergyPlus/ScheduleManager.cc
mitchute: I see. Thanks for clarifying.

jmythms:
Thanks for figuring it out so quickly. I will take out the string_view stuff for this PR 👍🏽 I did learn so much from it though!

Myoldmopar:
@jmythms what's the status of this now?  I see it is still a draft PR, so I won't dig in much, but at a minimum, CI is complaining about some unused variables.  Let me know if you need anything from me here.  This'll be a good one though!

nrel-bot-3:
@jmythms @lgentile it has been 28 days since this pull request was last updated.

jmythms:
#### Closing this PR:
The problems intended to be fixed in this PR have already been addressed in PR#[8996](https://github.com/NREL/EnergyPlus/pull/8996) and is part of Develop. 
- Reducing repeating the file open-close operations have been [fixed](https://github.com/NREL/EnergyPlus/blob/b448a8250a1db9995f7f7f91cfe686e85f956a3f/src/EnergyPlus/ScheduleManager.cc#L473).
- Parsing CSV files line by line have also been [implemented](https://github.com/NREL/EnergyPlus/blob/b448a8250a1db9995f7f7f91cfe686e85f956a3f/src/EnergyPlus/InputProcessing/CsvParser.cc#L205). 

Running the [HybridZoneModel_shadow113.idf](testfiles/HybridZoneModel_shadow113.idf) file through the Clion Profiler, we see the CPU time spent on ProcessScheduleInput() has reduced from 
73.98%  - 7415d8224 (Develop at the time of last update of this PR). 
19%  - This PR.
10% - b448a825 (Current Develop, which also includes additional improvements to ProcessNumber function, etc).

As the identified problems have already been solved, and reading the file is just taking ~10% CPU time of the simulation process for a 'large' CSV file, I propose to close this PR. 

The small advantage I see from continuing this work is reducing the dependency of E+ on third-party libraries like nlohmann. 
Since nlohmann is already used in the input processor since 2019, and the nlohmann license grants permission to modify it if necessary, I think it is too minor an advantage to pursue. 

Open to discussion, if necessary. 

@Myoldmopar 
image

PR #8809, opened 2021-06-08
Adjustment to View Factor Fix for Surfaces that are Nearly Half of Zone Surface Area

### LINKED ISSUE #8700: Internal Mass is affected by the Building North Angle Axis variation
Issue overview
--------------
Imagine a single thermal zone building (e.g., a cube) and we calculate the total annual energy consumption using a HVAC Template Zone Ideal Load Air System for every angle around 360° of the building north axis. If we plot the graph ([Fig1_Without_Internal_Mass](https://www.dropbox.com/s/m1vke18jvx3a3cw/Fig1_Without_Internal_Mass.png?dl=0)) we will have a smooth curve. However, if we add Internal Mass to the zone, the curve gets strangely disrupted ([Fig2_With_Internal_Mass](https://www.dropbox.com/s/zuvi5wxfh7f15cc/Fig2_With_Internal_Mass.png?dl=0)). This is awkward as the mass within a zone should not create such change when varying just one degree in the Building north.

### Details
Some additional details for this issue:
 - macOS Big Sur 11.2.3
 - EnergyPlus 9.5.0

### Checklist
 - [X] [IDF without Internal Mass](https://www.dropbox.com/s/epluk451s0j46c6/ESP_Toledo.082720_SWEC_hexahedron_WithoutInternalMass.idf?dl=0)
 - [X] [IDF with Internal Mass](https://www.dropbox.com/s/uh3phv47e7rlzl2/ESP_Toledo.082720_SWEC_hexahedron_WithInternalMass.idf?dl=0)

RKStrand:
Pull request overview
---------------------
 - Fixes #8700 
 - The original problem was demonstrated by a user who ran a fairly simple single zone run with a large amount of internal mass at integer values of North Axis that varied from 0 to 359 (literally, north axis = 0, 1, 2, 3, 4, ...359).  Upon graphing the results of these cases, the user noted that the results were not a nice smooth curve but rather had many discontinuities in it.  After testing and tracking down a warning message in the error file, it was determined that the problem was actually stemming from a problem in the view factor calculations.  The specification of an internal mass that had a surface area that was just below by a tiny amount the surface area of the remaining surfaces in the zone the correction for such cases was not done.  As a result, the view factors calculations did not successfully iterate to completion and this resulted in the odd variations in cooling load.  The fix was to allow the view factor correction for such extreme cases to be done when the largest area is larger than 0.99 times the remaining zone surface area.  In addition, several other minor code and documentation issues were also addressed.  The revised code now runs the user file successfully to completion and no longer exhibits the strangely fluxuating results as north axis varies.
 - Adds reporting of InternalMass surfaces to the Envelope Summary, Opaque Interior subtable. Previously internal mass was omitted from this report.

Audit diffs are due to additional table entries for internal mass.

### Pull Request Author
 - [X] Pull requests that impact EnergyPlus code must also include unit tests to cover enhancement or defect repair
 - [X] If any diffs are expected, author must demonstrate they are justified using plots and descriptions (see above)
 - [X] If changes fix a defect, the fix should be demonstrated in plots and descriptions (see above)

### Reviewer


src/EnergyPlus/HeatBalanceIntRadExchange.cc
```diff
@@ -1710,6 +1747,19 @@ namespace HeatBalanceIntRadExchange {
...
             }
         }
+        if (severeErrorPresent) {
+            ShowFatalError(state,
+                           "FixViewFactors: View factor calculations significantly out "
+                           "of tolerance.  See above messages for more information.");
+        }
+    }
+
+    bool DoesZoneHaveInternalMass(EnergyPlusData &state, int const numZoneSurfaces, const Array1D_int &surfPointer)
```
mjwitte, line 1762: `state` is unused here.
Also, minor detail, but `numZoneSurfaces` isn't necessary. The loop statement could be ` for (int surfNum : surfPointer)`.
If I'm following correctly, this only happens once per enclosure? So, there's no concern about repeating this, correct?

src/EnergyPlus/HeatBalanceIntRadExchange.cc
```diff
@@ -1543,9 +1552,20 @@ namespace HeatBalanceIntRadExchange {
...
+        // calculated using the standard procedure below will not converge and may result in invalid
+        // view factors where either reciprocity or completeness is not satisfied.  However, when
+        // the largest surface is just slightly smaller than the rest of the surface areas in the
+        // zone, it has been shown that there can still be problems.  The correction below can
+        // be helpful in avoiding these problems.  So, the criteria below (with the 0.9 term added
+        // into the comparison in the next line of code) intends to capture more cases that are
+        // "unbalanced" in surface area distribution so other strange cases can take advantage of
+        // this correction.  The use of 0.9 is simply to provide some reasonable boundary numerically
+        // and does not have some derived theoretical basis.
+        if (LargestArea > 0.9 * (sum(A) - LargestArea) && (N > 3)) {
```
mjwitte: In the diff comments, you suggest the the files with diffs have a surface that is triggering this rule, but I'm not seeing anything like that in 5ZoneSupRetPlenRAB. I see a couple of large surfaces (TOP-1 and BOTTOM-1) but these should be balanced by an equal area of floor/ceiling surfaces, plus the wall surfaces. Even in the plenum zones I'm puzzled why results would change. I haven't looked at any of the other files with diffs. Is it possible some other change is driving these diffs?

src/EnergyPlus/HeatBalanceIntRadExchange.cc
```diff
@@ -1676,22 +1696,39 @@ namespace HeatBalanceIntRadExchange {
...
                                              RowSum));
-                    ShowContinueError(state, "If zone is unusual, or tolerance is on the order of 0.001, view factors are probably OK.");
+                    ShowContinueError(state,
+                                      "If zone is unusual or tolerance is on the order of 0.001, view "
+                                      "factors might be OK but results should be checked carefully.");
+                    if (anyIntMassInZone) {
+                        ShowContinueError(state,
+                                          "For zones with internal mass like this one, this"
+                                          "can happen when the internal mass has an area that"
+                                          "is much larger than the other surfaces in the zone.");
```
mjwitte, line 1725: Would it be appropriate to suggest splitting the internal mass into two or more surfaces?

src/EnergyPlus/HeatBalanceIntRadExchange.cc
mjwitte: Oh, wait, I see what's going on, I think. For a plenum zone, say there are 2 large surfaces of 100m2 each and 4 small ones that add up to 10m2. So this would be if (100 > 0.9*(210-100)=99 which trips the rule. But I don't see why that isn't a valid enclosure as-is.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
RKStrand: Yes, should happen once (at most twice--not exactly sure why it's coming through twice sometimes?) per enclosure.  So, basically, get rid of state and numZoneSurfaces and adjust loop to the suggestion.  Will do.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
RKStrand: Yeah, there's some question as to what is the limit of "valid".  The old rule was a comparison without the 0.9.  Maybe 0.9 is not great either.  What about 0.99?  The problem is that if a user gets close but not exactly there, the file could run into problems like was seen in this defect.  How do you feel about 0.99?  It would probably make most of the diffs from develop go away.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
RKStrand: I will adjust the error message to suggest splitting the thermal mass into two or more surfaces.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
rraustad: "When one surface has an area that exceeds the sum of all other surface areas in a zone".

I assume the surfaces in question here are the zone floor, walls and ceiling. If so, then the largest surface's area should not be greater than 1/2 the total surface area. I can't think of a geometry where that would be the case unless the walls aren't vertical but I assume someone put this here for a reason. Not even sure if a unit test could sort this out.

mjwitte: So perhaps a better solution to the numerical problem in the users file would be to use a small absolute tolerance, like
` if (LargestArea > (sum(A) - LargestArea - somesmallvalue)`

rraustad: @RKStrand are the walls in the defect file vertical? What is the specific geometry that trips this limit?

rraustad: Even a pyramid wouldn't trip this limit. Are walls missing?

RKStrand: The zone in question here is a simple and fairly normal box (4 vertical walls, roof, floor) with one exception--it has an internal mass surface with an area equal to just slightly below the total area of the remaining surfaces.  As a result the "correction" in the current code gets skipped and the view factor calculations never iterate to completion.  This resulted in weird trends in the results.  I agree that it is not possible to have a valid geometry (i.e., a full enclosure) where one surface has an area larger than the remaining surface areas--it's no longer a full enclosure.  That's why the correction is there to avoid numerical issues when you are trying to figure out view factors for an incomplete enclosure.  In the user file that resulted in the defect, it just missed the cutoff.

RKStrand: @mjwitte I can do the small absolute tolerance instead of the percentage.  I think I was thinking that because everything is scaled to area that a fraction would be better.  Of course in either case, we have to establish either what the small value or the small fraction is.  I could go to 0.99 or 0.999 as a fraction.  I could also go to say 1m2 as a small tolerance.  Which do you prefer?  I'm good with any of those.

mjwitte: I don't have a strong opinion, just use something reasonable that fixes the defect file without tripping 5ZoneSupRetPlenRAB.

src/EnergyPlus/HeatBalanceIntRadExchange.cc
RKStrand: You sure I don't need state?  It's used in the line:

if (**_state_**.dataSurface->Surface(surfPointer(surfNum)).Class == SurfaceClass::IntMass)...

My compiler (after getting rid of state) is unhappy and I've tried a variety of variations.  Of course, maybe I just don't understand "state"...

mjwitte: Oh, sorry, missed that.

RKStrand: The `for (int surfNum : surfPointer)` didn't work...this happens once per zone.  The second time it comes into this routine (for the second zone), the counter picks up where it left off and then I get a crash.  In the file that I was running, there were 9 surfaces in the first zone and six in the second.  The first time through there weren't any problems.  The second time through for zone 2 the counter starts at 10 which is more than the number of surfaces in the zone and I get a pointer issue.  I think I need to go back to what I originally had?

RKStrand:
@mjwitte I addressed the internal mass error messaging issue and adjusted the fraction from 0.9 to 0.99.  Hopefully this resolves everything and the ci tests come back better than before.  Let me know if you still want me to adjust the passed variables to the new subroutine (see note above--I couldn't get your suggestion to work).  Thanks!

src/EnergyPlus/HeatBalanceIntRadExchange.cc
mjwitte: It's fine as-is, but for future reference, there's no counter with the suggested form, so this should work fine for multiple calls.
```
for (int surfNum : surfPointer){
 if (state.dataSurface->Surface(surfNum).Class == SurfaceClass::IntMass) return true;
}
```

RKStrand:
Not sure why it is failing on "AirTerminalSingleDuctMixer_SimFCU_ATMInletSideTest".  It runs on my machine just fine.  Good news is that the 0.99 gets rid of all the other differences.  Somehow GitHub didn't push one minor change to a comment line up so I'm doing that now...

mjwitte (approved):
@RKStrand Looks good. Merging.

